PR #1, opened 2026-07-14
history-service: parallel empty-bucket skip for paginated reads

hmchangw:
Paginated history reads walk Cassandra buckets one synchronous round-trip
at a time (fillPage). For sparse rooms or deep lookback this is N sequential
RTTs to fill a single page — the dominant read-latency cost, and a
connection-pool starvation risk under concurrency.

Decouple the walk orchestration from the gocql fetch so it is unit-testable,
then add an opt-in parallel mode: after a run of consecutive empty buckets a
read probes several buckets concurrently to skip the run. The parallel path
only ever commits to skipping EMPTY buckets — the first bucket that yields
rows is handed back to the serial path, so page contents and the resume
cursor always come from a single exact-limit read. This keeps parallel output
byte-identical to serial (proven by a differential test) while collapsing long
empty-bucket walks into ~N/concurrency rounds.

- walker.go: extract walkBuckets (serial + escalating-parallel) over a
  bucketFetcher abstraction; fillPage becomes the gocql adapter, behavior
  unchanged at concurrency=1.
- repository.go: WithReadConcurrency functional option; serial remains the
  default so existing callers are unaffected.
- config: MESSAGE_READ_BUCKET_CONCURRENCY (default 1 = off) and
  MESSAGE_READ_ESCALATE_AFTER (default 4), validated and logged at startup.
- tests: serial-vs-parallel differential (both directions, varied density and
  page size), escalation/hand-back, maxBuckets cap, floor termination, and a
  barrier test proving the batch fetches concurrently.

Co-Authored-By: Claude Opus 4.8 
Claude-Session: https://claude.ai/code/session_01LEUfLtmkFZr3EzjiVWCVPJ

### DIFF
--- a/history-service/internal/cassrepo/walker.go
+++ b/history-service/internal/cassrepo/walker.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 
 	"github.com/gocql/gocql"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/hmchangw/chat/pkg/msgbucket"
 )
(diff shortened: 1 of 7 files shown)


julianshen:
**Actionable comments posted: 3** — 🔴 1 blocking · 🟠 2 warning

<details>

* `history-service/internal/cassrepo/messages_by_room.go`
* `history-service/internal/cassrepo/repository.go`
* `history-service/internal/cassrepo/walker.go`
* `history-service/internal/cassrepo/walker_parallel_test.go`

</details>
<details>

* `opencode-review`

</details>

history-service/internal/cassrepo/walker.go
```diff
@@ -85,98 +86,168 @@ func (r pageResult[T]) toPage() Page[T] {
...
+				g.Go(func() error {
+					// firstBucket is always false here: escalation needs >=1 prior
+					// empty bucket, so walked>0 and this is never the first step.
+					res, err := fetch(gctx, batch[i], false, nil, pageSize)
+					if err != nil {
+						return err
+					}
+					results[i] = res
+					return nil
+				})
```
julianshen, line 191: _🎯 correctness_ | _🔴 Blocking_ | _`opencode-review`_

**Any error from a speculative bucket aborts the entire walk before results are examined.**

For example, if the first probed bucket contains enough data but a later bucket has a Cassandra/decryption error, serial mode returns the page while parallel mode fails; errgroup cancellation can also cancel the earlier query. Collect errors per index without shared error cancellation and process results in traversal order, ignoring errors after the first data bucket. Add a differential test covering data in an earlier bucket and an error in a later speculative bucket.

<details>

```
In `history-service/internal/cassrepo/walker.go` around line 191: Any error from a speculative bucket aborts the entire walk before results are examined. For example, if the first probed bucket contains enough data but a later bucket has a Cassandra/decryption error, serial mode returns the page while parallel mode fails; errgroup cancellation can also cancel the earlier query. Collect errors per index without shared error cancellation and process results in traversal order, ignoring errors after the first data bucket. Add a differential test covering data in an earlier bucket and an error in a later speculative bucket.

```

</details>

history-service/internal/cassrepo/walker.go
```diff
@@ -85,98 +86,168 @@ func (r pageResult[T]) toPage() Page[T] {
...
-		// accumulated rows and abort the walk so a later bucket can't overwrite
-		// the error or serve a partial page.
-		if scanErr != nil {
-			return pageResult[T]{}, fmt.Errorf("scan bucket %d: %w", bucket, scanErr)
-		}
+			results := make([]bucketResult[T], len(batch))
+			g, gctx := errgroup.WithContext(ctx)
+			for i := range batch {
+				g.Go(func() error {
+					// firstBucket is always false here: escalation needs >=1 prior
```
julianshen, line 183: _⚡ performance_ | _🟠 Warning_ | _`opencode-review`_

Each speculative probe reads and scans up to pageSize rows even though it only needs to determine whether the bucket is empty. This can amplify Cassandra reads and decryption work by concurrency*pageSize, after which the first data bucket is read again. Probe with limit 1 (then re-read the selected bucket using the precise remaining limit); this also avoids surfacing errors from rows that the current serial page would not reach.

<details>

```
In `history-service/internal/cassrepo/walker.go` around line 183: Each speculative probe reads and scans up to pageSize rows even though it only needs to determine whether the bucket is empty. This can amplify Cassandra reads and decryption work by concurrency*pageSize, after which the first data bucket is read again. Probe with limit 1 (then re-read the selected bucket using the precise remaining limit); this also avoids surfacing errors from rows that the current serial page would not reach.

```

</details>

history-service/internal/cassrepo/walker.go
```diff
@@ -85,98 +86,168 @@ func (r pageResult[T]) toPage() Page[T] {
...
 			return pageResult[T]{Rows: out, NextCursor: "", HasNext: false}, nil
 		}
 
-		q := queryFn(bucket, walked == 0).WithContext(ctx)
-		q = q.PageSize(pageSize - len(out))
-		if pageState != nil {
-			q = q.PageState(pageState)
-		}
+		if parallel {
+			// Build a batch of buckets to probe concurrently, bounded by the
```
julianshen, line 169: _🏷️ reliability_ | _🟠 Warning_ | _`opencode-review`_

**cfg.concurrency is used directly as the slice capacity and has no configured upper bound.**

A large environment value can cause excessive allocation or OOM even when maxBuckets permits only a small batch. Bound the capacity/concurrency by the remaining bucket budget and enforce a reasonable configuration maximum.

<details>

```
In `history-service/internal/cassrepo/walker.go` around line 169: cfg.concurrency is used directly as the slice capacity and has no configured upper bound. A large environment value can cause excessive allocation or OOM even when maxBuckets permits only a small batch. Bound the capacity/concurrency by the remaining bucket budget and enforce a reasonable configuration maximum.

```

</details>

julianshen:

<details>

* `history-service/internal/cassrepo/messages_by_room.go`
* `history-service/internal/cassrepo/repository.go`
* `history-service/internal/cassrepo/walker.go`
* `history-service/internal/cassrepo/walker_parallel_test.go`

</details>
<details>

* `opencode-review`

</details>

history-service/internal/cassrepo/walker.go
```diff
@@ -85,98 +86,168 @@ func (r pageResult[T]) toPage() Page[T] {
...
+			results := make([]bucketResult[T], len(batch))
+			g, gctx := errgroup.WithContext(ctx)
+			for i := range batch {
+				g.Go(func() error {
+					// firstBucket is always false here: escalation needs >=1 prior
+					// empty bucket, so walked>0 and this is never the first step.
+					res, err := fetch(gctx, batch[i], false, nil, pageSize)
+					if err != nil {
+						return err
+					}
```
julianshen, line 188: _🎯 correctness_ | _🔴 Blocking_ | _`opencode-review`_

**The parallel batch returns any speculative fetch error before checking which bucket first contained data.**

If an earlier bucket has enough rows to fill the page but a later concurrently probed bucket fails (for example, transient Cassandra or decrypt error), this request fails even though the serial walker would return the earlier page without querying that later bucket. Preserve per-index errors and resolve results in traversal order: return an error only if it precedes the first non-empty bucket; otherwise hand that bucket back to the serial path and ignore/cancel later speculative work. Add a differential test with a full nearer bucket and an error in a farther bucket.

<details>

```
In `history-service/internal/cassrepo/walker.go` around line 188: The parallel batch returns any speculative fetch error before checking which bucket first contained data. If an earlier bucket has enough rows to fill the page but a later concurrently probed bucket fails (for example, transient Cassandra or decrypt error), this request fails even though the serial walker would return the earlier page without querying that later bucket. Preserve per-index errors and resolve results in traversal order: return an error only if it precedes the first non-empty bucket; otherwise hand that bucket back to the serial path and ignore/cancel later speculative work. Add a differential test with a full nearer bucket and an error in a farther bucket.

```

</details>

hmchangw:
Closing in favor of #169, which supersedes this work.

Both PRs target the same bottleneck — paginated history reads walk Cassandra buckets one serial round-trip at a time, so sparse/deep ranges pay N sequential RTTs — and both use the same core refactor: extract a `bucketFetcher` seam and a `walkBuckets` traversal, proven byte-identical to serial via a differential/parity test.

They differ in the concurrency strategy:

- **This PR (#1)** escalates to a parallel **empty-bucket skip**: it stays serial until it hits a run of empty buckets, then probes ahead concurrently and hands the first data bucket back to the serial path. Because escalation only triggers on consecutive *empty* buckets, it doesn't speed up sparse-but-active rooms (few rows per bucket, no long empty runs) — a common shape.
- **#169** fans out over *all* buckets in waves and sizes each wave to the observed rows-per-bucket density (wide over empty/sparse runs, narrow over dense ones), so sparse-active rooms benefit too.

#169 is the fuller approach and has been carried forward with the adaptive widening plus a cleanup pass, so the two would also directly conflict if both landed.

Two good ideas from this PR that informed the decision and are worth carrying into #169 as a follow-up: the **opt-in config gate with a serial default** (#169 is currently always-on), and the **differential-test** methodology. Thanks for both.

---

PR #2, opened 2026-07-14
fix(room-worker): write individual room_members whenever the table has rows, not only for orgs (#499)

hmchangw:
## Summary

`member.list` (`room-service` `ListRoomMembers`) reads the `room_members` collection whenever it has **any** row, else falls back to `subscriptions`. But the add path (`room-worker` `processAddMembers`) only wrote a `room_members` row when `writeIndividuals` was true, which gated on `HasOrgRoomMembers` (**org** rows only). So in a room whose `room_members` has individual rows but no org rows (an org was added and later removed, leaving the individual rows), a **direct** member add wrote only a subscription — so the new member was invisible to `member.list`, even though the `member.add` event still fired.

Fixes #499.

## Changes

- `room-worker/store.go` + `store_mongo.go`: add `HasAnyRoomMembers(roomID)` — `CountDocuments({rid}) &gt; 0` (mirrors `HasOrgRoomMembers` without the `member.type` filter).
- `room-worker/handler.go`: `loadAddMemberInputs` now also loads `hasAnyRoomMembers` (fail-closed, concurrent with the other reads); `writeIndividuals` gates on `len(Orgs) &gt; 0 || hasAnyRoomMembers` instead of `hadOrgsBefore`. This aligns the write-gate with `member.list`&#39;s read-source. The first-org backfill still gates on `hadOrgsBefore` (org rows) — unchanged.
- Tests: a failing-first unit test (a direct add to a room whose `room_members` has individual-only rows now writes an individual row) + a lite-mode regression guard (an empty table stays subscription-only); existing add-member expectations updated for the added read.

## Verification

- Unit: the new test is red before the gate change and green after; full `room-worker` + `room-service` tests pass; integration-tagged vet + gofmt clean.
- Dev-stack e2e: reproduced the org-removed state, direct-added a member, and confirmed `member.list` now returns them (details below on request).

## Summary by CodeRabbit

- **Bug Fixes**
  - Direct member additions now continue writing individual membership records whenever a room already has any member records (not just org-based ones).
  - Preserved subscription-only behavior when the room has no existing member records.
  - Backfill behavior for first org additions now triggers based on whether the room-members table is empty, with improved closed-on-error handling and clearer error context.

### DIFF
--- a/room-worker/handler.go
+++ b/room-worker/handler.go
@@ -713,12 +713,11 @@ func (h *Handler) processRemoveOrg(ctx context.Context, req *model.RemoveMemberR
 type addMemberInputs struct {
 	room              *model.Room
 	candidates        []AddMemberCandidate
-	hadOrgsBefore     bool
 	hasAnyRoomMembers bool
 }
 
-// loadAddMemberInputs runs GetRoomMeta, ListAddMemberCandidates,
-// HasOrgRoomMembers, and HasAnyRoomMembers concurrently, collapsing the serial
+// loadAddMemberInputs runs GetRoomMeta, ListAddMemberCandidates, and
+// HasAnyRoomMembers concurrently, collapsing the serial
 // Mongo round trips into one. A plain errgroup.Group (not WithContext) is used deliberately: these are
 // independent reads, so a failure in one need not cancel the others — matching
 // the prior serial code, which returned the first error without cancellation.
@@ -744,15 +743,6 @@ func (h *Handler) loadAddMemberInputs(ctx context.Context, req *model.AddMembers
 			return fmt.Errorf("list add-member candidates: %w", err)
 		}
 		out.candidates = candidates
-		return nil
-	})
-	g.Go(func() error {
-		// Fail closed: defaulting hadOrgsBefore=false on error would trigger spurious first-org backfill.
-		hadOrgsBefore, err := h.store.HasOrgRoomMembers(ctx, req.RoomID)
-		if err != nil {
-			return fmt.Errorf("check existing org room members: %w", err)
-		}
-		out.hadOrgsBefore = hadOrgsBefore
 		return nil
 	})
 	g.Go(func() error {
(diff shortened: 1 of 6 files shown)


julianshen:

<details>

* `room-worker/handler.go`
* `room-worker/loadinputs_test.go`
* `room-worker/mock_store_test.go`
* `room-worker/store.go`

</details>
<details>

* `opencode-review`

</details>

room-worker/handler.go
```diff
@@ -940,13 +940,13 @@ func (h *Handler) processAddMembers(ctx context.Context, data []byte) (err error
...
-	// subs are already written so needSub is empty, hadOrgsBefore stays false
-	// until BulkCreateRoomMembers commits, and the backfill re-runs cleanly.
-	if len(req.Orgs) > 0 && !hadOrgsBefore {
+	// Backfill existing subscribers into room_members on the first org add,
+	// i.e. while the table is still empty — that's when individual tracking begins.
+	// Errors propagate: log-and-continue would silently corrupt room_members
+	// (existing subs would never get IRM rows). Retry is safe — subs are already
+	// written so needSub is empty, the table stays empty until BulkCreateRoomMembers
+	// commits, and the backfill re-runs cleanly.
+	if len(req.Orgs) > 0 && !inputs.hasAnyRoomMembers {
```
julianshen, line 949: _🎯 correctness_ | _🔴 Blocking_ | _`opencode-review`_

**Using `!inputs.hasAnyRoomMembers` as the backfill gate makes retry recovery unsafe.**

`BulkCreateRoomMembers` uses unordered `InsertMany`, which can persist only some rows before returning an error; a retry then sees any partially written individual row, skips backfill, and permanently omits other existing subscribers from `room_members`, which `member.list` now treats as authoritative. Keep `HasAnyRoomMembers` for `writeIndividuals`, but retain a separate org-transition/backfill signal (or explicitly reconcile missing subscriber rows) rather than assuming the table stays empty until the bulk write commits.

<details>

```
In `room-worker/handler.go` around line 949: Using `!inputs.hasAnyRoomMembers` as the backfill gate makes retry recovery unsafe. `BulkCreateRoomMembers` uses unordered `InsertMany`, which can persist only some rows before returning an error; a retry then sees any partially written individual row, skips backfill, and permanently omits other existing subscribers from `room_members`, which `member.list` now treats as authoritative. Keep `HasAnyRoomMembers` for `writeIndividuals`, but retain a separate org-transition/backfill signal (or explicitly reconcile missing subscriber rows) rather than assuming the table stays empty until the bulk write commits.

```

</details>

mliu33 (approved):
Thanks

PR #72, opened 2026-07-15
feat(room-worker): member_added room event carries members in member.list format

amdla:
## Problem

`member.list` (enrich=true) returns fully-structured entries — `sectName`/`employeeId` on individuals, `orgName`/`orgCode`/`memberCount`/`orgDescription` on orgs — but the real-time `member_added` room event only carried bare account strings. A client keeping its member panel live couldn't splice new members into the list it renders from the RPC; it had to refetch `member.list` after every add, and org adds carried no org entry at all.

## Change

The room-scoped `chat.room.{roomID}.event.member` event now carries a `members` array in the `member.list` (enrich=true) display shape, and fires exactly when the room's member list actually changes.

**Event payload (`pkg/model.MemberAddEvent`)**
- Gains `Members []RoomMemberEntry` — the display payload only, no membership `id`/`rid`/`ts` envelope. `members` echoes exactly the requested orgs and users: one org entry per requested org (`orgName`, `orgCode`, `memberCount`, `orgDescription`), then one individual entry per requested user that was newly subscribed **or** upgraded (`engName`, `chineseName`, `sectName`, `employeeId`). Org-expanded accounts are represented by their org entry, not listed individually.
- `Accounts` is now `omitempty` and **stripped from the client-facing copy** — the frontend renders from `members`. It is kept on the internal copies that need it: the same-site INBOX search feed and the cross-site OUTBOX copies (which carry no `members` and re-resolve display data at the destination), because `inbox-worker` creates destination subscriptions and `search-sync-worker` indexes by account.

**Fire semantics (`room-worker/processAddMembers`)** — fires when member.list composition changes, and no more:
- a new subscription joins;
- an existing org member is **upgraded** to an individual membership — fires carrying that individual, with no new subscription and no `members_added` system message;
- a **genuinely new org** is added (a new pre-write `ExistingOrgMembers` read distinguishes a new org from a re-add);
- **re-adding an already-present org is a no-op** — no event.

**`room_members` writes**: `BulkCreateRoomMembers` upserts each row on the `(rid, member.type, member.id)` unique index via `$setOnInsert`, so a re-add / JetStream redelivery is an idempotent no-op.

**Shared org-display plumbing**
- `pkg/orgdisplay` (new): pure move of room-service's org-display rollup so room-service and room-worker share one implementation; adds `Code` for `orgCode` — the plain section/department name (dept-first), distinct from the possibly TC-combined `orgName`, with no orgID fallback.
- `pkg/pipelines`: gains `OrgDisplayUsers`, the shared users read behind both member.list enrichment and event enrichment.
- `pkg/userstore`: shared projection gains `sectName`.

**Docs**: `member_added` and `RoomMemberEntry` tables updated in `docs/client-api.md` and `docs/client-api/events.md`.

## Error semantics

The org display and org-existence reads run with the other up-front reads in `loadAddMemberInputs`, **before any write** — a transient failure Naks a still-clean delivery. `buildAddedMembers` is a pure function.

## Testing

- `pkg/model` round-trip/omitempty; `pkg/orgdisplay` unit (`Build`/`Name`/`Code`/`Description`); handler tests for direct adds, org+direct adds, org→individual upgrade (fires with the individual, no sys-msg), re-add no-op, new-org-with-no-new-members, orgID fallback, the write-gate × enrichment cross-product, and per-lane payload shape (frontend strips `accounts` and carries `members`; INBOX/OUTBOX keep `accounts` and omit `members`); `ExistingOrgMembers` integration; the end-to-end enrichment/federation integration test.
- `make lint`: 0 issues. `make test` race suite green across the repo (85 packages); room-worker and room-service integration green. Mocks regenerated via `make generate`.


julianshen:
**No actionable comments.** ✅

<details>
<summary>📒 Files reviewed (19)</summary>

* `docs/client-api.md`
* `docs/client-api/events.md`
* `pkg/model/event.go`
* `pkg/model/model_test.go`
* `pkg/orgdisplay/orgdisplay.go`
* `pkg/orgdisplay/orgdisplay_test.go`
* `pkg/pipelines/integration_test.go`
* `pkg/pipelines/orgdisplay.go`
* `pkg/userstore/integration_test.go`
* `pkg/userstore/userstore.go`
* `room-service/orgdisplay.go`
* `room-service/orgdisplay_test.go`
* `room-worker/handler.go`
* `room-worker/integration_test.go`
* `room-worker/mock_store_test.go`
* `room-worker/store.go`

</details>
<details>

* `opencode-review`

</details>

docs/client-api.md
```diff
@@ -1229,6 +1229,7 @@ On `added` / `role_updated` / `mute_toggled` / `favorite_toggled` the embedded `
 | `roomName` | string | |
 | `roomType` | string | `"channel"`, `"dm"`, `"botDM"`, or `"discussion"`. Omitted when empty. |
 | `accounts` | string[] | The newly added accounts. |
+| `members` | [RoomMemberEntry](#roommemberentry)[] | The member.list-shaped display entries (the [RoomMemberEntry](#roommemberentry) payload only — no membership `id`/`rid`/`ts` envelope): org entries first (`orgName`, `orgCode`, `memberCount`, `orgDescription`), then one individual entry per newly subscribed direct add (`engName`, `chineseName`, `sectName`, `employeeId`). Unlike [List Members](#list-members) (`enrich: true`), individual entries here omit `isOwner` (new members are never owners) and `name` (bot display name). Accounts joined via org expansion appear in `accounts` only — their org entry represents them, mirroring `member.list`. |
```
mliu33: Like I commented in previous PR, we don't need "accounts" field anymore since "members" already return all data required by the frontend

docs/client-api.md
```diff
@@ -1238,7 +1239,7 @@ On `added` / `role_updated` / `mute_toggled` / `favorite_toggled` the embedded `
 A `members_added` system message also flows through the message pipeline and arrives as a `new_message` room event.
 
 > [!NOTE]
-> **No-op:** if every requested account is already a member (and no new orgs), or the add only upgrades an existing org member to an individual membership, the requester still gets an `AsyncJobResult` with `status: "ok"` but **no** `subscription.update` / `room.key` / `member_added` events follow.
+> **No-op:** if no org is requested and every requested account is already a member, or the add only upgrades an existing org member to an individual membership, the requester still gets an `AsyncJobResult` with `status: "ok"` but **no** `subscription.update` / `room.key` / `member_added` events follow. (Re-adding an org that is already present is **not** a no-op: `member_added` still fires, carrying the org's entry with `accounts: []`.)
```
mliu33: Re-adding an org that is already present should not fire the event.

docs/client-api/events.md
```diff
@@ -665,15 +665,18 @@ the event on the same room stream they already subscribe to.
...
+re-added (see the no-op note below for what does **not** fire). Triggered by
 [Add Members](request-reply.md#add-members) and indirectly by
 [Create Room](request-reply.md#create-room).
 
 A `members_added` system message also flows through the message pipeline as a
 `new_message` room event.
 
-> **No-op:** if every requested account is already a member (or the add only upgrades an
-> existing org member to individual membership), no `member_added` event fires.
+> **No-op:** if no org is requested and every requested account is already a member, or the add
```
mliu33: I just found a bug. If the add upgrades an existing org member to an individual membership, we also need to publish member_add event with members containing that individual.

Could you help fix it? Thanks a lot!

room-worker/handler.go
```diff
@@ -1078,6 +1086,10 @@ func (h *Handler) processAddMembers(ctx context.Context, data []byte) (err error
 			HistorySharedSince: historySharedSince,
 			Timestamp:          now.UnixMilli(),
 		}
+		// Marshalled BEFORE Members is attached — INBOX payloads must omit it (see model.MemberAddEvent.Members).
+		inboxPayload, _ := json.Marshal(memberAddEvt)
+
+		memberAddEvt.Members = buildAddedMembers(&req, subs, allowedIndividual, inputs.orgDisplayRows, userMap)
```
mliu33: The "Members" should only contain exactly the same org and individual requested by the user input. We don't need to include org expanded users.

For example, 

- User add "orgA + userA + userB"
- MemberAddEvent.Members = [orgA_RoomMemgerEntry, userA_RoomMemgerEntry, userB_RoomMemeberEntry]

Also, we can remove "accounts" filed from memebeAddEvent which originally contain all newly added individuals.

mliu33 (commented):
Outstanding work, some suggestions, thanks a lot!

docs/client-api.md
```diff
@@ -1220,25 +1220,25 @@ On `added` / `role_updated` / `mute_toggled` / `favorite_toggled` the embedded `
...
 
 > [!NOTE]
-> **No-op:** if every requested account is already a member (and no new orgs), or the add only upgrades an existing org member to an individual membership, the requester still gets an `AsyncJobResult` with `status: "ok"` but **no** `subscription.update` / `room.key` / `member_added` events follow.
+> **No-op:** when the request changes nothing — every requested account already subscribed, no org member upgraded to an individual membership, and every requested org already present — the requester still gets an `AsyncJobResult` with `status: "ok"` but **no** `subscription.update` / `room.key` / `member_added` events follow. In particular, **re-adding an already-present org is a no-op**. An **org→individual upgrade** (an existing org member added individually) is **not** a no-op: `member_added` fires with that individual in `members`, but no `members_added` system message is posted (no one newly joined).
```
mliu33, line 1241: Please check current old behavior if adding an existing org member as individual should fire members_added system message? I think we might still need to publish system msg for that upgrading individual member

room-worker/handler.go
```diff
@@ -1106,43 +1136,48 @@ func (h *Handler) processAddMembers(ctx context.Context, data []byte) (err error
...
-			Channels:        nonNil(req.Channels),
-			AddedUsersCount: len(subs),
-		}
-		sysMsgData, _ := json.Marshal(membersAdded)
-		seed := messageDedupSeed(ctx, "processAddMembers", req.RoomID,
-			fmt.Sprintf("%s:%s:%d", req.RoomID, req.RequesterAccount, req.Timestamp))
-		content := addedContent(requester, sysIndividuals, req.Orgs, func(a string) *model.User {
-			if u, ok := userMap[a]; ok {
-				return &u
+		// members_added sys-msg: a pure org→individual upgrade adds no new member and
```
mliu33, line 1139: Please check old behavior whether we should have system message for org->individual upgrade (I think we do) because the added individual will show up on the room member info panel in the frontend

mliu33 (approved):
Excellent work, thanks! One comment needed to check

PR #73, opened 2026-07-15
feat: room last-message preview (lastMessage) on message_deleted + encryption support

amdla:
## Summary

The mobile app shows each room's **last-message preview** in the room list (like Messenger). That server-side support either didn't exist or was incomplete: nothing ever carried a room's last-message *content*, and deleting the newest message left the sidebar pointing at a deleted message. This PR adds that support end-to-end and makes it correct under both message **edits** and **deletes**.

Behavioural contract:
- **Delete** a message → clients receive the room's **last non-deleted, non-system message** so the preview can move back (our difference from Instagram: we show the previous real message, not a "message deleted" placeholder).
- **Edit** a message → the preview shows the **newest message's current content**; editing an *old* message updates its content in place but never reorders it to the front (ordering is by `created_at`, which edits don't touch).

The "fetch the room's surviving last message" step is a new **internal RPC** on `history-service`; `broadcast-worker` calls it on delete and embeds the result in the `message_deleted` event. It also works with room **encryption** (currently off in prod during testing; this PR makes the code ready for the switch to encrypted).

### Contracts — `pkg/model`, `pkg/subject`
- **`LastMessagePreview`** (new, `pkg/model/lastmsg.go`) — trimmed preview shape shared by the delete event, the `rooms.lastMsg` denormalization, and the RPC: `{ messageId, type, senderAccount, senderName, msg | encMsg, createdAt, editedAt, attachmentCount }`. Content-only encryption via `encMsg` (a roomcrypto envelope, the *same* format clients already decrypt for `encryptedNewContent`); metadata stays plaintext. Exactly one of `msg`/`encMsg` is set.
- **`DeleteRoomEvent.lastMessage`** (`pkg/model/event.go`) — optional; the room's surviving preview after the delete. Absent on hidden `TShow=false` thread-reply deletes (the room preview can't change) and after a visible delete that empties the room (→ clear the preview).
- **`Room.lastMsg`** (`pkg/model/room.go`) — the denormalized preview persisted in Mongo.
- **`subject.MsgRoomLast`** — `chat.server.request.msg.{siteID}.room.last`, a server-to-server subject (mirrors `RoomsInfoBatch`/`ThreadSubscriptionList`); not client-facing, so it's intentionally absent from `client-api.md`.

### `history-service` — the RPC + a bug fix
- **`GetLastRoomMessage`** RPC (`internal/cassrepo/last_message.go`, `internal/service/last_message.go`): walks `messages_by_room` newest→oldest within the standard bucket-walk bounds, skipping soft-deleted rows and system-message types, and returns the first survivor (decrypting only the winning row). Uses a generous 100-row scan page so tombstone runs don't degenerate into one round-trip per row, and caps the total rows scanned per walk (the `rooms.get` ponytail precedent) so a mass-deleted tail degrades to "no preview" instead of draining a partition. Empty/all-deleted/unknown room → nil preview, not an error.
- **Bug fix — thread "last reply at" dropped on reply deletes.** `SoftDeleteMessage` already recomputed `thread_last_msg_at` from the surviving replies and persisted it, but only returned the count, so the canonical delete event left `NewThreadLastMsgAt` nil — which downstream reads as "no replies remain". It now returns and embeds the recomputed timestamp, so a live thread badge shows the correct "last reply" time after deleting the newest reply of a still-populated thread. (Pre-existing bug, adjacent to this code; fixed here to avoid a second pass over the same path.)

### `broadcast-worker` — event enrichment + Mongo maintenance + encryption
- **`LastMessageFetcher`** (`lastmsg_fetcher.go`): NATS client for the RPC, cloned from the existing `parent_fetcher.go` pattern (2s timeout, errcode envelope detection, sonic-safe decode).
- **`handleDeleted`**: **fetch-first** — resolves the survivor before publishing anything, and skips the RPC entirely when the stored `rooms.lastMsg` already identifies it (the common non-latest delete); any RPC/Mongo failure Naks and redelivers, so *every* delivered delete event carries a preview (deterministic absence). Hidden thread-reply deletes skip the fetch.
- **`rooms.lastMsg` invariant** kept consistent across the lifecycle: create stores the preview via the existing coalescer (system messages advance the pointer but never become the stored preview); delete does a **guarded rewind** keyed on the preview's identity (`lastMsg.messageId`, with a drift-state phase-2 path) to the survivor so a concurrent newer message is never clobbered; edit patches `lastMsg.msg`/`encMsg` when the edited message is still the previewed one. The coalescer reconciles its in-memory buffer on rewind/edit so a pending create can't resurrect a just-deleted preview.
- **Encryption**: in encrypted channel rooms the preview carries `encMsg` (content encrypted with the room key, computed once and shared between the event and Mongo — and, on edits, reused for `encryptedNewContent`) and `msg` is blanked; Mongo stores metadata-only so no plaintext content is at rest. DMs stay plaintext by design (`roomEncrypted = ENCRYPTION_ENABLED && type == channel`). Key-store errors Nak before any write or publish.
- Preview bodies are trimmed to a shared 256-rune snippet (`model.TrimPreview`) at every producer, so full message content never rides the RPC into Mongo or the delete event.

### Docs
- `docs/client-api.md` + derived views (`events.md`, `request-reply.md`): new `LastMessagePreview` shared schema (incl. `encMsg`), `message_deleted` `lastMessage` field + presence semantics + JSON examples (plaintext and encrypted), `newThreadLastMsgAt` wording corrected for reply deletes.

## Design decisions (chosen with the author)
- **RPC-from-broadcast-worker** rather than embedding the preview in the canonical `.deleted` event: keeps the worst-case bucket walk off the user's synchronous `msg.delete` request and off the durable canonical stream; failure retries ride JetStream redelivery.
- **Internal (server-to-server) RPC**, not client-facing: avoids account impersonation and per-viewer history-window distortion of a room-scoped value.
- **Fetch-first + Nak-on-failure**, not best-effort degrade: guarantees the preview on every delete event (best-effort would leave clients showing the deleted message — the bug we're fixing).
- **Trimmed preview + encrypted content twin**, not a full message blob: sonic-safe, no quoted-parent leak, one decode path for clients.

## Testing
- **TDD throughout** (red confirmed before green for every change).
- `make lint` — 0 issues; full unit suite green (83 packages, race detector); `gosec` clean.
- **Integration suite executed** against real containers (Cassandra / Mongo / NATS / Valkey / ES / MinIO): history-service (incl. new `GetLastRoomMessage` cases + tlm assertions), broadcast-worker (incl. encrypted `encMsg` create/delete/edit against real Mongo, drift-state rewind, legacy-room edit guard), and the rest of the affected services — all green.
- One integration-test bug caught and fixed here (`62b69a9`): a `bson.M` nested-doc assertion that mongo-driver v2 decodes as `bson.D` — the mocked unit tests couldn't surface it.
- `govulncheck`/`semgrep` are network-blocked in the sandbox and run in CI.

## Follow-ups (out of scope, noted for tracking)
- **Bug B — thread-lane plaintext leak.** `handleThreadCreated`/`handleThreadUpdated` publish plaintext content to thread followers even in encrypted channel rooms (the main-timeline paths encrypt; the hidden-thread lanes don't). Investigated and **deliberately not fixed here** — it's a systemic gap in the thread fan-out that warrants a dedicated, security-reviewed fix. Delete is unaffected (no content).
- **Cold-start previews.** `rooms.lastMsg` is persisted but not yet projected into the client-facing subscription/room-list response (user-service enrichment). Wiring it there (so mobile gets previews on cold start, not just via live delete events) is a follow-up; the shared schema is ready to link against.
- **Sidebar sort behaviour.** Deleting a room's newest message now rewinds `lastMsgAt`, so a room can move down the list — a product-visible change worth a confirm.
- **Preview content at rest + history-window scope (security sign-off wanted).** `rooms.lastMsg.msg` is denormalized in Mongo as a plaintext 256-rune snippet for non-encrypted rooms (DMs, and channels while encryption is off) — a new plaintext-content-at-rest surface distinct from Cassandra's at-rest encryption. Separately, the survivor preview is resolved server-to-server without a per-member history-window gate, so a member whose `historySharedSince` postdates the survivor can receive its snippet in `lastMessage`. Both are deliberate for now (snippet, not full body); flagging for an explicit security ack or a scoped follow-up.
- **Converge with the `rooms.get` last-message lane.** main's parallel `rooms.get` batch (`model.LastMessage`) and this PR's single-room RPC (`LastMessagePreview`) now share the 256-rune trim and the system-skip set, but still differ in wire shape and in sender-name fallback order (`EngName→UserDisplayName→Account` vs `EngName→AppName→Account`), and `rooms.get` does not yet skip system-message types. Converging them onto one shape + one skip rule (e.g. reimplementing `roomLastMessage` atop `GetLastRoomMessage`) is a follow-up so the room-list preview can't differ by refresh path.

## Summary by CodeRabbit

* **New Features**
  * Message deletion events now include the room's newest surviving message preview, helping clients refresh room lists without fetching history.
  * Last-message previews now support encrypted channels while preserving secure content handling.
  * Thread updates include the timestamp of the newest remaining reply after reply deletion.

* **Bug Fixes**
  * Room previews now stay synchronized after message creation, edits, and deletions.
  * Deleted, system, and thread-only messages are excluded when determining the surviving room preview.

* **Documentation**
  * Updated client API guidance and examples for last-message previews and encrypted content.


julianshen:
**Actionable comments posted: 6** — 🔴 3 blocking · 🟠 3 warning

<details>
<summary>📒 Files reviewed (40)</summary>

* `broadcast-worker/debug_log_test.go`
* `broadcast-worker/handler_lastmsg_test.go`
* `broadcast-worker/lastmsg_fetcher.go`
* `broadcast-worker/lastmsg_fetcher_test.go`
* `broadcast-worker/main.go`
* `broadcast-worker/mock_lastmsgfetcher_test.go`
* `broadcast-worker/mock_store_test.go`
* `broadcast-worker/pretouch.go`
* `broadcast-worker/store.go`
* `docs/client-api.md`
* `docs/client-api/events.md`
* `docs/reviews/branch-claude-adoring-bell-egysmi-2026-07-15.md`
* `history-service/internal/cassrepo/last_message.go`
* `history-service/internal/cassrepo/last_message_integration_test.go`
* `history-service/internal/cassrepo/last_message_unit_test.go`
* `history-service/internal/cassrepo/write.go`
* `history-service/internal/cassrepo/write_integration_test.go`
* `history-service/internal/service/last_message.go`
* `history-service/internal/service/last_message_test.go`
* `history-service/internal/service/migration.go`
* `history-service/internal/service/migration_test.go`
* `history-service/internal/service/rooms.go`
* `pkg/model/event.go`
* `pkg/model/lastmsg.go`
* `pkg/model/lastmsg_policy_test.go`
* `pkg/model/model_test.go`
* `pkg/model/room.go`
* `pkg/subject/subject.go`

</details>
<details>

* `opencode-review`

</details>

broadcast-worker/coalescer.go
```diff
@@ -58,13 +63,21 @@ func newCoalescingStore(inner Store, bulk bulkRoomLastMsgWriter) *coalescingStor
...
+// buffered user preview: the flush then lands the drift state
+// {lastMsgId: system, lastMsg: newest user message}, mirroring what the
+// direct (non-coalesced) writes produce.
+func (c *coalescingStore) UpdateRoomLastMessage(_ context.Context, roomID, msgID string, at time.Time, mentionAll bool, preview *model.LastMessagePreview) error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	cur := c.pending[roomID]
 	if at.After(cur.at) {
 		cur.msgID = msgID
 		cur.at = at
```
julianshen, line 88: _🎯 correctness_ | _🔴 Blocking_ | _`opencode-review`_

**Preview replacement is incorrectly nested under `at.After(cur.at)`.**

For arrivals user m1@t1, system@t3, then delayed user m2@t2, the pointer correctly remains at t3 but m2's newer user preview is ignored, leaving m1. Track and compare preview timestamps independently from the all-message pointer, and add this out-of-order drift case to the tests.

<details>

```
In `broadcast-worker/coalescer.go` around line 77: Preview replacement is incorrectly nested under `at.After(cur.at)`. For arrivals user m1@t1, system@t3, then delayed user m2@t2, the pointer correctly remains at t3 but m2's newer user preview is ignored, leaving m1. Track and compare preview timestamps independently from the all-message pointer, and add this out-of-order drift case to the tests.

```

</details>

broadcast-worker/handler.go
```diff
@@ -496,7 +553,42 @@ func (h *Handler) handleDeleted(ctx context.Context, evt *model.MessageEvent) er
...
+		survivor = room.LastMsg
+	} else {
+		survivor, err = h.lastMsgFetcher.FetchLastMessage(ctx, msg.RoomID)
+		if err != nil {
+			return fmt.Errorf("fetch last message for room %s: %w", msg.RoomID, err)
+		}
+		// Encrypted rooms: transform the fetched survivor in place (EncMsg
+		// ciphertext, Msg blanked) BEFORE any write, so the same preview object
+		// feeds both the Mongo rewind and the published event — one ciphertext,
+		// two destinations. Encryption failure Naks; nothing has been written
```
julianshen, line 591: _🎯 correctness_ | _🔴 Blocking_ | _`opencode-review`_

**The same non-system `survivor` is used to rewind both `lastMsgId` and `lastMsg`.**

This cannot preserve the documented drift semantics where the pointer includes system messages but the preview does not. For user A@t1, system S@t2, user B@t3, deleting B makes the history fetch skip S and Mongo rewinds the pointer to A instead of S. Fetch/represent the latest surviving pointer separately from the latest user preview.

<details>

```
In `broadcast-worker/handler.go` around line 575: The same non-system `survivor` is used to rewind both `lastMsgId` and `lastMsg`. This cannot preserve the documented drift semantics where the pointer includes system messages but the preview does not. For user A@t1, system S@t2, user B@t3, deleting B makes the history fetch skip S and Mongo rewinds the pointer to A instead of S. Fetch/represent the latest surviving pointer separately from the latest user preview.

```

</details>

history-service/internal/service/last_message.go
```diff
@@ -0,0 +1,76 @@
...
+		return nil, errcode.BadRequest("roomId is required")
+	}
+	c.WithLogValues("room_id", req.RoomID)
+	now := time.Now().UTC()
+
+	lastMsgAt, createdAt, err := s.resolveRoomTimes(c, req.RoomID, nil, now)
+	if err != nil {
+		// A room unknown to Mongo has no last message — answer nil rather than
+		// erroring, so callers treat it like an empty room.
+		if errors.Is(err, mongo.ErrNoDocuments) {
```
julianshen, line 32: _🎯 correctness_ | _🔴 Blocking_ | _`opencode-review`_

The recovery query derives its ceiling from the room's persisted `lastMsgAt`, which may intentionally lag while creates are buffered by the coalescer. If Mongo is at t0, m1@t1 and m2@t2 are buffered, and m2 is deleted before flush, this endpoint searches only through t0 and returns nil; the coalescer then drops the pending entry and loses m1. The request should supply a trustworthy upper bound such as the deleted event timestamp, or this endpoint should search through now rather than the stale denormalized pointer.

<details>

```
In `history-service/internal/service/last_message.go` around line 32: The recovery query derives its ceiling from the room's persisted `lastMsgAt`, which may intentionally lag while creates are buffered by the coalescer. If Mongo is at t0, m1@t1 and m2@t2 are buffered, and m2 is deleted before flush, this endpoint searches only through t0 and returns nil; the coalescer then drops the pending entry and loses m1. The request should supply a trustworthy upper bound such as the deleted event timestamp, or this endpoint should search through now rather than the stale denormalized pointer.

```

</details>

broadcast-worker/coalescer.go
```diff
@@ -73,6 +86,59 @@ func (c *coalescingStore) UpdateRoomLastMessage(_ context.Context, roomID, msgID
...
+// delegating the guarded Mongo rewind: the flush write is unguarded, so a
+// buffered create of the deleted message would otherwise resurrect its
+// preview (full content) right after the rewind removed it.
+func (c *coalescingStore) RewindRoomLastMessage(ctx context.Context, roomID, deletedMsgID string, survivor *model.LastMessagePreview, updatedAt time.Time) error {
+	c.mu.Lock()
+	if cur, ok := c.pending[roomID]; ok {
+		switch {
+		case cur.msgID == deletedMsgID:
+			if survivor != nil {
+				cur.msgID = survivor.MessageID
```
julianshen: _🎯 correctness_ | _🟠 Warning_ | _`opencode-review`_

**Rewinding a buffered delete target sets `cur.at` to the older survivor creation time.**

Bulk persistence uses this value for both `lastMsgAt` and room `updatedAt`, so a later flush can overwrite the delegate's deletion-time `updatedAt` with an older timestamp. Store pointer time and mutation/update time separately, preserving the supplied `updatedAt` for the bulk write.

<details>

```
In `broadcast-worker/coalescer.go` around line 99: Rewinding a buffered delete target sets `cur.at` to the older survivor creation time. Bulk persistence uses this value for both `lastMsgAt` and room `updatedAt`, so a later flush can overwrite the delegate's deletion-time `updatedAt` with an older timestamp. Store pointer time and mutation/update time separately, preserving the supplied `updatedAt` for the bulk write.

```

</details>

history-service/internal/cassrepo/last_message.go
```diff
@@ -0,0 +1,120 @@
...
+	"time"
+
+	"github.com/gocql/gocql"
+
+	"github.com/hmchangw/chat/history-service/internal/models"
+	"github.com/hmchangw/chat/pkg/model"
+)
+
+// lastMessageScanPageSize keeps each Cassandra round-trip generous: the walk
+// wants a single row, but a naive PageSize(1) would degenerate into one
```
julianshen: _🎯 correctness_ | _🟠 Warning_ | _`opencode-review`_

After 250 deleted/system rows, the scan returns nil even when a qualifying older message exists within the configured floor and bucket range. The caller interprets nil as an empty room and may clear `lastMsg`. Continue scanning to the existing bucket/floor bound, or return an explicit incomplete/error result instead of a false empty result.

<details>

```
In `history-service/internal/cassrepo/last_message.go` around line 15: After 250 deleted/system rows, the scan returns nil even when a qualifying older message exists within the configured floor and bucket range. The caller interprets nil as an empty room and may clear `lastMsg`. Continue scanning to the existing bucket/floor bound, or return an explicit incomplete/error result instead of a false empty result.

```

</details>

broadcast-worker/handler.go
```diff
@@ -729,21 +821,41 @@ func buildDeleteRoomEvent(room *model.Room, evt *model.MessageEvent) model.Delet
 	}
 }
```
julianshen, line 838: _🎯 correctness_ | _🟠 Warning_ | _`opencode-review`_

The new empty-content fast path changes encrypted edit behavior: the previous implementation encrypted even an empty `NewContent`, while this now produces both an empty plaintext field and nil `EncryptedNewContent`. Attachment-only messages can legitimately be edited to empty text, and encrypted consumers may require an encrypted payload. Keep encoding empty edit content, using a separate optional helper only where an omitted preview ciphertext is desired.

<details>

```
In `broadcast-worker/handler.go` around line 822: The new empty-content fast path changes encrypted edit behavior: the previous implementation encrypted even an empty `NewContent`, while this now produces both an empty plaintext field and nil `EncryptedNewContent`. Attachment-only messages can legitimately be edited to empty text, and encrypted consumers may require an encrypted payload. Keep encoding empty edit content, using a separate optional helper only where an omitted preview ciphertext is desired.

```

</details>

PR #78, opened 2026-07-16
Exclude bots from room read-floor and read-receipt queries

saurav-err-137:
## Summary

Exclude bot subscriptions from the room-wide read-floor (`minUserLastSeenAt`) and read-receipt list calculations in `room-service`. Bots are already distinguished in member counts and push notifications; this extends that distinction to read state.

## Changes

- **`store_mongo.go`**: Add `"u.isBot": {"$ne": true}` filter to `MinSubscriptionLastSeenByRoomID` and `ListReadReceipts` queries
  - `$ne: true` (not `isBot: false`) preserves backward compatibility with legacy subscription docs that lack the `u.isBot` field, treating them as humans
  - Updated docstrings to reflect bot exclusion and correct prior documentation that stated bots were counted
  - Residual filter applied to existing `{roomId, lastSeenAt}` index; no new index added (bots are few per room, and `$ne` yields no tight index bounds)

- **`integration_test.go`**: Extend test coverage for both methods
  - `TestMongoStore_MinSubscriptionLastSeenByRoomID_Integration`: 
    - Update existing "botdm" case to expect human's `lastSeenAt` instead of `nil`
    - Add "passive-bot" case: multiple humans + never-reading bot → floor = human minimum
    - Add "bot-earlier" case: bot with earlier `lastSeenAt` than all humans → floor still = human minimum (bot excluded)
    - Add "bot-only" case: bot-only room → `nil` (no countable subscriptions)
  - `TestMongoStore_ListReadReceipts_Integration`:
    - Add bot user and subscription with `lastSeenAt` well past message time
    - Verify bot is excluded from receipt list despite having read after the message

- **`docs/superpowers/specs/2026-07-16-bot-exclusion-read-floor-and-receipts-design.md`** (new): Comprehensive design spec covering background, scope, implementation, indexing rationale, testing strategy, and rollout notes

## Implementation Details

- **Backward compatible**: No schema or wire-format change; `u.isBot` flag already exists in production
- **Self-correcting**: Floor is recomputed on every `message.read`, so corrected values land on next read in affected rooms with no migration
- **No wiring changes**: The floor's write path (`messageRead` handler → `UpdateSubscriptionRead` → `MinSubscriptionLastSeenByRoomID` → `UpdateRoomMinUserLastSeenAt`) flows the fix straight into `rooms.minUserLastSeenAt` at a single choke point; user-service subscription-list `$lookup` inherits the fix for free
- **Out of scope**: Thread read-floor and thread read-receipts deferred (require schema addition to `thread_subscriptions`)

https://claude.ai/code/session_01Lg7w9sovhKWgtki6B9LRm2

### DIFF
--- a/room-service/store_mongo.go
+++ b/room-service/store_mongo.go
@@ -1150,16 +1150,17 @@ func (s *MongoStore) GetUserSiteID(ctx context.Context, account string) (string,
 }
 
 // MinSubscriptionLastSeenByRoomID returns the room's strict read floor: the
-// minimum lastSeenAt across ALL of the room's subscriptions, but only when
-// EVERY subscription has a usable lastSeenAt (> zero). If any subscription has
+// minimum lastSeenAt across all of the room's non-bot subscriptions, but only
+// when EVERY such subscription has a usable lastSeenAt (> zero). If any has
 // no usable lastSeenAt — missing, null, or the BSON zero date, i.e. a member
 // who was invited but has never opened the room — it returns nil, meaning "not
 // everyone has read yet". It also returns nil for a room with no subscriptions.
-// Bots are ordinary subscriptions and are counted: a botDM room (the bot never
-// reads) therefore always resolves to nil. The caller $unsets
-// rooms.minUserLastSeenAt on a nil result.
+// Bots (u.isBot) are excluded: a passive bot never freezes the floor, and a
+// botDM resolves to the human's lastSeenAt. A room with only bot subscriptions
+// therefore resolves to nil. The caller $unsets rooms.minUserLastSeenAt on a
+// nil result.
 func (s *MongoStore) MinSubscriptionLastSeenByRoomID(ctx context.Context, roomID string) (*time.Time, error) {
-	// The whole result is determined by a single document: the room's
+	// The whole result is determined by a single document: the room's non-bot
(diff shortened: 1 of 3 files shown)


room-service/store_mongo.go
```diff
@@ -1169,13 +1170,16 @@ func (s *MongoStore) MinSubscriptionLastSeenByRoomID(ctx context.Context, roomID
...
+	// predicate is applied as a residual filter (bots are few per room, and
+	// $ne yields no tight index bound), so this stays a bounded index seek on the
+	// message-read hot path rather than the prior full-room $group scan. $ne:true
+	// (not isBot:false) keeps legacy subs missing the flag counted as humans.
 	var doc struct {
 		LastSeenAt time.Time `bson:"lastSeenAt"`
 	}
 	err := s.subscriptions.FindOne(ctx,
-		bson.M{"roomId": roomID},
+		bson.M{"roomId": roomID, "u.isBot": bson.M{"$ne": true}},
```
mliu33, line 1182: We also need to do the same when calculating min last seen for thread room

mliu33 (approved):
Great work. Thanks. We also need to excluded bot and admin account in thread room min last seen.

Also need to exclude Admin user account

PR #79, opened 2026-07-17
feat(message-worker): durable OUTBOX federation for thread subscriptions

mliu33:
## What & why

`message-worker` federates thread subscriptions cross-site — when a thread reply involves a participant (the parent-message author or the replier) whose home site differs from the local site, the remote site must create/update that user's `ThreadSubscription`.

It did this with a **direct INBOX publish** on the `MESSAGES_CANONICAL` consumer's default `MaxDeliver=5`. So a destination-site outage longer than ~5 redeliveries **permanently dropped** the event — the local write had already committed, leaving local and remote state diverged with no repair. This is exactly the failure mode PR #410 fixed for **membership**, which `thread_subscription_upserted` was never migrated onto.

This PR routes `thread_subscription_upserted` through the local **OUTBOX** relay, so `outbox-worker` forwards it to the destination INBOX with retry-forever (`MaxDeliver=-1`). A destination down for an hour (up to OUTBOX retention) now **delays** the event instead of dropping it.

## Changes

- **`pkg/outbox`** — add `thread_subscription_upserted` to `ConcurrentEventTypes` (the order-insensitive lane); update producer docs to include `message-worker`.
- **`message-worker`** — `publishThreadSubInboxIfRemote` now calls the shared `outbox.Publish` onto the local OUTBOX instead of a direct `InboxExternal` publish. Envelope bytes and the dedup-ID seed are unchanged.
- **Tests** — outbox-worker end-to-end forward coverage for the new type (via a shared `assertConcurrentLaneForwards` harness); message-worker unit tests retargeted to the OUTBOX envelope.

## Design decisions

- **Concurrent lane, not FIFO** — `inbox-worker.UpsertThreadSubscription` is `$setOnInsert` (immutable identity) + `$max hasMention` (monotonic), so it is genuinely order-insensitive and identical to today's (unordered) delivery semantics. No ordering regression.
- **Zero production change** to `outbox-worker`, `inbox-worker`, `main.go`, or `bootstrap.go` — the consumer is event-type-generic, and `message-worker` already held the publish closure `outbox.Publish` needs (mirrors `room-worker`'s `federate`).
- **Hot path neutral-to-better** — replaces a cross-gateway INBOX publish with a local OUTBOX publish (same count, lower latency, higher availability) on a high-throughput worker.

## Scope

Durability **parity with #410** — delay-not-drop *within* OUTBOX retention. Beyond-retention **reconciliation** for thread subscriptions is an accepted follow-up, matching membership's own (still-unbuilt) backstop. No client-facing RPC or `pkg/model` client/event struct changed, so `docs/client-api.md` is untouched.

## Testing

- `make lint` — 0 issues
- `make test` (full suite, `-race`) — pass
- `make test-integration SERVICE=outbox-worker` — pass, incl. the new concurrent-lane forward test
- `make sast` — gosec / govulncheck / semgrep all pass
- Coverage: `publishThreadSubInboxIfRemote` **91.7%**, `pkg/outbox` **86.7%**

Design & plan: `docs/superpowers/specs/2026-07-17-message-worker-thread-subscription-outbox-durability-design.md`, `docs/superpowers/plans/2026-07-17-message-worker-thread-subscription-outbox-durability.md`

---

https://claude.ai/code/session_01FNMS2LV1evxmEZTKXxvFKX

### DIFF
--- a/docs/superpowers/plans/2026-07-17-message-worker-thread-subscription-outbox-durability.md
+++ b/docs/superpowers/plans/2026-07-17-message-worker-thread-subscription-outbox-durability.md
@@ -1,0 +1,534 @@
+# message-worker Thread-Subscription OUTBOX Durability — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Route message-worker's cross-site `thread_subscription_upserted` event through the local OUTBOX stream so a destination-site outage delays-not-drops it within retention, matching the #410 membership-federation durability guarantee.
+
+**Architecture:** Add `thread_subscription_upserted` to the OUTBOX concurrent partition (`pkg/outbox.ConcurrentEventTypes`), then change message-worker's `publishThreadSubInboxIfRemote` from a direct `InboxExternal` publish to the shared `outbox.Publish(...)` onto the local OUTBOX. `outbox-worker`'s per-destination concurrent consumer forwards it to the destination INBOX with `MaxDeliver=-1` — no code change in `outbox-worker` or `inbox-worker`, since both are event-type-generic.
+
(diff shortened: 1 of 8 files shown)


vjauhari-work (approved):
LGTM!

PR #81, opened 2026-07-17
feat(user-service): per-user settings RPCs (settings.get/set) + settings.update event

### LINKED ISSUE #80: [Feature] Per-user settings RPCs (settings.get / settings.set) + settings.update event
> **Verified against:** `main`@a6fa309 — code read after a fresh `git pull` (code is source of truth over docs).

## Summary
Add a per-user settings surface: two client RPCs (`settings.get` / `settings.set`) over a `settings` sub-document on the user record, plus a `settings.update` event so other logged-in devices sync live. No user-settings API exists today — the only user-global mutable state is `statusText`/`statusIsShow` (`UserStatusSet`).

Settings (all optional; **absent = client-defined default** — the server stores only what the user explicitly set and never fills defaults):

```json
{
  "fullWidth": true,
  "translateMessageInto": "en-US",
  "showMessagePreviewInSidebarList": true,
  "muteAllNotifications": false,
  "showMessagesAndPreviewsInNotifications": true,
  "showNotificationsDuringCallsAndMeetings": true,
  "scrollToBottomInChat": false
}
```

## Message-path trace  (REQUIRED — mark change / no-change / N/A per segment; a miss must show as a visible blank)
1. Client request struct + docs/client-api.md — **change**: new `SettingsGetRequest` (empty) / `SettingsSetRequest` (7 optional pointer fields); new client-api sections + derived views.
2. Gatekeeper mapping (request → canonical) — **N/A** (not on the message path; user-service request/reply).
3. Canonical model / event structs (pkg/model) — **change**: `UserSettings` (7 pointer fields, `json`+`bson`, all `omitempty`); `User.Settings *UserSettings`; `SettingsUpdateEvent{Timestamp, Settings}`.
4. Workers (message-worker / broadcast-worker / notification-worker) — **no change** (verified: no worker reads user settings today).
5. Storage tables (Cassandra / Mongo) — **change (Mongo only)**: `users.settings` embedded sub-document; no migration (absent field = never set); no new indexes. No Cassandra.
6. Read / history path (history-service) — **no change**.
7. Client-facing events (Leg-2 fan-out) — **change**: `chat.user.{account}.event.settings.update` published by user-service after a successful set (same pattern as `subscription.update`).

## Affected code  (verified, file:line — code is source of truth over docs)
- `user-service/service/` — new `settings.go` handlers beside `SetStatus` (`user-service/mongorepo/users.go:107` FindOneAndUpdate precedent).
- `user-service/mongorepo/users.go` — `GetUserSettings` (projection `{settings:1}`) + `UpdateUserSettings` (partial `$set` on `settings.<field>` per present pointer; `activeUserFilter` scoping like `SetUserStatus`).
- `pkg/subject/subject.go` — `UserSettingsGet/SetPattern` (`chat.user.{account}.request.user.{siteID}.settings.get|set`) + `SettingsUpdate(account)` event subject.
- `pkg/model/user.go` (Settings field) + new `pkg/model/usersettings.go`.

## Contracts & docs
- docs/client-api.md updated same PR? **Y** — new "User settings" section (request/response field tables + JSON example + the "absent = client default" contract) + `docs/client-api/request-reply.md` and `docs/client-api/events.md` derived views.
- Schema change? **N** (embedded Mongo sub-document, no migration, no index).

## Errors
- `errcode.BadRequest` for a malformed `translateMessageInto` (non-empty language-tag shape check; permissive, no whitelist). Infra failures return wrapped errors (they collapse to the generic error category at the boundary).

## Tests
- TDD red→green; unit (table-driven): get-unset → `{}`, get-after-set round-trip, partial set touches only sent fields (concurrent-field isolation), invalid language tag → BadRequest, event emitted on set with the resulting settings; integration (pkg/testutil Mongo): persisted sub-doc shape + partial `$set` semantics. ≥80% coverage.

## Acceptance criteria
- `settings.get` returns exactly the stored sub-document (empty object when never set) — the server never injects defaults.
- `settings.set` with any subset of the 7 fields updates only those fields; unsent fields keep their stored value (or stay absent).
- A successful set publishes `settings.update` carrying the full post-update settings.
- `docs/client-api.md` documents the fields, the partial-set semantics, and the absent-= -client-default contract.

## Dependencies / open questions
- None blocking. Cross-client default consistency is FE-owned by design (documented in the API section).

## Out of scope
- Server-side defaults; per-device settings; applying `muteAllNotifications` inside notification-worker (a consumer of these settings is a separate feature).

## PR / merge
- pr-self-audit runs as a **pre-review gate; its result is NOT included in the PR body** (policy 2026-06-17)
- Draft PR → review → CI green → request review; maintainer merges (no solo-merge).

chenjr0719:
## Summary

Per-user settings surface: `settings.get` / `settings.set` RPCs over a `settings` sub-document on the user record, plus a `settings.update` event so the caller's other devices sync live. The server stores **only what the user explicitly set** — every field is optional and an absent field means the client applies its own default (stated in the API doc).

Closes #80

## What's included

- **`pkg/model`** — `UserSettings` (7 optional pointer fields: `fullWidth`, `translateMessageInto`, `showMessagePreviewInSidebarList`, `muteAllNotifications`, `showMessagesAndPreviewsInNotifications`, `showNotificationsDuringCallsAndMeetings`, `scrollToBottomInChat`), `User.Settings`, `SettingsUpdateEvent`.
- **`pkg/subject`** — `settings.get` / `settings.set` request builders + patterns and the `settings.update` event subject.
- **`user-service`** — `GetSettings` (never-set → `{}`; the server never injects defaults) and `SetSettings` (**partial update**: only fields present in the request are written, via per-field `$set`; empty request → `bad_request`; permissive language-tag check on `translateMessageInto`). A successful set publishes `settings.update` with the full post-update settings over core NATS (ephemeral client delivery, mirroring the existing subscription-update pattern).

## NATS subjects

| Subject | Kind |
|---|---|
| `chat.user.{account}.request.user.{siteID}.settings.get` | request/reply |
| `chat.user.{account}.request.user.{siteID}.settings.set` | request/reply |
| `chat.user.{account}.event.settings.update` | server→client event |

## Verification

- Unit (table-driven): get-unset → `{}`, stored round-trip, partial-set field isolation, invalid/valid language tags, event carries the full post-update settings, publish failure is best-effort, not-found/store-error paths.
- Integration (build-tagged, live Mongo): persisted sub-document shape + per-field `$set` semantics.
- `docs/client-api.md` + derived views updated in the same PR (field tables, JSON examples, partial-set semantics, the absent-= -client-default contract).

## Summary by CodeRabbit

* **New Features**
  * Added per-user settings APIs to retrieve stored preferences and apply partial updates.
  * Successful updates now emit a client-facing `settings.update` fan-out containing the full post-update settings snapshot to the caller’s other connected devices.
  * Cross-site propagation now replicates settings updates to other sites via inbox events (with best-effort delivery).
* **Documentation**
  * Documented the new settings RPCs, validation behavior, error responses, and clarified which operation emits client events.
* **Tests**
  * Added unit and integration coverage for settings serialization, persistence, event payload shape, fan-out behavior, and inbox update handling.

### DIFF
--- a/inbox-worker/integration_test.go
+++ b/inbox-worker/integration_test.go
@@ -1483,28 +1483,35 @@ func TestInboxWorker_UpdateUserSettings_Integration(t *testing.T) {
 		userCol: db.Collection("users"),
 	}
 
-	_, err := db.Collection("users").InsertOne(ctx, model.User{ID: "u1", Account: "alice", SiteID: "site-b"})
-	require.NoError(t, err)
-
 	t1 := time.UnixMilli(1000).UTC()
 	t2 := time.UnixMilli(2000).UTC()
 	yes, no := true, false
 
+	// Each subtest owns its own account so it passes standalone under -run.
+	seed := func(t *testing.T, account string) {
+		t.Helper()
+		_, err := db.Collection("users").InsertOne(ctx, model.User{ID: "u-" + account, Account: account, SiteID: "site-b"})
+		require.NoError(t, err)
+	}
+
 	t.Run("writes the full settings sub-document", func(t *testing.T) {
-		require.NoError(t, store.UpdateUserSettings(ctx, "alice", &model.UserSettings{MuteAllNotifications: &yes}, t1))
+		seed(t, "settings-write")
+		require.NoError(t, store.UpdateUserSettings(ctx, "settings-write", &model.UserSettings{MuteAllNotifications: &yes}, t1))
 
 		var got model.User
-		require.NoError(t, store.userCol.FindOne(ctx, bson.M{"account": "alice"}).Decode(&got))
+		require.NoError(t, store.userCol.FindOne(ctx, bson.M{"account": "settings-write"}).Decode(&got))
 		require.NotNil(t, got.Settings)
 		require.NotNil(t, got.Settings.MuteAllNotifications)
 		assert.True(t, *got.Settings.MuteAllNotifications)
 	})
 
 	t.Run("replaces wholesale — a field dropped by the user is dropped here", func(t *testing.T) {
-		require.NoError(t, store.UpdateUserSettings(ctx, "alice", &model.UserSettings{FullWidth: &no}, t2))
+		seed(t, "settings-replace")
+		require.NoError(t, store.UpdateUserSettings(ctx, "settings-replace", &model.UserSettings{MuteAllNotifications: &yes}, t1))
+		require.NoError(t, store.UpdateUserSettings(ctx, "settings-replace", &model.UserSettings{FullWidth: &no}, t2))
 
 		var got model.User
-		require.NoError(t, store.userCol.FindOne(ctx, bson.M{"account": "alice"}).Decode(&got))
+		require.NoError(t, store.userCol.FindOne(ctx, bson.M{"account": "settings-replace"}).Decode(&got))
 		require.NotNil(t, got.Settings)
 		assert.Nil(t, got.Settings.MuteAllNotifications, "whole-object replace must not merge the previous value")
 		require.NotNil(t, got.Settings.FullWidth)
@@ -1512,10 +1519,12 @@ func TestInboxWorker_UpdateUserSettings_Integration(t *testing.T) {
 	})
 
 	t.Run("stale event is rejected by the settingsUpdatedAt high-water guard", func(t *testing.T) {
-		require.NoError(t, store.UpdateUserSettings(ctx, "alice", &model.UserSettings{MuteAllNotifications: &yes}, t1))
+		seed(t, "settings-stale")
+		require.NoError(t, store.UpdateUserSettings(ctx, "settings-stale", &model.UserSettings{FullWidth: &no}, t2))
+		require.NoError(t, store.UpdateUserSettings(ctx, "settings-stale", &model.UserSettings{MuteAllNotifications: &yes}, t1))
 
 		var got model.User
-		require.NoError(t, store.userCol.FindOne(ctx, bson.M{"account": "alice"}).Decode(&got))
+		require.NoError(t, store.userCol.FindOne(ctx, bson.M{"account": "settings-stale"}).Decode(&got))
 		assert.Nil(t, got.Settings.MuteAllNotifications, "stale settings must not overwrite newer ones")
 		require.NotNil(t, got.Settings.FullWidth)
 	})
--- a/inbox-worker/integration_test.go
+++ b/inbox-worker/integration_test.go
@@ -1472,3 +1472,59 @@ func TestInboxWorker_UpdateUserStatus_Integration(t *testing.T) {
 		assert.Equal(t, int64(0), count)
 	})
 }
+
+func TestInboxWorker_UpdateUserSettings_Integration(t *testing.T) {
+	db := setupMongo(t)
+	ctx := context.Background()
+
+	store := &mongoInboxStore{
+		subCol:  db.Collection("subscriptions"),
+		roomCol: db.Collection("rooms"),
+		userCol: db.Collection("users"),
+	}
+
+	_, err := db.Collection("users").InsertOne(ctx, model.User{ID: "u1", Account: "alice", SiteID: "site-b"})
+	require.NoError(t, err)
+
+	t1 := time.UnixMilli(1000).UTC()
+	t2 := time.UnixMilli(2000).UTC()
+	yes, no := true, false
+
+	t.Run("writes the full settings sub-document", func(t *testing.T) {
+		require.NoError(t, store.UpdateUserSettings(ctx, "alice", &model.UserSettings{MuteAllNotifications: &yes}, t1))
+
+		var got model.User
+		require.NoError(t, store.userCol.FindOne(ctx, bson.M{"account": "alice"}).Decode(&got))
+		require.NotNil(t, got.Settings)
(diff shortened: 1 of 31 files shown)


user-service/service/settings.go
```diff
@@ -0,0 +1,88 @@
...
+
+// validateSettings rejects an empty request (nothing to write) and a
+// malformed translateMessageInto.
+func validateSettings(set *model.UserSettings) error {
+	if set.IsEmpty() {
+		return errcode.BadRequest("no settings provided")
+	}
+	if set.TranslateMessageInto != nil && !translateTagRe.MatchString(*set.TranslateMessageInto) {
+		return errcode.BadRequest("invalid translateMessageInto language tag")
+	}
```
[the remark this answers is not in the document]
chenjr0719: Good catch — applied. `""` is now accepted for `translateMessageInto` and stores the explicit "translation off" choice (an unset would fall back to the client-side default, erasing the user's intent). Tests + all three API-doc surfaces updated (`cd73893`, `65b3fd4`).

user-service/service/settings.go
```diff
@@ -0,0 +1,90 @@
...
+	if set.TranslateMessageInto != nil && *set.TranslateMessageInto != "" && !translateTagRe.MatchString(*set.TranslateMessageInto) {
+		return errcode.BadRequest("invalid translateMessageInto language tag")
+	}
+	return nil
+}
+
+// publishSettingsUpdate fans out the per-user settings.update event over core
+// NATS (ephemeral client delivery, like subscription.update); best-effort —
+// errors are logged, the next set re-broadcasts the full settings.
+func (s *UserService) publishSettingsUpdate(c *natsrouter.Context, account string, settings *model.UserSettings) {
```
mliu33: We also need to fanout to inbox Jetstream for all other sites like setUsetStatus. Because there are some notification settings which local site notification worker would need to determine whether we should send mobile notification to the target user

chenjr0719: Good call — added in `c68dfde`, mirroring `publishStatus`:

- **`pkg/model`** — new inbox event type `user_settings_updated` + a `UserSettingsUpdated{account, settings, timestamp}` payload.
- **`user-service`** — `publishSettingsInbox` fans the **full post-update settings** to every other site's external INBOX lane (skips self, errors logged and never fail the set). The client `settings.update` event and the cross-site fanout now share one timestamp so they agree on ordering.
- **`inbox-worker`** — a `user_settings_updated` case + handler that mirrors the settings onto the local users doc, guarded by a `settingsUpdatedAt` high-water mark exactly like `statusUpdatedAt`, so an out-of-order or duplicate delivery can't regress. An account that doesn't live on the receiving site is a silent no-op.

One design note worth flagging: because the event carries the **full post-update settings** rather than the partial delta, the replica write is a whole-object replace (last-write-wins, same as status) — not a field merge. So a setting the user clears is genuinely cleared on every site, instead of lingering from an earlier value. There's a test pinning that.

Verified on our dev stack with two sites configured, the second one backed by its own separate datastore so a write landing there could only have arrived over the inbox lane: the settings replicate, an event replayed with an older timestamp is correctly rejected with no regression, and a follow-up partial update arrives as the full merged object.

Note the notification worker doesn't read these settings yet — this change gets them onto every site so it can.

inbox-worker/integration_test.go
```diff
@@ -1472,3 +1472,59 @@ func TestInboxWorker_UpdateUserStatus_Integration(t *testing.T) {
...
+	})
+
+	t.Run("unknown account is a no-op", func(t *testing.T) {
+		require.NoError(t, store.UpdateUserSettings(ctx, "ghost", &model.UserSettings{FullWidth: &yes}, t2))
+
+		count, err := store.userCol.CountDocuments(ctx, bson.M{"account": "ghost"})
+		require.NoError(t, err)
+		assert.Equal(t, int64(0), count)
+	})
+}
```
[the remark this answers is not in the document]
chenjr0719: Good catch — fixed in `7463061`.

Confirmed it was a real order dependency, not just a style point: run the stale-event subtest alone with `-run '.../stale'` and there's no prior `settingsUpdatedAt`, so the guard's `$exists:false` branch matches, the write applies, and the assertion inverts. The subtest was only passing because a sibling had left a newer high-water mark behind.

Each subtest now seeds its own account and establishes its own high-water mark. Verified by running the stale case standalone (passes) and the full test (passes).

inbox-worker/main.go
```diff
@@ -197,6 +197,35 @@ func (s *mongoInboxStore) UpdateUserStatus(ctx context.Context, account, statusT
...
+	}}
+	set := bson.M{"settings": settings, "settingsUpdatedAt": updatedAt}
+	res, err := s.userCol.UpdateOne(ctx, filter, bson.M{"$set": set})
+	if err != nil {
+		return fmt.Errorf("update user settings for %q: %w", account, err)
+	}
+	if res.MatchedCount == 0 {
+		// Same reasoning as UpdateUserStatus: the write already committed, so a failed
+		// existence check must not Nak/retry — it only picks the warn message.
+		count, cerr := s.userCol.CountDocuments(ctx, bson.M{"account": account})
```
mliu33: No need to do another DB call to check user doc count. Please also help remove this logic in updateUserStatus function

mliu33 (approved):
Super, thanks

PR #84, opened 2026-07-19
room-service: make mark-thread-as-read a no-op for non-followers

mliu33:
The message.thread.read RPC previously returned a NotFound
("thread subscription not found") when the caller was a room member
but did not follow the thread (no ThreadSubscription). Reacting on a
thread message does not create a ThreadSubscription, so a user who
opened an unfollowed thread and reacted would get an error when the
client marked the thread read.

Treat a missing ThreadSubscription as an idempotent no-op: skip all
thread-read writes, cross-site federation, and floor recompute, and
return {status: "accepted"}. Room-membership and validation errors are
unchanged, and the not-room-member sentinel still takes precedence.

Removed the now-dead errThreadSubNotFound sentinel and updated the
client API docs (canonical + request/reply view).

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_013yz5qYbhQNNR1BfopFxdyU

## Summary by CodeRabbit

* **Behavior Changes**
  * Marking a thread as read now returns an accepted no-op when the caller is not following the thread.
  * Requests for threads that do not belong to the specified room are handled the same way.
  * No unnecessary updates or notifications are triggered for these requests.

* **Documentation**
  * Updated RPC behavior and error documentation to reflect the accepted no-op response.
  * Removed the obsolete “thread subscription not found” error from the documented responses.


room-service/handler.go
```diff
@@ -1560,7 +1560,10 @@ func (h *Handler) messageThreadRead(c *natsrouter.Context, req model.MessageThre
 	case errors.Is(subErr, model.ErrSubscriptionNotFound):
 		return nil, errNotRoomMember
 	case errors.Is(tsubErr, model.ErrThreadSubscriptionNotFound):
-		return nil, errThreadSubNotFound
+		// The caller is a room member but does not follow this thread, so there is
```
julianshen, line 1563: _🎯 correctness_ | _🟠 Warning_ | _`opencode-review`_

The thread-subscription-not-found branch runs before generic `subErr` handling, so it returns `accepted` even when room membership could not be validated (for example, `CheckMembership` returned `context.Canceled`, as the updated test explicitly demonstrates). This contradicts the documented membership gate and masks datastore/request failures as successful no-ops. Handle every non-nil `subErr` before treating a missing thread subscription as success; update the siblings-cancelled test to expect the membership error.

<details>

```
In `room-service/handler.go` around line 1563: The thread-subscription-not-found branch runs before generic `subErr` handling, so it returns `accepted` even when room membership could not be validated (for example, `CheckMembership` returned `context.Canceled`, as the updated test explicitly demonstrates). This contradicts the documented membership gate and masks datastore/request failures as successful no-ops. Handle every non-nil `subErr` before treating a missing thread subscription as success; update the siblings-cancelled test to expect the membership error.

```

</details>

julianshen:
**Actionable comments posted: 1** — 🟠 1 warning

<details>

* `docs/client-api.md`
* `room-service/handler.go`
* `room-service/helper.go`
* `room-service/helper_test.go`

</details>
<details>

* `opencode-review`

</details>

chenjr0719 (commented):
Reviewed this and verified the before/after behavior end-to-end on our dev stack — reads clean, no blockers from me.

**Before / after** (driving the exact case this fixes: a room member who does **not** follow the thread calls `message.thread.read`):

| | Result |
|---|---|
| Before this change | error — `thread subscription not found` |
| With this change | idempotent no-op → `{ "status": "accepted" }` |

Confirmed live: the non-follower flow flips from the error to `accepted`, and no thread-read writes / floor recompute run on that path. Reacting to a thread message then marking it read is a real UX gap, so this is a good fix.

One small note (not a blocker): a `threadId` that belongs to a *different* room than the subject now folds into the same `accepted` no-op rather than surfacing an explicit signal — the updated docs call this out as intentional. Flagging only in case you would want to keep an explicit signal there for client-side debugging.

chenjr0719 (approved):
Approving — verified the before/after on our dev stack (details in my earlier comment): the non-follower `message.thread.read` flips from the `thread subscription not found` error to an idempotent `accepted` no-op, with no thread-read writes on that path. Clean, well-scoped fix.

PR #77, opened 2026-07-16
feat: implement clear-all-thread-unread cross-site RPC

saurav-err-137:
## Summary

Implements a new client-facing RPC that clears the unread status of all of a user's threads across all sites in a single call — the server-side "mark all threads read" operation. The implementation follows the established `user-service` aggregator → per-site `room-service` fan-out pattern, with graceful per-site degradation.

### Subjects & Wire Types (`pkg/subject`, `pkg/model`)
- Added four subject builders: `UserThreadReadAll`, `UserThreadReadAllPattern` (client-facing), `RoomThreadReadAll`, `RoomThreadReadAllSubscribe` (internal server-to-server)
- Added four model structs: `ThreadReadAllRequest`, `ThreadReadAllResponse`, `RoomThreadReadAllRequest`, `RoomThreadReadAllResponse`

### Room Service Store (`room-service/store.go`, `room-service/store_mongo.go`)
- Added `ClearThreadSubscriptionsForAccount(ctx, account, now)` — marks all thread subscriptions as read (`lastSeenAt=now`, `hasMention=false`), returns affected rows for federation
- Added `ClearSubscriptionThreadUnreadForAccount(ctx, account)` — removes `threadUnread` and clears `alert` on subscriptions with unread threads
- Integration tests verify both methods correctly update MongoDB and leave other accounts untouched

### Room Service Handler (`room-service/handler.go`)
- Implemented `clearAllThreadRead` handler — the per-site leaf of the aggregator
- Concurrently clears thread-subscription and room-subscription state via errgroup
- For cross-site users, federates one `ThreadReadEvent` per cleared thread via the existing `thread_read` event path
- Deliberately skips thread-room read-floor recompute and `thread_message_read` fan-out (bulk dismiss must not advance sender receipts)
- Registered at `subject.RoomThreadReadAllSubscribe(siteID)`

### User Service Aggregator (`user-service/service/threadunread.go`)
- Implemented `ClearAllThreadUnread` handler — reads user's home-replica thread subscriptions, derives distinct owning sites, fans out to each site's `room-service`
- Uses bounded concurrency (`maxSiteFanout` semaphore) and per-site degradation: failed sites are recorded in `UnavailableSites` rather than failing the entire call
- Mirrors the existing `GetThreadUnreadSummary` aggregator pattern

### Room Client (`user-service/roomclient/client.go`)
- Added `ClearAllThreadUnread(ctx, siteID, account)` method to issue the bulk RPC to room-service
[...]

PR #87, opened 2026-07-20
Batch clear-all-thread-unread federation and drop clearedThreads (PR #77 follow-up)

saurav-err-137:
Follow-up to the merged #77, addressing @mliu33's review feedback. Three changes:

## 1. Batched federation instead of per-thread
The cross-site "mark all threads read" dismiss now federates as **one** `thread_read_all` event per remote peer, replacing the one-`thread_read`-per-thread loop.

- New inbox event `model.InboxThreadReadAll` (`"thread_read_all"`) + `ThreadReadAllEvent` payload, added to the `pkg/outbox` concurrent partition.
- `room-service` publishes a single event when the user's home site is remote.
- `inbox-worker` gains `ApplyThreadReadAll`: a home-replica bulk clear that advances every one of the account's thread subscriptions under a per-doc `$lt` guard (clearing `hasMention`) and clears every subscription's `threadUnread`/`alert` — mirroring the source-site write. This drops federation volume from O(threads) to O(remote-sites).

## 2. No Find/Update snapshot gap
`ClearThreadSubscriptionsForAccount` is now a single account-scoped `UpdateMany` instead of a `Find`-then-`UpdateMany`, so a thread subscription inserted between the two calls can no longer be cleared without being accounted for. The concern is designed out rather than patched.

## 3. Dropped `clearedThreads` from the response
Removed from both the client-facing `ThreadReadAllResponse` and the internal `RoomThreadReadAllResponse` (unused by the frontend). Success is now `{ unavailableSites? }`; the roomclient and aggregator return no count.

## Docs
`docs/client-api.md` and the derived `docs/client-api/request-reply.md` reflect the new response shape; the design spec carries a "Post-merge follow-up" addendum documenting the reversal from per-thread to batched federation.

## Testing
- Full unit suite passes; lint, gosec, and govulncheck are clean.
- New/updated tests: inbox-worker `thread_read_all` apply (happy/malformed/store-error), the `$lt` guard not regressing a future read, account isolation, and room-service single-event federation.
- Integration tests (room-service, inbox-worker, user-service) run in CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WLMQ1PMcmqvKqXbnqnQ9W8)_

### DIFF
--- a/docs/client-api.md
+++ b/docs/client-api.md
@@ -5121,11 +5121,10 @@ Empty object.
 
 | Field | Type | Notes |
 |---|---|---|
-| `clearedThreads` | number | Total thread subscriptions cleared across all responding sites. `0` when nothing was unread. |
 | `unavailableSites` | string[] | Optional. Sites whose per-site clear failed; their threads may remain unread. Omitted when all responded. |
 
 ```json
-{ "clearedThreads": 7 }
+{}
 ```
 
 ##### Error response
--- a/docs/client-api/request-reply.md
+++ b/docs/client-api/request-reply.md
@@ -1755,7 +1755,8 @@ failures degrade into `unavailableSites` rather than erroring.
 `{siteID}` is the **caller's own home site**. Clears the unread status of all of the
 user's threads across every site — the "mark all threads read" action. `user-service`
 asks each owning site's `room-service` to clear that user's thread-subscription read
-state and room-subscription thread-unread state.
+state and room-subscription thread-unread state; each remote site converges the user's
+home replica via one `thread_read_all` inbox event.
 
 #### Request body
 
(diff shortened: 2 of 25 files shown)


docs/client-api.md
```diff
@@ -5121,11 +5121,10 @@ Empty object.
 
 | Field | Type | Notes |
 |---|---|---|
-| `clearedThreads` | number | Total thread subscriptions cleared across all responding sites. `0` when nothing was unread. |
 | `unavailableSites` | string[] | Optional. Sites whose per-site clear failed; their threads may remain unread. Omitted when all responded. |
 
 ```json
-{ "clearedThreads": 7 }
+{}
```
mliu33, line 5127: I was actually talking about unavailableSites lol..

mliu33 (approved):
Great, thanks!

PR #91, opened 2026-07-20
feat(msgraph): send User-Agent on app-only Graph requests

saurav-err-137:
## Summary

Add configurable `User-Agent` header support to the app-only Microsoft Graph client, enabling room-service Teams meeting calls to pass corporate proxy/WAF fronting that rejects non-browser agents. The presence client already had this capability; the app-only client now gets the same treatment.

## Changes

- **`pkg/msgraph/msgraph.go`**
  - Add `userAgent string` field to `graphClient` struct
  - Resolve `userAgent` in `New()` from `Config.UserAgent`, falling back to `defaultUserAgent` (browser string)
  - Move `defaultUserAgent` constant from `presence.go` and generalize its doc comment (now applies to both app-only and presence clients)
  - Set `User-Agent` header explicitly at all four request sites: token fetch (`accessToken`), meeting creation (`CreateOnlineMeeting`), directory lookup (`resolveChunk`), and user listing (`fetchUsersPage`)
  - Update `Config.UserAgent` doc comment to reflect that it is honored by both app-only and presence clients

- **`pkg/msgraph/presence.go`**
  - Remove the `defaultUserAgent` constant declaration (now defined in `msgraph.go`, same package)

- **`pkg/msgraph/msgraph_test.go`**
  - Add four new tests asserting `User-Agent` header presence:
    - `TestCreateOnlineMeeting_SendsDefaultUserAgent` — verifies default browser string on token and meeting requests
    - `TestCreateOnlineMeeting_UserAgentOverride` — verifies custom override is used when `Config.UserAgent` is set
    - `TestResolveAccountIDs_SendsUserAgent` — verifies directory requests carry the header
    - `TestListUsers_SendsUserAgent` — verifies user listing requests carry the header

- **`room-service/main.go`**
  - Add `GraphUserAgent string` config field (env: `GRAPH_USER_AGENT`, default: empty)
  - Wire it into the `msgraph.New()` call so room-service's meeting requests use the configured agent

## Implementation Details

- Follows the existing `presence.go` pattern (Approach A): resolved field + explicit per-site header sets, mirroring the presence client's implementation
- `GRAPH_USER_AGENT` env var name chosen for cross-service consistency with `user-presence-service`
- Browser string default (`Mozilla/5.0 ...`) is the value most likely to pass both Microsoft Graph and corporate proxy/WAF filters
- All four app-only request sites (token, meeting, directory, list-users) now send the header, ensuring consistent behavior across all consumers
- No change to `docs/client-api.md` — this is an outbound HTTP header to Microsoft Graph, not a client-facing RPC schema change
- Tests follow TDD pattern: failing assertions first, then minimal implementation

https://claude.ai/code/session_01DbrQYGHCQ6rLw5NB2mRrwD

### DIFF
--- a/docs/superpowers/plans/2026-07-20-msgraph-app-only-user-agent.md
+++ b/docs/superpowers/plans/2026-07-20-msgraph-app-only-user-agent.md
@@ -1,0 +1,308 @@
+# App-only Graph client User-Agent Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make every request the shared app-only `msgraph` client sends carry a configurable `User-Agent` header (browser-string default), so room-service Teams meeting calls can pass a fronting proxy/WAF like the presence client already does.
+
+**Architecture:** Add a resolved `userAgent` field to `graphClient` (from `Config.UserAgent`, falling back to the shared `defaultUserAgent`), and set the header explicitly at all four request sites — mirroring the existing `presence.go` pattern (Approach A). Wire a `GRAPH_USER_AGENT` env var through room-service into `msgraph.Config.UserAgent`.
+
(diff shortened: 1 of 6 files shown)


mliu33 (approved):
Thanks!

mliu33:
FYI: @cocoamix00 we need to pass custom user-agent header in order to call MS api from the pod

PR #98, opened 2026-07-21
fix(upload-service): env-gate the setCookie Partitioned attribute

### LINKED ISSUE #97: [Fix] setCookie: Partitioned attribute blocks the top-level-navigation file download
# [Fix] setCookie: Partitioned attribute blocks the top-level-navigation file download

> **Verified against:** `main`@4c841fb — code read after a fresh `git pull` (code is source of truth over docs).

## Summary

`POST /file/setCookie` (upload-service) issues the `ssoToken` session cookie with `Partitioned: true`. A file download that performs a **top-level redirect/navigation** to the download URL then fails to authenticate: `Partitioned` (CHIPS) scopes the cookie to the top-level site it was set under, so on a top-level navigation to the download origin the cookie is not attached → the protected download is unauthorized.

## Reproduce steps

Preconditions: a browser session where `POST /file/setCookie` was called while the chat web app is the top-level site (so the `ssoToken` cookie is partitioned under the app's origin).

1. Trigger a file download that navigates (top-level redirect) to the file-download URL served by upload-service.
2. Observe: the request to the download endpoint carries **no `ssoToken` cookie** → auth fails → the download does not complete.

Expected: the download authenticates via the cookie. Actual: the cookie is absent on the top-level navigation, download blocked.

## Root cause (verified, file:line)

- `upload-service/handler.go:116` — `HandleSetCookie` sets `Partitioned: true` alongside `SameSite=None` (`:115`). `Partitioned` keys the cookie to the top-level site of the context that set it (the chat app). The download is a **top-level navigation to a different origin**, so the partitioned cookie is not attached.
- The download endpoints **stream** the file — `HandleDownloadFile` (`handler.go:322`, `c.DataFromReader`) and `HandleDownloadMinioS3File` (`:368`) — i.e. a real navigation to the service's own URL, **not** a signed-URL S3 redirect. So the browser must present the cookie to that origin on a top-level navigation, which a partitioned cookie won't do.
- Token read is header-first, cookie-fallback (`upload-service/middleware.go` `tokenFromRequest`); only this navigation-download path depends on the cookie (header callers are unaffected).

## Message-path trace

N/A — this is an HTTP cookie-attribute change on an upload-service handler, not a NATS message path. Segments 1–7 (client-req / gatekeeper / model / workers / storage / read / client-events) are all **N/A**.

## Fix (Primary recommended + alternatives)

`Partitioned` was added deliberately as the mitigation for the `SameSite=None` cookie (per the code comment) so the cross-site ``-embed download survives third-party-cookie deprecation (CHIPS). The two flows pull opposite ways:

- `` embed → cross-site subresource → wants **Partitioned**.
- redirect-download → top-level navigation → wants **not-Partitioned**.

**Primary (recommended):** env-gate the attribute — `SETCOOKIE_PARTITIONED` (bool, `envDefault:"true"`). Ops turns it off where the redirect-download is used; reversible without a code change; keeps CHIPS where the embed flow matters.

**Alt A — drop `Partitioned` outright:** simplest, unblocks the download now. But the cookie becomes a plain cross-site cookie (still `Secure` + `HttpOnly`), and the ``-embed flow then leans on third-party cookies that browsers are deprecating (Chrome phasing out; Safari/Firefox block by default).

**Alt B — durable, FE-side:** change the download from a top-level redirect to a credentialed `fetch()` → blob, so it stays in the app's partition and `Partitioned` works for *both* flows. Requires an FE change.

**Tradeoff:** Primary/Alt A unblock the redirect-download now; Alt B is the long-term-correct fix for the embed side but needs FE work. Recommend **Primary** so the choice is per-env without losing the CHIPS mitigation where the embed flow is required.

## Contracts & docs

No wire/API contract change. If any doc describes the cookie attributes, update it in the same PR. Security note: without `Partitioned`, the `SameSite=None` cookie is trackable cross-site (mitigated by `Secure` + `HttpOnly`).

## Tests

- Unit: assert `HandleSetCookie` sets/omits `Partitioned` per the env flag (Primary) — or omits it (Alt A) — and preserves `Secure` + `HttpOnly` + `SameSite=None`.
- Manual: the top-level-navigation download authenticates end-to-end after the change.

## Acceptance criteria

- A top-level-navigation file download authenticates via the `ssoToken` cookie.
- `Secure` + `HttpOnly` + `SameSite=None` are preserved.
- (Primary) the `Partitioned` attribute is env-controlled, default keep.

- The durable ``-embed fix (fetch→blob, Alt B) — separate FE-side work.
- Which flows must work on which browsers (embed on Safari/Firefox needs `Partitioned` or the fetch→blob approach) — informs whether the default should stay `true`.

## PR / merge

Draft PR on the source repo; `/simplify` + terse comments before ready; external review + maintainer merges.

chenjr0719:
## Summary

Make the `Partitioned` attribute on the sso-token cookie (`HandleSetCookie`) configurable via an environment variable, and default it **off**. `Partitioned` (CHIPS) scopes the cookie to the top-level site it was set under, so a top-level-navigation file download to the download origin does not carry the cookie and the protected download is unauthorized. Defaulting it off unblocks that download; the flag re-enables it where the cross-site ``-embed flow needs CHIPS.

## Changes

- `upload-service/main.go`: add `SetCookiePartitioned bool` config field (`env:"SETCOOKIE_PARTITIONED"`, default `false`), passed into `NewHandler`.
- `upload-service/handler.go`: add `setCookiePartitioned` to `Handler`; `HandleSetCookie` now sets `Partitioned: h.setCookiePartitioned` instead of a hardcoded `true`.
- `upload-service/handler_test.go`: thread the new constructor parameter through existing call sites; the pre-existing attribute test sets the flag explicitly to assert the enabled (Partitioned-present) path.
- `upload-service/handler_setcookie_test.go` (new): table test driving `HandleSetCookie` with the flag `true` and `false`, asserting `Partitioned` present/omitted while `Secure`, `HttpOnly`, and `SameSite=None` hold in both cases.
- `docs/client-api.md`: note that `Partitioned` is env-controlled, omitted by default, added only when enabled.

## Verification

`go build`, `go vet`, and `go test ./upload-service/...` all pass.

Closes #97

## Summary by CodeRabbit

* **New Features**
  * Added the `SETCOOKIE_PARTITIONED` configuration option to control whether SSO cookies include the `Partitioned` attribute.
  * The option defaults to disabled, preserving secure cookie attributes such as `Secure`, `HttpOnly`, and `SameSite=None`.

* **Documentation**
  * Updated the client API documentation to describe the configurable cookie behavior.

### DIFF
--- a/docs/client-api.md
+++ b/docs/client-api.md
@@ -454,8 +454,10 @@ Cross-origin callers must send the request with credentials (e.g. `fetch(..., { 
 Response header:
 
 ```
-Set-Cookie: ssoToken=<token>; Path=/; HttpOnly; Secure; SameSite=None; Partitioned
+Set-Cookie: ssoToken=<token>; Path=/; HttpOnly; Secure; SameSite=None
 ```
+
+`Partitioned` is env-controlled server-side (`SETCOOKIE_PARTITIONED`, default `false` — omitted) and is added only when enabled.
 
 ```json
 { "success": true }
--- a/upload-service/handler.go
+++ b/upload-service/handler.go
@@ -66,17 +66,20 @@ type Handler struct {
 	preview        previewFunc
 	nowMilli       func() int64
 	cacheMaxAge    int
+
+	setCookiePartitioned bool
 }
 
 // NewHandler wires the handler dependencies. maxImages/maxImageSize gate the image
 // endpoint; maxAttachments/maxFileSize/mimeFilter/preview gate the file endpoint; s3
-// backs the MinIO/S3 download endpoint; cacheMaxAge is its Cache-Control max-age in seconds.
+// backs the MinIO/S3 download endpoint; cacheMaxAge is its Cache-Control max-age in
+// seconds; setCookiePartitioned gates the Partitioned attribute on HandleSetCookie.
 func NewHandler(store Store, dc driveClient, s3 objectStore, maxImages, maxAttachments int, maxImageSize, maxFileSize int64,
-	mimeFilter *mediaTypeFilter, preview previewFunc, cacheMaxAge int) *Handler {
+	mimeFilter *mediaTypeFilter, preview previewFunc, cacheMaxAge int, setCookiePartitioned bool) *Handler {
 	return &Handler{
 		store: store, drive: dc, s3: s3, maxImages: maxImages, maxAttachments: maxAttachments,
 		maxImageSize: maxImageSize, maxFileSize: maxFileSize, mimeFilter: mimeFilter,
-		preview: preview, cacheMaxAge: cacheMaxAge,
+		preview: preview, cacheMaxAge: cacheMaxAge, setCookiePartitioned: setCookiePartitioned,
 		nowMilli: func() int64 { return time.Now().UTC().UnixMilli() },
 	}
 }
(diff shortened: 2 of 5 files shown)


mliu33 (approved):
Thanks

PR #15, opened 2026-07-14
feat(history-service): rooms.get — batch room last-message read (#393)

hmchangw:
## Summary

Implements `rooms.get` (chat#393) — the room last-message endpoint (the customer&#39;s
`/api/v1/rooms.get`, used by mobile to render the last-message snippet in the room
list). A new history-service, per-site, account-scoped **batch** RPC that resolves each
requested room&#39;s latest message at read time (A2). Maintainer-approved approach (A2,
option E2) per the #393 design-lock.

## What&#39;s included

- **Subject** `chat.user.{account}.request.history.{siteID}.rooms.get` — account+site from
  the subject, `{roomIds[]}` in the body. One batch per site; the caller groups a user&#39;s
  rooms by site and fans out (same shape as `ThreadSubscriptionList`).
- **`RoomsGet` handler** (history-service) → `roomLastMessage`: per room, the existing
  access check (`checkAccessAndRoomTimes`) then the latest message via the existing
  `messages_by_room … DESC` reads (`GetMessagesBefore` / `GetMessagesBetweenDesc`,
  limit 1) — access-window + history-floor aware (mirrors LoadHistory). No new repo method.
- **A2 read-only:** no denormalized write, no event, no edit/delete hooks, **no
  walk-back** — a soft-deleted latest message is returned as-is with `deleted=true`.
- **Best-effort batch:** bounded-concurrency per-room resolve; a room that&#39;s
  not-accessible / empty / errored is omitted, never failing the batch. Batch ≤ 100;
  `content` preview-trimmed to 256 runes.
- **`docs/client-api.md`** §3.2 &#34;Get Rooms Last Message&#34; (doc-ratchet).

## #393 completion — web-sidebar integration

- **Shared wire types moved to `pkg/model`** (`LastMessage`, `RoomsGetRequest`,
  `RoomsGetResponse`) so user-service can consume them without importing
  history-service internals; history-service&#39;s own `models` package now aliases them
  (`type LastMessage = model.LastMessage`, etc. — no behavior change).
- **`SubscriptionRoom.LastMessage *model.LastMessage`** (`pkg/model/subscription.go`) —
  new optional field on the room-derived view returned by `subscription.list`.
- **`historyclient.Client.RoomsGet`** — new per-site RPC call mirroring the existing
  `GetThreadList` pattern (marshal request, NATS request/reply, `errcode.Parse`,
  unmarshal). Added to the `HistoryClient` interface; mocks regenerated.
[...]

PR #107, opened 2026-07-22
feat(message-worker): Teams message-history migration consumer

chenjr0719:
## Summary

message-worker consumes Teams message-history batches published to
`chat.msg.canonical.{siteID}.teams.batch`, transforms each source message, and runs it
through the existing persist pipeline (persist + thread/mention/quote, migration-gated).
No reply, no re-publish: broadcast + notification skip automatically (they key on the
`.created` event this path doesn't emit), giving a silent, no-fan-out migration. It's the
consume sibling of the oplog-transformer migration.

The **forward** shape is stubbed (depends on a `Forwarded` model field that lands with the
forward feature); room + subscription creation is handled by the Teams onboarding pipeline,
out of scope here.

## Changes

- **Durable consumer** on `chat.msg.canonical.{siteID}.teams.batch` (message-worker); each
  batch message → transform → the service's own `processMessage` pipeline with
  `isMigration=true` (suppresses thread side-effects). No reply, no canonical re-publish.
- **`MessageTransformer` seam** + `DefaultTransformer`: source-agnostic (opaque JSON);
  HTML→supported-markdown (unsupported markup degrades to raw text), user vs system type,
  reply(quote) shape via `QuotedParentMessage` scoped to the outer message's room. Forward
  branch stubbed.
- **Sender resolution** via the service's own HR identity store, per-batch cache:
  employeeId-first (globally unique) → exactly-one display-name fallback → idempotent
  `employeeId` upsert. Mentions resolve through the same resolver.
- **Deterministic message id** = stable hash of (roomId, teamsId), so a batch re-run
  overwrites the same row (idempotent). Empty-roomId messages are skipped (no orphan /
  cross-conversation id collision).
- **Not client-facing** — no `docs/client-api.md` entry.

## NATS subjects

- **Consumed (server-produced):** `chat.msg.canonical.{siteID}.teams.batch` —
  `TeamsBatchRequest{ messages: [raw] }`. Per-message error isolation (a bad message is
  logged + skipped, the batch continues; an infra failure Naks for redelivery).

## Verification

- In-repo integration test (testcontainer Cassandra + Mongo): batch → persist to Cassandra
  + idempotent re-run (deterministic id → single row).
- Unit: transformer shapes (HTML→md, reaction map, reply/quote, system), sender-resolution
  matrix, deterministic-id stability, empty-roomId skip.
- `golangci-lint` on the touched packages: 0 issues.
- Wire-level (verified on our dev stack): a server publish of a batch to
  `chat.msg.canonical.{siteID}.teams.batch` is consumed and the migrated message lands as a
  normal message (`type=""`) in the room's message store — full publish → consume → transform
  → persist path.

## Known limitation

Migrated messages are **persisted but not indexed in search** — search indexing keys on
the `.created` canonical event this path intentionally does not emit; search backfill is a
follow-up. (See the design spec.)

Related: PRD #15.

## Summary by CodeRabbit

* **New Features**
  * Added Teams message-history migration using durable batch processing.
  * Migrated messages preserve senders, mentions, replies, formatting, system-message types, and deterministic IDs for safe reprocessing.
  * Added dedicated indexing for migrated Teams history, mirroring the same idempotent skip behavior.
  * Invalid/unsupported items are skipped without blocking the rest of a batch.
* **Documentation**
  * Added migration design and implementation plans, including indexing and reprocessing behavior.
* **Tests**
  * Added unit and integration coverage for transformation, identity resolution, batching error handling, persistence, and idempotent reruns.


mliu33:
You can rebase with latest main branch now

docs/client-api.md
```diff
@@ -6897,3 +6897,44 @@ See [Error envelope](#6-error-envelope-reference). HTTP statuses:
 #### Triggered events — error path
 
 `None.`
+
+---
+
+## S1. Server-to-server — Teams message-history migration
```
mliu33: This is not client-facing related rpc. No need to include in this document

docs/client-api.md
```diff
@@ -6897,3 +6897,44 @@ See [Error envelope](#6-error-envelope-reference). HTTP statuses:
...
+
+A push-based request/reply RPC that ingests a batch of Teams-history messages. The
+handler (in `message-worker`) transforms each message into a canonical message,
+publishes it onto the canonical stream with the `X-Migration: live` header, and lets
+the existing pipeline persist **and index** it silently — no broadcast, no
+notification. It is the push sibling of the oplog-transformer migration.
+
+### Subject
+
+`chat.msg.canonical.{siteID}.teams.batch` — request/reply.
```
mliu33: This should not be request/reply.

Client should just publish into Jetsream, and worker consumes and processes the message in asynchronous way

docs/client-api.md
```diff
@@ -6897,3 +6897,44 @@ See [Error envelope](#6-error-envelope-reference). HTTP statuses:
...
+| `results[].status` | string | `persisted` (canonical publish acked), `skipped` (no id), or `error`. |
+| `results[].error` | string | Present only when `status` is `error`. |
+
+Per-message error isolation: one bad message yields an `error` result; the batch
+continues. Message ids are a deterministic hash of the Teams id, so re-running a
+batch overwrites the same rows (idempotent).
+
+### Triggered events
+
+`chat.msg.canonical.{siteID}.created` with header `X-Migration: live` — persisted +
```
mliu33: No, we should not publish event into itself again. The client who publishes event into subject "chat.msg.canonical.{siteID}.teams.batch", which already persists into the corresponding Jetsream at targeted site.

message-worker/hridentity.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+var _ HRIdentityStore = (*mongoHRIdentityStore)(nil)
+
+func newMongoHRIdentityStore(db *mongo.Database) *mongoHRIdentityStore {
+	return &mongoHRIdentityStore{users: mongoutil.NewCollection[model.User](db.Collection("users"))}
+}
+
+func (s *mongoHRIdentityStore) FindUserByEmployeeId(ctx context.Context, siteID, employeeId string) (*model.User, error) {
+	u, err := s.users.FindOne(ctx, bson.M{"siteId": siteID, "employeeId": employeeId})
```
mliu33: No need to provide siteID, which is redundant as employeeID is unique

message-worker/main.go
```diff
@@ -202,6 +204,16 @@ func main() {
 		}
 	}()
 
+	// Teams message-history batch migration RPC: transforms + republishes onto the
+	// canonical stream with X-Migration: live (persist + index, no re-delivery).
+	teamsMigration := newTeamsBatchHandler(
+		newMongoHRIdentityStore(db), cfg.SiteID,
+		migration.NewCanonicalPublisher(cfg.SiteID, js.PublishMsg, func() int64 { return time.Now().UTC().UnixMilli() }),
```
mliu33: Let's not do another publish again since the event published by the client is already persisted in the Jetstream

mliu33 (changes requested):

julianshen:
**Actionable comments posted: 5** — 🔴 1 blocking · 🟠 3 warning · 🔵 1 suggestion

<details>

* `data-migration/oplog-transformer/handler.go`
* `data-migration/oplog-transformer/handler_test.go`
* `data-migration/oplog-transformer/integration_test.go`
* `data-migration/oplog-transformer/main.go`
* `docs/client-api.md`
* `docs/superpowers/plans/2026-07-22-teams-message-migration.md`
* `docs/superpowers/specs/2026-07-22-teams-message-migration-design.md`
* `message-worker/hridentity.go`
* `message-worker/main.go`
* `message-worker/mock_hridentity_test.go`
* `message-worker/teamsbatch.go`
* `message-worker/teamsbatch_integration_test.go`
* `message-worker/teamsbatch_test.go`
* `message-worker/teamssender.go`
* `message-worker/teamssender_test.go`
* `message-worker/teamstransform.go`
* `message-worker/teamstransform_test.go`
* `pkg/migration/canonical.go`
* `pkg/migration/canonical_test.go`
* `pkg/model/teams_migration.go`
* `pkg/subject/subject.go`

</details>
<details>

* `opencode-review`

</details>

message-worker/teamsbatch.go
```diff
@@ -0,0 +1,99 @@
...
+	head, ok := peekTeamsHead(raw)
+	res := model.TeamsBatchResult{TeamsMsgID: head.ID}
+	if !ok {
+		res.Status = model.TeamsBatchError
+		res.Error = "malformed message payload"
+		return res
+	}
+	if head.ID == "" {
+		res.Status = model.TeamsBatchSkipped
+		return res
```
julianshen: _🎯 correctness_ | _🔴 Blocking_ | _`opencode-review`_

**Only the source ID is validated; an empty roomId is published and used as the deterministic-ID scope.**

This creates orphaned messages and permits cross-conversation ID collisions. The new test explicitly treats such a message as persisted. Reject/skip empty roomId before transformation and update the test.

<details>

```
In `message-worker/teamsbatch.go` around line 61: Only the source ID is validated; an empty roomId is published and used as the deterministic-ID scope. This creates orphaned messages and permits cross-conversation ID collisions. The new test explicitly treats such a message as persisted. Reject/skip empty roomId before transformation and update the test.

```

</details>

message-worker/teamstransform.go
```diff
@@ -0,0 +1,213 @@
...
+func (t *DefaultTransformer) buildQuotedParent(ctx context.Context, parent *teamsMessage) (*cassandra.QuotedParentMessage, error) {
+	s, err := t.resolver.resolve(ctx, parent.From.ID, parent.From.DisplayName)
+	if err != nil {
+		return nil, fmt.Errorf("resolve quoted sender: %w", err)
+	}
+	return &cassandra.QuotedParentMessage{
+		MessageID: deterministicMessageID(parent.RoomID, parent.ID),
+		RoomID:    parent.RoomID,
+		Sender:    cassandra.Participant{ID: s.UserID, Account: s.Account, EngName: s.EngName},
+		CreatedAt: parent.CreatedDateTime,
```
julianshen, line 106: _🎯 correctness_ | _🟠 Warning_ | _`opencode-review`_

Quoted-parent IDs are scoped using parent.RoomID, although a reply's parent belongs to the outer message's conversation and nested parent payloads may omit roomId. That produces an ID different from the separately migrated parent. Pass the outer room ID into buildQuotedParent and use it for both MessageID and RoomID.

<details>

```
In `message-worker/teamstransform.go` around line 140: Quoted-parent IDs are scoped using parent.RoomID, although a reply's parent belongs to the outer message's conversation and nested parent payloads may omit roomId. That produces an ID different from the separately migrated parent. Pass the outer room ID into buildQuotedParent and use it for both MessageID and RoomID.

```

</details>

message-worker/hridentity.go
```diff
@@ -0,0 +1,95 @@
...
+			slog.WarnContext(ctx, "skip user identity upsert: empty employeeId", "account", u.Account)
+			continue
+		}
+		models = append(models, mongo.NewUpdateOneModel().
+			// Site-scoped to match the reads: a migration for another site must create
+			// that site's user, never clobber an existing user's siteId/identity.
+			SetFilter(bson.M{"siteId": u.SiteID, "employeeId": u.EmployeeID}).
+			SetUpdate(bson.M{
+				"$set": bson.M{
+					"account": u.Account, "siteId": u.SiteID,
```
julianshen: _🏷️ data-integrity_ | _🟠 Warning_ | _`opencode-review`_

The lookup-then-upsert sequence is subject to a race: if HR sync or another request inserts the employee between the lookup and this update, $set overwrites that live user's account and names, contradicting the no-clobber intent. Put migration identity fields in $setOnInsert, or otherwise make the conditional creation atomic without updating an existing row.

<details>

```
In `message-worker/hridentity.go` around line 70: The lookup-then-upsert sequence is subject to a race: if HR sync or another request inserts the employee between the lookup and this update, $set overwrites that live user's account and names, contradicting the no-clobber intent. Put migration identity fields in $setOnInsert, or otherwise make the conditional creation atomic without updating an existing row.

```

</details>

docs/superpowers/plans/2026-07-22-teams-message-migration.md
```diff
@@ -0,0 +1,58 @@
...
+**Scope:** Phase 1 (messages) only. The forward branch is stubbed pending the
+`Forwarded` model field; room + subscription replace is Phase 2.
+
+---
+
+## Task 1: Lift shared helpers into `pkg/`
+
+- [x] Move the oplog-transformer's canonical publisher into `pkg/migration.CanonicalPublisher` (`PublishInsert` + `SetMigrationLive`); refactor the oplog-transformer to import it.
+- [x] Move `EmployeeIDFromGraphID` into `pkg/hrstore`; refactor `teams-hr-sync` to import it.
+- [x] Both existing callers + their tests use the lifted helpers (no duplication).
```
julianshen: _🏷️ intent-match_ | _🟠 Warning_ | _`opencode-review`_

The checked-off plan says EmployeeIDFromGraphID was moved into pkg/hrstore and callers were refactored, but this diff instead adds another private employeeIDFromGraphID copy in message-worker/hridentity.go and no pkg/hrstore change. Use the shared helper as documented or correct the plan/design so completed work is not misrepresented.

<details>

```
In `docs/superpowers/plans/2026-07-22-teams-message-migration.md` around line 16: The checked-off plan says EmployeeIDFromGraphID was moved into pkg/hrstore and callers were refactored, but this diff instead adds another private employeeIDFromGraphID copy in message-worker/hridentity.go and no pkg/hrstore change. Use the shared helper as documented or correct the plan/design so completed work is not misrepresented.

```

</details>

message-worker/teamsbatch_integration_test.go
```diff
@@ -0,0 +1,91 @@
...
+	})
+	req := model.TeamsBatchRequest{Messages: []json.RawMessage{raw}}
+	wantID := deterministicMessageID("room-1", "tm-1")
+
+	// Run twice: same batch → same deterministic id → idempotent.
+	for i := 0; i < 2; i++ {
+		reply, err := teams.handleBatch(natsrouter.NewContext(nil), req)
+		require.NoError(t, err)
+		require.Len(t, reply.Results, 1)
+		assert.Equal(t, model.TeamsBatchPersisted, reply.Results[0].Status)
```
julianshen: _🧪 tests_ | _🔵 Suggestion_ | _`opencode-review`_

**Despite claiming to exercise the batch RPC, this test calls teams.handleBatch directly with a synthetic context.**

It does not validate NATS registration, request decoding, reply encoding, or router behavior. Add an actual request/reply integration test against the registered subject.

<details>

```
In `message-worker/teamsbatch_integration_test.go` around line 73: Despite claiming to exercise the batch RPC, this test calls teams.handleBatch directly with a synthetic context. It does not validate NATS registration, request decoding, reply encoding, or router behavior. Add an actual request/reply integration test against the registered subject.

```

</details>

chenjr0719:
Reworked per your direction — message-worker now consumes `chat.msg.canonical.{siteID}.teams.batch` and runs each message through the existing persist pipeline (threads / mentions / quotes), no reply and no re-publish. Broadcast + notification skip it automatically since they key on the `.created` event — the silent migration we want.

One tradeoff to flag before merge: **search indexing.** `search-sync-worker` indexes off the canonical `.created` events, not off Cassandra — so with no `.created` publish, migrated messages persist but won't be indexed in Elasticsearch (not searchable). I've documented that as a phase-1 limitation (bulk history first, search backfill later).

If you'd rather migrated history be searchable at load time, the pipeline-consistent way is to emit a `.created` event carrying `X-Migration: live` — search-sync would index it, and broadcast/notification already skip that header. That's the re-publish you flagged, so I've left it out per your call; happy to add it back if searchability at migration time is required. Let me know which you prefer.

(Also removed the request/reply RPC, the `client-api.md` entry, and dropped `siteId` from the identity upsert per your note — employeeID is globally unique. CodeRabbit re-flagged the upsert site-scoping; I kept it `employeeId`-only per your data-model call.)

mliu33 (approved):

message-worker/main.go
```diff
@@ -202,6 +202,43 @@ func main() {
 		}
 	}()
 
+	// Teams message-history batch migration: a durable consumer on the canonical
+	// stream, filtered to .teams.batch. Each batch is transformed + fed through the
+	// existing persist pipeline (isMigration=true) — NOT re-published to canonical,
+	// so broadcast/notification/search-sync never fire (silent no-fan-out migration).
+	teamsMigration := newTeamsBatchHandler(newMongoHRIdentityStore(db), cfg.SiteID, handler.processMessage)
```
mliu33: I think we can reuse the same Jetstream consumer. Just adding the subject into subject filter. This will not create additional burden for nats server.

In addition, this is a one-time job. Creating an extra durable JS consumer is not ideal.

message-worker/teamsbatch.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	if head.ID == "" {
+		slog.DebugContext(ctx, "teams batch: skip message with no id")
+		res.Status = model.TeamsBatchSkipped
+		return res, nil
+	}
+	// No target room → the deterministic id would collide across conversations and the
+	// message would orphan; skip rather than persist it.
+	if head.RoomID == "" {
+		slog.DebugContext(ctx, "teams batch: skip message with no roomId", "teamsMsgId", head.ID)
```
mliu33: Use warn level for logging

message-worker/teamsbatch.go
```diff
@@ -0,0 +1,144 @@
...
+func (h *teamsBatchHandler) migrateOne(ctx context.Context, tr MessageTransformer, raw json.RawMessage) (model.TeamsBatchResult, error) {
+	head, ok := peekTeamsHead(raw)
+	res := model.TeamsBatchResult{TeamsMsgID: head.ID}
+	if !ok {
+		slog.WarnContext(ctx, "teams batch: malformed message payload")
+		res.Status, res.Error = model.TeamsBatchError, "malformed message payload"
+		return res, nil
+	}
+	if head.ID == "" {
+		slog.DebugContext(ctx, "teams batch: skip message with no id")
```
mliu33: Use warn logging level for any skip ones

message-worker/teamsbatch.go
```diff
@@ -0,0 +1,144 @@
...
+		// Per-message, deterministic → log the chain, record error, but don't Nak the batch.
+		slog.ErrorContext(ctx, "teams batch: transform failed", "teamsMsgId", head.ID, "error", err)
+		res.Status, res.Error = model.TeamsBatchError, "transform failed"
+		return res, nil //nolint:nilerr // per-message transform error is isolated, not a batch Nak
+	}
+	// Teams ids are unique only per conversation → scope by roomId to avoid collisions.
+	msg.ID = teamsmigrate.DeterministicMessageID(head.RoomID, head.ID)
+
+	evt := model.MessageEvent{Event: model.EventCreated, Message: msg, SiteID: h.siteID}
+	data, err := sonic.Marshal(evt)
```
mliu33: Why do we need to json marshal in the worker ? I think there is no need to marshal and process. We should be able to just use model struct and write into Cassandra

message-worker/teamsbatch.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	jsretry.Settle(ctx, msg, jsretry.DefaultBackoff, h.handleBatch(ctx, req))
+}
+
+// handleBatch runs each message; a fresh resolver per call scopes the sender cache to
+// the batch. Returns the first infra error so the caller Naks — per-message transform
+// errors are logged and don't block the batch.
+func (h *teamsBatchHandler) handleBatch(ctx context.Context, req model.TeamsBatchRequest) error {
+	resolver := newSenderResolver(h.store, h.siteID)
+	tr := h.newTransformer(resolver)
```
mliu33: Line 59 and 60 could be initialized in the NewHandler initialization. Not efficient to initialize them in handleBatch function every time.

In addition, we should have global LRU cache shared instead of one cache per batch

message-worker/teamssender.go
```diff
@@ -0,0 +1,111 @@
...
+// never overwrites an existing identity.
+func (r *senderResolver) resolve(ctx context.Context, teamsUserID, displayName string) (resolvedSender, error) {
+	if teamsUserID == "" {
+		return resolvedSender{}, fmt.Errorf("empty teams user id")
+	}
+	if s, ok := r.cache[teamsUserID]; ok {
+		return s, nil
+	}
+
+	empID := teamsmigrate.EmployeeIDFromGraphID(teamsUserID)
```
mliu33: Where do u get this from ? The mapping collection will only exist in central site...right ?

message-worker/teamssender.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	if u != nil {
+		s := senderFromUser(u)
+		r.cache[teamsUserID] = s
+		return s, nil
+	}
+
+	// Create: no UPN exists at the message layer, so account = employeeId; displayName
+	// lands in chineseName to mirror the HR mapping (teams-hr-sync writes it there).
+	nu := model.User{Account: empID, SiteID: r.siteID, EmployeeID: empID, ChineseName: displayName}
```
mliu33: We should apply hashing to have the same format of account as what we have now

message-worker/teamssender.go
```diff
@@ -0,0 +1,111 @@
...
+	if err := r.store.UpsertUserIdentities(ctx, []model.IUserWithChange{{User: nu}}); err != nil {
```
mliu33, line 85: We can't upsert user directly at local site. We need to call server-to-server user-service RPC to upsert a new user.

This user-service RPC should persist into the DB and publish fanout event to all other sites because every site needs to have this user.

Or message-worker could do the fanout job directly here without making another rpc call. This might be easier and more efficient as one-time work

message-worker/teamstransform.go
```diff
@@ -0,0 +1,132 @@
...
+type MessageTransformer interface {
+	Transform(ctx context.Context, raw json.RawMessage) (model.Message, error)
+}
+
+// DefaultTransformer is the built-in Teams transform. Inject a custom
+// MessageTransformer to migrate a different source. The payload types + payload-only
+// helpers live in pkg/teamsmigrate; the sender resolution stays here (it needs Mongo).
+type DefaultTransformer struct {
+	resolver identityResolver
+}
```
mliu33, line 25: The mention resolver should follow the same approach as sender resolver because mentioned user might not exist

pkg/teamsmigrate/teamsmigrate.go
```diff
@@ -0,0 +1,110 @@
...
+	var oid bson.ObjectID
+	copy(oid[:], sum[:12])
+	return oid.Hex()
+}
+
+// DeterministicMessageID is a stable, valid message id derived from the Teams id
+// scoped by conversation, so a batch re-run overwrites the same row (idempotent).
+// Teams message ids are unique only within a conversation, so roomId (the nextgen
+// room, 1:1 with the Teams chat) is folded into the seed to avoid cross-chat collisions.
+func DeterministicMessageID(chatScope, teamsID string) string {
```
mliu33, line 72: Why not just use Teams message id alone, which is already unique and should be sufficient

search-sync-worker/teams_migration.go
```diff
@@ -0,0 +1,99 @@
...
+	return &teamsMigrationCollection{indexPrefix: indexPrefix, siteID: siteID, devMode: devMode}
+}
+
+// Same MESSAGES_CANONICAL stream as messageCollection; only the subject filter differs.
+func (c *teamsMigrationCollection) StreamConfig(siteID string) jetstream.StreamConfig {
+	return userMessagesStreamCfg(siteID)
+}
+
+// A distinct durable so its cursor is independent of the .created (message-sync) consumer.
+func (c *teamsMigrationCollection) ConsumerName() string { return "message-sync-teams" }
```
mliu33: Let's not create a new Durable Js consumer. Reuse the existing one with additional subject in the subject filter.

teams-hr-sync/write_store.go
```diff
@@ -78,7 +77,10 @@ func (s *mongoWriteStore) UpsertUserIdentities(ctx context.Context, users []mode
 					"engName": u.EngName, "chineseName": u.ChineseName,
 					"employeeId": u.EmployeeID,
 				},
-				"$setOnInsert": bson.M{"_id": idgen.GenerateUUIDv7()},
+				// _id = employeeId (the stable per-employee id, matching the employees
+				// upsert + message-worker's migration) so a user's id is derivable from
+				// the Teams id — search-sync indexes migrated messages without a Mongo lookup.
+				"$setOnInsert": bson.M{"_id": u.EmployeeID},
```
mliu33, line 83: Let's try to hash into current 17 char base62 string

mliu33:
@julianshen this PR is a good example why human review is still needed lol

chenjr0719:
On cross-site user creation (your `teamssender.go` comment) — I traced the codebase and there's **no server-to-server create-user RPC** today. `user-service` exposes only read/set RPCs (`UserMe`, `UserStatus`, `UserSettings`, subscriptions, SSO); the only create path is `admin-service`'s local HTTP `POST /users`, which forces the local `siteId` and does **no fanout**. The **only** mechanism that creates a `users` doc AND reaches all sites is the HR feed: `chat.hr.{central}.users.upsert` → each site's `hr-sync-worker` upserts locally — and it's **employee-only** (keyed on `employeeId`; empty ones are skipped). `hr_employee` + `users` are both per-site (each site's `hr-sync-worker` writes them from the central feed).

So the plan for the migration's sender resolution is: look up `users` → else `hr_employee` (an **employee** → reuse the HR-feed `users.upsert` publish, no new RPC) → else it's an **external / non-employee** sender.

The open question is that **external / non-employee** case — there's no existing cross-site create path for them (auth doesn't auto-provision, bots/rooms/DMs don't create users, `teams_user` is a local lookup table, not the federated `users` collection). How would you like external senders handled?
- (a) extend the `users.upsert` HR lane to accept a synthetic non-employee key, or
- (b) a new create-user publish/RPC that fans out to all sites, or
- (c) scope this migration to employee senders only (external out of scope).

Happy to build whichever you prefer.

message-worker/teamssender.go
chenjr0719: Going with your "message-worker fans out directly" alternative, reusing the existing HR lane rather than a new RPC: on a not-found sender, message-worker publishes an `IUserWithChange` (with `employeeId = EmployeeIDFromGraphID(teamsUserId)` — the same hash `teams-hr-sync` uses) to `chat.hr.{central}.users.upsert`. The existing hr-sync-workers then create the user at every site — no new user-service RPC or inbox handler. Because the hash is deterministic, an employee created this way **converges** with the HR sync (same `_id`/employeeId), and the message can stamp `UserID = hash` immediately (eventually-consistent create). This also covers the mention resolve-or-create the same way.

One concern to flag on the **external / non-employee** case: those senders aren't in the HR feed, so they'd land in the `users` collection via this lane carrying a **synthetic `employeeId`** (the teams-id hash). Functionally fine and it converges, but semantically they aren't real HR employees. Acceptable, or would you prefer a dedicated non-employee create path (e.g. a `user.upserted` inbox event) for external users? Building the HR-lane version now — easy to switch if you'd rather.

message-worker/hridentity.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+		models = append(models, mongo.NewUpdateOneModel().
+			// employeeId is globally unique, so it alone keys the identity — a person
+			// resolves to one row regardless of site.
+			SetFilter(bson.M{"employeeId": u.EmployeeID}).
+			SetUpdate(bson.M{
+				"$set": bson.M{
+					"account": u.Account, "siteId": u.SiteID,
+					"engName": u.EngName, "chineseName": u.ChineseName,
+					"employeeId": u.EmployeeID,
```
mliu33: Set filter on employeeID, but also set employeeId ?

message-worker/teamsbatch.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		slog.WarnContext(ctx, "teams batch: skip message, transform failed", "teamsMsgId", head.ID, "error", err)
+		res.Status, res.Error = model.TeamsBatchError, "transform failed"
+		return res, nil //nolint:nilerr // per-message transform error is isolated, not a batch Nak
+	}
+	// Teams ids are unique only per conversation → scope by roomId to avoid collisions.
+	msg.ID = teamsmigrate.DeterministicMessageID(head.RoomID, head.ID)
+
+	// Cache hit after Transform (same sender key), so it adds no store round trip.
+	sender, err := h.resolver.resolve(ctx, head.From.ID, head.From.DisplayName)
```
mliu33, line 110: Remember to change to use "account" field to lookup sender later on

message-worker/hridentity.go
```diff
@@ -0,0 +1,87 @@
...
+		if u.EmployeeID == "" {
+			// employeeId is the identity key; an empty one would clobber every keyless row.
+			// account omitted — it may be an employee identifier (privacy).
+			slog.WarnContext(ctx, "skip user identity upsert: empty employeeId")
+			continue
+		}
+		models = append(models, mongo.NewUpdateOneModel().
+			// employeeId is globally unique, so it alone keys the identity — a person
+			// resolves to one row regardless of site.
+			SetFilter(bson.M{"employeeId": u.EmployeeID}).
```
mliu33: For user collection, use "account" field as filter

message-worker/teamssender.go
mliu33: Yes, so these external users should not have employeeID field set in user collection. 

- publish to hr user upsert subject. The user doc should not have employeeId field

For user collection, the upsert filter should always use "account" field,which is global unique

search-sync-worker/messages.go
```diff
@@ -137,6 +160,49 @@ func (c *messageCollection) resolveThreadParentCreatedAt(evt *model.MessageEvent
...
+			continue // system message — not indexed content
+		}
+		// UserAccount reuses the employeeId hash best-effort (no UPN at the message layer).
+		empID := teamsmigrate.EmployeeIDFromGraphID(tm.From.ID)
+		doc := MessageSearchIndex{
+			MessageID:   teamsmigrate.DeterministicMessageID(tm.RoomID, tm.ID),
+			RoomID:      tm.RoomID,
+			SiteID:      c.siteID,
+			UserID:      empID,
+			UserAccount: empID,
```
mliu33: Make sure to lookup correct user account. And make sure the message index doc format is the same as existing one

message-worker/teamssender.go
mliu33: Please fix this part

mliu33:
Make remaining fixes in separate PR

PR #12, opened 2026-07-14
fix(room-worker,inbox-worker): scrub thread state on member removal (#308)

hmchangw:
## Problem

`replyAccounts` is append-only; nothing ever removed an entry. When a user left a room their `subscriptions` doc was deleted but they stayed in `replyAccounts` (and their `thread_subscriptions` lingered), so they remained a thread "follower" indefinitely and kept receiving live thread replies.

## Change

Scrub thread state on **every removal path that truly ends membership**, so thread follower state no longer outlives room membership:

- **`room-worker`** (home site, owns `thread_rooms`): after the individual-only and org removal paths delete the subscription, `$pull` the departed accounts from `replyAccounts` across the room's threads (`PullThreadFollowers`) and delete their `thread_subscriptions` (`DeleteThreadSubscriptions`). The org path reuses the already-computed "truly removed" set, so dual-membership / sibling-org survivors are never scrubbed.
- **`inbox-worker`** (remote site, holds no `thread_rooms`): on a federated `member_removed` event, delete the removed accounts' `thread_subscriptions` too.

Both new store methods are no-ops on empty input and scoped to `(roomId, userAccount)`, so a user's thread state in other rooms is never touched.

## Testing
- Unit tests (TDD) for each removal path — including the dual-member / sibling-org "not scrubbed" cases and the federated `thread_subscriptions` cleanup.
- Integration tests for the new store methods (`$pull` and delete, with cross-room isolation and empty-input no-op).
- `go vet` clean; unit and integration suites green.

## Relationship to #309 (PR #108)
This is the **source-side** half of the thread-reachability invariant "a non-member must not receive a room's thread events": it keeps the follower data honest. The complementary **delivery-side** gate — intersecting channel thread fan-out recipients with current membership — was split into its own PR (#108, #309). The two are independent at the code level (disjoint services) and can merge in either order; together they close the leak from both ends.

> Note: previously this PR also contained the #309 fan-out gate (`broadcast-worker`). That change was split out into PR #108 so each issue lands independently.

## Summary by CodeRabbit

* **Bug Fixes**
  * When members leave rooms, thread follower data and per-thread read-state subscriptions are cleaned up, reducing stale notifications and read-state inconsistencies (with correct handling for individual vs. org removal, and dual-membership / sibling-org survivors preserved).

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed departed members’ thread followers and read-state data when they leave a room.
[...]

PR #108, opened 2026-07-22
fix(broadcast-worker): gate channel thread fan-out on membership (#309)

hmchangw:
## Problem

`broadcast-worker`'s `channelThreadFanOut` built its channel-thread recipient list from `thread_rooms.replyAccounts` (thread followers) plus the reply's `@`-mentions and published a per-account event to each recipient's personal subject — with **no check against current room membership**. The push path (`notification-worker`) *did* bound recipients to members, so the live broadcast could leak to accounts the notification path correctly skipped.

## Change

Gate channel thread fan-out on membership, enforced at delivery so no thread handler can bypass it:

- **Followers** (from `replyAccounts`) are trusted as current members — kept member-clean at the source by the removal-path scrub (#308, PR #12) — and delivered directly.
- **`@`-mentions** are *not* trusted (a reply can mention a non-member), so they are intersected against current membership via `store.FilterRoomMembers` (a projected `$in` on the `(roomId, u.account)` index, through the shared `pkg/pipelines.SubscribedAccounts`).
- The **sender** is always included (multi-device echo), added directly rather than via `replyAccounts` so it is race-free against the async `replyAccounts` writer.

The membership query runs **only when the reply has `@`-mentions**, and is scoped to just the mentioned accounts — so the common no-mention thread reply issues **zero** membership queries, and the query never scales with room size. The previous history-window gate (`GetHistorySharedSince` + `mentionVisible`) and the per-event parent resolution (including the history-service round-trip via `parentFetcher.FetchParent`) are removed, since the membership gate needs neither the parent's timestamp nor its author.

## Behavior preserved
Sender-first ordering, deduplication, and bot exclusion are unchanged. DM thread fan-out (which already derives recipients from `ListSubscriptions`) is untouched. This covers the channel branch of every thread handler — created, edited, and deleted.

## Relationship to #308 (PR #12)
This is the **delivery-side** half of the thread-reachability invariant "a non-member must not receive a room's thread events." #308 (PR #12) keeps the follower data honest at the **source** (removal scrubs `replyAccounts` / `thread_subscriptions`); this PR enforces the invariant at the **boundary** and backstops any residual drift. The two are independent at the code level (disjoint services) and can merge in either order; together they close the leak from both ends.

## Testing
- Unit tests (TDD) for the fan-out gate: no-mention (zero membership queries), member vs non-member mentions, sender echo, dedup, and bot exclusion.
- Integration test for the new `FilterRoomMembers` store method.
- `go vet` clean; unit and integration suites green.

## Out of scope
Mentions of non-members are filtered at **delivery** here, not rejected at **ingress** — moving mention validation into `message-gatekeeper` is #307.

https://claude.ai/code/session_01154fUerPxSjncqjPUbGEuX

---
_Generated by [Claude Code](https://claude.ai/code/session_01154fUerPxSjncqjPUbGEuX)_

### DIFF
--- a/broadcast-worker/handler.go
+++ b/broadcast-worker/handler.go
@@ -228,7 +228,7 @@ func (h *Handler) handleThreadCreated(ctx context.Context, evt *model.MessageEve
 	// the recipient set.
 	var fanOut []string
 	if meta.Type == model.RoomTypeChannel {
-		fanOut, err = h.channelThreadFanOut(ctx, msg.RoomID, evt.SiteID, parentMsgID, msg.UserAccount, parsed.Accounts, msg.ThreadParentMessageCreatedAt, evt.ThreadParentSenderAccount)
+		fanOut, err = h.channelThreadFanOut(ctx, meta.ID, parentMsgID, msg.UserAccount, parsed.Accounts)
 		if err != nil {
 			return fmt.Errorf("channel thread fan-out for parent %s: %w", parentMsgID, err)
 		}
@@ -328,7 +328,7 @@ func (h *Handler) handleThreadUpdated(ctx context.Context, evt *model.MessageEve
 	switch room.Type {
 	case model.RoomTypeChannel:
 		parsed := mention.Parse(msg.Content)
-		fanOut, err := h.channelThreadFanOut(ctx, room.ID, room.SiteID, parentMsgID, msg.UserAccount, parsed.Accounts, msg.ThreadParentMessageCreatedAt, evt.ThreadParentSenderAccount)
+		fanOut, err := h.channelThreadFanOut(ctx, room.ID, parentMsgID, msg.UserAccount, parsed.Accounts)
 		if err != nil {
 			return fmt.Errorf("channel thread fan-out for thread update of parent %s: %w", parentMsgID, err)
 		}
@@ -381,7 +381,7 @@ func (h *Handler) handleThreadDeleted(ctx context.Context, evt *model.MessageEve
 		// receive the delete. Only the channel path uses mentions; the DM path
 		// fans out to all members.
 		parsed := mention.Parse(msg.Content)
-		fanOut, err := h.channelThreadFanOut(ctx, room.ID, room.SiteID, parentMsgID, msg.UserAccount, parsed.Accounts, msg.ThreadParentMessageCreatedAt, evt.ThreadParentSenderAccount)
+		fanOut, err := h.channelThreadFanOut(ctx, room.ID, parentMsgID, msg.UserAccount, parsed.Accounts)
 		if err != nil {
 			return fmt.Errorf("channel thread fan-out for thread delete of parent %s: %w", parentMsgID, err)
 		}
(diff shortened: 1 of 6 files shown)


julianshen:
**Actionable comments posted: 2** — 🔴 2 blocking

<details>

* `broadcast-worker/mock_store_test.go`
* `broadcast-worker/store.go`

</details>
<details>

* `opencode-review`

</details>

broadcast-worker/handler.go
```diff
@@ -982,73 +979,35 @@ func threadFanOutAccounts(senderAccount, parentSenderAccount string, followers m
...
-	return threadFanOutAccounts(sender, parent.SenderAccount, followers, allowed), nil
+	var memberMentions map[string]struct{}
+	if len(mentions) > 0 {
+		memberMentions, err = h.store.FilterRoomMembers(ctx, roomID, mentions)
+		if err != nil {
+			return nil, fmt.Errorf("filter mentioned members for room %s: %w", roomID, err)
+		}
+	}
+	return threadFanOutAccounts(sender, followers, mentions, memberMentions), nil
 }
```
julianshen, line 1012: _🎯 correctness_ | _🔴 Blocking_ | _`opencode-review`_

**The new recipient set no longer includes the thread parent author.**

On the first reply, GetThreadFollowers can legitimately be empty because replyAccounts is populated asynchronously; the removed regression test explicitly covered this case. Consequently, the parent author can miss the first reply entirely. Preserve the parent sender from the event or FetchParent fallback and add it to threadFanOutAccounts.

<details>

```
In `broadcast-worker/handler.go` around line 1012: The new recipient set no longer includes the thread parent author. On the first reply, GetThreadFollowers can legitimately be empty because replyAccounts is populated asynchronously; the removed regression test explicitly covered this case. Consequently, the parent author can miss the first reply entirely. Preserve the parent sender from the event or FetchParent fallback and add it to threadFanOutAccounts.

```

</details>

broadcast-worker/handler.go
```diff
@@ -982,73 +979,35 @@ func threadFanOutAccounts(senderAccount, parentSenderAccount string, followers m
...
 	followers, err := h.store.GetThreadFollowers(ctx, parentMsgID)
 	if err != nil {
 		return nil, fmt.Errorf("get thread followers for parent %s: %w", parentMsgID, err)
 	}
-	return threadFanOutAccounts(sender, parent.SenderAccount, followers, allowed), nil
+	var memberMentions map[string]struct{}
+	if len(mentions) > 0 {
+		memberMentions, err = h.store.FilterRoomMembers(ctx, roomID, mentions)
+		if err != nil {
+			return nil, fmt.Errorf("filter mentioned members for room %s: %w", roomID, err)
```
julianshen, line 1007: _🔒 security_ | _🔴 Blocking_ | _`opencode-review`_

Replacing GetHistorySharedSince/mentionVisible with a current-membership-only check removes the existing history-access gate. A member who joined after the thread parent and is not permitted to see it can now receive the thread reply when mentioned, leaking content and thread metadata. Keep the query scoped to mentioned accounts, but retain validation that each member's history window admits the parent; restore coverage for restricted members.

<details>

```
In `broadcast-worker/handler.go` around line 1007: Replacing GetHistorySharedSince/mentionVisible with a current-membership-only check removes the existing history-access gate. A member who joined after the thread parent and is not permitted to see it can now receive the thread reply when mentioned, leaking content and thread metadata. Keep the query scoped to mentioned accounts, but retain validation that each member's history window admits the parent; restore coverage for restricted members.

```

</details>

hmchangw:
Was implemented elsewhere

PR #116, opened 2026-07-23
thread.list: carry message bodies as opaque json.RawMessage

saurav-err-137:
## Summary

Fixes a latent correctness bug in `user-service`'s `ListUserThreads` RPC where threads with reacted parent or last messages silently disappear from the inbox. The root cause: `cassandra.Message.Reactions` is a struct-keyed map with no `UnmarshalJSON`, so decoding a leaf response carrying reactions fails and marks the entire site unavailable.

The fix removes the decode entirely on the path that never needs it: retype `ThreadListItem.ParentMessage` and `.LastMessage` from `*cassandra.Message` to `json.RawMessage`, have `history-service` pre-marshal the bodies at build time, and let `user-service` forward the bytes untouched. The client-facing wire JSON is byte-for-byte unchanged — still the full `Message` object.

## Changes

- **`pkg/model/threadlist.go`**: Retype `ParentMessage` and `LastMessage` fields to `json.RawMessage`; swap import from `cassandra` to `encoding/json`. Wire form unchanged; `omitempty` preserves "absent when not hydrated" behavior.

- **`pkg/model/threadlist_test.go`**: Rewrite `TestThreadListItemJSON_WithMessages` to build items with marshaled bodies and assert round-trip fidelity. Add `TestThreadListItemJSON_MessageWithReactions_RoundTrips` regression test — verifies a reactions-bearing body survives JSON marshal/unmarshal without error (the old typed field would have failed here).

- **`history-service/internal/service/threads.go`**: In `buildThreadItems`, marshal each hydrated `models.Message` to `json.RawMessage` bytes before assignment. Skip the row with a warning on marshal error (effectively infallible but guards resilience; consistent with existing "skip half-hydrated rows" rule).

- **`history-service/internal/service/threadlist_test.go`**: Add `decodeThreadMsg` helper to unmarshal opaque bodies in assertions. Update existing assertions to decode before reading typed fields. Add `TestHistoryService_ListThreadSubscriptions_ParentWithReactions` to verify reactions-bearing bodies build and carry through as grouped-by-emoji wire form.

- **`user-service/service/threads_test.go`**: Add `TestUserService_ListUserThreads_ForwardsMessageBodiesWithReactions` regression test — asserts a reactions-bearing body aggregates and forwards verbatim without marking the site unavailable. No production change: `historyclient.GetThreadList` now copies `json.RawMessage` bytes instead of parsing.

## Implementation Details

- **TDD**: All tests written first (Red), then implementation (Green). Compilation coupling intentional: Task 1 retyping breaks `history-service` until Task 2 lands; full suite green only at Task 3.
- **Error handling**: Marshal failures in `buildThreadItems` logged with `slog.WarnContext` (structured fields, no secrets), row skipped to keep page resilient.
- **Coverage**: Happy path + reactions regression at all three levels (model round-trip, builder, aggregator forward). Minimum 80% coverage maintained.
- **Docs**: No change to `docs/client-api.md` — wire-preserving retype of a server-to-server carrier; client sees the same `Message` object under `parentMessage`/`lastMessage`.

All tests pass with race detector; `make lint` clean.

https://claude.ai/code/session_012qe55SeznCzmAoMdtrHa3n

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved thread list reliability for messages containing reactions.
  * Preserved parent and latest message content unchanged when forwarding thread data.
  * Prevented affected thread entries from being omitted due to message reaction data.
* **Documentation**
  * Added implementation and design documentation for reliable message forwarding in thread lists.

### DIFF
--- a/docs/superpowers/plans/2026-07-23-user-service-opaque-thread-messages.md
+++ b/docs/superpowers/plans/2026-07-23-user-service-opaque-thread-messages.md
@@ -1,0 +1,392 @@
+# Opaque thread.list Message Bodies — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Carry `ThreadListItem.ParentMessage`/`LastMessage` as opaque `json.RawMessage` so `user-service` forwards thread.list message bodies without decoding `cassandra.Message.Reactions` (a struct-keyed map with no JSON decoder), and have `history-service` emit them pre-marshaled.
+
+**Architecture:** Retype the two shared fields on `pkg/model.ThreadListItem` from `*cassandra.Message` to `json.RawMessage`. `history-service` (`buildThreadItems`) marshals each typed body to bytes at build time; `user-service` needs no production change — the raw bytes pass through decode and re-encode untouched. The client-facing wire JSON is byte-for-byte unchanged (still the full `Message` object), so no docs edit is required.
+
+**Tech Stack:** Go 1.25, `encoding/json` (stdlib), `stretchr/testify`, `go.uber.org/mock`. Tests run via `make test SERVICE=<name>` (race detector on by default).
+
+## Global Constraints
+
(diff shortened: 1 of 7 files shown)


mliu33 (approved):
Great work. Thanks!

PR #120, opened 2026-07-24
feat(es-index-migrator): add ES index rebuild/backfill service

ngangwar962:
## Summary

- Extracts `search-sync-worker`'s previously-private ES document shapes, Painless scripts, and bulk-result classification into shared `pkg/searchindex`/`pkg/searchengine` packages (Phase A), with `search-sync-worker` becoming a thin caller of the same code — verified behavior-preserving.
- Adds a new standalone service, `data-migration/es-index-migrator` (Phase B): a one-time/rebuild batch job that backfills a site's `messages`/`spotlight`/`user-room` Elasticsearch indexes directly from that site's own current-stack Cassandra (`messages_by_room`, streamed bucket-by-bucket) and MongoDB (`subscriptions`), reusing Phase A's shared packages so its output is indistinguishable from what `search-sync-worker`'s live path would produce. No dependency on `pkg/atrest`/Vault — this job's source rows were populated with plaintext content directly, never through the live at-rest-encryption write path.
- Design doc: `docs/search_index_migration_spec.md`. Implementation plan: `docs/superpowers/plans/2026-07-24-es-index-migrator.md`.

Built via 16 TDD tasks, each independently reviewed, plus a final whole-branch review with one fix wave addressing its findings (concurrency, test fidelity, live-worker parity — see Test plan).

## Test plan

- [x] `go build ./...`, `go vet -tags integration ./...`, `golangci-lint run ./...` all clean across the whole repo
- [x] `go test ./... -race` clean across the whole repo (unit tests only)
- [x] Every new/changed unit test passes with `-race`; the concurrent-collection fix was additionally stress-tested with `-race -count=20`
- [ ] **Run the `//go:build integration` suite against real Cassandra/MongoDB/Elasticsearch in CI** — this session's sandbox has Docker Hub image pulls blocked by network policy, so `messagesource_integration_test.go`, `subscriptionsource_mongo_integration_test.go`, and `integration_test.go` were verified by careful reading, schema cross-checks against `docs/cassandra_message_model.md`, and `go vet -tags integration`, but never actually executed against live containers
- [ ] **Confirm package coverage ≥ 80%** (CLAUDE.md floor) once integration tests can run — unit-only coverage sits at 58.7%–74.3% for `data-migration/es-index-migrator` because the Cassandra/Mongo source readers and `main.go`'s wiring are covered only by the not-yet-executed integration tests
- [ ] **Manually verify `make up SERVICE=data-migration/es-index-migrator`** brings the service up cleanly against a live `make deps-up` stack — the compose file was only validated statically (`docker compose config` resolution, hostname/network cross-check against `docker-local/compose.deps.yaml`)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_018zn5WjL8K7QmoprVVJUzhQ

---
_Generated by [Claude Code](https://claude.ai/code/session_018zn5WjL8K7QmoprVVJUzhQ)_

## Summary by CodeRabbit

* **New Features**
  * Added a one-time Elasticsearch index migration tool to rebuild messages, spotlight, and user-room indexes.
  * Idempotently bootstraps templates and stored scripts, then runs concurrent backfills with robust bulk indexing and accurate failure/skipped-row counting.
  * User-room indexing now includes bot subscriptions (no longer skipped).
* **Documentation**
  * Added migration specifications and an implementation plan.
* **Tests**
  * Expanded unit and integration coverage, including idempotency verification on reruns.
* **Chores**
  * Added a CI pipeline with coverage enforcement, multi-stage Docker build, and a local docker-compose one-shot runner.


pkg/searchindex/template.go
```diff
@@ -0,0 +1,206 @@
...
+// UserRoomTemplateName returns the ES index template name for the
+// user-room index. Unlike messages/spotlight, user-room index names are
+// unversioned (a single fixed index), so no version-stripping applies.
+func UserRoomTemplateName(indexName string) string {
+	return fmt.Sprintf("%s_template", indexName)
+}
+
+// UserRoomTemplateBody builds the ES index template for user-room; index_patterns is the exact
+// configured index name so a custom USER_ROOM_INDEX still maps correctly, and roomTimestamps is `flattened` to avoid per-key mapping bloat.
+func UserRoomTemplateBody(indexName string) json.RawMessage {
```
Joey0538: please add devMode bool param here

and the below part at the start of the method
replicas := 1
if devMode {
	replicas = 0
}

Joey0538 (commented):
## Code Review

Nice extraction — Phase A is a clean lift-and-shift and Phase B reuses it faithfully. Two items I'd want addressed before merge, plus some suggestions and nits below.

---

### 🔴 Blocker 1 — Silent data loss on per-row build errors

**File:** `data-migration/es-index-migrator/runner.go` (~lines 1750-1767, 1789-1800, 1819-1830)

When `buildMessageAction` / `buildSpotlightAction` / `buildUserRoomAction` returns an error, the row is logged and the callback returns `nil`. Nothing increments `flusher.failedCount`, so `main.go:889` (`if failed > 0`) exits **0** on a run that silently dropped rows. For a rebuild whose whole purpose is completeness, this is the wrong default.

**Suggested fix — either count skips or fail-fast:**
```go
action, err := buildMessageAction(msg)
if err != nil {
    slog.ErrorContext(ctx, "build message action", "error", err, "messageID", msg.ID)
    flusher.recordSkipped(1) // ensure exit code reflects reality
    return nil
}
```

---

### 🔴 Blocker 2 — Coverage below the 80% floor

**File:** `data-migration/es-index-migrator/deploy/azure-pipelines.yml` (~line 411)

PR body flags service-package coverage at 58.7–74.3%; the pipeline enforces 80%. Either the integration run brings it above the line on a Docker-enabled runner (unverified in the authoring session), or CI will red-flag on push. Please confirm before merge — and if this batch-job shape genuinely can't reach 80% via integration coverage, add an explicit justified override in the pipeline.

---

### 🟡 Suggestion — Unbounded memory on subscription load

**File:** `data-migration/es-index-migrator/subscriptionsource_mongo.go` (lines 2156 `Distinct`, 2175 `cur.All`)

Both peak the full result set in memory. For a large tenant this is a scale ceiling.

**Suggested fix — stream with a cursor:**
```go
cur, err := s.coll.Find(ctx, filter, options.Find().SetProjection(bson.M{"roomId": 1}))
if err != nil { return fmt.Errorf("find subscriptions: %w", err) }
defer cur.Close(ctx)
for cur.Next(ctx) {
    var sub Subscription
    if err := cur.Decode(&sub); err != nil { return fmt.Errorf("decode subscription: %w", err) }
    if err := yield(sub); err != nil { return err }
}
return cur.Err()
```

---

### 🟡 Suggestion — Room-wide abort on a single encrypted row

**File:** `data-migration/es-index-migrator/messagesource_cassandra.go` (line 1176, `rejectEncryptedMessage`)

First encrypted row aborts the room's stream, and errgroup's cancel-on-first-error can then skip untouched rooms. Can't distinguish "stray row" from "site already went live". Consider a soft-fail threshold instead of hard-abort-on-1.

```go
if encryptedCount > cfg.MaxEncryptedRowsPerRoom {
    return fmt.Errorf("room %s: encrypted rows exceed threshold — site likely already live", roomID)
}
slog.WarnContext(ctx, "skipping encrypted row", "roomID", roomID)
continue
```

---

### 🟡 Suggestion — Bootstrap overwrites ES templates every run

**File:** `data-migration/es-index-migrator/main.go` (`bootstrapPrerequisites`)

Every invocation re-PUTs three templates + two scripts, clobbering any operator tweak. Consider gating behind a `BOOTSTRAP_ES=true` env flag consistent with the repo's `BOOTSTRAP_STREAMS` pattern, or at minimum call it out in the runbook.

---

### 🟢 Nit — `buildTeamsActions` bypasses `NewMessageDoc`

**File:** `search-sync-worker/messages.go` (lines 7079-7095)

Constructs `searchindex.MessageDoc` directly, skipping `IsBot`/`AttachmentText` population — breaks the shared builder's implicit invariant. Either route through `NewMessageDoc` or add a comment on `MessageDoc` noting direct construction is a supported path.

---

### 🟢 Nit — Duplicated Cassandra setup helper

- `data-migration/es-index-migrator/integration_test.go` lines 693-742 (`setupCassandraForRun`)
- `data-migration/es-index-migrator/messagesource_integration_test.go` lines 1319-1372 (`setupCassandra`)

Near-duplicates. DRY into one helper in a shared `_test.go`.

---

### 🟢 Nit — Fragile driver-panic test

**File:** `data-migration/es-index-migrator/subscriptionsource_mongo_test.go` (line 2277)
**Test:** `TestNewMongoSubscriptionSource_NilDatabasePanics`

Asserts on the v2 Mongo driver panicking on `nil.Collection(...)` — a driver change would silently break it. Add an explicit `if db == nil { return nil, errors.New("nil database") }` in the constructor and assert on that instead.

---

### 🟢 Nit — Redundant helper in `spotlight.go`

**File:** `data-migration/es-index-migrator/spotlight.go` (lines 7541-7547)

`convertJoinedAt` does what an inline `if evt.JoinedAt > 0 { time.UnixMilli(...).UTC() }` does two lines above. Inline it and drop the `joinedAt int64` intermediate.

---

Overall structure and test discipline are strong — fix the silent-failure counting and confirm CI coverage, and this is good to land.

---

Joey0538 (approved):
Great work!

Issue #124, opened 2026-07-25
Teams migration follow-up: fan out newly-created external users (account-keyed)


chenjr0719:
## Summary

Follow-up to the Teams message-history migration consumer (#107, merged). Per the merge note *"make remaining fixes in separate PR"* — this tracks the one confirmed defect plus a design decision it depends on.

## Defect — a newly-created migrated user is not fanned out to other sites

When the migration resolver meets a Teams sender who is **not** already a known user (an external user, no `employeeId`), it creates the user with a **local Mongo write only** — no fanout:

- `message-worker/hridentity.go:58` `UpsertUserIdentities` → `s.users.BulkWrite` (local `users` collection).
- Resolver calls it directly for a not-found sender (`teamssender.go`).

The migration is a one-time central-site job, but by design every site needs the user doc (to render the sender on a migrated message). So a newly-created external user lands **only at the central site** and is missing elsewhere.

## The constraint that makes this non-trivial

The existing user fanout path is **employeeId-keyed and skips keyless users**, so migrated externals can't simply ride it:

- Producer (`teams-hr-sync/publisher.go:61`) only publishes users derived from **employees** (`UserFromEmployee`) — all carry an `employeeId`.
- Consumer (`hr-sync-worker/store.go`) `UpsertUserIdentities` filters by **`employeeId`** and **skips empty ones** (`store.go:75` *"skip user identity upsert: empty employeeId"*; filter `store.go:80`; `_id` derives from employeeId).

A migrated external user has **no `employeeId`** (that was the whole point of #107's account-keying), so publishing it to the current `chat.hr.{central}.users.upsert` fanout would be **skipped at every consuming site**.

Note on identity: user `_id` in `users` is **per-site** (each site `$setOnInsert`s its own `GenerateUUIDv7()` — `hr-sync-worker/store.go:87`), while **`account` is the stable cross-site key**, and messages carry `userAccount` as the cross-site identity (`pkg/model/message.go:92`). So the fanout only needs to make the user *exist* per-site keyed by account; the per-site `_id` difference is expected.

## Fix — account-key the fanout for keyless externals

Extend the fanout to keyless users, mirroring #107's account-keying:

- **`hr-sync-worker`'s `UpsertUserIdentities`** (the fanout consumer): add a branch for a keyless external — when `employeeId` is empty, filter by `account` (globally unique) and `$setOnInsert` a fresh `_id`, instead of skipping. Employee rows keep the `employeeId`-keyed path unchanged.
[...]

PR #125, opened 2026-07-25
feat: converge migrated Teams external user identity (deterministic _id, account-keyed upsert)

chenjr0719:
## Summary

A migrated Teams external user (a sender/member with no HR record) was resolved by writing a user locally **and** publishing a fanout, each site minting its own random `_id` — so the same person split across sites.

Per review, the identity is now publisher-owned and converges:

- The resolver — both the **message sender** (`message-worker`) and the **room-creation member** (`room-worker`) paths — sets a **deterministic `_id`** from the Graph id (`idgen.DeterministicID`) and publishes it. **Nothing is written locally first**, so a redelivery re-derives the same `_id` instead of splitting the user.
- Migrated externals carry **no `employeeId`** — they aren't employees.
- `hr-sync-worker` `UpsertUserIdentities` now upserts by **`account`** (the globally-unique key that covers HR-feed users and externals alike), honoring the publisher-supplied `_id` on insert (empty → mint per-site, unchanged for the HR feed).

Refs #124. Follows up the merged #132 (room-creation consumer) with the identity alignment mliu33 asked for.

## Changes

- `message-worker/teamssender.go` — resolver sets deterministic `_id`, no `employeeId`, publish-first, no local write; removes the now-dead local `UpsertUserIdentities` (interface, impl `hridentity.go`, mock).
- `room-worker/teamsroomcreate.go` — same in `resolveMember`: deterministic `_id`, no `employeeId`.
- `hr-sync-worker/store.go` — `UpsertUserIdentities` keys on `account` (was `employeeId`); sets `employeeId` only when present; `$setOnInsert` honors the published `_id`.

## NATS subjects

- Publishes: `chat.hr.{siteID}.users.upsert` (external-user fanout, publish-first). Consumed by every site's `hr-sync-worker`.

## Verification

- `chat-lint` 0 issues; `gosec` (matches CI sast) clean; `message-worker` + `room-worker` + `hr-sync-worker` unit tests green; integration tests compile (`go vet -tags integration`), including an `hr-sync-worker` integration case asserting a migrated external (no `employeeId`) is upserted account-keyed with the publisher-supplied `_id` honored, and an empty-account row skipped.

🤖 Generated with Claude Code

## Summary by CodeRabbit

* **New Features**
  * Improved Teams migration identity handling with stable user identities.
  * Newly discovered external users are published for synchronization and retain normalized employee information.
* **Bug Fixes**
  * Existing identities are updated instead of duplicated.
  * Records without required account information are skipped.
  * Migrated external users no longer receive an employee ID incorrectly.

### DIFF
--- a/message-worker/teamssender.go
+++ b/message-worker/teamssender.go
@@ -7,6 +7,7 @@ import (
 	lru "github.com/hashicorp/golang-lru/v2"
 
 	"github.com/hmchangw/chat/pkg/displayfmt"
+	"github.com/hmchangw/chat/pkg/idgen"
 	"github.com/hmchangw/chat/pkg/model"
 )
 
@@ -32,21 +33,19 @@ type HRIdentityStore interface {
 	AccountByTeamsID(ctx context.Context, teamsUserID string) (string, error)
 	// FindUserByAccount returns the single user with account (globally unique), or (nil,nil).
 	FindUserByAccount(ctx context.Context, account string) (*model.User, error)
-	// UpsertUserIdentities $sets IDENTITY FIELDS ONLY (siteId, engName, chineseName)
-	// keyed by account; it never touches roles/services/password.
-	UpsertUserIdentities(ctx context.Context, users []model.IUserWithChange) error
 }
 
 // senderResolver reuses the #70 HR store to map Teams users to nextgen identities.
 // The cache is process-wide (injected), shared across every batch the handler runs.
 type senderResolver struct {
-	store  HRIdentityStore
-	siteID string
-	cache  *lru.Cache[string, resolvedSender]
+	store        HRIdentityStore
+	siteID       string
+	cache        *lru.Cache[string, resolvedSender]
+	publishUsers func(ctx context.Context, users []model.IUserWithChange) error
 }
 
-func newSenderResolver(store HRIdentityStore, siteID string, cache *lru.Cache[string, resolvedSender]) *senderResolver {
-	return &senderResolver{store: store, siteID: siteID, cache: cache}
+func newSenderResolver(store HRIdentityStore, siteID string, cache *lru.Cache[string, resolvedSender], publishUsers func(ctx context.Context, users []model.IUserWithChange) error) *senderResolver {
+	return &senderResolver{store: store, siteID: siteID, cache: cache, publishUsers: publishUsers}
 }
 
 // resolve order: (1) map the AAD id to its account via teams_user — the globally-unique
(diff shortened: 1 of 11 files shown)


message-worker/teamssender.go
```diff
@@ -79,24 +78,18 @@ func (r *senderResolver) resolve(ctx context.Context, teamsUserID, displayName s
...
-		s = senderFromUser(created)
-	} else {
-		s = resolvedSender{Account: account, ChineseName: displayName, DisplayName: displayfmt.CombineWithFallback("", displayName, account)}
+	// New external user: the publisher owns _id == employeeId == the deterministic
+	// Graph-id hash, so this person is one identity whether resolved here or via the
+	// HR feed, and every site's hr-sync-worker converges on it. displayName lands in
+	// chineseName to mirror the HR mapping. Nothing is written locally first, so a
+	// redelivery re-derives the same _id instead of splitting the user.
+	id := teamsmigrate.EmployeeIDFromGraphID(teamsUserID)
+	nu := model.User{ID: id, Account: account, SiteID: r.siteID, ChineseName: displayName, EmployeeID: id}
```
mliu33: Make sure to fix and align the "_id" and "account" fields of the external user created in room creation for Teams migration.

- we don't lookup AccountByTeamsID
- no need "employeeId" field in user collection as external users are not actual employees

message-worker/teamssender.go
```diff
@@ -79,24 +78,18 @@ func (r *senderResolver) resolve(ctx context.Context, teamsUserID, displayName s
...
-	if created != nil {
-		s = senderFromUser(created)
-	} else {
-		s = resolvedSender{Account: account, ChineseName: displayName, DisplayName: displayfmt.CombineWithFallback("", displayName, account)}
+	// New external user: the publisher owns a deterministic _id (Graph-id hash) so
+	// this person is one identity at every site (hr-sync-worker keys on account).
+	// No employeeId — externals aren't employees. displayName lands in chineseName
+	// to mirror the HR mapping. Nothing is written locally first, so a redelivery
+	// re-derives the same _id instead of splitting the user.
+	id := idgen.DeterministicID([]byte(teamsUserID))
```
mliu33, line 86: Just remember to remove AccountByTeamsID() in line 63

room-worker/teamsroomcreate.go
```diff
@@ -180,11 +179,11 @@ func (h *Handler) resolveMember(ctx context.Context, account, graphID string) (*
...
-	// _id == employeeId == the deterministic Graph-id hash, so this person is one
-	// identity whether resolved here or via the HR feed, and hr-sync-worker keys
-	// its upsert on the (now non-empty) employeeId.
-	id := teamsmigrate.EmployeeIDFromGraphID(graphID)
-	nu := model.User{ID: id, Account: account, SiteID: h.siteID, EmployeeID: id}
+	// A deterministic _id from the Graph id means this person maps to one identity
+	// at every site (hr-sync-worker keys the upsert on account). No employeeId —
+	// migrated externals aren't employees.
+	id := idgen.DeterministicID([]byte(graphID))
+	nu := model.User{ID: id, Account: account, SiteID: h.siteID}
```
mliu33, line 186: Is Room-create job able to provide both user "account" and "displayName" when calling room creation rpc ? c.c @chenjr0719 @cocoamix00

mliu33 (approved):
I will approve and merge this first, but pls checkout the comments, thanks!

room-worker/teamsroomcreate.go
chenjr0719: Sounds good — plan for the follow-up:

- **Message path** (`teamssender.go`): today it maps the AAD id → account via a `teams_user` lookup (`AccountByTeamsID`) and resolves displayName. If the Teams batch payload carries `account` + `displayName` per sender (and per mention), we drop `AccountByTeamsID` entirely and build the identity straight from the payload.
- **Room-create path** already receives `account` per member, so no lookup there — it only needs `displayName` added if we want the external's name populated.

@cocoamix00 — could you confirm the Teams migration producer can include `account` + `displayName` in the `teams.batch` payload (for both the sender and any mentioned users)? Once that's in, I'll open the follow-up PR dropping the lookup. Tracked on our side.

chenjr0719: @cocoamix00 — and if it's easier, we're happy to handle the producer-side change ourselves (adding `account` + `displayName` to the `teams.batch` payload) — just say the word and we'll pick it up. 👍

cocoamix00: @chenjr0719 ok,  I will add dislayName to teams payload

chenjr0719: Thanks @cocoamix00! For the **room-create** payload that's exactly it — add `displayName` per member (`account` is already there), and I'll wire room-worker to read it and set the external's name.

One more for the **message batch** (`teams.batch`) — that's the mirror case: its payload already carries `displayName` but **not** `account`, which is why the message resolver still does the `AccountByTeamsID` lookup. If that payload can also include `account` (sender + mentions), I'll drop the lookup there too. Both tracked on #141.

chenjr0719: Correction on my last — scratch the message-batch part, that's a separate pipeline and not yours. Just the **room-create `displayName`** from you (which you've got 🙏). The message-batch `account` piece we'll route to the right owner separately.

PR #127, opened 2026-07-26
feat(loadgen): add Cassandra Run A soak harness

flywindy:
## Summary

Add a Kubernetes-ready Cassandra Run A soak harness to the existing `tools/loadgen` binary. The harness drives the real `message-gatekeeper` → `message-worker` → Cassandra write path and the `history-service` → Cassandra read path, while keeping the test topology and cleanup ownership isolated from borrowed staging users.

This is a component-level Cassandra soak tool. It does not define or certify the end-to-end newchat SLI/SLO contract.

## What changed

- Add `seed --workload=soak`, `soak`, and `teardown --workload=soak` lifecycle commands.
- Build channel and DM topology from up to 20,000 real MongoDB users without modifying or deleting those users.
- Seed test-owned rooms, subscriptions, ownership metadata, and per-room encryption material.
- Add the configurable Run A workload model:
  - top-level sends and thread replies;
  - `LoadHistory`, `GetThreadMessages`, and `GetMessageByID` reads;
  - reaction add/remove, edit, soft-delete, pin/unpin, and pinned-list actions;
  - Zipf room selection for hot-room behavior;
  - independent reaction rate and reactions-per-hot-message controls.
- Add a bounded per-room recent-message catalog with configurable `persist_grace` before messages become mutation or thread-parent candidates.
- Add transient retry/backoff, NATS reconnect behavior, pod restart warm-up, and mutation target-missing accounting.
- Add read-back verification through `GetMessageByID` and bucket-walking `LoadHistory` pagination.
- Extend loadgen metrics and reports with per-RPC achieved rate, latency, retry, error classification, verification, saturation, and `mutation_target_missing` signals.
- Add coverage enforcement for the Run A aggregate and core logic.
- Add English operator and implementation documentation under `docs/load-testing/`.

## Kubernetes and Argo CD lifecycle

The Helm chart models the company deployment constraints explicitly:

1. `phase=seed` creates a bounded, single-run Job.
2. `phase=soak` creates a single-replica Deployment with `Recreate` strategy and runs continuously until removed by a release.
3. `phase=stopped` removes workload resources while retaining evidence and test data.
4. `phase=teardown` creates an explicitly approved, bounded cleanup Job.

The chart mounts NATS credentials from a Secret, keeps workload configuration in a ConfigMap, disables service-account token mounting, runs as non-root with a read-only root filesystem, and exposes Prometheus metrics through a ClusterIP Service.

## Data ownership and safety

- Existing MongoDB users are borrowed read-only and are never deleted by teardown.
- MongoDB cleanup is restricted to entities marked as owned by the matching `SOAK_RUN_ID`.
- A heartbeat lease prevents teardown while a soak instance is still active.
- Cassandra cleanup defaults to `none` so results remain available for analysis.
- Cassandra truncation requires an explicit cleanup mode and an exact keyspace confirmation, and is intended only for a dedicated test keyspace.
- No credentials, `.env` files, generated NATS credentials, or staging endpoints are committed.

## Configurable assumptions

The currently unconfirmed production inputs remain configuration rather than hard-coded behavior:

- I8: reaction message scope;
- I10: global versus per-site rate scope;
- I12: messages per active user per day.

The supplied defaults are conservative and are reported as provisional at startup.

## Validation

- `make generate SERVICE=tools/loadgen`
- `make fmt`
- `make lint` — 0 issues
- `make test SERVICE=pkg/subject`
- `make test SERVICE=tools/loadgen`
- `make test-integration SERVICE=tools/loadgen`
- `make coverage-loadgen-soak`
  - Run A aggregate: **83.18%**
  - Run A core: **90.08%**
- `make validate-loadgen-k8s KUBE_DRY_RUN=true`
  - strict Helm lint for validation and local values;
  - render checks for seed, soak, stopped, and teardown phases;
  - successful Kubernetes client dry-run against local kind.
- `make test` — repository-wide race-enabled unit suite passed
- `make sast`
  - gosec: PASS
  - govulncheck: PASS, 0 reachable vulnerabilities
  - Semgrep: PASS, 0 findings
- Local kind smoke exercised the real gatekeeper/worker/history path, verified Cassandra ciphertext with no plaintext message body, validated pod restart warm-up, and confirmed ownership-scoped teardown without deleting borrowed users.

## Deferred and limitations

- Run B/C pathological experiments and direct-CQL injection are deferred.
- The optional service-side observability domain metric is not included.
- The recent-message catalog is intentionally not persisted across pod restarts; the Deployment warms it again from fresh sends and pinned-list reads.
- Local smoke validates correctness and deployability, not staging capacity or a production SLO verdict.
- Staging retention and Cassandra cleanup policy must be confirmed before the first real soak run.

## Summary by CodeRabbit

- **New Features**
  - Added Cassandra Run A soak testing (seed/run/teardown), soak workflow orchestration, and new Cassandra soak Prometheus metrics.
  - Added Helm chart + GitOps phase workflow (seed/soak/stopped/teardown) with metrics endpoint, values schema, and local validation.
  - Added NATS message subject builders for history/mutations plus wildcard message gets.
- **Documentation**
  - Added load-testing entry docs and Cassandra Run A implementation + soak test plans/runbooks.
- **Bug Fixes**
  - Improved local credential generation temp handling and artifact wiring.
  - Enhanced build automation with Windows-compatible output handling.
- **Tests**
  - Added extensive soak unit/integration coverage and edge-case validations.


tools/loadgen/soak_config.go
```diff
@@ -0,0 +1,217 @@
...
+		return fmt.Errorf("SOAK_MAX_USERS must be between 1 and %d", maxBorrowedSoakUsers)
+	}
+	if cfg.ActiveUsers <= 0 || cfg.ActiveUsers > cfg.MaxUsers {
+		return fmt.Errorf("SOAK_ACTIVE_USERS must be between 1 and SOAK_MAX_USERS")
+	}
+	if cfg.RoomCount <= 0 {
+		return fmt.Errorf("SOAK_ROOM_COUNT must be greater than zero")
+	}
+	if cfg.ChannelMembers < 2 || cfg.ChannelMembers > cfg.MaxUsers {
+		return fmt.Errorf("SOAK_CHANNEL_MEMBERS must be between 2 and SOAK_MAX_USERS")
```
flywindy, line 152: **Validate `SOAK_CHANNEL_MEMBERS` against the gatekeeper large-room threshold.**

The bound here is `2..SOAK_MAX_USERS` (up to 20,000), but `message-gatekeeper` enforces `LARGE_ROOM_THRESHOLD` (`message-gatekeeper/main.go:35`, `envDefault:"500"`): in a room whose `UserCount` exceeds the threshold, only owners/admins/bots may send **top-level** messages — everyone else gets `errLargeRoomPostRestricted` (`message-gatekeeper/handler.go:302-316`). Thread replies are exempt.

Failure scenario: an operator sets `channelMembers: 600` to model a big channel. Every channel room is now over the threshold, so top-level sends from ordinary members are rejected en masse. Because each seeded channel has a single owner, essentially the whole send lane collapses into a rejection storm, and the errors surface as generic send failures — expensive to diagnose after hours of soak.

Suggest failing closed at config time, e.g. cap `ChannelMembers` at the threshold (ideally a new `SOAK_LARGE_ROOM_THRESHOLD` defaulting to 500 so it can track the deployed gatekeeper setting) and reject anything above it with a message that names the gatekeeper behavior. One validation line prevents a whole class of wasted runs.

---

tools/loadgen/soak_main.go
```diff
@@ -0,0 +1,703 @@
...
+			continue
+		}
+		selector.members[subscription.RoomID] = append(
+			selector.members[subscription.RoomID],
+			soakSendTarget{
+				UserID: subscription.User.ID, Account: subscription.User.Account,
+				RoomID: subscription.RoomID,
+			},
+		)
+	}
```
flywindy, line 229: **`SOAK_ACTIVE_USERS` does not constrain runtime sender identities.**

Topology construction honors the active subset correctly — active users are placed first and idle borrowed users backfill the remaining seats (`soak_topology.go:129-147`). But the runtime selector builds its sender pool from *every* subscribed member of every room, so with `MAX_USERS=20000` / `ACTIVE_USERS=2000` the send lane issues messages as ~20k distinct identities, not 2k.

Two consequences:
- Divergence from the documented workload model, which specifies ~2k–5k concurrently active users driving the load.
- The startup log line `"activeUsers", len(topology.ActiveUsers)` is misleading on the soak path, because `LoadTopology` reconstructs `ActiveUsers` from all subscribers (`soak_store.go:265-278`), not from the seeded active subset.

Cassandra-shape impact is small (partitions key on room, not sender), so this is not a correctness bug — but it is a fidelity/reporting gap. Preferred fix: persist the active subset in the manifest (or tag subscriptions) and have the selector draw senders from it. Minimum fix: rename the runtime concept and adjust the log/docs so `ACTIVE_USERS` is clearly seed-time-only.

---

tools/loadgen/soak_mutation.go
```diff
@@ -0,0 +1,676 @@
...
+	request soakRPCRequest,
+	response any,
+) (soakRPCResult, int, time.Duration, bool, error) {
+	startedAt := m.clock.Now()
+	var result soakRPCResult
+	for attempt := 0; attempt <= m.cfg.MutationRetries; attempt++ {
+		var err error
+		result, err = m.rpc.Call(ctx, request, response)
+		if err == nil {
+			return result, attempt, m.clock.Now().Sub(startedAt), false, nil
```
flywindy: Minor: reported mutation latency includes retry backoff sleeps.

`startedAt` is captured once and the elapsed time is measured after the whole retry loop, so a not-found target that burns `MUTATION_RETRIES` sleeps (100ms → 200ms → 400ms by default) inflates the recorded latency by ~700ms even though each individual RPC was fast. During a warm-up window — or any period where persist lag exceeds `PERSIST_GRACE` — this inflates mutation p95/p99 and could be misread as Cassandra slowdown.

Consider recording per-attempt latency (or subtracting slept time) so the latency signal stays attributable to the service, with the retry count already carried separately in `outcome.Retries`.

---

tools/loadgen/soak_workload.go
```diff
@@ -0,0 +1,514 @@
...
+		maxInFlight,
+		recordUnderrun,
+		recordSaturation,
+		action,
+	)
+}
+
+func soakIntegerRate(rate float64) int {
+	return max(1, int(math.Round(rate)))
+}
```
flywindy: Minor: sub-1/s rates are silently rounded up to 1/s.

`max(1, round(rate))` means any configured rate in `(0, 1)` becomes 1/s — e.g. `SOAK_VERIFY_RATE=0.5` or `SOAK_PINNED_LIST_RATE=0.2` runs at 2× and 5× the requested rate respectively. Config validation only rejects `<= 0`, so this passes silently.

Given `pacedDispatch` takes an integer rate, either document the 1/s floor on the affected env vars, or reject fractional rates in `validateSoakConfig` so the discrepancy is visible at startup rather than in the report.

---

tools/loadgen/soak_send.go
```diff
@@ -0,0 +1,399 @@
...
+		parent, ok := s.catalog.PickEligible(
+			target.RoomID,
+			target.Account,
+			soakCatalogThreadParent,
+		)
+		if !ok || !s.catalog.IncrementThreadReplies(target.RoomID, parent.ID) {
+			kind = soakSendTopLevel
+		} else {
+			threadParentID = parent.ID
+		}
```
flywindy, line 158: Minor: the thread-reply budget is incremented optimistically at publish time and never rolled back.

`IncrementThreadReplies` runs before publishing, but a subsequent gatekeeper rejection, malformed reply, or `Expire()` timeout only calls `catalog.Reject(...)` — the parent's `threadReplies` counter keeps the increment. Over a long run the effective per-thread reply budget drifts below the configured `ThreadReplyLimit`.

The bias is conservative (threads stay under the cap rather than over it), so no behavior change is strictly needed — but a short comment stating the intent would stop a future reader from treating it as a bug, or decrement it alongside the existing `Reject` calls if exactness matters for the wide-partition follow-on experiments.

---

tools/loadgen/soak_store.go
```diff
@@ -0,0 +1,543 @@
...
+
+	var users []model.User
+	if err := cursor.All(ctx, &users); err != nil {
+		return nil, fmt.Errorf("decode borrowed soak users: %w", err)
+	}
+	return users, nil
+}
+
+func (s *mongoSoakStore) ResetOwned(ctx context.Context, runID string) error {
+	filter := bson.D{{Key: "soakRunId", Value: runID}}
```
flywindy: Minor: `soakRunId` queries have no supporting index on the shared collections.

`ResetOwned`, `LoadTopology`, and the teardown paths all filter `rooms` / `subscriptions` on `soakRunId`, which is a field this harness adds and nothing indexes. On a staging database where those collections already hold real data, each of these is a collection scan.

These are administrative, low-frequency operations, so this is acceptable — but with 10k seeded rooms and their subscriptions on a busy staging `subscriptions` collection, teardown may be slow enough to matter during the evidence-retention window. Worth either creating a partial index at seed time (and dropping it at teardown) or noting the expected cost in the operator README so nobody assumes teardown hung.

---

flywindy (commented):
## Review summary

Reviewed the full runtime (catalog, send, read, mutation, RPC, workload, store, teardown, config, topology, wiring), the `pkg/subject` change, the Helm chart, and the Makefile targets, cross-checking the load-bearing claims against `message-gatekeeper` and `history-service`. CI is green across lint / test / test-integration / sast.

**Overall: strong work, suitable to merge once the first finding is addressed.** The hard parts are handled correctly, and several of them are things that are easy to get wrong:

- **Reaction toggle ambiguity.** `ReactMessage` is a toggle, so a blind retry after an ambiguous timeout would invert state. The `soakRetryAmbiguous` + `ResolveAmbiguity` read-back path handles this properly, and reaction rate is modeled independently rather than as `send_rate × reactions_per_message`.
- **Read amplification.** The measured read lane is pinned to `MaxPages: 1` with the rationale documented in `soakMeasuredReadConfig`, leaving bucket-walks to the 1/s verifier. This avoids the obvious trap of every scheduled read walking 100 pages once rooms accumulate history.
- **Teardown safety.** Triple keyspace confirmation, heartbeat lease against live runs, no row-by-row Cassandra deletes, ownership-scoped Mongo deletion, idempotent repeat teardown, and borrowed users strictly read-only.
- **Catalog concurrency.** Consistent `globalMu → shard.mu` lock ordering with no inverse path, bounded pending map so lost gatekeeper replies cannot leak, and pinned entries protected from eviction.
- **Send correlation.** Reply subjects are verified against the expected `UserResponse(account, requestID)` before matching, the accepted `model.Message` is asserted field-by-field, and `AcceptAt` adopts the server's `CreatedAt` — which is what keeps bucket math honest.
- Subject builders line up exactly with the registered `history-service` patterns, payload sizing accounts for encryption overhead and respects the 20 KiB gatekeeper limit, and the coverage contract is enforced by tooling rather than asserted in prose.

### Findings

Inline comments cover, in priority order:

1. **`SOAK_CHANNEL_MEMBERS` is not validated against the gatekeeper's large-room threshold** — the one item worth fixing before merge. A value above 500 turns the send lane into a rejection storm that is expensive to diagnose mid-soak.
2. **`SOAK_ACTIVE_USERS` does not constrain runtime sender identities** — seed-time honors it, runtime does not; also makes the startup log misleading. Fidelity/reporting gap rather than a correctness bug.
3. Mutation latency includes retry backoff sleeps (inflates p95/p99 during target-missing churn).
4. Sub-1/s rates are silently rounded up to 1/s.
5. Thread-reply budget increments optimistically and is never rolled back (conservative bias; a comment may be enough).
6. `soakRunId` filters have no supporting index on shared collections (teardown cost on a populated staging database).

### Notes, no action requested

- The pivot from Job to Deployment for the soak phase is well-supported by `runMode=continuous`, the `stopped` phase, and the heartbeat lease — and the implementation plan document was updated to match, so there is no doc drift.
- Heartbeat failure is fatal regardless of `StopOnActionError`. That reads as deliberate and correct: once the lease lapses, the teardown guard is no longer protecting anything.
- The kind smoke test (real gatekeeper/worker/history path, ciphertext verification) is a manual claim that CI cannot reproduce. That is reasonable given the encryption path can only be exercised end-to-end, but it would help to link the reproduction steps from the operator README so the check is repeatable before the first staging run.

---

tools/loadgen/soak_config.go
flywindy: Addressed in efd90e6. Added configurable SOAK_LARGE_ROOM_THRESHOLD with default 500, validated channel membership against it, and wired the value through Helm, documentation, and tests.

tools/loadgen/soak_main.go
flywindy: Addressed in efd90e6. Seed now persists the exact active user IDs, runtime selectors use only subscriptions owned by that active subset, and topology tests ensure every room has an active member and every active user has a writable room.

tools/loadgen/soak_mutation.go
flywindy: Addressed in efd90e6. Mutation latency now sums actual RPC call time and excludes target retry backoff. A clock-advancing test covers the distinction.

tools/loadgen/soak_workload.go
flywindy: Addressed in efd90e6. Soak lanes now preserve float rates and the pacer supports sub-hertz cadence instead of rounding values below 1 per second up to 1.

docker-local/setup.sh
```diff
@@ -86,19 +91,19 @@ docker run --rm \
     nsc generate creds --account chatapp --name backend > /output/backend.creds
   '
 
-cp "$TMPDIR/backend.creds" "$BACKEND_CREDS"
+cp "$SETUP_TMP_DIR/backend.creds" "$BACKEND_CREDS"
 # 0644, not 0600: service containers run as non-root (uid 10001) and
 # bind-mount this file read-only at /etc/nats/backend.creds, so the
 # in-container user must be able to read it. Acceptable only because
 # this is a throwaway local-dev credential generated by this script.
 chmod 644 "$BACKEND_CREDS"
```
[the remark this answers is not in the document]
flywindy: No change in this PR. This is an explicitly documented throwaway local-development credential, and 0644 is required so the non-root UID 10001 service containers can read the read-only bind mount. Production uses an operator-owned Kubernetes Secret. Moving to 0600 would require a separate volume or init ownership redesign outside the Run A scope.

tools/loadgen/soak_main.go
```diff
@@ -0,0 +1,703 @@
...
+	if topology == nil || len(topology.Rooms) == 0 {
+		return nil, fmt.Errorf("soak topology has no rooms")
+	}
+	if cfg == nil {
+		return nil, fmt.Errorf("soak configuration is required")
+	}
+	picker, err := newSoakRoomPicker(seed, len(topology.Rooms))
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
flywindy: Addressed in efd90e6. The room picker construction error is now wrapped with contextual information before it reaches the caller.

tools/loadgen/soak_rpc.go
```diff
@@ -0,0 +1,346 @@
...
+		}
+
+		delay := c.backoff(result.Retries)
+		if err := c.sleeper.Sleep(ctx, delay); err != nil {
+			return result, fmt.Errorf("wait to retry %s: %w", request.Action, err)
+		}
+		result.Retries++
+	}
+
+	return result, errSoakRetryExhausted
```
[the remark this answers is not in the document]
flywindy: Addressed in efd90e6. Retry exhaustion now preserves both errSoakRetryExhausted and the underlying request error for errors.Is and errors.As classification, and the impossible fallthrough returns an explicit invariant error.

tools/loadgen/soak_seed.go
```diff
@@ -0,0 +1,162 @@
...
+		SubscriptionCount: len(topology.Subscriptions),
+		StartedAt:         now,
+		UpdatedAt:         now,
+	}
+	if err := store.PutManifest(ctx, &manifest); err != nil {
+		return soakTopology{}, fmt.Errorf("record seeding manifest: %w", err)
+	}
+	if err := store.ResetOwned(ctx, input.RunID); err != nil {
+		return soakTopology{}, fmt.Errorf("reset partial topology for run %q: %w", input.RunID, err)
+	}
```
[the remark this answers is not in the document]
flywindy: Addressed in efd90e6. Seed performs a manifest preflight and refuses to reset a running soak with a fresh heartbeat. Tests verify rejection happens before any topology write.

tools/loadgen/soak_send.go
```diff
@@ -0,0 +1,399 @@
...
+	}
```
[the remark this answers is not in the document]
flywindy: Addressed in efd90e6. Thread reply reservations are decremented on terminal marshal, tracking, rejection, malformed reply, assertion, acceptance, and expiry failures. Tests verify a rejected reply returns capacity to the parent.

tools/loadgen/soak_send.go
```diff
@@ -0,0 +1,399 @@
...
+		return result
+	}
+	if !s.catalog.AcceptAt(
+		pending.Target.RoomID,
+		pending.MessageID,
+		response.CreatedAt,
+	) {
+		result.ErrorClass = soakErrorAssertion
+		return result
+	}
```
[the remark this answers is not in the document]
flywindy: Addressed in efd90e6. The AcceptAt failure path now rejects the pending catalog candidate immediately instead of leaving it until global eviction.

tools/loadgen/soak_store.go
```diff
@@ -0,0 +1,543 @@
...
+		FindOne(ctx, bson.D{{Key: "_id", Value: runID}}).
+		Decode(&manifest)
+	if err == nil {
+		return &manifest, nil
+	}
+	if errors.Is(err, mongo.ErrNoDocuments) {
+		return nil, nil
+	}
+	return nil, fmt.Errorf("find soak manifest %q: %w", runID, err)
+}
```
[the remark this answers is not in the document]
flywindy: Addressed in efd90e6. FindManifest now reuses the explicit soak manifest projection.

tools/loadgen/soak_store.go
```diff
@@ -0,0 +1,543 @@
...
+	}
+	roomOwnedFilter := bson.D{
+		{Key: "soakRunId", Value: runID},
+		{Key: "_id", Value: bson.D{{Key: "$in", Value: roomIDs}}},
+	}
+	if _, err := s.db.Collection("rooms").DeleteMany(ctx, roomOwnedFilter); err != nil {
+		return fmt.Errorf("delete rooms for soak run %q: %w", runID, err)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
flywindy: Addressed in efd90e6. Seed checks generated room IDs against existing rooms through the primary key before writing. Teardown verifies rooms by _id plus soakRunId before deleting dependent thread data or wrapped DEKs. Integration tests cover conflict rejection and preservation of non-owned artifacts.

tools/loadgen/soak_send.go
flywindy: Addressed in efd90e6. Exact thread reply accounting was chosen: terminal failure and expiry paths compensate the optimistic reservation, while a publish error keeps the pending candidate available for its stable retry path.

tools/loadgen/soak_store.go
flywindy: Addressed in efd90e6 without adding indexes to shared collections. Reset, reload, and teardown start from ownership room IDs; the ledger uses its run-prefixed _id range, rooms use _id, subscriptions and thread rooms use existing roomId indexes, and thread subscriptions use the existing threadRoomId index. Teardown also has configurable serial batch pacing and per-batch timeouts.

tools/loadgen/soak_topology.go
```diff
@@ -0,0 +1,358 @@
...
+	}
+}
+
+func eligibleSoakUsers(users []model.User, siteID string) []model.User {
+	eligible := make([]model.User, 0, len(users))
+	for i := range users {
+		user := &users[i]
+		if user.ID == "" ||
+			user.SiteID != siteID ||
+			user.Deactivated ||
```
flywindy: **Blocking: this no longer compiles against `main`.**

`main` landed #129 (`refactor(user): replace deactivated flag with active *bool`), which removed `model.User.Deactivated` in favor of `Active *bool` plus an `IsActive()` helper where a missing field means active. The branch still compiles in isolation, which is why local runs pass, but CI builds the merge commit and fails:

```
tools/loadgen/soak_topology.go:42:9: user.Deactivated undefined
tools/loadgen/soak_topology_test.go:17:54: unknown field Deactivated in struct literal
```

Reproduced locally with `git merge origin/main && go build ./tools/loadgen/...`. This is what turned all four CI jobs red — lint/test/test-integration/sast each fail at the typecheck stage, so nothing downstream actually ran.

Four sites need updating, and note the Mongo-side ones are semantically different from a rename:

- `soak_topology.go:42` — use `!user.IsActive()` instead of `user.Deactivated`.
- `soak_topology_test.go:17` — `Deactivated: true` becomes `Active: ptr(false)` (or whatever pointer helper the package uses).
- `soak_store.go:50` — the filter `{"deactivated": {$ne: true}}` must become `{"active": {$ne: false}}`. The polarity flips: absent-means-active is now expressed on the `active` field. `user-service/mongorepo/users.go:53` has the canonical form (`activeUserFilter`) worth mirroring.
- `soak_store.go:70` — projection key `deactivated` → `active`, otherwise `IsActive()` reads a field that was never fetched and every borrowed user silently looks active.

The last two matter beyond compilation: if the projection and filter disagree with the new schema, the seed would borrow deactivated users, and those accounts may fail gatekeeper checks at send time.

---

tools/loadgen/soak_store.go
```diff
@@ -0,0 +1,764 @@
...
+var (
+	_ soakSeedStore      = (*mongoSoakStore)(nil)
+	_ soakTeardownStore  = (*mongoSoakStore)(nil)
+	_ soakLifecycleStore = (*mongoSoakStore)(nil)
+)
+
+func soakUserFilter(siteID string) bson.D {
+	return bson.D{
+		{Key: "siteId", Value: siteID},
+		{Key: "deactivated", Value: bson.D{{Key: "$ne", Value: true}}},
```
flywindy: Same #129 fallout as `soak_topology.go:42`, but this one is a silent behavioral bug rather than a compile error — worth calling out separately because the compiler will not catch it.

The `deactivated` field no longer exists in the users schema, so `{"deactivated": {$ne: true}}` matches every document unconditionally. Combined with the stale `deactivated` projection on line 70, deactivated accounts get borrowed into the topology and `IsActive()` sees a zero-valued field.

Mirror the canonical filter from `user-service/mongorepo/users.go:53`:

```go
{Key: "active", Value: bson.D{{Key: "$ne", Value: false}}},
```

and project `active` instead of `deactivated`. `soak_seed_test.go:24,31` assert on the old key names and will need the same update.

---

flywindy (commented):
## Follow-up review of `efd90e6`

All six findings from the previous round are addressed, several of them more thoroughly than I suggested. One new blocker appeared, but it is unrelated to the fixes themselves.

### Blocking: branch no longer compiles against `main`

`main` merged #129 (`Deactivated` → `Active *bool`) after this branch was cut. The branch builds in isolation, so local runs stay green, but CI builds the merge commit and fails typecheck — which is why lint, test, test-integration, and sast all went red at once. Reproduced locally with `git merge origin/main && go build ./tools/loadgen/...`. Details and the four call sites are in the inline comments; note that the Mongo filter and projection changes are semantic, not a rename.

### Previous findings — all resolved

| # | Finding | Resolution |
|---|---|---|
| M1 | `SOAK_CHANNEL_MEMBERS` vs gatekeeper large-room threshold | `SOAK_LARGE_ROOM_THRESHOLD` (default 500) added, with validation that channel membership cannot exceed it |
| M2 | `SOAK_ACTIVE_USERS` inert at runtime | Manifest now persists `activeUserIds`; `LoadTopology` rebuilds the active set from it with a count-mismatch guard, and selector/reader/mutator all filter through `isActiveSoakSubscription`. This is the stronger of the two options I offered |
| m3 | Mutation latency included backoff sleeps | Accumulates per-attempt `rpcLatency`; sleeps excluded |
| m4 | Sub-1/s rates rounded up | Pacer is float-native end to end (`newRatePacer`, `pacedDispatchRate`); `soakIntegerRate` deleted |
| m5 | Thread-reply budget never rolled back | `DecrementThreadReplies` plus a `rejectPending` helper releases the reservation on every reject path |
| m6 | `soakRunId` collection scans | Scans eliminated rather than indexed: every query is anchored on `_id $in` / `roomId $in` from paged ownership chunks, with `soakRunId` demoted to a safety predicate. Teardown also gained batch size, inter-batch delay, and per-batch timeout |

### Improvements beyond the review

Several of these are things I did not flag and are good catches:

- **Unbounded catalog under heavy pinning.** The old eviction loops `break` when every candidate is pinned, so `globalCap` could be exceeded indefinitely. `removeOldestGlobalLocked` now force-evicts. Real latent memory bug on a 72-hour run.
- **Teardown ownership re-verification.** `DeleteOwnedRoomBatch` now re-reads rooms and deletes only IDs still tagged with the run, so a recycled room ID cannot be collaterally deleted.
- **Seed-time safety.** `FindConflictingRoomIDs` refuses to seed when a generated room ID collides with non-owned data, and `isActiveSoakManifest` refuses to reseed over a live run.
- **Cursor-advance guards** on both ownership paging loops prevent an infinite loop on a misbehaving cursor.
- **`completeSoakRun`** no longer uses `context.Background()`; it gets a 5s timeout.
- Metrics server gained read/write/idle timeouts.

### Verification

`make test SERVICE=tools/loadgen` passes on the branch head in isolation (262s). I did not re-run the full suite post-merge since it does not compile yet.

Once the #129 rebase is done, this is ready from my side — no other outstanding concerns.

---

tools/loadgen/soak_seed_integration_test.go
```diff
@@ -0,0 +1,238 @@
...
+	require.NotEmpty(t, conflictingRoomID)
+	_, err = db.Collection("rooms").InsertOne(ctx, bson.D{
+		{Key: "_id", Value: conflictingRoomID},
+		{Key: "siteId", Value: "site-a"},
+	})
+	require.NoError(t, err)
+
+	store := &mongoSoakStore{db: db}
+	keyStore := roomkeystore.NewMongoStore(db.Collection("rooms"), time.Hour)
+	t.Cleanup(func() { _ = keyStore.Close() })
```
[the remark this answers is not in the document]
flywindy: Agreed. This is a valid cleanup-observability issue. I will make the t.Cleanup callback report keyStore.Close() failures through the test handle and keep this thread open until that change is pushed.

tools/loadgen/soak_topology.go
flywindy: Addressed in `ae534f9` after merging current `main` in `02db214`. Topology eligibility now uses `!user.IsActive()`, and the test models an inactive user with `Active: &false`. The full GitHub Actions run is green: lint, repo-wide test, SAST, and tools integration all passed.

tools/loadgen/soak_store.go
flywindy: Addressed in `ae534f9`. The Mongo filter now uses `active: {$ne: false}`, the projection includes `active`, and the exact filter/projection test was updated to preserve the missing-means-active semantics. The resulting CI run is fully green.

flywindy:
Review follow-up for current head `ae534f9`:

- The #129 `Deactivated` -> `Active *bool` blocker is fixed. The branch now uses `IsActive()`, the canonical Mongo `active: {$ne: false}` filter, and an `active` projection. All GitHub Actions checks are green.
- Accepted: report `keyStore.Close()` failures in integration cleanup.
- Accepted: validate DM-room capacity using the number of unique pairs involving at least one active user, so impossible configurations fail early with a useful error.
- Accepted with a smaller implementation: remove the room-conflict test dependency on error text and assert the failure plus zero side effects; a new production sentinel solely for this test is unnecessary.
- Accepted: add coverage for a running manifest with a missing heartbeat, which must fail closed as active.
- No change for the zero-rate pacer suggestion: `soakWorkload.Run` skips every lane whose `rate <= 0` before invoking `dispatchSoakLane`, so `newRatePacer(0, ...)` is not reachable from the configured soak path and no work is released.
- No change for the `SOAK_RUN_ID` charset suggestion: ownership bounds are built from the complete run ID plus `:`/`;`, so characters inside the run ID do not shift the prefix range. Adding an allowlist would introduce an unrelated compatibility restriction without fixing a demonstrated targeting issue.

I am leaving the accepted-but-not-yet-pushed items open until their implementation and validation are on the branch.

tools/loadgen/soak_topology.go
```diff
@@ -0,0 +1,358 @@
...
+			User: model.SubscriptionUser{
+				ID:      members[i].ID,
+				Account: members[i].Account,
+			},
+			RoomID:       room.ID,
+			SiteID:       room.SiteID,
+			Roles:        roles,
+			Name:         name,
+			RoomType:     room.Type,
+			IsSubscribed: true,
```
flywindy, line 341: Non-blocking fidelity note from the `main` merge, not from your fix.

#130 added `Subscription.Open` — the per-user sidebar-visibility flag — declared as `Open bool` with `bson:"open"` and **no** `omitempty`. Since `buildSoakSubscriptions` does not set it, every seeded subscription is persisted with an explicit `open: false`, whereas production subscriptions are born open (room-worker's `newSub` sets it true).

No functional impact on this soak: gatekeeper's `GetSubscription` and history-service's subscription lookup both filter only on `u.account` + `roomId`, so sends, history reads, and mutations are unaffected. But `user-service`'s `subscription.list` matches `open: {$ne: false}`, so seeded rooms would be invisible in a real client's sidebar on staging — mildly confusing if someone inspects the environment mid-run, and a divergence from production shape if a future workload ever exercises `subscription.list`.

Suggest setting `Open: true` here alongside `IsSubscribed: true` so seeded rows match what room-worker would create.

---

flywindy (commented):
## Follow-up review of `ae534f9` — LGTM

The `main` merge and the #129 alignment both look right. CI is green across all five checks, and I re-verified locally.

### Blocker resolved

All four call sites were updated, and the two semantic ones landed correctly rather than as a mechanical rename:

- `soak_topology.go:42` — `!user.IsActive()`, which correctly preserves the missing-means-active rule.
- `soak_store.go:50` — `{"active": {$ne: false}}`, matching `user-service/mongorepo/users.go:53` (`activeUserFilter`) exactly. The polarity flip is what matters here and it is right.
- `soak_store.go:70` — projection switched to `active`, so `IsActive()` reads a field that is actually fetched.
- `soak_topology_test.go` — `Active: &inactive`.

`soak_seed_test.go` also got stronger rather than merely updated: it now asserts the exact filter value (`bson.D{{"$ne", false}}`) instead of just key presence, so a future polarity regression fails the test.

### Verification

- `git merge-base --is-ancestor origin/main HEAD` — `main` is fully merged, so CI is testing the real post-merge state.
- `go build ./tools/loadgen/... ./pkg/...` and `go vet ./tools/loadgen/` — clean.
- `make test SERVICE=tools/loadgen` — pass (300s).
- `make test SERVICE=pkg/subject` — pass.
- No residual `Deactivated` / `"deactivated"` references outside the two `model_test.go` assertions that deliberately check the field is gone.
- CI: lint, sast, test, test-integration, prewarm all green.

### One non-blocking note

The merge brought #130's `Subscription.Open` flag, and seeded subscriptions now persist `open: false` because the field has no `omitempty`. No effect on the soak's read/write paths — neither gatekeeper nor history-service filters on it — but it diverges from production shape and hides seeded rooms from `subscription.list`. Inline comment has the detail; a one-line `Open: true` would close it.

### Verdict

LGTM from my side (GitHub will not let me formally Approve my own PR). All findings from both prior rounds are resolved, CI is green, and the remaining item is cosmetic — fold it into this branch or a follow-up, your call.

---

tools/loadgen/soak_seed_integration_test.go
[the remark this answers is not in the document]
flywindy: Addressed in `a3ec39e`. All three loadgen integration cleanups now assert `keyStore.Close()` with `require.NoError`, so cleanup failures surface in test results. The updated loadgen integration suite passed with the race detector, along with repo-wide tests, lint, coverage, Helm validation, and SAST.

flywindy:
Final review follow-up in `a3ec39e`:

- Fixed DM capacity validation to count only active-active and active-borrowed pairs; an impossible 1-active/4-borrowed case now fails before generating any subscription IDs.
- Seeded subscriptions now set `Open: true`, matching production birth semantics. Both topology unit coverage and Mongo integration coverage verify client-visible subscriptions are persisted open.
- All loadgen integration cleanup callbacks now surface `keyStore.Close()` errors.
- Conflict tests no longer depend on error message text; they assert failure and zero side effects.
- Added the missing-heartbeat fail-closed test for a running manifest.
- The previously discussed zero-rate pacer and `SOAK_RUN_ID` charset suggestions remain unchanged for the reasons documented above.

Validation: `make lint`, `make test`, `make test-integration SERVICE=tools/loadgen`, `make coverage-loadgen-soak` (82.68% aggregate / 90.13% core), `make validate-loadgen-k8s`, and complete SAST all passed.

PR #134, opened 2026-07-27
feat: per-user customizable chatlist sections

chenjr0719:
## Summary

Per-user customizable **chatlist sections**: users group their chats into named
sections with manual ordering. A chat's section **membership** and **order** live
on its subscription (`sectionId` / `sectionOrder`) — exactly how `favorite` is
stored — while a small per-user overlay carries the section **definitions** only.
So a membership change is **one** subscription write + **one** cross-site event,
and the definition fan-out stays **O(sections), not O(chats)**.

## What's included

**Subscription-backed membership + order (room-service)**
- `Subscription.SectionId *string` / `SectionOrder float64` / `SectionUpdatedAt *time.Time`.
- `moveChat` RPC: sets `sectionId` + a **fractional** `sectionOrder`. Placement is
  `afterRoomId` (after that room), `beforeRoomId` (before it — top-insertion when
  it's the section head), or append (omit both); the two anchors are mutually
  exclusive. A bounded rebalance re-spaces the section only when float precision is
  exhausted. One call covers assign / reorder-within / move-between / remove
  (`sectionId=null`). Fans out `subscription.update` (`action:"section_moved"`) + a
  cross-site INBOX event, high-water-mark-guarded by `sectionUpdatedAt` — mirrors
  `favorite.toggle`.
- `sectionOrder` is a meaningful `float64` (a top-inserted chat is legitimately `0`),
  so it carries no `omitempty` on the wire — the client always receives the new position.

**Section-definition registry (user-service)**
- `ChatlistState{SectionOrder, Sections, LastUpdatedAt}` + `ChatlistSection{ID, Name,
  BuiltIn, SortMode}` on `User.Chatlist`, with a `chatlistUpdatedAt` HWM.
- RPCs: `chatlist.get`, `section.create` / `rename` / `delete`, `section.reorder`
  (section display order — a full permutation of every section id), `section.setSortMode`
  (`custom` | `mostRecent`).
- Fans out `chatlist.update` + a cross-site INBOX event (mirrors settings); the
  inbox-worker applies `user_chatlist_updated`.

**Built-in "teams" section**: a product constant; the Teams-migration's existing
subscription write also stamps `sub.sectionId="teams"` — no separate import or
reconcile. An unknown `sectionId` renders under the built-in Chats section
(orphan-tolerant).

**Docs** (`client-api.md`, `client-api/events.md`, `client-api/request-reply.md`):
moveChat, the six registry RPCs, the subscription's new fields, and the client
read-model grouping rules.

## NATS subjects

Request-reply:
- `chat.user.{account}.request.room.{roomID}.{siteID}.chat.move`
- `chat.user.{account}.request.user.{siteID}.chatlist.get`
- `chat.user.{account}.request.user.{siteID}.chatlist.section.{create,rename,delete,reorder,setsortmode}`

Events:
- `chat.user.{account}.event.chatlist.update` (section definitions; O(sections))
- `chat.user.{account}.event.subscription.update` — `action:"section_moved"` (per-chat membership/order)

Cross-site INBOX: `subscription_section_moved` (HWM `sectionUpdatedAt`) ·
`user_chatlist_updated` (HWM `chatlistUpdatedAt`).

## Verification

- Unit + integration tests across room-service, user-service, the outbox, and the inbox-worker.
- End-to-end on a live dev stack: every RPC (create / rename / delete / reorder /
  setSortMode / moveChat with after / before / append) plus the `section_moved` and
  `chatlist.update` fan-outs, with the moved subscription's `sectionId` / `sectionOrder`
  confirmed at rest.
- **Cross-site federation** confirmed both ways — a section move on one site applies to
  the subscription on the other under the HWM guard, and section definitions federate
  the same way.
- A full front-end walkthrough of every case (create, rename, reorder, delete →
  orphan-fallback, move in / out, move-to-top, built-in Teams landing, and a live
  cross-tab update with no reload).

## Summary by CodeRabbit

* **New Features**
  * Added customizable chatlist sections with creation, renaming, deletion, reordering, and sort-mode controls.
  * Added chat movement between sections, custom ordering, and removal from sections.
  * Added synchronization of chatlist and section changes across connected devices and sites.
  * Teams chats are automatically placed in the Teams section.

* **Documentation**
  * Documented chatlist APIs, events, validation rules, and fractional ordering behavior.

### DIFF
--- a/room-service/handler.go
+++ b/room-service/handler.go
@@ -2250,6 +2250,7 @@ func (h *Handler) moveChat(c *natsrouter.Context, req model.MoveChatRequest) (*m
 	now := time.Now().UTC()
 
 	var order float64
+	var rebalanced []model.Subscription
 	if req.SectionID != nil {
 		var needRebalance bool
 		var err error
@@ -2258,7 +2259,7 @@ func (h *Handler) moveChat(c *natsrouter.Context, req model.MoveChatRequest) (*m
 			return nil, fmt.Errorf("compute section order: %w", err)
 		}
 		if needRebalance {
-			if err := h.store.RebalanceSection(ctx, account, *req.SectionID); err != nil {
+			if rebalanced, err = h.store.RebalanceSection(ctx, account, *req.SectionID, now); err != nil {
 				return nil, fmt.Errorf("rebalance section: %w", err)
 			}
 			if order, _, err = h.store.ComputeSectionOrder(ctx, account, *req.SectionID, req.AfterRoomID, req.BeforeRoomID); err != nil {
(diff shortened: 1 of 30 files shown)


chenjr0719:
Thanks @coderabbitai — addressed the round in `7a4ec224`:

**Applied:**
- **Cross-site HWM (the real one):** the origin `UpdateUserChatlist` now stamps the top-level `chatlistUpdatedAt` (= the timestamp shared with both fanouts), so a local edit bumps the high-water mark and an older inbound cross-site event can't regress it.
- **Built-in overlay id bound:** each item id is now length-bounded (`MaxChatIDLen`), so a client can't persist arbitrary long strings into its replicated doc.
- Docs: `chatlist.get` documented as read-only (emits no event); the model comments corrected (a built-in section's `Items` is the optional pin overlay, not always empty).
- Tests: store-error path, equal-timestamp duplicate delivery (the strict `$lt` guard), bad-id reject, and a chatlist-only projection on the integration readbacks.

**Two I didn't apply, with rationale:**
- *Request/reply doc link* — the href `client-api/request-reply.md#chatlist-sections` is already a valid **relative** path to the sibling doc (it resolves to the correct anchor), not a self-page link — so no change needed.
- *bson tags on the event struct* — `ChatlistUpdateEvent` is a wire-only fanout payload, never persisted; it intentionally follows the same json-only shape as the sibling `SettingsUpdateEvent`. The persisted structs (`ChatlistState` / `ChatlistSection`) already carry bson tags.

room-service/handler.go
```diff
@@ -2217,6 +2218,137 @@ func (h *Handler) favoriteToggle(c *natsrouter.Context) (*model.FavoriteToggleRe
...
+	if req.SectionID != nil {
+		if *req.SectionID == "" {
+			return nil, errcode.BadRequest("sectionId is empty", errcode.WithReason(errcode.UserChatlistSectionNotFound))
+		}
+		if model.IsBuiltinSection(*req.SectionID) {
+			// static built-in check only. A custom section's *existence*
+			// is not verified here (that would couple room-service to the
+			// user-service registry on every move) — orphan tolerance renders an
+			// unknown sectionId under Chats client-side.
+			return nil, errcode.BadRequest("cannot move a chat into a built-in section", errcode.WithReason(errcode.UserChatlistBuiltinTarget))
```
mliu33, line 2241: Does the spec say that users can't move chat into favorite section ?

Favorite section should not be built-in section

room-service/store_mongo.go
```diff
@@ -1094,6 +1094,166 @@ func (s *MongoStore) ToggleSubscriptionFavorite(ctx context.Context, roomID, acc
...
+		return nil, fmt.Errorf("move section for %q in room %q: %w", account, roomID, err)
+	}
+	return &result, nil
+}
+
+// ComputeSectionOrder midpoints a new position within (account, sectionID). Two
+// small indexed reads in the placed case (the afterRoomID sub, then the next
+// higher order); one in the append case (the section max).
+func (s *MongoStore) ComputeSectionOrder(ctx context.Context, account, sectionID, afterRoomID, beforeRoomID string) (float64, bool, error) {
+	secFilter := bson.M{"u.account": account, "sectionId": sectionID}
```
mliu33, line 1131: Make sure we have this index in MongoDB

room-service/store_mongo.go
```diff
@@ -1094,6 +1094,166 @@ func (s *MongoStore) ToggleSubscriptionFavorite(ctx context.Context, roomID, acc
...
+		if errors.Is(err, mongo.ErrNoDocuments) {
+			return next - 1, false, nil // head
+		}
+		if err != nil {
+			return 0, false, fmt.Errorf("lookup prev order: %w", err)
+		}
+		mid, exhausted := sectionMidpoint(prev.SectionOrder, next)
+		return mid, exhausted, nil
+	}
+	if afterRoomID == "" {
```
mliu33, line 1162: What does it mean when user pass beforeRoomID="" and afterRoomID="" ?

Always put at the end ?

room-service/store_mongo.go
```diff
@@ -1094,6 +1094,166 @@ func (s *MongoStore) ToggleSubscriptionFavorite(ctx context.Context, roomID, acc
...
+	var subs []model.Subscription
+	if err := cur.All(ctx, &subs); err != nil {
+		return nil, fmt.Errorf("rebalance decode section: %w", err)
+	}
+	if len(subs) == 0 {
+		return nil, nil
+	}
+	models := make([]mongo.WriteModel, 0, len(subs))
+	for i := range subs {
+		subs[i].SectionOrder = float64(i + 1)
```
mliu33, line 1245: We can use +10 to increase the gap between each order to have even lower chance of rebalancing

user-service/service/chatlist.go
```diff
@@ -0,0 +1,320 @@
...
+func (s *UserService) CreateChatlistSection(c *natsrouter.Context, req models.ChatlistSectionCreateRequest) (*model.ChatlistState, error) {
+	account := c.Param("account")
+	c.WithLogValues("account", account)
+	name := strings.TrimSpace(req.Name)
+	if err := validateSectionName(name); err != nil {
+		return nil, err
+	}
+	sortMode := req.SortMode
+	if sortMode == "" {
+		sortMode = model.SortModeCustom
```
mliu33, line 54: Custom is the default sort ?

mliu33 (approved):
Excellent work, thanks!

room-service/handler.go
mliu33: @chenjr0719 create a new PR to remove "favorite" as built in section

room-service/store_mongo.go
mliu33: @chenjr0719 consider increasing the gap between each order to reduce collision probability

room-service/store_mongo.go
chenjr0719: Added the `{u.account, sectionId, sectionOrder}` compound index in #207.

room-service/store_mongo.go
chenjr0719: Done in #207 — rebalance now spaces orders by 10 (`(i+1)*10`) instead of 1, so midpoint inserts rarely exhaust the gap.

room-service/store_mongo.go
chenjr0719: Yes — with both `beforeRoomID` and `afterRoomID` empty it appends at the section end (max order + gap). Added a doc comment clarifying that in #207.

user-service/service/chatlist.go
chenjr0719: Yes — an empty `SortMode` on section create defaults to `SortModeCustom`.

room-service/handler.go
chenjr0719: Tracked as a follow-up in #206 — the exact behavior (keep favoriting flag-driven vs. make Favorites a move-into section) is pending a spec review before we implement. Thanks for flagging it!

PR #138, opened 2026-07-28
Filter member statuses by statusIsShow opt-in flag

yenta:
## Summary
Enforce the `statusIsShow` flag when listing member statuses in a room. Members who have opted out of surfacing their status (`statusIsShow=false`) are now excluded from the `ListMemberStatuses` response, even if they have a non-empty `statusText`. This includes legacy user documents that lack the `statusIsShow` field entirely (which decode as `false`).

## Changes

- **store_mongo.go**: Updated the aggregation pipeline in `ListMemberStatuses` to filter on `statusIsShow: true` in addition to the existing non-empty `statusText` check. Updated the method's documentation to clarify that members with `statusIsShow=false` are dropped before the post-join limit.

- **store.go**: Updated the `ListMemberStatuses` interface documentation to clarify that subscriptions with `statusIsShow=false` are dropped and that every returned row has `statusIsShow=true`.

- **integration_test.go**: 
  - Fixed the existing test case to set `StatusIsShow: true` for Bob (aligning with the test's intent to verify status display).
  - Added two new test cases:
    - `"members with statusIsShow=false are excluded"` — verifies that members with an explicit `statusIsShow=false` are omitted from results.
    - `"member whose user doc lacks statusIsShow is excluded"` — verifies that legacy documents missing the `statusIsShow` field (decoding as `false`) are also omitted.

- **docs/client-api.md**: Updated API documentation to clarify that members with `statusIsShow=false` are omitted from the `members` array, and that the `statusIsShow` field in `MemberStatus` is always `true` (since members with `false` are filtered out).

- **docs/client-api/request-reply.md**: Updated success response documentation to note that only members with `statusIsShow=true` are returned.

## Implementation Details

The MongoDB aggregation pipeline now includes `statusIsShow: true` in the `$match` stage that filters members before applying the limit. This ensures:
- Members who explicitly opted out (`statusIsShow=false`) are excluded.
- Legacy documents missing the field (which decode as `false` in Go) are also excluded.
- The filtering happens before the `$limit`, so the limit applies only to members who meet both criteria (non-empty status text AND opted in).

https://claude.ai/code/session_01CqafXQZhtyc2PXUcdJxnRK

## Summary by CodeRabbit

* **Bug Fixes**
  * Member status results now exclude users who hide their status or have no visible status text (`statusIsShow` must be true and `statusText` must be non-empty).
  * Visible-status filtering is applied before enforcing the requested limit, so results match what’s actually reportable.

* **Documentation**
  * Updated “Get Member Statuses” API documentation to clarify returned fields and omission behavior (including legacy documents where `statusIsShow` is missing).

* **Tests**
  * Expanded integration coverage for status visibility opt-out and legacy omission scenarios.

### DIFF
--- a/docs/client-api.md
+++ b/docs/client-api.md
@@ -1679,7 +1679,7 @@ See [Error envelope](#6-error-envelope-reference). Common errors: `"only room me
 
 | Field | Type | Required | Notes |
 |---|---|---|---|
-| `limit` | number | no | Upper bound on returned rows. When omitted, the server uses `min(3, room.userCount)` (an empty room returns an empty list); when supplied, must be `> 0` and `<= room.userCount`. Fewer rows may come back — members with an empty `statusText` are omitted (see `members`). |
+| `limit` | number | no | Upper bound on returned rows. When omitted, the server uses `min(3, room.userCount)` (an empty room returns an empty list); when supplied, must be `> 0` and `<= room.userCount`. Fewer rows may come back — members with an empty `statusText` or with `statusIsShow=false` are omitted (see `members`). |
 
 ```json
 { "limit": 5 }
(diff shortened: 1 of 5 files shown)


mliu33 (approved):
Thanks

PR #139, opened 2026-07-28
Local/global room subject separation to reduce NATS gateway interest-map memory

hmchangw:
## Why

In the NATS supercluster, gateway **interest** propagation is per-account and all-or-nothing: because every chat user shares one NATS account, each connected client's per-room subscriptions (`chat.room.{roomID}.>`) get replicated into every peer site's interest map — including the large majority of rooms whose members are all on a single site and never need cross-site delivery. That replicated interest state is the memory pressure.

This PR lets same-site rooms deliver on a **leaf-filtered local namespace** (`chat.local.room.{roomID}.event`) that the platform team excludes from cross-gateway interest, while cross-site rooms stay on the propagated global subject (`chat.room.{roomID}.event`).

## How

- **Classification** — a sticky tri-state `Room.CrossSite *bool` (nil=unclassified→global fail-safe, &true=cross-site→global, &false=same-site→local), set by `room-worker` at every create/add path (channel, DM, `serverCreateDM`, member-add), reusing the exact points where cross-site federation is already decided. Never demoted (Phase 1). Data starts from scratch, so every room is classified at creation/member-add — no backfill of pre-existing rooms is needed.
- **Propagation** — surfaced through `roommetacache`, `RoomsInfoBatch`, the user-service `$lookup` + enrichment, into the `subscription.list` room payload as `crossSite`.
- **Routing gate** — `subject.RoomEventTargets(id, crossSite, crossSiteAt, mode, now)` behind `ROOM_SUBJECT_MODE` (`global` default → `dual` → `local`), routed through a single per-service `publishRoomEvent` helper at every room-event publish site (broadcast-worker ×3, room-service ×2, room-worker rename). A `.semgrep` rule fails CI on any inline room-subject publish that bypasses the helper. Cross-site rooms are **always** global in every mode.
- **Transition grace window** — a same-site→global flip records `Room.CrossSiteAt`; for `RoomLocalityGrace` (15 min) afterward, publishers emit the room's `.event` to **both** the local and global subjects (in local/dual mode). So a same-site member still subscribed to the local subject keeps receiving in real time until it re-subscribes — the flip gap is bounded **server-side**, independent of whether the re-subscribe nudge is delivered or how often the client reconciles. Bounded to actively-flipping rooms only; rooms born cross-site get no grace.
- **Client** — `chat-frontend` (internal test client) subscribes each room to the local or global subject by the flag, with re-subscribe-on-flip and a missing→global fail-safe.

## New NATS subjects

One new **local** room-event subject is introduced alongside the existing global one. The unit is the whole `chat.local.room.{roomID}.>` subtree, but only the `.event` lane has publishers in this PR.

| Subject | Direction | Emitted for | Gateway behavior |
|---|---|---|---|
| `chat.local.room.{roomID}.event` **(new)** | server → client | Same-site rooms (`crossSite=false`): in `ROOM_SUBJECT_MODE=local`, and in `dual` (both lanes); plus a flipped room during its grace window | Denied at the leaf node from cross-gateway interest propagation → site-local |
| `chat.room.{roomID}.event` *(existing)* | server → client | Cross-site rooms, not-yet-classified rooms, and everything under the default `global` mode | Propagated across the supercluster (unchanged) |

New JWT **subscribe** grant: `chat.local.room.>` (added to the auth-service scope template and `docker-local/setup.sh`; the prod template is infra-owned). Clients only *subscribe* to room subjects — no publish grant changes. The locality-aware `stream.msg` / `event.metadata.update` builders cover the rest of the `chat.local.room.{roomID}.>` subtree but have no local publishers yet.

## What the client must do

1. **Read `crossSite`** (tri-state) on each room in the `subscription.list` payload and choose the room-event subject:
   - `true` → `chat.room.{roomID}.event` (cross-site → global)
   - `false` → `chat.local.room.{roomID}.event` (confirmed same-site → local)
   - **absent/unknown → `chat.room.{roomID}.event`** (fail-safe global — a room the server hasn't classified yet)
2. **Carry the `chat.local.room.>` subscribe grant** in the client JWT (scope-template / infra change).
3. **Re-subscribe on flip** — when a room's `crossSite` changes (a same-site room gains an off-site member → global), move its subscription to the new subject. Correctness rests on the persisted flag + history/gap fetch, so a client that reconciles `subscription.list` on reconnect / periodically is correct; the **transition grace window** keeps it gapless for any client that re-subscribes within 15 min. For *prompt* re-routing the official client should also subscribe to `chat.user.{account}.event.room.update` (the server's locality nudge — distinct from `event.room.metadata.update`) and refresh `subscription.list` on receipt.
4. **No publish change** — clients still publish sends on the user-scoped tree; only room-event *subscriptions* are affected.

Old clients that ignore `crossSite` keep working: they stay on `chat.room.>`, and with the default `global` mode (and the zero-gap `dual` window) the server keeps publishing there.

## Fail-safe invariant

Global is the safe default everywhere: misrouting a global room to local would silently drop cross-gateway delivery, so the system only routes local when `crossSite == false` is authoritative. **Default `ROOM_SUBJECT_MODE=global` makes this entire change a no-op** (byte-identical publishing) until ops opts in — safe to merge inert.

## Rollout (see the runbook in the design doc)

1. Platform: add `chat.local.room.>` to the prod subscribe template + configure the leaf-node deny for `chat.local.>`. (No data backfill — deployments start from fresh data; every room is classified at creation/member-add.)
2. Deploy publishers `global → dual` (dual publishes same-site rooms to **both** prefixes → zero-gap client migration), deploy the frontend, then `dual → local`.

**Release gate (official client, not the chat-frontend test client):** with the transition grace window, a per-room local→global flip is gapless for any client that re-subscribes within `RoomLocalityGrace`, and never loses data (persisted + backfilled). Before enabling `local` in prod, the official client must implement the flip contract in "What the client must do" §3 (bootstrap-by-`crossSite`, reconnect reconcile, gap-fetch, and ideally the nudge handler) so its reconcile interval stays within the grace window.

## Testing

- Unit tests (TDD, mocked stores) and `golangci-lint` are green tree-wide; a `.semgrep` guard rule prevents future inline room-subject publishes.
- Mongo-backed **integration tests were written and vet-checked under `-tags integration`** — CI runs them.
- Frontend: vitest suite green, typecheck + build clean.

The fail-safe invariant was verified at every publish site, create/add path, and projection. Several silent-failure bugs (two dropped `crossSite` Mongo projections, a `SubscriptionRoom.crossSite` `omitempty` that would have defeated the feature, two missing create-path classifications, a broadcast-worker early-return that skipped the dual-mode global fallback, and a room-meta cache-invalidation race on the flip) were caught and fixed.

Design + plan: `docs/superpowers/specs/2026-07-28-local-global-room-subjects-design.md`, `docs/superpowers/plans/2026-07-28-local-global-room-subjects.md`.

https://claude.ai/code/session_01NKLjVmfDJQKfYgmLLofk3v

---
_Generated by [Claude Code](https://claude.ai/code/session_01NKLjVmfDJQKfYgmLLofk3v)_

### DIFF
--- a/broadcast-worker/handler.go
+++ b/broadcast-worker/handler.go
@@ -66,9 +66,10 @@ type Handler struct {
 	parentFetcher ParentFetcher
 	encrypt       bool
 	encoder       *roomcrypto.Encoder
+	routeMode     subject.RoomRouteMode
 }
 
-func NewHandler(store Store, userStore userstore.UserStore, pub Publisher, keyStore RoomKeyProvider, parentFetcher ParentFetcher, encrypt bool) *Handler {
+func NewHandler(store Store, userStore userstore.UserStore, pub Publisher, keyStore RoomKeyProvider, parentFetcher ParentFetcher, encrypt bool, routeMode subject.RoomRouteMode) *Handler {
 	return &Handler{
 		store:         store,
 		userStore:     userStore,
@@ -77,6 +78,7 @@ func NewHandler(store Store, userStore userstore.UserStore, pub Publisher, keySt
 		parentFetcher: parentFetcher,
 		encrypt:       encrypt,
 		encoder:       roomcrypto.NewEncoder(),
+		routeMode:     routeMode,
 	}
 }
 
@@ -198,9 +200,9 @@ func (h *Handler) handleCreated(ctx context.Context, evt *model.MessageEvent) er
 
 	switch meta.Type {
 	case model.RoomTypeChannel:
-		return h.publishChannelEvent(ctx, meta, clientMsg, evt.Timestamp, resolved.MentionAll, resolved.Participants)
+		return h.publishChannelEvent(ctx, &meta, clientMsg, evt.Timestamp, resolved.MentionAll, resolved.Participants)
 	case model.RoomTypeDM, model.RoomTypeBotDM:
-		return h.publishDMEvents(ctx, meta, clientMsg, evt.Timestamp, resolved.Accounts)
+		return h.publishDMEvents(ctx, &meta, clientMsg, evt.Timestamp, resolved.Accounts)
 	default:
 		slog.WarnContext(ctx, "unknown room type, skipping fan-out",
 			"type", meta.Type,
@@ -259,7 +261,7 @@ func (h *Handler) handleThreadCreated(ctx context.Context, evt *model.MessageEve
 		// Do NOT call SetSubscriptionMentions here: TShow=false replies are invisible
 		// in the main channel, so a room-level mention badge would appear with no
 		// visible message to explain it.
-		roomEvt := buildRoomEvent(meta, clientMsg, evt.Timestamp)
+		roomEvt := buildRoomEvent(&meta, clientMsg, evt.Timestamp)
 		roomEvt.MentionAll = resolved.MentionAll
 		if len(resolved.Participants) > 0 {
 			roomEvt.Mentions = resolved.Participants
@@ -274,7 +276,7 @@ func (h *Handler) handleThreadCreated(ctx context.Context, evt *model.MessageEve
 		// owned by message-worker (markThreadMentions), so broadcast-worker doesn't
 		// touch subscriptions here. lastMsgAt is intentionally NOT updated (would
 		// wrongly mark hasUnread for non-participants).
-		return h.publishDMEvents(ctx, meta, clientMsg, evt.Timestamp, resolved.Accounts)
+		return h.publishDMEvents(ctx, &meta, clientMsg, evt.Timestamp, resolved.Accounts)
 	default:
 		slog.WarnContext(ctx, "unknown room type, skipping thread fan-out",
 			"type", meta.Type,
@@ -460,8 +462,8 @@ func (h *Handler) publishThreadMetadata(ctx context.Context, room *model.Room, n
 	}
 	switch room.Type {
 	case model.RoomTypeChannel:
-		if err := h.pub.Publish(ctx, subject.RoomEvent(room.ID), payload); err != nil {
-			return fmt.Errorf("publish thread metadata for channel room %s: %w", room.ID, err)
+		if err := h.publishRoomEvent(ctx, room.ID, room.CrossSite, room.CrossSiteAt, payload, "thread metadata"); err != nil {
+			return err
 		}
 	case model.RoomTypeDM, model.RoomTypeBotDM:
 		for _, account := range room.Accounts {
@@ -666,10 +668,7 @@ func (h *Handler) publishMutation(ctx context.Context, room *model.Room, roomEvt
 
 	switch room.Type {
 	case model.RoomTypeChannel:
-		if err := h.pub.Publish(ctx, subject.RoomEvent(room.ID), payload); err != nil {
-			return fmt.Errorf("publish %s event for room %s message %s: %w", roomEvtType, room.ID, messageID, err)
-		}
-		return nil
+		return h.publishRoomEvent(ctx, room.ID, room.CrossSite, room.CrossSiteAt, payload, fmt.Sprintf("%s event (message %s)", roomEvtType, messageID))
 
 	case model.RoomTypeDM, model.RoomTypeBotDM:
 		for _, account := range room.Accounts {
@@ -797,7 +796,7 @@ func (h *Handler) encryptRoomEvent(ctx context.Context, roomID string, clientMsg
 	return nil
 }
 
-func (h *Handler) publishChannelEvent(ctx context.Context, meta roommetacache.Meta, clientMsg *model.ClientMessage, timestamp int64, mentionAll bool, mentions []model.Participant) error {
+func (h *Handler) publishChannelEvent(ctx context.Context, meta *roommetacache.Meta, clientMsg *model.ClientMessage, timestamp int64, mentionAll bool, mentions []model.Participant) error {
 	evt := buildRoomEvent(meta, clientMsg, timestamp)
 	evt.MentionAll = mentionAll
 	if len(mentions) > 0 {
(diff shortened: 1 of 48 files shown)


hmchangw:
Pushed fixes addressing the automated review (`4e8f242`, `47fd4d4`, `fb172d0`):

- **`sast` CI failure** (gosec G117) — false positive on the `SubscriptionRoom` serialization test (marshals a struct carrying a `PrivateKey` wire field, unset). Suppressed with a justified `// #nosec G117`, matching the existing precedent in the file. (`4e8f242`)

- **🔴 Critical — AddMembers cross-site mark not retry-safe.** Correct — on a JetStream redelivery where subscriptions already exist, `needSub` is empty and the mark was skipped before the early-return. Fixed by projecting each member's home `SiteID` onto `AddMemberCandidate` and classifying from the **full candidate set before the early-return** (`SetRoomCrossSite` is an idempotent `$set`). Added redelivery regression tests. (`47fd4d4`)

- **🟠 Major — preserve unknown locality instead of routing it locally.** Correct and important. `Room.CrossSite` (and the `SubscriptionRoom`/`EnrichedSubscription`/`RoomInfo`/`roommetacache.Meta` mirrors) are now **`*bool` tri-state**, following the repo's existing `User.Active *bool` / `IsActive()` pattern: `nil` (unclassified/unbackfilled) → **global fail-safe** (omitted on the wire → frontend `?? true`), `&false` (confirmed same-site) → local, `&true` → global. New `Room.RoutesGlobal()` helper; `RoomEventTargets` maps `nil`→global. Every create path now classifies definitively, and the backfill writes an explicit `true`/`false` for every scanned room. This also resolves the wire-contract note (`crossSite` now omits when unknown, serializes explicit `false` only for confirmed same-site). (`fb172d0`)

- **🟠 Major — newly-added same-site room routes global until refresh in `local` mode / room-update nudge.** Acknowledged as a **documented release-gate follow-up**: the frontend has no live `subscription.list`-refresh handler yet, so a connected client re-routes on reconnect. This cannot manifest in `global` (default) or `dual` (migration) mode — only after an explicit cut to `local` — and the rollout runbook gates the `dual→local` step behind that frontend work. Tracked, not addressed here.

Remaining Minor items (config-test env restore; baseline-subscribe doc bullet; design non-goals dual-mode wording) are noted and can be folded in if desired.

---

hmchangw:
Addressed the remaining actionable items from the latest review round (commits `7f51b49`, `3b83e58`, `cc1bc31`):

- **`backfill-crosssite/main.go` — SIGINT/SIGTERM cancellation.** `context.Background()` → `signal.NotifyContext`, so an operator interrupt / pod eviction cancels the in-flight scan/writes. `stop()` is called explicitly before `os.Exit` (run() owns the deferred teardown), avoiding the `exitAfterDefer` gocritic. Used `signal.NotifyContext` rather than `pkg/shutdown.Wait` — the latter *blocks until a signal* and is built for long-running daemons, so it would hang this run-to-completion batch after the work finishes.
- **`loadgen/daily_pool.go` — both room-event lanes.** Both pools now subscribe each room on the global **and** local subjects, so a run stays correct under `ROOM_SUBJECT_MODE=local`/`dual` instead of silently receiving nothing and reporting degraded E2E latency. Added local-lane regression tests for both pools.
- **`chat-frontend` `useRoomSubscriptions.js` — stale-session guard.** `safeDispatch` (and the post-decryption `finalize`) now drop work whose effect generation is no longer current, closing the logout→login window where an in-flight decryption could leak a prior session's event into the new session. `cancelledRef` alone didn't cover it — the new effect resets it to `false`. Added a red/green regression test; full vitest suite (701) + typecheck green.

**Already resolved at the current head** (no change needed): broadcast-worker's `publishRoomEvent` accumulates the publish error and returns after the loop (dual-mode global fallback preserved); `SetRoomCrossSite` busts the room-meta cache at both handler sites; all three `config_test.go` files seed `ROOM_SUBJECT_MODE` unconditionally with `t.Setenv`; the `subscriptions_test` "decodes as nil, not false" comment and the `RoomInfo.CrossSite` bson tag are in place.

**Skipped:** the `Test<Type>_<Method>` rename nits in `pkg/subject/subject_test.go` — that file's established convention is `Test<Func>` (54 of 68 tests), so the flagged names already match it — and the `RoomEventsContext.test.jsx` "mock at the API boundary" nit (a broad test-philosophy refactor of already-passing tests, low value).

The Mongo-backed integration tests still need CI to execute (no testcontainer registry in the build env), as noted in the PR description.

---

broadcast-worker/handler.go
```diff
@@ -886,7 +898,7 @@ func (h *Handler) publishDMEvents(ctx context.Context, meta roommetacache.Meta,
 	return nil
 }
 
-func buildRoomEvent(meta roommetacache.Meta, clientMsg *model.ClientMessage, eventTimestamp int64) model.RoomEvent {
+func buildRoomEvent(meta *roommetacache.Meta, clientMsg *model.ClientMessage, eventTimestamp int64) model.RoomEvent {
```
mliu33, line 901: Why are we changing to pointer type for meta ? If meta is nil, this function could possibly throw runtime error when accessing meta.Name, right ?

room-worker/handler.go
```diff
@@ -1236,6 +1255,22 @@ func (h *Handler) processAddMembers(ctx context.Context, data []byte) (err error
 		}
 		accountsBySite[siteID] = append(accountsBySite[siteID], acc)
 	}
+
+	// Remote-site bucket non-empty → mark global (sticky). room.CrossSite still
+	// holds the PRIOR state (SetRoomCrossSite doesn't mutate it), so gate the nudge
+	// on RoutesGlobal(): only a previously-confirmed same-site (&false) room flips.
+	if len(accountsBySite) > 0 {
+		if err := h.store.SetRoomCrossSite(ctx, req.RoomID); err != nil {
```
mliu33, line 1263: Why are we setting cross site again when we already do this in line 864?

room-worker/handler.go
```diff
@@ -1260,6 +1295,32 @@ func (h *Handler) processAddMembers(ctx context.Context, data []byte) (err error
 	return nil
 }
 
+// nudgeLocalToGlobalTransition best-effort nudges existing members (subscribers minus
+// newAccounts) to re-fetch subscription.list and re-subscribe global after a flip, via
+// their per-user tree. Pure latency opt — crossSite is source of truth, so nudge failures
+// are swallowed and clients self-correct on reconnect.
+func (h *Handler) nudgeLocalToGlobalTransition(ctx context.Context, roomID string, newAccounts []string, now time.Time) {
```
mliu33, line 1302: I think it's better to keep it simple that we always publish both global and local (dual by default). Retriggering all existing members to refetch subscription.list is heavy pressure on server side. Client will obtain correct room cross site on the next logic, which is good enough.

So I think we can remove this function. Besides, this function will fetch all room subscribers, which is a heavy operation for large rooms

mliu33 (approved):
Excellent, thanks! Just some suggestions

PR #140, opened 2026-07-28
docs(load-testing): add SLI/SLO spec (system/sli-slo.md)

flywindy:
## Summary

Adds the platform's user-facing SLI/SLO specification at
`docs/load-testing/system/sli-slo.md` — the acceptance criteria that all
system-level load tests assert against. (Moved here from `docs/specs/o11y/` to
honor the `docs/load-testing/README` intended structure: user-facing SLI/SLO
belongs under `system/`.)

**10 SLOs across 5 critical user journeys** (J1 is split — see Approach):

| Journey | SLOs |
|---|---|
| J1 Send a message | persistence 99.9% (SLO-1a) · channel broadcast **enqueue-accepted** 99.9% (SLO-1b) · enqueue ≤1s 99% (SLO-2) |
| J2 Interactive workflows | login ≤1s 99% (SLO-3) · enter channel ≤500ms 95% (SLO-4) · enter thread ≤300ms 99% (SLO-5) |
| J3 Notifications | push-stream handoff 99.9% (SLO-6) |
| J4 Search | availability 99.5% (SLO-7) · latency ≤1s 95% (SLO-8) |
| J5 Federation | forward freshness ≤30s 99% (SLO-9) |

## Approach

- **Journey-first, not service-first** (Google CUJ methodology; matches how chat
  products like Slack's Service Delivery Index measure themselves).
- **Every SLI is an event-based `good / valid` ratio** over a 28-day window.
  Percentiles are **diagnostic only, never SLO targets**. Latency SLOs are
  written as "X% of valid requests within the bound".
- **J1 is split, not one "delivery success".** The persist and publish paths
  consume `MESSAGES_CANONICAL` independently (no message-ID join), so J1 is two
  success ratios (1a persistence, 1b channel enqueue) plus a latency ratio (2) —
  each its own budget. A correlated single-J1 outcome is v2.
- **Honest v1 scoping, aligned to the code:**
  - A channel broadcast is **one core-NATS room-subject publish**; `PublishMsg`
    returns on **local enqueue**, not server-confirmed publication. SLO-1b/2 are
    therefore **enqueue-acceptance** (`broadcast_channel_enqueue_total` /
    `_age_seconds`), not publication — true publication latency is deferred to P7.
  - The denominator is `broadcast_path=room_subject`-scoped (channel thread
    replies route to thread fan-out and are excluded); per-recipient
    `all/partial/failed` lives only on DM/thread (v1-unscored).
  - Async ratios are marked **approximate (lag-enforced)**; consumer backlog /
    a connection-risk backstop are the primary signals. Denominators are anchored
    **upstream** so an outage reads as a dropped ratio, not missing data.
- **SLI spec / implementation split** — each SLI is written in user language,
  separately from what is measurable today (✅) vs on the roadmap (🔧).
- **Measurement roadmap (§8, P1–P7)** — all items land in `newchat` (app code)
  or infra; no flywindy/o11y SDK change is required.
- **Load-testing alignment (§10)** — SLOs are the acceptance criteria; driver is
  `tools/loadgen` (NATS), not k6. The hard gate reuses the production
  predicate/target over an **isolated, quiescent test site** with a
  warm-up → send → settle window (never the 28-day aggregate rule).

## Notes for reviewers

- All targets are **achievable-first starting values** — calibrated over 4–6
  weeks of observation before approval (no paging alerts until then).
- **SLO ≠ SLA**: internal document; any external SLA must be derived (looser,
  availability-focused, post-calibration).
- English is canonical. A zh-TW translation is maintained separately (Artifact),
  not part of this PR.
- Docs-only change; no code touched.

### DIFF
--- a/docs/load-testing/README.md
+++ b/docs/load-testing/README.md
@@ -6,6 +6,9 @@ component-specific and system-wide test plans.
 
 ## Current Plans
 
+- [`system/sli-slo.md`](system/sli-slo.md) is the platform's user-facing
+  SLI/SLO specification — the acceptance criteria all system-level load tests
+  assert against.
 - [`cassandra/soak-test-plan.md`](cassandra/soak-test-plan.md) is the
   authoritative specification for the Cassandra Run A pre-production soak.
 - [`cassandra/run-a-implementation-plan.md`](cassandra/run-a-implementation-plan.md)
--- a/docs/load-testing/system/sli-slo.md
+++ b/docs/load-testing/system/sli-slo.md
@@ -8,9 +8,11 @@
 | **Approval date** | — |
 | **Revisit date** | ~6 weeks after approval (end of calibration, §0.2) |
 
-Companions: `o11y-metrics-inventory.md`, `o11y-trace-design.md`,
-`o11y-performance-and-sampling.md`. Error budget policy: placeholder
-(`o11y-error-budget-policy.md`, to be written).
+Companions (o11y specs): `../../specs/o11y/o11y-metrics-inventory.md`,
+`../../specs/o11y/o11y-trace-design.md`,
+`../../specs/o11y/o11y-performance-and-sampling.md`. Error budget policy:
+placeholder (`o11y-error-budget-policy.md`, to be written). Load-test alignment:
+`../end-to-end-load-test-plan.md` (§10 here is the summary; the plan owns execution).
 
 ---
 
@@ -113,7 +115,7 @@ material; read receipts / unread badges → dashboard, v2 candidate (§8 P7).
 |---|---|---|---|---|
 | SLO-1a | J1 | canonical message persisted to Cassandra / **canonical messages published** | 99.9% | 🔧 P2 · approximate (lag-enforced) |
 | SLO-1b | J1 | **channel** room-subject broadcast **enqueue-accepted** (local core-NATS, single publish) / **canonical messages routed to the room-subject path** (`broadcast_path=room_subject`) | 99.9% | 🔧 P2 · approximate · enqueue-only (see §2) |
-| SLO-2 | J1 | room-subject broadcast published **within 1 s** of canonical acceptance (late = bad) / canonical messages on the room-subject path (`broadcast_path=room_subject`) | 99% | 🔧 P2 · approximate |
+| SLO-2 | J1 | room-subject broadcast **enqueue-accepted within 1 s** of canonical acceptance (late = bad) / canonical messages on the room-subject path (`broadcast_path=room_subject`) | 99% | 🔧 P2 · approximate · enqueue-only |
 | SLO-3 | J2 | successful login **within 1 s** / eligible login attempts | 99% | ✅ (auth leg — proxy) |
 | SLO-4 | J2 | channel load succeeds **within 500 ms** / eligible channel loads | 95% | 🔧 P1 |
 | SLO-5 | J2 | thread open succeeds **within 300 ms** / eligible thread opens | 99% | 🔧 P1 |
@@ -173,7 +175,7 @@ a protocol-receive prober (loadgen) and a render prober (browser) — neither is
   `room_subject`, else DM/BotDM → `dm`. This matters because a **channel thread
   reply** (`TShow=false`) is a channel room yet routes to per-account thread
   fan-out, **not** `publishChannelEvent` — a `room_type="channel"` denominator
-  would count it while no `broadcast_channel_publish_total` fires, wrongly
+  would count it while no `broadcast_channel_enqueue_total` fires, wrongly
   depressing SLO-1b/2. (gatekeeper resolves room type at the emit site to set the
   label; `thread`/`dm` slices are v1-unscored.)
   - **SLO-1a** uses the **all-`broadcast_path` total** (persistence covers every message).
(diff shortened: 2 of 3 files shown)


flywindy:
Overall, the direction is strong: organizing SLOs around CUJs rather than services, starting with calibration, separating enforcement from observational metrics, and treating consumer lag as a leading indicator all fit this project's event-driven architecture well. Before adopting the specification, I recommend clarifying the following:

1. **Zero-traffic slices should not automatically count as good.** For low-volume journeys such as notifications and federation, idle slices would significantly inflate compliance. Exclude them from the denominator and define a minimum-sample / `insufficient_data` policy, or supplement them with scheduled synthetic probes.
2. **Percentile SLOs do not yet define an error budget.** SLO-2/4/5/8/9 should state whether percentiles are calculated per 5-minute slice or over the full 28-day window, and what percentage of slices must pass. Without this, the burn-rate alerts in §7 cannot be applied consistently. Prefer `successful within bound / valid attempts` for budget accounting, with percentiles retained as diagnostic signals.
3. **SLO-1's `delivered ∧ persisted` condition cannot be derived correctly from independent cross-service counters.** Prometheus counters cannot join outcomes by message ID, and broadcast and persistence may complete in different slices. Either split this into persistence and publication SLIs, or emit a single workflow outcome from a completion ledger or synthetic probe.
4. **The client-error exclusion is too broad.** Excluding every errcode except `internal` would prevent `unavailable`, server-side timeouts, and 429/capacity rejection from burning budget. Eligibility should be based on whether the request was valid and expected to succeed, not solely on the wire error category.
5. **A server publish is not a per-recipient delivery.** The channel path publishes once to a room subject, and `UserCount` is not the number of connected recipients. Until client observation or probing is available, name this `broadcast publication success/latency` to avoid implying that recipients received or rendered the message.
6. **Measurability and scope need further tightening.** SLO-3 currently measures only the auth leg, not full app entry. SLO-8's histogram has no outcome/status label, so fast failures can improve p95. SLO-9 covers only the OUTBOX relay, not direct-INBOX publication or remote apply, and its budget should be isolated by `{origin_site,dest_site}`.

The first four items should be resolved before approval because they directly determine whether the SLO and error-budget figures are trustworthy. The remaining items can be addressed by tightening names, measurable-status labels, and scope.

flywindy:
Re-reviewed at `47d99fd8`. Most previous concerns are now addressed: the document explicitly marks approximate accounting, narrows notification scope to stream handoff, and describes loadgen readiness honestly. A few issues remain:

1. **The channel fan-out contract still contradicts the code.** The channel path performs one core-NATS room-subject publish; it does not publish one user-room event per account. Therefore channel v1 cannot emit a meaningful `all/partial/failed` recipient outcome?only `success/failed` for the local publish call. Keep recipient delivery as the observational last-mile SLI; defer per-recipient outcomes to DM/thread or a client prober.
2. **SLO-1b/2 need a channel-scoped denominator.** The contract still names the unfiltered `messages_canonical_published_total`, while v1 explicitly excludes DM/thread. Specify a bounded `room_type="channel"` label or a separate channel counter; SLO-1a may continue using the all-message total.
3. **The standard NATS exporter does not expose `oldest_pending_age`.** Consumer state provides pending/ack-pending counts and sequence positions, but not the timestamp of the oldest pending message. P3 must either add a custom monitor that resolves the earliest relevant stream message timestamp, or enforce SLO-9 initially with pending/ack-pending growth instead of claiming the exporter supplies this gauge.
4. **Mark SLO-6 as approximate and keep units consistent.** The summary row should say `recipients accepted / notifiable recipients`. Batch redelivery can still double-count successful duplicate PubAcks because the current emitter discards `PubAck.Duplicate`; P7 already acknowledges that an outcome ledger is needed for exact accounting.
5. **Do not claim loadgen unlocks render-level last mile.** Existing E2 correlation proves protocol-level NATS receipt, not decrypt/render/state application. Split P6 into a loadgen protocol-receive prober and browser synthetic/RUM for render.

Minor cleanup: distinguish permanent poison drops from `MaxDeliver` exhaustion, change SLO-7's denominator to `eligible search requests`, and update the PR description, which still refers to percentile/time-slice SLOs and full loadgen validation.

I would resolve items 1?3 before approval; the rest are small consistency fixes.

flywindy:
Re-reviewed the latest head (091dc84). The previous findings are addressed. Two remaining clarifications: (1) SLO-1b/2 boundary: nc.PublishMsg returning nil only proves local Core-NATS enqueue, not server-confirmed publication. A reconnect-buffer loss can therefore look successful while JetStream lag stays zero. Either name v1 enqueue acceptance and add connection/async-error backstops, or introduce a server-confirmed boundary (for example, a flush checkpoint or durable PubAck path) before calling it publication. (2) Loadgen hard gate: reuse the production good/valid predicates and targets, but evaluate isolated, run-window counter deltas. Reading the 28-day production aggregate directly would let historical/background traffic mask failures in the current run. Implementation note: P3 recording rules should filter is_consumer_leader=true before aggregating JetStream consumer state to avoid duplicate/follower series. Once the first two points are clarified, the spec looks ready.

flywindy:
Re-reviewed the latest head (`f584e33`) against the current code and loadgen. The previous findings are addressed; no architectural blocker remains. A few consistency fixes before approval:

- §10 should say **the same production source counters and good/valid predicates, evaluated over isolated run-window deltas**—not “the same production recording rules.” The load plan should also drain/reset after warm-up so warm-up completions cannot inflate the numerator.
- E2 protocol receipt is observational and does not measure the enforced SLO-1b/2 enqueue boundary. Mark enqueue assertion as missing, or split the two boundaries explicitly.
- Finish the terminology cleanup: `publication age/channel publication` → enqueue age/acceptance; P4 is a `connection-risk backstop`, not an enqueue-loss backstop; P7 should migrate **SLO-1b/2**, not only SLO-1b.
- Fix or remove the broken `../end-to-end-load-test-plan.md` link; the intended plan is a sibling under `system/`.

After these small fixes, LGTM.

mliu33 (approved):

PR #148, opened 2026-07-29
feat/implement-translation-api

YI13:
## Summary

Adds a new **`translation-service`** — an on-demand text-translation RPC over NATS **request/reply**. A client sends a `TranslateRequest` and receives a `TranslateResult` (the translated text) as the synchronous reply, or the standard errcode error envelope on failure. It uses the same `natsrouter.Register` shape every other data-returning RPC in the codebase uses.

The translation backend is pluggable: this PR ships a **mock** backend (no network) plus a **third-party SSE-streaming** backend that is fully built and unit-tested, enabled by config once an endpoint is provided (no code change).

## Client contract

| Action | Subject | Payload |
|---|---|---|
| **Request** | `chat.user.{account}.request.translate.{siteID}` | `TranslateRequest{ text, targetLang }` |
| **Reply** | auto-generated `_INBOX.>` (NATS request/reply) | `TranslateResult{ translatedText, targetLang }` on success, or `{ code, reason?, error }` on failure |

- `targetLang` is a **BCP-47 tag** — the value stored in user settings `translateMessageInto` (e.g. `en-US`, `zh-Hant-TW`) — normalized server-side to the backend code (`zhTW`/`zhCN`/`en`/`de`/`ja`). A bare `zh` and any unsupported tag return `unsupported_lang`.
- Errors flow through the standard errcode envelope: `empty_text` / `unsupported_lang` (`bad_request`), `unavailable` under handler saturation, `internal` for a backend failure. The raw backend cause is logged server-side and never returned to the client.

## What's included

- **Wire contract** — `pkg/subject` builders (`TranslateRequest`, `TranslateRequestPattern`), `pkg/model` types (`TranslateRequest`, `TranslateResult`), `pkg/errcode` reasons (`unsupported_lang`, `empty_text`).
- **Translator backend** — pluggable interface with a `mock` and a `stream` impl. The streaming client POSTs `{text, targetLang, applyWiki:false}`, concatenates the SSE `returnData.translation` chunks **verbatim (whitespace preserved)** until `data: [DONE]`, and errors on any non-`96200` chunk or a stream that ends without `[DONE]`.
- **Token auth** — J1→J2 exchange, cached until shortly before expiry (minus skew). On a `failed to verify jwt` rejection the J2 token is force-refreshed and the call retried once.
- **Handler** — `natsrouter.Register` returning `(*TranslateResult, error)`; validates (empty text, unsupported language), calls the backend, and replies. Handler saturation replies `unavailable` rather than dropping the request.
- **main / config** — `caarlos0/env`, `TRANSLATION_BACKEND` selection (fails fast when `stream` lacks its endpoint / accessToken URL / J1 token), `obs.Init`, graceful shutdown.
- **deploy/** — Dockerfile (`golang:1.25.12-alpine` → `alpine:3.21`), docker-compose, azure-pipelines.
- **Docs** — `docs/client-api.md` §3.6 + the derived `request-reply.md` / `events.md` views, plus the design spec/plan.

## Testing

- TDD throughout. Unit tests cover happy path, empty text, unsupported/ambiguous language, backend error, token refresh-and-retry, and all SSE branches (merge/whitespace, non-`96200`, missing `[DONE]`, HTTP error, malformed chunk).
- Integration test (`//go:build integration`) drives the real `natsrouter` via `nc.Request`, asserting both a success `TranslateResult` reply and a `bad_request` envelope reply.
- `go test -race` green; `golangci-lint` 0 issues; `go build` clean; `go.mod` unchanged.

## Transport note

Earlier revisions used a fire-and-forget publish + async result on `chat.user.{account}.response.{requestID}`. Per review, the RPC was changed to **synchronous request/reply** in `e41178a`: a data-returning RPC gets an immediate, correlated reply; validation errors surface at once; and handler saturation replies `unavailable` instead of a silent drop.

## Summary by CodeRabbit

- **New Features**
  - Added a synchronous text-translation request/reply API.
  - Updated translation subjects to include a `.text` action segment.
  - Supports translated-text replies with echoed target language, plus mock and streaming backends.

- **Bug Fixes**
  - Improved validation errors for empty text, unsupported/ambiguous languages, and stricter `targetLang` BCP-47 tags.
  - Clearer error behavior for malformed requests, handler saturation, and backend translation failures.

- **Documentation**
  - Updated client API and request/reply documentation for the new subject and error semantics.
  - Updated design/spec guidance and local-site requirements.

- **Chores**
  - Added translation-service Docker and CI/CD configuration.


translation-service/main.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/hmchangw/chat/pkg/natsrouter"
+	"github.com/hmchangw/chat/pkg/natsutil"
+	"github.com/hmchangw/chat/pkg/obs"
+	"github.com/hmchangw/chat/pkg/shutdown"
+	"github.com/hmchangw/chat/pkg/subject"
+)
+
+type NATSConfig struct {
+	URL       string `env:"URL,required"`
+	CredsFile string `env:"CREDS_FILE" envDefault:""`
```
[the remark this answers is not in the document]
YI13: Leaving this one as-is intentionally — a few reasons:

- **Repo convention.** Every NATS-connecting service in the monorepo (~30 of them — `message-worker`, `broadcast-worker`, `room-service`, `outbox-worker`, …) declares the creds file as `env:"NATS_CREDS_FILE" envDefault:""`; none use `,required`. Making `translation-service` the sole exception would diverge from the shared `NATSConfig` pattern rather than follow it.
- **No-auth NATS is a supported mode.** `pkg/natsutil.Connect` only stats/attaches `nats.UserCredentials` when the path is non-empty; an empty path connects without credentials. That's exactly how the dev/local stacks that run a plain `nats --jetstream` server (e.g. the `data-migration/*` composes) connect, and requiring the file here would break them.
- **Not a secret leak.** The field is a *path* resolved at deploy time; the credential material itself is never defaulted, hard-coded, or logged. The "never default secrets/connection strings" rule targets secret values, which still holds — `NATS_URL` remains `required`.

If the team wants to tighten this, the right move is a repo-wide change to the shared NATS-config convention (all services at once), not a single-service divergence — happy to do that as a separate PR if that's preferred.

---

translation-service/main.go
```diff
@@ -0,0 +1,106 @@
...
+	publish := func(ctx context.Context, subj string, data []byte) error {
+		return nc.PublishMsg(ctx, natsutil.NewMsg(ctx, subj, data))
+	}
+
+	handler := NewHandler(translator, publish)
+	// Cap in-flight handlers so a slow translate backend can't accumulate goroutines
+	// and outbound connections without ceiling; saturation drops the fire-and-forget
+	// request (logged) rather than piling up.
+	router := natsrouter.Default(nc, "translation-service", natsrouter.WithMaxConcurrency(cfg.MaxWorkers))
+	natsrouter.RegisterVoid(router, subject.TranslateRequestPattern(cfg.SiteID), handler.Translate)
```
chenjr0719: Transport shape: this registers as `RegisterVoid` (fire-and-forget, no reply). In this repo `RegisterVoid` is otherwise used only for one-way signals (presence heartbeats) — every data-returning RPC uses `Register` with a typed sync reply. As a request/reply RPC (text → translated text) the client would get an immediate accept/validation-error and there'd be no silent drop under `MaxConcurrency` saturation (a `RegisterVoid` message is dropped with only a warn log when the semaphore is full). Is the async publish + response-subject shape intentional, or should it align to the request/reply RPC in the original design?

translation-service/lang.go
```diff
@@ -0,0 +1,57 @@
+package main
+
+import "strings"
+
+// normalizeTargetLang maps a client-supplied BCP-47 language tag (the same shape
```
chenjr0719: Minor: this doc block runs a bit long — could trim to the *why* (the ambiguous-`zh` rejection is the non-obvious part worth keeping). The repo's `/simplify` would catch this and the `newTokenProvider` block too.

chenjr0719 (commented):
Nice work — this is a well-built service. Splitting my notes into two parts: an implementation review (code quality) and a design question that's really yours to call.

## Implementation
Clean backend abstraction — the `Translator` interface with a mock + stream impl, env-selected and fail-fast so a deploy can't silently serve mock. Errors are sanitized at the client boundary (generic message out, raw cause logged server-side, the upstream `returnMessage` never propagated), a `TranslateResult` is published on every error path, the container runs non-root, and the docs are complete across `client-api.md` + `events.md` + `request-reply.md`. A few polish notes:
- **Comments** — a couple of doc blocks run long (`normalizeTargetLang`, the `newTokenProvider` rationale); could trim to the *why*. Running the repo's `/simplify` would catch them (inline note below).
- **Token fetch** — `tokenProvider` holds its lock across the HTTP token exchange; bounded (5s cap) and deliberate, so fine. `singleflight` would let waiters share one fetch without serializing during a refresh, if preferred.
- **`Timestamp`** — `int64` ms (mirrors the `AsyncJobResult` envelope); other client-facing time fields lean on RFC3339 `time.Time` — flagging for consistency, likely intentional.

## Transport shape + scope (your call)
Cross-checking against how this was originally envisioned — a request/reply RPC (text in → translated text out), no stored translations, with the history-load RPCs later carrying a `translate` param:
- Storage (stateless), the single-translate shape, and the real-time re-fetch path all match. ✅
- **Transport:** the design describes a request/reply RPC returning text synchronously; this is currently a fire-and-forget publish (`RegisterVoid`) + an async result on the response subject (inline note below). One practical consequence: under the concurrency cap a saturated handler drops the request silently, so the client only learns via its own timeout — a request/reply shape surfaces that immediately.
- **Scope:** the standalone RPC is here; the `loadHistory`/`loadSurrounding` translate-on-read param isn't. Follow-up, or in scope for this PR?

Both of those last two are design decisions — deferring to you.

translation-service/main.go
YI13: Good call — aligned to request/reply in `e41178a`. The handler now registers via `natsrouter.Register` and returns `(*TranslateResult, error)`, so:

- the client gets an immediate, correlated reply on `_INBOX` (no dependency on the `chat.user.{account}.>` subscription);
- validation errors (`empty_text` / `unsupported_lang`) come back synchronously as the standard errcode envelope instead of a `status:"error"` result;
- under `WithMaxConcurrency` saturation the router now replies `unavailable` ("service busy") — the silent-drop path you flagged is gone.

`TranslateRequest`/`TranslateResult` were slimmed to the sync shape (dropped `requestId`/`timestamp`/`status`/error-envelope fields; errors flow through the errcode envelope), and `client-api.md` §3.6 + the derived views + the design spec were updated to match.

---

translation-service/lang.go
YI13: Trimmed in `26ebc66` — `normalizeTargetLang` (and the `newTokenProvider` timeout-cap note) now keep just the non-obvious *why* (the ambiguous-`zh` rejection; the lock-starvation rationale), leaving the mechanics to the code.

---

translation-service/token.go
```diff
@@ -0,0 +1,105 @@
...
+func newTokenProvider(accessTokenURL, j1Token string, timeout, skew time.Duration) *tokenProvider {
+	// The token exchange is a quick auth call; cap its client timeout well below a
+	// translate call so a hung accessToken endpoint can't hold the single-flight lock
+	// (and thus every waiting translate) for the full translate timeout.
+	tokenTimeout := timeout
+	if tokenTimeout > 5*time.Second {
+		tokenTimeout = 5 * time.Second
+	}
+	return &tokenProvider{
+		client:  restyutil.New(accessTokenURL, restyutil.WithTimeout(tokenTimeout)),
```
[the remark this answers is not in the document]
YI13: Fixed in `9853ab4` — `newTranslator` now fails fast on a non-positive `TRANSLATION_HTTP_TIMEOUT` for the `stream` backend (before any resty client is built), with `main_test` coverage for `0s` and a negative value. The `mock` backend makes no HTTP calls, so it's unaffected.

---

docs/client-api.md
```diff
@@ -5563,6 +5565,77 @@ See [Error envelope](#6-error-envelope-reference).
...
 ---
 
+### 3.6 translation-service
+
+#### Translate Text
+
+**Subject:** `chat.user.{account}.request.translate.{siteID}`
+**Reply subject:** auto-generated `_INBOX.>` (NATS request/reply)
+
+Synchronous RPC. `translation-service` validates the request, calls the translation backend, and replies with a `TranslateResult` (the translated text) on success, or the standard [error envelope](#6-error-envelope-reference) on failure. Under handler saturation the router replies `unavailable` (`"service busy"`) so the client can retry.
```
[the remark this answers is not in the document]
YI13: Documented in `9853ab4` — `{siteID}` is the caller's own (local) site ID (translation is stateless and not federated, so no origin-site rule like `msg.send`). Added to `client-api.md` §3.6, the `request-reply.md` view, and the design spec.

---

translation-service/integration_test.go
```diff
@@ -0,0 +1,69 @@
...
+		require.NoError(t, err)
+
+		msg, err := nc.Request(context.Background(), reqSubject, body, 2*time.Second)
+		require.NoError(t, err)
+
+		var env errcode.Error
+		require.NoError(t, json.Unmarshal(msg.Data, &env))
+		assert.Equal(t, errcode.CodeBadRequest, env.Code)
+		assert.Equal(t, errcode.TranslateEmptyText, env.Reason)
+	})
```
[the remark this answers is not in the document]
YI13: Added in `9853ab4` — a `malformed payload replies with bad_request` subtest that sends invalid JSON via `nc.Request` and asserts the router's standardized `bad_request` envelope.

---

pkg/model/model_test.go
```diff
@@ -4931,3 +4931,19 @@ func TestMessage_TypeBSONOmitEmpty(t *testing.T) {
...
+	roundTrip(t, &r, &model.TranslateRequest{})
+}
+
+func TestTranslateResultJSON(t *testing.T) {
+	r := model.TranslateResult{
+		TranslatedText: "你好 世界",
+		TargetLang:     "zh-Hant-TW",
+	}
+	roundTrip(t, &r, &model.TranslateResult{})
+}
```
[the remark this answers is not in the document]
YI13: Renamed in `9853ab4` — `TestTranslateResult_JSON` (and `TestTranslateRequest_JSON` alongside it, for consistency) now follow the `Test<Type>_<Method>` convention.

---

pkg/subject/subject.go
```diff
@@ -1464,3 +1464,16 @@ func OrgSyncUsersUpsert(centralSiteID string) string {
 func EmployeesQuit(siteID string) string {
 	return fmt.Sprintf("chat.hr.%s.employees.quit", siteID)
 }
+
+// TranslateRequest is the subject a client sends a TranslateRequest to via NATS
+// request/reply. The service registers TranslateRequestPattern and replies with a
+// TranslateResult (or an errcode envelope) on the auto-generated _INBOX subject.
+func TranslateRequest(account, siteID string) string {
```
chenjr0719, line 1474: Minor consistency on the subject shape. The repo's top-level-resource RPCs all carry a trailing action segment after `{siteID}` — `search.{siteID}.messages`, `emoji.{siteID}.list`, `orgs.{siteID}.members`. This one is `…request.translate.{siteID}` with no action segment. Suggest **`chat.user.{account}.request.translate.{siteID}.text`** to match that family — and the `.text` segment leaves room for a future `.batch`.

(The `{account}` token is correct as-is: it's the NATS publish-permission scope — a client may only publish under its own `chat.user.{account}.>` — so the handler not reading it is expected, not dead code.)

chenjr0719 (changes requested):
Thanks for the rework — the sync request/reply conversion reads clean: `natsrouter.Register` returning the result, a reply on every error path (validation → errcode envelope, backend failure → generic `internal` with the raw cause logged rather than returned, saturation → `unavailable`), config that fails fast when the stream backend lacks its endpoint/token, and a thorough test matrix. One change to align the subject with the repo's conventions (inline).

**FE-integration:** the scope here is the on-demand translation RPC, which is exactly right — a real-time message event arrives, the client calls this RPC with the message text + the user's `translateMessageInto`, and gets the translated text back with nothing stored. No history-service change is needed for this scope. 👍

pkg/model/translation.go
```diff
@@ -0,0 +1,22 @@
...
+
+// TranslateRequest is the client→server payload for the synchronous translate
+// RPC on chat.user.{account}.request.translate.{siteID}. It is a NATS
+// request/reply call: the reply is a TranslateResult on success, or an errcode
+// error envelope on failure, correlated by NATS via the _INBOX reply subject.
+// TargetLang is a BCP-47 tag (the user's settings.translateMessageInto value);
+// translation-service normalizes it to the backend's language code
+// (zhTW/zhCN/en/de/ja).
+type TranslateRequest struct {
+	Text       string `json:"text"       bson:"text"`
```
chenjr0719: `TranslateRequest` / `TranslateResult` carry `bson` tags, but nothing here is persisted — the backend explicitly doesn't store translations, so these are wire-only request/reply structs. Consider dropping the `bson` tags (json-only) to signal they're never written to a store.

chenjr0719 (commented):
Two more minor notes to go with the earlier review:

**Plan doc** — `docs/superpowers/plans/2026-07-23-translation-api.md` has a top disclaimer that it reflects the *old async* design, but the body still fully describes `RegisterVoid` / `response.{RequestID}` / "no synchronous reply". Worth updating or trimming so a future reader isn't misled — the client-facing `client-api.md` section is correctly sync, only this design-history doc lags.

pkg/subject/subject.go
YI13: Done in `f334586` — the subject is now `chat.user.{account}.request.translate.{siteID}.text`, matching the `<resource>.{siteID}.<action>` family and leaving room for a future `.batch`. Updated the `pkg/subject` builders + test and every doc reference (`client-api.md` §3.6, the `request-reply.md` view, and the design spec). The handler registers via `TranslateRequestPattern` and the integration test uses the builders, so both realign automatically. And agreed on `{account}` — left as-is.

---

pkg/model/translation.go
YI13: Both minor notes done in `f334586`:

- **bson tags** — dropped from `TranslateRequest`/`TranslateResult`; they're json-only wire structs now, with a doc-comment note that the backend never persists them.
- **Plan doc** — you're right that the body still described the old async design under the top disclaimer. Rather than rewrite the whole task-by-task log, I trimmed it to a short "superseded → see the (sync) design spec + `client-api.md` §3.6" redirect, so a future reader can't be misled.

---

translation-service/translator_stream.go
```diff
@@ -0,0 +1,159 @@
...
+			if chunk.ReturnCode != successReturnCode {
+				// ReturnMessage is third-party-controlled — do not propagate it into
+				// caller-visible errors; keep only the trusted numeric returnCode.
+				return "", false, fmt.Errorf("translate backend rejected request (returnCode %d)", chunk.ReturnCode)
+			}
+			merged.WriteString(chunk.ReturnData.Translation) // verbatim, whitespace preserved
+		case trimmed != "":
+			nonSSE.WriteString(trimmed)
+		}
+		if readErr != nil {
```
mliu33, line 130: The readErr check should be placed before switch case ?

mliu33 (approved):
Excellent work, thanks! Just one minor suggestion

pkg/model/translation.go
```diff
@@ -0,0 +1,23 @@
...
+// RPC on chat.user.{account}.request.translate.{siteID}.text. It is a NATS
+// request/reply call: the reply is a TranslateResult on success, or an errcode
+// error envelope on failure, correlated by NATS via the _INBOX reply subject.
+// TargetLang is a BCP-47 tag (the user's settings.translateMessageInto value);
+// translation-service normalizes it to the backend's language code
+// (zhTW/zhCN/en/de/ja). These are wire-only structs (json tags only) — the
+// backend never persists translations.
+type TranslateRequest struct {
+	Text       string `json:"text"`
+	TargetLang string `json:"targetLang"`
```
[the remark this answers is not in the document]
YI13: Keeping these json-only intentionally. `TranslateRequest` / `TranslateResult` are wire-only request/reply payloads — `translation-service` is stateless and never persists a translation, so `bson` tags would be dead metadata. This was an explicit request from a human reviewer ([#discussion_r3681412613](https://github.com/hmchangw/newchat/pull/148#discussion_r3681412613)) to signal they're never written to a store; the "both json + bson tags" convention targets model structs that are actually persisted (rooms, subscriptions, messages, …). Leaving as-is.

---

translation-service/translator_stream.go
YI13: Good eye, but the order here is deliberate. `bufio.Reader.ReadString('\n')` returns the final line **together with** `io.EOF` when the stream's last line has no trailing newline — so on that read the returned `line` is still valid data. Checking `readErr` before the `switch` would drop that final line unprocessed, and it could be the `data: [DONE]` terminator or the last translation chunk (which would then trip the "stream ended without [DONE]" / lost-fragment paths). Processing the line first and handling `readErr` after is the standard `ReadString` idiom. Happy to add a short inline comment to make that intent explicit if it helps.

---

YI13: Added the clarifying comment in `434a2c6`.

---

chenjr0719 (approved):
The request/reply subject is now `...request.translate.{siteID}.text` as discussed, and it reads clean — the token-timeout cap, language normalization, and always-reply-on-error are all in place. Approving. Thanks for the iterations!

PR #155, opened 2026-07-30
Replace user ID lookups with account-based lookups in message-worker

general-lex:
## Summary
Refactors the message-worker to look up users by their account name instead of their user ID. This change aligns with a broader shift toward account-based user identification throughout the system and simplifies the lookup path by eliminating the need to extract user IDs from message metadata.

## Key Changes
- **User lookup method**: Changed from `FindUserByID(ctx, evt.Message.UserID)` to `FindUserByAccount(ctx, evt.Message.UserAccount)` in the main message processing flow
- **Thread reply handling**: Updated `lookupOwnerSiteID()` to accept and look up users by account instead of ID, affecting both first-reply and subsequent-reply paths
- **Logging**: Updated log fields and error messages to reference `account` instead of `user_id` for consistency
- **Test updates**: Updated all 60+ test cases in `handler_test.go` and related test files to mock `FindUserByAccount` calls with account names (e.g., `"alice"`, `"parent-user"`) instead of `FindUserByID` calls with user IDs (e.g., `"u-1"`, `"u-parent"`)

## Implementation Details
- The `cassParticipant` struct already carries both `ID` and `Account` fields, so the parent sender lookup in thread reply handling now uses the account field
- System message handling preserves the existing behavior of proceeding with a nil sender when user lookup fails
- Error messages now reference the account being looked up rather than the user ID, improving clarity in logs and error traces
- All cross-site inbox publish operations for thread replies now use the account-based lookup, maintaining consistency across the federation layer

https://claude.ai/code/session_01D9A8zRRFA4jXcDqkYmzdgV

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved message sender and thread ownership resolution by consistently using account identifiers.
  - Updated cross-site inbox routing for thread replies to resolve the correct parent owner site.
  - Improved warning and error details when sender information is missing or cannot be found.

- **Tests**
  - Updated message-processing and breadcrumb coverage to reflect account-based sender lookups.

### DIFF
--- a/message-worker/debug_log_test.go
+++ b/message-worker/debug_log_test.go
@@ -96,7 +96,7 @@ func TestHandler_processMessage_DebugBreadcrumbs(t *testing.T) {
 	store := NewMockStore(ctrl)
 	us := NewMockUserStore(ctrl)
 	ts := NewMockThreadStore(ctrl)
-	us.EXPECT().FindUserByID(gomock.Any(), "u-1").Return(user, nil).AnyTimes()
+	us.EXPECT().FindUserByAccount(gomock.Any(), "alice").Return(user, nil).AnyTimes()
 	store.EXPECT().SaveMessage(gomock.Any(), gomock.Any(), gomock.Any(), "site-a").Return(nil).AnyTimes()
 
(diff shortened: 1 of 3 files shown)


mliu33 (approved):
Thanks!

PR #157, opened 2026-07-30
Replace card registration with validation-only endpoint

vjauhari-work:
## Summary

Replaces the `/api/v1/cards/register` endpoint (which persisted cards to MongoDB) with a read-only `/api/v1/cards/validate` endpoint that checks card structure and version ordering against the in-memory cache snapshot, without writing anything to the database.

## Key Changes

- **Handler**: Renamed `HandleRegister` → `HandleValidate`; removed all persistence logic (store.InsertCard, store.ListVersions, store.GetCard calls). Version ordering is now judged against `cache.Versions(path)` instead of querying Mongo, and the endpoint returns `200 OK` instead of `201 Created`.

- **Cache API**: Replaced `Add(card)` method with `Versions(path)` method. The cache no longer supports incremental updates after load; it is a read-only snapshot refreshed only via `Load()`. This simplifies concurrency (no more CAS retry logic) and makes version ordering deterministic.

- **Store interface**: Removed `GetCard()`, `InsertCard()`, and `ListVersions()` methods. The store now only exposes `ListCards()` (used by cache refresh) and `EnsureIndexes()`.

- **Store implementation** (`store_mongo.go`): Deleted `GetCard()`, `InsertCard()`, `ListVersions()`, and the JSON-to-BSON conversion helpers (`jsonToBSON`, `normalizeNumbers`). Updated `docToCard()` to drop both `_id` and `migratedAt` (storage-only fields) from the template.

- **Validation**: Renamed `validateRegister()` → `validateCard()` and removed the `CardUsage` field from `cardDoc` (it is optional in the stored document but not required for validation).

- **Readiness check**: The endpoint now returns `503 Service Unavailable` if the cache has not yet loaded, since there is no snapshot to order against. This prevents false positives (accepting every version as "highest" before the first refresh).

- **Tests**: Rewrote `TestHandleRegister_*` tests as `TestHandleValidate_*` with new scenarios:
  - `TestHandleValidate_HappyPath`: Validates without touching the store.
  - `TestHandleValidate_DoesNotPersistOrCache`: Confirms the card is neither cached nor servable after validation.
  - `TestHandleValidate_VersionOrdering`: Comprehensive matrix of version comparisons against cached snapshots.
  - `TestHandleValidate_CacheNotLoaded`: Confirms 503 before first load.
  - `TestHandleValidate_RegisterRouteRemoved`: Confirms the old `/register` route is gone.
  - Updated integration test `TestRegisterEndToEnd` → `TestValidateEndToEnd` to validate against a pre-loaded cache instead of inserting.

- **Documentation** (`docs/client-api.md`): Updated endpoint description from "publishes new templates" to "validates candidate templates before publishing them"; clarified that `_id` and `migratedAt` are stripped from responses; fixed field name from `cardVersion` to `_tcardVersion` in examples.

- **Routes**: Removed the `/api/v1/cards/register` POST route; `/api/v1/cards/validate` is now the only write endpoint.

## Implementation Details

- Version ordering uses semantic version comparison (via existing `isHighest()` helper) against the cache snapshot, not a database query. Non-semver cached versions are ignored (e.g., `"v1"` does not block `"1.0.0"`).
- The cache snapshot is as fresh as the last `/refresh` call; cards written to Mongo out-of-band since then are not yet counted in ordering checks.
- Storage-only fields (`_id`, `migratedAt`) are dropped at the document-to-cache conversion layer (`docToCard`), not on the way out, so the cache itself never holds them.
- Only top-level keys are removed; nested `id` fields (e.g., inside `body` array elements) are preserved as template content.

https://claude.ai/code/session_014pauDAgCXN8TcqbUF4xZUm

## Summary by CodeRabbit

* **New Features**
  * Added advisory card validation at `POST /api/v1/cards/validate`.
  * Added directory listings for card paths without the template suffix.
  * Template responses now expose `_tcardVersion` and preserve nested content fields.

* **Changes**
  * Removed the card registration endpoint; validation no longer stores or publishes cards.
  * Validation now checks cached version ordering and reports clearer unavailable, invalid, and conflict responses.

* **Documentation**
  * Updated API documentation, response examples, routing behavior, and error details.


vjauhari-work:
Both nitpicks verified against `c62329a`. Declining both — reasoning below.

**`store_mongo.go:33-40` — exclusion projection.** An inclusion projection isn't expressible for this query. Card templates are schemaless: `docToCard` passes through every top-level key it doesn't specifically handle, so the template *is* the arbitrary remainder of the document. An allowlist would mean enumerating keys that are unknown by construction, and would drop the entire template body. The suggested fix — "specify only the required fields while preserving support for schemaless card templates" — can't satisfy both halves at once.

The exclusion list is the narrowest projection this query admits, and it's largely pre-existing: `{_id: 0}` predates this PR, which only adds `migratedAt: 0`. Keeping it, with the in-code comment the finding already acknowledges as reasonable.

**`cache.go:71-85` — `Versions` full-map scan.** Accurate as a description, but not a trade worth making here:

- Measured at ~14 ns/entry: 1.2 µs at 100 entries, and it takes roughly **70,000** cached `(path, version)` entries to reach 1 ms. A catalog is `{namespace}/{locale}/{card}` × a few versions each.
- It *replaced* a per-request Mongo `ListVersions` round-trip, so it is already 10–100× faster than the code it supplants — on an admin endpoint hit a handful of times a day.
- The sketch would trade a microsecond for a correctness regression. `entries` and `byPath` as two separate `atomic.Pointer` fields cannot be swapped together, so a reader landing between the two stores observes a torn snapshot — new entries with a stale index, or the reverse. The single-pointer swap is precisely what makes the current cache consistent under concurrent `Load`. A safe version needs both maps behind one struct pointer.

If the catalog ever does grow, `List` is the one to fix first: it runs on every public `GET /api/v1/cards/...` and layers a `sort.Slice` with a semver-parsing comparator on top of the same scan. At that point both move behind a single pointer together.

**Docstring coverage (66.67%).** Not acting on this one either. It's a pre-merge advisory rather than a repo gate, and it runs against this codebase's convention, which caps comments at 2 lines and actively removes WHAT-comments. The uncovered functions are mostly test functions and small helpers.

All five CI checks are green on `c62329a`: `lint`, `test`, `test-integration (tcard-service)`, `sast`, `prewarm`.

---

GITMateuszCharczuk (approved):
LGTM!

PR #166, opened 2026-08-01
Thread-unread tracking + per-recipient push badge counts

mliu33:
## Summary

Two arcs in one branch:

1. **Cleanup (2026-08-01 spec):** removed the then-vestigial `Subscription.ThreadUnread` (no producer, all consumers on the derived model), slimmed `ThreadReadEvent`, and simplified the read paths.
2. **Badge feature (2026-08-02 spec):** a new product requirement — every push notification must carry the recipient's badge unread count (rooms with unread main/thread messages, each room counts once, capped "9+") — justified reintroducing `ThreadUnread` **with a real producer** plus a badge-count pipeline. Specs: `docs/superpowers/specs/2026-08-01-threadunread-field-removal-design.md`, `docs/superpowers/specs/2026-08-02-thread-unread-badge-design.md`.

## Badge feature (Phase A — thread-unread tracking)

- **pkg/model / pkg/outbox**: `Subscription.ThreadUnread []string` (parent message IDs), `ThreadUnreadAddedEvent` (`thread_unread_added`, concurrent outbox lane), `ThreadReadEvent.NewThreadUnread`, badge RPC types, `PushNotificationEvent.UnreadCounts`.
- **message-worker (producer)**: per thread reply, one `UpdateMany $addToSet` over followers ∪ mentionees − sender at the room's home site + one federated event per remote follower home site (reply-unique dedup IDs; migration-tagged messages skipped).
- **inbox-worker**: applies `thread_unread_added` to home replicas; thread-read applies restored with the subscription mirror gated on the `$lt` guard.
- **room-service**: `message.thread.read` `$pull`/`$unset` + federated `newThreadUnread`; `thread.read.all` bulk clear — all **without alert coupling** (alert stays untouched branch-wide).
- **user-service**: `subscription.count (unread=true)` thread phase is now fully site-local (reads `threadUnread` off the already-fetched page) — the per-site `GetThreadRoomInfoBatch` fan-out is gone from the count path. Wire contract unchanged.

## Badge feature (Phase B — per-recipient push badge counts)

- **pkg/badgecache**: per-user Valkey SET of unread room IDs (atomic Lua: exists-gated `Bump`, `Seed`∪trigger, `Reseed`; counts capped at 10; fail-open everywhere; 24h TTL). The set-based design makes redeliveries/bursts idempotent and "each room counts once" structural.
- **user-service**: server-to-server `chat.server.request.user.{siteID}.badge.count.batch` — cache hit = `SADD`+`SCARD` (zero Mongo), miss = seed from `unreadRooms` ∪ the triggering room (so the notifying message is always counted, independent of async pipeline ordering); `subscription.count` reseeds the cache on every poll. Valkey optional (`envDefault:""` = noop cache).
- **room-service / inbox-worker**: badge clear hooks at every read-ingest point (home-local reads + federated `subscription_read`/`thread_read`/`thread_read_all`/`member_removed`), gated on live post-state, best-effort after the Mongo writes.
- **notification-worker**: resolves each survivor's **home site from the users collection at member-cache fill time** (`roomsubcache.Member.HomeSiteID`, cache schema v3), groups survivors per home site, one concurrent `badge.count.batch` RPC per site, stamps `UnreadCounts` per batch; per-site failures degrade to absent entries. Gated by `BADGE_COUNT_RPC_ENABLED` (default false).
- **docs**: `client-api.md` + derived views, `notification-worker-downstream-contracts.md` (payload, RPC contract, accuracy model), `CDC_COVERAGE.md`.

## Process & verification

- Executed as 11 subagent-driven tasks, each independently reviewed; per-task reviews caught and fixed a dedup-key collision that could drop badge re-notifies, an order-dependent integration subtest, missing o11y instrumentation, and a stale-event badge-clear race.
- A final whole-branch review (fresh reviewer over the full 17-commit diff) caught a cross-task semantic collision none of the scoped reviews could see — `Subscription.siteId` (room's home site) was being used as the recipient's home site for badge RPC routing — fixed by resolving home sites from the users collection at cache-fill time, with production-shaped tests (all subs on one site, user home sites differing).
- Gates on the final tree: `make lint` 0 issues; full `make test` green (race detector); `make sast` gosec/govulncheck/semgrep PASS; `make test-integration` green for message-worker, inbox-worker, room-service, user-service, pkg/badgecache, pkg/roomsubcache against real containers.

## Rollout

- Phase A is independently shippable (cheaper counts, no badge); Phase B is inert until `BADGE_COUNT_RPC_ENABLED=true` + Valkey env is set on user-service/room-service/inbox-worker.
- Mixed-version federation safe in both directions (wire-compat tests included). Deploy inbox-worker before message-worker for the new event type.

https://claude.ai/code/session_01VHCj1koQBpJGC6wh3vF6Sv


room-service/handler.go
```diff
@@ -1342,6 +1361,14 @@ func (h *Handler) messageRead(c *natsrouter.Context) (*model.StatusReply, error)
 		return nil, err
 	}
 
+	// Best-effort badge cache clear: only when the reader has no thread-unread
+	// left on this subscription (fetched before the read applied) and is
+	// home-local — a cross-site reader's home replica is cleared by
+	// inbox-worker once the federated subscription_read event lands.
+	if len(sub.ThreadUnread) == 0 && userSiteID == h.siteID && h.badge != nil {
```
mliu33: A normal message read (main room) will not update the sub.ThreadUnread filed. Why do we need to clear cache here ?

room-service/main.go
```diff
@@ -26,6 +29,11 @@ import (
 	"github.com/hmchangw/chat/pkg/subject"
 )
 
+// badgeCacheTTL bounds how long an account's badge unread-room set survives
+// without a Bump/Seed/Reseed refresh (pkg/badgecache.New's ttl param) — see
+// user-service's identical constant, the accelerator's other writer.
+const badgeCacheTTL = 24 * time.Hour
```
mliu33: Make it configurable from env variable

user-service/main.go
```diff
@@ -22,6 +26,10 @@ import (
 	"github.com/hmchangw/chat/user-service/service"
 )
 
+// badgeCacheTTL bounds how long an account's badge unread-room set survives
+// without a Bump/Seed/Reseed refresh (pkg/badgecache.New's ttl param).
+const badgeCacheTTL = 24 * time.Hour
```
mliu33: Make it configurable from env variable

user-service/service/badge.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/hmchangw/chat/pkg/errcode"
+	"github.com/hmchangw/chat/pkg/model"
+	"github.com/hmchangw/chat/pkg/natsrouter"
+	"github.com/hmchangw/chat/pkg/natsutil"
+)
+
+// badgeCountCap mirrors pkg/badgecache's own 10-cap (10 renders as "9+" in the
+// UI) so the cache-down fallback path (cappedUnion) agrees with the
+// cache-backed paths on what the badge count tops out at.
+const badgeCountCap = 10
```
mliu33: Make it configurable from env variable

user-service/service/subscriptions.go
```diff
@@ -548,45 +547,52 @@ func (s *UserService) GetByRoomID(c *natsrouter.Context, req models.GetByRoomIDR
...
 	crossBySite := map[string][]model.EnrichedSubscription{}
 	roomIDsBySite := map[string][]string{}
 	for i := range subs {
 		if subs[i].SiteID == s.siteID {
 			if unread(subs[i].LastSeenAt, timeutil.TimeToMillis(subs[i].LastMsgAt)) {
-				unreadTotal++
+				ids = append(ids, subs[i].RoomID)
 			} else {
-				pendingRooms[subs[i].RoomID] = s.siteID
+				pendingRooms[subs[i].RoomID] = subs[i].ThreadUnread
```
mliu33: I don't understand why are we storing ThreadUnread into pendingRooms ? If local subscription.ThreadUnread > 1, this room is considered unread.

user-service/service/subscriptions.go
```diff
@@ -600,169 +606,86 @@ func (s *UserService) countUnread(ctx context.Context, account string, total int
 		for site := range crossBySite {
 			sites = append(sites, site)
 		}
+		// roomCand is a read room's thread-bump candidacy: just enough to key pendingRooms
+		// without copying the whole (large) EnrichedSubscription through the results channel.
+		type roomCand struct {
+			roomID       string
+			threadUnread []string
```
mliu33: What is the purpose of ThreadUnread here ?

user-service/service/subscriptions.go
```diff
@@ -600,169 +606,86 @@ func (s *UserService) countUnread(ctx context.Context, account string, total int
...
+					if !ok {
 						continue
 					}
-					if unread(siteSubs[j].LastSeenAt, lastMsg[rid]) {
-						n++
+					if unread(siteSubs[j].LastSeenAt, ms) {
+						res.unreadIDs = append(res.unreadIDs, rid)
 					} else {
-						read = append(read, rid)
+						res.readCands = append(res.readCands, roomCand{roomID: rid, threadUnread: siteSubs[j].ThreadUnread})
```
mliu33: I don't understand what "readCands" is for as we only need to track total unread room list

user-service/service/subscriptions.go
```diff
@@ -600,169 +606,86 @@ func (s *UserService) countUnread(ctx context.Context, account string, total int
...
 			}()
 		}
 		wg.Wait()
 		for i := range results {
-			unreadTotal += results[i].unread
-			for _, rid := range results[i].readRooms {
-				pendingRooms[rid] = sites[i]
+			ids = append(ids, results[i].unreadIDs...)
+			for _, cand := range results[i].readCands {
+				pendingRooms[cand.roomID] = cand.threadUnread
```
mliu33: What is this for? If no used, we should remove

user-service/service/subscriptions.go
```diff
@@ -600,169 +606,86 @@ func (s *UserService) countUnread(ctx context.Context, account string, total int
...
-			if !ok || bumped[cand.roomID] {
-				continue
-			}
-			ms := info.LastMsgAt
-			if unread(cand.lastSeenAt, &ms) {
-				bumped[cand.roomID] = true
-			}
+	// Thread phase: a message-read room still counts once if it has >=1 unread followed
+	// thread. pendingRooms is keyed by roomID, so each room bumps at most once here.
+	for roomID, threadUnread := range pendingRooms {
```
mliu33: No need to calculate this because user unread subscription count request always goes to the user-service at same site as user origin. At user home site, the user subscription list at local mongoDB is always complete as room origin site services will always publish outbox event to replicate subscription ThreadUnread field to user origin site.

message-worker/handler.go
```diff
@@ -545,17 +582,80 @@ func (h *Handler) markThreadMentions(ctx context.Context, msg *model.Message, th
...
+	if err := h.threadStore.AddThreadUnread(ctx, roomID, parentMessageID, accounts); err != nil {
+		return fmt.Errorf("add thread unread: %w", err)
+	}
+
+	bySite := map[string][]string{}
+	for _, a := range accounts {
+		// A genuine lookup failure NAKs (mirrors every other lookupOwnerSiteID call
+		// site in this file); "not found" already degrades to ("", nil) inside the
+		// helper, so it — like a same-site home — simply needs no federation here.
+		site, err := h.lookupOwnerSiteID(ctx, a, "thread-unread recipient")
```
mliu33: It's better to make one db call to lookup siteIDs for all users once

room-service/handler.go
mliu33: The clear isn't about `ThreadUnread` — it's about the badge set itself. The Valkey set holds every room that is unread for *any* reason (main messages or threads). Reading the main room ends the room's message-level unread, so its entry has to be removed here or the badge count stays inflated until TTL/reseed. The `len(sub.ThreadUnread) == 0` guard is what protects the case you're describing: if the sub still carries unread threads, the room is still legitimately unread and the entry is left in place. Expanded the comment in 08cf88b to spell this out.

---

julianshen:
**30 findings · 22 unique issues · 21 inline comments posted** — 🔴 7 blocking · 🟠 23 warning

### 💬 Additional comments (1)


**`user-service/mongorepo/subscriptions.go:478`**

_🧹 read-consistency_ | _🟠 Warning_ | _`mongodb`_

**A secondary read can be certified as fresh in the badge cache.**

GetActiveSubscriptions runs through enrichedSecondary, while the deployment defaults to secondaryPreferred. Immediately after room-service or inbox-worker writes a read and clears the badge marker, user-service can read the pre-update subscription from a lagging secondary and Reseed it with a new freshness marker. Cache-first counts and badge pushes may then serve the resurrected unread room for the full cache TTL. Use a primary-backed aggregation for snapshots that create freshness markers, providing read-after-write visibility.

### Local review memory

- `docs/superpowers/plans/2026-08-01-threadunread-field-removal.md:394` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟡 Minor_ _⚡ Quick win_ **Fix the markdownlint violation.** Remove the space before the closing code-span delimiter. MD038 reports this line. &lt;details> &lt;summary>🧰 Tools&lt;/summary> &lt;details> <summa… ([thread](https://github.com/hmchangw/newchat/pull/166#discussion_r3695716983))
- `docs/superpowers/specs/2026-08-02-thread-unread-badge-design.md:95` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟠 Major_ _🏗️ Heavy lift_ **Do not treat add and pull as order-insensitive.** $addToSet is idempotent with other additions, but it is not commutative with the $pull used by ThreadReadEvent . If… ([thread](https://github.com/hmchangw/newchat/pull/166#discussion_r3698162851))
- `docs/superpowers/specs/2026-08-02-thread-unread-badge-design.md:208` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟠 Major_ _🏗️ Heavy lift_ **Narrow the global ±1 divergence guarantee.** The race matrix also permits stale entries from mute/leave changes, fail-open clears, and Valkey outages. These failures ca… ([thread](https://github.com/hmchangw/newchat/pull/166#discussion_r3698162858))
- `docs/superpowers/specs/2026-08-02-thread-unread-badge-design.md:214` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟠 Major_ _🏗️ Heavy lift_ **Fix the mixed-version rollout claim for NewThreadUnread .** Old inbox-workers ignore this field. If a new room-service producer sends a non-empty post- $pull snapshot t… ([thread](https://github.com/hmchangw/newchat/pull/166#discussion_r3698162861))
- `docs/superpowers/plans/2026-08-02-thread-unread-badge.md:125` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟡 Minor_ _⚡ Quick win_ **Add BSON tags to every new pkg/model field.** BadgeCountBatchRequest , BadgeCountBatchResponse , ThreadReadEvent.NewThreadUnread , and ThreadUnreadAddedEvent only def… ([thread](https://github.com/hmchangw/newchat/pull/166#discussion_r3698324280))
- `docs/superpowers/plans/2026-08-02-thread-unread-badge.md:333` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟠 Major_ _🏗️ Heavy lift_ **Prevent a read clear from deleting newer unread state.** Lines 303-314 can lose an unread marker. After $pull returns an empty array, a concurrent AddThreadUnread can… ([thread](https://github.com/hmchangw/newchat/pull/166#discussion_r3698324287))
- `message-worker/integration_test.go:811` — @coderabbitai: _🎯 Functional Correctness_ _🟡 Minor_ _⚡ Quick win_ **Test a non-recipient in the target room.** Insert a third subscription for carol in r1 . Do not include carol in accounts . Assert that carol.ThreadUnread remains nil. The… ([thread](https://github.com/hmchangw/newchat/pull/166#discussion_r3698324294))
- `pkg/model/event.go:242` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟡 Minor_ _⚡ Quick win_ **Add a BSON tag to ThreadUnreadAddedEvent.Timestamp .** ThreadUnreadAddedEvent has JSON tags only. The coding guidelines require every NATS event struct in pkg/model to… ([thread](https://github.com/hmchangw/newchat/pull/166#discussion_r3698324300))
- `pkg/model/model_test.go:5339` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟡 Minor_ _⚡ Quick win_ **Add round-trip coverage for BadgeCountBatchRequest and BadgeCountBatchResponse .** These tests cover Subscription.ThreadUnread , ThreadUnreadAddedEvent , and ThreadRe… ([thread](https://github.com/hmchangw/newchat/pull/166#discussion_r3698324313))

<details>
<summary>📒 Files reviewed (81)</summary>

* `data-migration/CDC_COVERAGE.md`
* `docs/client-api.md`
* `docs/superpowers/plans/2026-08-01-threadunread-field-removal.md`
* `docs/superpowers/plans/2026-08-02-thread-unread-badge.md`
* `docs/superpowers/plans/2026-08-10-badge-cache-invalidation.md`
* `docs/superpowers/specs/2026-08-01-threadunread-field-removal-design.md`
* `docs/superpowers/specs/2026-08-02-thread-unread-badge-design.md`
* `docs/superpowers/specs/2026-08-10-badge-cache-invalidation-design.md`
* `inbox-worker/consumer_config_test.go`
* `inbox-worker/deploy/docker-compose.yml`
* `inbox-worker/handler.go`
* `inbox-worker/main.go`
* `message-worker/handler.go`
* `message-worker/mock_store_test.go`
* `message-worker/store.go`
* `notification-worker/badge_client_test.go`
* `outbox-worker/consumer_config_test.go`
* `pkg/badgecache/badgecache_integration_test.go`
* `pkg/badgecache/badgecache_test.go`
* `pkg/emoji/standard_emoji_gen.go`
* `pkg/model/event.go`
* `pkg/model/model_test.go`
* `pkg/model/push.go`
* `pkg/model/room.go`
* `pkg/model/subscription.go`
* `pkg/outbox/outbox.go`
* `pkg/outbox/outbox_test.go`
* `pkg/roomsubcache/metrics_test.go`
* `pkg/roomsubcache/roomsubcache_test.go`
* `pkg/subject/subject.go`
* `portal-service/mock_store_test.go`
* `room-service/deploy/docker-compose.yml`
* `room-service/handler.go`
* `room-service/main.go`
* `room-service/mock_store_test.go`
* `room-service/store.go`
* `user-service/main.go`
* `user-service/mongorepo/threadsubscriptions.go`
* `user-service/mongorepo/threadsubscriptions_test.go`
* `user-service/service/badge.go`
* `user-service/service/badge_test.go`
* `user-service/service/me_test.go`
* `user-service/service/mocks/mock_repository.go`
* `user-service/service/sso_test.go`
* `user-service/service/status_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

pkg/outbox/outbox.go
```diff
@@ -32,6 +32,7 @@ var ConcurrentEventTypes = []model.InboxEventType{
 	// inbox-worker's UpsertThreadSubscription is $setOnInsert (immutable identity)
 	// + $max hasMention (monotonic), so out-of-order/duplicate applies converge.
 	model.InboxThreadSubscriptionUpserted,
+	model.InboxThreadUnreadAdded,
```
julianshen, line 37: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**The concurrent event is not order-independent.**

`thread_unread_added` uses `$addToSet`, but `ApplyThreadRead` replaces the entire array with the source's `NewThreadUnread` snapshot. These operations do not commute. For example, if an add for `p3` reaches the home replica before an older thread-read snapshot `[p2]`, the subsequent `$set` permanently loses `p3`. Reads of different threads can similarly resurrect an already-cleared ID because their guards are scoped to different ThreadSubscriptions. Restore an operation-based read event carrying the parent ID and apply `$pull` without overwriting unrelated IDs, or provide an ordering/versioning mechanism covering both event types. Add an out-of-order add/read regression test.

<details>

```
In `pkg/outbox/outbox.go` around line 35: `thread_unread_added` uses `$addToSet`, but `ApplyThreadRead` replaces the entire array with the source's `NewThreadUnread` snapshot. These operations do not commute. For example, if an add for `p3` reaches the home replica before an older thread-read snapshot `[p2]`, the subsequent `$set` permanently loses `p3`. Reads of different threads can similarly resurrect an already-cleared ID because their guards are scoped to different ThreadSubscriptions. Restore an operation-based read event carrying the parent ID and apply `$pull` without overwriting unrelated IDs, or provide an ordering/versioning mechanism covering both event types. Add an out-of-order add/read regression test.

```

</details>

pkg/badgecache/badgecache.go
```diff
@@ -0,0 +1,275 @@
...
+// KEYS[1]=set key, KEYS[2]=marker key. ARGV[1]=roomID, ARGV[2]=ttlSeconds.
+// Returns -1 on miss, else SCARD after SADD+EXPIRE.
+var bumpScript = redis.NewScript(`
+if redis.call('EXISTS', KEYS[1]) == 0 and redis.call('EXISTS', KEYS[2]) == 0 then
+  return -1
+end
+redis.call('SADD', KEYS[1], ARGV[1])
+redis.call('EXPIRE', KEYS[1], ARGV[2])
+redis.call('SET', KEYS[2], '1', 'EX', ARGV[2])
+return redis.call('SCARD', KEYS[1])
```
julianshen, line 30: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**A legacy set is promoted to fresh without recomputation.**

The bump script treats a set without a freshness marker as a hit and immediately writes the marker. That contradicts `Count`'s stale-without-marker semantics and can bless a stale pre-upgrade set, allowing cache-first counts to serve it for another TTL. Simply making bump require the marker is insufficient because `Seed` currently merges into the legacy set. Make marker absence a bump miss, and have the first seed atomically discard any markerless set before adding the recomputed rooms; later concurrent seeds with a marker may continue merging trigger rooms.

<details>

```
In `pkg/badgecache/badgecache.go` around line 37: The bump script treats a set without a freshness marker as a hit and immediately writes the marker. That contradicts `Count`'s stale-without-marker semantics and can bless a stale pre-upgrade set, allowing cache-first counts to serve it for another TTL. Simply making bump require the marker is insufficient because `Seed` currently merges into the legacy set. Make marker absence a bump miss, and have the first seed atomically discard any markerless set before adding the recomputed rooms; later concurrent seeds with a marker may continue merging trigger rooms.

```

</details>

inbox-worker/main.go
```diff
@@ -538,33 +559,38 @@ func (s *mongoInboxStore) ApplyThreadRead(ctx context.Context, roomID, threadRoo
...
 	return nil
 }
 
+// AddThreadUnread marks parentMessageID unread for accounts' subscriptions in
+// roomID via a single $addToSet UpdateMany. Idempotent under JetStream
+// redelivery; accounts not subscribed simply match nothing.
+func (s *mongoInboxStore) AddThreadUnread(ctx context.Context, roomID, parentMessageID string, accounts []string) error {
+	if len(accounts) == 0 {
+		return nil
+	}
```
julianshen, line 624: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**A missing home-replica subscription silently loses the unread event.**

`AddThreadUnread` ignores `MatchedCount`, so the event is acknowledged even when `member_added` has not yet created one or more destination subscriptions. Because `thread_unread_added` rides the concurrent lane, it can overtake membership replication and the unread mark is then lost permanently. Preserve ordering behind membership events or detect unmatched distinct accounts and return a retryable error, with coverage for delivery before `member_added`.

<details>

```
In `inbox-worker/main.go` around line 584: `AddThreadUnread` ignores `MatchedCount`, so the event is acknowledged even when `member_added` has not yet created one or more destination subscriptions. Because `thread_unread_added` rides the concurrent lane, it can overtake membership replication and the unread mark is then lost permanently. Preserve ordering behind membership events or detect unmatched distinct accounts and return a retryable error, with coverage for delivery before `member_added`.

```

</details>

notification-worker/main.go
```diff
@@ -114,9 +129,51 @@ func (m *mongoMemberLoader) Load(ctx context.Context, roomID string) ([]roomsubc
 	if err := cur.Err(); err != nil {
 		return nil, fmt.Errorf("iterate subscriptions: %w", err)
 	}
+	if err := m.fillHomeSites(ctx, roomID, out); err != nil {
+		return nil, err
+	}
```
julianshen, line 132: _🏷️ reliability_ | _🟠 Warning_ | _`tgd-review`_

**Badge-only enrichment failures abort the push path.**

Home-site enrichment exists only to route optional badge RPCs, but any users-collection error currently fails the entire member load. This prevents or retries the push even though the subscription members were already loaded, contradicting the documented fail-open badge behavior. Log the enrichment failure and continue with empty home-site values so pushes still ship without counts.

<details>

> ‼️ **IMPORTANT**

```suggestion
if err := m.fillHomeSites(ctx, roomID, out); err != nil {
		slog.WarnContext(ctx, "member home-site enrichment failed; continuing without badge counts", "room_id", roomID, "error", err)
	}
```

</details>

<details>

```
In `notification-worker/main.go` around line 132: Home-site enrichment exists only to route optional badge RPCs, but any users-collection error currently fails the entire member load. This prevents or retries the push even though the subscription members were already loaded, contradicting the documented fail-open badge behavior. Log the enrichment failure and continue with empty home-site values so pushes still ship without counts.

```

</details>

chat-frontend/src/api/types.ts
```diff
@@ -68,7 +68,6 @@ export interface Subscription {
   joinedAt: string
   lastSeenAt?: string
   hasMention: boolean
```
julianshen, line 70: _🏷️ intent-match_ | _🟠 Warning_ | _`tgd-review`_

**The frontend type drops a field reintroduced by this PR.**

The final model, user-service response, and client API documentation all expose optional `threadUnread`, but this diff removes it from the frontend `Subscription` interface. Restore `threadUnread?: string[]` so the declared API type matches the response schema.

<details>

```
In `chat-frontend/src/api/types.ts` around line 70: The final model, user-service response, and client API documentation all expose optional `threadUnread`, but this diff removes it from the frontend `Subscription` interface. Restore `threadUnread?: string[]` so the declared API type matches the response schema.

```

</details>

inbox-worker/main.go
```diff
@@ -540,76 +550,87 @@ func (s *mongoInboxStore) ApplySubscriptionRestriction(ctx context.Context, room
...
+func (s *mongoInboxStore) ApplyThreadRead(ctx context.Context, roomID, threadRoomID, account, parentMessageID string, lastSeenAt time.Time) error {
+	filter := threadReadGuard(bson.M{"threadRoomId": threadRoomID, "userAccount": account}, lastSeenAt)
+	tsRes, err := s.threadSubCol.UpdateOne(ctx, filter, threadReadUpdate(lastSeenAt))
 	if err != nil {
-		return fmt.Errorf("apply thread read on thread subscription for %q in thread room %q: %w",
-			account, threadRoomID, err)
+		return fmt.Errorf("apply thread read for %q in thread room %q: %w", account, threadRoomID, err)
 	}
-	if tsRes.MatchedCount == 0 {
+	// Guard rejected (stale/duplicate) or legacy event without a parent ID:
```
julianshen, line 584: _🏷️ retries_ | _🔴 Blocking_ | _`distributed-system`_

**A partial thread-read apply cannot recover on redelivery.**

The guarded ThreadSubscription update and Subscription `$pull` are separate writes. Timeline: the first update succeeds, the `$pull` fails before executing, and the event is NAKed. On redelivery, `lastSeenAt` already equals the event timestamp, so `MatchedCount == 0` returns success without retrying the pull. The parent ID then remains unread permanently. Make both writes transactional, or distinguish an equal timestamp (retry of the same event) from a greater timestamp (stale event) and retry the pull for equality.

<details>

```
In `inbox-worker/main.go` around line 584: The guarded ThreadSubscription update and Subscription `$pull` are separate writes. Timeline: the first update succeeds, the `$pull` fails before executing, and the event is NAKed. On redelivery, `lastSeenAt` already equals the event timestamp, so `MatchedCount == 0` returns success without retrying the pull. The parent ID then remains unread permanently. Make both writes transactional, or distinguish an equal timestamp (retry of the same event) from a greater timestamp (stale event) and retry the pull for equality.

```

</details>

pkg/model/event.go
```diff
@@ -197,31 +198,48 @@ type SubscriptionReadEvent struct {
...
 type ThreadReadEvent struct {
-	Account         string   `json:"account"`
-	RoomID          string   `json:"roomId"`
-	ThreadRoomID    string   `json:"threadRoomId"`
-	ParentMessageID string   `json:"parentMessageId"`
-	NewThreadUnread []string `json:"newThreadUnread"`
-	Alert           bool     `json:"alert"`
-	LastSeenAt      int64    `json:"lastSeenAt"`
-	Timestamp       int64    `json:"timestamp"`
+	Account      string `json:"account"`
```
julianshen, line 213: _🏷️ compatibility_ | _🔴 Blocking_ | _`distributed-system`, `mongodb` (2 rules)_

**New thread-read events can erase data on old inbox workers.**

The new producer omits NewThreadUnread, but the pre-change inbox-worker decodes an absent field as nil and responds by $unsetting the entire subscription.threadUnread array. During a rolling deployment, a new room-service talking to an old inbox-worker can therefore erase unread IDs for unrelated threads. Require inbox-worker deployment before room-service emission, or retain a backward-compatible/versioned payload until old consumers are gone.

<details>

- **`distributed-system`**:213: During a rolling deploy, a new room-service can omit `newThreadUnread` while an old inbox-worker still decodes that field as nil and interprets it as `$unset` of the entire array. Reading p1 can therefore erase an unrelated unread p2 on the home replica. Deploy inbox-worker before the producer and document/enforce that order, or retain a transition field understood by old workers until they are gone.

</details>

<details>

```
In `pkg/model/event.go` around line 213: The new producer omits NewThreadUnread, but the pre-change inbox-worker decodes an absent field as nil and responds by $unsetting the entire subscription.threadUnread array. During a rolling deployment, a new room-service talking to an old inbox-worker can therefore erase unread IDs for unrelated threads. Require inbox-worker deployment before room-service emission, or retain a backward-compatible/versioned payload until old consumers are gone.

```

</details>

pkg/badgecache/badgecache.go
```diff
@@ -0,0 +1,246 @@
...
+// cap as "9+"). Override via New's maxCount (BADGE_COUNT_CAP).
+const DefaultMaxCount = 10
+
+// bumpScript SADDs one room and refreshes both TTLs. Miss (-1, no writes)
+// only when neither set nor marker exists; a marker-only fresh all-read state
+// is a hit that creates the set. KEYS=[set, marker], ARGV=[roomID, ttlSec];
+// returns SCARD on hit.
+var bumpScript = redis.NewScript(`
+if redis.call('EXISTS', KEYS[1]) == 0 and redis.call('EXISTS', KEYS[2]) == 0 then
+  return -1
```
julianshen, line 25: _🎯 correctness_ | _🔴 Blocking_ | _`distributed-system`, `silent-failure`, `tgd-review` (3 rules)_

**A legacy set is promoted to fresh without being recomputed.**

The bump script treats an existing set without a freshness marker as a hit and writes the marker. Pre-marker sets may be stale, so this can make cache-first counts trust stale contents for the full TTL. A missing marker must invalidate any legacy set and force the caller through the Mongo seed path.

<details>

> ‼️ **IMPORTANT**

```suggestion
if redis.call('EXISTS', KEYS[2]) == 0 then
  redis.call('DEL', KEYS[1])
  return -1
end
```

</details>

<details>

- **`silent-failure`**:25: The bump script reports a hit when either the set or marker exists. During rollout, an old set without a freshness marker can therefore receive one after only adding the triggering room, even though its other contents were never recomputed. Once BADGE_COUNT_CACHE_FIRST is enabled, Count treats that stale set as authoritative until invalidation or TTL. Treat every missing marker as a cache miss, and make the subsequent seed atomically discard a markerless legacy set before writing the computed room IDs and marker. Add coverage for a stale set without a marker.
- **`distributed-system`**:24: The freshness marker is the validity contract, but the bump script treats an unmarked legacy set as a cache hit and writes a marker. During rollout, a stale pre-marker set can receive one notification, become marked fresh, and later be served by cache-first count for another full TTL without a Mongo reconciliation. A bump should miss whenever the marker is absent; marker-only fresh-zero state still remains a hit.

<details>

```text
if redis.call('EXISTS', KEYS[2]) == 0 then
```

</details>


</details>

<details>

```
In `pkg/badgecache/badgecache.go` around line 23: The bump script treats an existing set without a freshness marker as a hit and writes the marker. Pre-marker sets may be stale, so this can make cache-first counts trust stale contents for the full TTL. A missing marker must invalidate any legacy set and force the caller through the Mongo seed path.

```

</details>

inbox-worker/main.go
```diff
@@ -540,76 +550,87 @@ func (s *mongoInboxStore) ApplySubscriptionRestriction(ctx context.Context, room
...
-	tsRes, err := s.threadSubCol.UpdateOne(ctx, tsFilter, tsUpdate)
+}
+
+// ApplyThreadRead advances the home-replica ThreadSubscription under the $lt
+// guard, then — only when that guarded update matched — $pulls parentMessageID
+// from the Subscription's threadUnread. The MatchedCount gate keeps a
+// stale/duplicate event from touching the Subscription.
+func (s *mongoInboxStore) ApplyThreadRead(ctx context.Context, roomID, threadRoomID, account, parentMessageID string, lastSeenAt time.Time) error {
+	filter := threadReadGuard(bson.M{"threadRoomId": threadRoomID, "userAccount": account}, lastSeenAt)
+	tsRes, err := s.threadSubCol.UpdateOne(ctx, filter, threadReadUpdate(lastSeenAt))
```
julianshen, line 580: _🏷️ atomicity_ | _🔴 Blocking_ | _`mongodb`, `performance-and-concurrency` (2 rules)_

**A retry can permanently skip the subscription pull.**

ApplyThreadRead first advances thread_subscriptions and only then updates subscriptions. If the first update succeeds but the subsequent $pull fails, redelivery uses the same lastSeenAt; the strict $lt guard then matches zero thread-subscription documents and returns before retrying the $pull. The stale threadUnread ID is therefore acknowledged permanently. Put the guarded thread-subscription update and subscription $pull in one MongoDB transaction, propagate the session to both operations, and retry transient transaction/unknown-commit errors.

<details>

- **`performance-and-concurrency`**:560: `ApplyThreadRead` first advances the ThreadSubscription and then performs the Subscription `$pull`. If the first write succeeds but the second write fails or the worker crashes between them, redelivery uses the same `lastSeenAt`; the strict `$lt` guard rejects it, so the function returns before retrying the pull. The unread ID then remains indefinitely. Allow an equal timestamp to match so an idempotent redelivery can complete the second write; genuinely older events remain rejected.

</details>

<details>

```
In `inbox-worker/main.go` around line 580: ApplyThreadRead first advances thread_subscriptions and only then updates subscriptions. If the first update succeeds but the subsequent $pull fails, redelivery uses the same lastSeenAt; the strict $lt guard then matches zero thread-subscription documents and returns before retrying the $pull. The stale threadUnread ID is therefore acknowledged permanently. Put the guarded thread-subscription update and subscription $pull in one MongoDB transaction, propagate the session to both operations, and retry transient transaction/unknown-commit errors.

```

</details>

pkg/outbox/outbox.go
```diff
@@ -32,6 +32,9 @@ var ConcurrentEventTypes = []model.InboxEventType{
 	// inbox-worker's UpsertThreadSubscription is $setOnInsert (immutable identity)
 	// + $max hasMention (monotonic), so out-of-order/duplicate applies converge.
 	model.InboxThreadSubscriptionUpserted,
+	// thread_unread_added $addToSets one parent ID and thread_read $pulls one —
```
julianshen, line 35: _🏷️ consistency_ | _🔴 Blocking_ | _`mongodb`_

**A delayed add can resurrect a thread already read.**

The operation-based $pull fixes loss of unrelated IDs, but $addToSet and $pull still do not commute for the same parent ID. Because thread_unread_added uses the concurrent lane while thread_read uses a different lane, an add published before a read can be applied afterward and re-add the parent ID indefinitely. Move thread_unread_added onto the same ordered delivery path as thread_read, or add a per-thread activity/read high-water mark that rejects additions older than the read.

<details>

```
In `pkg/outbox/outbox.go` around line 35: The operation-based $pull fixes loss of unrelated IDs, but $addToSet and $pull still do not commute for the same parent ID. Because thread_unread_added uses the concurrent lane while thread_read uses a different lane, an add published before a read can be applied afterward and re-add the parent ID indefinitely. Move thread_unread_added onto the same ordered delivery path as thread_read, or add a per-thread activity/read high-water mark that rejects additions older than the read.

```

</details>

inbox-worker/main.go
```diff
@@ -540,76 +550,87 @@ func (s *mongoInboxStore) ApplySubscriptionRestriction(ctx context.Context, room
...
 	if err != nil {
-		return fmt.Errorf("apply thread read on thread subscription for %q in thread room %q: %w",
-			account, threadRoomID, err)
+		return fmt.Errorf("apply thread read for %q in thread room %q: %w", account, threadRoomID, err)
 	}
-	if tsRes.MatchedCount == 0 {
+	// Guard rejected (stale/duplicate) or legacy event without a parent ID:
+	// leave threadUnread alone.
+	if tsRes.MatchedCount == 0 || parentMessageID == "" {
```
julianshen, line 586: _🏷️ nats-redelivery_ | _🔴 Blocking_ | _`nats`_

**A retry can acknowledge a thread read without applying its pull.**

ApplyThreadRead commits the guarded ThreadSubscription update before applying the Subscription $pull. If the first write succeeds but the second write returns a transient error, the handler NAKs the event. On redelivery, lastSeenAt already equals the event timestamp, so the $lt guard yields MatchedCount == 0 and this branch returns success without retrying the pull. JetStream then ACKs the event while threadUnread remains stale. Make both writes transactional, or persist enough event/revision state to distinguish a completed application from this partial application and safely resume the pull. Add a test where the subscription write fails once and the same event is redelivered.

<details>

```
In `inbox-worker/main.go` around line 586: ApplyThreadRead commits the guarded ThreadSubscription update before applying the Subscription $pull. If the first write succeeds but the second write returns a transient error, the handler NAKs the event. On redelivery, lastSeenAt already equals the event timestamp, so the $lt guard yields MatchedCount == 0 and this branch returns success without retrying the pull. JetStream then ACKs the event while threadUnread remains stale. Make both writes transactional, or persist enough event/revision state to distinguish a completed application from this partial application and safely resume the pull. Add a test where the subscription write fails once and the same event is redelivered.

```

</details>

message-worker/handler.go
```diff
@@ -170,9 +170,21 @@ func (h *Handler) processMessage(ctx context.Context, data []byte, isMigration b
...
-		if err := h.markThreadMentions(ctx, &evt.Message, threadRoomID, evt.SiteID, isMigration); err != nil {
+		mentionedAccounts, err := h.markThreadMentions(ctx, &evt.Message, threadRoomID, evt.SiteID, isMigration)
+		if err != nil {
 			return fmt.Errorf("mark thread mentions: %w", err)
 		}
+		// Suppressed for migrated replies like the thread-sub writes above: the
+		// source already carries threadUnread state, and re-deriving it here
+		// would re-mark threads the migration already resolved.
+		if !isMigration {
+			recipients := make([]string, 0, len(followers)+len(mentionedAccounts))
```
julianshen, line 181: _🎯 cassandra-correctness_ | _🟠 Warning_ | _`cassandra`, `silent-failure` (2 rules)_

**Unread state is published before Cassandra accepts the reply.**

For a thread reply, `fanOutThreadUnread` updates local subscriptions and publishes remote `thread_unread_added` events before `SaveThreadMessage` persists the reply to Cassandra. If the Cassandra write then fails, recipients can receive unread state for a message that is not readable. On redelivery, a local recipient who cleared the thread after the first attempt can also have the same parent ID re-added before Cassandra succeeds. Move this fan-out after a successful `SaveThreadMessage`; retries can then rely on the existing idempotent message-save/redelivery path before replaying the `$addToSet`/deduplicated outbox work.

<details>

- **`silent-failure`**:182: fanOutThreadUnread updates local subscriptions and durably publishes remote outbox events before SaveThreadMessage runs. If the Cassandra write then fails, the handler returns an error but the unread state remains locally and may already be replicated remotely. A permanent failure or exhausted retry therefore leaves users with an unread marker for a reply that was never persisted. Move the unread fan-out after SaveThreadMessage succeeds; redelivery remains safe because the save and unread operations are idempotent.

</details>

<details>

```
In `message-worker/handler.go` around line 181: For a thread reply, `fanOutThreadUnread` updates local subscriptions and publishes remote `thread_unread_added` events before `SaveThreadMessage` persists the reply to Cassandra. If the Cassandra write then fails, recipients can receive unread state for a message that is not readable. On redelivery, a local recipient who cleared the thread after the first attempt can also have the same parent ID re-added before Cassandra succeeds. Move this fan-out after a successful `SaveThreadMessage`; retries can then rely on the existing idempotent message-save/redelivery path before replaying the `$addToSet`/deduplicated outbox work.

```

</details>

user-service/service/subscriptions.go
```diff
@@ -548,45 +547,49 @@ func (s *UserService) GetByRoomID(c *natsrouter.Context, req models.GetByRoomIDR
...
-// came out read feed the thread phase (countThreadOnlyUnread); everything degrades
-// best-effort — an unreachable site is skipped rather than nuking the count to total.
-func (s *UserService) countUnread(ctx context.Context, account string, total int) (*models.CountResponse, error) {
-	// Short-circuit zero: min(0, maxSubs)=0 would build a $limit:0 MongoDB rejects.
-	if total == 0 {
-		return &models.CountResponse{Count: 0}, nil
-	}
-	// Cap at maxSubs — query-side total can exceed the cap; min keeps the fetch bounded and consistent with the list endpoints.
-	subs, err := s.subs.GetActiveSubscriptions(ctx, account, min(total, s.maxSubs))
+// unreadRooms returns the account's active room IDs with unread activity: a
```
julianshen, line 574: _🏷️ consistency_ | _🟠 Warning_ | _`distributed-system`_

**An in-flight reseed can overwrite a newer invalidation.**

Timeline: unreadRooms reads an old set of room IDs, a concurrent read updates Mongo and calls ClearAll, then this request executes Reseed with the old snapshot and writes a fresh marker. Subsequent cache-first counts serve the resurrected rooms until another invalidation or TTL. Seed has the same race in BadgeCountBatch. Add a per-account generation/fencing token that ClearAll advances and Seed/Reseed compare before publishing a snapshot.

<details>

```
In `user-service/service/subscriptions.go` around line 574: Timeline: unreadRooms reads an old set of room IDs, a concurrent read updates Mongo and calls ClearAll, then this request executes Reseed with the old snapshot and writes a fresh marker. Subsequent cache-first counts serve the resurrected rooms until another invalidation or TTL. Seed has the same race in BadgeCountBatch. Add a per-account generation/fencing token that ClearAll advances and Seed/Reseed compare before publishing a snapshot.

```

</details>

notification-worker/main.go
```diff
@@ -119,9 +127,51 @@ func (m *mongoMemberLoader) Load(ctx context.Context, roomID string) ([]roomsubc
 	if err := cur.Err(); err != nil {
 		return nil, fmt.Errorf("iterate subscriptions: %w", err)
 	}
+	if err := m.fillHomeSites(ctx, roomID, out); err != nil {
```
julianshen, line 130: _🏷️ operational_ | _🟠 Warning_ | _`distributed-system`, `mongodb`, `nats`, `tgd-review` (4 rules)_

**Optional home-site lookup failures abort push delivery.**

fillHomeSites queries users only to route optional badge RPCs, yet any MongoDB error is returned from the member loader and aborts or retries the entire notification. This contradicts the badge phase's fail-open contract even though subscription recipients were already loaded. Log the enrichment error and continue with empty HomeSiteID values so pushes still ship without unreadCounts.

<details>

- **`distributed-system`**:130: The subscription member list is already available when `fillHomeSites` queries the users collection. If that badge-only query fails, Load returns an error, so the notification is retried or dropped rather than emitted without badge counts. Log the enrichment failure and continue with empty HomeSiteID values; those accounts will then follow the existing no-count degradation path.
- **`nats`**:130: Home-site enrichment is used only to route optional badge.count.batch requests, yet any users-collection failure aborts the completed member load. HandleMessage consequently fails and the JetStream message is retried without emitting the push, contrary to the badge phase's fail-open contract. Continue with empty HomeSiteID values so the push ships without badge counts.
- **`tgd-review`**:130: Home-site enrichment is needed only for optional badge RPC routing, yet a users-collection failure aborts the member load and therefore the push. This also happens when badge RPCs are disabled. Continue with empty `HomeSiteID` values so notification delivery remains fail-open.

<details>

```text
if err := m.fillHomeSites(ctx, roomID, out); err != nil {
		slog.WarnContext(ctx, "member home-site enrichment failed; continuing without badge counts", "room_id", roomID, "error", err)
	}
```

</details>


</details>

<details>

```
In `notification-worker/main.go` around line 130: fillHomeSites queries users only to route optional badge RPCs, yet any MongoDB error is returned from the member loader and aborts or retries the entire notification. This contradicts the badge phase's fail-open contract even though subscription recipients were already loaded. Log the enrichment error and continue with empty HomeSiteID values so pushes still ship without unreadCounts.

```

</details>

user-service/service/badge.go
```diff
@@ -0,0 +1,61 @@
...
+// NATS: chat.server.request.user.{siteID}.badge.count.batch
+func (s *UserService) BadgeCountBatch(c *natsrouter.Context, req model.BadgeCountBatchRequest) (*model.BadgeCountBatchResponse, error) {
+	if len(req.Accounts) == 0 || req.RoomID == "" {
+		return nil, errcode.BadRequest("roomId and accounts are required")
+	}
+	resp := &model.BadgeCountBatchResponse{Counts: make(map[string]int, len(req.Accounts))}
+	hits := s.badge.BumpBatch(c, req.Accounts, req.RoomID)
+	for _, account := range req.Accounts {
+		if n, ok := hits[account]; ok {
+			resp.Counts[account] = n
```
julianshen, line 26: _🏷️ backpressure_ | _🟠 Warning_ | _`distributed-system`_

**Cold badge accounts are seeded serially under one RPC deadline.**

After the batched cache lookup, every miss runs unreadRooms sequentially. Two cold accounts whose cross-site metadata calls each take several seconds can exceed notification-worker's single 5-second request timeout; the caller then discards even counts completed earlier, and the push waits for the timeout. Process misses with bounded concurrency and per-account budgets, or chunk requests so one slow account cannot delay or discard the entire site's result.

<details>

```
In `user-service/service/badge.go` around line 26: After the batched cache lookup, every miss runs unreadRooms sequentially. Two cold accounts whose cross-site metadata calls each take several seconds can exceed notification-worker's single 5-second request timeout; the caller then discards even counts completed earlier, and the push waits for the timeout. Process misses with bounded concurrency and per-account budgets, or chunk requests so one slow account cannot delay or discard the entire site's result.

```

</details>

room-service/main.go
```diff
@@ -254,6 +264,34 @@ func main() {
...
+		// pkg/valkeyutil's instrumentCluster.
+		if _, err := o11yredis.Wrap(valkeyClient, sdk.TracerProvider(), sdk.MeterProvider()); err != nil {
+			slog.Error("instrument valkey client failed", "error", err)
+			os.Exit(1)
+		}
+		pingCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+		err := valkeyClient.Ping(pingCtx).Err()
+		cancel()
+		if err != nil {
+			slog.Error("valkey connect failed", "error", err)
```
julianshen, line 286: _🏷️ availability_ | _🟠 Warning_ | _`distributed-system`_

**An optional badge-cache outage prevents core services from starting.**

When VALKEY_ADDRS is configured—which the compose file does by default—a failed startup Ping calls os.Exit. Thus a Valkey outage during restart takes room-service offline even though badge operations are documented as fail-open and the handler supports a nil cache. The same startup pattern exists in inbox-worker and user-service. On connection failure, log the degradation, close the client, and continue with the nil/no-op badge implementation.

<details>

```
In `room-service/main.go` around line 286: When VALKEY_ADDRS is configured—which the compose file does by default—a failed startup Ping calls os.Exit. Thus a Valkey outage during restart takes room-service offline even though badge operations are documented as fail-open and the handler supports a nil cache. The same startup pattern exists in inbox-worker and user-service. On connection failure, log the degradation, close the client, and continue with the nil/no-op badge implementation.

```

</details>

pkg/model/subscription.go
```diff
@@ -36,10 +36,15 @@ type Subscription struct {
...
 	HasMention         bool             `json:"hasMention" bson:"hasMention"`
-	ThreadUnread       []string         `json:"threadUnread,omitempty" bson:"threadUnread,omitempty"`
-	Alert              bool             `json:"alert" bson:"alert"`
-	Muted              bool             `json:"muted" bson:"muted"`
-	Favorite           bool             `json:"favorite" bson:"favorite"`
+	// ThreadUnread lists parent-message IDs of threads in this room with
+	// replies the user hasn't read. Written by message-worker on each thread
+	// reply (federated to the user's home replica via thread_unread_added),
+	// cleared by message.thread.read / thread.read.all. Feeds the badge count
+	// only — precise thread state stays on the derived lastSeenAt model.
```
julianshen, line 43: _🏷️ data-modeling_ | _🟠 Warning_ | _`mongodb`_

**The embedded threadUnread array has unbounded growth.**

Each distinct unread thread appends another parent-message ID, but neither room membership limits nor $addToSet bound the number of threads a user can follow over a room's lifetime. A user who never clears thread reads can eventually approach MongoDB's 16 MiB document limit; subsequent AddThreadUnread writes then fail and message-worker retries before persisting the reply. Store unread entries as separately indexed documents, such as a unique {roomId, account, parentMessageId} collection, or enforce a documented hard bound with an overflow fallback.

<details>

```
In `pkg/model/subscription.go` around line 43: Each distinct unread thread appends another parent-message ID, but neither room membership limits nor $addToSet bound the number of threads a user can follow over a room's lifetime. A user who never clears thread reads can eventually approach MongoDB's 16 MiB document limit; subsequent AddThreadUnread writes then fail and message-worker retries before persisting the reply. Store unread entries as separately indexed documents, such as a unique {roomId, account, parentMessageId} collection, or enforce a documented hard bound with an overflow fallback.

```

</details>

julianshen (commented):
tGD inline review

user-service/service/subscriptions.go
```diff
@@ -548,45 +547,49 @@ func (s *UserService) GetByRoomID(c *natsrouter.Context, req models.GetByRoomIDR
...
+	// set and marker.
+	if s.badgeCacheFirst {
+		if n, fresh := s.badge.Count(c, account); fresh {
+			return &models.CountResponse{Count: n}, nil
+		}
+	}
+	ids, err := s.unreadRooms(c, account)
+	if err != nil {
+		return nil, err
+	}
```
julianshen, line 568: _🏷️ cache-consistency_ | _🟠 Warning_ | _`mongodb`_

**Partial cross-site results are marked as fully fresh.**

unreadRooms intentionally skips a site when GetRoomsMeta fails but still returns nil error. CountSubscriptions then unconditionally calls Reseed, and BadgeCountBatch similarly calls Seed, both of which write the freshness marker. A transient remote Mongo/RPC outage therefore becomes a persistent undercount served from cache until TTL rather than a one-request degradation. Return a completeness flag from unreadRooms and only Seed/Reseed the marker when every required site responded.

<details>

```
In `user-service/service/subscriptions.go` around line 568: unreadRooms intentionally skips a site when GetRoomsMeta fails but still returns nil error. CountSubscriptions then unconditionally calls Reseed, and BadgeCountBatch similarly calls Seed, both of which write the freshness marker. A transient remote Mongo/RPC outage therefore becomes a persistent undercount served from cache until TTL rather than a one-request degradation. Return a completeness flag from unreadRooms and only Seed/Reseed the marker when every required site responded.

```

</details>

notification-worker/handler.go
```diff
@@ -284,6 +306,69 @@ func (h *Handler) HandleMessage(ctx context.Context, data []byte) (retErr error)
...
+	}
+
+	var (
+		mu     sync.Mutex
+		merged = make(map[string]int, len(accounts))
+		g      errgroup.Group
+	)
+	for siteID, siteAccounts := range bySite {
+		g.Go(func() error {
+			counts, err := h.deps.BadgeClient.Counts(ctx, siteID, roomID, siteAccounts)
```
julianshen, line 338: _🏷️ nats-backpressure_ | _🟠 Warning_ | _`nats`_

**Badge RPC fan-out has no concurrency or batch bound.**

fetchUnreadCounts starts one unrestricted goroutine and NATS request per distinct HomeSiteID, and each site's full badge audience is sent as one unbounded accounts array. A large multi-site room can therefore create room-size concurrent requests; a large single-site room instead gives BadgeCountBatch hundreds of cold misses that it processes sequentially, each potentially issuing nested room RPCs. The caller times out after 5 seconds, but NATS request cancellation is not propagated to the responder, so that work can continue after the reply inbox is gone. Chunk accounts to a configured maximum, apply an errgroup/worker limit, validate the same maximum at user-service, and bound responder work with its own deadline.

<details>

```
In `notification-worker/handler.go` around line 338: fetchUnreadCounts starts one unrestricted goroutine and NATS request per distinct HomeSiteID, and each site's full badge audience is sent as one unbounded accounts array. A large multi-site room can therefore create room-size concurrent requests; a large single-site room instead gives BadgeCountBatch hundreds of cold misses that it processes sequentially, each potentially issuing nested room RPCs. The caller times out after 5 seconds, but NATS request cancellation is not propagated to the responder, so that work can continue after the reply inbox is gone. Chunk accounts to a configured maximum, apply an errgroup/worker limit, validate the same maximum at user-service, and bound responder work with its own deadline.

```

</details>

pkg/badgecache/badgecache.go
```diff
@@ -0,0 +1,246 @@
...
+end
+redis.call('SADD', KEYS[1], ARGV[1])
+redis.call('EXPIRE', KEYS[1], ARGV[2])
+redis.call('SET', KEYS[2], '1', 'EX', ARGV[2])
```
julianshen, line 29: _⚡ memory-exhaustion_ | _🟠 Warning_ | _`oom`_

**Badge sets have no cardinality bound.**

Each notification adds another room ID to the account's Valkey SET, while every bump refreshes the TTL for the entire set. The display cap only limits the returned count; it does not limit stored members. An active account receiving messages across continually created or joined rooms can therefore retain room IDs indefinitely, eventually exhausting Valkey memory and causing eviction, OOM, or cache-failure cascades. Add a hard per-account cardinality limit in the Lua seed/bump/reseed paths; if exceeded, delete the set and freshness marker so callers fall back to bounded recomputation. Ensure Valkey also has an explicit max-memory/eviction policy.

<details>

```
In `pkg/badgecache/badgecache.go` around line 29: Each notification adds another room ID to the account's Valkey SET, while every bump refreshes the TTL for the entire set. The display cap only limits the returned count; it does not limit stored members. An active account receiving messages across continually created or joined rooms can therefore retain room IDs indefinitely, eventually exhausting Valkey memory and causing eviction, OOM, or cache-failure cascades. Add a hard per-account cardinality limit in the Lua seed/bump/reseed paths; if exceeded, delete the set and freshness marker so callers fall back to bounded recomputation. Ensure Valkey also has an explicit max-memory/eviction policy.

```

</details>

pkg/model/subscription.go
```diff
@@ -36,10 +36,15 @@ type Subscription struct {
...
+	ThreadUnread []string `json:"threadUnread,omitempty" bson:"threadUnread,omitempty"`
+	Alert        bool     `json:"alert" bson:"alert"`
```
julianshen, line 45: _⚡ memory-exhaustion_ | _🟠 Warning_ | _`oom`_

**Unread thread IDs grow without a hard limit.**

Every distinct replied-to parent ID is `$addToSet` into this array, and entries persist until that thread or all threads are read. A room member can repeatedly create threads and mention another member, growing the victim's subscription document without a cardinality limit. Mongo's 16 MB document ceiling is only a weak bound: user-service fetches up to `maxSubs` subscriptions and decodes their full arrays, so multiple bloated subscriptions can cause large allocations, GC thrashing, or process OOM; reaching the Mongo ceiling also makes subsequent message processing fail and retry. Cap retained IDs per subscription or move this state to a bounded/separately paged representation, and avoid projecting the full array on paths that only need a non-empty check.

<details>

```
In `pkg/model/subscription.go` around line 45: Every distinct replied-to parent ID is `$addToSet` into this array, and entries persist until that thread or all threads are read. A room member can repeatedly create threads and mention another member, growing the victim's subscription document without a cardinality limit. Mongo's 16 MB document ceiling is only a weak bound: user-service fetches up to `maxSubs` subscriptions and decodes their full arrays, so multiple bloated subscriptions can cause large allocations, GC thrashing, or process OOM; reaching the Mongo ceiling also makes subsequent message processing fail and retry. Cap retained IDs per subscription or move this state to a bounded/separately paged representation, and avoid projecting the full array on paths that only need a non-empty check.

```

</details>

user-service/service/badge.go
julianshen, line 26: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Cold badge batches trigger serialized, uncoalesced recomputations.**

Every cache miss enters this account loop and calls `unreadRooms`, which performs a Mongo aggregation and may fan out cross-site RPCs. After Valkey eviction, restart, or read-driven invalidation, a large recipient batch therefore performs up to one full recomputation per account serially; concurrent messages containing the same accounts duplicate that work. This can exceed notification-worker's 5-second timeout while creating a Mongo/RPC stampede. Batch the source-of-truth reads where possible, process misses through a bounded worker pool, and coalesce concurrent recomputations per account (or omit counts while the cache is unavailable).

<details>

```
In `user-service/service/badge.go` around line 26: Every cache miss enters this account loop and calls `unreadRooms`, which performs a Mongo aggregation and may fan out cross-site RPCs. After Valkey eviction, restart, or read-driven invalidation, a large recipient batch therefore performs up to one full recomputation per account serially; concurrent messages containing the same accounts duplicate that work. This can exceed notification-worker's 5-second timeout while creating a Mongo/RPC stampede. Batch the source-of-truth reads where possible, process misses through a bounded worker pool, and coalesce concurrent recomputations per account (or omit counts while the cache is unavailable).

```

</details>

notification-worker/handler.go
```diff
@@ -284,6 +306,69 @@ func (h *Handler) HandleMessage(ctx context.Context, data []byte) (retErr error)
...
+		return nil
+	}
+
+	var (
+		mu     sync.Mutex
+		merged = make(map[string]int, len(accounts))
+		g      errgroup.Group
+	)
+	for siteID, siteAccounts := range bySite {
+		g.Go(func() error {
```
julianshen, line 337: _🎯 concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Badge fan-out is unbounded and gates push delivery.**

A message launches one goroutine and NATS request per distinct recipient home site with no concurrency limit, then waits for every request before emitting any push batch. A highly federated room can therefore create many simultaneous RPCs per message, and one degraded site delays all recipients for up to the 5-second badge timeout. Under message bursts this multiplies in-flight goroutines and requests and occupies notification workers. Apply a bounded per-message/site concurrency limit and a short overall badge budget; once the budget expires, emit pushes with only the counts already available.

<details>

```
In `notification-worker/handler.go` around line 337: A message launches one goroutine and NATS request per distinct recipient home site with no concurrency limit, then waits for every request before emitting any push batch. A highly federated room can therefore create many simultaneous RPCs per message, and one degraded site delays all recipients for up to the 5-second badge timeout. Under message bursts this multiplies in-flight goroutines and requests and occupies notification workers. Apply a bounded per-message/site concurrency limit and a short overall badge budget; once the budget expires, emit pushes with only the counts already available.

```

</details>

message-worker/store_mongo.go
```diff
@@ -184,6 +184,22 @@ func (s *threadStoreMongo) AddReplyAccounts(ctx context.Context, threadRoomID st
...
+// AddThreadUnread marks parentMessageID unread for accounts' subscriptions in
+// roomID via a single $addToSet UpdateMany. Idempotent under JetStream
+// redelivery; accounts not subscribed simply match nothing.
+func (s *threadStoreMongo) AddThreadUnread(ctx context.Context, roomID, parentMessageID string, accounts []string) error {
+	if len(accounts) == 0 {
+		return nil
+	}
+	if _, err := s.subscriptions.UpdateMany(ctx,
+		bson.M{"roomId": roomID, "u.account": bson.M{"$in": accounts}},
+		bson.M{"$addToSet": bson.M{"threadUnread": parentMessageID}},
```
julianshen, line 196: _🏷️ resource-growth_ | _🟠 Warning_ | _`performance-and-concurrency`_

**The unread-thread array grows without a bound.**

Each distinct followed thread adds another parent ID to `Subscription.threadUnread`, with no cap, retention, or automatic cleanup when users only read the main room. In long-lived thread-heavy rooms, inactive users accumulate an ever-growing Mongo array. `$addToSet` becomes progressively more expensive and the subscription document can eventually approach MongoDB's document-size limit, causing message processing failures. Store unread threads as separately indexed records with retention, or enforce a bounded representation that preserves the badge's room-level existence semantics.

<details>

```
In `message-worker/store_mongo.go` around line 196: Each distinct followed thread adds another parent ID to `Subscription.threadUnread`, with no cap, retention, or automatic cleanup when users only read the main room. In long-lived thread-heavy rooms, inactive users accumulate an ever-growing Mongo array. `$addToSet` becomes progressively more expensive and the subscription document can eventually approach MongoDB's document-size limit, causing message processing failures. Store unread threads as separately indexed records with retention, or enforce a bounded representation that preserves the badge's room-level existence semantics.

```

</details>

message-worker/handler.go
```diff
@@ -586,17 +623,84 @@ func (h *Handler) markThreadMentions(ctx context.Context, msg *model.Message, th
...
+	}
+	siteByAccount := make(map[string]string, len(users))
+	for i := range users {
+		siteByAccount[users[i].Account] = users[i].SiteID
+	}
+	bySite := map[string][]string{}
+	var missing []string
+	for _, a := range accounts {
+		site, ok := siteByAccount[a]
+		switch {
```
julianshen, line 681: _🏷️ data-integrity_ | _🟠 Warning_ | _`silent-failure`_

**Missing user records silently skip unread federation.**

FindUsersByAccounts may return only part of the requested recipient set. Accounts absent from that result are merely logged and omitted from bySite, after which the message is acknowledged. For a remotely homed follower, its home subscription never receives thread_unread_added, so subscription.count can permanently miss that unread thread. Treat missing routing records for active recipients as retryable, derive the home site from an authoritative source already carried by the follower data, or verify the account is no longer active before allowing the skip.

<details>

```
In `message-worker/handler.go` around line 681: FindUsersByAccounts may return only part of the requested recipient set. Accounts absent from that result are merely logged and omitted from bySite, after which the message is acknowledged. For a remotely homed follower, its home subscription never receives thread_unread_added, so subscription.count can permanently miss that unread thread. Treat missing routing records for active recipients as retryable, derive the home site from an authoritative source already carried by the follower data, or verify the account is no longer active before allowing the skip.

```

</details>

user-service/service/subscriptions.go
julianshen, line 574: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**A degraded unread-room result is cached as fresh.**

`unreadRooms` absorbs cross-site `GetRoomsMeta` failures and returns a partial ID list with no error, after which `Reseed` writes the freshness marker. `BadgeCountBatch` similarly seeds partial results. Once cache-first counting is enabled, that undercount is treated as authoritative until invalidation or TTL. Propagate a completeness flag and only Seed/Reseed with a marker when every required site responded; degraded results may still be returned best-effort but must remain stale.

<details>

```
In `user-service/service/subscriptions.go` around line 574: `unreadRooms` absorbs cross-site `GetRoomsMeta` failures and returns a partial ID list with no error, after which `Reseed` writes the freshness marker. `BadgeCountBatch` similarly seeds partial results. Once cache-first counting is enabled, that undercount is treated as authoritative until invalidation or TTL. Propagate a completeness flag and only Seed/Reseed with a marker when every required site responded; degraded results may still be returned best-effort but must remain stale.

```

</details>

hmchangw (approved):

PR #168, opened 2026-08-01
Enrich "added" subscription.update with room info + inline key; room.key becomes rotation-only

mliu33:
## Problem

A newly added room member's `subscription.update` (`action: "added"`) carried only the bare persisted `Subscription` plus a flat `roomName` — none of the `SubscriptionRoom` fields the FE gets from `subscription.list` (canonical name, `privateKey`/`keyVersion`, `crossSite`, `lastMsgId`/`lastMsgAt`, `minUserLastSeenAt`, `userCount`/`appCount`, `lastMentionAllAt`). The FE could not render the sidebar row like a list item and had to correlate a separate `room.key` event for the key.

Design spec: `docs/superpowers/specs/2026-08-01-subscription-update-room-info-design.md` · Plan: `docs/superpowers/plans/2026-08-01-subscription-update-room-info.md`

## Change

**room-worker — publish-time enrichment (no new RPCs, no schema changes):**
- New `subscriptionRoomFor(room, pair)` builds the `SubscriptionRoom` view; a shared `publishSubscriptionAdded` fans it out on all four `"added"` paths (member.add, create-room, DM-sync, self-DM). `Subscription.Room` is `bson:"-"` — never persisted; `previewMessage` deliberately stays nil (a no-shared-history member must not see the prior last message).
- **member.add**: key pair checked first (`requireKeyPair`, permanent error when absent — unchanged semantics), then the room is re-read post-count-writes so `userCount` includes the members just added; both reads happen before any publish so a failure redelivers with nothing half-sent.
- **create-room**: cross-site marking now lands **before** the fan-out (a channel is born provisionally `crossSite=false`; fanning that out would have routed remote-member creates to the local namespace) — with a regression test. Counts derive from the initial roster.
- **Initial `room.key` fan-outs removed** on member.add and create-room — the key rides the event (`subscription.room.privateKey`/`keyVersion`, same base64 form). Kept: rotation fan-out on member removal, the `room.key.get` recovery RPC, and the Teams-migration bulk fan-out (documented — that path emits no live added event).
- Cross-site adds need no inbox/outbox changes: the enriched event is built at the room's home site and routes to the member's home site over the supercluster as before.

**chat-frontend:**
- `subToRoom` and a new `keyEntryFor` are exported from the api layer and shared by the sidebar bootstrap and the live `added` path, so both map the wire shape and the key contract identically. On `added`, the room renders immediately and the inline key seeds `RoomKeysContext` (with `crossSite ?? true` fail-safe preserved).

**Docs:** `docs/client-api.md` + derived `docs/client-api/events.md` and `request-reply.md` updated — `added` events now document the embedded `room` object; `room.key` documented as rotation-only.

## Compatibility note

FE builds predating this change read keys only from `room.key` events; deploy the updated FE with or before the backend so new rooms' keys aren't deferred to the next `subscription.list`.

## Verification

- TDD throughout (red → green per path); `make lint`, full `make test` (103 pkgs, race) green
- `make sast` green (gosec / govulncheck / semgrep)
- Full `make test-integration`: 106 pkgs green incl. room-worker/room-service/inbox-worker/user-service; the only failures are `pkg/roomcrypto` + `pkg/roomkeysender` (untouched by this branch) whose Node test containers cannot reach npm from this sandbox (`EAI_AGAIN`)
- chat-frontend: 805 tests green, `npm run typecheck` clean
- Touched functions coverage: `subscriptionRoomFor` 100%, `processAddMembers` 93%, `finishCreateRoom` 87%, `publishSubscriptionAdded`/`publishSubscriptionUpdates` 87%
- Multi-agent review pass (reuse/simplification/efficiency/altitude) applied; skipped-by-design: merging the member.add `GetRoom`+`keyStore.Get` reads into one query (store-interface churn beyond this diff) and returning counts from `ReconcileMemberCounts` (same)

https://claude.ai/code/session_01NPWf56GBovdH3qECcywGVC

---
_Generated by [Claude Code](https://claude.ai/code/session_01NPWf56GBovdH3qECcywGVC)_

## Summary by CodeRabbit

* **New Features**
  * Subscription updates now include room details, sidebar metadata, and initial encryption keys when available.
  * Initial room encryption keys are delivered inline during room creation, member additions, and reconnect bootstrap.
  * Room-key events are reserved for key rotation scenarios.

* **Bug Fixes**
  * Improved local and cross-site room enrichment for live updates.
  * Added safeguards against stale session data and unavailable room metadata or keys.

* **Documentation**
  * Updated client API and event documentation for subscription and encryption-key delivery.


mliu33:
Review findings addressed in 8e8378a (single commit for easy diffing).

**Fixed**
- **Redelivery-unsafe re-read (critical, `room-worker/handler.go`):** the room key is now fetched and validated *before* `BulkCreateSubscriptions` — a key failure leaves no partial state, so a clean redelivery replays the whole sequence. The post-commit `GetRoom` re-read is now best-effort: on failure it degrades to the pre-add `GetRoomMeta` view (key still inline) instead of returning an error, so committed subs always get their `added` event in the same delivery. Covered by `TestHandler_ProcessAddMembers_KeyStoreGetFailureFailsBeforePublish` (asserts no `BulkCreateSubscriptions` on key failure) and the new `TestHandler_ProcessAddMembers_GetRoomFailureDegradesToMetaView`. A persisted-completion flag wasn't needed once the only remaining post-commit read became non-failing.
- **`isCurrent()` guards (frontend):** the `subscription.update` callback and the sidebar-bootstrap `.then` now gate on the effect generation instead of the resettable `cancelledRef`; added the suggested logout→login regression test (stale bootstrap resolving after re-login seeds no keys, opens no subs) and confirmed it fails against the old guard.
- **`publishSubscriptionUpdates` channel guard** now returns after logging (fail closed).
- **Docs:** `Subscription.room` row documents both population sources; canonical action enum gains `"read"` and the `roomName` wording now states the actual `omitempty` behavior (matching `events.md`); example `timestamp` values corrected to match their 2026 RFC3339 dates in both files; `request-reply.md` aligned with the canonical cross-site bot-membership rule (backend allows bots whose home site differs — `docs/client-api.md` §Add Members is the source of truth).

**Skipped (with reasons)**
- The three findings against code snippets in `docs/superpowers/plans/2026-08-01-….md` (marshal-error handling, `p_admin` app classification, `room.name` mapping): the plan is a historical execution artifact; the production code already behaves as requested — `publishSubscriptionAdded` logs-and-continues on marshal failure, `newSub` stamps `p_admin` as `IsBot` via `model.IsPlatformAdminAccount` so the count loop classifies it as an app, and the FE maps the wire shape through the shared `subToRoom` used by `subscription.list`.
- The two plan-doc nitpicks (make targets, request-reply.md listing): same rationale.

---

chenjr0719 (approved):
Reviewed end-to-end and traced the room-key path through the added-subscription fan-out. Clean and carefully done:

- The inline key is delivered on a per-user, JWT-scoped subject — not broadcast.
- It's excluded from BSON, so it's never persisted at rest.
- The fan-out logs only account + request id, never the payload.
- Only newly-subscribed accounts receive it (matches the prior key-fan-out target); existing members aren't re-sent.
- The key is fetched + validated pre-commit, so a redelivery can't silently drop the added event and its key.

Also verified on a live run: adding a member produced an `added` subscription.update carrying the room key inline (privateKey + keyVersion) on the recipient's per-user subject. Cross-site key delivery is unaffected — both the prior room.key and the inline key ride per-account subjects, and the cross-site path isn't touched here.

One non-blocking question: the added room view carries `lastMsgId`/`lastMsgAt`/`minUserLastSeenAt` while `previewMessage` is nil — I read that as intended (metadata only, matching the subscription.list shape); just confirming.

Approving.

PR #170, opened 2026-08-03
feat(teams): verify Teams room creation per site

cocoamix00:
## Verification status

The integration tests **have now been executed** against a real MongoDB and NATS. An earlier revision of this description said they never had — that is no longer true, and the detail below replaces it.

| Suite | Result |
|---|---|
| `make test-integration SERVICE=teams-room-inspector` | ok |
| `make test-integration SERVICE=teams-room-verify` | ok |
| `make test-integration SERVICE=teams-room-creation` | ok |
| `make test-integration SERVICE=teams-chat-sync` | ok |
| `make test` repo-wide, `-race` | exit 0, 0 failures |
| `make lint` | 0 issues |

The three specific risks raised in review are now disproven by named passing tests:

- `TestMongoStore_EnsureIndexes` — the new `needVerify_pending` partial index really is created.
- `TestMongoStore_MarkVerified_ClearsOnlyMatchingRefs` — the `updatedAt` compare-and-set round-trips correctly through BSON. The "a `time.Time` vs millisecond-datetime mismatch means nothing ever clears" risk does not materialise.
- `TestMongoStore_RoomStates` — the `$group` aggregation and its `{_id, count}` decoding work against a real server.

### Coverage

| Package | Unit only | With integration |
|---|---|---|
| `teams-room-verify` | 80.4% | **86.9%** |
| `teams-room-inspector` | 47.7% | **64.8%** |

`teams-room-inspector` is below the 80% floor. Its logic is fully covered — `handler.go` 100%, `routes.go` 100%, `store_mongo.go` 88% — and **excluding `main.go` the package measures 95.8%**. Everything still uncovered is inside `run()`: connecting Mongo, binding the listener, blocking on signals, draining. Reaching those 14 remaining statements requires a boot-and-shutdown test that signals the test binary, which is deliberately left to a follow-up.

For reference, the merged sibling `teams-room-creation` measures 56.1% unit-only, and the repo's own coverage gate (`tools/coveragecheck`, used for the soak suite) already excludes `soak_main.go` on the same reasoning.

### Still unverified

**The container images have not been built.** Both Dockerfiles fail at `apk add --no-cache ca-certificates` because build containers here get no proxy configuration and cannot reach `dl-cdn.alpinelinux.org`. This is environmental, not a defect in these Dockerfiles — as a control, the **already-merged** `teams-room-creation/deploy/Dockerfile` fails at the identical line, and 30 of the repo's 32 service Dockerfiles use that same instruction. The Go builder stage *does* succeed in both images, which is the part specific to this PR: the `COPY` set is sufficient and the binaries compile in-container. CI with normal network egress should build them; please confirm it does.

**`make sast` is partial**: gosec passes; govulncheck is blocked by the proxy (`vuln.go.dev` unreachable) and semgrep by a pipx/uv version mismatch. Both need a CI re-run.

## What this adds

`teams-room-creation` publishes room-creation events per site and clears `needCreateRoom`, but nothing confirmed the destination site actually built the room. A `room-worker` that skipped a chat (per-chat reconcile failure is WARN-and-continue) left a silent gap: flag cleared, room absent.

This adds a verification lane.

| Component | Shape | Scope |
|---|---|---|
| `teams-room-creation` (change) | existing CronJob | global |
| `teams-room-verify` (new) | CronJob, run-to-completion | global |
| `teams-room-inspector` (new) | Gin HTTP service | one deployment per site |

1. **`teams-room-creation`** now raises `needVerify=true` in the same compare-and-set write that clears `needCreateRoom`. Because that write only happens after JetStream acknowledges the batch, the flag means precisely *a creation event for this chat was durably published*.
2. **`teams-room-inspector`** — per-site, read-only. `POST /internal/teams/rooms/verify` takes up to 500 Teams chat ids and reports, per chat, whether the room exists and how many subscriptions point at it. Two projected queries, no `$lookup`. Reads through a secondary-preferred client.
3. **`teams-room-verify`** — global CronJob. Lists flagged chats, groups by `siteId`, batches, fans out to each site's inspector over a semaphore, compares, and clears the flag **only** for chats that converged. Mismatches are logged and keep their flag for the next run.

Room ids are derived by `teamsmigrate.RoomIDFromChatID` — the same function `room-worker` uses when creating the room, so writer and reader cannot drift apart silently.

## Design decisions worth reviewing

**Expected count is `accountsPresent`, not `len(Members)`.** `room-worker` subscribes only distinct, non-empty accounts — it skips guests/externals with no account and collapses duplicates. The design originally specified the raw member count; review showed that made any guest-containing chat **unable to ever converge**, since `MarkVerified` is the only writer that clears the flag and the mismatch branch never produces a ref. The flagged set would have grown without bound, burying genuine `missing_room` alarms. The raw count is still logged as `expected_members` beside `accounts_present` for diagnosis.

**Report-only, no self-healing.** Verification never re-flags `needCreateRoom`. A mismatch is reported and the chat stays flagged.

**No auth on the inspector endpoint.** Cluster-internal; network policy is the boundary, matching the fact that no service in this repo does service-to-service HTTP auth. The body is capped by `http.MaxBytesReader` since the endpoint is unauthenticated.

**Batch cap is shared.** `model.TeamsRoomVerifyMaxChatIDs` is enforced by the inspector *and* by `teams-room-verify`'s config validation, so a `VERIFY_BATCH_SIZE` above it fails at startup instead of 400ing every batch and leaving chats flagged forever — a failure that reads exactly like a healthy run.

**Secondary reads can produce false alarms.** Replication lag can make a just-created room look missing. Self-correcting (the chat keeps its flag and is rechecked), but schedule the verify cron with a comfortable offset behind the creation cron.

## Configuration

`teams-room-verify`: `MONGO_URI`, `MONGO_DB`, `TEAMS_VERIFY_SITE_URLS` (JSON `siteId → inspector base URL`, required), `VERIFY_BATCH_SIZE` (200, capped at 500), `MAX_WORKERS` (8).

`teams-room-inspector`: `MONGO_URI`, `MONGO_DB`, `SITE_ID`, `PORT` (8080).

Requires `needVerify_pending`, a new partial index on `teams_chat` created by `teams-chat-sync` — the documented owner of that collection's indexes.

## Known follow-ups (not in this PR)

- A boot-and-shutdown test for `teams-room-inspector`'s `run()`, which would close the remaining coverage gap. Kept separate because it signals the test binary (`shutdown.Wait` blocks on a real SIGTERM), which changes how that package's tests behave.
- `planBatches` is duplicated verbatim between `teams-room-creation` and `teams-room-verify` (~23 identical lines). Extraction touches a shipped service and belongs in its own PR.
- `run()` doesn't observe `ctx.Err()`, so a SIGTERM mid-pass emits a burst of warnings resembling an outage before exiting.
- A member whose `resolveMember` fails still mismatches — correct behavior, but the log gives no hint that resolution was the cause.
- No positive boundary test at exactly 500 chat ids; the inspector's bind error is discarded rather than attached via `errcode.WithCause`.

## Docs

`docs/superpowers/specs/2026-08-01-teams-room-verify-design.md` (design) and `docs/superpowers/plans/2026-08-01-teams-room-verify.md` (implementation plan) are included and reflect the shipped behavior. No `docs/client-api.md` change: these are service-to-service HTTP contracts, not `chat.user.*` NATS RPCs or `auth-service` routes.

## Summary by CodeRabbit

* **New Features**
  * Added automated verification of Teams room creation across sites.
  * Added inspection endpoints reporting room existence, member counts, and subscription counts.
  * Added batching, concurrency controls, health checks, and scheduled verification processing.
  * Chats are marked for verification after room creation and cleared only after successful validation.

* **Bug Fixes**
  * Prevents incomplete, mismatched, or misrouted room data from being marked as verified.

* **Tests**
  * Added comprehensive unit, integration, serialization, and endpoint coverage for verification workflows.


mliu33 (approved):
Outstanding work! Thanks!

Issue #171, opened 2026-08-03
[perf] subscription.list preview walks up to 50 message rows per room to return one


chenjr0719:
## Summary
`subscription.list` (sidebar/room-list load) builds each room's **preview message** by walking a **50-message page per room** — materializing + AES-decrypting up to 50 full rows — to return the first eligible one. Cost scales **O(N_subscriptions) × 50 heavy rows per load**. The filter (skip a system/deleted tail) is correct; the 50-row *first* page is over-eager tuning.

## Reproduce steps
Verified on a dev stack (`main@a9489c6d`):

**Correctness (current behavior — all pass):**
1. eligible-head (normal last message) → preview = it.
2. system-tail (add a member → a system "joined" becomes the newest) → preview **skips it, stays on the real message**.
3. quoted-head (a quoted reply is newest) → preview = the reply's own content, **no `quotedParent` field** (`PreviewMessage` carries none).

**Latency (`msg.history`, a 110-message room, 3 runs each):** limit 10 → ~3.3ms, 20 → ~4.6ms, 50 → ~6.1ms, 100 → ~6.9ms (sub-linear on a warm single node). The preview walk uses **page=50 ≈ ~6ms per room**, ×N rooms (concurrency 16). The amplification is **aggregate** (rows/decrypts across all rooms per load), not single-call latency — it shows most under scale + concurrency.

## Root cause (verified, file:line on `main`)
- `user-service/service/subscriptions.go:209` → `enrichLastMessage:327` → `history.RoomsGet:347` (batched per site — correct; the cost is inside `RoomsGet`).
- `history-service/internal/service/rooms.go:42` per-room fan-out (semaphore 16) → `roomLastPreviewMessage:71` → `GetMessagesBefore:87` with `lastMsgWalkPageSize = 50` (`rooms.go:19`). The cassandra repo projects full `baseColumns` and decrypts every scanned row, to return the first eligible (`rooms.go:96-104`).
- Secondary: `apps.assistant.name` `FindOne` (`mongorepo/app.go:27`) is uncached + unindexed, a scan per bot-DM room.
- Provenance: the batched per-room preview mechanism landed in #393; the eligible-message filter walk + `lastMsgWalkPageSize = 50` landed in #104 (2026-07-24, "room-list preview — filter system/quoted + enrich"), chosen to tolerate a trailing tail of ineligible messages — that 50-row first read is the amplification.

## Message-path trace
[...]

PR #172, opened 2026-08-03
perf(history): small first page for room preview, escalate past ineligible tail

chenjr0719:
## Summary

`subscription.list` with `includeLastMessage` (the client room-list) resolves each room's
last-message preview by walking message history at read time. The walk read a fixed
50-message page per room — but the latest message is eligible in the vast majority of
rooms, so it materialized and decrypted 50 rows to return one. On a room-list of N
subscriptions that is N × 50 rows read/decrypted per load.

This reads a small first page (default 3) and escalates to the full walk-back page
(default 50) only when the head is entirely system/deleted. Same message is selected —
only the row count read on the common path drops (50 → 3). Both page sizes are env-tunable
(`PREVIEW_FIRST_PAGE_SIZE`, `PREVIEW_WALK_PAGE_SIZE`), mirroring the service's existing
`LARGE_ROOM_THRESHOLD`.

Addresses #171.

## Changes

- `history-service/internal/service/rooms.go`: `roomLastPreviewMessage` reads a small
  first page; on an all-ineligible full first page it escalates to the full walk-back page
  for subsequent pages. The eligible-message selection and the max-page cap are unchanged.
- `history-service/internal/config`: `PreviewFirstPageSize` (3) + `PreviewWalkPageSize`
  (50) config fields, threaded into the service.
- Regression test for the escalation path (a full ineligible first page forces one
  escalation that surfaces the eligible message).

## Verification

- `history-service` unit tests green, including the new escalation case; `go vet` and
  golangci-lint clean.
- Measured before/after on a running stack — room-list latency (`subscription.list` with
  previews) over an account whose rooms are **all large** (each ≥50 messages, so every one
  hits the walk), varying the number of rooms enriched:

  | rooms | before p50 | after p50 | delta |
  |------:|-----------:|----------:|------:|
  | 1     | 64.8 ms    | 67.4 ms   | ~0 (within round-trip noise) |
  | 5     | 93.7 ms    | 73.2 ms   | ~22% |
  | 10    | 85.1 ms    | 83.9 ms   | ~noise |
  | 20    | 102.1 ms   | 85.0 ms   | ~17% |
  | 50    | 165.0 ms   | 110.4 ms  | ~33% |
  | 100   | 243.6 ms   | 148.4 ms  | ~39% |

  The per-room delta is small at low room counts (a few ms hidden in the fixed round-trip
  + subscription read) and grows with the number of large rooms — the aggregate rows
  materialized/decrypted per load drop ~16× per large room. Same message selected;
  preview correctness unchanged, no room lost its preview.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved room history previews to find the latest eligible message when the newest messages are system-generated or deleted.
  * Preview searches now expand automatically when the initial results do not contain an eligible message.

* **Configuration**
  * Added settings to control the initial preview size and expanded search page size.

### DIFF
--- a/history-service/internal/config/config.go
+++ b/history-service/internal/config/config.go
@@ -47,6 +47,8 @@ type Config struct {
 	MessageReadMaxBuckets   int             `env:"MESSAGE_READ_MAX_BUCKETS"    envDefault:"122"`
 	MessageHistoryFloorDays int             `env:"MESSAGE_HISTORY_FLOOR_DAYS"  envDefault:"365"`
 	LargeRoomThreshold      int             `env:"LARGE_ROOM_THRESHOLD"        envDefault:"500"`
+	PreviewFirstPageSize    int             `env:"PREVIEW_FIRST_PAGE_SIZE"     envDefault:"3"`
+	PreviewWalkPageSize     int             `env:"PREVIEW_WALK_PAGE_SIZE"      envDefault:"50"`
 	MaxPinnedPerRoom        int             `env:"MAX_PINNED_PER_ROOM"         envDefault:"10"`
 	PinEnabled              bool            `env:"PIN_ENABLED"                 envDefault:"true"`
 
--- a/history-service/internal/service/rooms.go
+++ b/history-service/internal/service/rooms.go
@@ -16,8 +16,7 @@ import (
 const (
 	maxRoomsGetBatch       = 100 // mirrors maxGetByIDsBatchSize
 	maxRoomsGetConcurrency = 16  // mirrors cassrepo.maxConcurrentIDReads
-	lastMsgWalkPageSize    = 50  // messages scanned per walk-back page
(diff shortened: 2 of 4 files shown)


chenjr0719:
Thanks — addressed both:

- **Walk-back coverage** — the small first page had cut the worst-case walk from 250 to 203 messages; bumped the page cap so it's back to ~253 (small first page + 5 full pages), keeping eligible-message selection unchanged.
- **Escalation test** — now asserts the first read uses the small page size (3) and the escalated read uses the full page size (50) with the cursor advanced past the oldest message of the first page.

Both green (unit tests + lint).

chenjr0719:
Superseded by #173, which covers the same read-path optimization more completely — a geometric small-page escalation (single query in the common case) with the ineligible-skip budget preserved, plus a short-TTL preview cache and a batch-size metric to gate the denormalization phase. Closing in favor of it.

PR #176, opened 2026-08-04
feat: guard MongoDB connection pools across all services

hmchangw:
## Summary

MongoDB-accessing services could exhaust their connection pool under load: the NATS router spawns one unbounded goroutine per request, and clients ran on the driver-default pool (100) with no operation timeout, so a burst or a slow Mongo saturates the pool and builds an unbounded, indefinitely-blocked checkout queue.

This PR introduces shared, env-tagged guard helpers and rolls them out to **every long-running service that connects to MongoDB** — an explicit pool cap everywhere, plus a bounded in-flight handler count and a per-request timeout on the request/reply and HTTP services.

## Shared helpers (`pkg/`)

- **`mongoutil.PoolConfig`** — `MONGO_MAX_POOL_SIZE` (default 500) / `MONGO_MIN_POOL_SIZE`; `Options()` + `Validate()`. Full-prefix env tags so it drops into flat or `MONGO_`-prefixed configs as a named field.
- **`natsrouter.GuardConfig`** — `MAX_CONCURRENCY` (default 256, 0 disables) / `REQUEST_TIMEOUT` (default 10s, 0 disables); `Options()` (WithMaxConcurrency), `TimeoutMiddleware()` (HandlerTimeout), `Validate()`.
- **`ginutil.Timeout`** — per-request context-deadline middleware for HTTP services.

## Rollout (by transport)

- **Request/reply (natsrouter)** → pool + concurrency cap + request timeout: history-service, bot-message-handler, bot-room-service, user-presence-service, user-service, search-service, room-service, room-worker, media-service (NATS side).
- **HTTP (Gin)** → pool + `ginutil.Timeout`: admin-service, botplatform-service, portal-service, tcard-service, media-service. **upload-service gets the pool cap only** — it streams large downloads already bounded by `MinioDownloadTimeout`/`WriteTimeout` (5m), which a short request deadline would cancel.
- **Workers** → explicit pool cap only (concurrency already bounded by `MAX_WORKERS`): message-worker, broadcast-worker, notification-worker, message-gatekeeper, inbox-worker, search-sync-worker, hr-sync-worker, bot-message-worker, and the teams-* sync services (pool cap applied to both read and write clients where split).

## Notable per-service decisions

- **history-service** was refactored onto the shared helpers (env vars/defaults unchanged) — it keeps all three guards.
- **bot-message-handler / bot-room-service** had a standalone `MAX_CONCURRENCY`; consolidated onto `GuardConfig` (same env var → runtime unchanged where compose pins it) and additionally gained `REQUEST_TIMEOUT`.
- **user-service** keeps its existing `HANDLER_TIMEOUT` for the per-request deadline and takes only the concurrency cap from Guard — no stacked second timeout.

## Design notes

- `MONGO_MAX_POOL_SIZE` is an authoritative headroom ceiling: connections are lazy and, in request/reply services, further gated by `MAX_CONCURRENCY` (a handler holds ≤1). Size `pods × pool` under the cluster connection limit.
- All new defaults match prior behavior where a value already existed; every guard is disable-able via `=0` and tunable per environment without code changes.

## Testing

- Shared helpers: TDD (Red→Green), race-tested, lint-clean.
- Whole repo builds (`go build ./...`); unit tests pass across all changed packages; `golangci-lint` clean on the changed packages.
- No client-facing handler or `pkg/model` struct changed, so `docs/client-api.md` is untouched.

### DIFF
--- a/history-service/cmd/main.go
+++ b/history-service/cmd/main.go
@@ -95,7 +95,11 @@ func main() {
 		os.Exit(1)
 	}
 
-	mongoClient, err := mongoutil.Connect(ctx, cfg.Mongo.URI, cfg.Mongo.Username, cfg.Mongo.Password, mongoutil.WithObservability(sdk))
+	mongoClient, err := mongoutil.Connect(ctx, cfg.Mongo.URI, cfg.Mongo.Username, cfg.Mongo.Password,
+		mongoutil.WithObservability(sdk),
+		mongoutil.WithMaxPoolSize(cfg.Mongo.MaxPoolSize),
+		mongoutil.WithMinPoolSize(cfg.Mongo.MinPoolSize),
+	)
 	if err != nil {
 		slog.Error("mongo connect failed", "error", err)
 		os.Exit(1)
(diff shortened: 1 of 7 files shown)


chenjr0719 (approved):
Reads clean. The shared `pkg/mongoutil` change is safe — `applyTuning` only writes a pool setting when the option pointer is non-nil, so services that don't pass it are untouched (and it's tested). Config validation is a thorough matrix (0=unbounded rejected, min<=max, 0-as-disable for concurrency/timeout, negatives rejected), and the pool-cap + bounded concurrency + per-request timeout combination is a sound load-shed design that reuses the already-tested router middleware. Ran the mongoutil + config package tests locally — green. LGTM.

PR #177, opened 2026-08-04
Display full date and time in message timestamps

hmchangw:
## Summary
Updated message timestamp formatting across the chat frontend to display both date and time instead of time only. This improves clarity when viewing messages across different days.

## Changes
- **MessageRow component**: Renamed `formatTime()` to `formatDateTime()` and updated it to use `toLocaleString()` with full date options (year, month, day) alongside the existing time format
- **SystemMessage component**: Updated inline timestamp formatting to match, displaying full date and time using `toLocaleString()`
- **Tests**: Updated corresponding test cases to verify both date (year) and time components are rendered:
  - MessageRow test now asserts the year "2026" appears in the timestamp
  - SystemMessage test updated similarly to check for both date and time presence

## Implementation Details
Both components now format timestamps using the same locale-aware approach with options for `year: 'numeric'`, `month: 'numeric'`, `day: 'numeric'`, `hour: '2-digit'`, and `minute: '2-digit'`. This ensures consistent formatting across the application and respects the user's locale settings.

https://claude.ai/code/session_01EQRfrgmsVmoCDnAorqKhV6

### DIFF
--- a/chat-frontend/src/components/shared/MessageList/MessageRow/MessageRow.jsx
+++ b/chat-frontend/src/components/shared/MessageList/MessageRow/MessageRow.jsx
@@ -9,9 +9,15 @@ import { useSubscription } from '@/context/RoomEventsContext'
 import { redactInaccessibleQuoteSnapshot } from '@/lib/redactQuote'
 import './style.css'
 
-function formatTime(dateStr) {
+function formatDateTime(dateStr) {
   const d = new Date(dateStr)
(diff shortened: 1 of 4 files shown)


chenjr0719 (approved):
Reads clean. The `formatTime`→`formatDateTime` rename + `toLocaleString` change is consistent across both components, and the tests assert the rendered year + HH:MM with correct timezone-stable reasoning (the fixed UTC timestamps can't shift out of 2026 by any offset). LGTM.

PR #169, opened 2026-08-03
Implement concurrent bucket fetching for history pagination

hmchangw:
## Summary

Refactor the bucketed-table page-fill logic to support concurrent bucket fetches in fan-out waves, improving pagination latency for sparse/empty bucket runs while maintaining strict row ordering and cursor alignment. The walk now probes only the start bucket initially (matching the old serial behavior for dense regions), then widens to concurrent fetches once underfill is detected.

## Key Changes

- **Extract `bucketFetcher` abstraction**: Define a new `bucketFetcher[T]` interface that encapsulates the contract for fetching up to `limit` rows from a single bucket, with internal short-page draining. This separates the I/O concern from the walk logic.

- **Implement `gocqlBucketFetcher`**: Adapt the existing gocql query + scan pattern into a `bucketFetcher` that drains short driver pages internally until the limit is reached or the bucket is exhausted. Preserves the original per-row error handling (e.g., decrypt failures abort the fetch).

- **Introduce `walkBuckets` function**: New direction-agnostic concurrent walk that:
  - Starts with a single-bucket wave (no speculative reads for dense start regions)
  - Widens to `fanout`-width waves once the start bucket underfills
  - Fetches each wave concurrently via `errgroup` with a configurable limit
  - Assembles rows in strict bucket order (no reordering despite concurrency)
  - Handles mid-bucket page boundaries by re-fetching with the exact consumed count to obtain an aligned resume cursor
  - Respects floor/ceiling bounds and maxBuckets budget

- **Update `fillPage` signature**: Add `fanout` parameter and delegate to `walkBuckets` via the new `gocqlBucketFetcher`.

- **Wire fanout configuration**: Add `defaultWalkFanout = 8` constant to `repository.go` and thread it through `GetMessagesBefore`, `GetMessagesBetweenDesc`, and `GetMessagesAfter` call sites.

- **Comprehensive test suite**: Add 13 new tests covering:
  - Single dense bucket (no fan-out)
  - Sparse fills across multiple buckets
  - Mid-bucket boundary cursor alignment
  - Floor termination
  - MaxBuckets cap
  - Resume from initial page state
  - Empty walks below floor
  - Fetcher error propagation
  - Ascending direction ordering
  - **Fanout parity**: Byte-identical results at fanout 1 vs. fanout 8 (core safety guarantee)
  - **Actual concurrency**: Proof that buckets are fetched in parallel (with barrier synchronization)

## Implementation Details

[...]

PR #178, opened 2026-08-04
feat: message forwarding — forwardedMessage snapshot on msg.send

cocoamix00:
## Summary

Adds message forwarding: a user forwards an existing message into other rooms via `msg.send` with new `forwardedMessageId` + `forwardedRoomId` fields. The new message carries an optional comment plus an immutable, server-built, text-only `forwardedMessage` snapshot of the source. Design: `docs/superpowers/specs/2026-08-03-message-forwarding-design.md`; plan: `docs/superpowers/plans/2026-08-03-message-forwarding.md`.

Key semantics:
- **Client fan-out**: one `msg.send` per destination room → per-room success/failure via the existing async replies. No new RPC.
- **Authorization for free**: message-gatekeeper fetches the source via the **source room's** `msg.get` subject, so history-service enforces the forwarder's subscription + access window there.
- **Hard-fail**: any source-fetch problem rejects the send (typed error, reply + Ack — never a placeholder, never a Nak loop).
- **Text-only snapshot, chain depth 1**: sources with attachments/cards/system type/deleted are rejected; forwarding a forward captures only its own comment.
- **Snapshot is immutable and never access-window-redacted** (source-room window enforced once, at forward time); body is encrypted at rest (`atrest.ForwardedContent`).

## Changes by area

- `pkg/model` + `pkg/model/cassandra`: `ForwardedMessage` UDT type, request/message wire fields
- Cassandra schema: `"ForwardedMessage"` UDT + `forwarded_message` column in `messages_by_room`, `messages_by_id`, `pinned_messages_by_room` (deliberately **not** `thread_messages_by_thread`) — additive migration `docker-local/cassandra/migrations/2026-08-forwarded-message.cql`, plus all doc/init/test-schema mirrors
- `message-gatekeeper`: forward validation, `FetchForwardedSource` (narrow sonic-safe projection), `resolveForwardSnapshot`
- `message-worker`: persists the snapshot on both main-timeline tables, plaintext + encrypted paths
- `history-service`: snapshot on read/pin/edit paths; pre-window pins redact it; encrypted edits preserve metadata and re-bundle the body
- `notification-worker`: `forwarded` flag on the push payload
- `pkg/atrest`: snapshot body joins the encrypted bundle
- Docs: `docs/client-api.md` + derived views (`request-reply.md`, `events.md`), `docs/cassandra_message_model.md`

## Source-room context on the snapshot

A client rendering a forward needs to know what kind of room it came from. The snapshot carries that inline, captured **once at forward time** — there is no read-time resolution anywhere:

- **`roomType`** (`channel` / `dm` / `botDM` / `discussion`), plus **`tshow`** and **`threadRoomId`** mirroring the source's own thread context. All three are immutable — a room's type never changes after creation, and a message's thread context is fixed at its send time — so they snapshot cleanly.
- **Identical everywhere**: because they ride the canonical message, the `msg.send` echo, broadcast events and every history read return the same object. (An earlier iteration of this branch resolved a `room` object per read, which set it *only* on history reads and left clients handling two shapes for one field.)
- **Deliberately no room name.** A name is mutable, so it cannot be frozen; resolving it per read would also leak a `dm` counterpart's name to readers outside the source room. Clients that display a name resolve it themselves from `roomId`. **Consequence worth weighing:** for a source room the reader is not a member of, the client has no name source — if "Forwarded from #prj-alpha" is required, a server-side lookup has to come back.
- **Best-effort**: the source-room meta lookup failing (or the site having no doc for that room) omits `roomType` and still delivers the forward — the forward is already fully validated by that point, so this cannot reject a valid send.

## Test plan

- TDD throughout; unit suites green with `-race` (`make test`, 106 packages), `make lint` 0 issues, gosec clean
- Coverage: `resolveForwardSnapshot` 95.5%, `FetchForwardedSource` 85.7%
- **Verified in CI** (could not run in the dev sandbox): all 7 `test-integration` suites pass, including the ones exercising the changed UDT and its four inline DDL mirrors — `history-service`, `message-worker`, `message-gatekeeper`, `pkg/atrest`, `pkg/model`, `notification-worker`, `tools`. The `sast` job (govulncheck + semgrep registry rulesets, both egress-blocked locally) also passes.

## Notes for reviewers

- **Stacked commit disclosure**: `69d22aa` (`perf(history-service): single-query room-preview reads (Phase 1) (#173)`) was present in this branch's base state and is not on `main` as of branch creation. It is not part of the forwarding work; if #173 already landed upstream it will dedup on merge, otherwise it merges alongside this PR.
- **`main` merged in**: `main` was merged into this branch to resolve a conflict in `history-service/internal/readcache/readcache.go`, where upstream's batched `GetRoomTimesByIDs` landed at the same insertion point as a method this branch then added. That method has since been removed, so only upstream's remains and the conflict is moot.
- **Deploy order (multi-site)**: apply the Cassandra migration and deploy history-service, message-worker, notification-worker on **every** site before deploying message-gatekeeper to **any** site — old workers silently drop the unknown `forwardedMessage` field, which would cause cross-site divergence for federated forwards. Documented in the spec's Deploy order section.
- **Open follow-ups, filed not fixed** (both pre-existing, both need a product call rather than a unilateral change): #210 — `forwardedContent` lets a forwarder substitute arbitrary text under the source author's identity; #211 — `atrest.ApplyDecryptedFields` would fabricate an incomplete snapshot if it ever saw an encrypted body with nil metadata (traced as unreachable under current write-path invariants).

## Summary by CodeRabbit

- **New Features**
  - Added message forwarding across rooms with optional replacement text.
  - Forwarded messages retain source details, sender, timestamps, mentions, and thread context.
  - Forwarding supports empty comments and appears in message history, pinned messages, events, and send responses.
  - Added forwarded-message indicators to push notifications.

- **Security & Validation**
  - Forwarding validates access, source availability, supported message types, and content limits.
  - Forwarded text is protected by at-rest encryption.

- **Documentation**
  - Updated API and data-model documentation with forwarding fields, restrictions, and error behavior.


cocoamix00:
## Review triage — all 14 open review comments

Each finding was re-verified against current code before acting (several predate later commits on this branch). Summary: **6 fixed**, **6 not actionable**, **1 filed separately**.

### Fixed in 5f81427a

| # | Finding | Change |
|---|---|---|
| 1 | `content must not be empty` error not qualified for forwards | Qualified in the canonical error table and the derived `request-reply.md` bullet |
| 3 | Error text said IDs "must be a 20-char base62 string" | `idgen.IsValidMessageID` accepts **17 or 20** (17 = legacy). Reworded all four sites (message/thread-parent/quoted-parent/forwarded) — text only, validation behavior unchanged |
| 4 | Plan doc used a hard-coded absolute scratchpad path | Replaced with a portable `coverage.out`. The raw `go test -coverprofile` usage itself stays — the Makefile genuinely has no coverage target, which the doc already self-justifies |
| 5 | "Never redacted" claim vs. pin redaction | **Doc fixed, code unchanged** — see below |
| 6 | Architecture-diagram fence missing a language (MD040) | Now ` ```text ` |
| 14 | `TestMessage_ForwardedMessage_JSON` didn't use the shared helper | Rewritten to round-trip a fully populated snapshot (incl. `mentions`, thread metadata, `room`) via `roundTrip` |

On **#5** specifically — the code is correct and stays as-is. `redactUnavailablePins` nil'ing `ForwardedMessage` is about the **destination** room's access window on the pin itself, not the source room's window (which is what "never redacted" refers to). It's deliberate and security-*positive*: without it a restricted caller could read a forward's body through a pin placeholder that is otherwise stubbed. `pin_internal_test.go` documents exactly this. Only the doc sentence was too absolute; it now carries the caveat.

### Not actionable

| # | Finding | Why |
|---|---|---|
| 2 | Broken `#attachment` link | Anchor resolves — the single `#### Attachment` heading matches all seven references. The cited line is inside an unrelated JSON example |
| 9 | "Reject every forward-of-forward" | Contradicts the design. Chain depth 1 is intentional: forwarding a forward captures that message's **own** comment; rejection happens only when the source forward has no forwardable content. Code, tests, and spec agree |
| 10 | Reject `TShow` on forwards | `tshow := req.TShow && req.ThreadParentMessageID != ""` is pre-existing generic normalization for **any** non-thread send, and `client-api.md` documents it as "ignored and normalized to `false` — the request is not rejected". Not forward-specific |
| 11 | Forwarded thread replies lose the snapshot | Unreachable. The gatekeeper rejects forward + thread-parent together, and message-worker only takes the thread path when `ThreadParentMessageID != ""`, so `SaveThreadMessage` can never see a non-nil `ForwardedMessage`. Noted as latent should that exclusion ever be relaxed |
| 12 | Add bson tags to `ForwardedMessage` | The `pkg/model/cassandra` package doc states: *"No bson tags — these types never reach MongoDB."* Every struct in the package follows this |
| 13 | Add bson tags to the request fields | `SendMessageRequest` is a pure wire DTO — **no** field on it carries bson tags. The two new fields match their neighbors |

Two severity corrections worth recording. The **`ForwardedRoomID` validation** finding is tagged SECURITY/Major but is **not exploitable**: `{roomID}` is a single-token NATS wildcard, so a malformed value breaks token alignment into a timeout (the documented hard-fail), not misrouting; and history-service independently enforces `getAccessSince` on that exact room plus a `msg.RoomID != roomID` check before returning anything, while the persisted `ForwardedMessage.RoomID` comes from the server-authoritative reply. It's a missed fail-fast producing a worse error message — real, but code quality, not a vulnerability. And **#5** above is labeled Major in the opposite direction from reality.

### Filed separately — #209

The **pin at-rest encryption** finding is real, reachable, and more serious than described — but it predates this PR. `insertPinnedMsg` has no `enc_payload`/`enc_meta` columns at all, so with `ATREST_ENABLED=true` pinning **any** message writes its decrypted body to `pinned_messages_by_room` in plaintext — `msg`, `attachments`, `card`, `card_action`, `quoted_parent_message`, and now `forwarded_message`. Introduced in `971f949e`; this PR only appended one more field to an already-broken insert. Tracked in #209 with repro conditions, the traced path, and the suggested `PinMessageEncrypted` fix, since the remedy touches the whole pin write path rather than anything forwarding-specific.

### Verification

`make test` (race, 106 packages) exit 0 / 0 failures · `make lint` 0 issues · mocks drift-free · gosec clean. CI additionally confirmed `test-integration (history-service)` green, which exercises the new batched room lookup against real containers.

---

cocoamix00:
## CI `test` failure — pre-existing flake on `main`, fixed in 9e3be9ba

The `test` job went red on 5f81427a with a **panic**, not an assertion failure:

```
panic: close of closed channel
  cassrepo.(*fakeCassandra).fetch      walker_test.go:144
  cassrepo.(*bucketWalk[...]).fetchWave.func1   walker.go:332
  errgroup.(*Group).Go.func1
FAIL  github.com/hmchangw/chat/history-service/internal/cassrepo
```

**Not caused by this branch.** `walker.go` / `walker_test.go` are untouched by the forwarding and enrichment work — they arrived here via the `main` merge (from #169, concurrent bucket fetching). Verified by reproducing on a pristine `origin/main` worktree: same panic, same line, within 200 `-race` runs.

### Root cause

The test double's barrier used check-then-act:

```go
select {
case <-gate:
default:
    close(gate)   // ← two goroutines can both land here
}
```

`reachedBarrier` is computed under the mutex, but the `select` is evaluated *outside* it and is not atomic with the `close`. When two fan-out fetchers hit the barrier together, both observe the channel still open, both take `default`, and the second `close` panics. It only needs the interleaving to land — hence green locally, red on a contended CI runner.

The production code (`walker.go`) is fine; this is entirely the fake.

### Fix

`sync.Once` around the close — the one construct that makes check-and-close atomic.

### Verification

| Check | Before | After |
|---|---|---|
| `-race -count=500` on the flaky test | panics (< 200 runs, both here and on `main`) | 500/500 pass |
| `-race -count=30` on the whole `cassrepo` package | — | pass |
| `make test` (race, full repo) | — | exit 0, 106 packages, 0 failures |
| `make lint` | — | 0 issues |

Flagging it rather than folding it in silently: an unrelated test file in this diff is otherwise unexplained. If you'd prefer it as a standalone PR against `main` — it fixes the flake for everyone, not just this branch — say so and I'll split it out.

---

cocoamix00:
## Second review round — 11 findings, triaged

The full (non-draft) CodeRabbit pass surfaced 11 actionable comments. Each verified against current code before acting.

### Fixed in `f677f791`

| Finding | Fix |
|---|---|
| `GetThreadMessages`' two early-return branches (`ThreadRoomID==""`, `TCount==0`) returned an un-enriched parent | Both now call `enrichForwardedRoom` before returning; added tests for both branches |
| `enrich_forward_test.go` only covered the `Limit:5` surrounding-message path, missing the `Limit:1` single-message fast path | Added a limit-1 test asserting enrichment on the spliced central message |
| No boundary tests at exactly `maxContentBytes` for `content`/`forwardedContent` (only the oversized case existed) | Added exact-boundary success + oversized-`content` failure cases, asserting the `maxContentBytes`/`attempted` error metadata |
| Spec doc's "Mirrors updated" list named a `docker-compose.yml` path that doesn't exist | Replaced with the four actual Go test files carrying inline DDL (`cassrepo`, `history-service/internal/service`, `message-worker`, `tools/loadgen` integration tests) |

### Verified as false positives (not fixed — reasoning below)

- **"MessageLink contract unresolved" in the design spec** — checked the exact cited code block: the spec's `ForwardedMessage` type literally has no `MessageLink` field, consistent with the shipped type and with `docs/client-api.md`. The reviewer is diffing the spec against the **plan** doc, which describes an earlier draft (before commit `fea89cd2`, "drop messageLink from the ForwardedMessage snapshot," already on this branch). The spec was never wrong; only the historical plan doc is stale, and plan docs aren't living contracts under CLAUDE.md.
- **"Clone before encrypted write" / mutation risk** — checked `buildCassandraMessage` directly: `cm.ForwardedMessage = &f` where `f := *msg.ForwardedMessage` is already a value-copy dereference, identical to the existing `QuotedParentMessage` clone three lines above (with a doc comment explicitly calling out why). `StripEncryptedFields` mutates the copy, never the caller's original. Verified safe by inspection, not by trusting the comment.
- **"idgen 20-char" wording, MD040 fence, nested markdown fences (plan doc)** — these three are re-surfaced instances of findings #3/#6 already fixed in the first round (commits `5f81427a`/`9e3be9ba`) or reference the same historical plan doc's already-superseded content; no new code-facing issue.

### Filed separately (real, pre-existing, out of this PR's scope)

- **#210** — `forwardedContent` lets a forwarder substitute arbitrary text under the source author's real `Sender` identity. Real, bounded (requires source-room access already), and already documented as an intentional tradeoff (client-side redaction) predating this branch — but a reviewer independently flagging it as Major security means it deserves an explicit product decision rather than resting on an existing code comment. Traced the full path and laid out four options; not resolving unilaterally since it changes shipped, documented behavior.
- **#211** — `atrest.ApplyDecryptedFields` fabricates a bare `ForwardedMessage{Msg: ...}` (no ID/Sender/CreatedAt) if it ever finds an encrypted body with nil metadata. Traced reachability: unreachable under the current write-path invariant (`SplitForEncryption`/`StripEncryptedFields` never produce that combination from a row this codebase wrote) — live defensive code with no current trigger, but a genuine data-integrity gap if that invariant is ever broken by corruption or a future change. Pre-existing (`249309d6`), not touched by this branch.

### Verification

`make test` (race, full repo) exit 0 / 106 packages / 0 failures · `make lint` 0 issues · new subtests individually confirmed running and passing (`go test -v -run ...`).

---

cocoamix00:
## Review round on `cc9e5303` — 1 finding, fixed

Fixed in `80b22830`: the superseded-notice banner I added left a doubled blank line, and in the plan doc that sat between two adjacent blockquotes — markdownlint MD028. Merged the notice and the following agentic-workers note into one contiguous blockquote (worth noting: collapsing to a *single* blank would not have fixed it — MD028 fires on any blank line between adjacent blockquotes), and collapsed the stray blank in the spec doc.

### On the docstring-coverage warning

The one failing pre-merge check is CodeRabbit's own 80% docstring threshold (34.33%). Not acting on it: this is a Go repo whose lint gate is `golangci-lint`, which passes clean, and CLAUDE.md sets no docstring-coverage requirement. Adding comments across unrelated files purely to move a bot's default metric would be churn against the "keep changes minimal and focused" rule. For what it's worth this branch moved the number *up* (31.88% → 34.33%).

### CI on the schema change

The refactor changed the `ForwardedMessage` Cassandra UDT and its four inline DDL mirrors, and I could not run integration tests locally (no Docker in this environment) — so CI was the first real check. All six suites that exercise the UDT passed on `cc9e5303`:

`history-service` ✅ · `message-worker` ✅ · `message-gatekeeper` ✅ · `pkg/atrest` ✅ · `pkg/model` ✅ · `notification-worker` ✅ · `lint` ✅

Unit tests (106 packages), `make lint` (0 issues) and `gosec` were green locally beforehand. `govulncheck` and the semgrep registry are unreachable from my environment (gateway 403 on `vuln.go.dev` and `semgrep.dev`), so the `sast` job in CI is the authoritative run for those two; I did run the repo's local `.semgrep` rules offline — one finding, in `tools/loadgen/soak_rpc.go`, pre-existing from #127 and untouched by this branch.

---

history-service/internal/cassrepo/pin.go
```diff
@@ -23,15 +23,15 @@ const (
...
 		created_at, tshow, thread_parent_id, thread_parent_created_at
-	) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`
+	) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`
 
 	deletePinnedRow = `DELETE FROM pinned_messages_by_room WHERE room_id = ? AND pinned_at = ? AND message_id = ?`
 
 	pinnedColumns = "room_id, pinned_at, message_id, sender, " +
-		"msg, mentions, attachments, card, card_action, quoted_parent_message, " +
+		"msg, mentions, attachments, card, card_action, quoted_parent_message, forwarded_message, " +
 		"visible_to, deleted, type, sys_msg_data, site_id, edited_at, " +
```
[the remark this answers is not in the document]
mliu33: Good catch...we are encrypting in "edit pin message", but did not encrypt when inserting into pin_messages table. Please help fix this discrepancy in a separate PR, thanks

julianshen:
**9 findings · 8 inline comments posted** — 🔴 2 blocking · 🟠 7 warning

### 💬 Additional comments (1)


**`docs/superpowers/specs/2026-08-03-message-forwarding-design.md`**

_🏷️ rolling-deploy_ | _🔴 Blocking_ | _`distributed-system`_

**An old worker can acknowledge and permanently discard the snapshot.**

The deploy section acknowledges that an old message-worker silently ignores `forwardedMessage`, but relies only on completing steps 1–4 everywhere before deploying gatekeeper and explicitly provides no feature flag. A failed rollout, lagging replica, or rollback leaves a reachable loss path: the new gatekeeper accepts a forward and returns success; JetStream assigns the canonical event to an old worker; JSON decoding ignores the unknown field; that worker persists an ordinary message and acknowledges the event. Broadcast may still show the snapshot, while history at that site permanently lacks it, and no replay repairs the divergence. Add an enforceable rollout/capability gate before accepting forwards, or version the event/consumer so an incompatible worker rejects rather than acknowledges forward events.

### Related work

- [PR #173](https://github.com/hmchangw/newchat/pull/173) — perf\(history-service\): single-query room-preview reads \(Phase 1\) (merged)
- [Issue #210](https://github.com/hmchangw/newchat/issues/210) — forwardedContent lets a forwarder attribute arbitrary text to the source message's author (open)
- [Issue #211](https://github.com/hmchangw/newchat/issues/211) — atrest.ApplyDecryptedFields fabricates a bare ForwardedMessage if metadata is ever missing (open)

### Local review memory

- `docs/client-api.md:5963` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟡 Minor_ _⚡ Quick win_ **Fix the broken Attachment link.** markdownlint-cli2 reports MD051 for the #attachment fragment at Line 5798. Replace it with the exact anchor generated by the Attach… ([thread](https://github.com/hmchangw/newchat/pull/178#discussion_r3708837654))
- `docs/superpowers/plans/2026-08-03-message-forwarding.md:914` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟠 Major_ _⚡ Quick win_ **Keep pin redaction consistent with the immutable snapshot contract.** The plan removes ForwardedMessage from pins even though the canonical API says forwarded snapshots ar… ([thread](https://github.com/hmchangw/newchat/pull/178#discussion_r3708837666))
- `message-gatekeeper/handler.go:287` — @coderabbitai: _🎯 Functional Correctness_ _🟡 Minor_ _⚡ Quick win_ **Reject TShow on forwarding requests.** Lines 273-275 reject only ThreadParentMessageID . A request with ForwardedMessageID , ForwardedRoomID , and TShow: true succeeds, then… ([thread](https://github.com/hmchangw/newchat/pull/178#discussion_r3708837674))
- `pkg/model/cassandra/message.go:106` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟠 Major_ _⚡ Quick win_ **Add BSON tags to ForwardedMessage .** ForwardedMessage has JSON and CQL tags only. Add camelCase BSON tags for every field. As per coding guidelines, “All model structs… ([thread](https://github.com/hmchangw/newchat/pull/178#discussion_r3708837685))
- `pkg/model/message.go:85` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟠 Major_ _⚡ Quick win_ **Add BSON tags to the new request fields.** Lines 94-95 add fields to a pkg/model struct with JSON tags only. Add matching BSON tags, or move this JSON-only DTO outside th… ([thread](https://github.com/hmchangw/newchat/pull/178#discussion_r3708837687))
- `pkg/model/model_test.go:2704` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟠 Major_ _⚡ Quick win_ **Use roundTrip with a fully populated forwarded snapshot.** This test checks only MessageID and Msg . It does not verify RoomID , Sender , CreatedAt , Mentions , M… ([thread](https://github.com/hmchangw/newchat/pull/178#discussion_r3708837695))
- `docs/superpowers/specs/2026-08-03-message-forwarding-design.md:204` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟠 Major_ _🏗️ Heavy lift_ **Require a cloned snapshot before encrypted writes.** The specification says buildCassandraMessage assigns msg.ForwardedMessage directly, but encrypted persistence bla… ([thread](https://github.com/hmchangw/newchat/pull/178#discussion_r3741057346))
- `pkg/model/message.go:96` — @coderabbitai: _🔒 Security & Privacy_ _🟠 Major_ _🏗️ Heavy lift_ **Do not permit source-body impersonation.** ForwardedContent is client-controlled, but the gatekeeper persists it as ForwardedMessage.Msg under the source sender. A user who can… ([thread](https://github.com/hmchangw/newchat/pull/178#discussion_r3741057355))
- `docs/superpowers/plans/2026-08-03-message-forwarding.md:146` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟡 Minor_ _⚡ Quick win_ **Repair the nested Markdown fences.** The markdown examples contain nested triple-backtick fences. The inner fence closes the outer block, which makes the rendered plan ma… ([thread](https://github.com/hmchangw/newchat/pull/178#discussion_r3741057332))
- `docs/superpowers/plans/2026-08-03-message-forwarding.md:215` — @coderabbitai: _🔒 Security & Privacy_ _🟠 Major_ _🏗️ Heavy lift_ **Implement the documented forwardedContent override.** The design specifies an optional forwardedContent value that replaces ForwardedMessage.Msg after source validation. This… ([thread](https://github.com/hmchangw/newchat/pull/178#discussion_r3741057334))

<details>
<summary>📒 Files reviewed (48)</summary>

* `docker-local/cassandra/init/09-udt-forwarded_message.cql`
* `docker-local/cassandra/init/10-table-messages_by_room.cql`
* `docker-local/cassandra/init/12-table-pinned_messages_by_room.cql`
* `docker-local/cassandra/init/13-table-messages_by_id.cql`
* `docker-local/cassandra/migrations/2026-08-forwarded-message.cql`
* `docs/cassandra_message_model.md`
* `docs/client-api.md`
* `docs/client-api/events.md`
* `docs/superpowers/plans/2026-08-03-message-forwarding.md`
* `docs/superpowers/specs/2026-08-03-message-forwarding-design.md`
* `history-service/internal/cassrepo/integration_test.go`
* `history-service/internal/cassrepo/messages_by_id_integration_test.go`
* `history-service/internal/cassrepo/messages_by_room.go`
* `history-service/internal/cassrepo/pin.go`
* `history-service/internal/cassrepo/pin_integration_test.go`
* `history-service/internal/cassrepo/walker_test.go`
* `history-service/internal/cassrepo/write.go`
* `history-service/internal/cassrepo/write_integration_test.go`
* `history-service/internal/service/integration_test.go`
* `history-service/internal/service/pin.go`
* `history-service/internal/service/pin_internal_test.go`
* `history-service/internal/service/reactions.go`
* `message-gatekeeper/fetcher_history_test.go`
* `message-gatekeeper/handler_test.go`
* `message-gatekeeper/helper.go`
* `message-gatekeeper/mock_store_test.go`
* `message-gatekeeper/sonic_wire_test.go`
* `message-gatekeeper/store.go`
* `message-worker/store_cassandra.go`
* `pkg/atrest/atrest.go`
* `pkg/atrest/split.go`
* `pkg/atrest/split_test.go`
* `pkg/model/cassandra/message.go`
* `pkg/model/cassandra/message_test.go`
* `pkg/model/message.go`
* `pkg/model/model_test.go`
* `pkg/model/push.go`
* `pkg/model/room.go`
* `pkg/model/search.go`
* `tools/loadgen/history_integration_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

history-service/internal/cassrepo/pin.go
```diff
@@ -67,7 +67,7 @@ func (r *Repository) PinMessage(ctx context.Context, msg *models.Message, pinned
 	batch.Query(pinMsgByID, pinnedAt, pinnedBy, msg.MessageID)
 	batch.Query(insertPinnedMsg,
 		msg.RoomID, pinnedAt, msg.MessageID, msg.Sender, msg.Msg, msg.Mentions,
-		msg.Attachments, msg.Card, msg.CardAction, msg.QuotedParentMessage, msg.VisibleTo,
+		msg.Attachments, msg.Card, msg.CardAction, msg.QuotedParentMessage, msg.ForwardedMessage, msg.VisibleTo,
```
julianshen, line 70: _🔒 security_ | _🔴 Blocking_ | _`tgd-review`_

**Pinning writes the decrypted forwarded body back to plaintext.**

The pin insert binds `msg.ForwardedMessage` directly while `insertPinnedMsg` stores neither `enc_payload` nor `enc_meta`. In a cipher-enabled deployment, history reads restore `ForwardedMessage.Msg` before the service pins the message, so this copies the sensitive body into the plaintext UDT, contradicting the documented at-rest contract that forwarded bodies live only in `enc_payload`. If the caller instead supplies the stripped UDT, the pinned copy loses its body. Add a cipher-aware pinned write/read path that stores a body-blanked UDT plus encrypted payload and metadata, and cover encrypted pin round-tripping with an integration test.

<details>

```
In `history-service/internal/cassrepo/pin.go` around line 70: The pin insert binds `msg.ForwardedMessage` directly while `insertPinnedMsg` stores neither `enc_payload` nor `enc_meta`. In a cipher-enabled deployment, history reads restore `ForwardedMessage.Msg` before the service pins the message, so this copies the sensitive body into the plaintext UDT, contradicting the documented at-rest contract that forwarded bodies live only in `enc_payload`. If the caller instead supplies the stripped UDT, the pinned copy loses its body. Add a cipher-aware pinned write/read path that stores a body-blanked UDT plus encrypted payload and metadata, and cover encrypted pin round-tripping with an integration test.

```

</details>

message-gatekeeper/handler.go
```diff
@@ -494,6 +524,87 @@ func (h *Handler) resolveQuoteSnapshot(ctx context.Context, account, roomID, sit
...
+			"request_id", req.RequestID, "forwarded_id", req.ForwardedMessageID, "error", err)
+		return nil, errcode.Unavailable("forward source temporarily unavailable")
+	}
+	if src.Deleted {
+		return nil, errcode.BadRequest("cannot forward a deleted message")
+	}
+	if model.IsSystemMessageType(src.Type) {
+		return nil, errcode.BadRequest("cannot forward a system message")
+	}
+	if len(src.Attachments) > 0 {
```
julianshen, line 559: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Card-action sources bypass the text-only forwarding check.**

The source projection and rejection logic inspect `card` but not the independent `cardAction` field. `model.Message` supports `CardAction`, and message-worker persists it separately from `Card`, so a message with only card-action content can be accepted as forwardable and have that content silently dropped from the snapshot. Add `CardAction json.RawMessage` to `forwardSourceProjection`, reject it alongside `Card`, and add a handler test for a card-action-only source.

<details>

```
In `message-gatekeeper/handler.go` around line 559: The source projection and rejection logic inspect `card` but not the independent `cardAction` field. `model.Message` supports `CardAction`, and message-worker persists it separately from `Card`, so a message with only card-action content can be accepted as forwardable and have that content silently dropped from the snapshot. Add `CardAction json.RawMessage` to `forwardSourceProjection`, reject it alongside `Card`, and add a handler test for a card-action-only source.

```

</details>

history-service/internal/cassrepo/pin.go
julianshen, line 70: _🔒 security_ | _🔴 Blocking_ | _`cassandra`_

**Pinning writes the forwarded body back to Cassandra in plaintext.**

Encrypted main-table rows store ForwardedMessage.Msg in enc_payload and leave forwarded_message.msg blank, but PinMessage binds the hydrated msg.ForwardedMessage directly into pinned_messages_by_room while insertPinnedMsg omits enc_payload and enc_meta. After GetMessageByID decrypts a forward, pinning it therefore persists the source body in plaintext, contrary to the at-rest encryption contract. On the encrypted path, insert a stripped forwarded_message UDT together with the existing ciphertext/meta, or re-encrypt the pin and update the pinned read path accordingly.

<details>

```
In `history-service/internal/cassrepo/pin.go` around line 70: Encrypted main-table rows store ForwardedMessage.Msg in enc_payload and leave forwarded_message.msg blank, but PinMessage binds the hydrated msg.ForwardedMessage directly into pinned_messages_by_room while insertPinnedMsg omits enc_payload and enc_meta. After GetMessageByID decrypts a forward, pinning it therefore persists the source body in plaintext, contrary to the at-rest encryption contract. On the encrypted path, insert a stripped forwarded_message UDT together with the existing ciphertext/meta, or re-encrypt the pin and update the pinned read path accordingly.

```

</details>

docker-local/cassandra/migrations/2026-08-forwarded-message.cql
```diff
@@ -0,0 +1,30 @@
...
+  msg TEXT,
+  room_id TEXT,
+  room_type TEXT,
+  sender FROZEN<"Participant">,
+  thread_parent_created_at TIMESTAMP,
+  thread_parent_id TEXT,
+  thread_room_id TEXT,
+  tshow BOOLEAN
+);
+ALTER TABLE chat.messages_by_room        ADD forwarded_message FROZEN<"ForwardedMessage">;
```
julianshen, line 28: _🏷️ schema-migration_ | _🔴 Blocking_ | _`distributed-system`_

**The schema migration is not safely resumable after partial success.**

The three ALTER statements are independent and non-transactional, but none is idempotent. If the first ALTER commits and the connection fails before the others, a stop-on-error replay fails on the existing first column and can leave the other tables unmodified. The same convergence problem exists for the explicitly acknowledged older UDT shape because CREATE TYPE IF NOT EXISTS silently retains it while the required ALTER TYPE operations are only comments. New history-service queries reference forwarded_message unconditionally, and message-worker inserts it, so a partially upgraded site can fail reads or repeatedly reject deliveries while other sites succeed; an older UDT can also silently omit the new snapshot fields. Make every schema mutation independently resumable using IF NOT EXISTS where supported or separately tracked reconciliation steps, and verify system_schema plus schema agreement at every site before enabling forwarding.

<details>

```
In `docker-local/cassandra/migrations/2026-08-forwarded-message.cql` around line 28: The three ALTER statements are independent and non-transactional, but none is idempotent. If the first ALTER commits and the connection fails before the others, a stop-on-error replay fails on the existing first column and can leave the other tables unmodified. The same convergence problem exists for the explicitly acknowledged older UDT shape because CREATE TYPE IF NOT EXISTS silently retains it while the required ALTER TYPE operations are only comments. New history-service queries reference forwarded_message unconditionally, and message-worker inserts it, so a partially upgraded site can fail reads or repeatedly reject deliveries while other sites succeed; an older UDT can also silently omit the new snapshot fields. Make every schema mutation independently resumable using IF NOT EXISTS where supported or separately tracked reconciliation steps, and verify system_schema plus schema agreement at every site before enabling forwarding.

```

</details>

history-service/internal/cassrepo/pin.go
julianshen, line 70: _🔒 security_ | _🔴 Blocking_ | _`tgd-review`_

**Pinning writes the forwarded body outside the encrypted payload.**

`PinMessage` binds the hydrated `msg.ForwardedMessage` directly into `forwarded_message`, while `insertPinnedMsg` writes neither `enc_payload` nor `enc_meta`. On cipher-enabled reads, `ForwardedMessage.Msg` has already been restored from the encrypted bundle, so pinning persists that body in plaintext. The new integration test further codifies the plaintext round-trip. Make pin creation encryption-aware: move the forwarded body into `enc_payload`, strip it from the UDT before binding, and add an encrypted pin test asserting the raw UDT body is empty.

<details>

```
In `history-service/internal/cassrepo/pin.go` around line 70: `PinMessage` binds the hydrated `msg.ForwardedMessage` directly into `forwarded_message`, while `insertPinnedMsg` writes neither `enc_payload` nor `enc_meta`. On cipher-enabled reads, `ForwardedMessage.Msg` has already been restored from the encrypted bundle, so pinning persists that body in plaintext. The new integration test further codifies the plaintext round-trip. Make pin creation encryption-aware: move the forwarded body into `enc_payload`, strip it from the UDT before binding, and add an encrypted pin test asserting the raw UDT body is empty.

```

</details>

message-worker/store_cassandra.go
```diff
@@ -87,21 +87,21 @@ func (s *CassandraStore) SaveMessage(ctx context.Context, msg *model.Message, se
 	batch.Query(
 		`INSERT INTO messages_by_room
 		   (room_id, bucket, created_at, message_id, sender, msg, site_id, updated_at,
-		    mentions, type, sys_msg_data, tshow, quoted_parent_message,
+		    mentions, type, sys_msg_data, tshow, quoted_parent_message, forwarded_message,
```
julianshen, line 90: _🏷️ cassandra-batches_ | _🟠 Warning_ | _`cassandra`_

**Max-size forwards can exceed Cassandra's batch failure threshold.**

SaveMessage duplicates both the comment and forwarded snapshot into messages_by_room and messages_by_id within one Cassandra batch. A valid request can contain a 20,480-byte comment plus a 20,480-byte forwarded body, producing more than 80 KiB across the two mutations before UDT metadata and protocol overhead; the encrypted path similarly duplicates an approximately 40 KiB enc_payload. This exceeds Cassandra's default 50 KiB batch_size_fail_threshold_in_kb, so valid forwards can fail with an oversized-batch error. Avoid batching the two large cross-partition mutations, or enforce a combined serialized-size limit safely below the configured batch threshold.

<details>

```
In `message-worker/store_cassandra.go` around line 90: SaveMessage duplicates both the comment and forwarded snapshot into messages_by_room and messages_by_id within one Cassandra batch. A valid request can contain a 20,480-byte comment plus a 20,480-byte forwarded body, producing more than 80 KiB across the two mutations before UDT metadata and protocol overhead; the encrypted path similarly duplicates an approximately 40 KiB enc_payload. This exceeds Cassandra's default 50 KiB batch_size_fail_threshold_in_kb, so valid forwards can fail with an oversized-batch error. Avoid batching the two large cross-partition mutations, or enforce a combined serialized-size limit safely below the configured batch threshold.

```

</details>

docker-local/cassandra/migrations/2026-08-forwarded-message.cql
```diff
@@ -0,0 +1,30 @@
...
+  message_id TEXT,
+  msg TEXT,
+  room_id TEXT,
+  room_type TEXT,
+  sender FROZEN<"Participant">,
+  thread_parent_created_at TIMESTAMP,
+  thread_parent_id TEXT,
+  thread_room_id TEXT,
+  tshow BOOLEAN
+);
```
julianshen, line 27: _🏷️ migration_ | _🟠 Warning_ | _`cassandra`_

**The schema migration cannot recover from partial application.**

Cassandra applies these three ALTER TABLE statements independently. If messages_by_room is altered and execution then stops, rerunning the file fails on that already-existing column and can leave messages_by_id or pinned_messages_by_room without forwarded_message. The newly changed readers and writers then fail with an undefined-column error. Make each addition conditional where supported by the deployed Cassandra version, or have the migration runner inspect system_schema and independently apply only missing columns.

<details>

```
In `docker-local/cassandra/migrations/2026-08-forwarded-message.cql` around line 27: Cassandra applies these three ALTER TABLE statements independently. If messages_by_room is altered and execution then stops, rerunning the file fails on that already-existing column and can leave messages_by_id or pinned_messages_by_room without forwarded_message. The newly changed readers and writers then fail with an undefined-column error. Make each addition conditional where supported by the deployed Cassandra version, or have the migration runner inspect system_schema and independently apply only missing columns.

```

</details>

message-gatekeeper/handler.go
```diff
@@ -239,33 +238,58 @@ func (h *Handler) processMessage(ctx context.Context, account, roomID, siteID st
...
-		return nil, errcode.BadRequest(fmt.Sprintf("invalid quoted parent message ID %q: must be a 20-char base62 string", req.QuotedParentMessageID))
+		return nil, errcode.BadRequest(fmt.Sprintf("invalid quoted parent message ID %q: must be a 17- or 20-char base62 string", req.QuotedParentMessageID))
 	}
 
-	// A message with attachments may carry empty content.
-	if req.Content == "" && len(req.Attachments) == 0 {
+	// Forward validation: both fields together, a valid source ID, and no
+	// combination with thread/quote/attachments — a forward is a main-timeline
+	// message carrying a comment plus a server-built snapshot (see
+	// docs/superpowers/specs/2026-08-03-message-forwarding-design.md).
```
julianshen, line 260: _🏷️ nats-subject_ | _🟠 Warning_ | _`nats`_

**ForwardedRoomID is interpolated into a NATS subject unvalidated.**

The new validation only requires ForwardedRoomID to be non-empty, but FetchForwardedSource passes it to subject.MsgGet as a subject token. A value containing `*` or `>` makes the request publish invalid, while `.` changes token boundaries and can route to an unintended broad subscription or produce no responders. The resulting permanent client-input failure is converted to `unavailable` and the JetStream send is acked, encouraging futile retries. Validate ForwardedRoomID using the application's room-ID rules and explicitly require one legal NATS token before issuing the request; reject violations as bad_request.

<details>

```
In `message-gatekeeper/handler.go` around line 260: The new validation only requires ForwardedRoomID to be non-empty, but FetchForwardedSource passes it to subject.MsgGet as a subject token. A value containing `*` or `>` makes the request publish invalid, while `.` changes token boundaries and can route to an unintended broad subscription or produce no responders. The resulting permanent client-input failure is converted to `unavailable` and the JetStream send is acked, encouraging futile retries. Validate ForwardedRoomID using the application's room-ID rules and explicitly require one legal NATS token before issuing the request; reject violations as bad_request.

```

</details>

message-gatekeeper/fetcher_history.go
```diff
@@ -104,3 +122,41 @@ func (f *historyParentFetcher) FetchQuotedParent(
...
+// unmarshal) is returned — the caller hard-fails the send on any of them.
+func (f *historyParentFetcher) FetchForwardedSource(
+	ctx context.Context,
+	account, srcRoomID, siteID, messageID string,
+) (*forwardSourceProjection, error) {
+	var src forwardSourceProjection
+	if err := f.requestMessageByID(ctx, "forward source", account, srcRoomID, siteID, messageID, &src); err != nil {
+		return nil, err
+	}
+	return &src, nil
```
julianshen, line 161: _🏷️ nats-request-reply_ | _🟠 Warning_ | _`nats`_

**Structurally empty history replies are accepted as successful fetches.**

sonic.Unmarshal accepts valid but structurally malformed replies such as `{}` or `null` into the zero-valued projection. FetchForwardedSource then returns success, and resolveForwardSnapshot can publish and acknowledge a canonical message with an empty source room, sender, and timestamp. It also never verifies that the returned room matches the room used for the authorization RPC. Treat missing required fields or a room mismatch as a transport/protocol failure so the send returns unavailable rather than reporting success.

<details>

> ‼️ **IMPORTANT**

```suggestion
if src.RoomID != srcRoomID || src.CreatedAt.IsZero() {
		return nil, fmt.Errorf("malformed forward source reply for message %s in room %s", messageID, srcRoomID)
	}
	return &src, nil
```

</details>

<details>

```
In `message-gatekeeper/fetcher_history.go` around line 161: sonic.Unmarshal accepts valid but structurally malformed replies such as `{}` or `null` into the zero-valued projection. FetchForwardedSource then returns success, and resolveForwardSnapshot can publish and acknowledge a canonical message with an empty source room, sender, and timestamp. It also never verifies that the returned room matches the room used for the authorization RPC. Treat missing required fields or a room mismatch as a transport/protocol failure so the send returns unavailable rather than reporting success.

```

</details>

docs/client-api.md
```diff
@@ -5915,12 +5958,15 @@ The same subject and request body cover three send variants: plain message, thre
 | Field | Type | Required | Notes |
 |---|---|---|---|
 | `id` | string | yes | The message's ID. Must be 20-char base62. The client generates this and uses it for client-side optimistic rendering. |
-| `content` | string | yes* | The message body, ≤ 20 KiB. *Required unless `attachments` is present — a message with attachments may have empty `content`. |
+| `content` | string | yes* | The message body, ≤ 20 KiB. *Required unless `attachments` is present or the message is a forward (`forwardedMessageId` set) — both may have empty `content`. |
 | `requestId` | string | yes | A 36-char hyphenated UUID (v4 or v7) the client generates. **Validated** — an empty or malformed `requestId` is rejected with no message published. The async reply is delivered to `chat.user.{account}.response.{requestId}`. |
```
julianshen, line 5962: _📝 documentation_ | _🟠 Warning_ | _`tgd-review`_

**The documented message-ID format disagrees with validation.**

The request table still requires a 20-character message ID, and the error table still documents the old 20-character error, while `processMessage` now describes and reports 17- or 20-character IDs for messages, thread parents, and quoted parents. Update the client API and request-reply documentation consistently so clients know which lengths are accepted.

<details>

```
In `docs/client-api.md` around line 5962: The request table still requires a 20-character message ID, and the error table still documents the old 20-character error, while `processMessage` now describes and reports 17- or 20-character IDs for messages, thread parents, and quoted parents. Update the client API and request-reply documentation consistently so clients know which lengths are accepted.

```

</details>

docker-local/cassandra/migrations/2026-08-forwarded-message.cql
```diff
@@ -0,0 +1,30 @@
...
+-- scripts already create new clusters with this column.
+--
+-- If a pre-release cluster already created an earlier ForwardedMessage without
+-- room_type/thread_room_id/tshow, CREATE TYPE IF NOT EXISTS is a no-op there —
+-- add the three columns explicitly instead:
+--   ALTER TYPE chat."ForwardedMessage" ADD room_type TEXT;
+--   ALTER TYPE chat."ForwardedMessage" ADD thread_room_id TEXT;
+--   ALTER TYPE chat."ForwardedMessage" ADD tshow BOOLEAN;
+
+CREATE TYPE IF NOT EXISTS chat."ForwardedMessage"(
```
julianshen, line 15: _🏷️ migration_ | _🔴 Blocking_ | _`silent-failure`_

**Existing older UDTs make the migration silently incomplete.**

The file explicitly acknowledges that `CREATE TYPE IF NOT EXISTS` is a successful no-op when an older `ForwardedMessage` already exists, but the required `ALTER TYPE` operations remain comments. Such a cluster can report successful migration while `room_type`, `thread_room_id`, and `tshow` remain absent, causing newly captured snapshot metadata to be omitted or writes to fail depending on driver behavior. Make migration execution inspect `system_schema` and independently add every missing UDT field, then verify the final shape and schema agreement before enabling forwarding.

<details>

```
In `docker-local/cassandra/migrations/2026-08-forwarded-message.cql` around line 15: The file explicitly acknowledges that `CREATE TYPE IF NOT EXISTS` is a successful no-op when an older `ForwardedMessage` already exists, but the required `ALTER TYPE` operations remain comments. Such a cluster can report successful migration while `room_type`, `thread_room_id`, and `tshow` remain absent, causing newly captured snapshot metadata to be omitted or writes to fail depending on driver behavior. Make migration execution inspect `system_schema` and independently add every missing UDT field, then verify the final shape and schema agreement before enabling forwarding.

```

</details>

message-worker/store_cassandra.go
```diff
@@ -87,21 +87,21 @@ func (s *CassandraStore) SaveMessage(ctx context.Context, msg *model.Message, se
...
 		    attachments, card, card_action)
-		 VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+		 VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
 		msg.RoomID, b, msg.CreatedAt, msg.ID, sender, msg.Content, siteID, msg.CreatedAt,
-		mentions, msg.Type, msg.SysMsgData, msg.TShow, msg.QuotedParentMessage,
+		mentions, msg.Type, msg.SysMsgData, msg.TShow, msg.QuotedParentMessage, msg.ForwardedMessage,
 		msg.Attachments, msg.Card, msg.CardAction,
 	)
```
julianshen, line 96: _🏷️ tombstones_ | _🟠 Warning_ | _`cassandra`_

**Nil forward binds create tombstones for every ordinary message.**

Both SaveMessage INSERTs, and likewise the encrypted variants, always include forwarded_message. For every non-forward message, gocql serializes the nil *ForwardedMessage as CQL NULL, which writes a cell tombstone into both messages_by_room and messages_by_id. Since room-history queries now select this column, a recent high-volume room bucket encounters one additional tombstone per ordinary message until gc_grace expires. PinMessage repeats the pattern for ordinary pins. Use statement variants that omit forwarded_message when absent, or bind Cassandra UNSET when the deployed protocol/driver supports it; apply the same treatment to pinned-row inserts.

<details>

```
In `message-worker/store_cassandra.go` around line 96: Both SaveMessage INSERTs, and likewise the encrypted variants, always include forwarded_message. For every non-forward message, gocql serializes the nil *ForwardedMessage as CQL NULL, which writes a cell tombstone into both messages_by_room and messages_by_id. Since room-history queries now select this column, a recent high-volume room bucket encounters one additional tombstone per ordinary message until gc_grace expires. PinMessage repeats the pattern for ordinary pins. Use statement variants that omit forwarded_message when absent, or bind Cassandra UNSET when the deployed protocol/driver supports it; apply the same treatment to pinned-row inserts.

```

</details>

pkg/model/cassandra/message.go
```diff
@@ -67,6 +67,44 @@ type QuotedParentMessage struct {
...
+type RoomType string
+
+// ForwardedMessage maps to the Cassandra "ForwardedMessage" UDT — the immutable
+// snapshot of a forwarded source message, captured at forward time by
+// message-gatekeeper. Text-only by design: sources carrying attachments or
+// cards are rejected at forward time, so the UDT has no attachment fields.
+// Never redacted on read (self-contained; the source room's access window was
+// enforced once, at forward time).
+type ForwardedMessage struct {
+	MessageID string      `json:"messageId"          cql:"message_id"`
```
julianshen, line 85: _🏷️ mongodb-data-model_ | _🟠 Warning_ | _`mongodb`_

**Forwarded snapshots encode with unintended BSON field names.**

Affected document shape: the nested `forwardedMessage` subdocument when `model.Message` is BSON-marshaled for a MongoDB insert or replacement. The outer field explicitly uses `bson:"forwardedMessage,omitempty"`, but this nested type has no BSON tags, so the MongoDB driver falls back to keys such as `messageid`, `roomid`, `createdat`, and `threadparentid` instead of the established camelCase names. Queries, projections, or updates using paths such as `forwardedMessage.messageId` will therefore miss these values. Add BSON tags matching the JSON names to every field.

<details>

```
In `pkg/model/cassandra/message.go` around line 85: Affected document shape: the nested `forwardedMessage` subdocument when `model.Message` is BSON-marshaled for a MongoDB insert or replacement. The outer field explicitly uses `bson:"forwardedMessage,omitempty"`, but this nested type has no BSON tags, so the MongoDB driver falls back to keys such as `messageid`, `roomid`, `createdat`, and `threadparentid` instead of the established camelCase names. Queries, projections, or updates using paths such as `forwardedMessage.messageId` will therefore miss these values. Add BSON tags matching the JSON names to every field.

```

</details>

message-gatekeeper/handler.go
```diff
@@ -494,6 +524,87 @@ func (h *Handler) resolveQuoteSnapshot(ctx context.Context, account, roomID, sit
...
 
+// resolveForwardSnapshot fetches the forward source (from the SOURCE room's
+// msg.get subject — history-service enforces the forwarder's subscription and
+// access window there) and projects it into the immutable snapshot. Hard-fail
+// policy: ANY failure rejects the send — no placeholder, no re-projection (a
+// forward without its content is meaningless). Transport failures collapse to
+// a typed unavailable error so the JetStream msg is replied+acked, never
+// Nak-looped on a dead history.
+func (h *Handler) resolveForwardSnapshot(ctx context.Context, account, siteID string, req *model.SendMessageRequest) (*cassandra.ForwardedMessage, error) {
+	if req.ForwardedMessageID == "" {
```
julianshen, line 535: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Multi-room forwards stampede the same source lookup.**

Forwarding one source into N rooms creates N concurrent `processMessage` calls by design. Each performs the identical `FetchForwardedSource(account, sourceRoom, site, messageID)` RPC and source-room metadata lookup, with no coalescing. Large destination selections or retries therefore generate N history-service requests—and corresponding Cassandra reads—for one source, consuming gatekeeper/history capacity and increasing tail latency. Coalesce concurrent authenticated source resolution with bounded singleflight keyed by account, site, source room, and message ID; allow each waiter to cancel independently and reuse the immutable room type.

<details>

```
In `message-gatekeeper/handler.go` around line 535: Forwarding one source into N rooms creates N concurrent `processMessage` calls by design. Each performs the identical `FetchForwardedSource(account, sourceRoom, site, messageID)` RPC and source-room metadata lookup, with no coalescing. Large destination selections or retries therefore generate N history-service requests—and corresponding Cassandra reads—for one source, consuming gatekeeper/history capacity and increasing tail latency. Coalesce concurrent authenticated source resolution with bounded singleflight keyed by account, site, source room, and message ID; allow each waiter to cancel independently and reuse the immutable room type.

```

</details>

pkg/atrest/split.go
```diff
@@ -56,4 +62,10 @@ func ApplyDecryptedFields(msg *cassandra.Message, enc *EncryptedFields) {
 		msg.QuotedParentMessage.Msg = enc.QuotedParentContent.Msg
 		msg.QuotedParentMessage.Attachments = enc.QuotedParentContent.Attachments
 	}
+	if enc.ForwardedContent != nil {
```
julianshen, line 65: _🏷️ error-handling_ | _🟠 Warning_ | _`silent-failure`_

**Missing snapshot metadata is masked by a body-only forward.**

When the encrypted bundle contains `ForwardedContent` but the Cassandra UDT is missing, `ApplyDecryptedFields` fabricates a non-nil `ForwardedMessage` containing only `Msg`. History can consequently return an apparently valid forward with no message ID, room, sender, or timestamp, hiding a partial write or corrupt row. Do not synthesize the snapshot; make the decryption/read path surface a corruption error or explicitly omit the forward with telemetry, and replace the test that currently requires body-only construction.

<details>

```
In `pkg/atrest/split.go` around line 65: When the encrypted bundle contains `ForwardedContent` but the Cassandra UDT is missing, `ApplyDecryptedFields` fabricates a non-nil `ForwardedMessage` containing only `Msg`. History can consequently return an apparently valid forward with no message ID, room, sender, or timestamp, hiding a partial write or corrupt row. Do not synthesize the snapshot; make the decryption/read path surface a corruption error or explicitly omit the forward with telemetry, and replace the test that currently requires body-only construction.

```

</details>

message-gatekeeper/handler.go
```diff
@@ -494,6 +524,87 @@ func (h *Handler) resolveQuoteSnapshot(ctx context.Context, account, roomID, sit
...
+		slog.WarnContext(ctx, "forward source fetch failed",
+			"request_id", req.RequestID, "forwarded_id", req.ForwardedMessageID, "error", err)
+		return nil, errcode.Unavailable("forward source temporarily unavailable")
+	}
+	if src.Deleted {
+		return nil, errcode.BadRequest("cannot forward a deleted message")
+	}
+	if model.IsSystemMessageType(src.Type) {
+		return nil, errcode.BadRequest("cannot forward a system message")
+	}
```
julianshen, line 558: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Card-action-only sources bypass the text-only forwarding check.**

The projection and rejection logic inspect Card but not the independently persisted CardAction field. A source containing only card-action content can therefore be accepted and have that content silently dropped from the forwarded snapshot. Project cardAction as a presence-only field, reject it alongside Card, and add a handler test.

<details>

```
In `message-gatekeeper/handler.go` around line 558: The projection and rejection logic inspect Card but not the independently persisted CardAction field. A source containing only card-action content can therefore be accepted and have that content silently dropped from the forwarded snapshot. Project cardAction as a presence-only field, reject it alongside Card, and add a handler test.

```

</details>

julianshen (commented):
tGD inline review

message-gatekeeper/handler.go
```diff
@@ -239,33 +238,58 @@ func (h *Handler) processMessage(ctx context.Context, account, roomID, siteID st
...
+		if req.ForwardedMessageID == "" || req.ForwardedRoomID == "" {
+			return nil, errcode.BadRequest("forwardedMessageId and forwardedRoomId must be set together")
+		}
+		if !idgen.IsValidMessageID(req.ForwardedMessageID) {
+			return nil, errcode.BadRequest(fmt.Sprintf("invalid forwarded message ID %q: must be a 17- or 20-char base62 string", req.ForwardedMessageID))
+		}
+		if req.QuotedParentMessageID != "" {
+			return nil, errcode.BadRequest("a forward cannot also quote a message")
+		}
+		if req.ThreadParentMessageID != "" {
```
julianshen, line 272: _🏷️ intent-match_ | _🟠 Warning_ | _`tgd-review`_

**Forward requests with tshow are accepted despite the design.**

The approved forwarding design explicitly lists forward plus tshow as invalid, but this block rejects only ThreadParentMessageID. A request with TShow=true is accepted and later normalized away. Reject req.TShow as bad_request here and add a validation test and matching error documentation.

<details>

```
In `message-gatekeeper/handler.go` around line 272: The approved forwarding design explicitly lists forward plus tshow as invalid, but this block rejects only ThreadParentMessageID. A request with TShow=true is accepted and later normalized away. Reject req.TShow as bad_request here and add a validation test and matching error documentation.

```

</details>

message-gatekeeper/handler.go
```diff
@@ -239,33 +238,58 @@ func (h *Handler) processMessage(ctx context.Context, account, roomID, siteID st
...
 
-	// A message with attachments may carry empty content.
-	if req.Content == "" && len(req.Attachments) == 0 {
+	// Forward validation: both fields together, a valid source ID, and no
+	// combination with thread/quote/attachments — a forward is a main-timeline
+	// message carrying a comment plus a server-built snapshot (see
+	// docs/superpowers/specs/2026-08-03-message-forwarding-design.md).
+	isForward := req.ForwardedMessageID != "" || req.ForwardedRoomID != ""
+	if isForward {
+		if req.ForwardedMessageID == "" || req.ForwardedRoomID == "" {
```
julianshen, line 263: _🏷️ input-validation_ | _🟠 Warning_ | _`tgd-review`_

**ForwardedRoomID is used as a NATS token without validation.**

Only non-emptiness is checked before ForwardedRoomID is interpolated into subject.MsgGet. Dots alter subject token boundaries, while wildcards or whitespace can make the publish invalid; these permanent input errors are then reported as transient unavailable failures. Validate the application room-ID format and require a legal single NATS token, returning bad_request on failure.

<details>

```
In `message-gatekeeper/handler.go` around line 263: Only non-emptiness is checked before ForwardedRoomID is interpolated into subject.MsgGet. Dots alter subject token boundaries, while wildcards or whitespace can make the publish invalid; these permanent input errors are then reported as transient unavailable failures. Validate the application room-ID format and require a legal single NATS token, returning bad_request on failure.

```

</details>

PR #179, opened 2026-08-04
feat(admin-service): on-duty room toggle

amdla:
## What

The admin console needs a switch that puts a channel "on duty" — the mode used when staff work off-network and the room has to be locked down to a known membership.

There is no new stored state. `onDuty` is a request parameter that maps onto the room's two pre-existing flags:

| `onDuty` | `restricted` | `externalAccess` |
|---|---|---|
| `true` | `true` | `true` |
| `false` | `false` | `false` |

It is persisted nowhere — not on the room, not on a subscription, not in the audit log. The durable state is those two flags, which already existed and already drive the membership and external-access checks.

## Endpoint

```
POST /v1/admin/rooms/:roomId/onduty
Authorization: Bearer <authToken>

{ "onDuty": true, "ownerAccount": "alice" }
{ "onDuty": false }
```

admin-service does not touch Mongo for this. It calls room-service's existing restrict RPC (`chat.server.request.room.{siteID}.restricted`) over NATS request/reply, so every rule stays in one place and cross-site federation comes along for free.

**Turning duty ON:**
- admin-only — session `admin` role + matching site in admin-service, then a platform-admin check in room-service
- `ownerAccount` is required and becomes the room's sole owner
- that account must already be a member of the room
- the room needs at least 5 members (`RESTRICTED_ROOM_MIN_MEMBERS`, default 5)
- channels only

**Turning duty OFF:** same admin gate; both flags cleared on the room and on every subscription. No owner is sent, so no role is touched and the member floor does not apply.

## No announcement

A duty switch is a hidden operation. room-service no longer publishes a system message for a restriction change — not behind a flag, not for any caller. The announcement is gone outright, so nothing renders in the timeline and nothing notifies.

What it does publish is a flat `room_restricted` **room event** on `chat.room.{roomID}.event` — a state update clients apply without rendering anything. That event type was already defined, round-trip tested and documented, but had no publisher; this wires it up.

Everything else still happens: the room document and every subscription are written, and the change federates to every site that homes a member.

## Reviewer notes

Two decisions worth a second opinion:

1. **The guards are transition-scoped.** `ownerAccount`-required and the ≥5 floor only run on a `false→true` transition. Re-calling on an already-restricted room with a different owner still rotates ownership and still validates that the new owner is a member — only the floor and the require-an-owner rule skip. This is deliberate: re-running the floor would mean a duty room that dropped below five members could never change its on-call owner again, and rotating the on-call owner is the common operation.

2. **Turning duty OFF does not restore prior roles.** Turning it ON flattens every non-owner to `member`, including anyone who held `admin`. Prior roles are recorded nowhere, so that flattening is permanent. If the flow this replaces restored them on the way out, this is a missing feature rather than a port — worth confirming before merge.

## Changes

- **admin-service** — new `room_onduty.go` (handler, request validation, timed NATS RPC, error mapping), route registration, NATS wiring in `main.go` via `natsutil.Connect` with drain on shutdown, and `NATS_URL` / `ROOM_RPC_TIMEOUT` config (the timeout is validated `> 0` and below the HTTP write timeout, so the server can't cut off its own reply).
- **room-service** — the restrict handler drops the system-message publish and emits the room event instead, best-effort and last, so a publish failure never fails the caller and clients are never told about a change the caller saw fail. Flag-only updates now preserve subscription roles; a named owner still rewrites them.
- **pkg/model** — `RoomRestrictedSysData` deleted (it had no producer and no consumer once the system message went away); corrected the `ownerAccount` comment on the room event.
- **docs** — `client-api.md` §9.12 and the restrict-RPC note, plus both derived views.

## Testing

| Check | Result |
|---|---|
| `make lint` | 0 issues |
| `make test` (whole repo, `-race`) | green |
| `make test-integration SERVICE=admin-service` | green — 4 duty tests against real Mongo + NATS |
| `make test-integration SERVICE=room-service` | green — incl. 4 store-branch tests for the role-preservation split |
| `gosec` / repo-local `semgrep` | 0 findings |
| `setRoomOnDuty` coverage | 90.7% of statements |

`govulncheck` was not run locally — `vuln.go.dev` is unreachable from this environment, same as on `main`. CI covers it.

## Summary by CodeRabbit

* **New Features**
  * Added an authenticated endpoint to toggle a room’s on-duty status.
  * Administrators can set restriction and external-access settings, with optional owner assignment or rotation.
  * Changes synchronize with the room service and emit room events.

* **Bug Fixes**
  * Improved handling of invalid requests, unavailable services, timeouts, and remote errors.
  * Administrative restriction changes no longer create timeline messages, notifications, or audit records.

* **Documentation**
  * Updated API and event documentation to describe owner behavior, synchronization, events, and error responses.

### DIFF
--- a/admin-service/config.go
+++ b/admin-service/config.go
@@ -1,6 +1,15 @@
 package main
 
-import "github.com/caarlos0/env/v11"
+import (
+	"fmt"
+	"time"
+
+	"github.com/caarlos0/env/v11"
+)
+
+// httpWriteTimeout bounds a response write; RoomRPCTimeout must stay under it so
+// the handler always wins the race against net/http closing the connection.
+const httpWriteTimeout = 30 * time.Second
 
 type Config struct {
 	Port                  string `env:"PORT" envDefault:"8082"`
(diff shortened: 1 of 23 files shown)


mliu33 (approved):
Outstanding work, thanks a lot!

PR #180, opened 2026-08-04
feat(translation): send J1 in accessToken body + source J1 from the ServiceAccount token file

YI13:
## Summary

Two related changes to `translation-service`'s third-party auth, shipped together:

### 1. J1 sent in the accessToken request body (original fix)

The J1→J2 exchange failed against the real accessToken API with **HTTP 400**:

```
translate backend: get access token: access token endpoint status 400
```

Root cause: the service sent **J1** as an `Authorization:` header with no request body, but the accessToken API expects the J1 in the JSON body as `{"key": <J1>}`. Fixed in `token.go`'s `fetchLocked` — `POST {"key": <J1>}` (resty sets `Content-Type: application/json`); the `Authorization` header is dropped.

### 2. J1 sourced from the Kubernetes ServiceAccount token file (new)

J1 is no longer a fixed string read once at startup. It now comes from a `j1Source`:

- `TRANSLATION_J1_TOKEN` (env literal) — **local/dev + test only**; wins when set.
- `TRANSLATION_J1_TOKEN_FILE` (default `/var/run/secrets/kubernetes.io/serviceaccount/token`) — production; the projected ServiceAccount token.

The file is **re-read on every J1→J2 exchange**, so a kubelet-rotated token is picked up without a restart. `newTranslator` probes the source once at startup, so a missing/empty mount fails fast (crash at boot) rather than per-request. This mirrors `pkg/atrest`'s `VAULT_TOKEN` convention (static env for local, platform credential for prod). No Vault involved — J1 *is* the file contents.

## Not changed

- The translate call still sends **J2** as the `Authorization` header (unchanged).
- No change to the NATS RPC contract or any client-facing schema → `docs/client-api.md` untouched.

## Ops prerequisites (outside this repo)

- Pod must mount the token (`automountServiceAccountToken` not `false`).
- The pod ServiceAccount's token must be a value the accessToken API accepts as J1.

## Testing

- `make test SERVICE=translation-service` green (`-race`); coverage **82.8%**.
- New tests: `j1source_test.go` (staticJ1/fileJ1 read+trim/reread/empty/missing, newJ1Source precedence); `TestTokenProvider_RereadsJ1EachFetch`, `TestTokenProvider_J1SourceError`; `TestNewTranslator_StreamJ1FromFile`, `TestNewTranslator_StreamFailsWhenJ1FileMissing`.
- `make lint` 0 issues; `make sast-gosec` PASS (G304 suppressed on the operator-configured file read).

## Design docs

- `docs/superpowers/specs/2026-08-04-translation-j1-service-account-design.md` (new)
- `docs/superpowers/specs/2026-07-23-translation-api-design.md` (J1-source description updated)

## Summary by CodeRabbit

* **New Features**
  * Added support for sourcing translation credentials from a configured token file or inline value.
  * Token files are re-read during exchanges to support credential rotation.
  * Translation service startup now validates that a usable credential source is available.
  * Credentials are sent in the token exchange request body.

* **Bug Fixes**
  * Improved concurrent token caching to avoid unnecessary exchanges.
  * Added clearer handling for missing, unreadable, or empty credential sources.


translation-service/token.go
```diff
@@ -74,12 +81,16 @@ func (p *tokenProvider) Refresh(ctx context.Context, stale string) (string, erro
 	return p.fetchLocked(ctx)
 }
 
-// fetchLocked calls the accessToken API with the J1 token and caches the J2
-// result. Caller must hold p.mu.
+// fetchLocked calls the accessToken API, sending the J1 token in the JSON body
+// ({"key": <J1>}), and caches the J2 result. Caller must hold p.mu.
 func (p *tokenProvider) fetchLocked(ctx context.Context) (string, error) {
+	key, err := p.readJ1()
```
mliu33, line 101: I just noticed that we are calling Mutex.Lock on Token() read path.

Don't lock on read path as multiple goroutines can't read cached token concurrently. I would suggest to use atomic.Pointer and only lock on refresh or use single flight to handle concurrency. Thanks!

mliu33 (approved):
Great work, thanks. Just one suggestion for performance.

translation-service/token.go
YI13: Done in `c5d95c3`. Moved the cached `(token, expiresAt)` into an immutable `tokenEntry` published via `atomic.Pointer`, so `Token()`'s hot cache-hit path is now lock-free. `mu` is kept only to serialize the fetch, with a re-check after acquiring it (double-checked locking) so we still do at most one exchange per expiry — the thundering-herd protection is preserved. `Refresh()` still locks.

Went with `atomic.Pointer` over `singleflight` to avoid pulling in a new dependency (`golang.org/x/sync`); the double-check gives the same single-fetch guarantee for this narrow case. Added `TestTokenProvider_ConcurrentReadsNoExtraFetch` (50 concurrent readers → one fetch, `-race` clean).

Thanks for the catch!

---

PR #186, opened 2026-08-04
Apply MAX_CONCURRENCY admission cap across request/reply services

hmchangw:
## Summary

The `natsrouter` admission-control cap (`WithMaxConcurrency`) was wired inconsistently across NATS request/reply services. A review found:

- Only `history-service`, `bot-message-handler`, and `bot-room-service` applied it via `MAX_CONCURRENCY`.
- `translation-service` applied a cap but keyed it off `MAX_WORKERS` — which everywhere else in the repo means the JetStream pull-worker pool size, so setting `MAX_CONCURRENCY` on it had no effect.
- Four services fronting constrained MongoDB/Elasticsearch/MinIO pools (`room-service`, `user-service`, `media-service`, `search-service`) ran the router unbounded — the same risk profile `history-service` was hardened against.

This PR makes the knob consistent.

## Changes

**translation-service** — rename `MaxWorkers`/`MAX_WORKERS` → `MaxConcurrency`/`MAX_CONCURRENCY` (default `100` preserved, so no behavior change beyond the env-var name). Gate the router option on `> 0` so `MAX_CONCURRENCY=0` disables cleanly without the non-positive `slog.Warn`.

**room-service, user-service, media-service, search-service** — add a `MaxConcurrency` config field (env `MAX_CONCURRENCY`, default `256`, matching `history-service`) and wire `WithMaxConcurrency` gated on `> 0`. `user-service` also rejects a negative `MAX_CONCURRENCY` in `config.Load`.

In every service, `0` disables the cap (unbounded spawn).

## Cap coverage after this change

| Service | Before | After |
|---|---|---|
| history-service, bot-message-handler, bot-room-service | `MAX_CONCURRENCY` | unchanged |
| translation-service | capped via `MAX_WORKERS` | `MAX_CONCURRENCY` |
| room-service, user-service, media-service, search-service | unbounded | `MAX_CONCURRENCY` (default 256) |
| user-presence-service, room-worker | unbounded | left as-is (mostly `RegisterVoid` / minimal RPC surface) |

## Behavioral note

Default `256` turns admission control **on** for the four previously-unbounded services. Under a genuine burst they now shed excess requests with `errcode.Unavailable("service busy")` instead of piling onto the DB pool — the intended hardening (mirrors `history-service`). Callers should retry on `unavailable`. If preferred, these could instead default to `0` (off) and be enabled per-environment.

## Testing

- Added config tests (TDD: written first, confirmed failing, then implemented) covering default value, env override, and — for `user-service` — negative rejection.
- `make test` passes for all five services.
- `make lint` → `0 issues`.

No client-facing handler schemas or `pkg/model` structs changed, so `docs/client-api.md` needs no update.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FuTiFb7nRq549k9MFrsCdh)_

### DIFF
--- a/media-service/config.go
+++ b/media-service/config.go
@@ -86,6 +86,11 @@ type config struct {
 
 	// AdminAcctPrefix overrides the platform-admin account prefix (ADMIN_ACCT_PREFIX); keep it identical across services.
 	AdminAcctPrefix string `env:"ADMIN_ACCT_PREFIX" envDefault:"p_admin"`
+
+	// MaxConcurrency caps in-flight request handlers so a burst is shed at the
+	// door (ErrUnavailable) instead of piling unbounded work onto MongoDB/MinIO.
+	// 0 disables the cap (unbounded spawn).
+	MaxConcurrency int `env:"MAX_CONCURRENCY" envDefault:"256"`
 }
 
 // clusterBaseURL returns the configured base URL for a site, or "" if unknown.
--- a/media-service/config_test.go
+++ b/media-service/config_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"os"
 	"testing"
 
 	"github.com/caarlos0/env/v11"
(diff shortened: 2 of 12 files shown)


mliu33 (approved):
Thanks

PR #193, opened 2026-08-05
feat: room preview denormalization (Phase 2) implementation

hmchangw:
## Summary

Memoizes each room's last eligible preview message on the Mongo room doc so the common
room-list read stops walking Cassandra.

The cost being attacked is **the walk**, not the RPC. Resolving a preview scans
`messages_by_room` backwards from the tail across bucketed partitions, skipping
ineligible messages (system type, deleted, hidden thread replies) up to a 250-message
budget — per room, for every room in a list. A 1000-room list costs ~1000 walks and one
RPC. This removes the walks and keeps the RPC.

Design doc: `docs/superpowers/specs/2026-08-05-room-preview-denormalization-design.md`,
which also carries the decision log for how this arrived at the shape below.

## Shape

**`history-service` owns the preview outright — sole writer, sole reader.** The stored
preview is an internal memo of that service's own walk, never a field other services
read. This is what keeps the change small: no cross-service write contract, no read
flag, no coordination between a writer worker and a reader service. `broadcast-worker/`
and `user-service/` are untouched by this PR.

**The body is encrypted at rest; the metadata is not.** Messages are already encrypted
in Cassandra, so persisting a plaintext copy of the body in Mongo would have created
exposure the system does not otherwise accept. The preview is therefore split:

- `previewMeta` (message id, sender, createdAt, mentions, visibleTo) stays plaintext —
  precisely the fields Cassandra also leaves unencrypted, so storing them introduces no
  new classification.
- Content and attachments are sealed into `previewCiphertext` + `previewNonce` under a
  site preview DEK.

**The preview DEK is separate from the message DEKs, sharing one Vault KEK.** A single
site-wide key sees far more GCM invocations than a per-room DEK, and a nonce collision
under it would expose the GHASH subkey for everything that key protects — so it must
never seal message bodies. It lives in its own `preview_deks` collection off the
existing Vault wrapper, so no human ever creates or holds key material.

**Rotation is an ops action, not a background job.** `PREVIEW_KEY_EPOCH` (validated
`&gt;= 1`) selects the key. A reader on a different epoch treats the stored preview as
absent and re-resolves, so no reader ever needs a retired key and no migration is
required — previews churn self-correctingly through a rolling deploy.

## Freshness

`previewForMsgId` is the freshness key: the newest message id the resolving walk
**observed** in Cassandra. The preview is current iff it equals `room.lastMsgId`. It is
deliberately *not* `previewMeta.messageId` — the two differ whenever the newest message
is ineligible and gets skipped, and keying on the latter would re-walk forever on a room
whose newest message is a system message.

`previewAsOf` is a separate write-ordering watermark, compared server-side inside the
update pipeline so a late write cannot clobber a newer one. It stays plaintext because
the guard has to evaluate it.

Reads resolve in three tiers: a current stored preview, else the walk, else degraded.
A walk that resolves persists its result best-effort (warm-back). A walk that degrades
writes nothing, so a transient Cassandra failure never overwrites good stored state.

## Write amplification

One write per room per message, not per subscription — a 10,000-member room costs the
same single write as a two-person DM.

## Reviewing this

Seven commits, each building and vetting on its own, in dependency order:

| Commit | Contents |
|---|---|
| `docs:` design spec | The design and its decision log. Read first. |
| `feat(model):` | `Room` gains the storage-only preview fields; `PreviewMeta` added; `PreviewMessage` returns to a pure wire type, never stored. |
| `feat(preview):` | `Sealed`, `Key`, `Seal`/`Open`, and watermark guards that iterate one `previewDocFields` list so set and clear cannot drift. Plus `Build`, `TruncateContent`, `BotAwareDisplayName`. |
| `feat(history-service):` persist | The preview DEK (`preview_deks`, `PREVIEW_KEY_EPOCH`, pinned to primary) and the room repo that projects/unseals. These ride together because threading the cipher changes `NewRoomRepo`'s signature. |
| `feat(history-service):` resolve | Batched read, three-tier resolve, warm-back, and delete-path clearing. |
| `docs(client-api):` | Preview content is a server-truncated 500-rune snippet. |
| `chore(search-service):` | Unrelated one-line goimports fix. |

## Verification

`make lint` (0 issues), `make test` (full suite, `-race`), `make sast` (gosec,
govulncheck, semgrep) all pass. `go vet -tags=integration ./...` is clean, and every
commit in the series builds and vets independently.

## Not in scope

- Automated key rotation — the epoch knob is deliberately manual.
- `visibleTo` remains a per-room field; whether a preview should be per-user is an open
  design question tracked separately.
- The pre-existing `rooms.get` batch-cap behaviour above 100 rooms is untouched and gets
  its own PR.

https://claude.ai/code/session_01EAJXUUvE9JtCisPHSunLGc

## Summary by CodeRabbit

- **New Features**
  - Added faster, more reliable room previews with cached results and automatic fallback when needed.
  - Preview updates remain accurate during message edits and deletions, avoiding stale results.
  - Bot previews can use configured app display names with sensible fallbacks.
  - Preview content and attachments are protected with encryption at rest.

- **Bug Fixes**
  - Preview text is consistently limited to 500 Unicode characters without splitting multibyte characters.

- **Documentation**
  - Updated the client API documentation to describe the server-generated preview snippet behavior.

## Summary by CodeRabbit

- **New Features**
  - Room previews now load faster and remain up to date after message edits or deletions.
  - Preview content is safely limited to 500 characters while preserving multibyte text correctly.
  - Previews can recover gracefully when cached data is unavailable or invalid.
  - Bot display names use configured app names when available, with a fallback otherwise.

- **Bug Fixes**
  - Prevented stale or degraded preview results from overwriting valid previews.
  - Hidden thread replies remain excluded from room previews.

- **Documentation**
  - Updated the client API documentation to describe the server-generated 500-rune preview snippet.

### DIFF
--- a/history-service/internal/mongorepo/room.go
+++ b/history-service/internal/mongorepo/room.go
@@ -3,25 +3,41 @@ package mongorepo
 import (
 	"context"
 	"fmt"
+	"log/slog"
 	"time"
 
 	"go.mongodb.org/mongo-driver/v2/bson"
 	"go.mongodb.org/mongo-driver/v2/mongo"
 
+	"github.com/hmchangw/chat/pkg/atrest"
 	"github.com/hmchangw/chat/pkg/model"
 	"github.com/hmchangw/chat/pkg/mongoutil"
+	"github.com/hmchangw/chat/pkg/preview"
 )
 
 const roomsCollection = "rooms"
 
-// RoomRepo reads room metadata from MongoDB.
+// PreviewDEKCollection holds the wrapped per-site preview DEKs. Separate from
+// atrest.CollectionName though both share the KEK — the split is operational:
+// losing a preview DEK is a cache miss, losing a room DEK is permanent history
+// loss, and co-locating them invites identical backup treatment.
+const PreviewDEKCollection = "preview_deks"
+
+// RoomRepo reads room metadata and owns the stored preview — the only place the
+// preview cipher is applied, so callers above it never hold a key.
 type RoomRepo struct {
-	rooms *mongoutil.Collection[model.Room]
+	rooms  *mongoutil.Collection[model.Room]
+	cipher atrest.Cipher // nil when ATREST_ENABLED=false — previews are then not stored
+	key    preview.Key
 }
 
-func NewRoomRepo(db *mongo.Database) *RoomRepo {
+// NewRoomRepo builds the repo. A nil cipher (ATREST_ENABLED=false) disables
+// preview storage rather than falling back to plaintext.
+func NewRoomRepo(db *mongo.Database, cipher atrest.Cipher, key preview.Key) *RoomRepo {
 	return &RoomRepo{
-		rooms: mongoutil.NewCollection[model.Room](db.Collection(roomsCollection)),
+		rooms:  mongoutil.NewCollection[model.Room](db.Collection(roomsCollection)),
+		cipher: cipher,
+		key:    key,
 	}
 }
 
@@ -57,23 +73,41 @@ func (r *RoomRepo) GetRoomTimes(ctx context.Context, roomID string) (lastMsgAt, 
 	return lastMsgAt, room.CreatedAt, nil
 }
 
-// RoomTimes holds the two room timestamps GetRoomTimesByIDs projects.
+// RoomTimes is what GetRoomTimesByIDs projects per room: the bucket-walk
+// timestamps, plus the memoized preview when one is stored and current.
 type RoomTimes struct {
 	LastMsgAt time.Time
 	CreatedAt time.Time
+	// Preview is the stored preview, already opened. Non-nil only when present,
+	// current (previewForMsgId == lastMsgId), on the configured epoch, and
+	// decryptable; anything else is nil and the caller walks Cassandra.
+	Preview *model.PreviewMessage
 }
 
 // GetRoomTimesByIDs batches GetRoomTimes across ids into a single $in query,
 // returning a map keyed by room ID. Rooms absent from Mongo are simply absent
 // from the map (not an error). Empty ids returns an empty, non-nil map with no query.
+//
+// The stored preview rides the same read, so serving a memoized preview costs
+// no extra round-trip beyond the room-times read the caller already needs.
 func (r *RoomRepo) GetRoomTimesByIDs(ctx context.Context, ids []string) (map[string]RoomTimes, error) {
 	out := make(map[string]RoomTimes, len(ids))
 	if len(ids) == 0 {
 		return out, nil
 	}
+	// lastMsgId rides along because freshness is an identity check against it
+	// (see pkg/preview and the design's §3).
 	rooms, err := r.rooms.FindMany(ctx,
 		bson.M{"_id": bson.M{"$in": ids}},
-		mongoutil.WithProjection(bson.M{"_id": 1, "lastMsgAt": 1, "createdAt": 1}),
+		mongoutil.WithProjection(bson.M{
+			"_id": 1, "lastMsgAt": 1, "createdAt": 1,
+			"lastMsgId":         1,
+			"previewMeta":       1,
+			"previewCiphertext": 1,
+			"previewNonce":      1,
+			"previewKeyEpoch":   1,
+			"previewForMsgId":   1,
+		}),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("get room times for %d rooms: %w", len(ids), err)
@@ -84,9 +118,86 @@ func (r *RoomRepo) GetRoomTimesByIDs(ctx context.Context, ids []string) (map[str
 		if room.LastMsgAt != nil {
 			lastMsgAt = *room.LastMsgAt
 		}
-		out[room.ID] = RoomTimes{LastMsgAt: lastMsgAt, CreatedAt: room.CreatedAt}
+		out[room.ID] = RoomTimes{
+			LastMsgAt: lastMsgAt,
+			CreatedAt: room.CreatedAt,
+			Preview:   r.openStoredPreview(ctx, room),
+		}
 	}
 	return out, nil
+}
+
+// openStoredPreview returns the room's memoized preview, or nil to make the
+// caller walk. Every rejection is a cache miss, never an error — a preview is
+// derived data, so failing to open one costs a walk and nothing else.
+func (r *RoomRepo) openStoredPreview(ctx context.Context, room *model.Room) *model.PreviewMessage {
+	if r.cipher == nil || room.PreviewMeta == nil || len(room.PreviewCiphertext) == 0 {
+		return nil
+	}
+	// Identity, not timestamps: current iff the newest message the walk observed
+	// is still the room's last message.
+	if room.PreviewForMsgID == "" || room.PreviewForMsgID != room.LastMsgID {
+		return nil
+	}
+	if room.PreviewKeyEpoch != r.key.Epoch {
+		// Expected during rotation (both epochs live across a rolling deploy), so
+		// Info — distinct from the same-epoch failure below.
+		slog.InfoContext(ctx, "stored preview on a retired key epoch, re-resolving",
+			"room_id", room.ID, "stored_epoch", room.PreviewKeyEpoch, "configured_epoch", r.key.Epoch)
+		return nil
+	}
+
+	pvw, err := preview.Open(ctx, r.cipher, r.key.SiteID, preview.Sealed{
+		Meta:       *room.PreviewMeta,
+		Ciphertext: room.PreviewCiphertext,
+		Nonce:      room.PreviewNonce,
+		KeyEpoch:   room.PreviewKeyEpoch,
+		ForMsgID:   room.PreviewForMsgID,
+	})
+	if err != nil {
+		// NOT expected: tampering, a truncated write, or a rotated-in-place DEK.
+		slog.WarnContext(ctx, "stored preview failed to decrypt on the configured epoch, re-resolving",
+			"room_id", room.ID, "epoch", room.PreviewKeyEpoch, "error", err)
+		return nil
+	}
+	return &pvw
+}
+
+// SetPreviewMessage seals pvw and stores it, watermark-guarded by asOf (see
+// pkg/preview.GuardedSetFields). A no-op when the cipher is disabled — the room
+// doc must never hold a plaintext body.
+//
+// forMsgID MUST be the newest message id the walk observed, never the lastMsgId
+// read out of Mongo; see previewWalk.NewestObservedID for why.
+//
+//nolint:gocritic // hugeParam: pvw's by-value shape is the RoomRepository.SetPreviewMessage contract shared with the mock and the readcache passthrough; the copy cost is negligible on this best-effort write path.
+func (r *RoomRepo) SetPreviewMessage(ctx context.Context, roomID string, pvw model.PreviewMessage, forMsgID string, asOf int64) error {
+	if r.cipher == nil {
+		return nil
+	}
+	sealed, err := preview.Seal(ctx, r.cipher, r.key, forMsgID, pvw)
+	if err != nil {
+		return fmt.Errorf("seal room preview %s: %w", roomID, err)
+	}
+	pipeline := mongo.Pipeline{{{Key: "$set", Value: preview.GuardedSetFields(sealed, asOf)}}}
+	if _, err := r.rooms.Raw().UpdateOne(ctx, bson.M{"_id": roomID}, pipeline); err != nil {
+		return fmt.Errorf("store room preview %s: %w", roomID, err)
+	}
+	return nil
+}
+
+// ClearPreview removes every stored preview field under the same guard,
+// advancing previewAsOf so an older redelivery cannot resurrect it. For a
+// mutation that leaves the room with no eligible message.
+func (r *RoomRepo) ClearPreview(ctx context.Context, roomID string, asOf int64) error {
+	if r.cipher == nil {
+		return nil
+	}
+	pipeline := mongo.Pipeline{{{Key: "$set", Value: preview.GuardedClearFields(asOf)}}}
+	if _, err := r.rooms.Raw().UpdateOne(ctx, bson.M{"_id": roomID}, pipeline); err != nil {
+		return fmt.Errorf("clear room preview %s: %w", roomID, err)
+	}
+	return nil
 }
 
 // GetRoomUserCount returns the room's userCount via a projected findOne.
--- a/history-service/internal/service/messages.go
+++ b/history-service/internal/service/messages.go
@@ -638,20 +638,35 @@ func (s *HistoryService) DeleteMessage(c *natsrouter.Context, siteID string, req
 	}, nil
 }
 
-// previewAfterMutation resolves the room's current last-eligible preview (same walk as
-// subscription.list) to carry on an edit/delete fan-out. Hidden thread replies (TShow==false)
-// never appear in the room timeline, so they're skipped — clients tell those apart via the
-// event's threadParentMessageId and drive the room preview themselves. TShow==true replies do
-// appear in the room, so they still get a preview. nil for a hidden thread reply, when the room
-// has no eligible message, or on a read error.
+// previewAfterMutation resolves the room's current last-eligible preview (the same
+// walk as subscription.list), both to carry on the edit/delete fan-out and to refresh
+// the stored copy. Bypasses the preview cache so a mutation always sees fresh state.
+// Hidden thread replies (TShow==false) never appear in the room timeline and are
+// skipped; clients drive those themselves off threadParentMessageId.
+//
+// The write happens here rather than riding the canonical event because a mutation is
+// invisible to the freshness check: editing or deleting the preview message does not
+// change lastMsgId, so nothing else would ever invalidate the memoized copy.
 func (s *HistoryService) previewAfterMutation(c *natsrouter.Context, msg *models.Message, roomID string, at time.Time) *models.PreviewMessage {
 	if msg.ThreadParentID != "" && !msg.TShow {
 		return nil
 	}
-	if preview, ok := s.roomLastPreviewMessage(c, roomID, nil, at); ok {
-		return &preview
+	w := s.roomLastPreviewMessage(c, roomID, nil, at)
+	switch w.State {
+	case previewFound:
+		s.storePreview(c, roomID, "mutation", &w, at.UnixMilli())
+		return &w.Preview
+	case previewEmpty:
+		// Walk completed with no eligible survivor, so the stored preview is
+		// definitively wrong. A degraded walk never reaches here.
+		if err := s.rooms.ClearPreview(c, roomID, at.UnixMilli()); err != nil {
+			slog.WarnContext(c, "preview clear after mutation failed", "room_id", roomID,
+				"request_id", natsutil.RequestIDFromContext(c), "error", err)
+		}
+		return nil
+	default: // previewDegraded — a survivor may still exist; clearing would lose it
+		return nil
 	}
-	return nil
 }
 
 // publishCanonicalBestEffort publishes a canonical event; failures are logged and swallowed (Cassandra is source of truth).
--- a/history-service/internal/service/messages.go
+++ b/history-service/internal/service/messages.go
@@ -638,6 +638,19 @@ func (s *HistoryService) DeleteMessage(c *natsrouter.Context, siteID string, req
 	}, nil
 }
 
+// storedPreview reads the room's memoized preview, or nil when none is current.
+// Best-effort: a read failure just sends the caller down the walk it would have
+// taken anyway.
+func (s *HistoryService) storedPreview(ctx context.Context, roomID string) *models.PreviewMessage {
+	rows, err := s.rooms.GetRoomTimesByIDs(ctx, []string{roomID})
+	if err != nil {
+		slog.WarnContext(ctx, "stored preview read failed, falling back to walk", "room_id", roomID,
+			"request_id", natsutil.RequestIDFromContext(ctx), "error", err)
+		return nil
+	}
+	return rows[roomID].Preview
+}
+
 // previewAfterMutation resolves the room's current last-eligible preview (the same
 // walk as subscription.list), both to carry on the edit/delete fan-out and to refresh
 // the stored copy. Bypasses the preview cache so a mutation always sees fresh state.
@@ -650,6 +663,14 @@ func (s *HistoryService) DeleteMessage(c *natsrouter.Context, siteID string, req
 func (s *HistoryService) previewAfterMutation(c *natsrouter.Context, msg *models.Message, roomID string, at time.Time) *models.PreviewMessage {
 	if msg.ThreadParentID != "" && !msg.TShow {
 		return nil
+	}
+	// A mutation that doesn't touch the stored preview cannot change it, so the
+	// walk is skipped and the preview re-reported unchanged. This is the common
+	// case: most edits and deletes target a message that is not the room's newest
+	// eligible one. One projected room-doc read replaces a Cassandra walk plus a
+	// re-seal and re-write.
+	if stored := s.storedPreview(c, roomID); stored != nil && stored.MessageID != msg.MessageID {
+		return stored
 	}
 	w := s.roomLastPreviewMessage(c, roomID, nil, at)
 	switch w.State {
--- a/history-service/internal/service/rooms.go
+++ b/history-service/internal/service/rooms.go
@@ -8,10 +8,12 @@ import (
 
 	"github.com/hmchangw/chat/history-service/internal/cassrepo"
 	"github.com/hmchangw/chat/history-service/internal/models"
+	"github.com/hmchangw/chat/history-service/internal/mongorepo"
 	"github.com/hmchangw/chat/pkg/errcode"
 	pkgmodel "github.com/hmchangw/chat/pkg/model"
 	"github.com/hmchangw/chat/pkg/natsrouter"
 	"github.com/hmchangw/chat/pkg/natsutil"
+	"github.com/hmchangw/chat/pkg/preview"
 )
 
 const (
@@ -49,7 +51,8 @@ func (s *HistoryService) RoomsGet(c *natsrouter.Context, req models.RoomsGetRequ
 
 	ids := dedupRoomIDs(req.RoomIDs)
 	now := time.Now().UTC()
-	metaByRoom := s.resolveRoomMetaHints(c, ids, req.Hints, now)
+	rows := s.readRoomRows(c, ids)
+	metaByRoom := s.resolveRoomMetaHints(ids, req.Hints, rows, now)
 
 	out := make(map[string]models.PreviewMessage, len(ids))
 	var mu sync.Mutex
@@ -68,7 +71,7 @@ func (s *HistoryService) RoomsGet(c *natsrouter.Context, req models.RoomsGetRequ
 		go func() {
 			defer wg.Done()
 			defer func() { <-sem }()
-			lm, ok := s.resolvePreview(c, roomID, metaByRoom[roomID], now)
+			lm, ok := s.resolvePreview(c, roomID, metaByRoom[roomID], rows[roomID].Preview, now)
 			if !ok {
 				return
 			}
@@ -82,44 +85,46 @@ func (s *HistoryService) RoomsGet(c *natsrouter.Context, req models.RoomsGetRequ
 	return &models.RoomsGetResponse{Rooms: out}, nil
 }
 
-// resolveRoomMetaHints maps req.Hints into a per-room *models.RoomMeta and batches ONE
-// GetRoomTimesByIDs read for the rooms whose hint doesn't resolve to a usable lastMsgAt
-// (missing entirely, or rejected by the same sanitizeLastMsgAt every per-room resolve uses).
-// A batch-read failure degrades: those rooms simply keep a nil meta, so the per-room
-// resolveRoomTimes falls back to its existing GetRoomTimes read — the batch failure never
-// fails the whole RPC.
+// readRoomRows issues the ONE batched room-doc read behind rooms.get, covering
+// every requested room rather than just the unhinted ones: RoomTimeHint carries
+// neither the memoized preview nor the lastMsgId its freshness check needs, so a
+// hinted room could otherwise never serve one — and that is every room with
+// messages.
+//
+// A read failure degrades to nil: each room falls back to its per-room
+// GetRoomTimes plus a walk. It never fails the RPC.
+func (s *HistoryService) readRoomRows(ctx context.Context, ids []string) map[string]mongorepo.RoomTimes {
+	rows, err := s.rooms.GetRoomTimesByIDs(ctx, ids)
+	if err != nil {
+		slog.WarnContext(ctx, "rooms.get batch room-doc read degraded, falling back per-room",
+			"room_count", len(ids), "request_id", natsutil.RequestIDFromContext(ctx), "error", err)
+		return nil
+	}
+	return rows
+}
+
+// resolveRoomMetaHints maps req.Hints into a per-room *models.RoomMeta, falling
+// back to the batched room-doc read (rows) for rooms whose hint doesn't resolve
+// to a usable lastMsgAt (missing entirely, or rejected by the same
+// sanitizeLastMsgAt every per-room resolve uses). A room with neither keeps a
+// nil meta, so the per-room resolveRoomTimes falls back to its own GetRoomTimes.
 func (s *HistoryService) resolveRoomMetaHints(
-	ctx context.Context,
 	ids []string,
 	hints map[string]pkgmodel.RoomTimeHint,
+	rows map[string]mongorepo.RoomTimes,
 	now time.Time,
 ) map[string]*models.RoomMeta {
 	metaByRoom := make(map[string]*models.RoomMeta, len(ids))
-	unhinted := make([]string, 0, len(ids))
 	for _, roomID := range ids {
-		hint, ok := hints[roomID]
-		if !ok || sanitizeLastMsgAt(hint.LastMsgAt, now) == nil {
-			unhinted = append(unhinted, roomID)
+		if hint, ok := hints[roomID]; ok && sanitizeLastMsgAt(hint.LastMsgAt, now) != nil {
+			metaByRoom[roomID] = &models.RoomMeta{LastMsgAt: hint.LastMsgAt, CreatedAt: hint.CreatedAt}
 			continue
 		}
-		metaByRoom[roomID] = &models.RoomMeta{LastMsgAt: hint.LastMsgAt, CreatedAt: hint.CreatedAt}
-	}
-	if len(unhinted) == 0 {
-		return metaByRoom
-	}
-
-	times, err := s.rooms.GetRoomTimesByIDs(ctx, unhinted)
-	if err != nil {
-		slog.WarnContext(ctx, "rooms.get batch room-times read degraded, falling back per-room",
-			"room_count", len(unhinted), "request_id", natsutil.RequestIDFromContext(ctx), "error", err)
-		return metaByRoom
-	}
-	for roomID, rt := range times {
-		if rt.LastMsgAt.IsZero() {
-			// Never-messaged room: nothing usable to hint. Leave the meta nil so
-			// resolveRoomTimes takes its plain meta==nil path (one full per-room
-			// GetRoomTimes, normalized directly) instead of treating a synthetic zero
-			// lastMsgAt as a "hint" and tripping the created>last consistency refetch.
+		rt, ok := rows[roomID]
+		if !ok || rt.LastMsgAt.IsZero() {
+			// Absent or never-messaged: nothing usable. A nil meta sends
+			// resolveRoomTimes down its plain per-room path, rather than treating a
+			// synthetic zero lastMsgAt as a hint and tripping the created>last refetch.
 			continue
 		}
 		last := rt.LastMsgAt.UnixMilli()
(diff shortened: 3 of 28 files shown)


docs/superpowers/plans/2026-08-05-room-preview-denormalization.md
```diff
@@ -0,0 +1,1479 @@
...
+- `make fmt`
+- `make lint`
+- `make generate` (confirm no mock drift: `git status` clean of unexpected `mock_*` diffs)
+- `make test`
+- `make test-integration SERVICE=broadcast-worker && make test-integration SERVICE=history-service && make test-integration SERVICE=user-service`
+- `make sast`
+
+Coverage check on the four touched surfaces:
+```bash
+go test -coverprofile=/tmp/claude-0/-home-user-newchat/9a9ab3f7-7dcc-504c-9983-330a09919f11/scratchpad/cov.out ./pkg/preview/... ./broadcast-worker/... ./history-service/... ./user-service/... && go tool cover -func=/tmp/claude-0/-home-user-newchat/9a9ab3f7-7dcc-504c-9983-330a09919f11/scratchpad/cov.out | tail -20
```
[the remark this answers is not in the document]
hmchangw: Not applicable here — those raw commands are what this repo prescribes for coverage. `CLAUDE.md:149` says verbatim:

> Use `go test -coverprofile=coverage.out` and `go tool cover -func=coverage.out` to verify coverage percentages

and there is no general coverage target in the Makefile (`coverage-loadgen-soak` is specific to the loadgen soak tool and its `tools/coveragecheck` thresholds). The plan already flags this line as the documented exception to the make-targets-only rule, so I've left it as-is.

The other two comments in this review are addressed in 25819f4: the "zero new write operations" claim is now scoped to the message hot path in both the spec goal and the plan header, naming the edit/delete and warm-back writes explicitly, and the Task 1 snippet gained its missing `encoding/json` import.

---

hmchangw:
Following up on the three findings in the CodeRabbit review that were reported in the review body rather than as inline threads (the three inline ones are resolved).

### `RoomsGet` batch limit (Major, `user-service/service/subscriptions.go`) — valid, but pre-existing on `main`

The arithmetic checks out. `history-service` rejects the request outright:

```go
// history-service/internal/service/rooms.go:42
if len(req.RoomIDs) > maxRoomsGetBatch { // maxRoomsGetBatch = 100
    return nil, errcode.BadRequest("too many roomIds")
}
```

and `MAX_SUBSCRIPTION_LIMIT` defaults to `1000`, so a page with >100 rooms on a single site loses every preview for that site (the error is swallowed into the "last-message enrichment degraded" warn).

It is not introduced by this PR. `origin/main` already has the unchunked call and the same safety claim in the comment:

```go
// origin/main, user-service/service/subscriptions.go
// … bounded well under history-service's 100-roomId batch cap, so no chunk-split is needed.
m, err := s.history.RoomsGet(c, site, roomIDsBySite[site], hints)
```

This PR narrows the exposure rather than widening it: with `SUBSCRIPTION_PREVIEW_FROM_DOC=true` the local site now sends only the *residual* rooms (those with no denormalized baseline), so it overflows strictly less often than `main`. The cross-site path is unchanged. No regression either way.

Deferring the fix to its own PR rather than folding it into this one, for two reasons:

1. It is a pre-existing defect in a path this PR only touches incidentally, and this PR is already ~4k lines across 44 files.
2. Chunking is not purely mechanical. Today `lastMsgBySite[i] == nil` means "site degraded, skip it". Split into N chunks, a single failing chunk yields *partial* previews for that site, so the degraded contract has to be redefined (merge what succeeded vs. treat the whole site as degraded). That deserves its own review, not a footnote in this one. The `chunkStrings` helper in `threadunread.go` is the obvious building block, and the 101-room test suggested in the review is the right coverage.

### Nitpicks — acknowledged, not taken here

- **TTL cache for `AppNameByAccount` on the bot preview path.** Fair observation: it is one `FindOne` per eligible bot message on the fan-out path. Deferring as speculative — it only affects bot senders, and I would rather add the cache against a measurement than on spec. The `roommetacache` pattern is the right shape if bot volume justifies it.
- **`roomsEnrichStages(dropDeleted, withPreview bool)` positional booleans.** Real readability point, but it is cosmetic churn across five call sites in a PR this size, and the current call sites are comment-annotated. Leaving as-is.

CI is green on `25819f4` — all 9 checks, including the `history-service` integration job that caught the nil-app-store nil-deref fixed in `26363d1`.

---

hmchangw:
Thanks — triaged all three findings from the review of `a7516e0`.

**MD018 in the design spec — fixed in `84e9243`.** Real: the line genuinely began with `#173`, so markdownlint parsed it as an ATX heading. Reflowed and backticked the PR references as suggested.

**Preview `CreatedAt` UTC assertion — not taken.** The assertion would be vacuous where it was proposed: `msgTime` in `TestHandler_HandleCreated_PassesEligiblePreview` is already `time.UTC`, so `assert.Equal(t, time.UTC, pvw.CreatedAt.Location())` passes whether or not `Build` normalizes. Making it meaningful there means giving the event a non-UTC `CreatedAt`, which then breaks the sibling `AdvanceSubscriptionLastSeen(..., msgTime)` expectation — a JSON round-trip yields a fixed-offset zone that is not `DeepEqual` to the original `time.Date(..., loc)` value.

The normalization is already pinned at the level that owns it, in `pkg/preview/preview_test.go`:

```go
func TestBuild_TruncatesAndNormalizesUTC(t *testing.T) {
	loc, err := time.LoadLocation("Asia/Taipei")
	...
	assert.Equal(t, time.UTC, got.CreatedAt.Location())
}
```

That test does exercise the conversion, so a change to the builder cannot drop it silently — which was the stated concern.

**Table-driven consolidation of the six mutation-preview tests — not taken.** Agreed on the general guideline, and CLAUDE.md does prefer table-driven tests for input/output variations. These six differ in more than inputs, though: each sets up a different mock expectation shape (Set / Clear / neither / error-swallowing / fan-out), so the table's `expect func(*MockStore)` column ends up carrying the actual variation while the shared columns carry little. Marked 💤 Low value in the review, and I'd rather not churn ~200 lines of passing test setup on a PR this size. Happy to revisit if you'd prefer it consolidated.

Also note this branch was force-pushed since the last review: the same 44-file change is now organized into 9 logical commits (one per layer), plus a round of `/simplify` cleanups — deduplicated watermark-guard builder in `pkg/preview`, `Build` now takes `model.PreviewMessage` instead of a parallel params struct, allocation-free `TruncateContent`, and the attachment decode hoisted so `handleCreated` decodes once for both the preview and the client message. `make lint` clean and `make test` green on the result.

---

hmchangw:
Triage of the 4 findings on Amendment 2 — 3 fixed in `589da3c`, 1 reframed.

**Line 274 — rotation / cached DEK adoption (Major). Valid, fixed.** A real hole: the design had each process unwrap once and cache permanently, while rotation changed the sentinel id, so a long-running process would keep writing under the retired key indefinitely. `PREVIEW_KEY_EPOCH` is now configuration rather than discovered state, which makes process restart the invalidation mechanism and avoids inventing a refresh protocol — caching permanently is safe precisely because the only thing that changes the epoch also restarts the process holding it. Reader/writer epoch handling and the rolling-deploy churn window (both epochs live, previews rewritten in both directions until the rollout completes) are now specified.

**Line 329 — fail-closed startup check (Major). Valid, fixed.** The assertion as written failed open: "attempt a read, fail if it succeeds" also passes when the query merely errors, so a timeout or a missing namespace would have been read as proof the grant is absent. It now proceeds only on an explicit MongoDB code 13 `Unauthorized`, treats every other outcome — success, timeout, network error, namespace-not-found, unrecognised error — as a startup failure, and uses a projected non-data-bearing query whose cursor is never decoded or logged.

**Line 244 — grammar (Minor). Valid, fixed.**

**Line 264 — storage contract still cleartext (Major). Correct observation, but not a defect to fix here.** `GuardedSetFields` does still build `previewMessage`, and `model.PreviewMessage` does still carry cleartext `Content`/`Attachments` — that divergence is the expected state, not an oversight. Amendment 2 is a design decision recorded ahead of the rework, which has not started. Changing the helpers now would leave the branch half-migrated with no encryption path behind it.

What was missing is that the amendment did not *say* so, which is a fair thing to have flagged. It now carries an explicit status header: design decision only, the branch still writes the plaintext shape, and a green CI run does not mean the code matches the spec.

Broader note for anyone reading this PR: it should not merge as-is. The denormalization writes ≤500 runes of message body plus attachments to Mongo in cleartext, bypassing `pkg/atrest`, which classifies exactly those fields as sensitive and nulls them in Cassandra. On `main` the room doc holds only `lastMsgAt`/`lastMsgId`, so this is new exposure rather than inherited. Amendment 2 records the agreed fix; the implementation is pending.

---

hmchangw:
@coderabbitai resume

Merged `main` (`465bab2`) into the branch — the merge conflict is resolved. Conflicts were confined to `history-service/internal/config`, where both sides appended a validation and its tests at the same spot; both are kept.

One semantic conflict git couldn't see, worth a look on the next pass: `main` flipped the Mongo client default to `secondaryPreferred` and pinned the message DEK collection to primary to compensate. This branch's second DEK collection (`preview_deks`) predates that flip and would have silently inherited `secondaryPreferred` — history-service mints the preview DEK on first use and reads it back immediately, so a lagging secondary could miss the key it had just written. It's now pinned to primary too.

The PR description has also been rewritten: it described a superseded design (write routed through `broadcast-worker`, plaintext preview, a `user-service` read flag). Both of those services are now byte-identical to `main`.

`make lint`, `make test`, and `make sast` pass on the merge result.

---

julianshen:
**Actionable comments posted: 18** — 🔴 9 blocking · 🟠 9 warning

<details>

* `docs/client-api.md`
* `docs/superpowers/specs/2026-08-05-room-preview-denormalization-design.md`
* `history-service/internal/mongorepo/room.go`
* `history-service/internal/mongorepo/room_test.go`
* `history-service/internal/mongorepo/setup_test.go`
* `history-service/internal/readcache/readcache.go`
* `history-service/internal/readcache/readcache_test.go`
* `history-service/internal/service/displayname_test.go`
* `history-service/internal/service/integration_test.go`
* `history-service/internal/service/reactions.go`
* `history-service/internal/service/rooms.go`
* `history-service/internal/service/rooms_test.go`
* `history-service/internal/service/rooms_warmback_test.go`
* `pkg/model/message.go`
* `pkg/model/room.go`
* `pkg/model/room_preview_test.go`
* `pkg/preview/preview.go`
* `pkg/preview/preview_test.go`
* `pkg/preview/seal.go`
* `pkg/preview/seal_test.go`
* `search-service/integration_index_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `silent-failure`
* `tgd-review`

</details>

history-service/internal/service/messages.go
```diff
@@ -638,20 +638,56 @@ func (s *HistoryService) DeleteMessage(c *natsrouter.Context, siteID string, req
...
+	// case: most edits and deletes target a message that is not the room's newest
+	// eligible one. One projected room-doc read replaces a Cassandra walk plus a
+	// re-seal and re-write.
+	if stored := s.storedPreview(c, roomID); stored != nil && stored.MessageID != msg.MessageID {
+		return stored
+	}
+	w := s.roomLastPreviewMessage(c, roomID, nil, at)
+	switch w.State {
+	case previewFound:
+		s.storePreview(c, roomID, "mutation", &w, at.UnixMilli())
```
julianshen, line 678: _🏷️ cassandra-consistency_ | _🔴 Blocking_ | _`cassandra`_

**A failed preview write leaves Cassandra mutations invisibly stale.**

After UpdateMessageContent or SoftDeleteMessage commits in Cassandra, SetPreviewMessage/ClearPreview failures are only logged and the mutation still succeeds. Edits and deletes do not advance rooms.lastMsgId, so the old previewForMsgId continues matching lastMsgId and openStoredPreview serves the old Mongo value without consulting Cassandra. A Mongo failure here can therefore expose pre-edit content or a deleted message indefinitely, until another message is created. Durably retry/enqueue the preview update or invalidation before acknowledging the mutation, or introduce a mutation version that readers verify so this partial failure becomes a Cassandra cache miss.

<details>

```
In `history-service/internal/service/messages.go` around line 678: After UpdateMessageContent or SoftDeleteMessage commits in Cassandra, SetPreviewMessage/ClearPreview failures are only logged and the mutation still succeeds. Edits and deletes do not advance rooms.lastMsgId, so the old previewForMsgId continues matching lastMsgId and openStoredPreview serves the old Mongo value without consulting Cassandra. A Mongo failure here can therefore expose pre-edit content or a deleted message indefinitely, until another message is created. Durably retry/enqueue the preview update or invalidation before acknowledging the mutation, or introduce a mutation version that readers verify so this partial failure becomes a Cassandra cache miss.

```

</details>

pkg/preview/preview.go
```diff
@@ -0,0 +1,124 @@
...
+	"previewMeta",
+	"previewCiphertext",
+	"previewNonce",
+	"previewKeyEpoch",
+	"previewForMsgId",
+}
+
+// guardedFields builds the watermark-guarded $set shared by the set and clear
+// paths: each field becomes onPass[field] when asOf >= the stored previewAsOf
+// (missing => 0; ties => last writer wins), and previewAsOf advances with it.
```
julianshen, line 91: _🏷️ ordering_ | _🔴 Blocking_ | _`distributed-system`_

**Wall-clock milliseconds cannot order concurrent preview writes.**

`asOf` is derived from handler wall clocks for mutations and compared at millisecond precision. If instance A commits edit A first with a clock ahead, then instance B commits edit B later in Cassandra with a lower timestamp, B's correct preview write is rejected. Equal-millisecond operations are also ordered only by Mongo arrival because `$gte` lets the last arrival win, which need not match Cassandra's winner. Since edits and deletes do not change `lastMsgId`, the stale preview still passes the freshness check indefinitely. Use a datastore-derived mutation revision or fencing token with a deterministic tie-breaker that reflects Cassandra's winning mutation, rather than `UnixMilli` wall time.

<details>

```
In `pkg/preview/preview.go` around line 91: `asOf` is derived from handler wall clocks for mutations and compared at millisecond precision. If instance A commits edit A first with a clock ahead, then instance B commits edit B later in Cassandra with a lower timestamp, B's correct preview write is rejected. Equal-millisecond operations are also ordered only by Mongo arrival because `$gte` lets the last arrival win, which need not match Cassandra's winner. Since edits and deletes do not change `lastMsgId`, the stale preview still passes the freshness check indefinitely. Use a datastore-derived mutation revision or fencing token with a deterministic tie-breaker that reflects Cassandra's winning mutation, rather than `UnixMilli` wall time.

```

</details>

history-service/internal/service/messages.go
```diff
@@ -638,20 +638,56 @@ func (s *HistoryService) DeleteMessage(c *natsrouter.Context, siteID string, req
...
+		return &w.Preview
+	case previewEmpty:
+		// Walk completed with no eligible survivor, so the stored preview is
+		// definitively wrong. A degraded walk never reaches here.
+		if err := s.rooms.ClearPreview(c, roomID, at.UnixMilli()); err != nil {
+			slog.WarnContext(c, "preview clear after mutation failed", "room_id", roomID,
+				"request_id", natsutil.RequestIDFromContext(c), "error", err)
+		}
+		return nil
```
julianshen, line 687: _🏷️ failure-handling_ | _🔴 Blocking_ | _`distributed-system`_

**A degraded mutation walk preserves a known-stale preview.**

When the stored preview is the message being edited or deleted, the code knows that memo is invalid. If the Cassandra replacement walk then times out or exhausts its budget, `previewDegraded` leaves the memo untouched. A successful delete therefore can continue serving the deleted body because `lastMsgId` does not change and the old memo still passes freshness. Clearing derived data does not lose the unknown survivor; it forces later reads to retry the walk. Retain whether the target matched the stored preview and conditionally invalidate it on degradation, ideally with a compare-and-delete on the observed preview message ID/watermark so a newer concurrent write is not removed.

<details>

```
In `history-service/internal/service/messages.go` around line 687: When the stored preview is the message being edited or deleted, the code knows that memo is invalid. If the Cassandra replacement walk then times out or exhausts its budget, `previewDegraded` leaves the memo untouched. A successful delete therefore can continue serving the deleted body because `lastMsgId` does not change and the old memo still passes freshness. Clearing derived data does not lose the unknown survivor; it forces later reads to retry the walk. Retain whether the target matched the stored preview and conditionally invalidate it on degradation, ideally with a compare-and-delete on the observed preview message ID/watermark so a newer concurrent write is not removed.

```

</details>

history-service/internal/service/rooms.go
```diff
@@ -129,43 +134,113 @@ func (s *HistoryService) resolveRoomMetaHints(
...
+//
+// asOf stays explicit at the call site: warm-back passes the preview's own
+// createdAt (always <= any event that observed it, so it cannot outrank a
+// mutation write), a mutation passes its edit/delete time. path distinguishes the
+// two in logs.
+func (s *HistoryService) storePreview(ctx context.Context, roomID, path string, w *previewWalk, asOf int64) {
+	if w.NewestObservedID == "" {
+		// No observed id means no key to invalidate against later — skip rather
+		// than store something permanently un-invalidatable.
+		return
```
julianshen, line 183: _🏷️ atomicity_ | _🔴 Blocking_ | _`distributed-system`_

**Best-effort preview writes make successful mutations permanently stale.**

The Cassandra edit/delete can commit while `SetPreviewMessage` or `ClearPreview` fails in Mongo; the error is only logged and the mutation still succeeds. Because mutations do not advance `lastMsgId`, the previous preview remains identity-fresh and is served indefinitely, including deleted content, until another message is created. Record a durable, retryable refresh or invalidation intent atomically with the mutation, such as an outbox consumed idempotently, and only acknowledge completion once that intent is durable.

<details>

```
In `history-service/internal/service/rooms.go` around line 183: The Cassandra edit/delete can commit while `SetPreviewMessage` or `ClearPreview` fails in Mongo; the error is only logged and the mutation still succeeds. Because mutations do not advance `lastMsgId`, the previous preview remains identity-fresh and is served indefinitely, including deleted content, until another message is created. Record a durable, retryable refresh or invalidation intent atomically with the mutation, such as an outbox consumed idempotently, and only acknowledge completion once that intent is durable.

```

</details>

history-service/internal/service/rooms.go
```diff
@@ -129,43 +134,113 @@ func (s *HistoryService) resolveRoomMetaHints(
...
+// preview (no walk), the in-process cache, then a Cassandra walk that warm-backs
+// what it finds.
+//
+// stored comes from the batch room-doc read, already freshness- and
+// epoch-checked; nil means walk. meta is the times hint (nil when none applies);
+// previewAfterMutation passes nil so mutations always see fresh state. The cache
+// is positives-only, so empty rooms and read failures re-resolve. Both walk paths
+// share one load closure, so a cache hit never re-warms.
+func (s *HistoryService) resolvePreview(ctx context.Context, roomID string, meta *models.RoomMeta, stored *models.PreviewMessage, now time.Time) (models.PreviewMessage, bool) {
+	// The point of the denormalization: the repo already checked freshness and
```
julianshen, line 147: _🏷️ rolling-deploy_ | _🔴 Blocking_ | _`distributed-system`_

**Mixed-version mutations leave stored previews stale.**

During a rolling deployment, a new instance can warm back preview m1, after which an old instance can edit or delete m1. The old binary mutates Cassandra but does not update the new Preview fields. A later new instance sees `previewForMsgId == lastMsgId` because mutations do not change `lastMsgId`, and returns the pre-mutation preview without walking. Use a two-phase rollout: first deploy mutation writers while stored-preview reads remain disabled, then enable reads only after all old instances are drained; alternatively add a mutation generation that all readers can validate.

<details>

```
In `history-service/internal/service/rooms.go` around line 147: During a rolling deployment, a new instance can warm back preview m1, after which an old instance can edit or delete m1. The old binary mutates Cassandra but does not update the new Preview fields. A later new instance sees `previewForMsgId == lastMsgId` because mutations do not change `lastMsgId`, and returns the pre-mutation preview without walking. Use a two-phase rollout: first deploy mutation writers while stored-preview reads remain disabled, then enable reads only after all old instances are drained; alternatively add a mutation generation that all readers can validate.

```

</details>

history-service/internal/service/rooms.go
julianshen, line 183: _🏷️ consistency_ | _🔴 Blocking_ | _`mongodb`_

**Mutation preview failures leave edited or deleted content current.**

SetPreviewMessage errors are logged and discarded here, and ClearPreview errors are likewise discarded in previewAfterMutation. After Cassandra commits an edit or deletion, lastMsgId does not change, so the old Mongo preview still passes previewForMsgId == lastMsgId. A transient Mongo/Vault failure can therefore make subsequent room lists serve old edited content or a deleted message indefinitely, until another message advances lastMsgId. Make post-mutation persistence part of the success contract with bounded retry/error propagation, or enqueue a durable idempotent repair before acknowledging the mutation.

<details>

```
In `history-service/internal/service/rooms.go` around line 183: SetPreviewMessage errors are logged and discarded here, and ClearPreview errors are likewise discarded in previewAfterMutation. After Cassandra commits an edit or deletion, lastMsgId does not change, so the old Mongo preview still passes previewForMsgId == lastMsgId. A transient Mongo/Vault failure can therefore make subsequent room lists serve old edited content or a deleted message indefinitely, until another message advances lastMsgId. Make post-mutation persistence part of the success contract with bounded retry/error propagation, or enqueue a durable idempotent repair before acknowledging the mutation.

```

</details>

history-service/internal/service/messages.go
```diff
@@ -638,20 +638,56 @@ func (s *HistoryService) DeleteMessage(c *natsrouter.Context, siteID string, req
...
+	// walk is skipped and the preview re-reported unchanged. This is the common
+	// case: most edits and deletes target a message that is not the room's newest
+	// eligible one. One projected room-doc read replaces a Cassandra walk plus a
+	// re-seal and re-write.
+	if stored := s.storedPreview(c, roomID); stored != nil && stored.MessageID != msg.MessageID {
+		return stored
+	}
+	w := s.roomLastPreviewMessage(c, roomID, nil, at)
+	switch w.State {
+	case previewFound:
```
julianshen, line 677: _🏷️ nats/request-reply_ | _🔴 Blocking_ | _`nats`_

**Preview storage can block the canonical mutation event.**

The Cassandra edit/delete has already committed when this synchronous preview write runs. SetPreviewMessage and ClearPreview use the NATS request context and may wait on Vault or Mongo until its deadline. Because previewAfterMutation must finish before MsgCanonicalUpdated/Deleted can be published, a slow store can leave the publisher with an expired context and also prevent a timely reply, while the mutation remains committed and canonical consumers never observe it. Compute the preview first, publish the canonical event before optional persistence, and move persistence to a bounded retry worker or give it a short child deadline that preserves time for publish/reply.

<details>

```
In `history-service/internal/service/messages.go` around line 677: The Cassandra edit/delete has already committed when this synchronous preview write runs. SetPreviewMessage and ClearPreview use the NATS request context and may wait on Vault or Mongo until its deadline. Because previewAfterMutation must finish before MsgCanonicalUpdated/Deleted can be published, a slow store can leave the publisher with an expired context and also prevent a timely reply, while the mutation remains committed and canonical consumers never observe it. Compute the preview first, publish the canonical event before optional persistence, and move persistence to a bounded retry worker or give it a short child deadline that preserves time for publish/reply.

```

</details>

history-service/internal/service/rooms.go
```diff
@@ -129,43 +134,113 @@ func (s *HistoryService) resolveRoomMetaHints(
...
+// asOf stays explicit at the call site: warm-back passes the preview's own
+// createdAt (always <= any event that observed it, so it cannot outrank a
+// mutation write), a mutation passes its edit/delete time. path distinguishes the
+// two in logs.
+func (s *HistoryService) storePreview(ctx context.Context, roomID, path string, w *previewWalk, asOf int64) {
+	if w.NewestObservedID == "" {
+		// No observed id means no key to invalidate against later — skip rather
+		// than store something permanently un-invalidatable.
+		return
+	}
```
julianshen, line 184: _🏷️ nats/reliability_ | _🔴 Blocking_ | _`nats`_

**Mutation preview failures are reported as success with no repair.**

storePreview logs and discards SetPreviewMessage failures, and the ClearPreview branch in messages.go does the same. Editing or deleting the current preview does not change lastMsgId, so the old document still satisfies previewForMsgId == lastMsgId. The mutation request and canonical publication therefore continue as successful, but history-service is now the sole writer and no canonical persistence signal remains to trigger a retry. For example, after deleting m1, a transient Mongo failure leaves subsequent rooms.get calls serving m1 indefinitely. Persist a retryable outbox or durable JetStream repair job keyed by roomID and asOf, acknowledging it only after Mongo succeeds; the existing watermark guard makes retries idempotent.

<details>

```
In `history-service/internal/service/rooms.go` around line 184: storePreview logs and discards SetPreviewMessage failures, and the ClearPreview branch in messages.go does the same. Editing or deleting the current preview does not change lastMsgId, so the old document still satisfies previewForMsgId == lastMsgId. The mutation request and canonical publication therefore continue as successful, but history-service is now the sole writer and no canonical persistence signal remains to trigger a retry. For example, after deleting m1, a transient Mongo failure leaves subsequent rooms.get calls serving m1 indefinitely. Persist a retryable outbox or durable JetStream repair job keyed by roomID and asOf, acknowledging it only after Mongo succeeds; the existing watermark guard makes retries idempotent.

```

</details>

history-service/internal/service/rooms.go
```diff
@@ -129,43 +134,113 @@ func (s *HistoryService) resolveRoomMetaHints(
...
+func (s *HistoryService) resolvePreview(ctx context.Context, roomID string, meta *models.RoomMeta, stored *models.PreviewMessage, now time.Time) (models.PreviewMessage, bool) {
+	// The point of the denormalization: the repo already checked freshness and
+	// epoch and opened it, so this is ready to serve.
+	if stored != nil {
+		return *stored, true
+	}
+	load := func(ctx context.Context) (models.PreviewMessage, bool, error) {
+		w := s.roomLastPreviewMessage(ctx, roomID, meta, now)
+		if w.State == previewFound {
+			s.storePreview(ctx, roomID, "warm_back", &w, w.Preview.CreatedAt.UnixMilli())
```
julianshen, line 155: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**Warm-back cannot replace a mutation-stamped preview.**

Post-mutation writes set previewAsOf to the edit/delete time, but this warm-back uses the preview message's original CreatedAt. After an edit or delete, that timestamp is necessarily older. If key rotation or a decrypt failure makes the stored preview unreadable, the Cassandra walk succeeds but GuardedSetFields rejects the warm-back forever. The room therefore keeps walking until later activity creates a newer timestamp, contradicting the lazy-rotation and self-healing behavior. Use an operation-resolution version that can advance the watermark safely, and test rotation/decrypt recovery after a mutation write.

<details>

```
In `history-service/internal/service/rooms.go` around line 155: Post-mutation writes set previewAsOf to the edit/delete time, but this warm-back uses the preview message's original CreatedAt. After an edit or delete, that timestamp is necessarily older. If key rotation or a decrypt failure makes the stored preview unreadable, the Cassandra walk succeeds but GuardedSetFields rejects the warm-back forever. The room therefore keeps walking until later activity creates a newer timestamp, contradicting the lazy-rotation and self-healing behavior. Use an operation-resolution version that can advance the watermark safely, and test rotation/decrypt recovery after a mutation write.

```

</details>

julianshen, line 155: _🏷️ cassandra-operations_ | _🟠 Warning_ | _`cassandra`_

**Older preview timestamps can permanently reject warm-backs.**

The messages_by_room walk is warm-backed using the eligible preview's CreatedAt, but post-mutation writes set previewAsOf to the later edit/delete time. For example, edit eligible m1 in February, then create an ineligible system message m2 in March. The walk observes and skips m2, resolves m1, but submits m1's January timestamp; GuardedSetFields rejects it against February's watermark, leaving previewForMsgId=m1 while lastMsgId=m2. Every cache expiry then repeats the Cassandra walk, potentially scanning the full 250-row budget. Key rotation has the same problem when an old-epoch preview has a mutation watermark. Carry the newest observed row's ordering timestamp for ordinary warm-backs and make the guard explicitly permit a higher key epoch to replace an older epoch without allowing epoch rollback.

<details>

```
In `history-service/internal/service/rooms.go` around line 155: The messages_by_room walk is warm-backed using the eligible preview's CreatedAt, but post-mutation writes set previewAsOf to the later edit/delete time. For example, edit eligible m1 in February, then create an ineligible system message m2 in March. The walk observes and skips m2, resolves m1, but submits m1's January timestamp; GuardedSetFields rejects it against February's watermark, leaving previewForMsgId=m1 while lastMsgId=m2. Every cache expiry then repeats the Cassandra walk, potentially scanning the full 250-row budget. Key rotation has the same problem when an old-epoch preview has a mutation watermark. Carry the newest observed row's ordering timestamp for ordinary warm-backs and make the guard explicitly permit a higher key epoch to replace an older epoch without allowing epoch rollback.

```

</details>

history-service/internal/service/rooms.go
```diff
@@ -129,43 +134,113 @@ func (s *HistoryService) resolveRoomMetaHints(
...
+// is positives-only, so empty rooms and read failures re-resolve. Both walk paths
+// share one load closure, so a cache hit never re-warms.
+func (s *HistoryService) resolvePreview(ctx context.Context, roomID string, meta *models.RoomMeta, stored *models.PreviewMessage, now time.Time) (models.PreviewMessage, bool) {
+	// The point of the denormalization: the repo already checked freshness and
+	// epoch and opened it, so this is ready to serve.
+	if stored != nil {
+		return *stored, true
+	}
+	load := func(ctx context.Context) (models.PreviewMessage, bool, error) {
+		w := s.roomLastPreviewMessage(ctx, roomID, meta, now)
```
julianshen, line 153: _🏷️ key-rotation_ | _🟠 Warning_ | _`distributed-system`_

**Key rotation is blocked by mutation watermarks.**

Warm-back always writes with the preview message's creation time. For example, a message created at 100 is edited at 200, so the mutation stores epoch 1 with `previewAsOf=200`. After deploying epoch 2, the reader rejects epoch 1, walks Cassandra, and attempts epoch 2 with `asOf=100`; the watermark rejects it forever. Every epoch-2 read then repeats the Cassandra walk, so rotation never completes for that room until later activity happens to carry a newer timestamp. Add a rekey path conditioned on the observed epoch, watermark, and freshness key, preserving the observed watermark while ensuring a concurrent mutation wins.

<details>

```
In `history-service/internal/service/rooms.go` around line 153: Warm-back always writes with the preview message's creation time. For example, a message created at 100 is edited at 200, so the mutation stores epoch 1 with `previewAsOf=200`. After deploying epoch 2, the reader rejects epoch 1, walks Cassandra, and attempts epoch 2 with `asOf=100`; the watermark rejects it forever. Every epoch-2 read then repeats the Cassandra walk, so rotation never completes for that room until later activity happens to carry a newer timestamp. Add a rekey path conditioned on the observed epoch, watermark, and freshness key, preserving the observed watermark while ensuring a concurrent mutation wins.

```

</details>

history-service/internal/mongorepo/room.go
```diff
@@ -3,25 +3,41 @@ package mongorepo
...
+	key    preview.Key
 }
 
-func NewRoomRepo(db *mongo.Database) *RoomRepo {
+// NewRoomRepo builds the repo. A nil cipher (ATREST_ENABLED=false) disables
+// preview storage rather than falling back to plaintext.
+func NewRoomRepo(db *mongo.Database, cipher atrest.Cipher, key preview.Key) *RoomRepo {
 	return &RoomRepo{
-		rooms: mongoutil.NewCollection[model.Room](db.Collection(roomsCollection)),
+		rooms:  mongoutil.NewCollection[model.Room](db.Collection(roomsCollection)),
```
julianshen, line 38: _🧹 read-consistency_ | _🟠 Warning_ | _`mongodb`_

**Preview freshness checks may read a lagging secondary.**

The rooms collection inherits the configured client read preference, unlike the DEK collections explicitly pinned to primary. With secondary or secondaryPreferred, GetRoomTimesByIDs can read an older lastMsgId and matching older previewForMsgId, incorrectly treating that preview as current. More seriously, previewAfterMutation can read an older preview whose MessageID differs from the message being mutated, skip its Cassandra walk and Mongo repair, and turn temporary replication lag into persistent stale state. Use a primary-pinned collection handle for preview-dependent room reads, or otherwise require primary reads for this query.

<details>

```
In `history-service/internal/mongorepo/room.go` around line 38: The rooms collection inherits the configured client read preference, unlike the DEK collections explicitly pinned to primary. With secondary or secondaryPreferred, GetRoomTimesByIDs can read an older lastMsgId and matching older previewForMsgId, incorrectly treating that preview as current. More seriously, previewAfterMutation can read an older preview whose MessageID differs from the message being mutated, skip its Cassandra walk and Mongo repair, and turn temporary replication lag into persistent stale state. Use a primary-pinned collection handle for preview-dependent room reads, or otherwise require primary reads for this query.

```

</details>

history-service/internal/service/rooms.go
```diff
@@ -129,43 +134,113 @@ func (s *HistoryService) resolveRoomMetaHints(
...
+		if w.State == previewFound {
```
julianshen, line 154: _🏷️ key-rotation_ | _🟠 Warning_ | _`mongodb`_

**Retired-key previews written by mutations cannot warm back.**

Warm-back orders its update by the preview's original CreatedAt. A preview stored after an edit or deletion instead has previewAsOf equal to the later mutation time. After PREVIEW_KEY_EPOCH is bumped, openStoredPreview rejects the old epoch, but this fallback write has an older asOf and GuardedSetFields silently rejects it. The room remains on the retired epoch and every subsequent list walks Cassandra until a later message changes the room. Propagate the rejected preview's watermark and identity, then conditionally reseal it with a compare-and-swap against the exact old epoch/preview state so a concurrent newer mutation cannot be overwritten.

<details>

```
In `history-service/internal/service/rooms.go` around line 154: Warm-back orders its update by the preview's original CreatedAt. A preview stored after an edit or deletion instead has previewAsOf equal to the later mutation time. After PREVIEW_KEY_EPOCH is bumped, openStoredPreview rejects the old epoch, but this fallback write has an older asOf and GuardedSetFields silently rejects it. The room remains on the retired epoch and every subsequent list walks Cassandra until a later message changes the room. Propagate the rejected preview's watermark and identity, then conditionally reseal it with a compare-and-swap against the exact old epoch/preview state so a concurrent newer mutation cannot be overwritten.

```

</details>

pkg/preview/preview.go
```diff
@@ -0,0 +1,124 @@
...
+func guardedFields(asOf int64, onPass map[string]any) bson.M {
+	cond := bson.M{"$gte": bson.A{asOf, bson.M{"$ifNull": bson.A{"$previewAsOf", 0}}}}
+	out := bson.M{"previewAsOf": bson.M{"$cond": bson.A{cond, asOf, "$previewAsOf"}}}
+	for _, f := range previewDocFields {
```
julianshen, line 95: _🏷️ atomicity_ | _🟠 Warning_ | _`mongodb`_

**Equal-millisecond mutations can overwrite newer previews.**

The watermark uses epoch milliseconds and accepts equality. Two edits/deletions in the same millisecond therefore have identical ordering keys; a delayed request or retry for the older mutation can arrive after the newer one and overwrite it. Because mutations do not advance lastMsgId, the regressed preview still passes the freshness check indefinitely. Carry a source-of-truth monotonic mutation version or deterministic unique tie-breaker and compare the complete ordering key; changing only >= to > would merely make first arrival win and would not establish the correct order.

<details>

```
In `pkg/preview/preview.go` around line 95: The watermark uses epoch milliseconds and accepts equality. Two edits/deletions in the same millisecond therefore have identical ordering keys; a delayed request or retry for the older mutation can arrive after the newer one and overwrite it. Because mutations do not advance lastMsgId, the regressed preview still passes the freshness check indefinitely. Carry a source-of-truth monotonic mutation version or deterministic unique tie-breaker and compare the complete ordering key; changing only >= to > would merely make first arrival win and would not establish the correct order.

```

</details>

history-service/internal/service/rooms.go
julianshen, line 154: _🏷️ nats/timeouts_ | _🟠 Warning_ | _`nats`_

**Warm-back delays the rooms.get reply despite being best-effort.**

After Cassandra has resolved the preview, the rooms.get loader still waits synchronously for DEK acquisition/encryption and Mongo UpdateOne before returning the available result. On a slow Vault or Mongo dependency, each room goroutine can remain blocked until the NATS request deadline, and RoomsGet waits for all of them, causing the requester to time out solely because an optional cache write stalled. Send warm-backs through a bounded background queue with controlled shutdown/retries, or use a short independent timeout and do not wait for completion in the reply path.

<details>

```
In `history-service/internal/service/rooms.go` around line 154: After Cassandra has resolved the preview, the rooms.get loader still waits synchronously for DEK acquisition/encryption and Mongo UpdateOne before returning the available result. On a slow Vault or Mongo dependency, each room goroutine can remain blocked until the NATS request deadline, and RoomsGet waits for all of them, causing the requester to time out solely because an optional cache write stalled. Send warm-backs through a bounded background queue with controlled shutdown/retries, or use a short independent timeout and do not wait for completion in the reply path.

```

</details>

history-service/internal/service/rooms.go
julianshen, line 153: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Key rotation can leave previews permanently on the old epoch.**

Post-mutation previews use the mutation time as previewAsOf, but this rotation warm-back uses the preview message's older CreatedAt. After PREVIEW_KEY_EPOCH changes, the walk succeeds but GuardedSetFields rejects the lower watermark, leaving the old-epoch ciphertext in place. Every subsequent request logs the epoch mismatch and falls back to Cassandra again (after each in-process cache expiry), potentially restoring the per-room walk load indefinitely. Repair epoch mismatches with a compare-and-swap matching the observed old epoch/watermark and permit only an epoch advance, so concurrent old-version instances cannot revert the repaired document.

<details>

```
In `history-service/internal/service/rooms.go` around line 153: Post-mutation previews use the mutation time as previewAsOf, but this rotation warm-back uses the preview message's older CreatedAt. After PREVIEW_KEY_EPOCH changes, the walk succeeds but GuardedSetFields rejects the lower watermark, leaving the old-epoch ciphertext in place. Every subsequent request logs the epoch mismatch and falls back to Cassandra again (after each in-process cache expiry), potentially restoring the per-room walk load indefinitely. Repair epoch mismatches with a compare-and-swap matching the observed old epoch/watermark and permit only an epoch advance, so concurrent old-version instances cannot revert the repaired document.

```

</details>

pkg/preview/preview.go
```diff
@@ -0,0 +1,124 @@
...
+	"previewCiphertext",
+	"previewNonce",
+	"previewKeyEpoch",
+	"previewForMsgId",
+}
+
+// guardedFields builds the watermark-guarded $set shared by the set and clear
+// paths: each field becomes onPass[field] when asOf >= the stored previewAsOf
+// (missing => 0; ties => last writer wins), and previewAsOf advances with it.
+func guardedFields(asOf int64, onPass map[string]any) bson.M {
```
julianshen, line 92: _🎯 concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Millisecond ties allow stale mutation writes to win.**

Mutation writers pass UnixMilli timestamps, and this >= guard accepts every write sharing the same millisecond. Two concurrent edits/deletes can therefore compute different previews with the same asOf; whichever Mongo update arrives last wins even if it represents the earlier mutation. The freshness ID remains unchanged for mutations, so the stale result can persist until later activity. Clock skew between service instances can similarly reject a causally newer write. Use an authoritative, totally ordered mutation version or a timestamp-plus-unique-tiebreaker watermark rather than wall-clock milliseconds with last-writer-wins ties.

<details>

```
In `pkg/preview/preview.go` around line 92: Mutation writers pass UnixMilli timestamps, and this >= guard accepts every write sharing the same millisecond. Two concurrent edits/deletes can therefore compute different previews with the same asOf; whichever Mongo update arrives last wins even if it represents the earlier mutation. The freshness ID remains unchanged for mutations, so the stale result can persist until later activity. Clock skew between service instances can similarly reject a causally newer write. Use an authoritative, totally ordered mutation version or a timestamp-plus-unique-tiebreaker watermark rather than wall-clock milliseconds with last-writer-wins ties.

```

</details>

pkg/preview/preview.go
julianshen, line 91: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Equal millisecond watermarks let stale mutations win.**

The guard accepts equality, while callers reduce edit/delete times to UnixMilli. Two concurrent mutations in the same millisecond can therefore both pass; if the older mutation's Mongo write completes last, it permanently replaces the newer preview even though Cassandra retained the newer mutation. lastMsgId does not change for edits/deletes, so freshness checking cannot repair this. Preserve a sufficiently precise, totally ordered mutation version and add a reverse-completion concurrency test.

<details>

```
In `pkg/preview/preview.go` around line 91: The guard accepts equality, while callers reduce edit/delete times to UnixMilli. Two concurrent mutations in the same millisecond can therefore both pass; if the older mutation's Mongo write completes last, it permanently replaces the newer preview even though Cassandra retained the newer mutation. lastMsgId does not change for edits/deletes, so freshness checking cannot repair this. Preserve a sufficiently precise, totally ordered mutation version and add a reverse-completion concurrency test.

```

</details>

hmchangw:
Closing this as **folded into #222**, not abandoned. The two PRs were framed as competing answers to "who writes the preview"; they turned out to be complementary, and #222 now carries both.

What went across, unchanged in substance:

- The **warm-back** — a resolved read-time walk persists what it found, keyed on the newest message the walk *observed* (not the preview's own id, not the doc's `lastMsgId`). #222 had dropped this and was re-walking Cassandra on every read of an unstored room.
- `SetPreviewMessage` / `ClearPreview` on `RoomRepo`, plus the new `UpdatePreviewBody` for the body-only mutation write, and the readcache pass-throughs.
- The **mutation writes staying in history-service**. #222 had routed these through broadcast-worker on a `MessageEvent.PreviewGone` flag; that flag is now deleted. The service that walks Cassandra is the one that writes, exactly as this design had it.
- `botAwareDisplayName` consolidated onto `pkg/preview`, with the unwired-app-store guard.
- `2026-08-05-room-preview-denormalization-design.md` itself, annotated with what the fold changed (this design names history-service the *sole* writer; it is now one of two — broadcast-worker writes on insert).

What did not, deliberately: broadcast-worker now writes the preview on insert, from the message it is already fanning out. That is the one thing #222 adds, and it is what makes the walk rare rather than merely cached.

**@julianshen — your review from earlier today is not lost, and I'd rather not waste it.** Most of the files it covers (`pkg/preview/*`, `mongorepo/room.go`, `service/rooms.go`, `readcache`, `pkg/model/room.go`) are in #222 in near-identical form, so the majority of those 18 findings very likely still apply there. I'm going to triage them against #222's tree and respond to each on that PR rather than asking you to re-review from scratch. Apologies for the churn — the fold landed after your pass.

---

PR #196, opened 2026-08-05
fix: walk back room last-message fields on last-message delete

chenjr0719:
## Summary

Deleting a room's **last** message left the room's denormalized last-message fields
(`lastMsgAt` — the `subscription.list` sort key — plus `lastMsgId` and `lastMentionAllAt`)
pointing at the now-deleted message. The room would keep sorting by, and preview, a message
that no longer exists. The send path was the only writer of these fields (coalesced), so a
delete never corrected them.

This walks the room fields back to the latest surviving message on delete — the room-level
mirror of the thread-level `newThreadLastMsgAt` walk-back the delete event already carries.
`history-service` recomputes the post-delete room state and `broadcast-worker` applies it
directly to the room document when the deleted message was the room's current last.

## Changes

- **`pkg/model/event.go`** — `MessageEvent` gains transport-only walk-back fields
  (`newRoomLastRecomputed`, `newRoomLastMsgAt`, `newRoomLastMsgId`, `newRoomLastMentionAllAt`;
  `bson:"-"`, stripped before the client). `newRoomLastRecomputed` gates the apply so a hidden
  thread reply, a read degrade, or an older producer leaves the room fields untouched (rather
  than wrongly clearing them).
- **`history-service`** — on a non-hidden-reply delete, `fillRoomLastWalkback` reuses the same
  preview walk (`roomLastPreviewMessage`, now with an error-surfacing variant to tell an empty
  room from a read failure) to source the surviving last message; an empty room signals a clear.
  When the deleted message was itself an `@all`, a bounded backward walk recomputes the latest
  surviving `@all` for `lastMentionAllAt`.
- **`broadcast-worker`** — `handleDeleted` applies the walk-back via a new direct (un-coalesced)
  `SetRoomLastMessage` store method when `msg.ID == room.LastMsgID`; nil last-message pointers
  clear the fields (room emptied). It re-parses the deleted message's content to decide whether
  `lastMentionAllAt` shifts. Best-effort — the field is decorative and self-heals on the next
  message, so a write blip never redelivers the fan-out.
- **`docs/client-api.md`, `docs/client-api/events.md`, `docs/client-api/request-reply.md`** —
  document the client-observable behavior (room re-sorts / group-mention badge clears after a
  last-message delete).

Edit ±`@all` recompute is deferred (the edit producer doesn't yet recompute `lastMentionAllAt`);
noted as out of scope here.

## Verification

- `chat-lint ./...` — 0 issues.
- `go test ./pkg/model/... ./broadcast-worker/... ./history-service/internal/service/...` — pass.
- New tests: model round-trip (fields omitted when zero, round-trip when set, absent from BSON);
  history-service delete walk-back (surviving last message, empty-room clear, `@all` recompute);
  broadcast-worker `handleDeleted` (applies when deleted was last, skips when non-last, clears
  when emptied, touches `lastMentionAllAt` only for an `@all` delete).

🤖 Generated with Claude Code

### DIFF
--- a/broadcast-worker/handler.go
+++ b/broadcast-worker/handler.go
@@ -502,6 +502,19 @@ func (h *Handler) handleDeleted(ctx context.Context, evt *model.MessageEvent) er
 	if err := h.publishMutation(ctx, room, model.RoomEventMessageDeleted, msg.ID, &del); err != nil {
 		return fmt.Errorf("publish delete mutation for room %s message %s: %w", room.ID, msg.ID, err)
 	}
+	// Walk the room's denormalized last-message fields back off the deleted message when it was the
+	// room's current last — otherwise room.lastMsgAt (subscription.list sort key) and the preview
+	// keep pointing at a deleted message. NewRoomLastRecomputed gates it (unset ⇒ hidden thread
+	// reply, read degrade, or old producer ⇒ leave the fields as-is). Best-effort: the field is
+	// decorative and self-heals on the next message, so a write blip mustn't redeliver the fan-out.
+	if evt.NewRoomLastRecomputed && msg.ID == room.LastMsgID {
+		setMentionAll := mention.Parse(msg.Content).MentionAll // lastMentionAllAt shifts only if the deleted msg was @all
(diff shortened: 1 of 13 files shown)


chenjr0719:
Closing this draft — deferring the last-message walk-back-on-delete change for now. The underlying stale sort-key behavior can be revisited in a fresh PR if we pick it back up.

PR #197, opened 2026-08-05
perf(user-service): stop reading every room to serve one subscription.list page

hmchangw:
Serving one page of the chat list no longer reads every room the account is subscribed to. No API change, no write-path change, no schema change.

Measured: **1.6× faster at 10 subscriptions, 5.3× at 100, 16× at 1000.** Numbers and method below.

---

## How to review this

The commits are meant to be read in order; each builds and tests green on its own.

| # | Commit | What to look for | Size |
|---|--------|------------------|------|
| 1 | `959039e` sort-key cache | Small self-contained LRU. Skim. | ~150 |
| 2 | `bcc6704` config knobs | Two env vars. Skim. | ~35 |
| 3 | `1cce7c0` **the actual change** | Everything worth arguing about is here. | ~900 |
| 4 | `b8ede7b` client-api docs | One paragraph of contract. | ~5 |
| 5 | `2b52f74` benchmark | Produced the numbers below. | ~130 |
| 6 | `e7274c9` sort-tie determinism | Adds `_id` as the last sort key. See "Ordering equivalence". | ~100 |
| 7 | `6d9c1f6` restored badge test | Test-only; a deletion in commit 3 that shouldn't have happened. | ~70 |

In commit 3, `subscriptions.go` is the file that matters. Read `AggregateSubscriptions` first — it is now ~25 lines of named steps — then follow the four calls it makes (`resolveSortKeys`, `buildListRows`, `sortListRows`, `fillListPage`) in that order. Everything else in the diff is tests and projections.

**The three things worth your scepticism**, in the order I'd want them challenged:

1. **Is the new ordering identical to the old?** The sort moved from Mongo into Go. See the mapping table below.
2. **Can anything stale reach the client?** Only the row order may lag. See "What can be stale".
3. **Do the benchmark numbers transfer to production?** They were taken against a loopback Mongo with a fully-cached `rooms` collection. See "Measured improvement" for which way each bias pushes.

---

## Problem

`AggregateSubscriptions` joined **every** subscription the account holds to its room, sorted that whole joined set inside Mongo, then applied `$skip`/`$limit`. An account with 500 subscriptions paid 500 room index seeks and a 500-row sort to return 20 rows. The sort key lives on the joined room, so it can't be pushed ahead of the join — the work is unavoidable in that shape.

Denormalizing `lastMsgAt` onto subscriptions was rejected: a message in a 5,000-member room would become 5,000 writes.

## Approach

Two observations:

- **Only two fields decide the order** — a room's `lastMsgAt` and `createdAt`. That's a small record, so a lot of rooms fit in memory.
- **Only rows on the page need to be complete.** Everything before paging is filtering and sorting.

So the read runs in four steps:

1. **Read the subscriptions**, projected to the four fields used to filter and sort (`_id`, `roomId`, `roomType`, `name`). One indexed query, small rows.
2. **Resolve each room's sort key** from an in-process cache; misses go into one batched `$in`. Rooms with no local document — every room owned by another site — are remembered as missing so they aren't looked up again.
3. **Filter, sort and page in Go.**
4. **Read the page**: two queries sized to the page — its rooms, and its full subscription documents.

Warm, the per-subscription cost drops from one room lookup to one small row — measured at ~155 µs vs ~10 µs below. The request is **not** O(page): step 1 still reads a row per subscription and sorts them all. It stays linear in the account's size, with a far smaller constant.

## Measured improvement

Benchmarked with commit `2b52f74` against a containerised Mongo, page size 20, warm cache:

| Subscriptions | Old | New | Speedup |
|---|---|---|---|
| 10 | 2.74 ms | 1.72 ms | **1.59×** |
| 50 | 9.09 ms | 2.73 ms | **3.32×** |
| 100 | 17.67 ms | 3.32 ms | **5.32×** |
| 500 | 78.78 ms | 7.32 ms | **10.76×** |
| 1000 | 153.70 ms | 9.52 ms | **16.15×** |

The shape matters more than any single row. Marginal cost per additional subscription:

| | old | new |
|---|---|---|
| 10 → 50 | 159 µs | 25.3 µs |
| 50 → 100 | 172 µs | 11.7 µs |
| 100 → 500 | 153 µs | 10.0 µs |
| 500 → 1000 | 150 µs | 4.4 µs |

The old path is almost perfectly linear at **~155 µs per subscription** — that is the per-subscription correlated `$lookup`, and it is the entire problem. The new path is ~1.5 ms fixed (three round trips plus the page reads) plus **~10 µs per subscription** for the step-1 read and the Go sort. That residual slope is the linear term noted in the trade-offs: real, but ~15× smaller. The ratio therefore keeps widening with N and had not flattened by 1000.

At 1000 rooms a user was waiting **154 ms** for a chat-list page. That is perceptible, so this is a latency change as well as a load change.

**How the harness biases the result**, both directions:

- **Mongo on loopback.** The new path makes three sequential round trips where the old made one, so it pays +2×RTT that this setup does not charge. This **overstates** the improvement, most at small N.
- **A tiny, fully-cached `rooms` collection.** The old path's N `_id` seeks are as cheap as they can ever be. This **understates** the improvement.
- **Warm cache.** `-benchtime 50x` is one cold iteration and 49 warm, so it reads as the steady state.

**The RTT caveat, quantified.** Break-even round-trip time — above which the extra two round trips cost more than the eliminated lookups save:

| Subscriptions | 10 | 50 | 100 | 500 | 1000 |
|---|---|---|---|---|---|
| Break-even RTT | 0.51 ms | 3.18 ms | 7.17 ms | 35.7 ms | 72.1 ms |

So a same-AZ Mongo at ~1 ms RTT would make the smallest accounts (≲20 rooms) marginally slower while everything above still wins comfortably. Issuing the two step-4 reads concurrently would remove most of that; it is listed under trade-offs and not done here.

**Still unmeasured: the cold path.** These are `BenchmarkSubscriptionList`. `BenchmarkSubscriptionListNoCache` isolates the phased read from the cache and has not been run. It is the relevant number after a restart, for dormant accounts, and for windowed requests.

**The cache does much less work for `updatedWithinDays` requests.** `resolveSortKeys` demotes any cached key that falls *outside* the window to a miss, so every dormant room is re-read on every windowed request — precisely the tail the window exists to hide. An account whose subscriptions are mostly dormant never reaches the warm case. `chat-frontend` never sends `updatedWithinDays`, so the web client is unaffected; other clients that do send it should not expect the table above.

**For the record**, an earlier revision of this description estimated the gain from BSON document sizes alone — 5.8× asymptotic, with a predicted crossover near 27–40 subscriptions below which the new path would be *slower*. Both were wrong: bytes were never the dominant term, the real driver is per-`$lookup`-invocation cost, and no crossover appears in this setup. That model has been removed rather than left alongside the measurements.

**How to confirm after deploy:** the `sub_sortkey/l1` cachemetrics hit rate, and `subscription.list` latency traces split by account size. The slope against subscription count should drop by roughly the 155 µs → 10 µs shown above.

## Ordering equivalence

The highest-risk part of the port. Every element of the old pipeline has a counterpart:

| Old (Mongo) | New (Go) |
|---|---|
| `roomsEnrichStages()` — `$lookup` + `preserveNullAndEmptyArrays` | `resolveSortKeys` — a room with no local document is a `Missing` key |
| missing room ⇒ null room fields, row kept | `enrichListRows` — no baseline ⇒ room fields left zero, row kept |
| `$match {lastMsgAt: {$gte: cutoff}}` | `buildListRows` — cutoff check |
| `__sortKey = $ifNull[lastMsgAt, createdAt]` | `listRow.sortAt` |
| `$sort {__selfDM: -1, __sortKey: -1, name: 1}` | `sortListRows` — same three keys, same directions |
| descending sort puts missing keys last | `sortAt == nil` sorts after every dated row |
| *(no counterpart — see below)* | `sortListRows` — `sub.ID` as a final tiebreak |
| `AggregatePagedHasMore` — `$skip`/`$limit`, over-read one | `fillListPage` — same offset/limit, same over-read |

Note for anyone comparing against an older read of this description: `main` no longer drops rooms whose name starts with `Del-` anywhere on this path — that filter was removed before this branch was rebased, and neither side of the table has it. A `Del-`-named room is enriched like any other, on both the old and the new path.

The one deliberate departure is that last row. Rows that tie on *every* other key fell back to `sort.SliceStable`, which preserves the order of the phase-one `Find` — a query with no sort, so that order is arbitrary between requests. Two subscriptions sharing an activity timestamp and a name could swap places between two pages of one `hasMore` drain, repeating one row and dropping another with no data change at all. Mongo's own sort was equally unstable, so this is not a regression, but it was free to remove: commit `e7274c9` adds the immutable `sub.ID` as the last key, after the self-DM pin, activity, undated-last and name. Covered by `TestSortListRows_TiesBreakOnSubscriptionID` and a precedence table-test pinning that the new key disturbs none of the four above it.

Otherwise pinned by `TestAggregateSubscriptions_NullSortKeysLastAndCreatedAtFallback_Integration`, which covers the fallback and the nil-last tiebreak together. Note this test was written for the new path — if you want the equivalence proven rather than asserted, that's the place to push.

## What can be stale, and what can't

The cache holds sort keys for 15s, so **the order rows come back in** can be that far out of date. Nothing else is:

| Invariant | Mechanism | Test |
|---|---|---|
| Row fields are current | Step 4 re-reads rooms and subs | — |
| A room entering the activity window appears immediately | A cached key failing the cutoff is re-read | `TestAggregateSubscriptions_WindowSeesFreshActivityDespiteWarmCache_Integration` |
| The page is full and `hasMore` is honest | Refill from later candidates when rows drop | `TestAggregateSubscriptions_PageRefillsPastFreshDeletes_Integration` |
| An un-favorited row leaves the page mid-request | Step 4 re-applies the list filter, not just `_id`s | `TestEnrichListRows_DropsRowsThatNoLongerMatchTheFilter_Integration` |

The window one is the subtle argument: `lastMsgAt` only moves forward, so a cached key that *passes* the window is certainly still inside it, but one that *fails* may have gone stale the moment the room got its first message in the window. Ordering may lag the TTL; whether a room appears at all may not.

One gap in that argument: the re-read exempts negative (`Missing`) entries. A room with no local document that *acquires* one with in-window activity therefore stays hidden from a windowed list for up to the TTL. Room documents are written before their subscriptions on every creation path and cross-site rooms never get a local document, so this needs an unusual ordering (a backfill or migration) to reach. The clean fix is to project `siteId` into `subLite` and keep negative entries sticky only for genuinely remote subscriptions; demoting every `Missing` hit under a cutoff would instead re-probe every cross-site room on every windowed request. Not done here.

The cache is per replica, so paging a long list across replicas is best-effort — a row can repeat or shift between pages. Clients already dedupe by `roomId`. Documented in `docs/client-api.md` §subscription.list and the derived view.

## Why it reads each collection twice

Subscriptions are read twice (small, then page-sized) and rooms twice (sort keys, then page baselines). That's deliberate, and the measurements above are *net of* both reads. Why the obvious collapses don't work:

- *Fetch full subscriptions in step 1.* That's the over-fetch the review round removed: the full document per subscription instead of a 130-byte row, to return 20 rows.
- *Serve the page's room fields from the cache.* It holds two dates; the page needs ten fields, including `encKey` and the counts. And it would serve stale counts.
- *Widen the cache to all ten fields.* Buys back the staleness the design exists to avoid, and puts room private-key bytes in a process-wide cache.

## Config

| Env | Default | Meaning |
|-----|---------|---------|
| `SUBS_SORTKEY_CACHE_SIZE` | `100000` | Rooms to keep sort keys for |
| `SUBS_SORTKEY_CACHE_TTL` | `15s` | How long a key may be used before re-reading |

Zero or less for either turns the cache off, and every list resolves keys from a fresh batched read. The `envDefault`s are the only place the defaults are written down. Together they are the kill switch if the hit rate or p99 disappoints after deploy — note they disable the *cache*, not the phased read path, which has no flag.

## Trade-offs I'd flag

- **The smallest accounts depend on Mongo RTT.** On loopback the new path wins even at 10 subscriptions (1.59×). At ~1 ms RTT the extra two round trips eat that margin below ~20 subscriptions. Break-even RTTs are tabulated above; concurrent step-4 reads would largely fix it.
- **Windowed lists get much less from the cache**, as described above.
- **A cold cache reads a page's rooms twice** — once for the sort key, once for the baseline. Bounded to page-size documents. Avoiding it means widening step 2's projection for every miss, which is worse whenever N ≫ page.
- **Cost still scales with the account's subscription count, not the page.** Step 1 reads one 130-byte row per subscription, and the sort is O(N log N) in the service process rather than in Mongo — measured at ~10 µs per subscription. Cheap and linear, but not O(page), and `MAX_CONCURRENCY` (256) is what bounds the aggregate heap.
- **The two step-4 reads run in series.** They depend only on values already in hand, so an errgroup would save a round trip per page.
- **A cold cache issues one unbounded `$in`.** Bounded by the account's subscription count, not the page size. Real accounts are nowhere near Mongo's BSON command limit, but there is no chunking.

## Review round (multi-agent, adversarially verified)

An 8-finder / 5-verifier pass found and fixed: step 1 fetching all ~25 fields per subscription, `withinDays` able to hide a room whose activity the cache hadn't seen, the persisted `origin` field missing from the projection, panics on out-of-range page values, refills reading one row at a time, missing `cachemetrics`, a dead `__sortKey`, and drift in the derived API view.

Two findings refuted and left alone: a claimed race caching a subscription as missing before its room exists (room documents are acked before their subscriptions on every creation path, and these reads hit the primary), and a goroutine-leak claim against the LRU library (repo-wide pre-existing pattern; the pinned version exposes no `Close`).

A later pre-merge risk review added commits 6 and 7 and the corrections marked above. It also refuted three standing review comments whose premise no longer holds on this base: "missing local rooms treated as cross-site / restore the site-aware filter" (`main`'s repo has no `siteID` field and its join uses `preserveNullAndEmptyArrays`, so keeping those rows *is* the old behaviour), and "hundreds of serial refill queries" (rows drop only when a subscription changes within the same request — milliseconds — not within the cache TTL, and the `Del-` rename premise is gone).

## Testing

TDD throughout; every test failed for the intended reason first.

- **Unit:** cache wrapper (disabled mode, roundtrip, negative entries, TTL expiry); the small projection pinned to exactly four fields; the room projection derived from the struct's bson tags, so the two drifting apart fails the test; sort-tie determinism and key precedence.
- **Integration:** the invariants table above, plus undated ordering and the `createdAt` fallback, `origin` surviving the round trip, Teams rooms excluded on this path (nothing covered that before — #255's tests exercise the stage builder alone), out-of-range page values degrading without panics, and the restored badge-projection pin (`TestGetActiveSubscriptions_ProjectsBadgeFields_Integration` — the raw pipeline output across local, empty and cross-site rooms, asserting exactly five fields so no room `encKey` reaches the badge path).
- The entire pre-existing `AggregateSubscriptions` integration contract passes unchanged.

`make lint` 0 issues, full-repo `make test` (race), the user-service integration suite, and `gosec` all green. govulncheck and the semgrep registry rulesets can't reach their hosts from the sandbox; local `.semgrep/` rules pass — rely on CI's `sast` job for those two.

## Out of scope — and one correction

`GetDMSubscription`, `GetSubscriptionByRoomID` and `GetActiveSubscriptions` still use the shared `$lookup` pipeline and are genuinely narrow: the first two match a unique key and `$limit 1`, and `GetActiveSubscriptions` caps with `$limit` **before** the join, so `$lookup` touches at most `maxSubs` rows whatever the account's size. Nothing analogous to fix.

**`CountActiveSubscriptions` is not in that category, and an earlier version of this description wrongly said it was.** It runs `$match` → `roomsEnrichStages()` → `$count` with no cap, so it joins every active subscription to its room to return a single integer. That is the same shape this PR removes from the list path, minus the sort — and at the ~155 µs per subscription measured above, an account with 1000 active rooms costs ~150 ms to *count*. It is reached from `subscription.count`'s non-unread branch; the unread branch goes through the bounded `GetActiveSubscriptions` and is served by the badge cache.

Not fixed here, and the fix is **not** simply an extension of this PR. Serving the count from the sort-key cache only helps on a hit: cold, resolving N rooms still costs N index lookups — the same N the `$lookup` does. It would collapse N subpipeline invocations into one query, but not the seeks. (It would hit more often than it sounds, since step 2 above resolves keys for *every* subscription, not just the page, so a list call warms exactly the set the count needs — but the cache is per replica and expires in 15s, so that is a tendency, not a guarantee.)

The change that removes the seeks outright is denormalisation: mark members' subscriptions when a room is soft-deleted, and the count becomes a plain indexed `countDocuments` with no room access at all. The write fan-out that ruled this out for `lastMsgAt` does not apply — `lastMsgAt` churns on every message, whereas room soft-deletion is rare and already a fan-out event. That carries its own correctness questions (ordering, at-least-once delivery, cross-site rooms whose deletion is not locally visible) and belongs in its own PR.

https://claude.ai/code/session_01EfPa3pTMGaKVzV1xe1sqp7

---
_Generated by [Claude Code](https://claude.ai/code/session_01EfPa3pTMGaKVzV1xe1sqp7)_


user-service/mongorepo/subscriptions.go
```diff
@@ -250,39 +298,222 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
+		return zero, err
+	}
+	rows := buildListRows(subs, keys, account, listType, favorite, withinDays)
+	sortListRows(rows)
+	pageRows, hasMore := pageListRows(rows, page)
+	data, err := r.enrichSubscriptionPage(ctx, pageRows)
+	if err != nil {
+		return zero, err
+	}
+	return mongoutil.OffsetPageHasMore[model.EnrichedSubscription]{Data: data, HasMore: hasMore}, nil
```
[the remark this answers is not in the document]
hmchangw: Fixed in 9d2ccf4 — went with the refill approach to keep the read path O(page): `fillListPage` walks the sorted candidates from the offset, enriching `limit+1`-sized batches and refilling from later live candidates whenever the fresh read drops a freshly soft-deleted row, so the page and `HasMore` always describe the live sequence. In your A–D example the response is now `[C, D]` with `HasMore=false` in the same request. Added the suggested paginated integration test (`TestAggregateSubscriptions_PageRefillsPastFreshSoftDeletes_Integration`) — written red against the old code first. Steady state is still a single batch; a refill round only occurs within the cache TTL of a soft-delete.

---

hmchangw:
Addressed the CodeRabbit review in 8108720:

- **Fill-batch floor** (actionable): `fillListPage` now enriches at least `minFillBatch` (32) candidates per round, so a degenerate limit can't turn a freshly-dead candidate prefix into one round trip per row.
- **`bsonTagPaths` hardening**: tag options stripped via `strings.Cut`, `bson:"-"` skipped, pointer-to-struct dereferenced — as suggested.
- **Projection pin**: new `TestSubscriptionFieldsProjection_MatchesModelTags` derives the expected set from `model.Subscription`'s persisted bson tags, so a model field missing from `subscriptionFieldsProjection` now fails a unit test.

Skipped (with reason): the phase-1 row-count metric and chunked sort-key `$in`. Both costs are cold-start-only (steady state is cache-served) and already observable — a cold-cache spike shows on the `sub_sortkey` miss counter and an oversized `$in` on Mongo slow-query logs. Worth revisiting only if traces show real outlier accounts.

Also removed two dead-code leftovers surfaced by a local review pass: the never-read `SubscriptionRepo.siteID` field (its comment described a filter that is actually `room.name`-based) and the orphaned `createdAt` key in `roomsEnrichStages`' `$lookup` projection.

---

julianshen:

### Related work

- [PR #255](https://github.com/hmchangw/newchat/pull/255) — feat\(search\): SHOW\_TEAMS\_ROOM config to hide Teams rooms from list/search/counts (merged)

<details>
<summary>📒 Files reviewed (18)</summary>

* `docs/client-api.md`
* `pkg/testutil/mongo.go`
* `user-service/main.go`
* `user-service/mongorepo/sortkeycache.go`
* `user-service/mongorepo/sortkeycache_test.go`
* `user-service/mongorepo/subscriptions_bench_test.go`
* `user-service/mongorepo/subscriptions_cache_test.go`
* `user-service/mongorepo/subscriptions_fill_test.go`
* `user-service/mongorepo/subscriptions_lite_test.go`
* `user-service/mongorepo/subscriptions_projection_test.go`
* `user-service/mongorepo/subscriptions_sort_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -259,40 +292,342 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
+// and the old join dropped them under a window too).
+func (r *SubscriptionRepo) resolveSortKeys(ctx context.Context, subs []subLite, cutoff *time.Time) (map[string]roomSortKey, error) {
+	keys := make(map[string]roomSortKey, len(subs))
+	// The unique (roomId, u.account) index makes one account's roomIDs distinct
+	// already — no dedup needed.
+	var misses []string
+	for i := range subs {
+		id := subs[i].RoomID
+		k, ok := r.sortKeys.get(ctx, id)
+		if ok && cutoff != nil && !k.Missing && (k.LastMsgAt == nil || k.LastMsgAt.Before(*cutoff)) {
```
julianshen: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**Negative cache entries can hide rooms from an activity window.**

If a missing room is cached and its document subsequently appears with recent activity, this condition preserves the `Missing` hit. `buildListRows` then drops it because `LastMsgAt` is nil, so the fresh baseline read is never reached and `withinDays` membership remains stale until TTL expiry. This contradicts the documented always-fresh window membership. Re-read negative entries for windowed requests and add an integration test covering missing-to-active.

<details>

> ‼️ **IMPORTANT**

```suggestion
// negative-cached as Missing so cross-site rooms don't re-probe Mongo on
// every unwindowed list. Windowed lists re-read Missing hits because a local
// room document may have appeared since the negative result was cached, and
// window membership must not lag the TTL.
func (r *SubscriptionRepo) resolveSortKeys(ctx context.Context, subs []subLite, cutoff *time.Time) (map[string]roomSortKey, error) {
	keys := make(map[string]roomSortKey, len(subs))
	// The unique (roomId, u.account) index makes one account's roomIDs distinct
	// already — no dedup needed.
	var misses []string
	for i := range subs {
		id := subs[i].RoomID
		k, ok := r.sortKeys.get(ctx, id)
		if ok && cutoff != nil && (k.Missing || k.LastMsgAt == nil || k.LastMsgAt.Before(*cutoff)) {
```

</details>

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 465: If a missing room is cached and its document subsequently appears with recent activity, this condition preserves the `Missing` hit. `buildListRows` then drops it because `LastMsgAt` is nil, so the fresh baseline read is never reached and `withinDays` membership remains stale until TTL expiry. This contradicts the documented always-fresh window membership. Re-read negative entries for windowed requests and add an integration test covering missing-to-active.

```

</details>

julianshen, line 484: _🏷️ consistency_ | _🟠 Warning_ | _`distributed-system`_

**Negative cache entries hide rooms entering an activity window.**

A supported late-materialization path breaks window freshness: a subscription initially has no local room document, so the first request caches `Missing`; the room then appears with an in-window `lastMsgAt`. Subsequent windowed requests exempt that cached marker from revalidation, and `buildListRows` drops the row before the fresh page read can discover it. The room remains absent until cache expiry, contradicting the documented always-fresh `updatedWithinDays` membership. Treat `Missing` as a miss whenever a cutoff applies and add a windowed late-room integration test.

<details>

> ‼️ **IMPORTANT**

```suggestion
		if ok && cutoff != nil && (k.Missing || k.LastMsgAt == nil || k.LastMsgAt.Before(*cutoff)) {
```

</details>

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 476: A supported late-materialization path breaks window freshness: a subscription initially has no local room document, so the first request caches `Missing`; the room then appears with an in-window `lastMsgAt`. Subsequent windowed requests exempt that cached marker from revalidation, and `buildListRows` drops the row before the fresh page read can discover it. The room remains absent until cache expiry, contradicting the documented always-fresh `updatedWithinDays` membership. Treat `Missing` as a miss whenever a cutoff applies and add a windowed late-room integration test.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -259,40 +292,342 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
+		return nil, fmt.Errorf("decode subscription page rows: %w", err)
+	}
+	subByID := make(map[string]*model.Subscription, len(fullSubs))
+	for i := range fullSubs {
+		subByID[fullSubs[i].ID] = &fullSubs[i]
+	}
+	// Iterate the candidate rows (not cursor order) so the page keeps the
+	// sorted sequence.
+	for i := range rows {
+		b, ok := baselines[rows[i].sub.RoomID]
```
julianshen, line 610: _🏷️ consistency_ | _🟠 Warning_ | _`distributed-system`_

**A vanished room remains in a window from its cached timestamp.**

If a recent room's sort key is cached and its room document then disappears before subscription cleanup, `resolveSortKeys` accepts the cached in-window timestamp. The fresh baseline read returns no room, but this loop only drops an existing `Del-` baseline and still returns the subscription with zero-valued room fields. Thus a fresh read knows the room no longer satisfies the activity window but serves it until cache expiry. Pass the cutoff into `enrichListRows`, drop missing baselines or baselines whose current `lastMsgAt` fails it, update absent cache entries appropriately, and let `fillListPage` refill the gap.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 606: If a recent room's sort key is cached and its room document then disappears before subscription cleanup, `resolveSortKeys` accepts the cached in-window timestamp. The fresh baseline read returns no room, but this loop only drops an existing `Del-` baseline and still returns the subscription with zero-valued room fields. Thus a fresh read knows the room no longer satisfies the activity window but serves it until cache expiry. Pass the cutoff into `enrichListRows`, drop missing baselines or baselines whose current `lastMsgAt` fails it, update absent cache entries appropriately, and let `fillListPage` refill the gap.

```

</details>

user-service/mongorepo/sortkeycache.go
```diff
@@ -0,0 +1,67 @@
...
+		c.metrics.Miss(ctx)
+	}
+	return k, ok
+}
+
+func (c *sortKeyCache) add(roomID string, key roomSortKey) {
+	if c == nil {
+		return
+	}
+	c.lru.Add(roomID, key)
```
julianshen, line 63: _🏷️ ordering_ | _🟠 Warning_ | _`distributed-system`_

**Late cache writes can overwrite newer room activity.**

Cache updates are unconditional, so concurrent reads can regress state. Request A can read an old `lastMsgAt`, a message advances the room, request B reads and caches the newer value, and then delayed request A calls `Add` and overwrites it. A room displaced by the regressed key may not be page-enriched again, allowing ordering staleness to last a fresh TTL after the stale write and exceed the documented TTL bound from the message update. Include a room revision or `_updatedAt` in `roomSortKey` and perform a synchronized version check before replacement; at minimum, enforce the code's stated monotonic-`lastMsgAt` assumption atomically so an older activity key cannot replace a newer one.

<details>

```
In `user-service/mongorepo/sortkeycache.go` around line 66: Cache updates are unconditional, so concurrent reads can regress state. Request A can read an old `lastMsgAt`, a message advances the room, request B reads and caches the newer value, and then delayed request A calls `Add` and overwrites it. A room displaced by the regressed key may not be page-enriched again, allowing ordering staleness to last a fresh TTL after the stale write and exceed the documented TTL bound from the message update. Include a room revision or `_updatedAt` in `roomSortKey` and perform a synchronized version check before replacement; at minimum, enforce the code's stated monotonic-`lastMsgAt` assumption atomically so an older activity key cannot replace a newer one.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -259,40 +292,342 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
+// the TTL the way ordering may. Rooms absent from the batched read are
+// negative-cached as Missing so cross-site rooms don't re-probe Mongo on
+// every list (Missing hits are not demoted: there is no local doc to re-read,
+// and the old join dropped them under a window too).
+func (r *SubscriptionRepo) resolveSortKeys(ctx context.Context, subs []subLite, cutoff *time.Time) (map[string]roomSortKey, error) {
+	keys := make(map[string]roomSortKey, len(subs))
+	// The unique (roomId, u.account) index makes one account's roomIDs distinct
+	// already — no dedup needed.
+	var misses []string
+	for i := range subs {
```
julianshen, line 481: _🎯 correctness_ | _🟠 Warning_ | _`mongodb`_

**Negative cache entries can hide rooms entering an activity window**

The cutoff revalidation explicitly exempts cached `Missing` entries. If a room is absent during one windowed request and its document is subsequently inserted with an in-window `lastMsgAt`, later requests within the cache TTL retain `Missing`; `buildListRows` then drops the subscription before the fresh page read can discover the room. This contradicts the documented fresh `updatedWithinDays` membership. Re-read missing entries for windowed requests, or retain enough site metadata to exempt only rooms known to be permanently remote.

<details>

> ‼️ **IMPORTANT**

```suggestion
		if ok && cutoff != nil && (k.Missing || k.LastMsgAt == nil || k.LastMsgAt.Before(*cutoff)) {
```

</details>

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 473: The cutoff revalidation explicitly exempts cached `Missing` entries. If a room is absent during one windowed request and its document is subsequently inserted with an in-window `lastMsgAt`, later requests within the cache TTL retain `Missing`; `buildListRows` then drops the subscription before the fresh page read can discover the room. This contradicts the documented fresh `updatedWithinDays` membership. Re-read missing entries for windowed requests, or retain enough site metadata to exempt only rooms known to be permanently remote.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -259,40 +292,342 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
+		switch {
+		case a.sortAt != nil && b.sortAt != nil:
+			if !a.sortAt.Equal(*b.sortAt) {
+				return a.sortAt.After(*b.sortAt)
+			}
+		case a.sortAt != nil:
+			return true
+		case b.sortAt != nil:
+			return false
+		}
```
julianshen, line 412: _🏷️ pagination_ | _🟠 Warning_ | _`mongodb`_

**Equal sort keys make offset pagination nondeterministic**

`sort.SliceStable` only preserves the order returned by the phase-one MongoDB `Find`, which has no sort and therefore provides no stable ordering across queries or failovers. Two subscriptions with the same activity timestamp and name—such as missing or cross-site rooms sharing a display name—can swap between requests, causing offset pages to duplicate or skip rows even when data and cache entries are unchanged. Add the immutable subscription ID as the final ordering key.

<details>

> ‼️ **IMPORTANT**

```suggestion
		if a.sub.Name != b.sub.Name {
			return a.sub.Name < b.sub.Name
		}
		return a.sub.ID < b.sub.ID
```

</details>

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 405: `sort.SliceStable` only preserves the order returned by the phase-one MongoDB `Find`, which has no sort and therefore provides no stable ordering across queries or failovers. Two subscriptions with the same activity timestamp and name—such as missing or cross-site rooms sharing a display name—can swap between requests, causing offset pages to duplicate or skip rows even when data and cache entries are unchanged. Add the immutable subscription ID as the final ordering key.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -259,40 +292,342 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
+// the activity-window cutoff) and computes each row's sort position from the
+// resolved sort keys. Missing rooms (cross-site) are kept — their deletion
+// isn't visible here — except under the window, which requires a dated room.
+// Window membership is fresh: resolveSortKeys re-reads any cache hit that
+// fails the cutoff before the row gets here.
+func buildListRows(subs []subLite, keys map[string]roomSortKey, account string, favorite bool, cutoff *time.Time) []listRow {
+	rows := make([]listRow, 0, len(subs))
+	for i := range subs {
+		key := keys[subs[i].RoomID]
+		if !key.Missing && strings.HasPrefix(key.Name, deletedRoomNamePrefix) {
```
julianshen: _🏷️ cross-site-replication_ | _🟠 Warning_ | _`nats`_

**Local missing rooms are exposed as cross-site subscriptions**

The phased path no longer carries either the service's local site ID or the subscription's siteId, so every missing room is retained as if it were cross-site. If a local room deletion is applied before its subscription cleanup—an ordinary cross-subject replication interleaving—the subscription is returned with an empty room baseline; the previous site-aware filter explicitly dropped local missing rooms. Restore the local site ID in the repository, include siteId in subLite, and retain a missing room only when the subscription belongs to another site.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 369: The phased path no longer carries either the service's local site ID or the subscription's siteId, so every missing room is retained as if it were cross-site. If a local room deletion is applied before its subscription cleanup—an ordinary cross-subject replication interleaving—the subscription is returned with an empty room baseline; the previous site-aware filter explicitly dropped local missing rooms. Restore the local site ID in the repository, include siteId in subLite, and retain a missing room only when the subscription belongs to another site.

```

</details>

user-service/mongorepo/subscriptions.go
julianshen, line 484: _🏷️ cross-site-replication_ | _🟠 Warning_ | _`nats`_

**Negative cache entries hide rooms that enter an activity window**

When a subscription is observed before its room document, Missing is cached. If room replication then creates the room with an in-window lastMsgAt, a subsequent withinDays request accepts the Missing cache hit without re-reading Mongo, and buildListRows drops it because LastMsgAt is nil. The room remains absent until TTL expiry, contradicting the documented fresh window membership. Demote Missing hits to misses for windowed requests, just like stale out-of-window keys.

<details>

> ‼️ **IMPORTANT**

```suggestion
		if ok && cutoff != nil && (k.Missing || k.LastMsgAt == nil || k.LastMsgAt.Before(*cutoff)) {
```

</details>

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 476: When a subscription is observed before its room document, Missing is cached. If room replication then creates the room with an in-window lastMsgAt, a subsequent withinDays request accepts the Missing cache hit without re-reading Mongo, and buildListRows drops it because LastMsgAt is nil. The room remains absent until TTL expiry, contradicting the documented fresh window membership. Demote Missing hits to misses for windowed requests, just like stale out-of-window keys.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -259,40 +292,342 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
+		return nil, fmt.Errorf("find room baselines: %w", err)
+	}
+	var docs []roomBaseline
+	if err := cur.All(ctx, &docs); err != nil {
+		return nil, fmt.Errorf("decode room baselines: %w", err)
+	}
+	baselines := make(map[string]*roomBaseline, len(docs))
+	for i := range docs {
+		baselines[docs[i].ID] = &docs[i]
+		r.sortKeys.add(docs[i].ID, roomSortKey{
```
julianshen, line 587: _🏷️ nats-pagination_ | _🟠 Warning_ | _`nats`_

**Cache refreshes can make later NATS pages skip rooms**

Each subscription.list page is a separate NATS request, but page enrichment mutates the sort cache after the response order has already been chosen. For example, cached order X,Y becomes stale when Y receives a message; page 1 still returns X, while enriching its limit+1 batch refreshes both keys. Page 2 then sorts Y,X and applies offset 1, returning X again and permanently omitting Y from the drain. This happens even on one responder without another database update; per-instance caches across responders add another path. Dedupe cannot recover an omitted row. Use a stable cursor/snapshot or keyset pagination for the drain, or otherwise prevent sort-key changes from altering its ordering across requests.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 582: Each subscription.list page is a separate NATS request, but page enrichment mutates the sort cache after the response order has already been chosen. For example, cached order X,Y becomes stale when Y receives a message; page 1 still returns X, while enriching its limit+1 batch refreshes both keys. Page 2 then sorts Y,X and applies offset 1, returning X again and permanently omitting Y from the drain. This happens even on one responder without another database update; per-instance caches across responders add another path. Dedupe cannot recover an omitted row. Use a stable cursor/snapshot or keyset pagination for the drain, or otherwise prevent sort-key changes from altering its ordering across requests.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -259,40 +292,342 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
+// the room got its first in-window message — window MEMBERSHIP must not lag
+// the TTL the way ordering may. Rooms absent from the batched read are
+// negative-cached as Missing so cross-site rooms don't re-probe Mongo on
+// every list (Missing hits are not demoted: there is no local doc to re-read,
+// and the old join dropped them under a window too).
+func (r *SubscriptionRepo) resolveSortKeys(ctx context.Context, subs []subLite, cutoff *time.Time) (map[string]roomSortKey, error) {
+	keys := make(map[string]roomSortKey, len(subs))
+	// The unique (roomId, u.account) index makes one account's roomIDs distinct
+	// already — no dedup needed.
+	var misses []string
```
julianshen, line 480: _🎯 concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**The sort-key cache stampedes on concurrent misses.**

After startup or synchronized TTL expiry, concurrent list requests can all observe the same room IDs as missing before any request populates the cache. Each request then issues and decodes its own large `$in` query, multiplying MongoDB load by the in-flight handler count. Coalesce fills, for example with account-keyed singleflight plus a cache recheck inside the leader, or a batched per-room loading mechanism; avoid holding one global lock across unrelated loads.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 472: After startup or synchronized TTL expiry, concurrent list requests can all observe the same room IDs as missing before any request populates the cache. Each request then issues and decodes its own large `$in` query, multiplying MongoDB load by the in-flight handler count. Coalesce fills, for example with account-keyed singleflight plus a cache recheck inside the leader, or a batched per-room loading mechanism; avoid holding one global lock across unrelated loads.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -259,40 +292,342 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
+	// Degrade non-normalized values the way the old Mongo-side $skip/$limit
+	// path did (no panic): a negative offset reads from the start, a negative
+	// limit behaves like limit 0 (empty page, hasMore from the over-read).
+	offset := max(page.Offset, 0)
+	limit := max(page.Limit, 0)
+	candidates := rows[min(offset, int64(len(rows))):]
+	// Cap the over-read at the candidate count: collected can never exceed it,
+	// and the cap keeps limit+1 from overflowing (and the slice capacity sane)
+	// for a pathological MaxInt64 limit.
+	need := min(limit, int64(len(candidates))) + 1
```
julianshen, line 436: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Stale deletions can trigger hundreds of serial refill queries.**

When many cached candidates have since been soft-deleted, every refill batch yields no rows and this loop advances only `max(limit+1, 32)` candidates while `enrichListRows` performs two sequential MongoDB reads. At the default limit of 40, 10,000 stale deleted rows cause about 488 queries in one request, likely exceeding the handler timeout and amplifying load across concurrent requests. Filter deleted baselines before fetching subscriptions and grow subsequent zero-yield batches substantially, or compact remaining candidates with bounded bulk reads before continuing.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 428: When many cached candidates have since been soft-deleted, every refill batch yields no rows and this loop advances only `max(limit+1, 32)` candidates while `enrichListRows` performs two sequential MongoDB reads. At the default limit of 40, 10,000 stale deleted rows cause about 488 queries in one request, likely exceeding the handler timeout and amplifying load across concurrent requests. Filter deleted baselines before fetching subscriptions and grow subsequent zero-yield batches substantially, or compact remaining candidates with bounded bulk reads before continuing.

```

</details>

user-service/mongorepo/subscriptions.go
julianshen, line 481: _🎯 correctness_ | _🟠 Warning_ | _`silent-failure`_

**Negative cache hits bypass the fresh activity-window check.**

When a missing room is cached and its document subsequently appears with recent activity, this condition leaves the `Missing` hit valid. `buildListRows` then drops the row because `LastMsgAt` is nil, so the fresh baseline read never runs and the API silently omits an in-window subscription until the TTL expires. The new integration test already demonstrates that a room can appear after a negative entry is cached. Demote missing hits whenever a window applies, update the now-contradictory comment, and add a windowed regression test.

<details>

> ‼️ **IMPORTANT**

```suggestion
		if ok && cutoff != nil && (k.Missing || k.LastMsgAt == nil || k.LastMsgAt.Before(*cutoff)) {
```

</details>

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 473: When a missing room is cached and its document subsequently appears with recent activity, this condition leaves the `Missing` hit valid. `buildListRows` then drops the row because `LastMsgAt` is nil, so the fresh baseline read never runs and the API silently omits an in-window subscription until the TTL expires. The new integration test already demonstrates that a room can appear after a negative entry is cached. Demote missing hits whenever a window applies, update the now-contradictory comment, and add a windowed regression test.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -259,40 +292,342 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
+		if ok {
+			keys[id] = k
+		} else {
+			misses = append(misses, id)
+		}
+	}
+	if len(misses) == 0 {
+		return keys, nil
+	}
+	cur, err := r.rooms.Find(ctx, bson.M{"_id": bson.M{"$in": misses}},
```
julianshen, line 496: _🏷️ scalability_ | _🟠 Warning_ | _`tgd-review`_

**One unbounded $in query can exceed MongoDB's BSON limit.**

On a cold or disabled cache, `misses` contains every matching subscription room ID and is placed into one BSON filter. Subscription count is not bounded by the response page limit, so a sufficiently large account can exceed MongoDB's 16 MiB BSON document limit and make listing fail. Resolve misses in bounded chunks and merge the projected results into `keys`.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 488: On a cold or disabled cache, `misses` contains every matching subscription room ID and is placed into one BSON filter. Subscription count is not bounded by the response page limit, so a sufficiently large account can exceed MongoDB's 16 MiB BSON document limit and make listing fail. Resolve misses in bounded chunks and merge the projected results into `keys`.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -259,40 +292,342 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
+// every list (Missing hits are not demoted: there is no local doc to re-read,
+// and the old join dropped them under a window too).
+func (r *SubscriptionRepo) resolveSortKeys(ctx context.Context, subs []subLite, cutoff *time.Time) (map[string]roomSortKey, error) {
+	keys := make(map[string]roomSortKey, len(subs))
+	// The unique (roomId, u.account) index makes one account's roomIDs distinct
+	// already — no dedup needed.
+	var misses []string
+	for i := range subs {
+		id := subs[i].RoomID
+		k, ok := r.sortKeys.get(ctx, id)
```
julianshen, line 483: _🏷️ consistency_ | _🔴 Blocking_ | _`distributed-system`_

**Negative cache hides newly materialized rooms from windowed lists.**

The new negative-cache integration test establishes this reachable sequence: a list observes no local room and caches Missing, then that local room document is inserted. If a subsequent rooms request uses withinDays before the TTL expires, this condition exempts the Missing hit from refresh; buildListRows sees no lastMsgAt and drops the row, so enrichListRows never reads the newly created room and cannot repair the cache. This violates the documented fresh-membership contract and hides a recently active room for up to the cache TTL after cross-service creation or replication ordering. Re-read Missing entries whenever a cutoff applies, or retain site identity and negative-cache only rooms proven to be cross-site.

<details>

> ‼️ **IMPORTANT**

```suggestion
		if ok && cutoff != nil && (k.Missing || k.LastMsgAt == nil || k.LastMsgAt.Before(*cutoff)) {
```

</details>

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 475: The new negative-cache integration test establishes this reachable sequence: a list observes no local room and caches Missing, then that local room document is inserted. If a subsequent rooms request uses withinDays before the TTL expires, this condition exempts the Missing hit from refresh; buildListRows sees no lastMsgAt and drops the row, so enrichListRows never reads the newly created room and cannot repair the cache. This violates the documented fresh-membership contract and hides a recently active room for up to the cache TTL after cross-service creation or replication ordering. Re-read Missing entries whenever a cutoff applies, or retain site identity and negative-cache only rooms proven to be cross-site.

```

</details>

user-service/mongorepo/subscriptions.go
julianshen: _🏷️ data-modeling_ | _🔴 Blocking_ | _`mongodb`_

**Missing local rooms are incorrectly treated as cross-site.**

The previous pipeline used the repository site ID to distinguish cross-site subscriptions from dangling local subscriptions and dropped the latter. The new room lookup represents every absent room as `Missing`, and this condition retains all such rows; `enrichListRows` then returns them with an empty room baseline. If a local room document is deleted or room/subscription cleanup is temporarily inconsistent, its subscription reappears as if it were cross-site. Retain the local site ID, include `siteId` in `subLite`, and drop `Missing` rows whose subscription belongs to the local site.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 369: The previous pipeline used the repository site ID to distinguish cross-site subscriptions from dangling local subscriptions and dropped the latter. The new room lookup represents every absent room as `Missing`, and this condition retains all such rows; `enrichListRows` then returns them with an empty room baseline. If a local room document is deleted or room/subscription cleanup is temporarily inconsistent, its subscription reappears as if it were cross-site. Retain the local site ID, include `siteId` in `subLite`, and drop `Missing` rows whose subscription belongs to the local site.

```

</details>

user-service/mongorepo/sortkeycache.go
julianshen, line 63: _🏷️ ordering_ | _🟠 Warning_ | _`distributed-system`_

**Late cache fills can overwrite newer room sort keys.**

Two concurrent misses can complete out of order: request R1 reads the old room key and stalls, a message advances lastMsgAt, request R2 reads and caches the new key, then R1 resumes and blindly replaces it with the old key. The late Add also starts a fresh TTL, so ordering can regress and remain stale beyond the documented TTL measured from the message update. This is reachable both while resolving all sort keys and while caching page baselines. Include a monotonic room revision such as _updatedAt in the projection and conditionally replace only older cached revisions, with a concurrency test covering the interleaving.

<details>

```
In `user-service/mongorepo/sortkeycache.go` around line 66: Two concurrent misses can complete out of order: request R1 reads the old room key and stalls, a message advances lastMsgAt, request R2 reads and caches the new key, then R1 resumes and blindly replaces it with the old key. The late Add also starts a fresh TTL, so ordering can regress and remain stale beyond the documented TTL measured from the message update. This is reachable both while resolving all sort keys and while caching page baselines. Include a monotonic room revision such as _updatedAt in the projection and conditionally replace only older cached revisions, with a concurrency test covering the interleaving.

```

</details>

user-service/mongorepo/subscriptions.go
julianshen, line 483: _🏷️ consistency_ | _🟠 Warning_ | _`mongodb`_

**Negative cache hits can make withinDays membership stale.**

`TestAggregateSubscriptions_NegativeCachesMissingRooms_Integration` establishes that a room document can appear after its absence was cached. For a `withinDays` request, this guard exempts `Missing` hits from refresh, so `buildListRows` rejects the row for lacking `lastMsgAt` and the fresh page read is never reached. A newly available room with recent activity therefore remains invisible until the default 15-second TTL, contrary to the documented fresh `updatedWithinDays` membership. Re-probe cached missing rooms when a window cutoff applies.

<details>

> ‼️ **IMPORTANT**

```suggestion
		if ok && cutoff != nil && (k.Missing || k.LastMsgAt == nil || k.LastMsgAt.Before(*cutoff)) {
```

</details>

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 475: `TestAggregateSubscriptions_NegativeCachesMissingRooms_Integration` establishes that a room document can appear after its absence was cached. For a `withinDays` request, this guard exempts `Missing` hits from refresh, so `buildListRows` rejects the row for lacking `lastMsgAt` and the fresh page read is never reached. A newly available room with recent activity therefore remains invisible until the default 15-second TTL, contrary to the documented fresh `updatedWithinDays` membership. Re-probe cached missing rooms when a window cutoff applies.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -259,40 +292,342 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
+		}
+		if ok {
+			keys[id] = k
+		} else {
+			misses = append(misses, id)
+		}
+	}
+	if len(misses) == 0 {
+		return keys, nil
+	}
```
julianshen, line 495: _🏷️ scalability_ | _🟠 Warning_ | _`mongodb`_

**A cold cache can create an oversized MongoDB find command.**

On a cold or expired cache, `misses` contains one room ID for every matching subscription because phase one reads the account's complete subscription set regardless of the requested page size. Serializing all IDs into one `$in` can exceed MongoDB's BSON command-size limit for an account with sufficiently many memberships, causing even a small first-page request to fail. Fetch room sort keys in bounded `$in` batches and merge the results before negative-caching IDs absent from each batch.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 487: On a cold or expired cache, `misses` contains one room ID for every matching subscription because phase one reads the account's complete subscription set regardless of the requested page size. Serializing all IDs into one `$in` can exceed MongoDB's BSON command-size limit for an account with sufficiently many memberships, causing even a small first-page request to fail. Fetch room sort keys in bounded `$in` batches and merge the results before negative-caching IDs absent from each batch.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -259,40 +292,342 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
-	// Scaling ceiling: the room join + activity sort run over the full matched set before
-	// the skip/limit page (the sort key lives on the joined room, so it can't be pushed past
-	// the lookup). Fine at realistic per-account sub counts; the fix for very large accounts is
-	// denormalizing room activity onto the subscription — a write-side change tracked separately.
-	// Primary: the subscription list must reflect a just-changed subscription immediately.
-	return r.enriched.AggregatePagedHasMore(ctx, pipeline, page)
+	// Primary throughout the phased path: the subscription list must reflect a
+	// just-changed subscription immediately, and the page rows are enriched from
+	// fresh room reads — only the cached sort keys may lag.
+	cur, err := r.subscriptions.Raw().Find(ctx, match,
```
julianshen, line 295: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Every page materializes and sorts all matching subscriptions.**

The page limit is applied only after Find().All loads every matching subscription, resolveSortKeys allocates O(N) maps/slices, and sortListRows performs an O(N log N) in-process sort. On a cold cache, the same request also creates one $in containing every room ID. A large or imported account therefore consumes memory and CPU proportional to its complete subscription history even for limit=40; concurrent requests multiply that heap usage and can produce oversized Mongo commands or process-level GC/OOM pressure. Unlike the previous pipeline, this also moves the unbounded pre-page working set into each service instance. Keep sorting and paging in Mongo until the activity key can be denormalized and indexed on subscriptions, or enforce a hard supported candidate bound and batch the room lookup.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 298: The page limit is applied only after Find().All loads every matching subscription, resolveSortKeys allocates O(N) maps/slices, and sortListRows performs an O(N log N) in-process sort. On a cold cache, the same request also creates one $in containing every room ID. A large or imported account therefore consumes memory and CPU proportional to its complete subscription history even for limit=40; concurrent requests multiply that heap usage and can produce oversized Mongo commands or process-level GC/OOM pressure. Unlike the previous pipeline, this also moves the unbounded pre-page working set into each service instance. Keep sorting and paging in Mongo until the activity key can be denormalized and indexed on subscriptions, or enforce a hard supported candidate bound and batch the room lookup.

```

</details>

user-service/mongorepo/subscriptions.go
julianshen, line 483: _🎯 concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Concurrent cache misses stampede the rooms collection.**

Entries populated by one batched resolution expire together. After startup or TTL expiry, concurrent handlers for the same popular account can all observe the same misses before any handler reaches add, then each issue an identical potentially large rooms $in query. The cutoff branch further turns every cached out-of-window positive entry into a miss, making this pattern recur for windowed lists containing dormant rooms. The resulting duplicate primary reads can create periodic load spikes and latency amplification up to the handler-concurrency limit. Coalesce in-flight refreshes per account or miss batch with singleflight, recheck the cache before querying, and let canceled waiters stop waiting without canceling a refresh still needed by other callers.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 475: Entries populated by one batched resolution expire together. After startup or TTL expiry, concurrent handlers for the same popular account can all observe the same misses before any handler reaches add, then each issue an identical potentially large rooms $in query. The cutoff branch further turns every cached out-of-window positive entry into a miss, making this pattern recur for windowed lists containing dormant rooms. The resulting duplicate primary reads can create periodic load spikes and latency amplification up to the handler-concurrency limit. Coalesce in-flight refreshes per account or miss batch with singleflight, recheck the cache before querying, and let canceled waiters stop waiting without canceling a refresh still needed by other callers.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -259,40 +292,342 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
+		return out, nil
+	}
+	roomIDs := make([]string, 0, len(rows))
+	subIDs := make([]string, 0, len(rows))
+	for i := range rows {
+		roomIDs = append(roomIDs, rows[i].sub.RoomID)
+		subIDs = append(subIDs, rows[i].sub.ID)
+	}
+	cur, err := r.rooms.Find(ctx, bson.M{"_id": bson.M{"$in": roomIDs}},
+		options.Find().SetProjection(roomBaselineProjection()))
```
julianshen, line 576: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Independent enrichment reads are serialized.**

The room-baseline Find/All completes before the subscription-page Find starts, although both queries depend only on roomIDs, subIDs, and match, all of which are already available. Every non-empty page therefore pays the two Mongo round trips in series; under pool or database pressure their queueing delays also accumulate. Run these reads concurrently with a cancellation-aware errgroup, or combine them into one page-sized aggregation, then build the output after both results succeed.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 571: The room-baseline Find/All completes before the subscription-page Find starts, although both queries depend only on roomIDs, subIDs, and match, all of which are already available. Every non-empty page therefore pays the two Mongo round trips in series; under pool or database pressure their queueing delays also accumulate. Run these reads concurrently with a cancellation-aware errgroup, or combine them into one page-sized aggregation, then build the output after both results succeed.

```

</details>

user-service/mongorepo/subscriptions.go
julianshen, line 483: _🏷️ data freshness_ | _🟠 Warning_ | _`silent-failure`_

**Negative cache hits can silently omit newly created rooms.**

When a windowed request caches a room as Missing, and that room document subsequently appears with an in-window lastMsgAt, this condition explicitly prevents re-reading the Missing entry. buildListRows then drops the subscription because its cached lastMsgAt is nil, returning a successful but incomplete page and incorrect hasMore despite the documented fresh updatedWithinDays membership. Demote Missing hits to misses for windowed requests as well.

<details>

> ‼️ **IMPORTANT**

```suggestion
		if ok && cutoff != nil && (k.Missing || k.LastMsgAt == nil || k.LastMsgAt.Before(*cutoff)) {
```

</details>

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 475: When a windowed request caches a room as Missing, and that room document subsequently appears with an in-window lastMsgAt, this condition explicitly prevents re-reading the Missing entry. buildListRows then drops the subscription because its cached lastMsgAt is nil, returning a successful but incomplete page and incorrect hasMore despite the documented fresh updatedWithinDays membership. Demote Missing hits to misses for windowed requests as well.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -259,40 +292,342 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
+		b, ok := baselines[rows[i].sub.RoomID]
+		if ok && strings.HasPrefix(b.Name, deletedRoomNamePrefix) {
+			continue
+		}
+		sub, found := subByID[rows[i].sub.ID]
+		if !found {
+			continue
+		}
+		var es model.EnrichedSubscription
+		es.Subscription = *sub
```
julianshen, line 616: _🏷️ data integrity_ | _🟠 Warning_ | _`silent-failure`_

**Missing local rooms are returned as successful zero-valued rows.**

If the fresh room-baseline query does not return a local room, the code still appends its subscription with empty room metadata. This masks an orphaned or hard-deleted local room as a valid successful result. The previous repository contract distinguished cross-site subscriptions from local subscriptions and dropped local rows with missing rooms; removing siteID makes that distinction impossible. Restore the local site identifier and skip a row when its baseline is absent and its subscription belongs to the local site, while retaining genuinely cross-site rows.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 615: If the fresh room-baseline query does not return a local room, the code still appends its subscription with empty room metadata. This masks an orphaned or hard-deleted local room as a valid successful result. The previous repository contract distinguished cross-site subscriptions from local subscriptions and dropped local rows with missing rooms; removing siteID makes that distinction impossible. Restore the local site identifier and skip a row when its baseline is absent and its subscription belongs to the local site, while retaining genuinely cross-site rows.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -259,40 +292,342 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
+// negative-cached as Missing so cross-site rooms don't re-probe Mongo on
+// every list (Missing hits are not demoted: there is no local doc to re-read,
+// and the old join dropped them under a window too).
+func (r *SubscriptionRepo) resolveSortKeys(ctx context.Context, subs []subLite, cutoff *time.Time) (map[string]roomSortKey, error) {
+	keys := make(map[string]roomSortKey, len(subs))
+	// The unique (roomId, u.account) index makes one account's roomIDs distinct
+	// already — no dedup needed.
+	var misses []string
+	for i := range subs {
+		id := subs[i].RoomID
```
julianshen, line 482: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Negative cache hits can make window membership stale.**

A room cached as Missing is never re-read for an updatedWithinDays request. If its room document subsequently appears with recent activity, buildListRows still sees a nil LastMsgAt and excludes it until the cache TTL expires. This contradicts the documented guarantee that updatedWithinDays membership is fresh; the existing negative-cache integration test even demonstrates that missing rooms can later appear. Demote Missing hits to misses whenever a cutoff applies.

<details>

> ‼️ **IMPORTANT**

```suggestion
		if ok && cutoff != nil && (k.Missing || k.LastMsgAt == nil || k.LastMsgAt.Before(*cutoff)) {
```

</details>

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 474: A room cached as Missing is never re-read for an updatedWithinDays request. If its room document subsequently appears with recent activity, buildListRows still sees a nil LastMsgAt and excludes it until the cache TTL expires. This contradicts the documented guarantee that updatedWithinDays membership is fresh; the existing negative-cache integration test even demonstrates that missing rooms can later appear. Demote Missing hits to misses whenever a cutoff applies.

```

</details>

user-service/mongorepo/subscriptions.go
julianshen: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Missing local rooms are now returned as cross-site rows.**

Missing only means that no room document was found, but this branch treats every missing room as cross-site and keeps it for unwindowed lists. The previous repository contract accepted the local site ID specifically to keep cross-site missing rows while dropping local subscriptions whose room was missing; this change removes that identity entirely. Orphaned local subscriptions will therefore be returned with zero-valued room enrichment. Retain the local site ID, project each subscription's siteId in subLite, and keep a missing room only when the subscription belongs to another site.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 369: Missing only means that no room document was found, but this branch treats every missing room as cross-site and keeps it for unwindowed lists. The previous repository contract accepted the local site ID specifically to keep cross-site missing rows while dropping local subscriptions whose room was missing; this change removes that identity entirely. Orphaned local subscriptions will therefore be returned with zero-valued room enrichment. Retain the local site ID, project each subscription's siteId in subLite, and keep a missing room only when the subscription belongs to another site.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -267,52 +303,361 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
+		case a.sortAt != nil && b.sortAt != nil:
+			if !a.sortAt.Equal(*b.sortAt) {
+				return a.sortAt.After(*b.sortAt)
+			}
+		case a.sortAt != nil:
+			return true
+		case b.sortAt != nil:
+			return false
+		}
+		return a.sub.Name < b.sub.Name
```
julianshen: _🏷️ ordering_ | _🟠 Warning_ | _`distributed-system`_

**Equal sort keys make offset pagination nondeterministic.**

When two subscriptions have equal activity timestamps and names, the stable sort preserves the order of the unsorted MongoDB `Find`. That order can change across requests or after failover, so page one can use A,B while page two uses B,A; applying the offset then duplicates one row and omits the other without any data change. Use the immutable subscription ID as the final ordering key.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 441: When two subscriptions have equal activity timestamps and names, the stable sort preserves the order of the unsorted MongoDB `Find`. That order can change across requests or after failover, so page one can use A,B while page two uses B,A; applying the offset then duplicates one row and omits the other without any data change. Use the immutable subscription ID as the final ordering key.

```

</details>

julianshen (commented):
tGD inline review

user-service/mongorepo/subscriptions_test.go
```diff
@@ -828,72 +828,30 @@ func TestAppSubscriptionRoundTrip_Integration(t *testing.T) {
 	})
 }
 
-// The badge path must return exactly its five fields, across local, empty and cross-site rooms.
-func TestGetActiveSubscriptions_ProjectsBadgeFields_Integration(t *testing.T) {
-	r, db := newTestSubscriptionRepo(t)
+// Every persisted subscription field must survive the list path's inclusion
```
julianshen, line 832: _🧪 tests_ | _🟠 Warning_ | _`tgd-review`_

**An unrelated projection regression test was removed.**

This new list-path test replaces TestGetActiveSubscriptions_ProjectsBadgeFields even though the active-subscription behavior remains. The removed test verified exact output fields across local, empty, and cross-site rooms and specifically guarded against leaking room encryption keys. Restore that test separately rather than trading its coverage for the new origin round-trip test.

<details>

```
In `user-service/mongorepo/subscriptions_test.go` around line 831: This new list-path test replaces TestGetActiveSubscriptions_ProjectsBadgeFields even though the active-subscription behavior remains. The removed test verified exact output fields across local, empty, and cross-site rooms and specifically guarded against leaking room encryption keys. Restore that test separately rather than trading its coverage for the new origin round-trip test.

```

</details>

julianshen (commented):
tGD inline review

user-service/mongorepo/subscriptions_bench_test.go
```diff
@@ -0,0 +1,137 @@
...
+// runs. BenchmarkSubscriptionList carries the same name and sub-benchmark
+// names on both branches so benchstat can pair them.
+const benchAccount = "bench-user"
+
+// newBenchRepo builds the repo under test over an isolated database.
+func newBenchRepo(b *testing.B) (*SubscriptionRepo, *mongo.Database) {
+	b.Helper()
+	db := testutil.MongoDB(b, "user-service-bench")
+	// BRANCH-SPECIFIC. On main this line reads:
+	//     r := NewSubscriptionRepo(db, "site-a")
```
julianshen, line 35: _📝 documentation_ | _🔵 Suggestion_ | _`tgd-review`_

**The documented main-branch benchmark constructor does not compile.**

The pre-change constructor accepts db followed by Option values, not a site-ID string. Following this branch-comparison instruction on main therefore causes a type error. The documented main-branch call should pass only db.

<details>

```
In `user-service/mongorepo/subscriptions_bench_test.go` around line 35: The pre-change constructor accepts db followed by Option values, not a site-ID string. Following this branch-comparison instruction on main therefore causes a type error. The documented main-branch call should pass only db.

```

</details>

julianshen (commented):
tGD inline review

user-service/mongorepo/subscriptions.go
```diff
@@ -256,25 +285,34 @@ func (r *SubscriptionRepo) AggregateSubscriptions(ctx context.Context, account,
...
-	return r.enriched.AggregatePagedHasMore(ctx, pipeline, page)
+	// All primary: the list must show a subscription the caller just changed,
+	// and the page's rows come from fresh room reads. Only sort keys may lag.
+	cur, err := r.subscriptions.Raw().Find(ctx, match,
+		options.Find().SetProjection(subscriptionLiteProjection()))
+	if err != nil {
+		return zero, fmt.Errorf("find subscriptions: %w", err)
+	}
+	var subs []subLite
+	if err := cur.All(ctx, &subs); err != nil {
```
julianshen, line 301: _⚡ memory-exhaustion_ | _🟠 Warning_ | _`oom`_

**Each page loads every matching subscription into the service heap**

The prior OOM concern remains: `cur.All` materializes the account's complete matching subscription set even when the requested page is only 40 rows. The same request then allocates `keys`, `misses`, room-key `docs`, and `rows` proportional to that full set; on a cold cache several of these coexist, and concurrent page requests multiply the peak heap. Subscription cardinality is not bounded by the page limit, so a long-lived or imported account with very many memberships can cause GC thrashing, container eviction, or process OOM. Keep sorting/paging in Mongo until the activity key is denormalized and indexed, or immediately enforce a hard candidate limit by reading at most limit+1 candidates and returning a resource-exhausted error before allocating the additional maps and slices.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 301: The prior OOM concern remains: `cur.All` materializes the account's complete matching subscription set even when the requested page is only 40 rows. The same request then allocates `keys`, `misses`, room-key `docs`, and `rows` proportional to that full set; on a cold cache several of these coexist, and concurrent page requests multiply the peak heap. Subscription cardinality is not bounded by the page limit, so a long-lived or imported account with very many memberships can cause GC thrashing, container eviction, or process OOM. Keep sorting/paging in Mongo until the activity key is denormalized and indexed, or immediately enforce a hard candidate limit by reading at most limit+1 candidates and returning a resource-exhausted error before allocating the additional maps and slices.

```

</details>

julianshen (commented):
tGD inline review

PR #205, opened 2026-08-07
fix(thread msg): return number of replies in thread list rpc

amdla:
Returns each thread's reply count from the cross-site thread inbox
(`chat.user.{account}.request.user.{siteID}.thread.list`) as an explicit
`tcount` field on every `ThreadListItem`.

## Why

The count was already in the response — but only inside `parentMessage`, which
this RPC carries as an opaque `json.RawMessage`: history-service pre-marshals it
and user-service forwards the bytes verbatim, never decoding (decoding would hit
`cassandra.Message.Reactions`, a struct-keyed map with no JSON decoder). So no
server stage could read the count, Go consumers needed a projection decode to
reach it, and both hops were documented *Optional* — `parentMessage` on the item,
`tcount` on the Message. It genuinely went missing too: the oplog transformer
writes `thread_rooms`/`thread_subscriptions` straight to Mongo and never touches
Cassandra, so migrated threads had no `tcount` column at all and clients could
not tell *unknown* from *zero*.

## What changed

`ThreadListItem` gains `TCount int` (`json:"tcount"`, **no** `omitempty`),
populated in `buildThreadItems` from the parent the page **already hydrates**:

```go
if parent.TCount != nil {
    item.TCount = *parent.TCount
}
```

- **Zero added reads, writes, or schema changes.** `tcount` is already in the
  Cassandra `baseColumns` projection, so this is a nil-check plus an 8-byte copy
  per row.
- **Plain `int`, always serialized** — clients never branch on `undefined`.
  Accepted cost: `0` conflates "all replies deleted", "never counted" (migrated
  threads, or briefly a just-created thread), and a not-yet-upgraded peer site.
- **Cap passes through untouched** — `pkg/threadcount.Cap` (99 means "99 or
  more"); no arithmetic anywhere in this change.
- **`parentMessage.tcount` is retained** for backward compatibility.
- **user-service production code is unchanged** — the aggregator already copies
  items verbatim; tests pin that property rather than code enforcing it.

Item-level `tcount` and the embedded `parentMessage.tcount` come from the same
in-memory parent, so they cannot disagree within a row.

### Rejected alternatives

- **Denormalize a count onto `thread_rooms`** — uncapped and Mongo-only, but adds
  a write to the reply hot path, creates a second source of truth that can drift
  from `tcount`, and needs a backfill.
- **Count from Cassandra at read time** — up to ~100 bounded partition scans per
  page per site; precisely what the cached `tcount` exists to avoid.

## Rollout note

During a mixed-version deploy, rows from a not-yet-upgraded leaf site decode to
`tcount: 0` (their leaf omits the field), and a not-yet-upgraded aggregator drops
the key entirely. Self-heals as sites upgrade; documented in the field table.
Clients that keep reading `parentMessage.tcount` are unaffected.

## Testing

Every link in the chain is covered at the level where it can actually fail:

| Link | Coverage |
|---|---|
| Mongo pipeline → row keys | integration (`mongorepo/threadsubscription_test.go`) |
| `tcount` written to Cassandra | integration (`cassrepo/write_integration_test.go`) |
| `tcount` scanned back out | integration — **new** (`cassrepo/messages_by_id_integration_test.go`) |
| Lift onto the item | unit — `4` / `nil` / `0` / `99`, plus a parent-body key-presence pin |
| Aggregator pass-through | unit — merge, global sort, all three enrichment branches, degraded lookups |

The scan test closes a real blind spot: everything above the repository mocks it,
so a broken `tcount` scan would have reported `0` for every thread with the whole
suite green. Each new assertion was mutation-checked (deliberately wrong
expectation → observed failure → restored).

Also pinned: a zero `tcount` serializes rather than being omitted, and a payload
lacking the key decodes to `0` (the documented old-leaf behavior).

Locally green: `make lint` (0 issues), unit suites for `pkg/model`,
`history-service`, `user-service` (all `-race`), and the full integration suites
for `history-service` and `user-service` against real Mongo/Cassandra/NATS
containers. `make sast`: gosec clean and the repo-local semgrep rules clean;
govulncheck and the semgrep registry rulesets are network-blocked in the dev
sandbox and run here in CI.

## Docs

`docs/client-api.md` — `tcount` row in the **ThreadListItem** table (verified
field-for-field against the struct: 14 fields, same order, optionality matching
`omitempty`), both JSON examples updated, `parentMessage` note de-staled. The
derived views need no edit — `docs/client-api/request-reply.md` links to the
canonical schema rather than restating it, and `thread.list` emits no events.

`docs/design/user-thread-list.md` — the original "reply count is **not** a
separate field" decision is marked superseded, with a pointer to the spec.

Unrelated drive-by, found while verifying: the second thread-list JSON example
carried `"unread": false` with no `lastSeenAt`, which `threadUnread` can never
emit (a nil `lastSeenAt` is always unread). Fixed.


pkg/model/threadlist.go
```diff
@@ -19,9 +19,13 @@ type ThreadListItem struct {
 	Unread     bool   `json:"unread"               bson:"unread"` // lastMsgAt > lastSeenAt
 
 	// LastMsgAt is the thread's last-activity time (UTC ms) and the global sort
-	// key for the inbox. Reply count rides on ParentMessage.TCount.
+	// key for the inbox. Reply count is surfaced as TCount below.
 	LastMsgAt int64 `json:"lastMsgAt" bson:"lastMsgAt"`
 
+	// TCount is the parent's non-deleted reply count, capped at pkg/threadcount.Cap
+	// (99 = "99 or more"). Always serialized; 0 when the column was never written.
```
mliu33, line 26: Do we have this 99 cap in old behavior ?

mliu33 (approved):
Great work, thanks!

Issue #204, opened 2026-08-07
Feature - Priority Contacts Notification Plan


evelynchang5525:
# Notification Settings — Backend Implementation Plan

## Current State

| Area | Status |
|---|---|
| `UserSettings` fields (`muteAllNotifications`, `alwaysAllowPriorityNotifications`, `showPreviewsInNotifications`, `showNotificationsInCall`) | **Stored** in `users` collection via `user-service`,  round-tripped to client — but **never evaluated** by `notification-worker` |
| `priorityContacts` | **Does not exist** — no model, no storage, no API handler |
| `notification-worker` push gate | Only checks per-room `muted` flag + presence status (`busy`/`in-call`); ignores all `UserSettings` |

Two things to build:

1. **Store `priorityContacts`** — model + persistence + CRUD API (user-service)
2. **Make `notification-worker` aware of settings + priority contacts** — fetch at push-decision time and apply the filtering logic

---

## Part 1 — Storing Priority Contacts

### 1-A. Data Storage Decision

**Embed `priorityContacts` as an array inside the existing `users` document** (`settings.priorityContacts`), rather than a separate collection.

Rationale:
- Priority contacts are owned by one user, accessed always in the context of that user's settings, and capped at 30 entries — a perfect fit for embedding.
- The existing `UpdateUserSettings` pattern (`$set settings.<field>`) already handles the full-replace semantics we need for the list. Add/Remove are issued as separate NATS RPCs that use `$addToSet`/ `$pull` directly at the Mongo layer — both are atomic, no read-modify-write race.
- The 30-entry cap is enforced at the handler level before the Mongo write (count check + guard), consistent with how other business rules are applied in `user-service`.
- Avoids a new collection, new indexes, and a join on every `GetSettings` call.
- Document size impact is negligible: 30 entries × ~30 bytes each ≈ < 1 KB added to the user document.

**Mongo operations used per action:**

| Action | Mongo op | Atomic? |
|---|---|---|
| Get list | `FindOne` project `settings.priorityContacts` | — |
| Add contact | `$addToSet { "settings.priorityContacts": entry }` | Yes — no race |
| Remove contact | `$pull { "settings.priorityContacts": { account: X } }` | Yes — no race |
| Count (cap check) | `$size` aggregation or application-level `len` after read | App-level before write |

### 1-B. `pkg/model` Changes
**Add `PriorityContacts` to `UserSettings`** — stored in `users` via bson, included in every `settings.get` response:

```go
[...]

PR #213, opened 2026-08-08
feat(user-service): priority contacts storage and API (Spec 1)

evelynchang5525:
## Summary

Implements **Spec 1 of 2** for priority contacts (issue #204): storage, CRUD API, and cross-site replication. Users can store up to 30 priority contacts, read them back enriched for display, and every site receives the list. Push-behavior changes are **Spec 2** — nothing consumes the list yet, which is intentional.

Design spec: `docs/superpowers/specs/2026-08-08-priority-contacts-storage-api-design.md`

## Key changes

**Model & infrastructure**
- `PriorityContacts []string` on `pkg/model.UserSettings`, with two deliberate omissions: it is **not** checked by `IsEmpty()` (so a `settings.set` carrying only this field correctly falls through to `bad_request`), and **not** referenced by `UpdateUserSettings` (only the dedicated RPCs write it).
- Two reason codes in `pkg/errcode/codes_user.go`: `priority_contact_limit`, `priority_contact_not_found`.
- Six subject builders in `pkg/subject/subject.go` (concrete + `Pattern` forms) for the three RPCs.

**Wire types** (`user-service/models/prioritycontacts.go`)
- `PriorityContactMutateRequest`, `PriorityContactItem`, `PriorityContactUser`, `PriorityContactApp`, `PriorityContactsResponse`.
- `PriorityContactItem` carries an explicit `type` discriminator (`"user"` / `"bot"`) that selects which nested object is present — `user` and `app` are mutually exclusive with each other. The discriminator exists so clients never infer a row's kind from a `.bot` suffix or from which fields came back populated.

**Repository** (`user-service/mongorepo/users.go`) — five new methods
- `GetUserPriorityContacts(ctx, account) (*model.User, error)` — returns the user, not a bare slice, so the service can tell a missing user (`not_found`) from an empty list (`{"contacts": []}`).
- `UserExists`, `GetPriorityContactUsers` (reads); `AddPriorityContact`, `RemovePriorityContact` (writes).
- `UpdateUserSettings` gained an `at time.Time` parameter so the origin write stamps `settingsUpdatedAt`.

**Service** (`user-service/service/prioritycontacts.go`) — three handlers
- `GetPriorityContacts`, `AddPriorityContact`, `RemovePriorityContact`.
- Enrichment partitions accounts by `model.IsBot`, then runs two **sequential** lookups (HR and apps) over at most 30 accounts, building rows in stored order. A lookup failure degrades the affected rows to `account` + `type` rather than failing the call — same posture as the existing thread-list enrichment.
- `publishSettingsFanouts` helper in `settings.go` publishes both fanouts off one shared timestamp, so no caller can emit the client event while silently skipping cross-site replication.

**Cross-site replication**
- Add and remove each publish two fanouts off one timestamp: the client event `chat.user.{account}.event.settings.update`, and one cross-site event per remote peer on `chat.inbox.{destSiteID}.external.user_settings_updated`.
- Both carry the **full** `settings` sub-document, because `inbox-worker` applies it as a whole-object `$set` — a partial payload would wipe unrelated settings at every remote site.

## Two pre-existing bugs fixed along the way

Both are in the settings replication lane, which this feature is the first to depend on heavily.

1. **The origin never stamped `settingsUpdatedAt`.** `inbox-worker`'s high-water guard has an `$exists: false` fallback, so with no origin stamp that branch always matched and a *stale* remote event could overwrite a *newer* local edit. `UpdateUserChatlist` already stamped for exactly this reason; settings now does too.
2. **Same-millisecond events were dropped.** The guard was `$lt`, so of two writes sharing a millisecond the second was discarded at every remote site. Relaxed to `$lte`, safe because the apply is an idempotent whole-object replace.

## Correctness notes worth a reviewer's attention

- **The 30-cap is enforced inside the update filter** (`$expr` + `$size` + `$ifNull`), not by a read-then-check-then-write. A driver-side guard would let two concurrent adds both observe 29 and overshoot.
- **Add is idempotent, with a subtlety:** a duplicate add *at* the cap succeeds without publishing (the write misses, the re-read finds the contact present, it returns early). A duplicate add *under* the cap still publishes — the list is unchanged but `settingsUpdatedAt` bumps and both fanouts fire.
- **Remove deliberately has no existence check** — removing an account that no longer exists is the cleanup case we want to permit. Self-removal is allowed; only self-*add* is rejected.
- **`GetPriorityContactUsers` is deliberately not active-filtered**, while `UserExists` is: strict at add time (you cannot add a deactivated user), lenient at render time (a contact deactivated *after* being added still renders their name). Only an account with no user document at all degrades to `account` + `type`.

## Known follow-up, before Spec 2

This branch is the first time `users.settings` for one account has more than one RPC entry point. Because `natsrouter` dispatches each request on its own goroutine, MongoDB's write-serialization order need not match the order handlers captured their timestamps — so a physically-later write can carry an earlier stamp and be silently rejected by a remote site's `$lte` guard, leaving that replica stale until the next settings write.

The design spec accepted a narrower version of this under a single-writer model; add and remove triple the writers and widen the window. It is invisible today because nothing reads the list — **Spec 2 is what makes it user-visible**, as a priority contact silently failing to pierce mute at a remote site. Worth resolving before that work lands.

## Testing

- Unit tests: mock-based, `package service`, covering all three handlers' happy paths, validation, error paths, and both-fanouts-one-timestamp assertions. 84.3% package coverage; new functions ≥90%.
- Integration tests: testcontainers, including a concurrency test that seeds 28 contacts and fires 8 concurrent adds against a cap of 30 to prove the filter-based guard holds.
- All CI checks green, including `test-integration` for `user-service`, `inbox-worker`, and `pkg/model`.

https://claude.ai/code/session_0149LEpkiftMNyp8ZzKFFasz

## Summary by CodeRabbit

- **New Features**
  - Added priority-contact management to view, add, and remove contacts.
  - Supports up to 30 users or apps with enriched display details.
  - Added validation, duplicate handling, unresolved-contact support, and clear limit/not-found errors.
  - Settings responses and update events now include priority contacts.
  - Notifications can bypass mute settings for priority contacts, except during calls.

- **Bug Fixes**
  - Improved settings synchronization when updates share the same timestamp.

- **Documentation**
  - Documented operations, responses, limits, errors, and event behavior.


evelynchang5525:
Thanks — addressed in fd198e9. Two of the three findings were right; one I'm declining, with reasoning below.

**Accepted — configuration defaults (line 250).** Correct, and it mattered: "It defaults **true**" had `PRESENCE_RPC_ENABLED` as its nearest antecedent, so an implementer could plausibly have read it as enabling presence. Both flags are now named explicitly. Kept the "a gate nobody turns on" rationale rather than taking the suggested diff verbatim, since that is the actual reason for the asymmetry. While fixing it I found the same section still carried a stale single-compose path, which is also corrected.

**Accepted — bounded lookup (nitpick, line 66).** Fair. Added `USER_SETTINGS_TIMEOUT` (2s, mirroring the existing `PRESENCE_RPC_TIMEOUT`) so the read is bounded rather than inheriting whatever deadline the consumer context carries; on expiry it fails open, so slow Mongo degrades to current behaviour instead of stalling fan-out. Not adding bespoke metrics: the worker already wires `mongoutil.WithObservability`, so this attaches a dependency to an instrumented client and latency/error signals come with it.

**Accepted in part — the `active` observation (line 83).** The diagnosis is right and the prose was self-contradictory: it claimed a deactivated user "is not a recipient" and then described them pushing. Candidates come from `subscriptions`, not `users`, so a deactivated account with a live subscription stays in the slice, misses the settings map, and pushes. Rewritten to say so plainly. Not filtering them out here, though — this service consults `users` nowhere today, so deactivated users *already* receive pushes. That is a real pre-existing gap, but fixing it inside this change would silently alter delivery for a population the spec never set out to touch. Recorded as out of scope, belonging in its own change against the member-loading path.

**Declined — fail-closed on unknown settings (line 83).** Keeping fail-open. The spec now argues the case explicitly rather than leaving it implicit:

- A settings read failure is not per-user — it is a Mongo hiccup affecting every message in flight. Fail-closed turns that into total notification silence site-wide for the duration. Silence is the failure mode users cannot detect; nobody notices the push that never arrived, so it surfaces only when someone misses something important. Fail-open's worst case is a muted user receiving a push during an incident: visible, self-correcting on recovery, and far cheaper than the alternative.
- It matches the two neighbouring stages on this exact path. `PresenceSnapshotter` swallows errors and defaults to push; the hook vetoer logs and allows. A gate failing closed between two that fail open would make the pipeline's partial-failure behaviour incoherent.
- On the Security & Privacy framing specifically: muting is a delivery preference, not access control. An unwanted push to the recipient's own device discloses nothing to a third party — they are already a member of the room the message went to — so there is no confidentiality argument for treating unknown as deny.

Worth noting for context that this file is a design document, not an implementation — no code in this PR is affected by any of the above.

---

evelynchang5525:
Thanks — three of the four are addressed in `3b03017`; the fourth I'm declining. All four land on the Spec 2 planning docs, not on the shipping Spec 1 code.

**1. "one indexed `$in` per message" is wrong — fixed.** Correct: `Snapshot` chunks at `USER_SETTINGS_BATCH_SIZE` (512) and issues one `Find` per chunk, so `ceil(candidates/512)` queries. Reachable above one chunk when a large room's candidate set survives the `EligibleForPush` throttle — an `@all` in a several-thousand-member room. Corrected in the plan's architecture line and the compose comment, and in the two spec passages carrying the same loose claim.

**2. Config defaults test isn't isolated from the process environment — fixed, good catch.** `env.ParseAs` builds from `os.Environ()`, so an inherited `USER_SETTINGS_ENABLED` would shadow the `envDefault:"true"` the test exists to pin, and the test would pass or fail on host state. Now `t.Setenv` (to register the restore) then `os.Unsetenv` for both vars before parsing — the same shape `TestConfig_Mode` already uses for `MODE` at `notification-worker/config_test.go:31-34`.

**3. Verify the index before enabling — noted, but the index is already guaranteed.** `users.account` is a *unique* index created by `UserRepo.EnsureIndexes` (`user-service/mongorepo/users.go:33-36`), and that call fails service startup with an operator-facing message if it can't be created — including dedicated branches for a pre-existing non-unique `account_1` (codes 85/86) and for duplicate accounts. Since user-service is what writes `priorityContacts` in the first place, no deployment can have this feature without it. I've added a constraint recording that the index is owned by another service, that `notification-worker` must not create it, and that whoever ships should still confirm `db.users.getIndexes()` per environment before leaving the gate on. I did not add an explain-plan rollout gate — that's heavier than a guaranteed unique index warrants.

**4. Circuit breaker / bounded parallelism for timeout stalls — declining.** The concern is real but it isn't new, and acting on it here would exceed what this spec set out to change:

- `Presence.Snapshot` already sits on this exact path with the same serial 2s `PRESENCE_RPC_TIMEOUT` and the same fail-open contract. A settings breaker without a presence breaker would make the pipeline's behaviour under partial failure incoherent — which is the same reasoning the spec uses to justify fail-open over fail-closed.
- The stall is already bounded on two axes: `USER_SETTINGS_TIMEOUT` per call, and `MAX_WORKERS` (default 100) concurrent handler goroutines with `jsretry` backoff behind them. Worst case is degraded throughput during a Mongo outage, not unbounded growth.
- `USER_SETTINGS_ENABLED=false` is the operational answer to a sustained Mongo incident, and it exists precisely so ops can take the new dependency out of the path without a rollback.

Adding a breaker is a sensible follow-up for the fan-out path as a whole — settings *and* presence together — rather than something to bolt onto one of the two gates in the change that introduces it.

---

evelynchang5525:
Thanks — two fixed in `659056d`, two declined with reasoning below.

**Fixed: stale settings reads from a secondary.** Good catch, and the reasoning is stronger than stated. This service's `MONGO_READ_PREFERENCE=secondaryPreferred` arrived from `main` *after* this feature's design was written, so neither change considered the combination. The consequence is user-visible in the worst way: someone enables "mute all", user-service writes to the primary, and this worker keeps pushing until the write replicates. It also silently contradicts a claim the design spec makes explicitly — "no cache, **no TTL staleness**" — since secondary reads reintroduce staleness, just measured in replication lag rather than TTL.

Fixed narrowly: only the `users` collection is routed to `readpref.Primary()`, via the existing `mongoutil.CollectionWithReadPreference` helper. `subscriptions`, `thread_rooms`, and `rooms` keep `secondaryPreferred` — membership and metadata tolerate lag, and the point of that setting was to keep fan-out lookups off the primary. No new client, no new config knob.

**Fixed: incomplete env isolation in the defaults test.** Correct — the test cleared `USER_SETTINGS_ENABLED` and `PRESENCE_RPC_ENABLED` but still asserted on `UserSettingsBatchSize` and `UserSettingsTimeout` without clearing those, so an inherited value would have shadowed the very defaults it exists to pin. All four are now cleared.

**Declined for now: content-derived `Nats-Msg-Id`.** The hazard is real and I agree with the analysis — it's already documented in `docs/notification-worker-downstream-contracts.md` as an accepted risk, including the 250-survivor/batch-100 walkthrough where a redelivery after a fail-open survivor-set shift silently drops a recipient. It isn't being fixed in this PR for two reasons: `id` and `Nats-Msg-Id` are consumed by push-service and documented as a contract, so changing their derivation needs that team's agreement; and the failure requires both a partial batch failure *and* a settings-state change between delivery attempts. It's queued as a decision for the repo owner — either a tracking issue or a dedicated follow-up PR — rather than bundled into a change that's already carrying two specs.

**Declined: generated mocks and test renaming.** Both would make the new tests inconsistent with the file they live in. `notification-worker`'s test suite uses hand-written stubs throughout — `stubMembers`, `stubPresence`, `stubFollowers`, `stubParent`, `recordingEmitter` — and every existing handler test is named `TestHandle_*`. `stubSettings` and `accountVetoer` follow that established local pattern, and `accountVetoer` exists specifically because the pre-existing `rejectHook` is all-or-nothing and the placement test needs to veto exactly one account. Switching two of roughly seven doubles to generated mocks, or renaming four of ~40 tests to a different scheme, would leave the file half-converted. If the repo wants `go.uber.org/mock` and `Test<Type>_<Method>` here, that's a worthwhile sweep of the whole package in its own PR — I'd rather not start it halfway.

---

mliu33 (approved):
Outstanding work, thanks!

PR #217, opened 2026-08-10
Add loadgen verify subcommand for correctness testing

hmchangw:
## Summary

Adds a new `loadgen verify` subcommand that asserts **messaging correctness under concurrency** — complementing the existing `daily` scenario which measures capacity and latency. The verify scenario samples a fraction of sends into a fixed set of "probe rooms", tracks each probe's expected recipient set, and validates delivery completeness, leakage, persistence, and membership-change effectiveness.

## Key Changes

- **New verify subcommand entry point** (`verify.go`): Orchestrates the full run lifecycle (preflight → fixture build → user activation → warmup → steady-state probing → drain → evaluation). Configurable via flags for preset selection, probe room count, member churn rate, timing windows, and lane selection (global/local/both).

- **Probe tracking** (`verify_probe.go`, `verify_probe_test.go`): Records per-recipient delivery for sampled messages, detects violations (missing recipients, duplicates, total loss, leakage), and generates summary counts.

- **Membership model** (`verify_membership.go`, `verify_membership_test.go`): Maintains loadgen's oracle of probe-room membership (built from fixtures + applied changes), tracks epochs across membership changes, and validates that add/remove operations take effect on both the subscription list and authorization path.

- **Persistence readback** (`verify_readback.go`, `verify_readback_test.go`): Batches probes for history-service verification, confirms correct `messageId`, `roomId`, `senderID`, and `threadParentId` metadata, and detects persistence misses or field mismatches.

- **Violation types & reporting** (`verify_rooms.go`, `verify_rooms_test.go`, `verify_verdict.go`, `verify_verdict_test.go`, `verify_report.go`, `verify_report_test.go`): Defines 9 violation kinds (missing_recipient, total_loss, duplicate_delivery, unexpected_recipient, persistence_miss, persistence_mismatch, membership_not_applied, membership_add_ineffective, membership_remove_ineffective). Evaluates inputs against thresholds to produce PASS/FAIL/INCONCLUSIVE verdicts with actionable detail. Renders console and JSON output.

- **Direct pool enhancements** (`daily_pool.go`, `daily_pool_test.go`): Adds delivery sink attachment for probe tracking, lane selection (global/local), and drop accounting for recipient connection failures. Maintains backward compatibility with `daily` (nil sink = no-op).

- **Integration with daily scenario** (`daily.go`, `main.go`): Wires verify into the CLI, reuses fixture building and user activation infrastructure, and supports designated user ordering to ensure probe-room members occupy the direct pool.

- **Design documentation** (`docs/superpowers/specs/2026-08-08-loadgen-verify-correctness-design.md`): Comprehensive spec covering correctness definition, run lifecycle, probe sampling strategy, membership change validation, and per-user lane privacy guarantees.

- **Implementation plan** (`docs/superpowers/plans/2026-08-08-loadgen-verify-correctness.md`): Task decomposition and file structure.

- **README updates** (`tools/loadgen/README.md`): Documents the verify scenario, flags, and caveats (not yet validated end-to-end).

## Notable Implementation Details

- **Probe sampling**: Configurable fraction (default 1%) of room sends are tracked to bound accounting cost while keeping the system under full realistic load.
- **Membership epochs**: Probes published at epoch N are judged against epoch N's membership even if delivered after epoch N+1 begins, enabling late-delivery correctness.
- **Lane-aware delivery**: Rooms subscribed on both global and local lanes; duplicates are only violations within a single lane (per spec §7.1).
- **Settle windows**: After each membership change, a quiet period allows the system to propagate the change before probing resumes, with configurable duration.
- **Verdict thresholds**: INCONCLUSIVE if tracked probes fall below `--min-probes` (default 50); FAIL if any violation detected; PASS otherwise.
- **Backward compatible**: Direct pool changes preserve `daily` behavior (nil sink, both lanes

https://claude.ai/code/session_01JGF2qn4Zxwu3BQD8VfsYzM

## Summary by CodeRabbit

* **New Features**
  * Added a `verify` load-generator command for deterministic concurrent messaging correctness checks.
  * Checks delivery completeness, duplicate or unexpected recipients, persistence, and membership changes.
  * Supports configurable probing, room selection, membership churn, delivery lanes, retries, and console or JSON reports.
  * Reports `PASS`, `FAIL`, or `INCONCLUSIVE` results with exit codes and violation details.
  * Added a deployment target for running verification and saving timestamped results.

* **Documentation**
  * Documented configuration, output, limitations, and verification behavior.


julianshen:
**11 findings · 10 unique issues · 10 inline comments posted** — 🔴 6 blocking · 🟠 4 warning · 🔵 1 suggestion

### ⚠️ Rules that failed (4)

* `distributed-system` — the reviewer returned no parseable findings array (15131 chars)
* `elasticsearch` — the reviewer returned no parseable findings array (0 chars)
* `nats` — the reviewer returned no parseable findings array (8128 chars)
* `tgd-review` — the reviewer returned no parseable findings array (0 chars)

### Local review memory

- `docs/superpowers/specs/2026-08-08-loadgen-verify-correctness-design.md:123` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟠 Major_ _⚡ Quick win_ **Align --seed with the implemented fixture contract.** The specification says that --seed drives fixture generation. tools/loadgen/verify.go passes dailyRunSeed to B… ([thread](https://github.com/hmchangw/newchat/pull/217#discussion_r3745781010))
- `docs/superpowers/specs/2026-08-08-loadgen-verify-correctness-design.md:619` — @coderabbitai: _🎯 Functional Correctness_ _🟡 Minor_ _⚡ Quick win_ **Include the mandatory DM band in the output example.** Section 6.0 requires every probe-room set to include DM, small, and medium rooms. The example reports only 32 small, 18 medi… ([thread](https://github.com/hmchangw/newchat/pull/217#discussion_r3745781012))
- `tools/loadgen/daily_pool.go:321` — @coderabbitai: _🎯 Functional Correctness_ _🟡 Minor_ _⚡ Quick win_ **Unsubscribe the subscriptions this call created when it fails.** release() clears the du.rooms claim, but it does not remove the subscriptions already opened in this call: - If… ([thread](https://github.com/hmchangw/newchat/pull/217#discussion_r3745781014))
- `tools/loadgen/README.md:1134` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟡 Minor_ _⚡ Quick win_ **Add a language identifier to the fenced block.** markdownlint reports MD040 because this fence has no language. Use text or console . &lt;details> &lt;summary>Proposed fix</… ([thread](https://github.com/hmchangw/newchat/pull/217#discussion_r3754719641))

<details>
<summary>📒 Files reviewed (23)</summary>

* `docs/superpowers/plans/2026-08-08-loadgen-verify-correctness.md`
* `docs/superpowers/specs/2026-08-08-loadgen-verify-correctness-design.md`
* `tools/loadgen/README.md`
* `tools/loadgen/daily.go`
* `tools/loadgen/daily_pool.go`
* `tools/loadgen/daily_pool_test.go`
* `tools/loadgen/deploy/Makefile`
* `tools/loadgen/main.go`
* `tools/loadgen/verify.go`
* `tools/loadgen/verify_membership.go`
* `tools/loadgen/verify_membership_test.go`
* `tools/loadgen/verify_pool_test.go`
* `tools/loadgen/verify_probe.go`
* `tools/loadgen/verify_probe_test.go`
* `tools/loadgen/verify_readback.go`
* `tools/loadgen/verify_readback_test.go`
* `tools/loadgen/verify_report.go`
* `tools/loadgen/verify_report_test.go`
* `tools/loadgen/verify_rooms.go`
* `tools/loadgen/verify_rooms_test.go`
* `tools/loadgen/verify_test.go`
* `tools/loadgen/verify_verdict.go`
* `tools/loadgen/verify_verdict_test.go`

</details>
<details>
<summary>⚙️ Rules run (5)</summary>

* `cassandra`
* `mongodb`
* `oom`
* `silent-failure`

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+	}
+	body, err := json.Marshal(model.AddMembersRequest{RoomID: roomID, Users: []string{target.Account}})
+	if err != nil {
+		return pendingChange{}, false, fmt.Errorf("marshal member add: %w", err)
+	}
+	if err := r.requestAccepted(ctx, subject.MemberAdd(requester.Account, roomID, r.siteID), body); err != nil {
+		slog.Warn("member add rejected", "room", roomID, "target", targetID, "err", err)
+		return pendingChange{}, false, nil
+	}
+	// A real client subscribes on joining. A half-subscribed user silently
```
julianshen, line 868: _🏷️ failure-handling_ | _🔴 Blocking_ | _`distributed-system`_

**Ambiguous membership writes are silently discarded.**

Both addMember and removeMember convert any requestAccepted error into (no change, nil error). A concrete partial-failure path is: room-service commits an add/remove, its reply is lost, the NATS request times out, and loadgen logs the error but neither updates MembershipModel nor records an inconclusive signal. The change is then absent from ChangeCounts and all membership checks, so the run can report PASS while its model has diverged from the system. Transport errors must abort churn and feed HarnessErr/another INCONCLUSIVE signal; a definite application-level rejection of a valid requested change should also be recorded rather than treated as no work.

<details>

```
In `tools/loadgen/verify.go` around line 868: Both addMember and removeMember convert any requestAccepted error into (no change, nil error). A concrete partial-failure path is: room-service commits an add/remove, its reply is lost, the NATS request times out, and loadgen logs the error but neither updates MembershipModel nor records an inconclusive signal. The change is then absent from ChangeCounts and all membership checks, so the run can report PASS while its model has diverged from the system. Transport errors must abort churn and feed HarnessErr/another INCONCLUSIVE signal; a definite application-level rejection of a valid requested change should also be recorded rather than treated as no work.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+		if has {
+			observed = []string{pc.userID}
+		}
+		r.mm.RecordOracle(pc.roomID, observed, pc.epoch)
+	}
+
+	accepted, seen := r.sendAsTarget(ctx, pc)
+	if seen {
+		r.mm.RecordSendResult(pc.roomID, pc.userID, accepted, pc.epoch)
+	}
```
julianshen, line 967: _🏷️ failure-handling_ | _🔴 Blocking_ | _`distributed-system`_

**A lost authorization reply can still produce PASS.**

sendAsTarget returns seen=false for subscription, flush, publish, or reply-timeout failures, and observe simply skips RecordSendResult. harvest then permanently removes that pending change. For example, the gatekeeper accepts a removed member's stale send but its response is lost; Effective remains short, no violation or trust reason is created, and evaluateVerify can return PASS. Return an error from sendAsTarget and retry within a bounded budget or mark the run INCONCLUSIVE whenever the authorization outcome cannot be observed.

<details>

```
In `tools/loadgen/verify.go` around line 967: sendAsTarget returns seen=false for subscription, flush, publish, or reply-timeout failures, and observe simply skips RecordSendResult. harvest then permanently removes that pending change. For example, the gatekeeper accepts a removed member's stale send but its response is lost; Effective remains short, no violation or trust reason is created, and evaluateVerify can return PASS. Return an error from sendAsTarget and retry within a bounded budget or mark the run INCONCLUSIVE whenever the authorization outcome cannot be observed.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+		r.tracker.RecordSuppressed()
+		track = false
+	}
+
+	msgID, err := r.publishSend(ctx, u.Account, roomID)
+	if err != nil {
+		slog.Warn("probe send failed", "user", u.ID, "room", roomID, "err", err)
+		return
+	}
+	if !track {
```
julianshen, line 675: _🏷️ delivery-ordering_ | _🔴 Blocking_ | _`distributed-system`_

**Deliveries can arrive before their probe is registered.**

emitOneProbe calls publishSend before RegisterProbe, while RecordDelivery intentionally ignores unknown message IDs. After nats.Publish returns, the emitter goroutine can be descheduled while the broker and downstream workers process the message and direct-pool callbacks record its deliveries. Those callbacks discard the still-unknown ID; registration then occurs with an empty received set and finalization reports false loss. Generate the message ID and register the probe before crossing the publish boundary, then explicitly abandon that record if publication fails.

<details>

```
In `tools/loadgen/verify.go` around line 675: emitOneProbe calls publishSend before RegisterProbe, while RecordDelivery intentionally ignores unknown message IDs. After nats.Publish returns, the emitter goroutine can be descheduled while the broker and downstream workers process the message and direct-pool callbacks record its deliveries. Those callbacks discard the still-unknown ID; registration then occurs with an empty received set and finalization reports false loss. Generate the message ID and register the probe before crossing the publish boundary, then explicitly abandon that record if publication fails.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+// room's original membership, never from the reserve: a floater removed by
+// churn would get Forbidden on readback (its access window no longer covers
+// the message) and turn a healthy run INCONCLUSIVE (spec §8).
+func (r *verifyRun) pickSender(roomID string, rnd *rand.Rand) *userState {
+	members := r.prs.MembersOf(roomID)
+	if len(members) == 0 {
+		return nil
+	}
+	return r.byID[members[rnd.Intn(len(members))]]
+}
```
julianshen, line 653: _🏷️ consistency_ | _🔴 Blocking_ | _`distributed-system`_

**The settle check does not fence probes from membership changes.**

InSettle is checked before publishing, but churn only opens the model's settle window after the remote membership RPC and local subscription work complete. A probe can therefore publish while an add/remove is in flight. Because membership and message operations use independent subjects/workers, the backend may apply the remove before fan-out while the probe captures the old expected set, producing a false missing_recipient; the inverse add interleaving is also possible. Add a per-room mutation barrier that marks the room changing before issuing the membership RPC, prevents new probes, and abandons or drains already in-flight probes before changing the epoch; resume only after settle.

<details>

```
In `tools/loadgen/verify.go` around line 653: InSettle is checked before publishing, but churn only opens the model's settle window after the remote membership RPC and local subscription work complete. A probe can therefore publish while an add/remove is in flight. Because membership and message operations use independent subjects/workers, the backend may apply the remove before fan-out while the probe captures the old expected set, producing a false missing_recipient; the inverse add interleaving is also possible. Add a per-room mutation barrier that marks the room changing before issuing the membership RPC, prevents new probes, and abandons or drains already in-flight probes before changing the epoch; resume only after settle.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+	// BuiltinPreset returns a value; BuildFixtures takes a pointer.
+	preset, ok := BuiltinPreset(vc.Preset)
+	if !ok {
+		slog.Error("unknown preset", "preset", vc.Preset)
+		return 2
+	}
+	if vc.Users > 0 {
+		preset.Users = vc.Users
+	}
+
```
julianshen, line 160: _🎯 mongodb-correctness_ | _🔴 Blocking_ | _`mongodb`_

**--users regenerates fixtures instead of only limiting activation.**

The database is seeded from the full named preset, but this overwrites preset.Users before BuildFixtures. Because fixtures, room memberships, and subscription documents are derived from the preset, `verify --preset=daily-heavy --users=5000` builds an expected membership graph that can differ from the Mongo users/rooms/subscriptions seeded for daily-heavy. The run can then report missing recipients or membership failures against a healthy database. Keep the preset unchanged for seeded-state preflight and BuildFixtures; use vc.Users only when calculating how many already-built users executeVerify activates.

<details>

```
In `tools/loadgen/verify.go` around line 160: The database is seeded from the full named preset, but this overwrites preset.Users before BuildFixtures. Because fixtures, room memberships, and subscription documents are derived from the preset, `verify --preset=daily-heavy --users=5000` builds an expected membership graph that can differ from the Mongo users/rooms/subscriptions seeded for daily-heavy. The run can then report missing recipients or membership failures against a healthy database. Keep the preset unchanged for seeded-state preflight and BuildFixtures; use vc.Users only when calculating how many already-built users executeVerify activates.

```

</details>

tools/loadgen/verify_probe.go
```diff
@@ -0,0 +1,243 @@
...
+	t.probes[msgID] = rec
+	t.mu.Unlock()
+}
+
+// RecordDelivery notes that userID received msgID on the given lane. Deliveries
+// for untracked messages are ignored cheaply — 99% of traffic is untracked.
+func (t *ProbeTracker) RecordDelivery(userID, msgID, roomID string, ln lane, _ time.Time) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
```
julianshen, line 91: _🏷️ routing_ | _🔴 Blocking_ | _`nats`_

**Deliveries from the wrong room still satisfy a probe.**

`RecordDelivery` looks up only `msgID` and never compares its `roomID` argument with `probeRecord.roomID`. Thus a broadcast carrying a probe ID but routed or encoded for another room can count toward completeness; if the expected users overlap that room, the run can PASS despite cross-room delivery. For room-lane subscriptions, `directPool.deliver` also discards the actual NATS subject/subscribed room, so subject-versus-payload routing cannot be checked. Reject and record room-ID mismatches, and pass the subscribed room or parsed NATS subject into the sink so the intended room, subject room token, and event room ID can all be compared.

<details>

```
In `tools/loadgen/verify_probe.go` around line 91: `RecordDelivery` looks up only `msgID` and never compares its `roomID` argument with `probeRecord.roomID`. Thus a broadcast carrying a probe ID but routed or encoded for another room can count toward completeness; if the expected users overlap that room, the run can PASS despite cross-room delivery. For room-lane subscriptions, `directPool.deliver` also discards the actual NATS subject/subscribed room, so subject-versus-payload routing cannot be checked. Reject and record room-ID mismatches, and pass the subscribed room or parsed NATS subject into the sink so the intended room, subject room token, and event room ID can all be compared.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+	}
+	if err := r.env.publish(ctx, subject.MsgSend(pc.account, pc.roomID, r.siteID), data); err != nil {
+		slog.Warn("membership probe send failed", "room", pc.roomID, "err", err)
+		return false, false
+	}
+
+	waitCtx, cancel := context.WithTimeout(ctx, defaultRequestTimeout)
+	defer cancel()
+	msg, err := sub.NextMsgWithContext(waitCtx)
+	if err != nil {
```
julianshen, line 1062: _🏷️ error-handling_ | _🔴 Blocking_ | _`silent-failure`_

**Authorization-probe failures can be mistaken for expected rejection.**

sendAsTarget returns seen=false for subscription, flush, publish, and NextMsgWithContext failures, so observe records nothing and the verdict can still PASS with Effective lower than Total. It also treats every errcode reply as accepted=false; for a removal, an Internal or other unrelated service error therefore satisfies the expected rejection and is counted as effective. Return an error from sendAsTarget, feed it into an INCONCLUSIVE input, and classify accepted=false only for the specific authorization rejection expected after removal.

<details>

```
In `tools/loadgen/verify.go` around line 1062: sendAsTarget returns seen=false for subscription, flush, publish, and NextMsgWithContext failures, so observe records nothing and the verdict can still PASS with Effective lower than Total. It also treats every errcode reply as accepted=false; for a removal, an Internal or other unrelated service error therefore satisfies the expected rejection and is counted as effective. Return an error from sendAsTarget, feed it into an INCONCLUSIVE input, and classify accepted=false only for the specific authorization rejection expected after removal.

```

</details>

tools/loadgen/verify_probe.go
```diff
@@ -0,0 +1,243 @@
...
+	if _, expected := rec.expected[userID]; !expected {
+		// Leakage is only meaningful on the per-user lane, where the system
+		// chooses the address. On the room lane, delivery follows from whoever
+		// subscribed to the topic, which loadgen controls (spec §7.3).
+		if ln == laneUser {
+			rec.leaked[userID] = struct{}{}
+		}
+		return
+	}
+	rec.received[deliveryKey{userID: userID, ln: ln}]++
```
julianshen, line 105: _🎯 correctness_ | _🔴 Blocking_ | _`silent-failure`_

**Wrong-room events are credited as successful probe deliveries.**

RecordDelivery receives roomID but never compares it with the registered probe's room. A misrouted RoomEvent carrying a probe message ID can therefore satisfy expected recipients and allow PASS, particularly when users overlap both rooms. Reject mismatched room metadata before recording the recipient so the intended-room delivery remains missing.

<details>

> ‼️ **IMPORTANT**

```suggestion
	if roomID != rec.roomID {
		return
	}
	if _, expected := rec.expected[userID]; !expected {
		// Leakage is only meaningful on the per-user lane, where the system
		// chooses the address. On the room lane, delivery follows from whoever
		// subscribed to the topic, which loadgen controls (spec §7.3).
		if ln == laneUser {
			rec.leaked[userID] = struct{}{}
		}
		return
	}
	rec.received[deliveryKey{userID: userID, ln: ln}]++
```

</details>

<details>

```
In `tools/loadgen/verify_probe.go` around line 96: RecordDelivery receives roomID but never compares it with the registered probe's room. A misrouted RoomEvent carrying a probe message ID can therefore satisfy expected recipients and allow PASS, particularly when users overlap both rooms. Reject mismatched room metadata before recording the recipient so the intended-room delivery remains missing.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+		return 2
+	}
+	if vc.Users > 0 {
+		preset.Users = vc.Users
+	}
+
+	// Same preflight daily runs, for the same reason and a worse consequence:
+	// against an unseeded (or mis-seeded) database the gatekeeper rejects every
+	// send with "user X is not subscribed to room Y", yet the publishes still
+	// succeed — so every probe would reach zero recipients and the run would
```
julianshen, line 164: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**--users regenerates fixtures instead of only limiting activation.**

Overwriting `preset.Users` before `BuildFixtures` changes the generated fixture population and room memberships. The seeded database uses the builtin preset, while the CLI and README define `--users` as an activation limit. A partial run can therefore select probe rooms and expected recipients from fixtures that do not match Mongo, producing false delivery or membership failures. Build the full builtin fixture unchanged and apply `vc.Users` only when calculating how many users to activate in `executeVerify`.

<details>

```
In `tools/loadgen/verify.go` around line 164: Overwriting `preset.Users` before `BuildFixtures` changes the generated fixture population and room memberships. The seeded database uses the builtin preset, while the CLI and README define `--users` as an activation limit. A partial run can therefore select probe rooms and expected recipients from fixtures that do not match Mongo, producing false delivery or membership failures. Build the full builtin fixture unchanged and apply `vc.Users` only when calculating how many users to activate in `executeVerify`.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+}
+
+// observe runs both membership oracles for one settled change: what
+// subscription.list reports, and whether the target's own send is authorized.
+func (r *verifyRun) observe(ctx context.Context, pc *pendingChange) {
+	has, err := r.oracleHasRoom(ctx, pc.account, pc.roomID)
+	if err != nil {
+		r.recordOracleErr(pc.roomID, pc.userID, err)
+	} else {
+		var observed []string
```
julianshen, line 957: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**An unobserved authorization check can still produce PASS.**

`observe` records the membership send result only when `sendAsTarget` returns `seen=true`. Subscription failures, flush failures, publish failures, and reply timeouts all return `(false, false)`, leaving `sendSeen` false. `MembershipModel.Finalize` then emits no violation, and `evaluateVerify` does not treat `Effective < Total` as inconclusive, so a run can PASS without performing the add/remove effectiveness assertion. Propagate these observation failures into a verdict-gating error, or explicitly track unobserved authorization checks and make them INCONCLUSIVE.

<details>

```
In `tools/loadgen/verify.go` around line 957: `observe` records the membership send result only when `sendAsTarget` returns `seen=true`. Subscription failures, flush failures, publish failures, and reply timeouts all return `(false, false)`, leaving `sendSeen` false. `MembershipModel.Finalize` then emits no violation, and `evaluateVerify` does not treat `Effective < Total` as inconclusive, so a run can PASS without performing the add/remove effectiveness assertion. Propagate these observation failures into a verdict-gating error, or explicitly track unobserved authorization checks and make them INCONCLUSIVE.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+		return
+	}
+	if !track {
+		return
+	}
+	// Registered after the publish returns, not before: a failed publish would
+	// otherwise leave a registered probe nothing could ever deliver, i.e. a
+	// manufactured total_loss. The reverse race is not reachable — a delivery
+	// has to traverse MESSAGES → gatekeeper → canonical → broadcast-worker →
+	// back, which no map insert loses to.
```
julianshen, line 682: _🎯 concurrency_ | _🔴 Blocking_ | _`tgd-review`_

**The settle check races with membership changes.**

`emitOneProbe` checks `InSettle`, publishes, and reads the epoch in separate operations. Meanwhile add/remove requests mutate the real system before `ApplyAdd`/`ApplyRemove` opens the model's settle window. A selected probe can therefore be published while membership is changing and then be registered against either the old or new epoch, manufacturing `missing_recipient` failures for an explicitly ambiguous delivery. Mark the room as changing before issuing the external membership request and coordinate probe snapshot/publication with churn so no tracked publish can cross the change boundary.

<details>

```
In `tools/loadgen/verify.go` around line 682: `emitOneProbe` checks `InSettle`, publishes, and reads the epoch in separate operations. Meanwhile add/remove requests mutate the real system before `ApplyAdd`/`ApplyRemove` opens the model's settle window. A selected probe can therefore be published while membership is changing and then be registered against either the old or new epoch, manufacturing `missing_recipient` failures for an explicitly ambiguous delivery. Mark the room as changing before issuing the external membership request and coordinate probe snapshot/publication with churn so no tracked publish can cross the change boundary.

```

</details>

tools/loadgen/daily_pool.go
```diff
@@ -45,30 +159,38 @@ func newDirectPool(natsURL, credsFile string, c *Collector) *directPool {
...
 	}
-	du := &directUser{id: u.ID, nc: nc}
+	du.nc = nc
+	lanes := p.laneSelection()
 	for _, roomID := range u.Rooms {
-		for _, global := range []bool{true, false} {
+		for _, global := range lanes {
 			sub, err := nc.Subscribe(subject.RoomEvent(roomID, global), func(m *nats.Msg) {
 				p.onBroadcast(m)
+				p.deliver(u.ID, m.Data, laneForSubject(m.Subject))
```
julianshen, line 180: _🏷️ backpressure_ | _🟠 Warning_ | _`nats`_

**Async subscription drops are attributed to the system under test.**

The new verification callbacks perform JSON decoding and then serialize all deliveries through `ProbeTracker`'s single mutex, but these async subscriptions have no explicit pending limits, slow-consumer accounting, or `nats.ErrorHandler`. If a subscription's bounded nats.go pending queue overflows under load, nats.go drops deliveries and reports `ErrSlowConsumer`; this harness does not capture that signal, so `Finalize` reports the resulting gap as `missing_recipient` or `total_loss`. Install an async error handler or inspect each subscription's dropped count, set intentional pending message/byte limits, and make any harness-side subscription drop render the run INCONCLUSIVE.

<details>

```
In `tools/loadgen/daily_pool.go` around line 180: The new verification callbacks perform JSON decoding and then serialize all deliveries through `ProbeTracker`'s single mutex, but these async subscriptions have no explicit pending limits, slow-consumer accounting, or `nats.ErrorHandler`. If a subscription's bounded nats.go pending queue overflows under load, nats.go drops deliveries and reports `ErrSlowConsumer`; this harness does not capture that signal, so `Finalize` reports the resulting gap as `missing_recipient` or `total_loss`. Install an async error handler or inspect each subscription's dropped count, set intentional pending message/byte limits, and make any harness-side subscription drop render the run INCONCLUSIVE.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+		"probeMembers", len(prs.Members),
+		"reserve", len(reserve),
+		"probeRate", vc.ProbeRate,
+		"memberChurn", vc.MemberChurn)
+
+	rep, err := executeVerify(ctx, cfg, &vc, &fx, prs, designated)
+	if err != nil {
+		slog.Error("verify run failed", "err", err)
+		return 2
+	}
```
julianshen, line 199: _🏷️ error-handling_ | _🟠 Warning_ | _`silent-failure`_

**Console write failures do not affect the exit code.**

The return values from fmt.Print are ignored. If stdout encounters a partial write or ENOSPC while the verdict is PASS, the command can exit 0 without delivering the report. Check the write error and return the INCONCLUSIVE exit code.

<details>

> ‼️ **IMPORTANT**

```suggestion
	if _, err := fmt.Print(renderVerifyConsole(rep)); err != nil {
		slog.Error("write verify console report failed", "err", err)
		return 2
	}
```

</details>

<details>

```
In `tools/loadgen/verify.go` around line 199: The return values from fmt.Print are ignored. If stdout encounters a partial write or ENOSPC while the verdict is PASS, the command can exit 0 without delivering the report. Check the write error and return the INCONCLUSIVE exit code.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+		return "", fmt.Errorf("marshal verify send: %w", err)
+	}
+	if r.env.collector != nil {
+		r.env.collector.RecordPublish(reqID, msgID, time.Now())
+	}
+	if err := r.env.publish(ctx, subject.MsgSend(account, roomID, r.siteID), data); err != nil {
+		if r.env.collector != nil {
+			r.env.collector.RecordPublishFailed(reqID, msgID)
+		}
+		return "", fmt.Errorf("publish verify send: %w", err)
```
julianshen, line 706: _🎯 concurrency_ | _🟠 Warning_ | _`tgd-review`_

**Registering probes after publish can discard early deliveries.**

`publishSend` queues the message before `RegisterProbe` inserts its tracking record. Delivery callbacks run concurrently on other NATS connections, so there is no happens-before guarantee that the full service pipeline cannot return a broadcast before registration; the comment relies only on an expected timing difference. Such a delivery is ignored as an unknown message and later reported missing. Generate the message ID and register the probe before publishing, with an unregister/abort path if publication fails.

<details>

```
In `tools/loadgen/verify.go` around line 706: `publishSend` queues the message before `RegisterProbe` inserts its tracking record. Delivery callbacks run concurrently on other NATS connections, so there is no happens-before guarantee that the full service pipeline cannot return a broadcast before registration; the comment relies only on an expected timing difference. Such a delivery is ignored as an unknown message and later reported missing. Generate the message ID and register the probe before publishing, with an unregister/abort path if publication fails.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+// own fixture members, and tracks the deterministic --probe-rate fraction.
+// Exits on ctx.Done.
+func (r *verifyRun) emitProbes(ctx context.Context) {
+	rooms := r.prs.RoomIDs()
+	if len(rooms) == 0 {
+		return
+	}
+	interval := time.Duration(float64(time.Second) / probeSendRate(r.vc))
+	if interval <= 0 {
+		interval = time.Millisecond
```
julianshen, line 618: _🏷️ intent-match_ | _🟠 Warning_ | _`tgd-review`_

**Probes do not sample the daily workload described by the spec.**

`emitProbes` is a separate regular ticker that injects up to 200 additional simple root-message sends per second, and only those sends are eligible for tracking. The existing Markov/Poisson daily sends into probe rooms remain untracked. Consequently `--probe-rate` is not the documented fraction of workload sends, probe traffic has different timing and sender selection, and `ThreadParentID` is always empty, so threaded-message persistence metadata is never exercised. Integrate probe selection into the ordinary daily send path, or narrow the documented guarantees and remove the unexercised thread-parent claim.

<details>

```
In `tools/loadgen/verify.go` around line 618: `emitProbes` is a separate regular ticker that injects up to 200 additional simple root-message sends per second, and only those sends are eligible for tracking. The existing Markov/Poisson daily sends into probe rooms remain untracked. Consequently `--probe-rate` is not the documented fraction of workload sends, probe traffic has different timing and sender selection, and `ThreadParentID` is always empty, so threaded-message persistence metadata is never exercised. Integrate probe selection into the ordinary daily send path, or narrow the documented guarantees and remove the unexercised thread-parent claim.

```

</details>

julianshen (commented):
tGD inline review

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+	if err != nil {
+		if errors.Is(err, flag.ErrHelp) {
+			return 0 // -h / --help already printed usage
+		}
+		slog.Error("verify flag parse failed", "err", err)
+		return 2
+	}
+
+	// BuiltinPreset returns a value; BuildFixtures takes a pointer.
+	preset, ok := BuiltinPreset(vc.Preset)
```
julianshen, line 152: _🎯 mongodb-correctness_ | _🔴 Blocking_ | _`mongodb`_

**--users mutates preset.Users before BuildFixtures, desyncing fixtures from seeded Mongo**

runVerify does `if vc.Users > 0 { preset.Users = vc.Users }` and then `BuildFixtures(&preset, dailyRunSeed, cfg.SiteID)`. The database was seeded by `loadgen seed PRESET=<preset>` from the full builtin preset, so Mongo holds the full-population users/rooms/subscriptions. With `--preset=daily-heavy --users=5000`, the fixture graph (and therefore the probe-room member union, the expected recipient sets, and the reserve candidate pool) is derived from a shrunken user set that does not match the subscription documents in Mongo. The gatekeeper and broadcast-worker authorize/fan out according to Mongo, while ProbeTracker judges against fixture-derived membership: extra real members receive DM-lane probes (false `unexpected_recipient`) and fixture members absent from Mongo show as false `missing_recipient` — a confident FAIL on a healthy system. `verifyDailySeeded(ctx, cfg, dailyConfig{Preset: vc.Preset, Users: vc.Users})` receives the mutated user count, so the seeded-state preflight cannot catch the divergence either. The README documents `--users` as an activation count only. Minimal fix: leave the builtin preset untouched when building fixtures, and apply `vc.Users` solely as the activation cap in executeVerify (which already computes `activate` from it); audit whether verifyDailySeeded should receive the preset's own user count.

<details>

```
In `tools/loadgen/verify.go` around line 152: runVerify does `if vc.Users > 0 { preset.Users = vc.Users }` and then `BuildFixtures(&preset, dailyRunSeed, cfg.SiteID)`. The database was seeded by `loadgen seed PRESET=<preset>` from the full builtin preset, so Mongo holds the full-population users/rooms/subscriptions. With `--preset=daily-heavy --users=5000`, the fixture graph (and therefore the probe-room member union, the expected recipient sets, and the reserve candidate pool) is derived from a shrunken user set that does not match the subscription documents in Mongo. The gatekeeper and broadcast-worker authorize/fan out according to Mongo, while ProbeTracker judges against fixture-derived membership: extra real members receive DM-lane probes (false `unexpected_recipient`) and fixture members absent from Mongo show as false `missing_recipient` — a confident FAIL on a healthy system. `verifyDailySeeded(ctx, cfg, dailyConfig{Preset: vc.Preset, Users: vc.Users})` receives the mutated user count, so the seeded-state preflight cannot catch the divergence either. The README documents `--users` as an activation count only. Minimal fix: leave the builtin preset untouched when building fixtures, and apply `vc.Users` solely as the activation cap in executeVerify (which already computes `activate` from it); audit whether verifyDailySeeded should receive the preset's own user count.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+	if target == nil || requester == nil {
+		return pendingChange{}, false, nil
+	}
+	body, err := json.Marshal(model.AddMembersRequest{RoomID: roomID, Users: []string{target.Account}})
+	if err != nil {
+		return pendingChange{}, false, fmt.Errorf("marshal member add: %w", err)
+	}
+	if err := r.requestAccepted(ctx, subject.MemberAdd(requester.Account, roomID, r.siteID), body); err != nil {
+		slog.Warn("member add rejected", "room", roomID, "target", targetID, "err", err)
+		return pendingChange{}, false, nil
```
julianshen, line 866: _🏷️ consistency_ | _🔴 Blocking_ | _`mongodb`_

**addMember/removeMember swallow all requestAccepted errors as silent no-ops**

Both addMember and removeMember turn any error from `r.requestAccepted(...)` into `(pendingChange{}, false, nil)` with only a `slog.Warn`. Concrete execution path: room-service commits the add/remove to Mongo, the reply is lost, the NATS request times out — loadgen neither updates MembershipModel nor records any trust signal. The change never enters `mm.changes`, so ChangeCounts, RecordOracle, RecordSendResult, and Finalize never see it, and evaluateVerify can return PASS while loadgen's model has permanently diverged from the system's Mongo-backed membership state. Subsequent probes into that room are then judged against a stale expected set, manufacturing `missing_recipient` instead of an honest INCONCLUSIVE. Minimal fix: distinguish transport failures (timeout, no reply) — which must abort churn and feed HarnessErr or a dedicated INCONCLUSIVE input, mirroring how a SubscribeRoom failure already does — from definite application-level rejections, which should still be recorded (e.g. as an unapplied change) rather than dropped.

<details>

```
In `tools/loadgen/verify.go` around line 866: Both addMember and removeMember turn any error from `r.requestAccepted(...)` into `(pendingChange{}, false, nil)` with only a `slog.Warn`. Concrete execution path: room-service commits the add/remove to Mongo, the reply is lost, the NATS request times out — loadgen neither updates MembershipModel nor records any trust signal. The change never enters `mm.changes`, so ChangeCounts, RecordOracle, RecordSendResult, and Finalize never see it, and evaluateVerify can return PASS while loadgen's model has permanently diverged from the system's Mongo-backed membership state. Subsequent probes into that room are then judged against a stale expected set, manufacturing `missing_recipient` instead of an honest INCONCLUSIVE. Minimal fix: distinguish transport failures (timeout, no reply) — which must abort churn and feed HarnessErr or a dedicated INCONCLUSIVE input, mirroring how a SubscribeRoom failure already does — from definite application-level rejections, which should still be recorded (e.g. as an unapplied change) rather than dropped.

```

</details>

tools/loadgen/verify_probe.go
julianshen, line 105: _🎯 correctness_ | _🔴 Blocking_ | _`mongodb`, `silent-failure` (2 rules)_

**RecordDelivery ignores room mismatch, crediting cross-room events as delivered**

RecordDelivery looks up the probe by msgID only and never compares the roomID carried by the RoomEvent with probeRecord.roomID (registered at verify.go:684 from the fixture room). A broadcast that was misrouted or encoded for another room but carries a probe message ID therefore satisfies every expected recipient who overlaps the two rooms, and Finalize reports the probe complete — converting a cross-room routing defect into apparent exactly-once correctness, the exact failure class this tool exists to catch. Minimal fix: reject deliveries whose roomID does not match the registered probe's room, so the intended room's delivery remains missing and surfaces as missing_recipient/total_loss instead of being papered over.

<details>

> ‼️ **IMPORTANT**

```suggestion
if roomID != rec.roomID {
	return
}
if _, expected := rec.expected[userID]; !expected {
	// Leakage is only meaningful on the per-user lane, where the system
	// chooses the address. On the room lane, delivery follows from whoever
	// subscribed to the topic, which loadgen controls (spec §7.3).
	if ln == laneUser {
		rec.leaked[userID] = struct{}{}
	}
	return
}
rec.received[deliveryKey{userID: userID, ln: ln}]++
```

</details>

<details>

- **`mongodb`**:96: RecordDelivery looks up `t.probes[msgID]` and credits the delivery without checking that the `roomID` argument matches `rec.roomID`. A RoomEvent carrying a tracked message ID but encoded/routed for a different room counts toward completeness (and its sender toward `Complete`); if the two rooms share members, a cross-room delivery bug passes as a clean run. Additionally, `directPool.deliver` passes only the payload's RoomID — the subscribed room / NATS subject token are discarded — so subject-versus-payload routing cannot be cross-checked at all. Minimal guard: reject deliveries whose payload room differs from the registered probe's room (empty roomID tolerated for payloads lacking roomId). Passing the subscribed room or parsed subject into the sink for a fuller three-way check is a worthwhile follow-up.

</details>

<details>

```
In `tools/loadgen/verify_probe.go` around line 96: RecordDelivery looks up the probe by msgID only and never compares the roomID carried by the RoomEvent with probeRecord.roomID (registered at verify.go:684 from the fixture room). A broadcast that was misrouted or encoded for another room but carries a probe message ID therefore satisfies every expected recipient who overlaps the two rooms, and Finalize reports the probe complete — converting a cross-room routing defect into apparent exactly-once correctness, the exact failure class this tool exists to catch. Minimal fix: reject deliveries whose roomID does not match the registered probe's room, so the intended room's delivery remains missing and surfaces as missing_recipient/total_loss instead of being papered over.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+	if err != nil {
+		r.recordOracleErr(pc.roomID, pc.userID, err)
+	} else {
+		var observed []string
+		if has {
+			observed = []string{pc.userID}
+		}
+		r.mm.RecordOracle(pc.roomID, observed, pc.epoch)
+	}
+
```
julianshen, line 963: _🏷️ silent-failure_ | _🔴 Blocking_ | _`silent-failure`_

**Failed authorization observations are dropped, letting the run PASS unchecked**

In observe(), sendAsTarget returns seen=false for every observation failure: SubscribeSync error, Flush error, marshal error, publish error, and NextMsgWithContext timeout (verify.go:1030-1063, each path logs at most a slog.Warn). When seen is false, RecordSendResult is skipped, and harvest (verify.go:937-947) then permanently removes the pending change. The result: ChangeCounts.Total counts the change but Effective stays short, no violation is emitted, and evaluateVerify (tools/loadgen/verify_verdict.go) has no signal for Effective < Total — so the verdict is PASS. Failure path: the gatekeeper accepts a removed member's stale send (the severe membership_remove_ineffective case) but its reply is lost or times out once; the effectiveness assertion silently never happened and the run reports PASS. Minimal fix: have sendAsTarget return an error for subscribe/flush/publish/reply-timeout failures (as opposed to a definitive accept/reject answer), feed it into a verdict-gating input (e.g. recordHarnessErr or a dedicated unobserved-authorization counter surfaced like OracleErrs), and/or retry within a bounded budget before declaring the outcome unobservable.

<details>

```
In `tools/loadgen/verify.go` around line 963: In observe(), sendAsTarget returns seen=false for every observation failure: SubscribeSync error, Flush error, marshal error, publish error, and NextMsgWithContext timeout (verify.go:1030-1063, each path logs at most a slog.Warn). When seen is false, RecordSendResult is skipped, and harvest (verify.go:937-947) then permanently removes the pending change. The result: ChangeCounts.Total counts the change but Effective stays short, no violation is emitted, and evaluateVerify (tools/loadgen/verify_verdict.go) has no signal for Effective < Total — so the verdict is PASS. Failure path: the gatekeeper accepts a removed member's stale send (the severe membership_remove_ineffective case) but its reply is lost or times out once; the effectiveness assertion silently never happened and the run reports PASS. Minimal fix: have sendAsTarget return an error for subscribe/flush/publish/reply-timeout failures (as opposed to a definitive accept/reject answer), feed it into a verdict-gating input (e.g. recordHarnessErr or a dedicated unobserved-authorization counter surfaced like OracleErrs), and/or retry within a bounded budget before declaring the outcome unobservable.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+		slog.Warn("membership probe send failed", "room", pc.roomID, "err", err)
+		return false, false
+	}
+
+	waitCtx, cancel := context.WithTimeout(ctx, defaultRequestTimeout)
+	defer cancel()
+	msg, err := sub.NextMsgWithContext(waitCtx)
+	if err != nil {
+		return false, false
+	}
```
julianshen, line 1064: _🏷️ silent-failure_ | _🔴 Blocking_ | _`silent-failure`_

**Any errcode reply counts as rejection, masking unrelated service errors**

sendAsTarget classifies every errcode envelope as accepted=false (`if _, isErr := errcode.Parse(msg.Data); isErr { return false, true }`). For a removal check, the expected outcome IS rejection, so any error reply — an Internal from a Mongo hiccup in history/room-service, a BadRequest from a malformed request, etc. — is recorded via RecordSendResult as a successful authorization rejection and increments ChangeCounts.Effective. A gatekeeper with genuinely stale membership (still accepting the removed user) is thereby masked whenever some other component happens to error first, converting an infrastructure failure into apparent correctness. Minimal fix: parse the specific errcode and compare it to the authorization rejection the gatekeeper returns ('user X is not subscribed to room Y'); treat any other error code as an unobserved outcome that surfaces as an inconclusive/trust signal rather than as evidence of effective removal.

<details>

```
In `tools/loadgen/verify.go` around line 1064: sendAsTarget classifies every errcode envelope as accepted=false (`if _, isErr := errcode.Parse(msg.Data); isErr { return false, true }`). For a removal check, the expected outcome IS rejection, so any error reply — an Internal from a Mongo hiccup in history/room-service, a BadRequest from a malformed request, etc. — is recorded via RecordSendResult as a successful authorization rejection and increments ChangeCounts.Effective. A gatekeeper with genuinely stale membership (still accepting the removed user) is thereby masked whenever some other component happens to error first, converting an infrastructure failure into apparent correctness. Minimal fix: parse the specific errcode and compare it to the authorization rejection the gatekeeper returns ('user X is not subscribed to room Y'); treat any other error code as an unobserved outcome that surfaces as an inconclusive/trust signal rather than as evidence of effective removal.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+func (r *verifyRun) addMember(ctx context.Context, roomID, targetID string) (pendingChange, bool, error) {
+	target := r.byID[targetID]
+	requester := r.roomRequester(roomID)
+	if target == nil || requester == nil {
+		return pendingChange{}, false, nil
+	}
+	body, err := json.Marshal(model.AddMembersRequest{RoomID: roomID, Users: []string{target.Account}})
+	if err != nil {
+		return pendingChange{}, false, fmt.Errorf("marshal member add: %w", err)
+	}
```
julianshen, line 863: _🏷️ silent-failure_ | _🔴 Blocking_ | _`silent-failure`_

**Membership RPC errors are swallowed, so diverged changes escape all checks**

addMember (verify.go:863-866) and removeMember (verify.go:892-895) convert every requestAccepted error — including NATS timeouts — into (no change, nil error) with only a slog.Warn. Concrete partial-failure path: room-service commits the add, its reply is lost, the request times out; loadgen neither updates MembershipModel nor records any inconclusive signal. The change never enters m.changes, so it is absent from ChangeCounts, from Finalize's dual-oracle comparison, and from the authorization-effectiveness checks entirely — while the real system now has a member loadgen does not model. Subsequent probes judge against the stale expected set and the divergence is invisible on the room lane (non-member room-lane deliveries are deliberately ignored), so the run can report PASS with the model and the system disagreeing. Minimal fix: distinguish transport errors (requestFn returned a Go error, or an ambiguous errcode) from a definite application-level rejection; on transport failure abort churn via recordHarnessErr like the SubscribeRoom path already does, and record definite rejections of valid changes rather than treating them as no work.

<details>

```
In `tools/loadgen/verify.go` around line 863: addMember (verify.go:863-866) and removeMember (verify.go:892-895) convert every requestAccepted error — including NATS timeouts — into (no change, nil error) with only a slog.Warn. Concrete partial-failure path: room-service commits the add, its reply is lost, the request times out; loadgen neither updates MembershipModel nor records any inconclusive signal. The change never enters m.changes, so it is absent from ChangeCounts, from Finalize's dual-oracle comparison, and from the authorization-effectiveness checks entirely — while the real system now has a member loadgen does not model. Subsequent probes judge against the stale expected set and the divergence is invisible on the room lane (non-member room-lane deliveries are deliberately ignored), so the run can report PASS with the model and the system disagreeing. Minimal fix: distinguish transport errors (requestFn returned a Go error, or an ambiguous errcode) from a definite application-level rejection; on transport failure abort churn via recordHarnessErr like the SubscribeRoom path already does, and record definite rejections of valid changes rather than treating them as no work.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+	// harvest a settled one. Well below the 5s default --settle.
+	verifyChurnPoll = 500 * time.Millisecond
+	// verifyChurnObservation is the budget one settled change needs AFTER its
+	// settle window: harvest has to page subscription.list and run a send/reply
+	// round trip before the steady window closes. The effective tailroom is
+	// this PLUS --settle — see churnTailroom.
+	verifyChurnObservation = 10 * time.Second
+	// verifyChurnTailroom is the floor on that tailroom, so a zero --settle
+	// still leaves room for the two observations.
+	verifyChurnTailroom = 10 * time.Second
```
julianshen, line 240: _🎯 correctness_ | _🟠 Warning_ | _`cassandra`_

**Persistence readback retry budget (~12s) is shorter than realistic async Cassandra write lag.**

message-worker persists probes to Cassandra asynchronously from MESSAGES-CANONICAL; verify itself documents that "storage lag is not storage loss" (spec §8). Yet after the 30s drain, a still-missing probe gets only verifyReadbackAttempts=4 with verifyReadbackBackoff=4s — roughly 12–16 seconds of additional wall clock including query time. Concrete execution path on default flags: under daily-heavy background load, a message-worker consumer backlog or a Cassandra compaction pause that delays visibility of a batch beyond ~45s post-publish makes those probes surface as KindPersistenceMiss violations, and evaluateVerify turns that into a confident FAIL (exit 1) against a healthy stack — precisely the false-failure mode the retry mechanism exists to prevent. Because a miss maps to FAIL rather than INCONCLUSIVE, an under-sized budget directly corrupts the tool's primary output. Minimal fix: once the Task 11 resource/write-lag measurements exist, size attempts/backoff from observed p99 message-worker write latency (with headroom), and/or expose the budget as a flag; until measured, err generously, since the asymmetry (miss→FAIL, error→INCONCLUSIVE) punishes an under-budget far more than an over-budget.

<details>

```
In `tools/loadgen/verify.go` around line 240: message-worker persists probes to Cassandra asynchronously from MESSAGES-CANONICAL; verify itself documents that "storage lag is not storage loss" (spec §8). Yet after the 30s drain, a still-missing probe gets only verifyReadbackAttempts=4 with verifyReadbackBackoff=4s — roughly 12–16 seconds of additional wall clock including query time. Concrete execution path on default flags: under daily-heavy background load, a message-worker consumer backlog or a Cassandra compaction pause that delays visibility of a batch beyond ~45s post-publish makes those probes surface as KindPersistenceMiss violations, and evaluateVerify turns that into a confident FAIL (exit 1) against a healthy stack — precisely the false-failure mode the retry mechanism exists to prevent. Because a miss maps to FAIL rather than INCONCLUSIVE, an under-sized budget directly corrupts the tool's primary output. Minimal fix: once the Task 11 resource/write-lag measurements exist, size attempts/backoff from observed p99 message-worker write latency (with headroom), and/or expose the budget as a flag; until measured, err generously, since the asymmetry (miss→FAIL, error→INCONCLUSIVE) punishes an under-budget far more than an over-budget.

```

</details>

tools/loadgen/verify.go
```diff
@@ -0,0 +1,1107 @@
...
+// message ID it minted.
+func (r *verifyRun) publishSend(ctx context.Context, account, roomID string) (string, error) {
+	msgID := idgen.GenerateMessageID()
+	reqID := idgen.GenerateRequestID()
+	data, err := json.Marshal(model.SendMessageRequest{ID: msgID, Content: verifyProbeContent, RequestID: reqID})
+	if err != nil {
+		return "", fmt.Errorf("marshal verify send: %w", err)
+	}
+	if r.env.collector != nil {
+		r.env.collector.RecordPublish(reqID, msgID, time.Now())
```
julianshen, line 700: _🎯 concurrency_ | _🟠 Warning_ | _`mongodb`_

**Registering the probe only after publishSend can discard early deliveries**

publishSend queues the message on NATS before RegisterProbe inserts the tracking record, and RecordDelivery deliberately ignores unknown message IDs. There is no happens-before guarantee that the full pipeline (gatekeeper → canonical → broadcast-worker → subscriber callback) cannot deliver before the emitter goroutine's map insert lands; such deliveries are silently dropped and the probe finalizes as a false `total_loss` (FAIL). The in-code comment argues the pipeline is far slower than the insert, which is a timing heuristic, not a guarantee. Minimal fix: mint the message ID and RegisterProbe before crossing the publish boundary, and add a tracker removal/abandon path so a failed publish does not leave a guaranteed-total-loss record.

<details>

```
In `tools/loadgen/verify.go` around line 700: publishSend queues the message on NATS before RegisterProbe inserts the tracking record, and RecordDelivery deliberately ignores unknown message IDs. There is no happens-before guarantee that the full pipeline (gatekeeper → canonical → broadcast-worker → subscriber callback) cannot deliver before the emitter goroutine's map insert lands; such deliveries are silently dropped and the probe finalizes as a false `total_loss` (FAIL). The in-code comment argues the pipeline is far slower than the insert, which is a timing heuristic, not a guarantee. Minimal fix: mint the message ID and RegisterProbe before crossing the publish boundary, and add a tracker removal/abandon path so a failed publish does not leave a guaranteed-total-loss record.

```

</details>

tools/loadgen/daily_pool.go
```diff
@@ -106,8 +231,100 @@ func (p *directPool) onBroadcast(m *nats.Msg) {
...
+		p.mu.Unlock()
+	}
+	for _, global := range p.laneSelection() {
+		subj := subject.RoomEvent(roomID, global)
+		sub, err := du.nc.Subscribe(subj, func(m *nats.Msg) {
+			p.deliver(userID, m.Data, laneForSubject(m.Subject))
+		})
+		if err != nil {
+			release()
+			return fmt.Errorf("subscribe room %s for %s: %w", roomID, userID, err)
```
julianshen, line 310: _🧹 cleanup_ | _🟠 Warning_ | _`silent-failure`_

**SubscribeRoom failure leaves live subscriptions behind while clearing the claim**

release() only deletes the du.rooms claim; it does not unsubscribe anything this call created. If the second lane's Subscribe fails (daily_pool.go:320-322) or Flush fails (:328-330), the first lane's subscription stays live and appended to du.subs while pool state records that the user does not hold the room. The connection keeps delivering room events through p.deliver, attributing them to userID for a room the membership model treats as not joined. Today the blast radius is limited because addMember treats any SubscribeRoom error as fatal and aborts churn (harnessErr → INCONCLUSIVE), but the rollback still leaves the pool, the live subscription set, and the model inconsistent until Close. Minimal fix: collect the subscriptions created in this call and Unsubscribe them inside release(), removing their entries from du.subs (or defer appending to du.subs until after Flush succeeds).

<details>

```
In `tools/loadgen/daily_pool.go` around line 310: release() only deletes the du.rooms claim; it does not unsubscribe anything this call created. If the second lane's Subscribe fails (daily_pool.go:320-322) or Flush fails (:328-330), the first lane's subscription stays live and appended to du.subs while pool state records that the user does not hold the room. The connection keeps delivering room events through p.deliver, attributing them to userID for a room the membership model treats as not joined. Today the blast radius is limited because addMember treats any SubscribeRoom error as fatal and aborts churn (harnessErr → INCONCLUSIVE), but the rollback still leaves the pool, the live subscription set, and the model inconsistent until Close. Minimal fix: collect the subscriptions created in this call and Unsubscribe them inside release(), removing their entries from du.subs (or defer appending to du.subs until after Flush succeeds).

```

</details>

tools/loadgen/verify_readback.go
```diff
@@ -0,0 +1,244 @@
...
+// again.
+func (r *Readback) verifyRoom(ctx context.Context, account, roomID string, targets []ReadbackTarget) ([]Violation, error) {
+	pending := make(map[string]ReadbackTarget, len(targets))
+	for _, t := range targets {
+		pending[t.MsgID] = t
+	}
+
+	var mismatches []Violation
+
+	for attempt := 0; attempt < r.attempts && len(pending) > 0; attempt++ {
```
julianshen, line 112: _🏷️ operational-risk_ | _🔵 Suggestion_ | _`cassandra`_

**Sequential per-account, per-room readback retries can stall finalize for many minutes.**

run.readback (tools/loadgen/verify.go) iterates sender accounts sequentially; Readback.Verify iterates rooms sequentially; verifyRoom sleeps the full backoff between attempts for each room independently. When persistence is broadly unavailable (history-service down, or message-worker stalled so every probe stays pending), the finalize cost is sum(accounts) x rooms x (attempts-1) x backoff with no overall deadline other than ctx cancellation: at the default 50 probe rooms with all probes missing, that is ~50 x 3 x 4s = ~10 minutes of pure sleep appended after the drain phase before the operator sees any verdict. This is a reachable operational path (a dependency outage mid-run), not a scale speculation. Minimal fix: bound the whole readback phase with a deadline derived from ctx (e.g., context.WithTimeout covering run.readback) and/or fan out per-room verification concurrently so one slow room does not serialize the budget of every other room.

<details>

```
In `tools/loadgen/verify_readback.go` around line 112: run.readback (tools/loadgen/verify.go) iterates sender accounts sequentially; Readback.Verify iterates rooms sequentially; verifyRoom sleeps the full backoff between attempts for each room independently. When persistence is broadly unavailable (history-service down, or message-worker stalled so every probe stays pending), the finalize cost is sum(accounts) x rooms x (attempts-1) x backoff with no overall deadline other than ctx cancellation: at the default 50 probe rooms with all probes missing, that is ~50 x 3 x 4s = ~10 minutes of pure sleep appended after the drain phase before the operator sees any verdict. This is a reachable operational path (a dependency outage mid-run), not a scale speculation. Minimal fix: bound the whole readback phase with a deadline derived from ctx (e.g., context.WithTimeout covering run.readback) and/or fan out per-room verification concurrently so one slow room does not serialize the budget of every other room.

```

</details>

PR #218, opened 2026-08-10
feat(threadcount): drop the 99 cap 

amdla:
Count/CountAndLatest scanned only until the live tally hit Cap=99, so any longer thread reported exactly 99 and clients rendered "99+". The count is now exact: the scan pages to the end of the partition.

Trade-off, stated plainly: Cap also bounded the per-write partition read. Every thread reply add and delete recomputes tcount, so that read is now linear in the thread's length instead of ~99 rows, making writes progressively more expensive as a thread grows. PageSize still bounds per-round-trip memory, not the total scanned. If this shows up on the hot path, an incremental counter is the fix, not a re-introduced ceiling.

Updates both authoritative writers' comments (message-worker add path, history-service delete path), the client-API docs for tcount and newTcount, the derived events view, and the Cassandra schema doc.

## Summary by CodeRabbit

* **New Features**
  * Thread reply counts now show the exact number of non-deleted replies, including threads with more than 99 replies.
  * Reply count updates and thread metadata now provide authoritative exact totals.

* **Bug Fixes**
  * Soft-deleted replies are excluded consistently while active replies are fully counted.
  * Long-thread counting now reports complete totals and avoids returning partial results when canceled.

* **Documentation**
  * Updated API and data model documentation to reflect exact reply-count behavior.

### DIFF
--- a/docs/cassandra_message_model.md
+++ b/docs/cassandra_message_model.md
@@ -143,7 +143,7 @@ CREATE TABLE IF NOT EXISTS messages_by_room(
   sender FROZEN<"Participant">,
   site_id TEXT,
   sys_msg_data BLOB,
-  tcount INT, // bounded non-deleted thread reply count, capped at 99 (pkg/threadcount.Cap); FE renders >= 99 as "99+"
+  tcount INT, // exact non-deleted thread reply count (pkg/threadcount)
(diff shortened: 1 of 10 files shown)


GITMateuszCharczuk (approved):
LGTM!

Issue #203, opened 2026-08-07
translation test


YI13:
2026/08/07 14:41:41 CLOSED-LOOP subject=chat.user.YCJUANJ.request.translate.00600000.text workers=1 conns=1 duration=5s lang=en
---------- results ----------
total=9  ok=9  busy(Unavailable)=0  other-err=0  timeout/no-responder=0
throughput=2 req/s (replies received: 2/s)
reply latency  p50=615.253425ms  p90=749.025038ms  p95=749.025038ms  p99=749.025038ms  max=749.025038ms

YI13:
>> building loadgen ...
>> Step 0: baseline latency (qps=1, 20s) ...
2026/08/07 15:18:39 OPEN-LOOP subject=chat.user.YCJUANJ.request.translate.00600000.text qps=1 max-inflight=2 conns=1 duration=20s lang=en
>> baseline: p50=380.3ms p95=1583.8ms ok/s=1
>> sizing for 50 RPS: expected in-flight ~80 -> -max-inflight=120
>> ACTION: ensure the service's MAX_CONCURRENCY >= 120 (else you'll see 'busy' before 50 RPS).

qps   | ok/s     | busy   | err5xx   | dropped  | p50ms    | p95ms    | p99ms   
------+----------+--------+----------+----------+----------+----------+---------
>> step qps=10 for 60s ...
2026/08/07 15:18:59 OPEN-LOOP subject=chat.user.YCJUANJ.request.translate.00600000.text qps=10 max-inflight=120 conns=1 duration=1m0s lang=en
10    | 0.4      | 0      | 579      | 0        | 9.8      | 81.1     | 457
>> step qps=25 for 60s ...
2026/08/07 15:19:59 OPEN-LOOP subject=chat.user.YCJUANJ.request.translate.00600000.text qps=25 max-inflight=120 conns=1 duration=1m0s lang=en
25    | 0.3      | 0      | 1480     | 0        | 8.9      | 130.9    | 414.1
>> step qps=40 for 60s ...
2026/08/07 15:20:59 OPEN-LOOP subject=chat.user.YCJUANJ.request.translate.00600000.text qps=40 max-inflight=120 conns=1 duration=1m0s lang=en
40    | 0.3      | 0      | 2380     | 0        | 46.5     | 267.7    | 468
>> step qps=45 for 60s ...
[...]

PR #219, opened 2026-08-10
feat(translation): map backend 429 to too_many_requests

YI13:
## Summary

When the third-party translate API returns **HTTP 429 (rate limited)**, `translation-service` was falling through to a raw `translate backend error (status 429)` which collapses to `internal` at the NATS boundary. The client received:

```json
{"code":"internal","error":"internal error"}
```

That's the least useful signal for the **most common** failure — the backend is tightly rate-limited (measured ≈ 10 RPM in the test env; see #203), so 429 is the dominant error, yet it looked like a generic server bug with no "back off / retry later" hint.

`translateOnce` already maps **5XX / transport failures → `unavailable`** (added recently). This PR gives **429 its own classification** — consistent with that, but a distinct, accurate code.

## Change

`translateOnce` classifies the status **before parsing the body**: a 429 now returns `errcode.TooManyRequests` with reason `rate_limited`, and the service does **NOT retry** (only a `failed to verify jwt` response triggers the existing refresh-and-retry).

Client now receives:

```json
{"code":"too_many_requests","reason":"rate_limited","error":"translation rate limited"}
```

(HTTP category 429.)

| Backend response | Before | After |
|---|---|---|
| 429 | `internal` | **`too_many_requests` / `rate_limited`** |
| 5XX / transport / mid-stream drop | `unavailable` / `upstream_unavailable` | unchanged |
| other 4XX / malformed stream / non-success returnCode | `internal` | unchanged |

**No retry on 429** — retrying a rate-limited backend just amplifies load; the caller is told to back off instead.

## Files

- `pkg/errcode/codes_translation.go` — new reason `TranslateRateLimited` (`"rate_limited"`, 429).
- `translation-service/translator_stream.go` — 429 → `TooManyRequests(rate_limited)` (added `net/http`).
- `translation-service/translator_stream_test.go` — new test: 429 → `too_many_requests` + `rate_limited`, translate called **once** (no retry).
- `docs/client-api.md` + `docs/client-api/request-reply.md` — document the new error case (client-facing handler error surface changed).

## Testing

- `make test SERVICE=translation-service` green (`-race`); `pkg/errcode` tests green.
- `make lint` 0 issues; `make sast-gosec` PASS; `go build ./...` OK.

## Not in scope

- `Retry-After` passthrough and a translation-result cache are separate follow-ups (the higher-leverage mitigations for a 10 RPM backend).

---
_Generated by [Claude Code](https://claude.ai/code/session_01H2Jd3N1mNhSxjKNLcYgcgS)_

## Summary by CodeRabbit

* **New Features**
  * Added explicit handling for translation service rate limiting.
  * HTTP 429 responses now return a `too_many_requests` error with the `rate_limited` reason and are not retried.

* **Documentation**
  * Updated the translation error reference to document rate-limiting behavior.
  * Clarified that internal errors include other 4XX backend responses.

* **Tests**
  * Added coverage verifying rate-limited translation requests return the expected error without retrying.

### DIFF
--- a/docs/client-api.md
+++ b/docs/client-api.md
@@ -5939,7 +5939,8 @@ See [Error envelope](#6-error-envelope-reference). The reply carries the `{ code
 | `bad_request` | `unsupported_lang` | `targetLang` does not resolve to a supported language (outside the [Supported languages](#supported-languages) set, or a bare `zh` with no script/region). |
 | `unavailable` | — | Handler saturation — the concurrency cap is full; retry. |
 | `unavailable` | `upstream_unavailable` | The third-party translation backend returned a 5XX or was unreachable (transport failure). Clients should show a "translation service temporarily unavailable" message and allow a retry. |
-| `internal` | — | Other translation backend failure (e.g. malformed stream, 4XX, non-success returnCode). The raw cause is logged server-side, never returned. |
+| `too_many_requests` | `rate_limited` | The third-party translation backend rate-limited the request (HTTP 429). The service does **not** retry; clients should back off and retry later. |
+| `internal` | — | Other translation backend failure (e.g. malformed stream, other 4XX, non-success returnCode). The raw cause is logged server-side, never returned. |
 
 ```json
 {
--- a/docs/client-api/request-reply.md
+++ b/docs/client-api/request-reply.md
@@ -2128,7 +2128,7 @@ with a `TranslateResult`, or the standard error envelope on failure.
 
 #### Error response
 
-Standard `{ code, reason?, error }` envelope. Key errors: `empty_text` (`bad_request`) for empty `text`; `unsupported_lang` (`bad_request`) for a `targetLang` outside the set; `unavailable` under handler saturation; `unavailable` / `upstream_unavailable` when the third-party backend returns a 5XX or is unreachable; `internal` for other backend failures. See [../client-api.md §3.6](../client-api.md#36-translation-service).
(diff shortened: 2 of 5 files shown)


YI13:
Re: the nitpick on `translation-service/translator_stream_test.go:108` (discarded `io.WriteString` error) — not adopting, for three reasons:

1. **File convention.** Every httptest handler in this file writes its response with `_, _ = io.WriteString(...)` (~15 sites: lines 36, 74, 75, 155, 159, 160, 180, 210, 218, 236, 253, 287, 315, 316, …). Changing only line 108 would make the new test inconsistent with the other 14 that sit right beside it.

2. **Not a meaningful error path.** This is an in-process `httptest` server; a write failure there just means the client hung up, which the test already asserts against via the returned error. The `_, _ =` idiom is the deliberate, standard convention for these in-test writers.

3. **The proposed diff is malformed.** It drops the closing `}` of the JSON literal (`` `{"message":"rate limit exceeded"` ``), so applying it verbatim would change the payload.

If we want to stop discarding these write errors, that's a whole-file cleanup (all ~15 sites) and belongs in its own PR, not scoped to this 429 change.

---

chenjr0719 (approved):
LGTM — mirrors the existing 5XX→unavailable mapping cleanly: the 429 is classified before parsing the body, returns `too_many_requests`/`rate_limited`, and (per the added test) is not retried, unlike a jwt failure. Docs are ratcheted (client-api + request-reply).

One optional, non-blocking future consideration: a backend 429 can carry a `Retry-After` — propagating it (like `pkg/msgraph` does for Graph throttling) would let clients back off precisely rather than blindly. The translation design doesn't establish a `Retry-After` contract for this backend, so it's not required here — worth revisiting if the upstream's error contract turns out to include one.

PR #220, opened 2026-08-10
fix(search): empty search results — index env-var drift, Teams INBOX envelope, system-message indexing

amdla:
`search.rooms` returned an empty array on the test server.

Root cause: search-service read the ES index names under a `SEARCH_` prefix while search-sync-worker and es-index-migrator read them unprefixed, so the reader was searching an index nobody writes. That fails silently — the read pattern is a wildcard queried with `allow_no_indices=true`, so a wrong index name returns `200` with zero hits, indistinguishable from "no results". Chasing it surfaced two more defects in the same indexing path, fixed here.

## search-service

- The three index env vars move off `SearchConfig` (which injects `envPrefix:"SEARCH_"`) onto root `Config`, so reader and writer agree on one spelling: `USER_ROOM_INDEX`, `SPOTLIGHT_INDEX`, `SPOTLIGHT_ORG_INDEX`.
- All three are now `required,notEmpty`. `USER_ROOM_INDEX` had neither, and being intentionally unversioned it also skipped the `StripVersion` shape check — an empty value parsed cleanly at startup and then failed per-request inside the terms-lookup.
- Empty room/org results log the pattern that was actually searched: WARN when it resolved to zero shards (always a misconfiguration, never a routine miss), flow level otherwise. This is the line that would have made the original incident a one-minute diagnosis instead of a code audit.

## room-worker

`federateTeamsMembership` published a bare `InboxMemberEvent` on the INBOX internal lane, where the consumer expects the `InboxEvent` envelope. The decode failed and search-sync-worker Acked the event away, so Teams-migrated memberships never reached the spotlight or user-room indexes — degrading room search *and* message search (the user-room index is the restricted-rooms map message search depends on).

The local lane now wraps like every other internal-lane publisher. The federated lane deliberately keeps passing the inner event, because `outbox.Publish` builds its own envelope — a test pins that asymmetry so a future "make both lanes consistent" edit can't double-wrap and break federation.

## search-sync-worker

- `parseMemberEvent` rejects unwrapped and typeless events by name. The old timestamp-only guard could not catch them: `timestamp` is a field on *both* structs, so a bare inner event cleared it and then died deep in the payload decode on an opaque `unexpected end of JSON input`.
- System messages (membership/rename chrome) are no longer indexed as searchable content. message-worker, notification-worker and history-service already gate on `IsSystemMessageType`; this collection was the outlier. One filter covers all three message consumers — they share `*messageCollection`.
- Poison-drop logs now carry the subject and consumer. Those errors are Acked, so the log line is the only trace the message ever existed.

## es-index-migrator

Same system-message filter on backfill, including selecting the `type` column the projection had excluded — without it the filter would have been a silent no-op and a rebuild would have re-introduced every system message.

## pkg/model

`systemMessageTypes` is hand-maintained and four services read it, so an unregistered constant changes all four behaviours at once with nothing failing. A guard now walks every non-test file in the package and fails when a `MessageType*` constant is missing from the map or from an explicit exemption list with a reason.

## Verification

- **CI green** on all 9 jobs, including `sast` (gosec + govulncheck + semgrep) and every `test-integration` target.
- **End-to-end on a local stack** (real NATS + Elasticsearch + Mongo + Cassandra): a Teams-migration batch was published on `chat.teams.room.canonical.{site}.create` and the resulting room was confirmed present in the spotlight index, in the user-room index, and in the `search.rooms` reply — with no `build action` error in the worker. This is the path no automated test reaches: `TestSpotlightSync_Integration` is genuinely end-to-end but builds its own envelope, so it would pass even if room-worker published garbage. The producer→consumer seam is otherwise pinned only by mirrored unit tests in two packages that cannot import each other.
- Also verified end-to-end: reader/writer index-name agreement, the WARN firing only for a genuinely missing index (and not on a routine miss), and the backfill filter being live rather than a no-op.

## Docs

`docs/client-api.md` and its request/reply view state that system messages are never returned by `search.messages` — the one client-observable behaviour change here. The seven excluded types match `systemMessageTypes` exactly; a client-set `type: "important"` is normal content and stays searchable.

`docs/client-api/events.md` is untouched: no server→client event struct changed. No request/response schema or error case changed either, so no field tables were touched — the only file altered under `pkg/model/` is a test.

## Note for reviewers

`Search Rooms` documents `roomType: "dm"` as a supported filter. It is accepted and the filter itself works, but DM rooms are currently indexed with an empty `roomName` (`resolveRoomName` returns `""` for non-channels), so they cannot match a name query on the live path — the backfill does not share the defect, which is why behaviour differs depending on how the index was populated. **Pre-existing, not introduced here**, and tracked as follow-up work along with four other defects in this subsystem (same-site DMs emitting no INBOX event, `room_renamed` never updating spotlight, integration tests building the index without the production analyzer, and spotlight having no mapping push).


room-worker/teamsroomcreate.go
```diff
@@ -261,8 +261,21 @@ func (h *Handler) federateTeamsMembership(ctx context.Context, room *model.Room,
...
+	// The internal lane carries the InboxEvent envelope like every other
+	// InboxInternal publisher — search-sync-worker decodes evt.Payload. The
+	// federated branch below stays unwrapped: outbox.Publish builds its own.
+	internalData, err := json.Marshal(model.InboxEvent{
+		Type:       eventType,
+		SiteID:     h.siteID,
+		DestSiteID: h.siteID,
+		Payload:    payload,
+		Timestamp:  acceptedAt.UnixMilli(),
+	})
```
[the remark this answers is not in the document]
amdla: search-sync-worker uses InboxEvent.Timestamp as the ES external document version, and this is a batch replay; the federated lane passes acceptedAt for the same reason

search-sync-worker/inbox_stream_test.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/hmchangw/chat/pkg/model"
+)
+
+// Covers the INBOX envelope contract. The unwrapped case matters most: a bare
+// InboxMemberEvent shares `timestamp`/`siteId` with InboxEvent, so it clears a
+// timestamp-only guard and then fails opaquely in the payload decode.
+func TestParseMemberEvent(t *testing.T) {
```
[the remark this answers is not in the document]
amdla: parseMemberEvent is a package-level function, not a method, and there's no MemberEvent type here

julianshen:
**Actionable comments posted: 11** — 🔴 2 blocking · 🟠 9 warning

<details>

* `data-migration/es-index-migrator/cassandra_test_helpers_test.go`
* `data-migration/es-index-migrator/messagesource_cassandra.go`
* `data-migration/es-index-migrator/runner.go`
* `data-migration/es-index-migrator/runner_test.go`
* `docs/client-api.md`
* `pkg/model/event_test.go`
* `room-worker/teamsroomcreate.go`
* `room-worker/teamsroomcreate_test.go`
* `search-service/deploy/docker-compose.yml`
* `search-service/handler.go`
* `search-service/handler_test.go`
* `search-service/main.go`
* `search-service/query_messages.go`
* `search-service/response.go`
* `search-service/response_test.go`
* `search-sync-worker/inbox_stream.go`
* `search-sync-worker/inbox_stream_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `silent-failure`
* `tgd-review`

</details>

search-sync-worker/messages.go
```diff
@@ -190,6 +190,12 @@ func (c *messageCollection) BuildAction(data []byte) ([]searchengine.BulkAction,
 	if !actionableEvent(evt.Event) {
 		return nil, nil
 	}
+	// Sys-messages (membership/rename) arrive on MESSAGES-CANONICAL like any
+	// other message but are UI chrome, not searchable content. Filtered before
+	// the parent lookup; `important` is client-set, not system, and stays.
+	if model.IsSystemMessageType(evt.Message.Type) {
```
julianshen, line 196: _🏷️ rolling-deploy_ | _🔴 Blocking_ | _`distributed-system`_

**Filtering does not remove messages indexed by an old worker.**

During a rolling deployment, an old worker can index system message m1 and acknowledge the event. A new worker then filters every replay or subsequent system-message event by returning no action, so it never issues an Elasticsearch delete; the existing document remains searchable indefinitely. The migrator likewise only skips these rows and cannot clean a reused destination index. Emit an idempotent delete for system-message events, or require a guaranteed-empty replacement index populated only after old workers are drained, followed by an alias switch and cleanup.

<details>

```
In `search-sync-worker/messages.go` around line 196: During a rolling deployment, an old worker can index system message m1 and acknowledge the event. A new worker then filters every replay or subsequent system-message event by returning no action, so it never issues an Elasticsearch delete; the existing document remains searchable indefinitely. The migrator likewise only skips these rows and cannot clean a reused destination index. Emit an idempotent delete for system-message events, or require a guaranteed-empty replacement index populated only after old workers are drained, followed by an alias switch and cleanup.

```

</details>

julianshen, line 196: _🎯 elasticsearch-correctness_ | _🔴 Blocking_ | _`elasticsearch`_

**Existing system-message documents remain searchable.**

Returning no bulk action only prevents future indexing. Any system message indexed before this deployment remains in Elasticsearch indefinitely, and even a later update or delete event for that message is acknowledged without deleting the existing document. This contradicts the new API guarantee that system messages are never returned. Add a cleanup/reindex migration for existing indices and make system-message events emit an appropriately versioned delete action so stale documents converge out of the index.

<details>

```
In `search-sync-worker/messages.go` around line 196: Returning no bulk action only prevents future indexing. Any system message indexed before this deployment remains in Elasticsearch indefinitely, and even a later update or delete event for that message is acknowledged without deleting the existing document. This contradicts the new API guarantee that system messages are never returned. Add a cleanup/reindex migration for existing indices and make system-message events emit an appropriately versioned delete action so stale documents converge out of the index.

```

</details>

search-service/response.go
```diff
@@ -15,6 +15,12 @@ type rawHit[T any] struct {
 }
 
 type rawResponse[T any] struct {
+	// Shards.Total is 0 when the read pattern matched no concrete index —
+	// allow_no_indices=true turns that into 200 + empty hits, so it is the only
+	// field distinguishing a misconfigured index from a query that found nothing.
+	Shards struct {
```
julianshen, line 21: _🏷️ partial-failure_ | _🟠 Warning_ | _`distributed-system`_

**Partial shard failures are returned as successful search results.**

The response parser retains only `_shards.total` and discards `successful`, `failed`, and failure details. With Elasticsearch's default partial-search behavior, a partition or unavailable replica can produce HTTP 200 with `total > 0`, `failed > 0`, and incomplete or empty hits. The handler then returns those hits as complete; for an empty response it even logs that no document matched because the total is nonzero. Disable partial results on every search request (`allow_partial_search_results=false`) or surface failed shard counts from the parser and return a backend-unavailable error whenever any shard failed.

<details>

```
In `search-service/response.go` around line 21: The response parser retains only `_shards.total` and discards `successful`, `failed`, and failure details. With Elasticsearch's default partial-search behavior, a partition or unavailable replica can produce HTTP 200 with `total > 0`, `failed > 0`, and incomplete or empty hits. The handler then returns those hits as complete; for an empty response it even logs that no document matched because the total is nonzero. Disable partial results on every search request (`allow_partial_search_results=false`) or surface failed shard counts from the parser and return a backend-unavailable error whenever any shard failed.

```

</details>

search-service/deploy/docker-compose.yml
```diff
@@ -30,9 +30,12 @@ services:
       # (OTEL_EXPORTER_PROMETHEUS_PORT, default :2112), not here.
       - SEARCH_HEALTH_ADDR=${SEARCH_HEALTH_ADDR:-:9090}
       # Local dev pins to site-suffixed indices; prod uses aliases owned by ops/IaC.
-      - SEARCH_USER_ROOM_INDEX=${SEARCH_USER_ROOM_INDEX:-user-room-mv-${SITE_ID:-site-local}}
-      - SEARCH_SPOTLIGHT_INDEX=${SEARCH_SPOTLIGHT_INDEX:-spotlight-${SITE_ID:-site-local}-v1}
-      - SEARCH_SPOTLIGHT_ORG_INDEX=${SEARCH_SPOTLIGHT_ORG_INDEX:-spotlightorg-${SITE_ID:-site-local}-v1}
+      # Unprefixed, and MUST match search-sync-worker's SPOTLIGHT_INDEX /
+      # USER_ROOM_INDEX exactly — a mismatch reads an index nobody writes and
+      # returns empty hits with no error.
+      - USER_ROOM_INDEX=${USER_ROOM_INDEX:-user-room-mv-${SITE_ID:-site-local}}
```
julianshen, line 36: _🏷️ deployment-compatibility_ | _🟠 Warning_ | _`distributed-system`_

**The environment rename makes rollback fail at startup.**

The previous binary requires `SEARCH_USER_ROOM_INDEX`, `SEARCH_SPOTLIGHT_INDEX`, and `SEARCH_SPOTLIGHT_ORG_INDEX`, while the updated deployment exports only their unprefixed replacements. After this configuration is deployed, rolling back or restarting an old-version instance makes environment parsing fail, potentially leaving no healthy search-service replicas. Export both spellings with identical values for a compatibility window, then remove the legacy variables only after rollback to the old binary is no longer supported.

<details>

```
In `search-service/deploy/docker-compose.yml` around line 36: The previous binary requires `SEARCH_USER_ROOM_INDEX`, `SEARCH_SPOTLIGHT_INDEX`, and `SEARCH_SPOTLIGHT_ORG_INDEX`, while the updated deployment exports only their unprefixed replacements. After this configuration is deployed, rolling back or restarting an old-version instance makes environment parsing fail, potentially leaving no healthy search-service replicas. Export both spellings with identical values for a compatibility window, then remove the legacy variables only after rollback to the old binary is no longer supported.

```

</details>

search-sync-worker/handler.go
```diff
@@ -83,13 +83,15 @@ func (h *Handler) Add(msg jetstream.Msg) {
...
+		slog.ErrorContext(ctx, "decode payload", "error", err, "subject", msg.Subject(), "consumer", h.collection.ConsumerName())
 		natsutil.Ack(msg, "decode payload failed")
 		return
 	}
 	actions, err := h.collection.BuildAction(data)
 	if err != nil {
-		slog.Error("build action", "error", err)
+		// Every BuildAction error is parse/validation poison — Ack drops it for
+		// good, so this line is the only trace; keep it identifying the message.
+		slog.ErrorContext(ctx, "build action", "error", err, "subject", msg.Subject(), "consumer", h.collection.ConsumerName())
```
julianshen, line 94: _🏷️ delivery-observability_ | _🟠 Warning_ | _`distributed-system`_

**Acked poison events still cannot be identified or replayed.**

A decode or BuildAction failure is permanently acknowledged, but the new log records only the subject and consumer, which are shared by every event. If a mixed-version or malformed membership event fails validation, its search update is dropped and operators have neither a stream sequence/message ID nor a dead-letter copy with which to identify and replay it. Log JetStream stream and consumer sequence plus the NATS message ID, and preferably publish the original event and error metadata to a bounded DLQ before acknowledging it.

<details>

```
In `search-sync-worker/handler.go` around line 94: A decode or BuildAction failure is permanently acknowledged, but the new log records only the subject and consumer, which are shared by every event. If a mixed-version or malformed membership event fails validation, its search update is dropped and operators have neither a stream sequence/message ID nor a dead-letter copy with which to identify and replay it. Log JetStream stream and consumer sequence plus the NATS message ID, and preferably publish the original event and error metadata to a bounded DLQ before acknowledging it.

```

</details>

search-service/main.go
```diff
@@ -93,6 +90,12 @@ type Config struct {
 	Mongo    MongoConfig    `envPrefix:"MONGO_"`
 	UsersAPI UsersAPIConfig `envPrefix:"USERS_API_"`
 	DebugLog logctx.Config  `envPrefix:"DEBUG_LOG_"`
+	// UNPREFIXED on purpose — must match search-sync-worker / es-index-migrator
+	// exactly. On SearchConfig they would pick up envPrefix:"SEARCH_" and drift
+	// from the writer silently (wildcard read + allow_no_indices ⇒ empty hits).
+	UserRoomIndex     string `env:"USER_ROOM_INDEX,required,notEmpty"`
```
julianshen, line 96: _🏷️ deployment_ | _🟠 Warning_ | _`elasticsearch`_

**The index environment rename has no rolling compatibility.**

Existing deployments necessarily provide SEARCH_USER_ROOM_INDEX, SEARCH_SPOTLIGHT_INDEX, and SEARCH_SPOTLIGHT_ORG_INDEX because those were the previously required names. New instances now require only the unprefixed names and will fail startup, while old instances cannot use only the new names. The compose change covers local deployment but not the production IaC referenced in its comment. Temporarily accept both spellings with a defined precedence, or require a staged rollout that first publishes both sets of variables before deploying this binary.

<details>

```
In `search-service/main.go` around line 96: Existing deployments necessarily provide SEARCH_USER_ROOM_INDEX, SEARCH_SPOTLIGHT_INDEX, and SEARCH_SPOTLIGHT_ORG_INDEX because those were the previously required names. New instances now require only the unprefixed names and will fail startup, while old instances cannot use only the new names. The compose change covers local deployment but not the production IaC referenced in its comment. Temporarily accept both spellings with a defined precedence, or require a staged rollout that first publishes both sets of variables before deploying this binary.

```

</details>

room-worker/teamsroomcreate.go
```diff
@@ -261,8 +261,21 @@ func (h *Handler) federateTeamsMembership(ctx context.Context, room *model.Room,
...
+	if err != nil {
+		return fmt.Errorf("marshal internal inbox envelope: %w", err)
+	}
 	seed := fmt.Sprintf("%s:%s:%d", room.ID, eventType, acceptedAt.UnixMilli())
```
julianshen, line 277: _🏷️ nats-operational-safety_ | _🟠 Warning_ | _`nats`_

**Previously ACKed bare membership events remain unrecoverable**

This fixes future publishes on the internal inbox subject, but it does not repair events emitted by the previous bare-payload producer. The diff confirms search-sync-worker treats those as BuildAction errors, and Handler.AddWithContext ACKs every such error, so JetStream will not redeliver them. Existing rooms can therefore retain missing search membership state after deployment; during a rolling deployment, events from old room-worker instances have the same fate. Add an explicit spotlight/subscription backfill or republish current membership snapshots with fresh deduplication IDs, and document or enforce rollout handling for old publishers.

<details>

```
In `room-worker/teamsroomcreate.go` around line 277: This fixes future publishes on the internal inbox subject, but it does not repair events emitted by the previous bare-payload producer. The diff confirms search-sync-worker treats those as BuildAction errors, and Handler.AddWithContext ACKs every such error, so JetStream will not redeliver them. Existing rooms can therefore retain missing search membership state after deployment; during a rolling deployment, events from old room-worker instances have the same fate. Add an explicit spotlight/subscription backfill or republish current membership snapshots with fresh deduplication IDs, and document or enforce rollout handling for old publishers.

```

</details>

search-service/handler.go
```diff
@@ -165,13 +166,32 @@ func (h *handler) searchRooms(c *natsrouter.Context, req model.SearchRoomsReques
...
 }
 
+// logEmptyResult names the searched pattern on a zero-result query, since
+// allow_no_indices=true makes a misconfigured index look like a normal miss.
+// WARN only for zero shards (always broken) and without the user-typed query;
+// routine misses log at flow level, which is X-Debug-gated and keeps it.
+func logEmptyResult(ctx context.Context, kind, pattern, account, query string, shards int) {
+	requestID := natsutil.RequestIDFromContext(ctx)
+	if shards == 0 {
+		slog.WarnContext(ctx, "empty search result: read pattern matched no index",
```
julianshen, line 186: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Missing indices trigger a synchronous warning for every request.**

When a read pattern matches zero shards, every room or organization search calls slog.WarnContext. Under a misconfigured deployment or an index-creation outage, all concurrent searches take this path; the standard slog handler serializes writes to its output, so search traffic can contend on logging and generate an unbounded volume of duplicate warnings, worsening latency and log backpressure during the incident. Rate-limit or coalesce this warning per kind/pattern (and expose a metric or perform an index check at startup) rather than emitting it once per request.

<details>

```
In `search-service/handler.go` around line 186: When a read pattern matches zero shards, every room or organization search calls slog.WarnContext. Under a misconfigured deployment or an index-creation outage, all concurrent searches take this path; the standard slog handler serializes writes to its output, so search traffic can contend on logging and generate an unbounded volume of duplicate warnings, worsening latency and log backpressure during the incident. Rate-limit or coalesce this warning per kind/pattern (and expose a metric or perform an index check at startup) rather than emitting it once per request.

```

</details>

search-service/handler.go
```diff
@@ -165,13 +166,32 @@ func (h *handler) searchRooms(c *natsrouter.Context, req model.SearchRoomsReques
...
 	return &model.SearchRoomsResponse{Rooms: rooms}, nil
 }
 
+// logEmptyResult names the searched pattern on a zero-result query, since
+// allow_no_indices=true makes a misconfigured index look like a normal miss.
+// WARN only for zero shards (always broken) and without the user-typed query;
+// routine misses log at flow level, which is X-Debug-gated and keeps it.
+func logEmptyResult(ctx context.Context, kind, pattern, account, query string, shards int) {
+	requestID := natsutil.RequestIDFromContext(ctx)
+	if shards == 0 {
```
julianshen, line 185: _🏷️ failure-handling_ | _🟠 Warning_ | _`silent-failure`_

**A missing index is returned as a successful empty search result.**

With `allow_no_indices=true`, a typoed or nonexistent read pattern produces HTTP 200 with `_shards.total == 0`. This branch only logs a warning; both callers then return an empty response with a nil error, so clients cannot distinguish total search unavailability from a legitimate miss. Return a backend/unavailable error when `shards == 0`, or configure Elasticsearch to reject missing indices, while retaining normal empty success when shards exist.

<details>

```
In `search-service/handler.go` around line 185: With `allow_no_indices=true`, a typoed or nonexistent read pattern produces HTTP 200 with `_shards.total == 0`. This branch only logs a warning; both callers then return an empty response with a nil error, so clients cannot distinguish total search unavailability from a legitimate miss. Return a backend/unavailable error when `shards == 0`, or configure Elasticsearch to reject missing indices, while retaining normal empty success when shards exist.

```

</details>

search-service/response.go
```diff
@@ -15,6 +15,12 @@ type rawHit[T any] struct {
 }
 
 type rawResponse[T any] struct {
+	// Shards.Total is 0 when the read pattern matched no concrete index —
+	// allow_no_indices=true turns that into 200 + empty hits, so it is the only
```
julianshen, line 19: _🏷️ failure-handling_ | _🟠 Warning_ | _`silent-failure`_

**Shard failures and timeouts are silently accepted as complete results.**

The response parser records only `_shards.total`. Elasticsearch can return HTTP 200 with `timed_out: true` or `_shards.failed > 0` and partial or empty hits. Those responses currently produce nil errors; when total is nonzero, an empty partial response is even classified as an ordinary miss. Parse the timeout and failed-shard fields and reject incomplete responses, or disable partial search results at the Elasticsearch request layer.

<details>

```
In `search-service/response.go` around line 19: The response parser records only `_shards.total`. Elasticsearch can return HTTP 200 with `timed_out: true` or `_shards.failed > 0` and partial or empty hits. Those responses currently produce nil errors; when total is nonzero, an empty partial response is even classified as an ordinary miss. Parse the timeout and failed-shard fields and reject incomplete responses, or disable partial search results at the Elasticsearch request layer.

```

</details>

search-sync-worker/messages.go
julianshen, line 196: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Existing indexed system messages are never removed**

Returning no action only prevents future system-message writes. The previous implementation indexed these messages, so active indices can already contain them; this branch also suppresses any later delete event that could remove such a document. Those existing documents therefore remain searchable despite the new API guarantee. Ensure deployment rebuilds and cuts over to a clean index, run an explicit delete-by-query migration, or otherwise emit deletion actions for previously indexed system messages.

<details>

```
In `search-sync-worker/messages.go` around line 196: Returning no action only prevents future system-message writes. The previous implementation indexed these messages, so active indices can already contain them; this branch also suppresses any later delete event that could remove such a document. Those existing documents therefore remain searchable despite the new API guarantee. Ensure deployment rebuilds and cuts over to a clean index, run an explicit delete-by-query migration, or otherwise emit deletion actions for previously indexed system messages.

```

</details>

amdla: Expected - we will clear the data and run it through indexer once again

amdla: Expected - we will clear the data and run it through indexer once again

mliu33 (approved):
Awesome, thanks

PR #223, opened 2026-08-11
[NATS Tuning] Make NATS drain actually block, and make slow consumers visible

hmchangw:
## The bug

`nats.Conn.Drain()` does not drain. From `nats.go@v1.50.0:6055-6075`, it flips the connection to `DRAINING_SUBS`, spawns `go nc.drainConnection()`, and returns `nil`. All the real work — draining each subscription, `FlushTimeout(5s)`, then `Close()` — happens on that goroutine.

Every service did this:

```go
func(ctx context.Context) error { return nc.Drain() },
```

The hook returned `nil` instantly, `shutdown.Wait` moved on, `main` returned, and `drainConnection` died wherever it got to. Whatever sat in the client write buffer at that moment was lost: JetStream acks (→ redelivery), request/reply responses (→ caller timeout), OUTBOX forwards.

The library's own doc comment names the fix — *"Use the ClosedCB option to know when the connection has moved from draining to closed."* Nothing in the repo did.

The trap was documented at exactly one site, `admin-service/main.go`, whose comment said Drain "returns immediately and finishes in the background". That reasoning was locally sound (an HTTP-only service with an idle connection) but was never generalised to the 22 sites where the connection is not idle.

## What changed

**`natsutil.Drain(ctx, conn)`** — starts the drain, then blocks on the connection's `CLOSED` status transition or `ctx`, whichever comes first. `ErrConnectionClosed` and `ErrConnectionReconnecting` are treated as benign (they surface on ordinary restarts today and log as shutdown errors). This mirrors the register-then-drain-then-wait pattern `pkg/natsrouter.Shutdown` already uses one level down for subscriptions.

**`DrainTimeout` set to 10s** (library default is 30s — larger than the entire shutdown budget, so it could never fire). Worst case is 10s + the internal `FlushTimeout(5s)` = 15s. Note `shutdown.Wait` creates **one** context for the whole shutdown and runs hooks sequentially over it, so the 25s is shared, not per-hook; 15s leaves ~10s for the remaining hooks, which are sub-second in practice.

**Slow consumer diagnostics.** The shared `nats.ErrorHandler` discarded its `*nats.Subscription`, so a slow consumer logged a bare `nats async error` — no subject, no queue, no drop count, nothing alertable. It now logs subject, queue, drop count and pending/limit counters, at ERROR once messages have actually dropped and WARN before that, and increments `nats_slow_consumer_events_total`.

That counter counts **episodes, not messages**, deliberately: `Subscription.Dropped()` is cumulative and never reset outside unsubscribe, and the callback fires only on the Active→SlowConsumer transition, so adding `Dropped()` per episode would re-add every earlier episode's drops (3 dropped then 5 more would report 3 + 8 = 11). The callback is also dispatched through an async queue, so a `Dropped()` read inside it isn't the count at episode time either. Exact numbers live in the log fields.

**30 call sites converted**, across five shapes:

| Shape | Count | Treatment |
|---|---|---|
| `shutdown.Wait` hook lambda | 19 | one-line swap; 5 needed the discarded `ctx` named |
| Inline in a combined hook | 2 | converted in place, existing error swallow kept |
| `defer` in a `run` function | 2 | `context.WithoutCancel` + 20s — the parent ctx is already cancelled when the defer runs |
| `nc.Close()` | 2 | `hr-sync-worker` split into its own hook; `teams-hr-sync` converted for consistency |
| `data-migration` shutdown paths | 5 | 4 hooks + `oplog-connector`'s own shutdown function |

**`media-service` shutdown chain fixed** (found by the final whole-branch review). It ran `go shutdown.Wait(...)` and then blocked on `srv.ListenAndServe()` directly, unlike every sibling Gin service. When the third hook closed the listener, `ListenAndServe` returned and the process exited — abandoning the rest of that hook and `obsShutdown` entirely. The structure predates this branch, but the branch worsens it: with one shared 25s context, a drain that now legitimately occupies up to 15s leaves the later hooks on a nearly-expired deadline, and `srv.Shutdown` with an expired context returns without waiting for in-flight requests. It now uses the same `srvErr`/`shutdownDone` pattern as `admin-service` and its six siblings; the hook list is unchanged.

**14 sites deliberately not converted.** The remaining `_ = nc.Drain()` calls in `data-migration` are startup error paths followed immediately by `os.Exit(1)` — nothing has been published, so there is nothing to flush, and blocking up to 15s before an error exit would be strictly worse. `tools/loadgen` (11 files) is a dev harness and is out of scope. `pkg/testutil/nats_binary.go` is test-harness teardown.

## Verification

- `make test` — full repo with `-race`, all packages pass
- `make lint` — 0 issues
- `make build` — every touched service
- `pkg/natsutil` coverage 87.8% (above the 80% floor, below the 90% target). `Drain` sits at 77.3%: the residual is a nil guard that cannot fire, a `default:` branch the library can never reach, and inherently racy windows. Not padded.

**SAST is partially unverified in this environment.** `gosec` passes with no medium+ findings, and semgrep's in-repo rules (`.semgrep/`, run with the gate's real flags) return 0 findings. `govulncheck` and semgrep's registry rulesets (`p/golang`, `p/security-audit`) could not run — the sandbox's egress policy returns 403 for `vuln.go.dev` and `semgrep.dev`. This branch touches no `go.mod`/`go.sum`, so `govulncheck`'s verdict should match `main`'s, but that is reasoning rather than a result. **CI must run the full gate.**

## Behaviour change to expect

Services now take up to ~15s to exit instead of exiting instantly. That is the point, but it will read as a pod-restart regression on first deploy. A service sitting at the full 10s drain timeout has a wedged subscription handler — a real finding, not a bug in this change.

## Known limitations (deliberately out of scope)

1. **The drain can be starved of budget.** `shutdown.Wait` shares one 25s context across sequential hooks, and in every JetStream worker the preceding `wg.Wait()` hook bounds itself on that *same* context. A wedged handler consumes the whole budget, and the drain then runs with an already-expired context. Fixing it means reserving a drain sub-budget inside `pkg/shutdown` — a different change.
2. **`hr-sync-worker` does not wait for in-flight handlers.** `cc.Stop()` is non-blocking and its cleanup calls `Unsubscribe()` rather than `Drain()`, removing the subscription from `nc.subs` synchronously — so `natsutil.Drain` has nothing left to wait on for that consumer. It reliably recovers an ack already queued in the write buffer (the bug fixed here), but not a handler still executing pre-ack. The loss window narrows from "every shutdown" to "a shutdown landing mid-message". Closing it needs a `sync.WaitGroup` around the `Consume` callback, and may be a codebase-wide question — CLAUDE.md's sequential-consumer pattern doesn't mandate one the way its high-throughput pattern does.
3. **`SetPendingLimits` tuning** for `broadcast-worker` fan-out. The remedy for slow-consumer pressure is raising the per-subscription buffer, but picking a number before the new counter shows real drop rates would be guessing. Diagnose first.
4. `media-service` drains NATS *before* shutting down its HTTP listener, unlike the workers. Pre-existing and left as-is — changing hook order is a behavioural change beyond the fix above.
5. A comment in `outbox-worker/main.go:200` still refers to `nc.Drain()` in prose; conceptually accurate, but the identifier no longer appears in that file.

Payload-size checks and `NoResponders` error mapping are **not** in this PR — they are #239.

## Notes for review

No wire, schema, or config change; nothing here is client-facing, so `docs/client-api.md` and its derived views are untouched. The spec and implementation plan are included under `docs/superpowers/`.

Each of the 7 implementation tasks was reviewed independently as it landed, followed by a whole-branch review that produced the `media-service` fix above. That review also verified, against the library source rather than the branch's claims, that `natsutil.Drain` registers its status listener *before* calling `Drain()`, that the three `WithoutCancel` batch jobs match each file's actual context construction, and that the new metric's `subject`/`queue` attributes are bounded (the codebase uses no `UseOldRequestStyle`, so mux'd request/reply shares one wildcard inbox per connection rather than a subscription per call).

Three defects originated in the plan rather than the implementation and were corrected in-branch: the metric overcounting described above, a comment asserting SIGTERM cancellation in a binary that installs no signal handler, and a build command that named a service while skipping the subpackage it edited.

Verified to merge cleanly against current `main`; the newly-landed `client-update-service` has no NATS connection, so the sweep remains complete.

## Summary by CodeRabbit

* **Improvements**
  * Services now complete a more graceful, context-aware shutdown, allowing in-flight messaging work to finish before connections close.
  * Shutdown operations are bounded by timeouts to prevent services from hanging indefinitely.
  * Connection cleanup is more reliable during restarts, including improved handling of already-closed or reconnecting connections.
  * Slow messaging consumers now generate clearer diagnostic logs and monitoring metrics, helping identify dropped or delayed messages.
* **Documentation**
  * Added guidance covering connection draining, shutdown behavior, observability, testing, and rollout considerations.

### DIFF
--- a/data-migration/oplog-connector/main.go
+++ b/data-migration/oplog-connector/main.go
@@ -94,7 +94,7 @@ func main() {
 		func(context.Context) error { conn.beginShutdown(); return nil },
 		func(ctx context.Context) error { return conn.awaitWatchers(ctx) },
 		func(ctx context.Context) error { return healthStop(ctx) },
-		func(context.Context) error { return conn.nc.Drain() },
+		func(ctx context.Context) error { return natsutil.Drain(ctx, conn.nc) },
 		func(ctx context.Context) error { mongoutil.Disconnect(ctx, conn.client); return nil },
 		func(ctx context.Context) error { return obsShutdown(ctx) },
 	)
@@ -231,7 +231,11 @@ func (c *connector) Close() {
 	if err := c.awaitWatchers(wctx); err != nil {
 		slog.Warn("watcher drain incomplete", "error", err)
 	}
-	_ = c.nc.Drain()
+	dctx, dcancel := context.WithTimeout(context.Background(), 20*time.Second)
+	defer dcancel()
+	if err := natsutil.Drain(dctx, c.nc); err != nil {
+		slog.Warn("nats drain incomplete", "error", err)
+	}
 	mongoutil.Disconnect(context.Background(), c.client)
 }
 
--- a/media-service/main.go
+++ b/media-service/main.go
@@ -104,7 +104,7 @@ func run() error {
 
 	go shutdown.Wait(ctx, 25*time.Second,
 		func(ctx context.Context) error { return router.Shutdown(ctx) },
-		func(_ context.Context) error { return nc.Drain() },
+		func(ctx context.Context) error { return natsutil.Drain(ctx, nc) },
 		func(ctx context.Context) error {
 			slog.Info("shutting down media-service")
 			return srv.Shutdown(ctx)
--- a/media-service/main.go
+++ b/media-service/main.go
@@ -102,20 +102,30 @@ func run() error {
 		WriteTimeout: 30 * time.Second,
 	}
 
-	go shutdown.Wait(ctx, 25*time.Second,
-		func(ctx context.Context) error { return router.Shutdown(ctx) },
-		func(ctx context.Context) error { return natsutil.Drain(ctx, nc) },
-		func(ctx context.Context) error {
-			slog.Info("shutting down media-service")
-			return srv.Shutdown(ctx)
-		},
-		// obsShutdown LAST so all prior teardown telemetry is exported.
-		func(ctx context.Context) error { return obsShutdown(ctx) },
-	)
+	srvErr := make(chan error, 1)
+	go func() {
+		slog.Info("media-service listening", "port", cfg.Port, "site", cfg.SiteID)
+		srvErr <- srv.ListenAndServe()
+	}()
 
-	slog.Info("media-service listening", "port", cfg.Port, "site", cfg.SiteID)
-	if err := srv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+	shutdownDone := make(chan struct{})
+	go func() {
+		defer close(shutdownDone)
+		shutdown.Wait(ctx, 25*time.Second,
+			func(ctx context.Context) error { return router.Shutdown(ctx) },
+			func(ctx context.Context) error { return natsutil.Drain(ctx, nc) },
+			func(ctx context.Context) error {
+				slog.Info("shutting down media-service")
+				return srv.Shutdown(ctx)
+			},
+			// obsShutdown LAST so all prior teardown telemetry is exported.
+			func(ctx context.Context) error { return obsShutdown(ctx) },
+		)
+	}()
+
+	if err := <-srvErr; err != nil && !errors.Is(err, http.ErrServerClosed) {
 		return fmt.Errorf("listen and serve: %w", err)
 	}
+	<-shutdownDone
 	return nil
 }
--- a/pkg/natsutil/connect_test.go
+++ b/pkg/natsutil/connect_test.go
@@ -70,6 +70,6 @@ func TestConnect_SetsDrainTimeout(t *testing.T) {
 	require.NoError(t, err)
 	t.Cleanup(conn.NatsConn().Close)
 
-	require.Equal(t, 15*time.Second, conn.NatsConn().Opts.DrainTimeout,
-		"DrainTimeout must be under the 25s shutdown budget, not the 30s library default")
+	require.Equal(t, 10*time.Second, conn.NatsConn().Opts.DrainTimeout,
+		"DrainTimeout must leave headroom in the shared 25s shutdown budget, not the 30s library default")
 }
--- a/pkg/natsutil/connect_test.go
+++ b/pkg/natsutil/connect_test.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 	"go.opentelemetry.io/otel/propagation"
@@ -61,4 +62,14 @@ func TestConnect_PresentCredsFilePassesPrecheck(t *testing.T) {
 	_, err := natsutil.Connect(context.Background(), "nats://127.0.0.1:1", path,
 		noop.NewTracerProvider(), propagation.TraceContext{}, false)
 	require.False(t, errors.Is(err, os.ErrNotExist), "precheck should pass when file exists, got: %v", err)
+}
+
+func TestConnect_SetsDrainTimeout(t *testing.T) {
+	conn, err := natsutil.Connect(context.Background(), startTestServer(t), "",
+		noop.NewTracerProvider(), propagation.TraceContext{}, false)
+	require.NoError(t, err)
+	t.Cleanup(conn.NatsConn().Close)
+
+	require.Equal(t, 15*time.Second, conn.NatsConn().Opts.DrainTimeout,
+		"DrainTimeout must be under the 25s shutdown budget, not the 30s library default")
 }
--- a/pkg/natsutil/drain.go
+++ b/pkg/natsutil/drain.go
@@ -56,6 +56,18 @@ func Drain(ctx context.Context, conn *o11ynats.Conn) error {
 	case <-ch:
 		return nil
 	case <-ctx.Done():
+		// select picks pseudo-randomly when both cases are ready at once, so a
+		// drain that finished in the same instant ctx expired can still land
+		// here. Re-check before reporting failure so a completed drain is
+		// never misreported as incomplete.
+		select {
+		case <-ch:
+			return nil
+		default:
+		}
+		if nc.IsClosed() {
+			return nil
+		}
 		return fmt.Errorf("nats drain incomplete: %w", ctx.Err())
 	}
 }
--- a/pkg/natsutil/drain.go
+++ b/pkg/natsutil/drain.go
@@ -1,0 +1,61 @@
+package natsutil
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+
+	o11ynats "github.com/flywindy/o11y/nats"
+	"github.com/nats-io/nats.go"
+)
+
+// Drain puts the connection into the drain state and blocks until it reaches
+// CLOSED or ctx expires.
+//
+// nats.Conn.Drain only *starts* the drain — it spawns drainConnection on a
+// goroutine and returns nil immediately — so a shutdown hook that returns
+// nc.Drain() directly returns before subscriptions have drained and before the
+// final publish flush, and the process exits with buffered JetStream acks and
+// request/reply responses still in the write buffer.
+func Drain(ctx context.Context, conn *o11ynats.Conn) error {
+	if conn == nil {
+		return nil
+	}
+	nc := conn.NatsConn()
+	if nc == nil {
+		return nil
+	}
+
+	// Register the listener before calling Drain so a fast close cannot land in
+	// the gap between the two calls, and re-check IsClosed after, to cover a
+	// close that completed before the listener was in place. Together these
+	// close both sides of the race.
+	ch := nc.StatusChanged(nats.CLOSED)
+	defer nc.RemoveStatusListener(ch)
+
+	if err := nc.Drain(); err != nil {
+		switch {
+		case errors.Is(err, nats.ErrConnectionClosed):
+			return nil
+		case errors.Is(err, nats.ErrConnectionReconnecting):
+			// Drain hard-closed the connection. Buffered publishes are gone,
+			// but there was no reachable server to flush them to — a normal
+			// shutdown-during-outage, not a failure worth reporting upward.
+			slog.WarnContext(ctx, "nats drain skipped: connection reconnecting")
+			return nil
+		default:
+			return fmt.Errorf("start nats drain: %w", err)
+		}
+	}
+	if nc.IsClosed() {
+		return nil
+	}
+
+	select {
+	case <-ch:
+		return nil
+	case <-ctx.Done():
+		return fmt.Errorf("nats drain incomplete: %w", ctx.Err())
+	}
+}
--- a/pkg/natsutil/drain_test.go
+++ b/pkg/natsutil/drain_test.go
@@ -21,12 +21,23 @@ import (
 // production option wiring — including DrainTimeout — is what gets exercised.
 func startTestServer(t *testing.T) string {
 	t.Helper()
+	_, url := startTestServerWithHandle(t)
+	return url
+}
+
+// startTestServerWithHandle is startTestServer but also returns the
+// *natsserver.Server handle, for tests that need to shut the server down
+// mid-test (e.g. to drive a connection into RECONNECTING) rather than only
+// at t.Cleanup.
+func startTestServerWithHandle(t *testing.T) (*natsserver.Server, string) {
+	t.Helper()
 	ns, err := natsserver.NewServer(&natsserver.Options{Port: -1})
 	require.NoError(t, err)
 	ns.Start()
 	require.True(t, ns.ReadyForConnections(5*time.Second))
 	t.Cleanup(ns.Shutdown)
-	return ns.ClientURL()
+	url := ns.ClientURL()
+	return ns, url
 }
 
 func testConn(t *testing.T, url string) *o11ynats.Conn {
@@ -58,9 +69,11 @@ func TestDrain_WaitsForInFlightHandler(t *testing.T) {
 	url := startTestServer(t)
 	conn := testConn(t, url)
 
+	entered := make(chan struct{})
 	release := make(chan struct{})
 	handlerDone := make(chan struct{})
 	_, err := conn.NatsConn().Subscribe("drain.slow", func(*nats.Msg) {
+		close(entered)
 		<-release
 		close(handlerDone)
 	})
@@ -70,6 +83,12 @@ func TestDrain_WaitsForInFlightHandler(t *testing.T) {
 	pub := testConn(t, url)
 	require.NoError(t, pub.NatsConn().Publish("drain.slow", []byte("x")))
 	require.NoError(t, pub.NatsConn().Flush())
+
+	// pub.Flush only guarantees the server processed the PUB, not that this
+	// subscriber's readLoop has dispatched it yet. Wait for the handler to
+	// actually start before racing it against Drain, so a slow dispatch can't
+	// let Drain's sub.Drain() see an empty queue and skip the handler.
+	<-entered
 
 	drained := make(chan error, 1)
 	go func() {
@@ -109,6 +128,31 @@ func TestDrain_CtxExpiredReturnsError(t *testing.T) {
 	err = natsutil.Drain(ctx, conn)
 	require.Error(t, err)
 	require.ErrorIs(t, err, context.DeadlineExceeded)
+}
+
+// The production path this covers: Connect sets MaxReconnects(-1), so a pod
+// that receives SIGTERM while NATS is unreachable sits in RECONNECTING
+// indefinitely and Drain is called against that state, not a healthy one.
+// nc.Drain() hard-closes the connection in that case (discarding the write
+// buffer) and returns ErrConnectionReconnecting rather than nil — Drain must
+// still report a clean nil, since the connection genuinely did reach CLOSED,
+// not surface the library's internal sentinel as a failure.
+func TestDrain_ReconnectingHardClosesAndReturnsNil(t *testing.T) {
+	ns, url := startTestServerWithHandle(t)
+	conn := testConn(t, url)
+
+	_, err := conn.NatsConn().Subscribe("drain.reconnecting", func(*nats.Msg) {})
+	require.NoError(t, err)
+
+	ns.Shutdown()
+	require.Eventually(t, conn.NatsConn().IsReconnecting, 5*time.Second, 10*time.Millisecond,
+		"connection must enter RECONNECTING once the server goes away")
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	require.NoError(t, natsutil.Drain(ctx, conn))
+	require.True(t, conn.NatsConn().IsClosed())
 }
 
 func TestDrain_AlreadyClosedIsNotAnError(t *testing.T) {
--- a/pkg/natsutil/drain_test.go
+++ b/pkg/natsutil/drain_test.go
@@ -1,0 +1,133 @@
+package natsutil_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	o11ynats "github.com/flywindy/o11y/nats"
+	natsserver "github.com/nats-io/nats-server/v2/server"
+	"github.com/nats-io/nats.go"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/trace/noop"
+
+	"github.com/hmchangw/chat/pkg/natsutil"
+)
+
+// startTestServer starts an embedded NATS server and returns its client URL.
+// Separate from startTestNATS (which hands back a raw *nats.Conn) because these
+// tests need to build a real *o11ynats.Conn through natsutil.Connect, so the
+// production option wiring — including DrainTimeout — is what gets exercised.
+func startTestServer(t *testing.T) string {
+	t.Helper()
+	ns, err := natsserver.NewServer(&natsserver.Options{Port: -1})
+	require.NoError(t, err)
+	ns.Start()
+	require.True(t, ns.ReadyForConnections(5*time.Second))
+	t.Cleanup(ns.Shutdown)
+	return ns.ClientURL()
+}
+
+func testConn(t *testing.T, url string) *o11ynats.Conn {
+	t.Helper()
+	conn, err := natsutil.Connect(context.Background(), url, "",
+		noop.NewTracerProvider(), propagation.TraceContext{}, false)
+	require.NoError(t, err)
+	return conn
+}
+
+func TestDrain_ClosesConnection(t *testing.T) {
+	conn := testConn(t, startTestServer(t))
+
+	_, err := conn.NatsConn().Subscribe("drain.test", func(*nats.Msg) {})
+	require.NoError(t, err)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	require.NoError(t, natsutil.Drain(ctx, conn))
+	require.True(t, conn.NatsConn().IsClosed(),
+		"Drain must not return until the connection has reached CLOSED")
+}
+
+// The regression this whole change exists for: Drain must not return while a
+// handler is still running. Before the fix nats.Conn.Drain returned instantly
+// and this assertion fails with handlerDone still false.
+func TestDrain_WaitsForInFlightHandler(t *testing.T) {
+	url := startTestServer(t)
+	conn := testConn(t, url)
+
+	release := make(chan struct{})
+	handlerDone := make(chan struct{})
+	_, err := conn.NatsConn().Subscribe("drain.slow", func(*nats.Msg) {
+		<-release
+		close(handlerDone)
+	})
+	require.NoError(t, err)
+	require.NoError(t, conn.NatsConn().Flush())
+
+	pub := testConn(t, url)
+	require.NoError(t, pub.NatsConn().Publish("drain.slow", []byte("x")))
+	require.NoError(t, pub.NatsConn().Flush())
+
+	drained := make(chan error, 1)
+	go func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		drained <- natsutil.Drain(ctx, conn)
+	}()
+
+	select {
+	case <-drained:
+		t.Fatal("Drain returned while the handler was still in flight")
+	case <-time.After(200 * time.Millisecond):
+	}
+
+	close(release)
+	<-handlerDone
+	require.NoError(t, <-drained)
+}
+
+func TestDrain_CtxExpiredReturnsError(t *testing.T) {
+	url := startTestServer(t)
+	conn := testConn(t, url)
+
+	release := make(chan struct{})
+	t.Cleanup(func() { close(release) })
+	_, err := conn.NatsConn().Subscribe("drain.wedged", func(*nats.Msg) { <-release })
+	require.NoError(t, err)
+	require.NoError(t, conn.NatsConn().Flush())
+
+	pub := testConn(t, url)
+	require.NoError(t, pub.NatsConn().Publish("drain.wedged", []byte("x")))
+	require.NoError(t, pub.NatsConn().Flush())
+
+	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+	defer cancel()
+
+	err = natsutil.Drain(ctx, conn)
+	require.Error(t, err)
+	require.ErrorIs(t, err, context.DeadlineExceeded)
+}
+
+func TestDrain_AlreadyClosedIsNotAnError(t *testing.T) {
+	conn := testConn(t, startTestServer(t))
+	conn.NatsConn().Close()
+
+	require.NoError(t, natsutil.Drain(context.Background(), conn))
+}
+
+func TestDrain_Idempotent(t *testing.T) {
+	conn := testConn(t, startTestServer(t))
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	require.NoError(t, natsutil.Drain(ctx, conn))
+	require.NoError(t, natsutil.Drain(ctx, conn))
+}
+
+func TestDrain_NilConnIsNotAnError(t *testing.T) {
+	require.NoError(t, natsutil.Drain(context.Background(), nil))
+}
--- a/pkg/natsutil/slowconsumer.go
+++ b/pkg/natsutil/slowconsumer.go
@@ -1,0 +1,90 @@
+package natsutil
+
+import (
+	"context"
+	"log/slog"
+
+	"github.com/nats-io/nats.go"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/metric/noop"
+)
+
+// slowConsumerDropped counts messages dropped by a subscription that could not
+// keep up, tagged by subject and queue group.
+var slowConsumerDropped metric.Int64Counter
+
+func init() {
+	var err error
+	slowConsumerDropped, err = otel.Meter("nats").Int64Counter(
+		"nats_slow_consumer_dropped_total",
+		metric.WithDescription("Messages dropped by a NATS subscription that could not keep up, by subject and queue group"),
+	)
+	if err != nil {
+		// Fall back to a no-op counter so recording stays safe even if the
+		// global meter provider is not initialised at package init time.
+		slowConsumerDropped, _ = noop.NewMeterProvider().Meter("nats").
+			Int64Counter("nats_slow_consumer_dropped_total")
+	}
+}
+
+// slowConsumerFields builds the slog fields describing a slow consumer.
+// Every accessor is best-effort: sub is nil for connection-level async errors,
+// and Dropped/Pending/PendingLimits all error on an already-closed
+// subscription, which happens routinely during shutdown. A diagnostic path
+// must never panic or mask the event it is reporting.
+func slowConsumerFields(sub *nats.Subscription) []any {
+	if sub == nil {
+		return []any{"subject", "unknown"}
+	}
+	fields := []any{"subject", sub.Subject, "queue", sub.Queue}
+	if dropped, err := sub.Dropped(); err == nil {
+		fields = append(fields, "dropped", dropped)
+	}
+	// nbytes, not bytes — a local named bytes shadows the stdlib package and
+	// trips the linter's shadow check.
+	if msgs, nbytes, err := sub.Pending(); err == nil {
+		fields = append(fields, "pending_msgs", msgs, "pending_bytes", nbytes)
+	}
+	if msgs, nbytes, err := sub.PendingLimits(); err == nil {
+		fields = append(fields, "limit_msgs", msgs, "limit_bytes", nbytes)
+	}
+	return fields
+}
+
+// subDropped returns the subscription's drop count, or 0 when it cannot be read.
+func subDropped(sub *nats.Subscription) int {
+	if sub == nil {
+		return 0
+	}
+	dropped, err := sub.Dropped()
+	if err != nil {
+		return 0
+	}
+	return dropped
+}
+
+// logSlowConsumer reports a slow consumer and records the drop count.
+func logSlowConsumer(log *slog.Logger, sub *nats.Subscription) {
+	dropped := subDropped(sub)
+	logSlowConsumerAt(log, dropped, slowConsumerFields(sub))
+	if dropped > 0 && sub != nil {
+		slowConsumerDropped.Add(context.Background(), int64(dropped),
+			metric.WithAttributes(
+				attribute.String("subject", sub.Subject),
+				attribute.String("queue", sub.Queue),
+			))
+	}
+}
+
+// logSlowConsumerAt picks the level: dropped messages on a core subscription are
+// unrecoverable loss and should page, so they log at ERROR. A slow consumer that
+// has not dropped anything yet is a warning.
+func logSlowConsumerAt(log *slog.Logger, dropped int, fields []any) {
+	if dropped > 0 {
+		log.Error("nats slow consumer, messages dropped", fields...)
+		return
+	}
+	log.Warn("nats slow consumer", fields...)
+}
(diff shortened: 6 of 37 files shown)


julianshen:
**Actionable comments posted: 13** — 🔴 6 blocking · 🟠 7 warning

<details>
<summary>📒 Files reviewed (37)</summary>

* `admin-service/main.go`
* `bot-room-service/main.go`
* `broadcast-worker/main.go`
* `data-migration/oplog-collections-transformer/main.go`
* `data-migration/oplog-connector/main.go`
* `data-migration/oplog-direct-transfer/main.go`
* `data-migration/oplog-transformer/main.go`
* `docs/superpowers/plans/2026-08-11-nats-connection-handling.md`
* `docs/superpowers/specs/2026-08-11-nats-connection-handling-design.md`
* `hr-sync-worker/main.go`
* `inbox-worker/main.go`
* `media-service/main.go`
* `message-worker/main.go`
* `outbox-worker/main.go`
* `pkg/natsutil/connect.go`
* `pkg/natsutil/connect_test.go`
* `pkg/natsutil/drain.go`
* `pkg/natsutil/drain_test.go`
* `pkg/natsutil/slowconsumer.go`
* `pkg/natsutil/slowconsumer_test.go`
* `push-notification-service/main.go`
* `room-service/main.go`
* `room-worker/main.go`
* `search-service/main.go`
* `teams-hr-sync/main.go`
* `teams-room-creation/main.go`
* `user-service/main.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `silent-failure`
* `tgd-review`

</details>

pkg/natsutil/drain.go
```diff
@@ -0,0 +1,73 @@
...
+	if err := nc.Drain(); err != nil {
+		switch {
+		case errors.Is(err, nats.ErrConnectionClosed):
+			return nil
+		case errors.Is(err, nats.ErrConnectionReconnecting):
+			// Drain hard-closed the connection. Buffered publishes are gone,
+			// but there was no reachable server to flush them to — a normal
+			// shutdown-during-outage, not a failure worth reporting upward.
+			slog.WarnContext(ctx, "nats drain skipped: connection reconnecting")
+			return nil
```
julianshen, line 46: _🏷️ delivery guarantees_ | _🔴 Blocking_ | _`distributed-system`_

**Draining while reconnecting silently discards buffered publishes.**

A publish made during a NATS outage can return nil after entering the reconnect buffer. If shutdown then calls Drain, nats.Conn.Drain hard-closes the reconnecting connection and returns ErrConnectionReconnecting, discarding that buffer. This branch converts the error to nil, so both the original publisher and shutdown caller observe success although the broker never received the message. Wait for reconnection within ctx if delivery must be preserved; at minimum, propagate the wrapped error so the failed drain is recorded.

<details>

> ‼️ **IMPORTANT**

```suggestion
		case errors.Is(err, nats.ErrConnectionReconnecting):
			// Drain hard-closed the connection, so reconnect-buffered
			// publishes could not be flushed and have been discarded.
			slog.WarnContext(ctx, "nats drain skipped: connection reconnecting")
			return fmt.Errorf("start nats drain: %w", err)
```

</details>

<details>

```
In `pkg/natsutil/drain.go` around line 41: A publish made during a NATS outage can return nil after entering the reconnect buffer. If shutdown then calls Drain, nats.Conn.Drain hard-closes the reconnecting connection and returns ErrConnectionReconnecting, discarding that buffer. This branch converts the error to nil, so both the original publisher and shutdown caller observe success although the broker never received the message. Wait for reconnection within ctx if delivery must be preserved; at minimum, propagate the wrapped error so the failed drain is recorded.

```

</details>

pkg/natsutil/drain.go
```diff
@@ -0,0 +1,73 @@
...
+		default:
+			return fmt.Errorf("start nats drain: %w", err)
+		}
+	}
+	if nc.IsClosed() {
+		return nil
+	}
+
+	select {
+	case <-ch:
```
julianshen, line 56: _🏷️ failure handling_ | _🔴 Blocking_ | _`distributed-system`_

**Drain timeouts still return success after dropping pending work.**

If a subscription handler remains blocked for the configured 10-second DrainTimeout, nats.go reports ErrDrainTimeout asynchronously and closes the connection. CLOSED then satisfies this case and Drain returns nil. Pending core messages can be lost and JetStream messages redelivered, while the shutdown hook reports a successful drain; the asynchronous error callback is not synchronized with this return and may not run before process exit. Plumb ErrDrainTimeout into the helper, or arrange for the caller-visible deadline to win before the library hard-close, and add a test using a short DrainTimeout plus a held handler that requires Drain to return an error.

<details>

```
In `pkg/natsutil/drain.go` around line 56: If a subscription handler remains blocked for the configured 10-second DrainTimeout, nats.go reports ErrDrainTimeout asynchronously and closes the connection. CLOSED then satisfies this case and Drain returns nil. Pending core messages can be lost and JetStream messages redelivered, while the shutdown hook reports a successful drain; the asynchronous error callback is not synchronized with this return and may not run before process exit. Plumb ErrDrainTimeout into the helper, or arrange for the caller-visible deadline to win before the library hard-close, and add a test using a short DrainTimeout plus a held handler that requires Drain to return an error.

```

</details>

pkg/natsutil/drain.go
```diff
@@ -0,0 +1,73 @@
...
+			slog.WarnContext(ctx, "nats drain skipped: connection reconnecting")
+			return nil
+		default:
+			return fmt.Errorf("start nats drain: %w", err)
+		}
+	}
+	if nc.IsClosed() {
+		return nil
+	}
+
```
julianshen, line 54: _🏷️ nats-lifecycle_ | _🔴 Blocking_ | _`nats`_

**A drain timeout is reported as a successful shutdown.**

With the configured 10-second DrainTimeout, a blocked subscription causes nats.go to set LastError to nats.ErrDrainTimeout and force the connection to CLOSED. This branch observes CLOSED and returns nil, so shutdown reports success even though pending handlers/messages were abandoned. Check LastError before treating CLOSED as successful, and add a test using a short DrainTimeout with a blocked handler.

<details>

> ‼️ **IMPORTANT**

```suggestion
case <-ch:
		if err := nc.LastError(); errors.Is(err, nats.ErrDrainTimeout) {
			return fmt.Errorf("nats drain incomplete: %w", err)
		}
		return nil
```

</details>

<details>

```
In `pkg/natsutil/drain.go` around line 53: With the configured 10-second DrainTimeout, a blocked subscription causes nats.go to set LastError to nats.ErrDrainTimeout and force the connection to CLOSED. This branch observes CLOSED and returns nil, so shutdown reports success even though pending handlers/messages were abandoned. Check LastError before treating CLOSED as successful, and add a test using a short DrainTimeout with a blocked handler.

```

</details>

pkg/natsutil/drain.go
```diff
@@ -0,0 +1,73 @@
...
+	ch := nc.StatusChanged(nats.CLOSED)
+	defer nc.RemoveStatusListener(ch)
+
+	if err := nc.Drain(); err != nil {
+		switch {
+		case errors.Is(err, nats.ErrConnectionClosed):
+			return nil
+		case errors.Is(err, nats.ErrConnectionReconnecting):
+			// Drain hard-closed the connection. Buffered publishes are gone,
+			// but there was no reachable server to flush them to — a normal
```
julianshen, line 43: _🏷️ error-handling_ | _🔴 Blocking_ | _`silent-failure`_

**Reconnect drain failures are converted into success.**

When NATS is reconnecting, nc.Drain() hard-closes the connection and returns ErrConnectionReconnecting, discarding buffered publishes. This branch explicitly acknowledges that data loss but returns nil, so every shutdown caller observes a successful drain. Return a wrapped error instead; callers may still choose how to report shutdown failure.

<details>

```
In `pkg/natsutil/drain.go` around line 43: When NATS is reconnecting, nc.Drain() hard-closes the connection and returns ErrConnectionReconnecting, discarding buffered publishes. This branch explicitly acknowledges that data loss but returns nil, so every shutdown caller observes a successful drain. Return a wrapped error instead; callers may still choose how to report shutdown failure.

```

</details>

pkg/natsutil/drain.go
```diff
@@ -0,0 +1,73 @@
...
+			// shutdown-during-outage, not a failure worth reporting upward.
+			slog.WarnContext(ctx, "nats drain skipped: connection reconnecting")
+			return nil
+		default:
+			return fmt.Errorf("start nats drain: %w", err)
+		}
+	}
+	if nc.IsClosed() {
+		return nil
+	}
```
julianshen, line 53: _🏷️ error-handling_ | _🔴 Blocking_ | _`silent-failure`_

**A CLOSED transition hides asynchronous drain failures.**

The background drain reaches CLOSED both after success and after its configured DrainTimeout or final FlushTimeout fails. Because the caller context is typically 25 seconds while the internal drain can fail and close after roughly 10–15 seconds, this branch returns nil before ctx expires even when pending messages or acknowledgements were discarded. Check nc.LastError() after every observed/fast-path CLOSED transition and return a wrapped drain or flush error when present.

<details>

```
In `pkg/natsutil/drain.go` around line 53: The background drain reaches CLOSED both after success and after its configured DrainTimeout or final FlushTimeout fails. Because the caller context is typically 25 seconds while the internal drain can fail and close after roughly 10–15 seconds, this branch returns nil before ctx expires even when pending messages or acknowledgements were discarded. Check nc.LastError() after every observed/fast-path CLOSED transition and return a wrapped drain or flush error when present.

```

</details>

pkg/natsutil/slowconsumer.go
```diff
@@ -0,0 +1,109 @@
...
+		metric.WithAttributes(
+			attribute.String("subject", sub.Subject),
+			attribute.String("queue", sub.Queue),
+		))
+}
+
+// logSlowConsumerAt picks the level: dropped messages on a core subscription are
+// unrecoverable loss and should page, so they log at ERROR. A slow consumer that
+// has not dropped anything yet is a warning.
+func logSlowConsumerAt(log *slog.Logger, dropped int, fields []any) {
```
julianshen, line 103: _🏷️ observability_ | _🟠 Warning_ | _`distributed-system`_

**A closed slow consumer downgrades confirmed message loss to WARN.**

A slow-consumer callback can be queued after messages are dropped, then execute after shutdown closes the subscription. Dropped() consequently fails, subDropped returns (0, false), and this level selection emits WARN even though ErrSlowConsumer itself confirms that core NATS messages were dropped. Alerts keyed to ERROR will miss this permanent loss. Thread the `ok` value into level selection and use ERROR when the count is positive or unavailable, while continuing to omit the unreadable count field.

<details>

```
In `pkg/natsutil/slowconsumer.go` around line 103: A slow-consumer callback can be queued after messages are dropped, then execute after shutdown closes the subscription. Dropped() consequently fails, subDropped returns (0, false), and this level selection emits WARN even though ErrSlowConsumer itself confirms that core NATS messages were dropped. Alerts keyed to ERROR will miss this permanent loss. Thread the `ok` value into level selection and use ERROR when the count is positive or unavailable, while continuing to omit the unreadable count field.

```

</details>

pkg/natsutil/drain.go
```diff
@@ -0,0 +1,73 @@
...
+		switch {
+		case errors.Is(err, nats.ErrConnectionClosed):
+			return nil
+		case errors.Is(err, nats.ErrConnectionReconnecting):
+			// Drain hard-closed the connection. Buffered publishes are gone,
+			// but there was no reachable server to flush them to — a normal
+			// shutdown-during-outage, not a failure worth reporting upward.
+			slog.WarnContext(ctx, "nats drain skipped: connection reconnecting")
+			return nil
+		default:
```
julianshen, line 47: _🏷️ message-loss_ | _🟠 Warning_ | _`nats`_

**A reconnecting connection discards buffered publishes but returns nil.**

Connect enables unlimited reconnects, during which Publish can queue data in the reconnect buffer. If shutdown occurs then, nats.Conn.Drain hard-closes the connection and returns ErrConnectionReconnecting, discarding that buffer. Returning nil makes the shutdown orchestrator record a successful drain despite unrecoverable message loss; propagate the sentinel after logging it.

<details>

> ‼️ **IMPORTANT**

```suggestion
return fmt.Errorf("nats drain while reconnecting: %w", err)
```

</details>

<details>

```
In `pkg/natsutil/drain.go` around line 47: Connect enables unlimited reconnects, during which Publish can queue data in the reconnect buffer. If shutdown occurs then, nats.Conn.Drain hard-closes the connection and returns ErrConnectionReconnecting, discarding that buffer. Returning nil makes the shutdown orchestrator record a successful drain despite unrecoverable message loss; propagate the sentinel after logging it.

```

</details>

pkg/natsutil/drain.go
julianshen, line 56: _🎯 concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Drain reports internal drain timeouts as success.**

When a subscription handler remains blocked past the configured 10-second DrainTimeout, nats.go records ErrDrainTimeout, closes the connection, and emits CLOSED. This branch consequently returns nil even though the drain timed out and buffered acknowledgements or replies may have been discarded. The shutdown hook therefore cannot distinguish a completed drain from forced closure. After every CLOSED observation, inspect nc.LastError() and return an error when it is nats.ErrDrainTimeout (and any drain-related flush error that nats.go records). Add a test using a short nats.DrainTimeout and a blocked handler.

<details>

```
In `pkg/natsutil/drain.go` around line 56: When a subscription handler remains blocked past the configured 10-second DrainTimeout, nats.go records ErrDrainTimeout, closes the connection, and emits CLOSED. This branch consequently returns nil even though the drain timed out and buffered acknowledgements or replies may have been discarded. The shutdown hook therefore cannot distinguish a completed drain from forced closure. After every CLOSED observation, inspect nc.LastError() and return an error when it is nats.ErrDrainTimeout (and any drain-related flush error that nats.go records). Add a test using a short nats.DrainTimeout and a blocked handler.

```

</details>

data-migration/oplog-connector/main.go
```diff
@@ -231,7 +231,11 @@ func (c *connector) Close() {
 	if err := c.awaitWatchers(wctx); err != nil {
 		slog.Warn("watcher drain incomplete", "error", err)
 	}
-	_ = c.nc.Drain()
+	dctx, dcancel := context.WithTimeout(context.Background(), 20*time.Second)
+	defer dcancel()
```
julianshen, line 235: _🏷️ shutdown_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Independent shutdown timeouts can exceed the termination grace period.**

Close first allows awaitWatchers up to 20 seconds and then creates a fresh 20-second drain budget. With a wedged watcher followed by a NATS drain needing its configured 10-second subscription timeout plus flush, shutdown can take about 35 seconds before Mongo cleanup, exceeding the documented 30-second pod grace period. SIGKILL can then interrupt the newly added drain and lose buffered acknowledgements. Bound the entire Close sequence to one shutdown deadline, or cap the watcher phase to roughly 10 seconds and reserve the remaining 15 seconds for NATS drain.

<details>

```
In `data-migration/oplog-connector/main.go` around line 235: Close first allows awaitWatchers up to 20 seconds and then creates a fresh 20-second drain budget. With a wedged watcher followed by a NATS drain needing its configured 10-second subscription timeout plus flush, shutdown can take about 35 seconds before Mongo cleanup, exceeding the documented 30-second pod grace period. SIGKILL can then interrupt the newly added drain and lose buffered acknowledgements. Bound the entire Close sequence to one shutdown deadline, or cap the watcher phase to roughly 10 seconds and reserve the remaining 15 seconds for NATS drain.

```

</details>

pkg/natsutil/drain_test.go
```diff
@@ -0,0 +1,177 @@
...
+	require.NoError(t, conn.NatsConn().Flush())
+
+	pub := testConn(t, url)
+	require.NoError(t, pub.NatsConn().Publish("drain.wedged", []byte("x")))
+	require.NoError(t, pub.NatsConn().Flush())
+
+	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+	defer cancel()
+
+	err = natsutil.Drain(ctx, conn)
```
julianshen, line 128: _🧪 tests_ | _🟠 Warning_ | _`tgd-review`_

**The deadline test can drain before its handler is actually wedged.**

A publisher Flush does not guarantee the subscriber callback has started, as the preceding test correctly explains. If Drain observes an empty queue first, this test can receive nil instead of exercising context expiry. Synchronize on handler entry before starting the 100ms deadline.

<details>

> ‼️ **IMPORTANT**

```suggestion
	entered := make(chan struct{})
	release := make(chan struct{})
	t.Cleanup(func() { close(release) })
	_, err := conn.NatsConn().Subscribe("drain.wedged", func(*nats.Msg) {
		close(entered)
		<-release
	})
	require.NoError(t, err)
	require.NoError(t, conn.NatsConn().Flush())

	pub := testConn(t, url)
	require.NoError(t, pub.NatsConn().Publish("drain.wedged", []byte("x")))
	require.NoError(t, pub.NatsConn().Flush())
	select {
	case <-entered:
	case <-time.After(5 * time.Second):
		t.Fatal("handler did not start")
	}

	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
	defer cancel()

	err = natsutil.Drain(ctx, conn)
```

</details>

<details>

```
In `pkg/natsutil/drain_test.go` around line 115: A publisher Flush does not guarantee the subscriber callback has started, as the preceding test correctly explains. If Drain observes an empty queue first, this test can receive nil instead of exercising context expiry. Synchronize on handler entry before starting the 100ms deadline.

```

</details>

pkg/natsutil/connect_test.go
```diff
@@ -62,3 +63,13 @@ func TestConnect_PresentCredsFilePassesPrecheck(t *testing.T) {
 		noop.NewTracerProvider(), propagation.TraceContext{}, false)
 	require.False(t, errors.Is(err, os.ErrNotExist), "precheck should pass when file exists, got: %v", err)
 }
```
julianshen, line 65: _🧪 tests_ | _🟠 Warning_ | _`tgd-review`_

**Slow-consumer wiring and metrics are not asserted by tests.**

The slow-consumer tests call logSlowConsumer directly, while the Connect test only checks DrainTimeout. Removing the new ErrorHandler branch or the counter increment would therefore leave the suite green even though the advertised visibility feature stopped working. Add an integration test that exercises the callback installed by Connect (or induces a real slow-consumer event) and observes the counter through an OpenTelemetry manual reader.

<details>

```
In `pkg/natsutil/connect_test.go` around line 65: The slow-consumer tests call logSlowConsumer directly, while the Connect test only checks DrainTimeout. Removing the new ErrorHandler branch or the counter increment would therefore leave the suite green even though the advertised visibility feature stopped working. Add an integration test that exercises the callback installed by Connect (or induces a real slow-consumer event) and observes the counter through an OpenTelemetry manual reader.

```

</details>

pkg/natsutil/drain.go
julianshen, line 46: _🏷️ delivery guarantees_ | _🔴 Blocking_ | _`distributed-system`_

**Reconnect hard-close is reported as a successful drain.**

With MaxReconnects(-1), a core NATS Publish can return successfully while its bytes remain in the reconnect buffer. If SIGTERM arrives during that outage, nc.Drain() hard-closes the connection and discards that buffer, returning ErrConnectionReconnecting. This branch then returns nil, so shutdown reports success even though accepted publishes or buffered acknowledgements were lost; the previous direct Drain hooks exposed this error. At minimum, propagate the error. Preserving delivery during outages additionally requires a durable outbox or a bounded strategy that waits for reconnection before shutdown.

<details>

> ‼️ **IMPORTANT**

```suggestion
		case errors.Is(err, nats.ErrConnectionReconnecting):
			// Drain hard-closed the connection, so buffered publishes could not
			// be flushed; report the incomplete drain to the shutdown caller.
			slog.WarnContext(ctx, "nats drain failed: connection reconnecting")
			return fmt.Errorf("start nats drain: %w", err)
```

</details>

<details>

```
In `pkg/natsutil/drain.go` around line 41: With MaxReconnects(-1), a core NATS Publish can return successfully while its bytes remain in the reconnect buffer. If SIGTERM arrives during that outage, nc.Drain() hard-closes the connection and discards that buffer, returning ErrConnectionReconnecting. This branch then returns nil, so shutdown reports success even though accepted publishes or buffered acknowledgements were lost; the previous direct Drain hooks exposed this error. At minimum, propagate the error. Preserving delivery during outages additionally requires a durable outbox or a bounded strategy that waits for reconnection before shutdown.

```

</details>

pkg/natsutil/drain.go
julianshen, line 56: _🏷️ failure handling_ | _🔴 Blocking_ | _`distributed-system`_

**A CLOSED event can hide an internal drain timeout.**

CLOSED only proves that the asynchronous drain terminated, not that it completed successfully. A handler can remain active beyond the configured 10-second DrainTimeout; nats.go then reports ErrDrainTimeout and closes the connection. This branch receives CLOSED and returns nil, allowing shutdown to exit while the handler's acknowledgement is lost, potentially redelivering a message after its external side effect committed. A final publish flush failure can follow the same path. Inspect and propagate the terminal drain error after every CLOSED/IsClosed path, or explicitly capture ErrDrainTimeout and flush errors from the async handler. Add a test with a short nats.DrainTimeout, a wedged handler, and a longer caller context that asserts Drain returns ErrDrainTimeout.

<details>

```
In `pkg/natsutil/drain.go` around line 56: CLOSED only proves that the asynchronous drain terminated, not that it completed successfully. A handler can remain active beyond the configured 10-second DrainTimeout; nats.go then reports ErrDrainTimeout and closes the connection. This branch receives CLOSED and returns nil, allowing shutdown to exit while the handler's acknowledgement is lost, potentially redelivering a message after its external side effect committed. A final publish flush failure can follow the same path. Inspect and propagate the terminal drain error after every CLOSED/IsClosed path, or explicitly capture ErrDrainTimeout and flush errors from the async handler. Add a test with a short nats.DrainTimeout, a wedged handler, and a longer caller context that asserts Drain returns ErrDrainTimeout.

```

</details>

pkg/natsutil/drain.go
julianshen, line 46: _🏷️ reliability_ | _🔴 Blocking_ | _`nats`_

**Reconnect-time data loss is reported as a successful drain.**

With MaxReconnects(-1), an outage can leave publishes in the reconnect buffer when shutdown begins. nats.Conn.Drain then hard-closes the connection and returns ErrConnectionReconnecting, discarding that buffer; this branch explicitly acknowledges the loss but returns nil, so every shutdown hook reports success. Propagate the sentinel so shutdown records that graceful delivery was not achieved.

<details>

> ‼️ **IMPORTANT**

```suggestion
		case errors.Is(err, nats.ErrConnectionReconnecting):
			return fmt.Errorf("start nats drain: %w", err)
```

</details>

<details>

```
In `pkg/natsutil/drain.go` around line 41: With MaxReconnects(-1), an outage can leave publishes in the reconnect buffer when shutdown begins. nats.Conn.Drain then hard-closes the connection and returns ErrConnectionReconnecting, discarding that buffer; this branch explicitly acknowledges the loss but returns nil, so every shutdown hook reports success. Propagate the sentinel so shutdown records that graceful delivery was not achieved.

```

</details>

pkg/natsutil/drain.go
julianshen, line 56: _🏷️ reliability_ | _🔴 Blocking_ | _`nats`_

**A forced close is indistinguishable from a completed drain.**

CLOSED is reached after both a clean drain and the configured 10-second DrainTimeout. A handler blocked beyond that timeout causes nats.go to emit ErrDrainTimeout and forcibly close the connection; this select receives CLOSED and returns nil before the shared 25-second context expires. Core messages can be lost and JetStream work can be interrupted or redelivered while shutdown reports success. Coordinate the async ErrDrainTimeout with this waiter, or make the helper own a timeout that can return an error before the client's forced close. Add a test using a short DrainTimeout and a blocked handler that asserts the failure is propagated.

<details>

```
In `pkg/natsutil/drain.go` around line 56: CLOSED is reached after both a clean drain and the configured 10-second DrainTimeout. A handler blocked beyond that timeout causes nats.go to emit ErrDrainTimeout and forcibly close the connection; this select receives CLOSED and returns nil before the shared 25-second context expires. Core messages can be lost and JetStream work can be interrupted or redelivered while shutdown reports success. Coordinate the async ErrDrainTimeout with this waiter, or make the helper own a timeout that can return an error before the client's forced close. Add a test using a short DrainTimeout and a blocked handler that asserts the failure is propagated.

```

</details>

pkg/natsutil/drain.go
```diff
@@ -0,0 +1,73 @@
...
+		select {
+		case <-ch:
+			return nil
+		default:
+		}
+		if nc.IsClosed() {
+			return nil
+		}
+		return fmt.Errorf("nats drain incomplete: %w", ctx.Err())
+	}
```
julianshen, line 72: _🏷️ timeout-handling_ | _🔴 Blocking_ | _`performance-and-concurrency`_

**An internal drain timeout is reported as success.**

When a subscription handler remains blocked past `DrainTimeout`, nats.go records `nats.ErrDrainTimeout` and closes the connection. Every CLOSED path here returns nil, so callers proceed as though handlers and buffered acknowledgements drained successfully. Under this interleaving the handler can continue while dependencies are disconnected, and its ack/reply is lost. Check `nc.LastError()` after observing CLOSED and propagate `ErrDrainTimeout`; add a regression test using a short `nats.DrainTimeout` and a blocked handler.

<details>

> ‼️ **IMPORTANT**

```suggestion
	closedResult := func() error {
		if err := nc.LastError(); errors.Is(err, nats.ErrDrainTimeout) {
			return fmt.Errorf("nats drain incomplete: %w", err)
		}
		return nil
	}
	if nc.IsClosed() {
		return closedResult()
	}

	select {
	case <-ch:
		return closedResult()
	case <-ctx.Done():
		// select picks pseudo-randomly when both cases are ready at once, so a
		// drain that finished in the same instant ctx expired can still land
		// here. Re-check before reporting failure so a completed drain is
		// never misreported as incomplete.
		select {
		case <-ch:
			return closedResult()
		default:
		}
		if nc.IsClosed() {
			return closedResult()
		}
		return fmt.Errorf("nats drain incomplete: %w", ctx.Err())
	}
```

</details>

<details>

```
In `pkg/natsutil/drain.go` around line 50: When a subscription handler remains blocked past `DrainTimeout`, nats.go records `nats.ErrDrainTimeout` and closes the connection. Every CLOSED path here returns nil, so callers proceed as though handlers and buffered acknowledgements drained successfully. Under this interleaving the handler can continue while dependencies are disconnected, and its ack/reply is lost. Check `nc.LastError()` after observing CLOSED and propagate `ErrDrainTimeout`; add a regression test using a short `nats.DrainTimeout` and a blocked handler.

```

</details>

pkg/natsutil/drain.go
julianshen, line 56: _🏷️ failure-handling_ | _🔴 Blocking_ | _`silent-failure`_

**An internal drain timeout is mistaken for successful closure.**

The NATS drain goroutine records ErrDrainTimeout and closes the connection when subscriptions exceed DrainTimeout; a final FlushTimeout failure also closes the connection after recording the error. Because this code treats every CLOSED event as success, the 10-second internal timeout can fire before the 25-second shutdown context and Drain returns nil despite an incomplete drain. Services then exit with handlers unfinished or buffered acknowledgements and replies unflushed. Inspect nc.LastError() after every CLOSED observation and return the drain/flush error. Add a test with a short DrainTimeout and a longer caller deadline; the current context-expiry test does not exercise this path.

<details>

```
In `pkg/natsutil/drain.go` around line 56: The NATS drain goroutine records ErrDrainTimeout and closes the connection when subscriptions exceed DrainTimeout; a final FlushTimeout failure also closes the connection after recording the error. Because this code treats every CLOSED event as success, the 10-second internal timeout can fire before the 25-second shutdown context and Drain returns nil despite an incomplete drain. Services then exit with handlers unfinished or buffered acknowledgements and replies unflushed. Inspect nc.LastError() after every CLOSED observation and return the drain/flush error. Add a test with a short DrainTimeout and a longer caller deadline; the current context-expiry test does not exercise this path.

```

</details>

pkg/natsutil/slowconsumer.go
```diff
@@ -0,0 +1,109 @@
...
+	if msgs, nbytes, err := sub.PendingLimits(); err == nil {
+		fields = append(fields, "limit_msgs", msgs, "limit_bytes", nbytes)
+	}
+	return fields
+}
+
+// logSlowConsumer reports a slow-consumer episode and counts it.
+func logSlowConsumer(log *slog.Logger, sub *nats.Subscription) {
+	dropped, ok := subDropped(sub)
+	logSlowConsumerAt(log, dropped, slowConsumerFields(sub, dropped, ok))
```
julianshen, line 89: _🏷️ observability_ | _🟠 Warning_ | _`nats`_

**Closed subscriptions downgrade confirmed message loss to WARN.**

ErrSlowConsumer is raised after the client drops a message. Because the error callback is asynchronous, the subscription can close before this callback reads Dropped(); subDropped then returns ok=false and dropped=0, and logSlowConsumerAt emits WARN rather than ERROR. This can suppress paging for actual loss during shutdown. Preserve the unknown state and log ERROR whenever Dropped cannot be read on an ErrSlowConsumer path; update the closed/nil-subscription tests accordingly.

<details>

```
In `pkg/natsutil/slowconsumer.go` around line 89: ErrSlowConsumer is raised after the client drops a message. Because the error callback is asynchronous, the subscription can close before this callback reads Dropped(); subDropped then returns ok=false and dropped=0, and logSlowConsumerAt emits WARN rather than ERROR. This can suppress paging for actual loss during shutdown. Preserve the unknown state and log ERROR whenever Dropped cannot be read on an ErrSlowConsumer path; update the closed/nil-subscription tests accordingly.

```

</details>

media-service/main.go
```diff
@@ -102,20 +102,30 @@ func run() error {
...
+	shutdownDone := make(chan struct{})
+	go func() {
+		defer close(shutdownDone)
+		shutdown.Wait(ctx, 25*time.Second,
+			func(ctx context.Context) error { return router.Shutdown(ctx) },
+			func(ctx context.Context) error { return natsutil.Drain(ctx, nc) },
+			func(ctx context.Context) error {
+				slog.Info("shutting down media-service")
+				return srv.Shutdown(ctx)
+			},
```
julianshen, line 120: _🏷️ shutdown-ordering_ | _🟠 Warning_ | _`performance-and-concurrency`_

**The HTTP listener stays open while NATS drains.**

On SIGTERM, the service first stops its NATS router and can now spend up to roughly 15 seconds draining the connection before `srv.Shutdown` closes the HTTP listener. Existing keep-alive clients or requests arriving during endpoint removal can therefore reach handlers after their NATS dependency is draining or closed, causing avoidable failures under production traffic. Close the HTTP listener and wait for in-flight HTTP handlers before tearing down NATS.

<details>

> ‼️ **IMPORTANT**

```suggestion
			func(ctx context.Context) error {
				slog.Info("shutting down media-service")
				return srv.Shutdown(ctx)
			},
			func(ctx context.Context) error { return router.Shutdown(ctx) },
			func(ctx context.Context) error { return natsutil.Drain(ctx, nc) },
```

</details>

<details>

```
In `media-service/main.go` around line 115: On SIGTERM, the service first stops its NATS router and can now spend up to roughly 15 seconds draining the connection before `srv.Shutdown` closes the HTTP listener. Existing keep-alive clients or requests arriving during endpoint removal can therefore reach handlers after their NATS dependency is draining or closed, causing avoidable failures under production traffic. Close the HTTP listener and wait for in-flight HTTP handlers before tearing down NATS.

```

</details>

data-migration/oplog-connector/main.go
```diff
@@ -231,7 +231,11 @@ func (c *connector) Close() {
 	if err := c.awaitWatchers(wctx); err != nil {
 		slog.Warn("watcher drain incomplete", "error", err)
 	}
-	_ = c.nc.Drain()
+	dctx, dcancel := context.WithTimeout(context.Background(), 20*time.Second)
```
julianshen, line 234: _🏷️ shutdown-timeout_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Sequential shutdown deadlines can exceed the pod grace period.**

`Close` first gives `awaitWatchers` an independent 20-second deadline and then gives NATS drain another 20-second deadline. If watchers stall near their timeout and NATS has an in-flight handler or flush, this path can run for about 35 seconds before Mongo disconnection, exceeding the documented 30-second termination grace period and ending in SIGKILL during the drain. Use one overall shutdown context, bounded below the process grace period, for both waits so the second phase receives only the remaining budget.

<details>

```
In `data-migration/oplog-connector/main.go` around line 234: `Close` first gives `awaitWatchers` an independent 20-second deadline and then gives NATS drain another 20-second deadline. If watchers stall near their timeout and NATS has an in-flight handler or flush, this path can run for about 35 seconds before Mongo disconnection, exceeding the documented 30-second termination grace period and ending in SIGKILL during the drain. Use one overall shutdown context, bounded below the process grace period, for both waits so the second phase receives only the remaining budget.

```

</details>

pkg/natsutil/drain.go
julianshen, line 46: _🏷️ failure-handling_ | _🟠 Warning_ | _`silent-failure`_

**Reconnect-time drain loss is returned as success.**

NATS hard-closes a reconnecting connection and returns ErrConnectionReconnecting. This branch explicitly acknowledges that buffered publishes are discarded, but then returns nil, so every shutdown caller observes a successful drain despite possible lost acknowledgements, replies, or publishes. Preserve the close behavior but propagate a wrapped error so shutdown handling can surface the incomplete durable work.

<details>

> ‼️ **IMPORTANT**

```suggestion
		case errors.Is(err, nats.ErrConnectionReconnecting):
			slog.WarnContext(ctx, "nats drain failed: connection reconnecting", "error", err)
			return fmt.Errorf("start nats drain: %w", err)
```

</details>

<details>

```
In `pkg/natsutil/drain.go` around line 41: NATS hard-closes a reconnecting connection and returns ErrConnectionReconnecting. This branch explicitly acknowledges that buffered publishes are discarded, but then returns nil, so every shutdown caller observes a successful drain despite possible lost acknowledgements, replies, or publishes. Preserve the close behavior but propagate a wrapped error so shutdown handling can surface the incomplete durable work.

```

</details>

pkg/natsutil/slowconsumer.go
julianshen, line 89: _🏷️ failure-handling_ | _🟠 Warning_ | _`silent-failure`_

**An unreadable drop count downgrades known message loss.**

This function is reached specifically for nats.ErrSlowConsumer, whose event indicates messages were dropped. If the asynchronous callback runs after the subscription closes, subDropped returns (0, false), but that zero is passed to logSlowConsumerAt and logged only at WARN without the message-loss wording. The code already identifies subscription closure during shutdown as routine, so this reachable race suppresses the ERROR alert for real data loss. Treat ok == false as an unknown count, not zero: log the slow-consumer event at ERROR while omitting the unavailable count, and update the nil/closed-path tests accordingly.

<details>

```
In `pkg/natsutil/slowconsumer.go` around line 89: This function is reached specifically for nats.ErrSlowConsumer, whose event indicates messages were dropped. If the asynchronous callback runs after the subscription closes, subDropped returns (0, false), but that zero is passed to logSlowConsumerAt and logged only at WARN without the message-loss wording. The code already identifies subscription closure during shutdown as routine, so this reachable race suppresses the ERROR alert for real data loss. Treat ok == false as an unknown count, not zero: log the slow-consumer event at ERROR while omitting the unavailable count, and update the nil/closed-path tests accordingly.

```

</details>

pkg/natsutil/slowconsumer.go
```diff
@@ -0,0 +1,109 @@
...
+	if sub == nil {
+		return
```
julianshen, line 91: _🏷️ observability_ | _🟠 Warning_ | _`tgd-review`_

**Closed slow-consumer callbacks downgrade confirmed loss to WARN.**

`nats.ErrSlowConsumer` is emitted only after a subscription overflows and drops a message, and its error callback is asynchronous. If the subscription closes before that callback executes, `Dropped()` fails, `subDropped` returns `(0, false)`, and this call logs at WARN despite confirmed message loss. Treat an unreadable count on a non-nil subscription as ERROR while still omitting the unavailable count, and add a closed-subscription level test.

<details>

```
In `pkg/natsutil/slowconsumer.go` around line 91: `nats.ErrSlowConsumer` is emitted only after a subscription overflows and drops a message, and its error callback is asynchronous. If the subscription closes before that callback executes, `Dropped()` fails, `subDropped` returns `(0, false)`, and this call logs at WARN despite confirmed message loss. Treat an unreadable count on a non-nil subscription as ERROR while still omitting the unavailable count, and add a closed-subscription level test.

```

</details>

data-migration/oplog-connector/main.go
```diff
@@ -231,7 +231,11 @@ func (c *connector) Close() {
...
+	defer dcancel()
+	if err := natsutil.Drain(dctx, c.nc); err != nil {
+		slog.Warn("nats drain incomplete", "error", err)
+	}
```
julianshen, line 238: _🏷️ shutdown_ | _🟠 Warning_ | _`tgd-review`_

**Sequential timeout budgets can exceed the termination grace period.**

`Close` first gives `awaitWatchers` 20 seconds and then creates a fresh 20-second drain context. This permits 40 seconds overall—or about 35 seconds with the default 10-second drain plus flush—despite the documented 30-second Kubernetes termination grace period. Use one outer shutdown budget, optionally with a 20-second child budget for watcher shutdown, and pass the remaining outer context to the NATS drain.

<details>

```
In `data-migration/oplog-connector/main.go` around line 238: `Close` first gives `awaitWatchers` 20 seconds and then creates a fresh 20-second drain context. This permits 40 seconds overall—or about 35 seconds with the default 10-second drain plus flush—despite the documented 30-second Kubernetes termination grace period. Use one outer shutdown budget, optionally with a 20-second child budget for watcher shutdown, and pass the remaining outer context to the NATS drain.

```

</details>

media-service/main.go
mliu33: We should fix it according to this suggestion. Thanks

mliu33 (approved):
Outstanding work, thanks a lot! Good catch!!

PR #233, opened 2026-08-11
feat(permission): foundations — model, reason catalog, subject builders, design docs

nalydadad:
Foundations for the user permission whitelist (an append-only ledger recording whether a user may view images from outside the corporate network). First PR of a 3-PR stack (foundations → backend → frontend); nothing references these symbols yet, so this merges independently.

- `pkg/model/permission.go` — `PermissionKey` + `external.image.view`, `PermissionGrant` (ledger row; pointer/omitempty window fields per the optional-fields convention), limits (`MaxSubjects=200`, `MaxReasonRunes=1000`), and `EvaluateGrant`: the single latest-wins, fail-closed evaluation function every entry point shares. Half-open `[from, until)` window semantics; instant-vs-instant comparison (no timezone logic at evaluation time).
- `pkg/errcode/codes_permission.go` — the 8 permission reason codes, registered in the catalog test.
- `pkg/subject` — `UserPermissionGet` builders for the `permission.get` RPC subject.
- `docs/superpowers/specs/2026-08-10-user-permission-whitelist-design.md` — the approved design (data model, validation rules, fixed UTC+8 date-interpretation rule, architecture rationale). The follow-up PRs implement it; reviewing the spec here means the later diffs can be checked against an already-merged reference.
- `docs/superpowers/plans/2026-08-11-user-permission-whitelist.md` — the implementation plan the work was executed from (kept for traceability).

Tests: `EvaluateGrant` at 100% coverage (boundary cases per spec §10); model round-trips; subject builder tables. Full-repo suite green on this branch (proves existing services are unaffected by the additive pkg changes).

### DIFF
--- a/docs/superpowers/plans/2026-08-11-user-permission-whitelist.md
+++ b/docs/superpowers/plans/2026-08-11-user-permission-whitelist.md
@@ -1,0 +1,6950 @@
+# User Permission Whitelist Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** An append-only permission ledger recording whether a user may view images from outside the corporate network — admin-written over HTTP (API + console), self-queried over NATS, directly readable by Metabase.
+
+**Architecture:** No new service. admin-service (Gin, `requireAdmin`) owns writes into a new `permission_grants` collection (transactional batch insert, latest-wins evaluation, half-open `[from, until)` validity windows anchored to a fixed UTC+8 rule) plus slim per-subject `admin_audit` entries; user-service answers `permission.get` over NATS request/reply where the subject's `{account}` token IS the authenticated identity; admin-frontend gains a Permissions console view. Evaluation is one pure function (`model.EvaluateGrant`) shared by every current and future entry point.
+
+**Tech Stack:** Go 1.25, Gin, NATS (natsrouter), MongoDB driver v2 (no ORM), go.uber.org/mock, testify, testcontainers (via pkg/testutil), React 19 + Vitest (admin-frontend).
+
+**Spec:** `docs/superpowers/specs/2026-08-10-user-permission-whitelist-design.md` — the design authority. When a step here seems to contradict it, the spec wins; stop and flag.
+
+## Global Constraints
+
+- All commands via the root Makefile: `make test SERVICE=<path>` (works for any package path, e.g. `SERVICE=pkg/model`), `make test-integration SERVICE=<name>`, `make generate SERVICE=<name>`, `make lint`, `make fmt`, `make sast`. Never raw `go` commands — sole exception: the coverage spot-check in Task 14 uses CLAUDE.md's documented `go test -coverprofile` procedure, and admin-frontend tests run `cd admin-frontend && npm test` (no make target exists).
+- Strict TDD red-green-refactor: failing test first, run and confirm FAIL, minimal implementation, confirm PASS, lint, commit. Never skip the red phase.
+- Coverage ≥80% per package; 90%+ on `EvaluateGrant` and handler validation.
+- errcode Tier-1 in handlers (`errcode.BadRequest(msg, errcode.WithReason(...))` etc.); adapters `errhttp.Write` (admin-service) / natsrouter automatic (user-service); infra failures return `fmt.Errorf("...: %w", err)`. Never log AND return the same error. Every new reason registered in `allReasons` (`pkg/errcode/codes_test.go`).
+- Mongo: explicit projection on every find; camelCase bson tags; `bson:"_id"` for IDs; no `$lookup`. `permission_grants` indexes are created by admin-service ONLY (user-service must not call EnsureIndexes for it).
+- NATS subjects only via `pkg/subject` builders — never raw `fmt.Sprintf`.
+- Transactions require a replica set: integration tests for transactional store methods use `testutil.MongoDBReplicaSet(t, prefix)`; plain `testutil.MongoDB(t, prefix)` is a STANDALONE Mongo where transactions fail. Read-only repo tests may use the standalone helper.
+- Unit tests: same package, table-driven, named subtests, testify, mockgen mocks (regenerate via `make generate`, never hand-edit). Integration tests: `//go:build integration`, `pkg/testutil` containers, `TestMain` via `testutil.RunTests`/`RunTestsWithPrewarm`.
+- Dates: the fixed interpretation rule is `time.FixedZone("UTC+8", 8*60*60)` — never `time.Local`, never `time.Parse` without a location (its UTC default is exactly the bug this design avoids). `+1 day` via civil arithmetic `time.Date(y, m, d+1, 0, 0, 0, 0, tzTaipei)`. No tzdata anywhere.
+- Commit per task with the exact message given in the task's final step; conventional-commit style; NO AI-provenance trailers.
+- One PR at the end; `docs/client-api.md` + `docs/client-api/request-reply.md` land in the same PR (Task 13). Delete any `docs/reviews/` files before creating the PR.
+
+## Task overview (dependency order)
+
+| # | Task | Package | Commit |
+|---|---|---|---|
+| 1 | PermissionGrant model + EvaluateGrant | pkg/model | `feat(model): permission grant ledger type + latest-wins evaluation` |
+| 2 | Permission reason catalog | pkg/errcode | `feat(errcode): permission reason catalog` |
+| 3 | permission.get subject builders | pkg/subject | `feat(subject): user permission.get subject builders` |
+| 4 | Store: grants + audit batch | admin-service | `feat(admin-service): permission grants store — transactional batch insert, ledger reads, audit batch` |
+| 5 | POST /v1/admin/permissions + body limit | admin-service | `feat(admin-service): POST /v1/admin/permissions with body limit` |
+| 6 | GET /v1/admin/permissions | admin-service | `feat(admin-service): GET /v1/admin/permissions ledger + current decision` |
+| 7 | PermissionRepo (primary reads) | user-service | `feat(user-service): permission grants repository (primary reads)` |
+| 8 | permission.get NATS handler | user-service | `feat(user-service): permission.get NATS handler` |
+| 9 | Permissions api client + reason copy | admin-frontend | `feat(admin-frontend): permissions api client` |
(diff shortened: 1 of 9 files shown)


julianshen:
**Actionable comments posted: 16** — 🔴 11 blocking · 🟠 5 warning

### 💬 Additional comments (16)


**`docs/superpowers/plans/2026-08-11-user-permission-whitelist.md`**

_🏷️ idempotency_ | _🔴 Blocking_ | _`distributed-system`_

**Transactions do not make retried writes idempotent.**

The plan generates fresh grant IDs and timestamps for every POST, while the transaction protects only one attempt. Concrete failure: the grant transaction commits, its 201 response is lost, another admin revokes the permission, and the original caller retries; the retry inserts a new grant that sorts after the revoke and unintentionally restores access. The plan's claim that a transaction prevents duplicates on resend is therefore incorrect. Require a stable caller-supplied operation/idempotency key, persist it atomically with the batch under a unique index, validate repeated keys against the original payload, and return the original result without appending rows. Add a test covering a committed-but-unacknowledged request followed by a revoke and retry.

---

**`docs/superpowers/specs/2026-08-10-user-permission-whitelist-design.md`**

_🏷️ ordering_ | _🔴 Blocking_ | _`distributed-system`_

**Client-clock ordering can make a later revoke lose.**

Latest-wins sorts by application-generated recordedAt and then an application-generated UUIDv7 _id. Concrete failure: replica A's clock is fast when it records a grant, then replica B's clock is slow when it records a later revoke; both readers permanently select the earlier grant as newest, so the revoke never takes effect. The _id tie-break does not repair this because UUIDv7 is also clock-derived and only resolves equal recordedAt values. NTP and a confirming GET cannot guarantee correctness. Allocate a monotonically increasing per-(site, permission, subject) sequence on the Mongo primary in the same transaction, store it on every ledger row, and use it as the authoritative sort key in all readers.

---

**`docs/superpowers/plans/2026-08-11-user-permission-whitelist.md`**

_🏷️ data-integrity_ | _🔴 Blocking_ | _`mongodb`_

**An omitted granted field is stored as a revocation.**

The planned request model uses `Granted bool`, so missing or JSON `null` values decode as `false`. A request with no `granted` or window fields therefore follows the revoke path and appends a permanent latest-wins revocation to `permission_grants`. The approved spec explicitly requires `granted` to be present. Use `*bool` or a presence-tracking JSON type, reject nil before any MongoDB lookup/write, and add tests for omitted and null values.

---

**`docs/superpowers/plans/2026-08-11-user-permission-whitelist.md`**

_🏷️ data-integrity_ | _🔴 Blocking_ | _`mongodb`_

**Applicant and approver accounts are never validated.**

The POST implementation calls `FindAccountStates(ctx, h.cfg.SiteID, subjects)`, so only subject accounts are queried. A valid subject combined with a nonexistent applicant or approver passes validation and permanently stores dangling account references in the append-only ledger. This contradicts the design and planned API documentation, which require all three account roles to exist. Query the deduplicated union of subjects, applicant, and approver in one `$in` lookup; require existence for all entries while applying `IsActive()` only to subjects.

---

**`docs/superpowers/plans/2026-08-11-user-permission-whitelist.md`**

_🎯 query-correctness_ | _🔴 Blocking_ | _`mongodb`_

**The planned GET drops two required MongoDB query shapes.**

The approved design allows `subjectAccount` and `permission` as independent optional filters, including site-wide and permission-only ledger browsing. Task 6 instead rejects an empty `subjectAccount`, while `ListPermissionGrants` always adds `subjectAccount` to its filter. Consequently the API cannot execute the site-wide query intended for the `{siteId:1, recordedAt:-1}` index or the permission-only audit query. Make both store filters conditional, remove the mandatory-subject validation, and return `currentlyGranted` only when both filters are supplied. Add tests for all four filter combinations.

---

**`docs/superpowers/plans/2026-08-11-user-permission-whitelist.md`**

_🏷️ reliability_ | _🔴 Blocking_ | _`silent-failure`_

**The Mongo healthcheck reports healthy before a primary is elected.**

Task 12's healthcheck exits successfully when `rs.initiate()` merely returns, and later checks only `rs.status().ok`. Neither condition proves the node is writable primary, so Compose can start dependent services before election completes; their index creation or transactions can then fail during startup. Make the healthcheck explicitly exit nonzero until `db.hello().isWritablePrimary` is true, while retaining the one-time initialization path.

---

**`docs/superpowers/plans/2026-08-11-user-permission-whitelist.md`**

_🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**Missing granted is silently treated as a revocation.**

Task 5 models `Granted` as a plain `bool` and never validates its presence. Go's JSON decoder maps both an omitted field and `null` to `false`, so either payload can pass as a revoke when no window is supplied. This contradicts design §4.4 rule 7, which explicitly requires `granted` and rejects omission or null. Represent it as `*bool`, reject nil with `missing_permission_fields`, and only then branch on the dereferenced value; add tests for omitted and null values.

---

**`docs/superpowers/plans/2026-08-11-user-permission-whitelist.md`**

_🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**Applicant and approver accounts are never validated.**

Task 5 calls `FindAccountStates(ctx, h.cfg.SiteID, subjects)`, so only subject accounts are queried. The approved design §4.4 requires all subject, applicant, and approver accounts to exist, while only subjects must be active; Task 13 also documents 404 for an unknown applicant or approver. Query the deduplicated union of all three account roles, reject any missing account, and apply the inactive check only to subjects. Add tests for unknown applicant and approver accounts.

---

**`docs/superpowers/plans/2026-08-11-user-permission-whitelist.md`**

_🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**The window check accepts a start date after the expiry date.**

`parseWindow` first advances the submitted expiry date by one day and then rejects only `from.After(until)`. Thus `effectiveFrom=2027-01-01` with `expiresAt=2026-12-31` produces equal instants and is accepted, despite the submitted start date being after the inclusive expiry date. Compare the civil input dates before shifting, or reject `!from.Before(until)` after constructing the exclusive end. Add this adjacent-day boundary case to `TestParseWindow`.

---

**`docs/superpowers/plans/2026-08-11-user-permission-whitelist.md`**

_🏷️ intent-match_ | _🔴 Blocking_ | _`tgd-review`_

**The GET plan removes two filter combinations required by the spec.**

Design §4.6 states that `subjectAccount` and `permission` are independently optional and defines all four combinations, with `currentlyGranted` present only when both are supplied. Task 4's store API always requires a subject, Task 6 rejects an empty `subjectAccount`, and Task 13 documents it as required while computing a decision whenever only `permission` is present. This eliminates site-wide and permission-wide queries and can produce a decision across multiple subjects. Update the repository filter construction, handler validation, response condition, tests, and documentation to support the four specified combinations.

---

**`docs/superpowers/plans/2026-08-11-user-permission-whitelist.md`**

_🏷️ intent-match_ | _🔴 Blocking_ | _`tgd-review`_

**The plan makes an optional reason mandatory.**

The approved design says revoke reasons are optional and §4.4 rule 6 accepts an omitted or empty reason, rejecting only values over `MaxReasonRunes`. Task 5 rejects `req.Reason == ""`, Task 10 disables submission for an empty reason, and Task 13 documents the field as required. Because the plan explicitly names the spec as the authority, these steps must allow an empty reason while retaining the maximum-rune validation.

---

**`docs/superpowers/plans/2026-08-11-user-permission-whitelist.md`**

_🏷️ atomicity_ | _🟠 Warning_ | _`distributed-system`_

**Ledger and audit writes can diverge permanently.**

The planned handler commits permission_grants first and invokes best-effort AppendAuditMany afterward. If the process crashes after the ledger commit, or the audit insert fails, the permission changes while AuditView permanently omits the admin action; the caller may observe either 201 or a timeout. This contradicts the stated goal that every permission change appears in admin_audit. Because both collections are in the same MongoDB database, minimally replace the two store calls with one transaction that inserts the grant rows, idempotency record, and audit rows atomically.

---

**`docs/superpowers/plans/2026-08-11-user-permission-whitelist.md`**

_🏷️ consistency_ | _🟠 Warning_ | _`distributed-system`_

**The admin GET can return a decision absent from its ledger.**

listPermissions first calls ListPermissionGrants and then GetLatestPermissionGrant as separate Mongo operations. Concrete interleaving: the list returns the current grant, a revoke commits, and GetLatestPermissionGrant reads that revoke; the response says currentlyGranted=false while its entries contain only the grant. Count and page contents can similarly disagree if an append occurs between CountDocuments and Find. Return entries, total, and the latest row from one repository operation using a single aggregation with $facet, or perform the reads in a snapshot transaction, so each response describes one database snapshot.

---

**`docs/superpowers/plans/2026-08-11-user-permission-whitelist.md`**

_🏷️ operations_ | _🟠 Warning_ | _`mongodb`_

**The replica-set healthcheck reports healthy before MongoDB is writable.**

The proposed healthcheck exits successfully when `rs.initiate()` merely returns, and `rs.status().ok` can also be 1 while the node is not PRIMARY. Docker can therefore release dependent services during election; startup index creation or transactional writes may fail with a not-primary/server-selection error and cause a crash loop. After initiating, exit nonzero until `db.hello().isWritablePrimary` is true; only then let the healthcheck succeed.

---

**`docs/superpowers/plans/2026-08-11-user-permission-whitelist.md`**

_🏷️ reliability_ | _🟠 Warning_ | _`silent-failure`_

**The Mongo readiness loop can wait forever without surfacing failure.**

Task 12's `until docker inspect ... = healthy; do sleep 2; done` has no deadline or exhausted-retry path. A broken replica-set configuration or permanently unhealthy container leaves the verification step hanging indefinitely instead of failing with diagnostics. Bound the attempts and, on exhaustion, print the container health/logs and exit nonzero.

---

**`docs/superpowers/plans/2026-08-11-user-permission-whitelist.md`**

_🏷️ plan-feasibility_ | _🟠 Warning_ | _`tgd-review`_

**Task 8 references the handler before creating it.**

Part B replaces `service.go` and registers `s.GetPermission`, but `service/permission.go` is not created until Part C. Consequently the mock-generation/package-loading step and the claimed wiring-checkpoint `make test SERVICE=user-service` cannot pass as ordered because `UserService.GetPermission` is undefined. Create the handler before registering it, or defer the registration and checkpoint until after Part C.

<details>
<summary>📒 Files reviewed (9)</summary>

* `docs/superpowers/plans/2026-08-11-user-permission-whitelist.md`
* `pkg/errcode/codes_permission.go`
* `pkg/errcode/codes_test.go`
* `pkg/model/model_test.go`
* `pkg/model/permission.go`
* `pkg/model/permission_test.go`
* `pkg/subject/subject.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `silent-failure`
* `tgd-review`

</details>

PR #234, opened 2026-08-11
feat(permission): admin permissions API + user permission.get RPC

nalydadad:
Backend of the permission whitelist, on top of the foundations PR: admin-service owns writes, user-service answers the self-query. Client-API docs land here with the handlers, per repo convention.

**admin-service**
- `POST /v1/admin/permissions` — batch grant/revoke (≤200 subjects, deduped) with applicant/approver/subject existence validation (subjects additionally must be active), civil-date validity window (`YYYY-MM-DD`, fixed UTC+8 interpretation, stored as half-open `[from, day+1)` instants), optional reason (≤1000 runes when present), required `granted` (`*bool` — omitted/null rejected rather than silently recorded as a revoke). Transactional `InsertMany` into the append-only `permission_grants` ledger + best-effort slim `admin_audit` entries. 256KB body limit.
- `GET /v1/admin/permissions` — ledger listing, `subjectAccount`/`permission` independently-optional filters, newest-first; `currentlyGranted` (latest-wins evaluation) present only when both filters are given.
- Store: transactional batch insert, ledger reads with `recordedAt`/`_id` tie-break sort, account-state batch lookup; the two `permission_grants` indexes are created here only (single index owner).

**user-service**
- `permission.get` NATS RPC — the `{account}` token in the subject IS the authenticated identity (request body carries no account field, structurally); read-only repo over the shared collection (primary reads for read-your-writes); same `EvaluateGrant`, fail-closed.

**Docs**: `docs/client-api.md` §9.13/§9.14 (+§6 reason catalog, shared `PermissionGrantView`/`GrantCreated` schemas) and the `permission.get` section + derived request-reply view.

Tests: TDD throughout; handler validation tables cover every rule incl. boundary rows (exactly-200 subjects, exactly-1000 runes, same-day window, from==until off-by-one); integration tests prove transaction all-or-nothing, latest-wins ordering incl. same-millisecond `_id` tie-break, index usage by name (explain), unfiltered/permission-only listing; identity-from-subject and fail-closed paths on the RPC. Full-repo suite, race detector, gosec/govulncheck/semgrep all green at this tree.

Design notes recorded during review, for the record: `granted` was tightened to `*bool` (matching the `room_onduty.OnDuty` precedent) after review flagged that a plain bool silently records a revoke on omission — an irreversible write in an append-only ledger; applicant/approver existence validation was restored to match the spec after review caught the doc/handler contradiction.


julianshen:
**Actionable comments posted: 16** — 🔴 3 blocking · 🟠 13 warning

### 💬 Additional comments (2)


**`admin-service/permissions.go`**

_🏷️ atomicity_ | _🟠 Warning_ | _`mongodb`_

**Permission grants can commit without their audit records.**

InsertPermissionGrants commits permission_grants in its own transaction before AppendAuditMany writes admin_audit outside that transaction. An audit InsertMany error, network ambiguity, or process crash can therefore leave committed grants with missing or partial audit rows, while the handler only logs the error and may return 201. Move both InsertMany operations into one store method and one transaction, passing the transaction context to both collections and returning any failure.

---

**`admin-service/permissions.go`**

_🏷️ idempotency_ | _🟠 Warning_ | _`mongodb`_

**Retrying an ambiguous request can create a new grant operation.**

The transaction protects one invocation only. Every replay generates fresh grant and audit _id values, and there is no persisted request key or unique operation constraint. If the first request commits but its response is lost, retrying the same body inserts another batch; if a revoke was recorded meanwhile, the delayed grant retry receives a newer recordedAt and can incorrectly become the latest decision. Require an idempotency key and persist it with the result in the same transaction under a unique site-scoped index, returning the original result on replay.

<details>

* `admin-service/main.go`
* `admin-service/middleware.go`
* `admin-service/middleware_test.go`
* `admin-service/routes.go`
* `admin-service/store.go`
* `docs/client-api.md`
* `user-service/main.go`
* `user-service/models/permission.go`
* `user-service/models/permission_test.go`
* `user-service/mongorepo/permissions.go`
* `user-service/mongorepo/permissions_test.go`
* `user-service/service/me_test.go`
* `user-service/service/mocks/mock_repository.go`
* `user-service/service/permission.go`
* `user-service/service/permission_test.go`
* `user-service/service/sso_test.go`
* `user-service/service/status_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `silent-failure`
* `tgd-review`

</details>

admin-service/permissions.go
```diff
@@ -0,0 +1,412 @@
...
+			ID:               idgen.GenerateUUIDv7(),
+			SiteID:           h.cfg.SiteID,
+			Permission:       model.PermissionKey(req.Permission),
+			SubjectAccount:   acct,
+			Granted:          granted,
+			ApplicantAccount: req.ApplicantAccount,
+			ApproverAccount:  req.ApproverAccount,
+			Reason:           req.Reason,
+			RecordedBy:       principal.Account,
+			RecordedAt:       now,
```
julianshen, line 246: _🏷️ ordering_ | _🔴 Blocking_ | _`distributed-system`_

**Application-clock ordering can make a successful revoke ineffective.**

The authoritative order is `RecordedAt`, but `now` is sampled before account validation and the transaction, then generated on whichever admin replica handles the request. A grant can receive a later timestamp, stall, and commit before a revoke whose replica clock is behind; both requests return 201, but both readers sort by `recordedAt desc, _id desc` and continue selecting the grant. Concurrent writes encoded in the same BSON millisecond can likewise be ordered by unrelated UUID tie-breakers. Persist a database-coordinated monotonic order per `(siteId, permission, subjectAccount)` in the same transaction and sort on that value; retain `RecordedAt` only as informational time.

<details>

```
In `admin-service/permissions.go` around line 266: The authoritative order is `RecordedAt`, but `now` is sampled before account validation and the transaction, then generated on whichever admin replica handles the request. A grant can receive a later timestamp, stall, and commit before a revoke whose replica clock is behind; both requests return 201, but both readers sort by `recordedAt desc, _id desc` and continue selecting the grant. Concurrent writes encoded in the same BSON millisecond can likewise be ordered by unrelated UUID tie-breakers. Persist a database-coordinated monotonic order per `(siteId, permission, subjectAccount)` in the same transaction and sort on that value; retain `RecordedAt` only as informational time.

```

</details>

admin-service/permissions.go
```diff
@@ -0,0 +1,412 @@
...
+			errcode.WithReason(errcode.PermissionInactiveSubject),
+			errcode.WithMetadata("accounts", strings.Join(inactive, ", "))))
+		return
+	}
+
+	principal := principalFrom(c)
+	grants := make([]*model.PermissionGrant, len(subjects))
+	for i, acct := range subjects {
+		g := &model.PermissionGrant{
+			ID:               idgen.GenerateUUIDv7(),
```
julianshen, line 237: _🏷️ idempotency_ | _🔴 Blocking_ | _`distributed-system`_

**Retries append a new operation instead of replaying the committed result.**

Every attempt generates fresh grant IDs. If Mongo commits the first transaction but the HTTP response is lost or times out, retrying the identical request avoids all duplicate-key protection and appends another ledger and audit batch. If a revoke commits between the original grant and its delayed retry, the retry becomes the latest row and restores access. Require a client-stable idempotency key, persist it with a unique index in the grant transaction, and return the previously stored result when that key is replayed.

<details>

```
In `admin-service/permissions.go` around line 257: Every attempt generates fresh grant IDs. If Mongo commits the first transaction but the HTTP response is lost or times out, retrying the identical request avoids all duplicate-key protection and appends another ledger and audit batch. If a revoke commits between the original grant and its delayed retry, the retry becomes the latest row and restores access. Require a client-stable idempotency key, persist it with a unique index in the grant transaction, and return the previously stored result when that key is replayed.

```

</details>

admin-service/store_mongo.go
```diff
@@ -341,6 +367,152 @@ func (s *storeMongo) ListAudit(ctx context.Context, siteID string, f AuditFilter
...
+// rows (spec §4.4 note on step 11).
+func (s *storeMongo) InsertPermissionGrants(ctx context.Context, grants []*model.PermissionGrant) error {
+	if len(grants) == 0 {
+		return nil
+	}
+	docs := make([]any, len(grants))
+	for i, g := range grants {
+		docs[i] = g
+	}
+	return s.withTransaction(ctx, func(ctx context.Context) error {
```
julianshen, line 402: _🏷️ idempotency_ | _🔴 Blocking_ | _`mongodb`_

**Request retries can duplicate grants and override a later revoke.**

The transaction makes one InsertMany call atomic, but it does not make the HTTP operation idempotent. createPermissions generates fresh UUIDs and a fresh recordedAt for every request, so a client retry after an ambiguous commit inserts new permission_grants rows. A delayed retry of a grant after another administrator has revoked it becomes the newest row and re-enables the permission. Require a client idempotency key, persist it with the ledger operation, and enforce a unique compound index so retries return the original result rather than appending another decision.

<details>

```
In `admin-service/store_mongo.go` around line 399: The transaction makes one InsertMany call atomic, but it does not make the HTTP operation idempotent. createPermissions generates fresh UUIDs and a fresh recordedAt for every request, so a client retry after an ambiguous commit inserts new permission_grants rows. A delayed retry of a grant after another administrator has revoked it becomes the newest row and re-enables the permission. Require a client idempotency key, persist it with the ledger operation, and enforce a unique compound index so retries return the original result rather than appending another decision.

```

</details>

admin-service/permissions.go
```diff
@@ -0,0 +1,412 @@
...
+			ActorUserID:   principal.UserID,
+			ActorAccount:  principal.Account,
+			Action:        action,
+			TargetAccount: g.SubjectAccount,
+			Details:       map[string]string{"permission": string(g.Permission)},
+			SiteID:        h.cfg.SiteID,
+			Timestamp:     now.UnixMilli(),
+		}
+	}
+	if err := h.store.AppendAuditMany(ctx, entries); err != nil {
```
julianshen, line 277: _🏷️ atomicity_ | _🟠 Warning_ | _`distributed-system`_

**Committed permission changes can permanently lose their audit entries.**

The permission transaction commits before `AppendAuditMany`, and audit failure is only logged. For example, client cancellation immediately after commit leaves the request context canceled, so the audit insert fails while the grant or revoke remains authoritative; a transient Mongo failure has the same outcome. There is no durable retry or reconciliation, despite the API promising one audit entry per subject. Insert grants and audit entries in one Mongo transaction, or atomically write an outbox record with the grants and retry audit delivery from that outbox.

<details>

```
In `admin-service/permissions.go` around line 297: The permission transaction commits before `AppendAuditMany`, and audit failure is only logged. For example, client cancellation immediately after commit leaves the request context canceled, so the audit insert fails while the grant or revoke remains authoritative; a transient Mongo failure has the same outcome. There is no durable retry or reconciliation, despite the API promising one audit entry per subject. Insert grants and audit entries in one Mongo transaction, or atomically write an outbox record with the grants and retry audit delivery from that outbox.

```

</details>

admin-service/store_mongo.go
```diff
@@ -341,6 +367,152 @@ func (s *storeMongo) ListAudit(ctx context.Context, siteID string, f AuditFilter
...
+func (s *storeMongo) AppendAuditMany(ctx context.Context, entries []*AuditEntry) error {
+	if len(entries) == 0 {
+		return nil
+	}
+	docs := make([]any, len(entries))
+	for i, e := range entries {
+		docs[i] = e
+	}
+	if _, err := s.adminAudit.InsertMany(ctx, docs); err != nil {
+		return fmt.Errorf("insert audit entries: %w", err)
```
julianshen, line 514: _🏷️ atomicity_ | _🟠 Warning_ | _`mongodb`_

**Permission rows and their audit entries commit independently.**

InsertPermissionGrants commits its transaction before createPermissions calls this separate InsertMany. A MongoDB error, process crash, or request-context cancellation after the ledger commit leaves permission_grants rows with no corresponding admin_audit entries, despite the documented one-entry-per-subject behavior. Move both inserts into one store operation and one transaction, propagating the same session context to both collections; alternatively persist a transactional outbox record with the grant and retry audit delivery durably.

<details>

```
In `admin-service/store_mongo.go` around line 511: InsertPermissionGrants commits its transaction before createPermissions calls this separate InsertMany. A MongoDB error, process crash, or request-context cancellation after the ledger commit leaves permission_grants rows with no corresponding admin_audit entries, despite the documented one-entry-per-subject behavior. Move both inserts into one store operation and one transaction, propagating the same session context to both collections; alternatively persist a transactional outbox record with the grant and retry audit delivery durably.

```

</details>

admin-service/store_mongo.go
```diff
@@ -53,6 +55,30 @@ func (s *storeMongo) EnsureIndexes(ctx context.Context) error {
...
+			{Key: "_id", Value: -1},
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("create permission_grants siteId_permission_subjectAccount_recordedAt_id index: %w", err)
+	}
+
+	// Backs the audit/BI browse (no subjectAccount equality, so index 1 above doesn't apply).
+	_, err = s.permGrants.Indexes().CreateOne(ctx, mongo.IndexModel{
+		Keys: bson.D{{Key: "siteId", Value: 1}, {Key: "recordedAt", Value: -1}},
```
julianshen: _🏷️ indexes_ | _🟠 Warning_ | _`mongodb`_

**The indexes cover only one of four supported filter combinations.**

The five-field index covers the both-filters query, but ListPermissionGrants also supports subject-only, permission-only, and no-filter queries. With an unconstrained permission or subjectAccount between the equality prefix and recordedAt, MongoDB cannot use that index for the requested sort. The fallback {siteId, recordedAt} index also omits the _id tie-break and requires residual scans for either single filter. Since permission_grants is append-only, these paths scan an ever-growing site range and may require a blocking sort. Add indexes {siteId, recordedAt, _id}, {siteId, permission, recordedAt, _id}, and {siteId, subjectAccount, recordedAt, _id}, and add explain tests for all four query shapes.

<details>

```
In `admin-service/store_mongo.go` around line 76: The five-field index covers the both-filters query, but ListPermissionGrants also supports subject-only, permission-only, and no-filter queries. With an unconstrained permission or subjectAccount between the equality prefix and recordedAt, MongoDB cannot use that index for the requested sort. The fallback {siteId, recordedAt} index also omits the _id tie-break and requires residual scans for either single filter. Since permission_grants is append-only, these paths scan an ever-growing site range and may require a blocking sort. Add indexes {siteId, recordedAt, _id}, {siteId, permission, recordedAt, _id}, and {siteId, subjectAccount, recordedAt, _id}, and add explain tests for all four query shapes.

```

</details>

admin-service/store_mongo.go
```diff
@@ -341,6 +367,152 @@ func (s *storeMongo) ListAudit(ctx context.Context, siteID string, f AuditFilter
...
+	if subjectAccount != "" {
+		filter["subjectAccount"] = subjectAccount
+	}
+	if permission != "" {
+		filter["permission"] = permission
+	}
+
+	skip := int64((page - 1) * limit)
+
+	total, err := s.permGrants.CountDocuments(ctx, filter)
```
julianshen, line 428: _🏷️ consistency_ | _🟠 Warning_ | _`mongodb`_

**The list response is assembled from inconsistent snapshots.**

CountDocuments and Find are separate primary reads, and listPermissions subsequently performs another GetLatestPermissionGrant read. An insert between these operations can make total disagree with entries or make currentlyGranted reflect a row absent from the returned ledger. Execute the count, page, and latest-row calculation in one aggregation or expose a store operation that runs all reads in one snapshot transaction with the session context propagated to every query.

<details>

```
In `admin-service/store_mongo.go` around line 425: CountDocuments and Find are separate primary reads, and listPermissions subsequently performs another GetLatestPermissionGrant read. An insert between these operations can make total disagree with entries or make currentlyGranted reflect a row absent from the returned ledger. Execute the count, page, and latest-row calculation in one aggregation or expose a store operation that runs all reads in one snapshot transaction with the session context propagated to every query.

```

</details>

admin-service/store_mongo.go
```diff
@@ -341,6 +367,152 @@ func (s *storeMongo) ListAudit(ctx context.Context, siteID string, f AuditFilter
...
+func (s *storeMongo) ListPermissionGrants(ctx context.Context, siteID, subjectAccount string, permission model.PermissionKey, page, limit int) ([]model.PermissionGrant, int64, error) {
+	filter := bson.M{"siteId": siteID}
+	if subjectAccount != "" {
+		filter["subjectAccount"] = subjectAccount
+	}
+	if permission != "" {
+		filter["permission"] = permission
+	}
+
+	skip := int64((page - 1) * limit)
```
julianshen, line 426: _🏷️ pagination_ | _🟠 Warning_ | _`mongodb`_

**Offset pagination duplicates rows when new grants are appended.**

The ledger is sorted newest-first and remains writable while clients paginate. If a new row is inserted after page 1, every existing offset shifts; page 2 can repeat page 1's last row, and clients stopping at the earlier total can miss historical rows. Large page numbers also require MongoDB to walk all skipped index entries. Replace page/skip pagination with a cursor containing the last (recordedAt, _id) pair and query the corresponding descending range.

<details>

```
In `admin-service/store_mongo.go` around line 423: The ledger is sorted newest-first and remains writable while clients paginate. If a new row is inserted after page 1, every existing offset shifts; page 2 can repeat page 1's last row, and clients stopping at the earlier total can miss historical rows. Large page numbers also require MongoDB to walk all skipped index entries. Replace page/skip pagination with a cursor containing the last (recordedAt, _id) pair and query the corresponding descending range.

```

</details>

admin-service/store_mongo.go
```diff
@@ -53,6 +55,30 @@ func (s *storeMongo) EnsureIndexes(ctx context.Context) error {
...
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("create permission_grants siteId_permission_subjectAccount_recordedAt_id index: %w", err)
+	}
+
+	// Backs the audit/BI browse (no subjectAccount equality, so index 1 above doesn't apply).
+	_, err = s.permGrants.Indexes().CreateOne(ctx, mongo.IndexModel{
+		Keys: bson.D{{Key: "siteId", Value: 1}, {Key: "recordedAt", Value: -1}},
+	})
```
julianshen, line 80: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**The browse index cannot satisfy the tie-break sort.**

ListPermissionGrants sorts by recordedAt and _id, but this fallback index contains only siteId and recordedAt. For an unfiltered site browse, MongoDB therefore needs a blocking sort rather than streaming the first page directly from the index. As the append-only ledger grows, each request performs another full-history scan/sort, increasing database CPU and response latency. Include _id in the index and add explain coverage for the no-filter query.

<details>

> ‼️ **IMPORTANT**

```suggestion
		Keys: bson.D{{Key: "siteId", Value: 1}, {Key: "recordedAt", Value: -1}, {Key: "_id", Value: -1}},
```

</details>

<details>

```
In `admin-service/store_mongo.go` around line 77: ListPermissionGrants sorts by recordedAt and _id, but this fallback index contains only siteId and recordedAt. For an unfiltered site browse, MongoDB therefore needs a blocking sort rather than streaming the first page directly from the index. As the append-only ledger grows, each request performs another full-history scan/sort, increasing database CPU and response latency. Include _id in the index and add explain coverage for the no-filter query.

```

</details>

admin-service/store_mongo.go
```diff
@@ -53,6 +55,30 @@ func (s *storeMongo) EnsureIndexes(ctx context.Context) error {
...
+			{Key: "recordedAt", Value: -1},
+			{Key: "_id", Value: -1},
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("create permission_grants siteId_permission_subjectAccount_recordedAt_id index: %w", err)
+	}
+
+	// Backs the audit/BI browse (no subjectAccount equality, so index 1 above doesn't apply).
+	_, err = s.permGrants.Indexes().CreateOne(ctx, mongo.IndexModel{
```
julianshen, line 78: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Subject-only lookups scan the entire site's ledger.**

The documented subjectAccount-only path has no supporting index: the first compound index cannot use subjectAccount efficiently when permission is unconstrained, while the fallback index has no subjectAccount key. For a site with M historical rows and a sparse subject, both CountDocuments and Find can examine O(M) entries for every page. Since this ledger grows without deletion, latency and MongoDB load grow with all site history rather than that subject's history. Add a {siteId: 1, subjectAccount: 1, recordedAt: -1, _id: -1} index and an explain test for this filter shape.

<details>

```
In `admin-service/store_mongo.go` around line 75: The documented subjectAccount-only path has no supporting index: the first compound index cannot use subjectAccount efficiently when permission is unconstrained, while the fallback index has no subjectAccount key. For a site with M historical rows and a sparse subject, both CountDocuments and Find can examine O(M) entries for every page. Since this ledger grows without deletion, latency and MongoDB load grow with all site history rather than that subject's history. Add a {siteId: 1, subjectAccount: 1, recordedAt: -1, _id: -1} index and an explain test for this filter shape.

```

</details>

admin-service/permissions.go
```diff
@@ -0,0 +1,412 @@
...
+	resp := listPermissionsResponse{
+		Entries: entries,
+		Total:   total,
+	}
+
+	// currentlyGranted is a latest-wins decision over (permission, subjectAccount) —
+	// meaningless without both, so it's computed and included only when the
+	// caller supplied both filters (spec §4.6).
+	if permission != "" && subjectAccount != "" {
+		latest, err := h.store.GetLatestPermissionGrant(ctx, h.cfg.SiteID, permission, subjectAccount)
```
julianshen, line 382: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Default filtered lookups repeat the latest-row query.**

With both filters and the default page 1, ListPermissionGrants has already fetched rows in the exact recordedAt/_id order used by GetLatestPermissionGrant, so grants[0] is the latest row and an empty result means no grant. This line adds a third sequential MongoDB command after the count and page query on every normal console lookup, adding one network round trip and 50% more database commands on this path. Reuse grants[0] on page 1, and retain GetLatestPermissionGrant only for later pages where the latest row is not present.

<details>

```
In `admin-service/permissions.go` around line 402: With both filters and the default page 1, ListPermissionGrants has already fetched rows in the exact recordedAt/_id order used by GetLatestPermissionGrant, so grants[0] is the latest row and an empty result means no grant. This line adds a third sequential MongoDB command after the count and page query on every normal console lookup, adding one network round trip and 50% more database commands on this path. Reuse grants[0] on page 1, and retain GetLatestPermissionGrant only for later pages where the latest row is not present.

```

</details>

admin-service/permissions.go
```diff
@@ -0,0 +1,412 @@
...
+// rule (spec §4.4 step 9). Granted is a pointer for the same reason: an
+// omitted "granted" must be rejected, not silently default to false/revoke —
+// see the missing-fields check below (no binding tag, to keep the reason
+// grouped with applicantAccount/approverAccount as missing_permission_fields
+// rather than the generic JSON-bind bad_request).
+type permissionRequest struct {
+	Permission       string   `json:"permission"`
+	SubjectAccounts  []string `json:"subjectAccounts"`
+	Granted          *bool    `json:"granted"`
+	EffectiveFrom    *string  `json:"effectiveFrom"` // "2026-09-01"; nil on grant = now
```
julianshen, line 40: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Explicit null revoke windows are accepted as if omitted.**

Unmarshalling JSON `null` into `*string` produces nil, so a revoke containing `"effectiveFrom": null` or `"expiresAt": null` passes the later nil check. The documented contract requires these fields to be entirely absent on revokes. Track key presence separately, such as with a presence-aware type or raw JSON fields, and add tests for both explicit-null cases.

<details>

```
In `admin-service/permissions.go` around line 42: Unmarshalling JSON `null` into `*string` produces nil, so a revoke containing `"effectiveFrom": null` or `"expiresAt": null` passes the later nil check. The documented contract requires these fields to be entirely absent on revokes. Track key presence separately, such as with a presence-aware type or raw JSON fields, and add tests for both explicit-null cases.

```

</details>

admin-service/main.go
```diff
@@ -70,6 +70,7 @@ func run() error {
 	r.Use(gin.Recovery())
 	r.Use(ginutil.RequestID())
 	r.Use(ginutil.AccessLog())
+	r.Use(bodyLimit(maxPermissionBodyBytes))
```
julianshen: _🏷️ intent-match_ | _🟠 Warning_ | _`tgd-review`_

**The permission body cap is applied to every admin endpoint.**

Registering `bodyLimit(maxPermissionBodyBytes)` as engine-wide middleware changes request handling for login, password, user, room, and other routes, although the constant, tests, and API documentation describe a permission-request limit. Scope this middleware to `POST /v1/admin/permissions` in route registration to avoid an undocumented behavior change to existing endpoints.

<details>

```
In `admin-service/main.go` around line 73: Registering `bodyLimit(maxPermissionBodyBytes)` as engine-wide middleware changes request handling for login, password, user, room, and other routes, although the constant, tests, and API documentation describe a permission-request limit. Scope this middleware to `POST /v1/admin/permissions` in route registration to avoid an undocumented behavior change to existing endpoints.

```

</details>

docs/client-api/request-reply.md
```diff
@@ -251,6 +251,8 @@ matching `siteId`). Full schemas, examples, and error tables are in
 | `GET /v1/admin/audit` | synchronous HTTP | List the audit log (§9.9). |
 | `POST /v1/admin/rooms/:roomId/onduty` | synchronous HTTP | Toggle a channel's on-duty state: maps the boolean onto `restricted` + `externalAccess` via room-service's restrict RPC, with `ownerAccount` required when turning on. Emits a `room_restricted` room event; no system message, so nothing is displayed (§9.12). |
 | `POST /v1/password/change` | synchronous HTTP | Logged-in admin's self-service password change (§9.11). |
+| `POST /v1/admin/permissions` | synchronous HTTP | Grant or revoke a permission for one or more subject accounts; appends to the permission ledger with one slim audit entry per subject (§9.13). |
+| `GET /v1/admin/permissions?subjectAccount=<account>` | synchronous HTTP | List an account's permission ledger newest-first, plus the current computed decision (§9.14). |
```
julianshen, line 255: _📝 documentation_ | _🔵 Suggestion_ | _`tgd-review`_

**The summary omits a filter required for currentlyGranted.**

This row promises the current computed decision while showing only `subjectAccount`. The implementation and §9.14 include `currentlyGranted` only when both `subjectAccount` and `permission` are supplied.

<details>

> ‼️ **IMPORTANT**

```suggestion
| `GET /v1/admin/permissions?subjectAccount=<account>&permission=<permission>` | synchronous HTTP | List an account's permission ledger newest-first, plus the current computed decision (§9.14). |
```

</details>

<details>

```
In `docs/client-api/request-reply.md` around line 255: This row promises the current computed decision while showing only `subjectAccount`. The implementation and §9.14 include `currentlyGranted` only when both `subjectAccount` and `permission` are supplied.

```

</details>

admin-service/store_mongo.go
```diff
@@ -341,6 +370,152 @@ func (s *storeMongo) ListAudit(ctx context.Context, siteID string, f AuditFilter
...
+	"approverAccount":  1,
+	"reason":           1,
+	"recordedBy":       1,
+	"recordedAt":       1,
+}
+
+// InsertPermissionGrants appends the batch atomically (withTransaction +
+// InsertMany), so a resend after a partial failure cannot produce duplicate
+// rows (spec §4.4 note on step 11).
+func (s *storeMongo) InsertPermissionGrants(ctx context.Context, grants []*model.PermissionGrant) error {
```
julianshen, line 394: _🏷️ idempotency_ | _🔴 Blocking_ | _`distributed-system`_

**Retries can turn one permission change into a newer duplicate.**

The transaction makes each batch atomic but does not make a resend idempotent: every HTTP attempt generates new grant IDs and a new recordedAt. If the transaction commits but the response is lost or commit acknowledgement times out, the caller can retry and append another logical operation. A delayed retry of a grant after another admin has revoked it becomes the newest row and re-enables the permission. Require a caller-stable idempotency key, persist it with a unique constraint in the same transaction, and return the previously stored result when that key is replayed.

<details>

```
In `admin-service/store_mongo.go` around line 394: The transaction makes each batch atomic but does not make a resend idempotent: every HTTP attempt generates new grant IDs and a new recordedAt. If the transaction commits but the response is lost or commit acknowledgement times out, the caller can retry and append another logical operation. A delayed retry of a grant after another admin has revoked it becomes the newest row and re-enables the permission. Require a caller-stable idempotency key, persist it with a unique constraint in the same transaction, and return the previously stored result when that key is replayed.

```

</details>

admin-service/permissions.go
```diff
@@ -0,0 +1,392 @@
...
+			ID:               idgen.GenerateUUIDv7(),
+			SiteID:           h.cfg.SiteID,
+			Permission:       model.PermissionKey(req.Permission),
+			SubjectAccount:   acct,
+			Granted:          granted,
+			ApplicantAccount: req.ApplicantAccount,
+			ApproverAccount:  req.ApproverAccount,
+			Reason:           req.Reason,
+			RecordedBy:       principal.Account,
+			RecordedAt:       now,
```
julianshen, line 246: _🏷️ ordering_ | _🔴 Blocking_ | _`distributed-system`_

**Replica clocks can make a completed revoke lose to an older grant.**

Latest-wins ordering is based on recordedAt from the admin-service replica's wall clock. For example, replica A is five minutes ahead and commits a grant; after its 201 response, replica B commits a revoke with the correct clock. The revoke has a smaller recordedAt, so both admin-service and user-service permanently select the earlier grant despite the causal ordering. The _id tie-breaker only helps equal timestamps. Allocate an authoritative monotonic sequence/version in MongoDB within the write transaction and sort by it; retain recordedAt only as display metadata.

<details>

```
In `admin-service/permissions.go` around line 246: Latest-wins ordering is based on recordedAt from the admin-service replica's wall clock. For example, replica A is five minutes ahead and commits a grant; after its 201 response, replica B commits a revoke with the correct clock. The revoke has a smaller recordedAt, so both admin-service and user-service permanently select the earlier grant despite the causal ordering. The _id tie-breaker only helps equal timestamps. Allocate an authoritative monotonic sequence/version in MongoDB within the write transaction and sort by it; retain recordedAt only as display metadata.

```

</details>

user-service/service/service.go
```diff
@@ -185,4 +193,5 @@ func (s *UserService) RegisterHandlers(r *natsrouter.Router) {
 	natsrouter.RegisterNoBody(r, subject.UserAppsCategoriesPattern(s.siteID), s.ListAppCategories)
 	natsrouter.Register(r, subject.UserSSOSetPattern(s.siteID), s.SSOSet)
 	natsrouter.RegisterOptionalBody(r, subject.UserSSORefreshPattern(s.siteID), s.SSORefresh)
+	natsrouter.Register(r, subject.UserPermissionGetPattern(s.siteID), s.GetPermission)
```
julianshen, line 196: _🏷️ nats-subject_ | _🔴 Blocking_ | _`nats`_

**The new NATS subject constructor is missing.**

The handler registration calls `subject.UserPermissionGetPattern`, but this PR adds no definition for that constructor in the subject package. The user service therefore cannot compile or register the new responder. Add a constructor matching `chat.user.*.request.user.{siteID}.permission.get`, using exactly one wildcard token for the account and the validated site ID as one literal token, plus a registration test that verifies `account` extraction.

<details>

```
In `user-service/service/service.go` around line 196: The handler registration calls `subject.UserPermissionGetPattern`, but this PR adds no definition for that constructor in the subject package. The user service therefore cannot compile or register the new responder. Add a constructor matching `chat.user.*.request.user.{siteID}.permission.get`, using exactly one wildcard token for the account and the validated site ID as one literal token, plus a registration test that verifies `account` extraction.

```

</details>

admin-service/permissions.go
```diff
@@ -0,0 +1,392 @@
...
+			ActorUserID:   principal.UserID,
+			ActorAccount:  principal.Account,
+			Action:        action,
+			TargetAccount: g.SubjectAccount,
+			Details:       map[string]string{"permission": string(g.Permission)},
+			SiteID:        h.cfg.SiteID,
+			Timestamp:     now.UnixMilli(),
+		}
+	}
+	if err := h.store.AppendAuditMany(ctx, entries); err != nil {
```
julianshen, line 277: _🏷️ atomicity_ | _🟠 Warning_ | _`distributed-system`_

**Ledger writes can commit without their audit records.**

The permission transaction commits before AppendAuditMany, and audit failure is only logged. A client disconnect immediately after the ledger commit cancels the shared request context, or MongoDB can fail between these calls, leaving an effective permission change with no audit row while the handler may still report 201. InsertMany can also partially insert an audit batch before returning an error. Insert grants and audit entries in one multi-collection MongoDB transaction, or transactionally write an outbox record that is retried until the audit entries are durably recorded.

<details>

```
In `admin-service/permissions.go` around line 277: The permission transaction commits before AppendAuditMany, and audit failure is only logged. A client disconnect immediately after the ledger commit cancels the shared request context, or MongoDB can fail between these calls, leaving an effective permission change with no audit row while the handler may still report 201. InsertMany can also partially insert an audit batch before returning an error. Insert grants and audit entries in one multi-collection MongoDB transaction, or transactionally write an outbox record that is retried until the audit entries are durably recorded.

```

</details>

admin-service/store_mongo.go
```diff
@@ -341,6 +370,152 @@ func (s *storeMongo) ListAudit(ctx context.Context, siteID string, f AuditFilter
...
+		filter["subjectAccount"] = subjectAccount
+	}
+	if permission != "" {
+		filter["permission"] = permission
+	}
+
+	skip := int64((page - 1) * limit)
+
+	total, err := s.permGrants.CountDocuments(ctx, filter)
+	if err != nil {
```
julianshen, line 429: _🏷️ consistency_ | _🟠 Warning_ | _`distributed-system`_

**Concurrent appends can produce a self-contradictory list response.**

CountDocuments and Find are separate reads, and the handler subsequently performs another GetLatestPermissionGrant read. If a revoke commits after the count but before Find, total can be 1 while entries contains both rows; if it commits after Find but before GetLatestPermissionGrant, entries can show only a grant while currentlyGranted is false. This breaks the endpoint's confirmation workflow. Return the count, requested page, and latest decision from one aggregation using $facet, or execute all reads in a transaction with snapshot read concern.

<details>

```
In `admin-service/store_mongo.go` around line 429: CountDocuments and Find are separate reads, and the handler subsequently performs another GetLatestPermissionGrant read. If a revoke commits after the count but before Find, total can be 1 while entries contains both rows; if it commits after Find but before GetLatestPermissionGrant, entries can show only a grant while currentlyGranted is false. This breaks the endpoint's confirmation workflow. Return the count, requested page, and latest decision from one aggregation using $facet, or execute all reads in a transaction with snapshot read concern.

```

</details>

user-service/service/permission.go
```diff
@@ -0,0 +1,28 @@
...
+	c.WithLogValues("account", account)
+	key := model.PermissionKey(req.Permission)
+	if !model.KnownPermission(key) {
+		return nil, errcode.BadRequest("unknown permission", errcode.WithReason(errcode.PermissionUnknownKey))
+	}
+	latest, err := s.permissions.GetLatestGrant(c, s.siteID, key, account)
+	if err != nil {
+		return nil, fmt.Errorf("get latest permission grant: %w", err)
+	}
+	granted := model.EvaluateGrant(latest, time.Now().UTC())
```
julianshen, line 26: _🏷️ time-consistency_ | _🟠 Warning_ | _`distributed-system`_

**Clock skew makes permission checks disagree at validity boundaries.**

Validity is evaluated with each user-service replica's local wall clock. Around effectiveFrom or expiresAt, two replicas with opposite clock skew can return different granted values for the same unchanged ledger row, so repeated NATS requests can alternate between allowed and denied. Evaluate the window against one authoritative time source, preferably MongoDB server time in the same query/aggregation that selects the latest row, so all replicas use the same decision instant.

<details>

```
In `user-service/service/permission.go` around line 26: Validity is evaluated with each user-service replica's local wall clock. Around effectiveFrom or expiresAt, two replicas with opposite clock skew can return different granted values for the same unchanged ledger row, so repeated NATS requests can alternate between allowed and denied. Evaluate the window against one authoritative time source, preferably MongoDB server time in the same query/aggregation that selects the latest row, so all replicas use the same decision instant.

```

</details>

admin-service/store_mongo.go
```diff
@@ -53,6 +55,33 @@ func (s *storeMongo) EnsureIndexes(ctx context.Context) error {
 		return fmt.Errorf("create admin_audit siteId_targetAccount_timestamp index: %w", err)
 	}
 
+	// Backs ListPermissionGrants/GetLatestPermissionGrant: equality prefix
```
julianshen, line 58: _🏷️ indexes_ | _🟠 Warning_ | _`mongodb`_

**Optional filter shapes scan the site-wide permission ledger.**

ListPermissionGrants documents subject-only and permission-only queries, but neither index covers those filters together with the recordedAt/_id sort. For subject-only requests, the five-field index cannot constrain subjectAccount because permission is missing from its prefix; the browse index can preserve order only by scanning all rows for the site and applying a residual filter. Permission-only requests likewise cannot use the first index for ordering because subjectAccount interrupts the sort suffix. Since this ledger is append-only, unrelated history makes these requests progressively more expensive. Add {siteId:1, subjectAccount:1, recordedAt:-1, _id:-1} and {siteId:1, permission:1, recordedAt:-1, _id:-1} indexes.

<details>

```
In `admin-service/store_mongo.go` around line 58: ListPermissionGrants documents subject-only and permission-only queries, but neither index covers those filters together with the recordedAt/_id sort. For subject-only requests, the five-field index cannot constrain subjectAccount because permission is missing from its prefix; the browse index can preserve order only by scanning all rows for the site and applying a residual filter. Permission-only requests likewise cannot use the first index for ordering because subjectAccount interrupts the sort suffix. Since this ledger is append-only, unrelated history makes these requests progressively more expensive. Add {siteId:1, subjectAccount:1, recordedAt:-1, _id:-1} and {siteId:1, permission:1, recordedAt:-1, _id:-1} indexes.

```

</details>

admin-service/store_mongo.go
julianshen, line 429: _🏷️ pagination_ | _🟠 Warning_ | _`mongodb`_

**Offset pagination duplicates rows during concurrent inserts.**

SetSkip implements page-number pagination over a ledger that is concurrently appended newest-first. If a new grant is inserted after page 1, every prior row shifts by one and page 2 can repeat the last row from page 1. Large page numbers also force MongoDB to walk and discard the entire offset. Replace page/skip with seek pagination using the last (recordedAt, _id) pair and a range predicate matching the descending compound indexes.

<details>

```
In `admin-service/store_mongo.go` around line 429: SetSkip implements page-number pagination over a ledger that is concurrently appended newest-first. If a new grant is inserted after page 1, every prior row shifts by one and page 2 can repeat the last row from page 1. Large page numbers also force MongoDB to walk and discard the entire offset. Replace page/skip with seek pagination using the last (recordedAt, _id) pair and a range predicate matching the descending compound indexes.

```

</details>

admin-service/store_mongo.go
```diff
@@ -341,6 +370,152 @@ func (s *storeMongo) ListAudit(ctx context.Context, siteID string, f AuditFilter
...
+	}
+	return s.withTransaction(ctx, func(ctx context.Context) error {
+		if _, err := s.permGrants.InsertMany(ctx, docs); err != nil {
+			return fmt.Errorf("insert permission grants: %w", err)
+		}
+		return nil
+	})
+}
+
+// ListPermissionGrants returns the ledger for the site newest-first
```
julianshen, line 410: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Two supported filter shapes lack sort-compatible indexes.**

The subject-only and permission-only GET paths are reachable API combinations, but neither existing index can both filter and provide the required recordedAt/_id ordering. As a site's append-only ledger grows, MongoDB must either scan the site-wide browse index with a residual filter or use a narrower index followed by a blocking sort; CountDocuments also lacks an efficient subject-only prefix. Add {siteId, subjectAccount, recordedAt:-1, _id:-1} and {siteId, permission, recordedAt:-1, _id:-1} indexes, and extend the explain tests to cover both filter shapes.

<details>

```
In `admin-service/store_mongo.go` around line 410: The subject-only and permission-only GET paths are reachable API combinations, but neither existing index can both filter and provide the required recordedAt/_id ordering. As a site's append-only ledger grows, MongoDB must either scan the site-wide browse index with a residual filter or use a narrower index followed by a blocking sort; CountDocuments also lacks an efficient subject-only prefix. Add {siteId, subjectAccount, recordedAt:-1, _id:-1} and {siteId, permission, recordedAt:-1, _id:-1} indexes, and extend the explain tests to cover both filter shapes.

```

</details>

admin-service/store_mongo.go
```diff
@@ -341,6 +370,152 @@ func (s *storeMongo) ListAudit(ctx context.Context, siteID string, f AuditFilter
...
+	}
+	if permission != "" {
+		filter["permission"] = permission
+	}
+
+	skip := int64((page - 1) * limit)
+
+	total, err := s.permGrants.CountDocuments(ctx, filter)
+	if err != nil {
+		return nil, 0, fmt.Errorf("count permission grants: %w", err)
```
julianshen, line 430: _🎯 concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Concurrent writes can produce internally inconsistent list responses.**

CountDocuments and Find are separate reads, followed by another independent GetLatestPermissionGrant read in the handler when both filters are present. For example, a revoke inserted after Find but before GetLatestPermissionGrant yields entries showing only the prior grant while currentlyGranted is false; an insert between CountDocuments and Find can also return entries inconsistent with total. Perform count, page retrieval, and latest-row retrieval from one snapshot, preferably with one aggregation using $facet, or use a snapshot transaction and return the latest row through the same store operation.

<details>

```
In `admin-service/store_mongo.go` around line 430: CountDocuments and Find are separate reads, followed by another independent GetLatestPermissionGrant read in the handler when both filters are present. For example, a revoke inserted after Find but before GetLatestPermissionGrant yields entries showing only the prior grant while currentlyGranted is false; an insert between CountDocuments and Find can also return entries inconsistent with total. Perform count, page retrieval, and latest-row retrieval from one snapshot, preferably with one aggregation using $facet, or use a snapshot transaction and return the latest row through the same store operation.

```

</details>

admin-service/store_mongo.go
```diff
@@ -341,6 +370,152 @@ func (s *storeMongo) ListAudit(ctx context.Context, siteID string, f AuditFilter
...
+	if subjectAccount != "" {
+		filter["subjectAccount"] = subjectAccount
+	}
+	if permission != "" {
+		filter["permission"] = permission
+	}
+
+	skip := int64((page - 1) * limit)
+
+	total, err := s.permGrants.CountDocuments(ctx, filter)
```
julianshen, line 428: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Unbounded offset pagination can repeatedly scan the full ledger.**

The documented page parameter has no maximum, and skip-based pagination requires MongoDB to walk all preceding matching index entries. On this append-only collection, deep valid pages perform an exact full count and then another near-full scan, while arbitrarily out-of-range pages still issue the second query after counting. Repeated deep-page requests can consume substantial MongoDB CPU and I/O. Bound page/offset, return an empty page immediately when skip >= total, and use cursor/keyset pagination if deep traversal must remain supported.

<details>

```
In `admin-service/store_mongo.go` around line 428: The documented page parameter has no maximum, and skip-based pagination requires MongoDB to walk all preceding matching index entries. On this append-only collection, deep valid pages perform an exact full count and then another near-full scan, while arbitrarily out-of-range pages still issue the second query after counting. Repeated deep-page requests can consume substantial MongoDB CPU and I/O. Bound page/offset, return an empty page immediately when skip >= total, and use cursor/keyset pagination if deep traversal must remain supported.

```

</details>

admin-service/permissions.go
```diff
@@ -0,0 +1,392 @@
...
+// rule (spec §4.4 step 9). Granted is a pointer for the same reason: an
+// omitted "granted" must be rejected, not silently default to false/revoke —
+// see the missing-fields check below (no binding tag, to keep the reason
+// grouped with applicantAccount/approverAccount as missing_permission_fields
+// rather than the generic JSON-bind bad_request).
+type permissionRequest struct {
+	Permission       string   `json:"permission"`
+	SubjectAccounts  []string `json:"subjectAccounts"`
+	Granted          *bool    `json:"granted"`
+	EffectiveFrom    *string  `json:"effectiveFrom"` // "2026-09-01"; nil on grant = now
```
julianshen, line 40: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Explicit null revoke windows are accepted as omitted.**

A `*string` cannot distinguish an absent JSON field from an explicit `null`; both decode to nil. Consequently, a revoke containing `"effectiveFrom": null` or `"expiresAt": null` passes the nil checks and is created, contrary to the documented requirement that these fields be entirely absent. Track field presence during unmarshalling and add tests for both null fields.

<details>

```
In `admin-service/permissions.go` around line 40: A `*string` cannot distinguish an absent JSON field from an explicit `null`; both decode to nil. Consequently, a revoke containing `"effectiveFrom": null` or `"expiresAt": null` passes the nil checks and is created, contrary to the documented requirement that these fields be entirely absent. Track field presence during unmarshalling and add tests for both null fields.

```

</details>

admin-service/middleware.go
```diff
@@ -84,3 +85,17 @@ func principalFrom(c *gin.Context) session.Session {
...
+// maxRequestBodyBytes caps every admin-service request body (registered on the
+// root engine in main.go), not any single endpoint's.
+const maxRequestBodyBytes = 1 << 20 // 1MB
+
+// bodyLimit caps request bodies at max bytes; a caller that exceeds it gets a
+// truncated read, which fails ShouldBindJSON downstream and surfaces as an
+// ordinary 400 (spec §9 — no 413 in this service's closed Code set).
+func bodyLimit(max int64) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		c.Request.Body = http.MaxBytesReader(c.Writer, c.Request.Body, max)
```
julianshen, line 98: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Oversized bodies can pass when JSON ends before the limit.**

`http.MaxBytesReader` reports an error only when a consumer reads beyond the limit. Gin's JSON binder decodes one value, so a valid request object followed by enough whitespace or trailing data to exceed 1MB can be accepted without reaching the reader limit. This contradicts the documented guarantee that every body over 1MB returns 400. Enforce known oversized `Content-Length` values and ensure chunked bodies are consumed through an explicit max+1-byte size check before invoking handlers.

<details>

```
In `admin-service/middleware.go` around line 98: `http.MaxBytesReader` reports an error only when a consumer reads beyond the limit. Gin's JSON binder decodes one value, so a valid request object followed by enough whitespace or trailing data to exceed 1MB can be accepted without reaching the reader limit. This contradicts the documented guarantee that every body over 1MB returns 400. Enforce known oversized `Content-Length` values and ensure chunked bodies are consumed through an explicit max+1-byte size check before invoking handlers.

```

</details>

admin-service/permissions_test.go
```diff
@@ -0,0 +1,930 @@
...
+
+// strPtr returns a pointer to s, for constructing permissionRequest's
+// optional *string date fields in test bodies.
+func strPtr(s string) *string { return &s }
+
+// futureDate returns the civil date (fixed UTC+8 rule) days from now as
+// YYYY-MM-DD, for handler test rows that must keep exercising their named
+// validation branch regardless of when the suite runs — a hardcoded date
+// would silently decay into the until-not-past branch once it lapses.
+func futureDate(days int) string {
```
julianshen, line 52: _🧪 tests_ | _🟠 Warning_ | _`tgd-review`_

**Success tests expire after December 2026.**

Although `futureDate` exists specifically to avoid time-dependent fixtures, numerous successful create-permission cases still use `expiresAt: "2026-12-31"`. Once that date passes in UTC+8, they will enter the expired-window branch and fail instead of exercising their intended behavior. Use dynamically generated future dates in all handler success fixtures and derive expected stored instants from those values.

<details>

```
In `admin-service/permissions_test.go` around line 52: Although `futureDate` exists specifically to avoid time-dependent fixtures, numerous successful create-permission cases still use `expiresAt: "2026-12-31"`. Once that date passes in UTC+8, they will enter the expired-window branch and fail instead of exercising their intended behavior. Use dynamically generated future dates in all handler success fixtures and derive expected stored instants from those values.

```

</details>

PR #236, opened 2026-08-11
feat(obs): upgrade to o11y v0.11.0

flywindy:
## Summary

- Upgrade `github.com/flywindy/o11y` from v0.9.1 to **v0.11.0** (this branch landed v0.10.0 first, then v0.11.0), which moves `akira-core/instrumentation-go/otel-nats` from v0.8.0 to v0.9.1.
- Enable Cassandra table attribution and prove `db.collection.name` on both operation-duration and query-attempt metrics with an integration test.
- Add trusted identity propagation for `user.name`, `chat.room.id`, and `chat.site.id` across NATS RPCs and relevant JetStream workers.
- Expose the otel-nats tracing toggle ladder (`relay > environment > SDK default`) through documented environment variables.
- Harden telemetry trust boundaries: public HTTP ingress accepts TraceContext only, while OIDC, Microsoft Graph, and translation-provider egress strip W3C baggage but preserve trace correlation.
- Document the v0.11.0 NATS span-name change in `docs/specs/o11y/o11y-trace-design.md`.

## Why

o11y v0.10.0 provides the supported baggage-attribute mechanism, Cassandra table metrics, and dynamic otel-nats feature flags needed by the observability rollout. The existing application had uneven router middleware stacks and no uniform baggage boundary policy, which could leave some NATS routes without identity attributes and could allow caller-controlled or internal PII baggage to cross HTTP trust boundaries.

o11y v0.11.0 is span-naming conformance work on top of that: it aligns NATS span names with the semconv v1.39.0 messaging shape and bounds their cardinality. It has no API removals and no dependency changes beyond otel-nats itself, so no service code changed for the bump.

## Breaking change: NATS span names (v0.11.0)

Span names now follow `{messaging.operation.name} {destination}`, and the destination is omitted when no low-cardinality value exists. There is no compatibility shim — **dashboards, TraceQL/trace-search queries, alerts, and `spanmetrics` connector rules keyed on the old names must be updated**.

| Old name (otel-nats ≤ v0.8.0) | New name |
|---|---|
| `send {subject}` (publish, core and JetStream) | `publish {subject}` |
| `{subject} request` | `request {subject}` |
| `receive {inbox}` (reply-receive) | `receive` |
| `publish {inbox}` (manual reply via `msg.Reply`) | `publish` |
| `process {inbox}` (handler subscribed on an inbox) | `process` |
| `receive`/`process {delivered subject}` for a JetStream consumer with a single-valued wildcard filter | `receive`/`process {filter subject}` |

A reply inbox is auto-generated and single-use, so semconv omits it from the span name rather than turning span names into an unbounded backend dimension. It stays queryable on `messaging.destination.name`, `messaging.message.conversation_id`, and the new `messaging.destination.temporary` / `messaging.destination.anonymous` booleans — **correlate replies by attribute, not by span name**. `messaging.destination.template` carries the bounded destination when it differs from the concrete subject.

## Behavior and operational impact

- `user.name` is emitted only when `O11Y_USER_BAGGAGE_ENABLED=true`; the default remains off because it is PII, including in `docker-local`.
- Room and site identifiers use application attributes `chat.room.id` and `chat.site.id`; OpenTelemetry has no generic chat-room/site semantic convention.
- Cassandra table labels add bounded series cardinality (o11y defaults to a 200-table bound).
- NATS tracing can be toggled with `OTEL_NATS_TRACING_ENABLED`, or dynamically through `OTEL_INSTRUMENTATION_GO_FLAGS_ENDPOINT` using `otel-nats-tracing` and the process-wide `otel-instrumentation-go-tracing` veto.
- Invalid tracing boolean/relay configuration fails connection construction instead of silently falling back.
- Public HTTP callers cannot inject baggage: `ginutil.CORS` still advertises the `baggage` request header so a browser on the previous contract is not blocked at preflight during rollout, but the contents are dropped at extraction by `obs.PublicIngressPropagator()`. Internal NATS hops retain TraceContext + Baggage propagation.

## Validation

- `make fmt`
- `make lint` — 0 issues
- `make test` (race detector, all packages)
- `make test-integration SERVICE=pkg/cassutil`
- `make sast-gosec` — 0 findings
- `make sast-vuln` / `make sast-semgrep` — covered by the CI `sast` gate (`vuln.go.dev` is unreachable from the environment the v0.11.0 bump was prepared in)
- `docker compose -f docker-local/compose.services.yaml config --services`
- `git diff --check origin/main...HEAD`

## Review notes

- No client-facing request/response or event schema changed, so `docs/client-api.md` and its derived views do not require updates.
- No store interface changed, so mock regeneration is not required.
- The `pkg/obs/obs.go` "master-off still installs baggage materialization options" review finding is a false positive: the `if !c.Enabled` branch `return`s before the baggage options are appended, so `off.options()` is 6 base + 4 pillar disables = 10, matching `TestOptions_MasterOffDisablesPillars`.
- Session-local/untracked files are excluded from the branch.

## Summary by CodeRabbit

* **New Features**
  * Improved distributed tracing across HTTP, messaging, authentication, and background processing.
  * Added trusted account, room, and site context to telemetry for easier diagnosis and monitoring.
  * Added opt-in user context and feature-flag relay configuration for local observability environments.
  * Improved tracing and identity context across client-facing and internal message flows.

* **Bug Fixes**
  * Prevented sensitive identity baggage from being forwarded to external services while preserving trace correlation.
  * Added clearer validation and fast failure for invalid tracing configuration.

* **Documentation**
  * Updated observability guidance, configuration details, and trace naming conventions.


julianshen:
**Actionable comments posted: 3** — 🟠 3 warning

<details>
<summary>📒 Files reviewed (54)</summary>

* `admin-service/main.go`
* `admin-service/room_onduty_integration_test.go`
* `auth-service/handler.go`
* `auth-service/handler_test.go`
* `auth-service/main.go`
* `bot-room-service/main.go`
* `client-update-service/main.go`
* `docker-local/compose.services.yaml`
* `docs/specs/o11y/o11y-integration.md`
* `docs/specs/o11y/o11y-trace-design.md`
* `go.mod`
* `go.sum`
* `history-service/internal/cassrepo/walker_test.go`
* `media-service/main.go`
* `message-gatekeeper/handler_test.go`
* `message-worker/handler.go`
* `pkg/cassutil/integration_test.go`
* `pkg/ginutil/middleware.go`
* `pkg/ginutil/middleware_test.go`
* `pkg/msgraph/chats.go`
* `pkg/msgraph/groups.go`
* `pkg/msgraph/msgraph.go`
* `pkg/msgraph/msgraph_test.go`
* `pkg/msgraph/presence.go`
* `pkg/natsrouter/middleware.go`
* `pkg/natsrouter/router.go`
* `pkg/natsutil/connect.go`
* `pkg/natsutil/connect_test.go`
* `pkg/obs/obs.go`
* `pkg/obs/obs_test.go`
* `pkg/oidc/oidc_test.go`
* `pkg/oidc/refresh.go`
* `pkg/subject/subject.go`
* `portal-service/main.go`
* `room-service/main.go`
* `room-worker/main.go`
* `search-service/main.go`
* `tcard-service/main.go`
* `teams-room-inspector/main.go`
* `translation-service/token.go`
* `translation-service/translator_stream.go`
* `translation-service/translator_stream_test.go`
* `upload-service/main.go`
* `user-service/main.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `silent-failure`
* `tgd-review`

</details>

pkg/obs/obs.go
```diff
@@ -101,6 +123,10 @@ func (c *Config) options() []o11y.Option {
 			o11y.WithProfilingEnabled(false),
 		)
 	}
+	opts = append(opts, o11y.WithBaggageAttributes(RoomIDKey, SiteIDKey))
+	if c.UserBaggageEnabled {
+		opts = append(opts, o11y.WithUserBaggage())
+	}
```
julianshen: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**Master-off still installs baggage materialization options.**

These options are appended even when `O11Y_ENABLED=false`, contradicting the zero-impact master-switch behavior and the updated `TestOptions_MasterOffDisablesPillars` expectation. With the existing six base options and four disable options, this unconditional append makes `off.options()` contain 11 options rather than the asserted 10; enabling user baggage would add another. Gate both baggage options on `c.Enabled`.

<details>

> ‼️ **IMPORTANT**

```suggestion
	if c.Enabled {
		opts = append(opts, o11y.WithBaggageAttributes(RoomIDKey, SiteIDKey))
		if c.UserBaggageEnabled {
			opts = append(opts, o11y.WithUserBaggage())
		}
	}
```

</details>

<details>

```
In `pkg/obs/obs.go` around line 126: These options are appended even when `O11Y_ENABLED=false`, contradicting the zero-impact master-switch behavior and the updated `TestOptions_MasterOffDisablesPillars` expectation. With the existing six base options and four disable options, this unconditional append makes `off.options()` contain 11 options rather than the asserted 10; enabling user baggage would add another. Gate both baggage options on `c.Enabled`.

```

</details>

pkg/ginutil/middleware.go
```diff
@@ -34,7 +34,7 @@ func CORS() gin.HandlerFunc {
 	return func(c *gin.Context) {
 		c.Header("Access-Control-Allow-Origin", "*")
 		c.Header("Access-Control-Allow-Methods", "GET, POST, PATCH, DELETE, OPTIONS")
-		c.Header("Access-Control-Allow-Headers", "Content-Type, Authorization, X-Request-ID, traceparent, tracestate, baggage")
+		c.Header("Access-Control-Allow-Headers", "Content-Type, Authorization, X-Request-ID, traceparent, tracestate")
```
julianshen: _🏷️ protocol-compatibility_ | _🟠 Warning_ | _`distributed-system`_

**Removing baggage breaks existing browser clients during rollout.**

A browser client using the previously advertised CORS contract can continue sending a `baggage` header while servers roll to this version. Its preflight then receives a response that omits `baggage`, so the browser blocks the actual API request before the new TraceContext-only propagator can safely ignore the header. This can make otherwise valid requests fail across all services using `ginutil.CORS`. Keep the header CORS-compatible while rejecting its contents through `obs.PublicIngressPropagator()`, and update the corresponding test expectation.

<details>

> ‼️ **IMPORTANT**

```suggestion
		c.Header("Access-Control-Allow-Headers", "Content-Type, Authorization, X-Request-ID, traceparent, tracestate, baggage")
```

</details>

<details>

```
In `pkg/ginutil/middleware.go` around line 37: A browser client using the previously advertised CORS contract can continue sending a `baggage` header while servers roll to this version. Its preflight then receives a response that omits `baggage`, so the browser blocks the actual API request before the new TraceContext-only propagator can safely ignore the header. This can make otherwise valid requests fail across all services using `ginutil.CORS`. Keep the header CORS-compatible while rejecting its contents through `obs.PublicIngressPropagator()`, and update the corresponding test expectation.

```

</details>

pkg/obs/obs.go
flywindy: Not applying this — the options are already gated, just by an early return rather than an `if c.Enabled` block.

`options()` returns inside the `if !c.Enabled` branch (`return append(opts, WithTraceEnabled(false), …)`), so line 126 is unreachable when the master switch is off. The counts the finding predicts don't occur: `off.options()` is 6 base + 4 pillar disables = **10**, and `on.options()` is 6 base + 1 `WithBaggageAttributes` = **7** — which is exactly what `TestOptions_MasterOffDisablesPillars` asserts, and that test passes.

Wrapping the same lines in `if c.Enabled { … }` after an early return would be dead structure, so leaving as is.

---

pkg/ginutil/middleware.go
flywindy: Agreed and fixed in 616bd59. `Access-Control-Allow-Headers` advertises `baggage` again, so a browser still sending it under the previous contract clears preflight during rollout; the contents are still rejected at extraction by `obs.PublicIngressPropagator()`, not by CORS. `TestCORS_PreflightOptions_AllowsTracePropagationHeaders` now asserts `Contains` with that reasoning in a comment.

---

pkg/obs/obs.go
julianshen: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**The master-off path still appends baggage options.**

These options are appended even when `O11Y_ENABLED=false`. Consequently, `off.options()` now contains 11 options rather than the 10 asserted by `TestOptions_MasterOffDisablesPillars`, and enabling `O11Y_USER_BAGGAGE_ENABLED` while the master switch is off appends another option. Gate both baggage options on `c.Enabled`, consistent with the master-switch contract and the updated test comment.

<details>

> ‼️ **IMPORTANT**

```suggestion
	if c.Enabled {
		opts = append(opts, o11y.WithBaggageAttributes(RoomIDKey, SiteIDKey))
		if c.UserBaggageEnabled {
			opts = append(opts, o11y.WithUserBaggage())
		}
	}
```

</details>

<details>

```
In `pkg/obs/obs.go` around line 126: These options are appended even when `O11Y_ENABLED=false`. Consequently, `off.options()` now contains 11 options rather than the 10 asserted by `TestOptions_MasterOffDisablesPillars`, and enabling `O11Y_USER_BAGGAGE_ENABLED` while the master switch is off appends another option. Gate both baggage options on `c.Enabled`, consistent with the master-switch contract and the updated test comment.

```

</details>

pkg/natsrouter/middleware.go
```diff
@@ -81,6 +82,24 @@ func RequireRequestID() HandlerFunc {
...
 }
 
+// traceIdentity enriches the active NATS entry span and all downstream spans
+// and contextual logs with trusted route identity. Router.addRoute installs it
+// for every router, including callers that use New with a hand-built middleware
+// stack. Params were extracted before middleware runs; account has also been
+// decoded from its NATS-safe bot transport form by route.extractParams.
+func traceIdentity() HandlerFunc {
+	return func(c *Context) {
+		ctx := obs.ContextWithIdentity(
```
julianshen: _🔒 security_ | _🟠 Warning_ | _`tgd-review`_

**Client-controlled NATS baggage remains trusted when params are absent.**

NATS routes include browser-facing `chat.user...` subjects, but this middleware starts from the baggage extracted by the composite NATS propagator. `ContextWithIdentity` only replaces non-empty route values, so a client can forge managed attributes absent from a route—for example, `chat.site.id` survives on the new test route because `site-1` is static rather than a `{siteID}` parameter. The forged value may already have been materialized on the entry span before this middleware runs. Treat externally reachable NATS ingress like public HTTP ingress: reject or clear managed baggage before span start, or explicitly classify public routes and clear all managed keys before applying trusted subject/payload identity.

<details>

```
In `pkg/natsrouter/middleware.go` around line 92: NATS routes include browser-facing `chat.user...` subjects, but this middleware starts from the baggage extracted by the composite NATS propagator. `ContextWithIdentity` only replaces non-empty route values, so a client can forge managed attributes absent from a route—for example, `chat.site.id` survives on the new test route because `site-1` is static rather than a `{siteID}` parameter. The forged value may already have been materialized on the entry span before this middleware runs. Treat externally reachable NATS ingress like public HTTP ingress: reject or clear managed baggage before span start, or explicitly classify public routes and clear all managed keys before applying trusted subject/payload identity.

```

</details>

docs/specs/o11y/o11y-trace-design.md
```diff
@@ -102,8 +102,29 @@ actual **frontend** span names follow `nats <operation> <subject>` —
 Tempo trace list is legible at a glance
 instead of a wall of `nats.request`. The subject is *also* carried on the
 `messaging.destination.name` attribute (with `chat.request_id` on requests), so
-you can filter by span name **or** by attribute. Backend consumer spans are
-named `process <subject>` by `otelnats` from the subject directly.
+you can filter by span name **or** by attribute.
+
+Backend span names are owned entirely by `otelnats` and follow the semconv
+v1.39.0 messaging shape `{operation} {destination}` (otel-nats v0.9.1, shipped
+with o11y v0.11.0):
```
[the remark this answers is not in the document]
flywindy: Fixed in 9b0ece8 — the §9 ground-truth citation now reads `o11y` v0.11.0 / `akira-core/otel-nats` v0.9.1, matching the pinned versions.

---

docs/specs/o11y/o11y-trace-design.md
```diff
@@ -166,22 +187,22 @@ flowchart TB
...
+  meta+members (+ `mongo roomkeys` if E2E) → N× `publish chat.room.{r}.event.*`.
 - **notification-worker** (1 trace, linked): `process` → `mongo`/`valkey` →
-  `send push.{s}.send` → links into the push service's trace.
+  `publish push.{s}.send` → links into the push service's trace.
 - **search-sync-worker** (linked): JetStream `Fetch` consumer span → `search-sync bulk flush` (links to every source message in the batch) → Elasticsearch bulk spans.
 
 **Total for one message ≈ 6 traces** — the **browser** publish span is its own
 trace (linked, not parent, into the gatekeeper), plus gatekeeper +
 message-worker + broadcast-worker + notification-worker + search-sync; the push
 service adds a 7th linked off notification. They are
```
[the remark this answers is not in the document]
flywindy: Fixed in 9b0ece8 — now "linked from notification".

---

docs/specs/o11y/o11y-trace-design.md
```diff
@@ -303,7 +324,7 @@ The DM-specific piece is **first-time room provisioning**: `user-service`'s
 
 - `〔Trace: user-service〕` `process …request…dm` → `mongo` → `send
   …room.{site}.create.dm` (PRODUCER) + reply to client.
-- `〔Trace: room-worker〕` (linked to the `send …create.dm`) `process` →
+- `〔Trace: room-worker〕` (linked to the `publish …create.dm`) `process` →
```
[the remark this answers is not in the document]
flywindy: Fixed in 9b0ece8 — scenario E's `publish …room.{site}.create.dm`, scenario G's `publish …canonical.edited`, and the §9 checklist's "link to the upstream `publish` span" all updated. The remaining `send` occurrences are deliberate: the pre-v0.11.0 names in the migration note, and `msg.send` / "message send", which are subject tokens and prose, not span names.

---

pkg/obs/obs.go
flywindy: Same finding as #discussion_r3764623802 — still a false positive, and unchanged since that reply.

`options()` returns inside the `if !c.Enabled` branch, so line 126 is unreachable with the master switch off; `off.options()` is 6 base + 4 pillar disables = **10** and `on.options()` is 6 + 1 = **7**, exactly what `TestOptions_MasterOffDisablesPillars` asserts. `O11Y_USER_BAGGAGE_ENABLED=true` with `O11Y_ENABLED=false` also cannot append anything, for the same reason — and `initSDK` additionally stores `userBaggageEnabled = cfg.Enabled && cfg.UserBaggageEnabled`.

---

julianshen: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**Baggage options bypass the observability master switch.**

These options are appended even when `c.Enabled` is false. This makes `TestOptions_MasterOffDisablesPillars` return 11 options instead of its expected 10 and can configure user baggage despite `O11Y_ENABLED=false`. Guard both baggage options with the master switch.

<details>

> ‼️ **IMPORTANT**

```suggestion
	if c.Enabled {
		opts = append(opts, o11y.WithBaggageAttributes(RoomIDKey, SiteIDKey))
		if c.UserBaggageEnabled {
			opts = append(opts, o11y.WithUserBaggage())
		}
	}
```

</details>

<details>

```
In `pkg/obs/obs.go` around line 126: These options are appended even when `c.Enabled` is false. This makes `TestOptions_MasterOffDisablesPillars` return 11 options instead of its expected 10 and can configure user baggage despite `O11Y_ENABLED=false`. Guard both baggage options with the master switch.

```

</details>

pkg/natsrouter/middleware.go
flywindy: Confirmed and fixed in 6b5b6b8. I reproduced it first: a client-facing route with a static site token, caller baggage `chat.site.id=forged-site`, and the forged value reached the handler — `TestRouter_DropsForgedIdentityOnClientFacingRoutes` fails without the fix.

What landed:

- **`obs.ContextWithPublicIdentity`** — drops every key in the new `obs.ManagedBaggageKeys()` before applying trusted values, so a key the boundary has no replacement for is removed rather than inherited. It reuses the existing `withoutBaggageAttribute` path, which also zeroes the entry-span attributes the OnStart processor already materialized (the second half of your finding — clearing baggage alone would have left those on the span).
- **Single source of truth** — `WithBaggageAttributes` registration now derives from `managedBaggageKeys`, so a future key cannot be materialized but left out of the clear list. Guarded by a test.
- **`pkg/natsrouter`** — `traceIdentity` uses the public variant on `chat.user.>` (`subject.IsClientFacing`). Internal subjects keep the preserve-upstream behavior on purpose: a federated event must stay attributed to its originating site, not the site that happens to process it.
- **Site label recovered, not just dropped** — `natsrouter.WithSiteID(cfg.SiteID)` labels client-facing routes whose subject pins the site to a static token, from the service's own config instead of the wire. Wired into all 10 router services.
- **`message-gatekeeper`** — the invalid-subject early return is the one client-reachable branch that returns before the subject yields an identity, so it never called `ContextWithIdentity` at all; it now drops the caller's baggage there.

One residual risk, documented in `docs/specs/o11y/o11y-integration.md`: the entry span starts before any middleware, so a forged value is still visible to a **head sampling decision**. Closing that would need a TraceContext-only propagator on a separate client-facing connection — noted as a follow-up rather than done here. A head sampler must not key on a managed attribute.

Checked the other `ContextWithIdentity` call sites while I was in there: the HTTP services all use `PublicIngressPropagator` (no inbound baggage at all), and the canonical-stream workers pass all three values non-empty, so both were already covered.

---

pkg/obs/obs.go
```diff
@@ -172,3 +231,86 @@ func initSDK(ctx context.Context, minLevel *slog.Level, wrap func(slog.Handler)
...
+// the forged values before any handler ran.
+func ContextWithPublicIdentity(ctx context.Context, account, roomID, siteID string) context.Context {
+	if !contextAttributesEnabled.Load() {
+		return ctx
+	}
+	for _, key := range managedBaggageKeys {
+		ctx = withoutBaggageAttribute(ctx, key)
+	}
+	return ContextWithIdentity(ctx, account, roomID, siteID)
+}
```
julianshen: _🔒 security_ | _🟠 Warning_ | _`nats`_

**The master switch can leave client NATS baggage unsanitized.**

`natsutil.Connect` now allows `OTEL_NATS_TRACING_ENABLED` or the relay to enable NATS tracing even when `O11Y_ENABLED=false`. In that configuration, the NATS wrapper can extract baggage from a client publish on `chat.user.*`, but this early return skips the public-boundary cleanup. Forged user, room, or site baggage can therefore be injected into replies and subsequent internal publishes, where an observability-enabled downstream service treats the internal hop as trusted. Always remove managed baggage at public NATS ingress; only gate the addition and materialization of trusted identity on `contextAttributesEnabled`.

<details>

> ‼️ **IMPORTANT**

```suggestion
	for _, key := range managedBaggageKeys {
		ctx = withoutBaggageAttribute(ctx, key)
	}
	if !contextAttributesEnabled.Load() {
		return ctx
	}
```

</details>

<details>

```
In `pkg/obs/obs.go` around line 283: `natsutil.Connect` now allows `OTEL_NATS_TRACING_ENABLED` or the relay to enable NATS tracing even when `O11Y_ENABLED=false`. In that configuration, the NATS wrapper can extract baggage from a client publish on `chat.user.*`, but this early return skips the public-boundary cleanup. Forged user, room, or site baggage can therefore be injected into replies and subsequent internal publishes, where an observability-enabled downstream service treats the internal hop as trusted. Always remove managed baggage at public NATS ingress; only gate the addition and materialization of trusted identity on `contextAttributesEnabled`.

```

</details>

pkg/obs/obs.go
```diff
@@ -172,3 +231,86 @@ func initSDK(ctx context.Context, minLevel *slog.Level, wrap func(slog.Handler)
...
+// baggage at extraction, but a NATS propagator is per-connection, so the drop
+// happens on the first hop of the handler chain. Dropping also zeroes the
+// entry span's attributes, which the SDK's OnStart processor materialized from
+// the forged values before any handler ran.
+func ContextWithPublicIdentity(ctx context.Context, account, roomID, siteID string) context.Context {
+	if !contextAttributesEnabled.Load() {
+		return ctx
+	}
+	for _, key := range managedBaggageKeys {
+		ctx = withoutBaggageAttribute(ctx, key)
```
julianshen: _🔒 security_ | _🟠 Warning_ | _`tgd-review`_

**Public NATS sanitization is skipped when tracing is overridden on.**

Connect now permits OTEL_NATS_TRACING_ENABLED or the relay to enable NATS propagation even when O11Y_ENABLED is false. In that configuration contextAttributesEnabled remains false, so this early return preserves caller-controlled managed baggage. A downstream service with observability enabled can then trust and export the forged identity after an internal hop. Public-ingress baggage must always be removed; only adding trusted replacement attributes should depend on observability being enabled.

<details>

> ‼️ **IMPORTANT**

```suggestion
	for _, key := range managedBaggageKeys {
```

</details>

<details>

```
In `pkg/obs/obs.go` around line 282: Connect now permits OTEL_NATS_TRACING_ENABLED or the relay to enable NATS propagation even when O11Y_ENABLED is false. In that configuration contextAttributesEnabled remains false, so this early return preserves caller-controlled managed baggage. A downstream service with observability enabled can then trust and export the forged identity after an internal hop. Public-ingress baggage must always be removed; only adding trusted replacement attributes should depend on observability being enabled.

```

</details>

message-gatekeeper/handler.go
```diff
@@ -110,6 +115,8 @@ func (h *Handler) HandleJetStreamMsg(ctx context.Context, msg jetstream.Msg) {
 		return
 	}
 
+	requester := subject.DecodeAccount(account)
+	ctx = obs.ContextWithIdentity(ctx, requester, roomID, siteID)
```
julianshen: _🔒 security_ | _🟠 Warning_ | _`tgd-review`_

**Valid gatekeeper messages use the non-public identity path.**

This subject is explicitly client-facing, but the valid path calls ContextWithIdentity rather than ContextWithPublicIdentity. When the NATS tracing environment or relay overrides a disabled SDK default, ContextWithIdentity does nothing and forged baggage can propagate through the canonical publish and reply. Use the public-boundary helper so all managed caller baggage is cleared before trusted subject values are applied.

<details>

> ‼️ **IMPORTANT**

```suggestion
	ctx = obs.ContextWithPublicIdentity(ctx, requester, roomID, siteID)
```

</details>

<details>

```
In `message-gatekeeper/handler.go` around line 119: This subject is explicitly client-facing, but the valid path calls ContextWithIdentity rather than ContextWithPublicIdentity. When the NATS tracing environment or relay overrides a disabled SDK default, ContextWithIdentity does nothing and forged baggage can propagate through the canonical publish and reply. Use the public-boundary helper so all managed caller baggage is cleared before trusted subject values are applied.

```

</details>

message-gatekeeper/handler_test.go
```diff
@@ -2266,3 +2270,112 @@ func TestHandler_processMessage_NonThreadMessage_NoParentFetch(t *testing.T) {
...
+	t.Setenv("O11Y_TRACE_ENABLED", "false")
+	t.Setenv("O11Y_METRICS_ENABLED", "false")
+	t.Setenv("O11Y_LOG_ENABLED", "false")
+	t.Setenv("O11Y_USER_BAGGAGE_ENABLED", "true")
+	t.Setenv("OTEL_SERVICE_NAME", "message-gatekeeper-identity-test")
+	t.Setenv("OTEL_EXPORTER_PROMETHEUS_PORT", "0")
+
+	_, shutdown, err := obs.Init(context.Background())
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = shutdown(context.Background()) })
```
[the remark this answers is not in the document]
flywindy: System.Collections.Hashtable.body

pkg/obs/obs.go
flywindy: Fixed in `cf059677`. `ContextWithPublicIdentity` now removes every managed baggage key before consulting `contextAttributesEnabled`; master-off only suppresses trusted identity emission, not trust-boundary sanitization. `TestContextWithPublicIdentity_MasterOffStillDropsManagedKeys` reproduces the env/relay propagation scenario and proves managed keys are removed while unrelated baggage is preserved.

pkg/obs/obs.go
flywindy: Fixed in `cf059677` together with the duplicate security finding in the adjacent thread. Public sanitization now always runs before the observability toggle check. The new master-off regression test failed with all three forged values before the fix and passes after it; trusted identity enrichment remains covered by the existing master-on tests.

message-gatekeeper/handler.go
flywindy: Fixed in `cf059677`. The valid message-gatekeeper path now calls `obs.ContextWithPublicIdentity`. `TestHandleJetStreamMsg_ValidSubject_MasterOffDropsForgedIdentity` proves forged user/room/site baggage reaches neither the canonical publish nor client reply when o11y emission is off, while the existing `TestHandleJetStreamMsg_AttachesIdentityToContext` still proves trusted decoded `user.name`, room, and site are added when enabled.

pkg/msgraph/msgraph.go
```diff
@@ -363,7 +373,7 @@ func (g *graphClient) accessToken(ctx context.Context) (string, error) {
 	form.Set("client_secret", g.cfg.ClientSecret)
 	form.Set("scope", graphScope)
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, g.tokenURL, strings.NewReader(form.Encode()))
+	req, err := newExternalRequestWithContext(ctx, http.MethodPost, g.tokenURL, strings.NewReader(form.Encode()))
```
mliu33, line 376: Don't log the body as this contains client secret

pkg/msgraph/presence.go
```diff
@@ -87,7 +87,7 @@ func (p *presenceClient) accessToken(ctx context.Context) (string, error) {
 	form.Set("username", p.creds.Username)
 	form.Set("password", p.creds.Password)
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, p.tokenURL, strings.NewReader(form.Encode()))
+	req, err := newExternalRequestWithContext(ctx, http.MethodPost, p.tokenURL, strings.NewReader(form.Encode()))
```
mliu33, line 90: Is this function going to log the body ?

pkg/obs/obs.go
```diff
@@ -172,3 +231,86 @@ func initSDK(ctx context.Context, minLevel *slog.Level, wrap func(slog.Handler)
...
+// otherwise leave in place — so an identifier the subject does not carry (a
+// static site token, say) cannot be supplied by the caller instead.
+//
+// This is the NATS counterpart of PublicIngressPropagator: HTTP can refuse
+// baggage at extraction, but a NATS propagator is per-connection, so the drop
+// happens on the first hop of the handler chain. Dropping also zeroes the
+// entry span's attributes, which the SDK's OnStart processor materialized from
+// the forged values before any handler ran.
+func ContextWithPublicIdentity(ctx context.Context, account, roomID, siteID string) context.Context {
+	for _, key := range managedBaggageKeys {
```
mliu33, line 281: This is redundant because contextWithBaggageAttribute already drops the key first before adding the given value

pkg/obs/obs.go
```diff
@@ -172,3 +231,86 @@ func initSDK(ctx context.Context, minLevel *slog.Level, wrap func(slog.Handler)
...
+			next, err := o11y.ContextWithUser(ctx, account)
+			if err != nil {
+				slog.WarnContext(ctx, "telemetry identity attribute omitted", "attribute", o11y.UserNameKey, "error", err)
+			} else {
+				ctx = next
+				o11y.SetUser(ctx, account)
+			}
+		}
+	}
+	ctx = contextWithBaggageAttribute(ctx, RoomIDKey, roomID)
```
mliu33, line 263: I think we should be able to set them in batch ?

mliu33 (approved):
Excellent work, thanks! Just some minor suggestions~

pkg/obs/obs.go
flywindy: Half right, and the half that isn't is the reason the loop exists — so it's now scoped rather than removed. Addressed in #259.

`contextWithBaggageAttribute` drops the key first **only when it has a value to set**:

```go
if value == "" {
    return ctx        // returns before withoutBaggageAttribute
}
ctx = withoutBaggageAttribute(ctx, key)
```

So:

- **Key with a trusted value** — you're right, it was dropped twice.
- **Key with none** (a subject pinning the site to a static token leaves no `{siteID}` param) — nothing drops it, and the caller's forged value survives. That's the vector #236 fixed.
- **`O11Y_ENABLED=false`** — `ContextWithIdentity` returns at its first line, so the loop is the only sanitization. NATS tracing can still be forced on by `OTEL_NATS_TRACING_ENABLED` or the flag relay, so this path matters.

`account` behaves the same way: its drop is inside `if account != ""`.

#259 pre-clears only the keys no trusted value will overwrite when emission is on, and still clears all of them when it's off. `TestUnsuppliedManagedKeys` pins that split against `ManagedBaggageKeys()`.

---

pkg/obs/obs.go
flywindy: Agreed — partly done in #259, with one half blocked on the SDK.

**Batched:** `withoutBaggageAttributes` now takes keys variadically and does one `o11y.ContextWithoutBaggageValues` plus one span write for the whole set instead of one of each per key (and skips keys the baggage doesn't carry). `contextWithBaggageAttributes` materializes the accepted values in a single `SetAttributes`. Combined with dropping the redundant clear from the adjacent thread: up to 6 baggage rebuilds → at most 3, up to 6 span writes → at most 3, per call on the public path.

**Not batched:** the baggage *set*. o11y only exposes `ContextWithBaggageValue(ctx, key, value)` — single member per call — and the only way to batch it here would be to build `baggage.New(members...)` ourselves, which gives up the SDK's validation (size limits, reserved-key collision checks). Better raised upstream as a `ContextWithBaggageValues` request than worked around in `pkg/obs`; I've noted it in the code comment.

Worth stating the scale honestly: baggage carries at most 3 members, so a rebuild is a small map copy. This runs once per NATS request and once per message in `message-gatekeeper`, so it is on the message hot path, but it was never going to be the bottleneck there — the change is worth making for the code shape more than the nanoseconds.

---

pkg/msgraph/msgraph.go
flywindy: Checked the whole path — the body is not logged anywhere. But you had to ask, and that's worth fixing, so #259 makes it explicit and enforces it.

`newExternalRequestWithContext` is one line: it strips baggage and hands `body` to `http.NewRequestWithContext`. It never reads or logs it. Nor does anything else here:

- The only `slog` call in `pkg/msgraph` is a throttling warning in `chats.go`.
- The token error paths surface the status and the OAuth error code only — `fmt.Errorf("token endpoint returned status %d: %s", resp.StatusCode, tr.Error)`. Even the unmarshal-failure branch leaves `body` out.
- The default client is a plain `http.Client`; if a caller injects `otelhttp` via `WithHTTPClient`, that records method/URL/status/size, never bodies — and the secret is in the POST form, not the URL.

#259 adds two things: the doc comment now states that `body` is passed through untouched and never logged (naming `client_secret` as the reason), and `.semgrep/msgraph-secrets.yml` makes it a CI gate — a form value, an `io.ReadAll` body, or a stringified body reaching `slog` or `fmt.Errorf` in this package fails the `sast` job. Verified both directions: clean on the package as it stands, and it catches all four shapes when introduced deliberately.

---

pkg/msgraph/presence.go
flywindy: No — same helper as the other thread, and it doesn't log the body. But **this call site is the higher-stakes one**, so thanks for pointing at it separately.

The ROPC form here carries more than the client secret:

```go
form.Set("client_secret", p.cfg.ClientSecret)
form.Set("username", p.creds.Username)
form.Set("password", p.creds.Password)
```

Nothing logs it today (`ropcToken`'s error path surfaces only the status and `tr.Error`, same as the client-credentials path), but this is a spot where one debug line becomes an incident. That's why #259 doesn't just document it — `.semgrep/msgraph-secrets.yml` fails the `sast` job if a form value or a response body reaches `slog`/`fmt.Errorf` anywhere in `pkg/msgraph`, so the next person adding a debug line here gets stopped by CI rather than by review.

---

PR #238, opened 2026-08-11
Valkey timeout and startup hardening

hmchangw:
Closes the two mechanical traps that let a Valkey outage escalate into a chat outage. Neither is a logic bug in the cache consumers — both live in how the client is constructed.

**The timeout trap.** The client was built with only `Addrs` and `Password`, so go-redis defaults applied and `ContextTimeoutEnabled` was `false` — a caller's context deadline did not bound socket reads at all. Against a Valkey that *blackholes* packets (the common degraded mode, rather than one that refuses connections), every hot-path read stalled for seconds before falling back. The code was fail-open; the network behaviour was fail-slow.

**The startup trap.** Seven services treated a failed startup `PING` as fatal. Running pods survived an outage, but any rollout, autoscale, node drain, or OOM-kill during one would put `message-gatekeeper` and `broadcast-worker` into `CrashLoopBackOff` — and the crashloop outlived the outage.

## Review order

`pkg/valkeyutil` is 12 of 38 files and everything else depends on its contract, so read it first. Each commit is one self-contained change and builds independently (verified), so commit-by-commit works.

| Commit | Scope |
|---|---|
| `d2d849a` | Design spec + implementation plan (working notes — see *Open questions*) |
| `dae29ef` | `pkg/valkeyutil`: profiles, breaker, non-fatal connect, `IncrEx` fix, `LogDegraded` |
| `b4fae2d` | `botplatform-service`: fail-open posture |
| `4ff5a3b` | `user-presence-service`: lazy connect + `StoreProfile` |
| `6904300` | Five cache services lazy; seeder keeps failing fast |
| `f049618` | Warn-flood suppression at the cache consumers |
| `eadb2e7` | `docs/health-probes.md` |
| `dc8f5a2` | Breaker error classification corrected — see below |

## Behaviour changes

These are the things that change in production, and the reason to read the PR rather than the diff:

1. **The circuit breaker is on by default** for existing `ConnectCluster` callers, without them opting in. Deliberate; `WithoutCircuitBreaker()` opts out.
2. **`botplatform-service` fails open.** During an outage bot requests are admitted **unthrottled and without duplicate suppression**. Bots are critical, so a lost ceiling beats a dead bot — this is a conscious trade, not an oversight. `bot_control_bypassed_total{control}` is the only durable signal that protection was off; alert on it.
3. **Six services no longer fail startup** on an unreachable Valkey. `tools/seed-sample-data` still does, which is correct for a one-shot CLI.
4. **`user-presence-service` gains o11y hooks.** It previously built a raw client outside `valkeyutil` and was the only Valkey client in the fleet emitting no command spans or pool metrics.
5. **`IncrEx` changes wire behaviour** from `INCR` + conditional `EXPIRE` to a pipelined `INCR` + `EXPIRE NX`. This fixes a latent bug: a landed `INCR` whose `EXPIRE` failed left a counter with no expiry, so the window never reset and the caller was throttled *permanently* once past the ceiling, long after Valkey recovered.
6. **Per-call warn logs drop to Debug while the circuit is open.** A live failure still logs at Warn.

## Design decisions worth a reviewer's attention

**`ContextTimeoutEnabled: true`** is the load-bearing field. Without it the whole profile mechanism is decorative.

**Breaker error classification uses all three of gobreaker's outcomes** (corrected in `dc8f5a2`; the earlier commits got this wrong and the reasoning is worth stating).

The original implementation forced every error into success-or-failure, on the stated premise that "gobreaker offers no neutral outcome, so each ambiguous error is scored by which wrong answer costs less." That premise is **false** for the `gobreaker/v2 v2.4.0` this PR pins: `Settings.IsExcluded` drops a call from the accounting entirely, is consulted before `IsSuccessful`, and its own documentation offers *ignoring context cancellations* as the example use case.

Both forced scores were live defects, failing in opposite directions:

- **Cancellation was scored a failure.** A cancelled caller is no evidence Valkey is unhealthy, but `ReadyToTrip` fires at five consecutive failures — so a burst opened the breaker on a healthy cache and stampeded Mongo, the self-inflicted outage the classifier exists to prevent. Cancellations arrive in exactly that shape, since one slow sibling makes an errgroup cancel all the others at once.
- **Pool timeout was scored a success**, which clears `ConsecutiveFailures`. A genuine outage interleaved with pool timeouts kept the count near zero and the breaker never tripped — the same "held closed through a real outage" failure the cancellation reasoning was trying to avoid, reached from the other side.

Both are now **excluded**: they move neither counter, so they can neither trip the breaker nor hold it closed. `context.DeadlineExceeded` is deliberately **left in the accounting as a failure** — with `ContextTimeoutEnabled` bounding socket reads, a deadline is precisely how a blackholing Valkey surfaces, which is the degraded mode this PR targets. Excluding it too would blind the breaker to the outage it was built for.

**Timeout values are code constants, not environment variables** — a departure from the repo's usual env-config rule. Exposing them adds seven services' worth of surface nobody will set correctly under incident pressure, and a wrong value silently reintroduces the trap. Flagged because a reviewer should weigh in.

**Profiles add retries.** The go-redis *cluster* default is `MaxRetries: -1` (no retries), so worst-case per-command latency is `(MaxRetries+1) × ReadTimeout` plus backoff, not `ReadTimeout` alone.

## Known limitations

- **One breaker spans the whole cluster.** A single-shard failover trips it and short-circuits keys on healthy shards too. Per-shard breakers need slot mapping; not attempted here.
- **The breaker decorates the `Client` facade**, so `user-presence-service`'s raw client (Lua, sorted sets) is not behind one and pays its `StoreProfile` budget per call instead.
- **Room-metadata invalidation stays best-effort**, so a rename during an outage can serve stale metadata until `ROOM_META_L2_TTL` (default 15m). The write itself still lands in Mongo.

## Verification

- CI was green on `eadb2e7`: `test`, `lint`, `sast`, and all 50 `test-integration` jobs. `dc8f5a2` is re-running.
- Locally on `dc8f5a2`: `make lint` 0 issues, `make test` green (109 packages), `go build ./...` and `go vet -tags=integration` clean, `gosec` clean. `govulncheck` and `semgrep` could not run in the dev sandbox (their rule/vuln feeds are network-fetched and blocked there) — CI covers all three.
- `pkg/valkeyutil` unit coverage is **71.4%**. The remainder is the `clusterClient` store methods, which need a live server and are covered by the integration suite per CLAUDE.md §4 — not a gap, but stated plainly rather than rounded up. `breaker.go` itself is at 100% for both classifier predicates.
- Every commit builds and vets independently; the branch is bisectable.

## Open questions for the reviewer

1. **`github.com/sony/gobreaker/v2` is a new dependency.** CLAUDE.md requires sign-off before adding one. Please confirm.
2. **Should the spec and plan ship?** `docs/superpowers/**` is the bulk of the diff and they are working notes rather than reference docs. Say the word and I will drop them, cutting the review surface substantially. Note the plan still contains the superseded `isSuccessful` body verbatim as a historical transcript; if these files ship rather than being dropped, that block wants a correction too.
3. **Timeout constants vs env vars** (see above).

https://claude.ai/code/session_01GtKXocLa5CCVCJdasafSK2

### DIFF
--- a/botplatform-service/middleware.go
+++ b/botplatform-service/middleware.go
@@ -104,6 +104,10 @@ type incrExClient interface {
 
 // botRateLimit enforces per-caller then global fixed-window counters (60s each); 0 disables.
 // Per-caller first so a rejected caller doesn't consume the global budget.
+//
+// Fails OPEN: an unreachable Valkey admits the request instead of rejecting it. Bots are
+// critical, so a lost ceiling beats a dead bot — during an outage nothing throttles a
+// looping caller, which is the accepted cost of staying up.
 func botRateLimit(client incrExClient, perCaller, perGlobal int) gin.HandlerFunc {
 	const window = time.Minute
 
@@ -125,8 +129,8 @@ func botRateLimit(client incrExClient, perCaller, perGlobal int) gin.HandlerFunc
 		if perCaller > 0 {
 			n, err := client.IncrEx(ctx, "botrl:caller:"+pr.UserID, window)
 			if err != nil {
-				errhttp.Write(ctx, c, errcode.Internal("bot rate limit caller", errcode.WithCause(err)))
-				c.Abort()
+				bypassControl(ctx, "rate limit", controlRateLimitCaller, err, "counter", "caller")
+				c.Next()
 				return
 			}
 			if n > int64(perCaller) {
@@ -140,8 +144,8 @@ func botRateLimit(client incrExClient, perCaller, perGlobal int) gin.HandlerFunc
 		if perGlobal > 0 {
 			n, err := client.IncrEx(ctx, "botrl:global", window)
 			if err != nil {
-				errhttp.Write(ctx, c, errcode.Internal("bot rate limit global", errcode.WithCause(err)))
-				c.Abort()
+				bypassControl(ctx, "rate limit", controlRateLimitGlobal, err, "counter", "global")
+				c.Next()
 				return
 			}
 			if n > int64(perGlobal) {
--- a/docs/superpowers/specs/2026-08-11-valkey-hardening-design.md
+++ b/docs/superpowers/specs/2026-08-11-valkey-hardening-design.md
@@ -1,0 +1,244 @@
+# Valkey Timeout and Startup Hardening
+
+An audit of the Valkey surface asked whether a Valkey outage brings the system
+down. It does not — the cache consumers are all written fail-open and degrade to
+Mongo or Elasticsearch. But two mechanical traps sit underneath that correctness
+and can convert a cache incident into a chat outage. This spec closes both.
+
+Neither trap is a logic bug in the consumers. Both live in how the client is
+constructed.
+
+## The two traps
+
+**The timeout trap.** `valkeyutil.ConnectCluster` builds its `redis.ClusterOptions`
+with only `Addrs` and `Password` (`pkg/valkeyutil/valkey.go:40`), and
+`presencestore.NewValkeyStore` does the same (`user-presence-service/presencestore/store.go:193`).
+Every timeout is therefore a go-redis default: 5s dial, 3s read, 3 retries. Against
+a Valkey that *refuses* connections this is harmless — the dial fails instantly. Against
+one that **blackholes** packets, which is the more common degraded mode, every
+hot-path GET burns seconds before the Mongo fallback runs. Per message. On the
+JetStream consumers that means throughput collapse, ack-pending saturation, and a
+growing stream backlog.
+
+The code is fail-open. The network behavior is fail-slow. Fail-open only works if
+it fails fast.
+
+**The startup trap.** `ConnectCluster` gates on a `PING` and returns an error on
+failure (`pkg/valkeyutil/valkey.go:52`); every caller treats that as fatal. Seven
+services exit: `notification-worker/main.go:170` and `search-service/main.go:156`
+unconditionally, and `message-gatekeeper/main.go:110`, `broadcast-worker/main.go:128`,
+`room-worker/main.go:155`, `botplatform-service/main.go:84`,
+`user-presence-service/main.go:100` whenever `VALKEY_ADDRS` is set — which
+production sets.
+
+Running pods survive an outage, because liveness is process-up-only and readiness
+probes NATS alone (`docs/health-probes.md`). But any rollout, HPA scale-up, node
+drain, or OOM-kill during the outage puts the core message path —
+`message-gatekeeper` and `broadcast-worker` — into `CrashLoopBackOff`. A cache
+outage that happens to overlap a deploy becomes a chat outage.
+
+## Scope
+
+- Bounded, profile-driven timeouts on every Valkey client.
+- A circuit breaker so the degraded path costs approximately nothing.
+- Removal of the fatal startup `PING` gate across all seven services.
+- Two posture changes in `botplatform-service`, described below.
+
+### Out of scope
+
+- The fail-open behavior of the cache consumers (`roommetacache`, `roomsubcache`,
+  `search-service`). It is already correct; this spec only makes it fast.
+- Valkey topology, replication, and failover — an ops concern.
+- `user-presence-service`'s lack of a fallback store. Valkey is legitimately its
+  store of record; serving errors is the right failure mode, not a design defect.
+
+## Timeout profiles
+
+Two named profiles in `pkg/valkeyutil`, selected per call site.
+
+| Knob | `CacheProfile` | `StoreProfile` | go-redis default |
+|------|---------------|----------------|------------------|
+| `DialTimeout` | 1s | 1s | 5s |
+| `ReadTimeout` / `WriteTimeout` | 150ms | 500ms | 3s |
+| `MaxRetries` | 1 | 2 | 3 |
+| `PoolTimeout` | 250ms | 1s | ~`ReadTimeout`+1s |
+| `ContextTimeoutEnabled` | true | true | **false** |
+
+`CacheProfile` serves the five cache consumers. `StoreProfile` serves
+`user-presence-service`, where Valkey is the store of record rather than a cache —
+a 150ms ceiling on a Lua `EVAL` under load would manufacture failures that no
+fallback can absorb.
+
+`ContextTimeoutEnabled` is the least obvious and most important entry. With it
+false, which is today's state, a caller's `context.WithTimeout` does **not** bound
+the socket read. The 10s `fetchTimeout` guards in `pkg/roommetacache/roommetacache.go:34`
+and `notification-worker/members.go:18` are consequently the only ceiling on a
+hung Valkey call, and they are two orders of magnitude above the intended budget.
+
+These are code constants, not environment variables. CLAUDE.md §6 requires
+deployment configuration to come from the environment, but these are internal
+tuning constants: exposing them would add seven services' worth of surface that
+nobody will set correctly under incident pressure, and a wrong value reintroduces
+exactly the trap being closed.
+
+## Circuit breaker
+
+Bounded timeouts still cost roughly 300ms on every hot-path call during an
+outage. At message volume that is its own throughput problem, so the client
+short-circuits once Valkey is demonstrably down.
+
+`github.com/sony/gobreaker/v2`, pinned at `v2.4.0`. This is a new third-party
+dependency, approved during design in preference to a hand-rolled breaker.
+
+One `*gobreaker.CircuitBreaker[any]` per `Client` instance, shared across all five
+interface methods — reachability is a property of the connection, not of the
+command. Values box through `any`, which is free relative to a network round-trip.
+
+| Setting | Value | Reason |
+|---------|-------|--------|
+| `ReadyToTrip` | `ConsecutiveFailures >= 5` | Rides out a single blip; trips on a real outage. |
+| `Timeout` | 5s | Open → half-open cooldown. |
+| `MaxRequests` | 1 | One half-open probe, so recovery never stampedes. |
+| `Interval` | 0 | No auto-reset while closed; consecutive-failure semantics. |
+| `OnStateChange` | transition log + metric | See Observability. |
+
+### IsSuccessful is load-bearing
+
+gobreaker counts every returned error as a failure. `valkeyutil.ErrCacheMiss` is
+not a failure — it is the cache working correctly — and a cold or sparse keyspace
+would otherwise trip the breaker and disable Valkey for a workload that is
+behaving perfectly. `Settings.IsSuccessful` must therefore classify both
+`ErrCacheMiss` and `context.Canceled` as successes. Only transport errors count
+toward the trip threshold.
+
+This is the single detail most likely to be got wrong, and it fails in the
+direction of a self-inflicted outage. It gets a dedicated test.
+
+### Callers need almost no changes
+
+When open, the breaker surfaces `gobreaker.ErrOpenState` / `ErrTooManyRequests`,
+which `valkeyutil` maps to an exported `ErrUnavailable` alongside a new
+`valkeyutil.IsUnavailable(err)` helper.
+
+Every fail-open consumer already branches `errors.Is(err, ErrCacheMiss)` → treat
+as miss, everything else → log and fall back. `ErrUnavailable` lands in the
+fallback branch with no code change at all.
+
+The one genuine edit is log volume: a per-message WARN during an outage is a log
+flood. The breaker logs the closed→open and open→closed transitions once, and the
+three existing call-site warnings are gated on `!IsUnavailable(err)`:
+
+- `pkg/roommetacache/valkey.go:46`
+- `notification-worker/members.go:46`
+- `search-service/handler.go:227`
+
+## Startup: lazy connect
+
+The fix is smaller than the trap suggests. go-redis already dials lazily and
+self-heals per call, so nothing needs to retry in the background. The crashloop
+comes entirely from the startup `PING` being treated as fatal.
+
+`valkeyutil.ConnectClusterLazy` performs identical construction and
+instrumentation, demotes the `PING` to a non-fatal reachability log, and always
+returns a usable `Client`. `presencestore.NewValkeyStore` gets the same treatment.
+All seven call sites switch to it and drop their `os.Exit(1)` / error return.
+
+No background goroutine is introduced, so there is no termination path to get
+wrong (CLAUDE.md §3, concurrency).
+
+`ConnectCluster` is retained for `tools/seed-sample-data`, a one-shot CLI where
+failing fast on an unreachable Valkey is the correct behavior.
+
+## Per-service posture after the change
+
+| Service | Boot with Valkey down | Runtime behavior |
+|---------|----------------------|------------------|
+| `message-gatekeeper` | starts | breaker opens; room-meta reads go to Mongo |
+| `broadcast-worker` | starts | breaker opens; room-meta reads go to Mongo |
+| `room-worker` | starts | `BustMeta` no-ops; L2 TTL reconciles |
+| `notification-worker` | starts | member fan-out falls back to Mongo |
+| `search-service` | starts | restricted-rooms falls through to Elasticsearch |
+| `user-presence-service` | starts | `errcode` errors per request; self-heals |
+| `botplatform-service` | starts | see below |
+
+### botplatform-service
+
+Two Valkey-backed controls sit on the same requests, and they degrade
+differently because they protect different things.
+
+**Rate limiting is suspended.** It is a protective control — it guards the
+platform, not the data — so availability wins during an outage. `botRateLimit`
+(`botplatform-service/middleware.go:128`) gains a fail-open path: on a Valkey
+error it logs WARN, emits a metric, and calls `c.Next()` instead of aborting with
+`errcode.Internal`.
+
+**Idempotency splits by endpoint.** It is a correctness control: the sentinel is
+what stops a retrying bot from double-executing within the 60s bucket. A
+duplicate message is visible but recoverable; a duplicate room creation or member
+add is expensive and awkward to undo. So:
+
+| Endpoints | Posture on Valkey error |
+|-----------|------------------------|
+| `POST /api/v1/rooms/:roomID/messages`, `POST /api/v1/dms/:userID/messages` | fail **open** — proceed without the sentinel |
+| `POST /api/v1/rooms`, `.../members/add`, `.../members/remove` | fail **closed** — `errcode.Internal`, as today |
+
+`routes.go` already builds the two classes through separate closures
+(`botplatform-service/routes.go:27` for messages, `:30` for room management), so
+this is a `failOpen bool` parameter on `botIdempotency` rather than a
+restructure. `botRateLimit` at `:25` takes the same parameter, set `true`.
+
+A suspended control must never be silent — both fail-open paths log and emit the
+metric below.
+
+## Observability
+
+| Signal | Type | Purpose |
+|--------|------|---------|
+| `valkey_breaker_transitions{from,to}` | counter | Emitted from `OnStateChange`. |
+| `valkey_breaker_state` | gauge | Current state, for dashboards. |
+| `bot_control_bypassed{control="ratelimit"\|"idempotency"}` | counter | A bypassed control is an alertable condition — abuse protection is off. |
+
+Existing `cachemetrics` hit/miss/error recording is unchanged.
+
+## Testing
+
+TDD throughout, per CLAUDE.md §4. Red before green on every item.
+
+**Breaker.** Table-driven unit tests over closed → open → half-open → closed,
+threshold boundaries, and concurrent callers under `-race`. `IsSuccessful` gets
+its own table asserting that `ErrCacheMiss` and `context.Canceled` never trip the
+breaker while transport errors do.
+
+**Bounded latency.** A `net.Listen` that accepts connections and never responds —
+a deterministic blackhole. This is the failure mode that connection-refused
+testing misses entirely, and it is the one the timeout profiles exist to bound.
+Assert a `Get` returns within the profile budget.
+
+**Lazy connect.** Assert a usable `Client` comes back from an unreachable address
+and that the first call returns an error rather than panicking.
+
+**botplatform.** Extend the existing tables in `middleware_ratelimit_test.go` and
+`middleware_idempotency_test.go` with a Valkey-error case per posture, including
+the fail-closed room-management case.
+
+**Integration.** `testutil.SharedValkeyCluster` for the healthy-path regression,
+with `testutil.FlushValkey` registered in `t.Cleanup` per CLAUDE.md §4.
+
+Coverage floor is 80%; `pkg/valkeyutil` is shared infrastructure and targets 90%.
+
+## Risks
+
+**The profiles are tighter than production has ever run.** A p99 Valkey latency
+above 150ms would start failing cache reads that today succeed slowly. The
+consumers all fall back correctly, so the failure mode is extra Mongo load rather
+than errors — but the `cachemetrics` error series should be watched after rollout,
+and `CacheProfile` loosened if it proves tight.
+
+**Suspended rate limiting removes abuse protection during an outage**, at exactly
+the moment the platform is already degraded. This is the accepted cost of the
+availability posture chosen during design; the `bot_control_bypassed` alert is the
+compensating control.
+
+**Fail-open message idempotency admits duplicate bot messages** during an outage.
+Bounded by the 60s bucket and limited to bots that retry, and duplicates remain
+impossible on the expensive room-management paths.
--- a/docs/superpowers/specs/2026-08-11-valkey-hardening-design.md
+++ b/docs/superpowers/specs/2026-08-11-valkey-hardening-design.md
@@ -102,17 +102,37 @@ command. Values box through `any`, which is free relative to a network round-tri
 | `Interval` | 0 | No auto-reset while closed; consecutive-failure semantics. |
 | `OnStateChange` | transition log + metric | See Observability. |
 
-### IsSuccessful is load-bearing
+### Error classification is load-bearing
 
-gobreaker counts every returned error as a failure. `valkeyutil.ErrCacheMiss` is
-not a failure — it is the cache working correctly — and a cold or sparse keyspace
-would otherwise trip the breaker and disable Valkey for a workload that is
-behaving perfectly. `Settings.IsSuccessful` must therefore classify both
-`ErrCacheMiss` and `context.Canceled` as successes. Only transport errors count
-toward the trip threshold.
+gobreaker's default counts every returned error as a failure.
+`valkeyutil.ErrCacheMiss` is not a failure — it is the cache working correctly —
+and a cold or sparse keyspace would otherwise trip the breaker and disable
+Valkey for a workload that is behaving perfectly. `Settings.IsSuccessful` must
+classify `ErrCacheMiss` as a success. Only transport errors count toward the
+trip threshold.
 
-This is the single detail most likely to be got wrong, and it fails in the
-direction of a self-inflicted outage. It gets a dedicated test.
+Some errors are evidence of nothing either way, and for those *both* scores are
+wrong. Counting them failures opens the breaker on a healthy cache and stampedes
+the fallback store; counting them successes clears `ConsecutiveFailures` and can
+hold the breaker closed straight through a real outage. gobreaker v2 provides a
+third outcome for exactly this — `Settings.IsExcluded` drops a call from the
+accounting entirely — and it is consulted before `IsSuccessful`. Two errors are
+excluded:
+
+| Error | Why it is excluded |
+|---|---|
+| `context.Canceled` | The caller went away. An outage generates these in bulk — one slow sibling makes an errgroup cancel all the others at once. |
+| `redis.ErrPoolTimeout` | Local saturation: our own concurrency outran the pool, so Valkey was never asked. |
+
+`context.DeadlineExceeded` is deliberately **not** excluded. With
+`ContextTimeoutEnabled` bounding socket reads, a deadline is precisely how a
+blackholing Valkey surfaces — the degraded mode this design targets — so it must
+keep counting as a failure.
+
+This is the single detail most likely to be got wrong, and each way of getting
+it wrong disables the protection in a different direction. Both predicates get
+dedicated tests, including a test that an excluded error does not reset the
+consecutive-failure count.
 
 ### Callers need almost no changes
 
--- a/pkg/valkeyutil/breaker.go
+++ b/pkg/valkeyutil/breaker.go
@@ -1,0 +1,158 @@
+package valkeyutil
+
+import (
+	"context"
+	"errors"
+	"log/slog"
+	"time"
+
+	"github.com/redis/go-redis/v9"
+	"github.com/sony/gobreaker/v2"
+)
+
+const (
+	// breakerFailureThreshold rides out a single blip but trips on a real
+	// outage.
+	breakerFailureThreshold = 5
+	// breakerCooldown is how long the breaker stays open before allowing one
+	// half-open probe through.
+	breakerCooldown = 5 * time.Second
+)
+
+// ErrUnavailable is returned when the circuit breaker is open — Valkey is
+// known-down and the call short-circuited without a network round-trip.
+//
+// Every fail-open consumer already branches errors.Is(err, ErrCacheMiss) for
+// the miss case and falls back on everything else, so ErrUnavailable lands in
+// the fallback path with no consumer change required.
+var ErrUnavailable = errors.New("valkey: circuit open")
+
+// IsUnavailable reports whether err came from an open circuit breaker rather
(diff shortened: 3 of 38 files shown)


julianshen:
**4 findings · 4 inline comments posted** — 🔴 1 blocking · 🟠 3 warning

> [!NOTE]

<details>
<summary>📒 Files reviewed (38)</summary>

* `botplatform-service/metrics.go`
* `botplatform-service/metrics_test.go`
* `botplatform-service/middleware_idempotency_test.go`
* `botplatform-service/middleware_ratelimit_test.go`
* `broadcast-worker/main.go`
* `docs/health-probes.md`
* `docs/superpowers/plans/2026-08-11-valkey-hardening.md`
* `docs/superpowers/specs/2026-08-11-valkey-hardening-design.md`
* `go.mod`
* `go.sum`
* `notification-worker/members.go`
* `notification-worker/members_test.go`
* `pkg/roommetacache/valkey.go`
* `pkg/valkeyutil/breaker.go`
* `pkg/valkeyutil/breaker_test.go`
* `pkg/valkeyutil/integration_test.go`
* `pkg/valkeyutil/metrics.go`
* `pkg/valkeyutil/metrics_test.go`
* `pkg/valkeyutil/observability.go`
* `pkg/valkeyutil/observability_integration_test.go`
* `pkg/valkeyutil/observability_test.go`
* `pkg/valkeyutil/profile.go`
* `pkg/valkeyutil/profile_test.go`
* `pkg/valkeyutil/valkey.go`
* `room-worker/main.go`
* `search-service/handler.go`
* `search-service/main.go`
* `user-presence-service/presencestore/store.go`
* `user-presence-service/presencestore/store_test.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

pkg/valkeyutil/valkey.go
```diff
@@ -119,17 +198,27 @@ func (r *clusterClient) SetNX(ctx context.Context, key, value string, ttl time.D
...
 		}
+		return n, nil
+	}
+	// EXPIRE NX on every increment, not a bare EXPIRE on the 0->1 transition: if a
+	// prior call's INCR landed and its EXPIRE did not, the counter is left with no
+	// expiry, never resets, and permanently throttles the caller once it passes the
+	// ceiling. NX repairs that on the next increment while leaving a live window
+	// alone, so this stays a fixed window rather than becoming a sliding one.
+	// Pipelined to keep the pair at one round trip; both commands share a key, so
+	// cluster routing keeps them on one node.
```
julianshen, line 214: _🏷️ delivery-semantics_ | _🔴 Blocking_ | _`distributed-system`_

**IncrEx is non-atomic and unsafe under automatic retries.**

CacheProfile now enables one retry, while IncrEx sends INCR and EXPIRE NX as a two-command pipeline. If Valkey executes the pipeline but its response is lost, go-redis can retry it and increment twice for one admitted HTTP request; with a limit of 1, that first request can be rejected with count 2, and larger limits are consumed prematurely. A partial write can also execute INCR without EXPIRE, carrying stale count into the window repaired by a later request. Use one atomic Lua command for increment-plus-expiry and disable ambiguous automatic retries for this non-idempotent operation, or add a stable request ID and deduplicate server-side.

<details>

```
In `pkg/valkeyutil/valkey.go` around line 214: CacheProfile now enables one retry, while IncrEx sends INCR and EXPIRE NX as a two-command pipeline. If Valkey executes the pipeline but its response is lost, go-redis can retry it and increment twice for one admitted HTTP request; with a limit of 1, that first request can be rejected with count 2, and larger limits are consumed prematurely. A partial write can also execute INCR without EXPIRE, carrying stale count into the window repaired by a later request. Use one atomic Lua command for increment-plus-expiry and disable ambiguous automatic retries for this non-idempotent operation, or add a stable request ID and deduplicate server-side.

```

</details>

botplatform-service/middleware.go
```diff
@@ -125,8 +129,8 @@ func botRateLimit(client incrExClient, perCaller, perGlobal int) gin.HandlerFunc
 		if perCaller > 0 {
 			n, err := client.IncrEx(ctx, "botrl:caller:"+pr.UserID, window)
 			if err != nil {
-				errhttp.Write(ctx, c, errcode.Internal("bot rate limit caller", errcode.WithCause(err)))
-				c.Abort()
+				bypassControl(ctx, "rate limit", controlRateLimitCaller, err, "counter", "caller")
+				c.Next()
```
julianshen, line 133: _⚡ memory-exhaustion_ | _🔴 Blocking_ | _`oom`_

**Valkey failure removes every admission bound from bot traffic.**

When Valkey is unavailable, both the per-caller and global error branches call c.Next(), and the code explicitly permits an unthrottled looping caller. An authenticated bot can therefore open concurrent requests faster than handlers complete; each admitted request retains its request/session state and proceeds through body-reading idempotency and downstream handling. During a Valkey outage this can multiply in-flight memory until GC thrashing, container eviction, or process OOM, potentially cascading across replicas. Keep the fail-open behavior for ordinary traffic, but add a fixed-size process-local fallback token bucket or concurrency semaphore on this error path, rejecting with 429 when that hard bound is exhausted. The fallback must have fixed cardinality rather than an unbounded per-caller map.

<details>

```
In `botplatform-service/middleware.go` around line 133: When Valkey is unavailable, both the per-caller and global error branches call c.Next(), and the code explicitly permits an unthrottled looping caller. An authenticated bot can therefore open concurrent requests faster than handlers complete; each admitted request retains its request/session state and proceeds through body-reading idempotency and downstream handling. During a Valkey outage this can multiply in-flight memory until GC thrashing, container eviction, or process OOM, potentially cascading across replicas. Keep the fail-open behavior for ordinary traffic, but add a fixed-size process-local fallback token bucket or concurrency semaphore on this error path, rejecting with 429 when that hard bound is exhausted. The fallback must have fixed cardinality rather than an unbounded per-caller map.

```

</details>

pkg/valkeyutil/profile.go
```diff
@@ -0,0 +1,70 @@
...
+	// CacheProfile serves consumers that have a backing store to fall back to
+	// (room meta, room subscriptions, search restricted-rooms). Failing fast
+	// matters more than succeeding slowly — the fallback is always available.
+	CacheProfile = Profile{
+		DialTimeout:  time.Second,
+		ReadTimeout:  150 * time.Millisecond,
+		WriteTimeout: 150 * time.Millisecond,
+		PoolTimeout:  250 * time.Millisecond,
+		MaxRetries:   1,
+	}
```
julianshen, line 40: _🏷️ retries_ | _🔴 Blocking_ | _`performance-and-concurrency`, `tgd-review` (2 rules)_

**Automatic retries replay control writes after ambiguous timeouts.**

CacheProfile is used by botplatform-service and now enables one automatic retry for SetNX and the IncrEx pipeline. If Valkey executes a command but its response exceeds ReadTimeout or is lost, go-redis retries it: SetNX then returns false because the first attempt acquired the sentinel, causing the original request to be rejected as a duplicate, while IncrEx increments the rate counter twice for one request. These paths are reached during transient network loss or latency above 150ms, and the breaker cannot detect them because the retry can return successfully. Use a no-retry profile (MaxRetries: -1) for the bot control client, or make these operations safely retryable with an operation token/Lua protocol.

<details>

- **`tgd-review`**:39: CacheProfile now enables one go-redis retry and botplatform uses this profile for IncrEx and SetNX. After a command reaches Valkey but its reply is lost, retrying IncrEx can increment the rate-limit counter twice. Retrying SetNX can acquire on the first attempt and return false on the retry, causing the original request to be rejected as an in-flight duplicate despite the intended fail-open posture. Use a no-retry profile (MaxRetries: -1 under ClusterOptions semantics) for the bot controls, or redesign these operations so retries can recover the original result safely.

</details>

<details>

```
In `pkg/valkeyutil/profile.go` around line 40: CacheProfile is used by botplatform-service and now enables one automatic retry for SetNX and the IncrEx pipeline. If Valkey executes a command but its response exceeds ReadTimeout or is lost, go-redis retries it: SetNX then returns false because the first attempt acquired the sentinel, causing the original request to be rejected as a duplicate, while IncrEx increments the rate counter twice for one request. These paths are reached during transient network loss or latency above 150ms, and the breaker cannot detect them because the retry can return successfully. Use a no-retry profile (MaxRetries: -1) for the bot control client, or make these operations safely retryable with an operation token/Lua protocol.

```

</details>

pkg/valkeyutil/breaker.go
```diff
@@ -0,0 +1,174 @@
...
+// so it must keep counting as a failure.
+func isExcluded(err error) bool {
+	return errors.Is(err, context.Canceled) || errors.Is(err, redis.ErrPoolTimeout)
+}
+
+// breakerClient wraps a Client with one shared circuit breaker. Reachability
+// is a property of the connection, not of the command, so all methods share a
+// single breaker rather than one per operation.
+type breakerClient struct {
+	inner Client
```
julianshen, line 102: _🏷️ failure-isolation_ | _🟠 Warning_ | _`distributed-system`_

**One breaker conflates healthy and failed cluster slots.**

The single consecutive-failure counter covers every Redis Cluster slot and command. If the global rate-limit key's node is unreachable while caller-key nodes remain reachable, each request records a successful caller increment followed by a failed global increment; the success resets ConsecutiveFailures, so the breaker never opens and every request continues paying the timeout throughout the partition. Conversely, a burst against one failed slot can open the breaker and short-circuit healthy slots. Scope breaker accounting by node/slot or at least by independent keyspace/control so partial failures neither remain fail-slow nor disable healthy shards.

<details>

```
In `pkg/valkeyutil/breaker.go` around line 102: The single consecutive-failure counter covers every Redis Cluster slot and command. If the global rate-limit key's node is unreachable while caller-key nodes remain reachable, each request records a successful caller increment followed by a failed global increment; the success resets ConsecutiveFailures, so the breaker never opens and every request continues paying the timeout throughout the partition. Conversely, a burst against one failed slot can open the breaker and short-circuit healthy slots. Scope breaker accounting by node/slot or at least by independent keyspace/control so partial failures neither remain fail-slow nor disable healthy shards.

```

</details>

botplatform-service/middleware.go
julianshen, line 133: _🏷️ partial-failure_ | _🟠 Warning_ | _`distributed-system`_

**A caller-counter failure also bypasses the healthy global limit.**

When the per-caller key maps to an unavailable cluster node but the global key's node is healthy, this branch immediately runs the handler and returns, so the global counter is never checked or incremented. Traffic affected by only one failed shard therefore escapes the service-wide ceiling, and telemetry records only the caller bypass even though both controls were skipped. Record the caller-control bypass, continue to the global check, and invoke c.Next only after all still-available controls have run. Add a test where only the caller key errors and assert that the global key is still incremented.

<details>

```
In `botplatform-service/middleware.go` around line 133: When the per-caller key maps to an unavailable cluster node but the global key's node is healthy, this branch immediately runs the handler and returns, so the global counter is never checked or incremented. Traffic affected by only one failed shard therefore escapes the service-wide ceiling, and telemetry records only the caller bypass even though both controls were skipped. Record the caller-control bypass, continue to the global check, and invoke c.Next only after all still-available controls have run. Add a test where only the caller key errors and assert that the global key is still incremented.

```

</details>

pkg/valkeyutil/breaker.go
```diff
@@ -0,0 +1,174 @@
...
+//     outage generates these in bulk — one slow sibling makes an errgroup cancel
+//     all the others at once.
+//   - Pool timeout: our own concurrency outran the pool, so Valkey was never
+//     asked.
+//
+// context.DeadlineExceeded is deliberately NOT excluded. Once
+// ContextTimeoutEnabled bounds socket reads, a deadline is exactly how a
+// blackholing Valkey surfaces — the degraded mode this package exists to catch —
+// so it must keep counting as a failure.
+func isExcluded(err error) bool {
```
julianshen, line 94: _🏷️ timeout-handling_ | _🟠 Warning_ | _`distributed-system`_

**Expired caller deadlines can open the breaker on healthy Valkey.**

context.DeadlineExceeded is always counted as a Valkey failure, although ContextTimeoutEnabled can return it immediately when the incoming request's deadline has already expired without contacting Valkey. During a backlog, five consecutive expired requests can therefore open the breaker against a healthy cluster, forcing cache fallbacks or bot-control bypasses for the cooldown period. Make execution context-aware and exclude deadlines attributable to an already-expired caller context while continuing to count actual socket/read timeouts; add a test proving expired contexts alone cannot trip the breaker.

<details>

```
In `pkg/valkeyutil/breaker.go` around line 94: context.DeadlineExceeded is always counted as a Valkey failure, although ContextTimeoutEnabled can return it immediately when the incoming request's deadline has already expired without contacting Valkey. During a backlog, five consecutive expired requests can therefore open the breaker against a healthy cluster, forcing cache fallbacks or bot-control bypasses for the cooldown period. Make execution context-aware and exclude deadlines attributable to an already-expired caller context while continuing to count actual socket/read timeouts; add a test proving expired contexts alone cannot trip the breaker.

```

</details>

pkg/valkeyutil/breaker.go
```diff
@@ -0,0 +1,174 @@
...
+// outage. A miss is a completed round trip and so unambiguously healthy; only
+// genuine transport failures may count toward the trip threshold.
+//
+// Errors that are evidence of nothing either way never reach here: isExcluded
+// removes them from the accounting first.
+func isSuccessful(err error) bool {
+	if err == nil {
+		return true
+	}
+	return errors.Is(err, ErrCacheMiss)
```
julianshen, line 71: _🏷️ mongodb-operational-risk_ | _🟠 Warning_ | _`mongodb`_

**Semantic cache errors can open the breaker and fan out to MongoDB.**

`isSuccessful` treats every non-miss error as a breaker failure. Five `WRONGTYPE` or similar data/command errors from one malformed Valkey key therefore open the client-wide breaker. Unrelated room metadata and notification member lookups then receive `ErrUnavailable` and fall back to their MongoDB queries at full request/message rate, turning one poisoned cache key into collection-wide MongoDB load. Exclude non-availability Redis command/data errors from breaker accounting while still returning them to callers; continue counting network, timeout, and availability errors. Add a test proving repeated `WRONGTYPE` errors do not short-circuit unrelated keys.

<details>

```
In `pkg/valkeyutil/breaker.go` around line 71: `isSuccessful` treats every non-miss error as a breaker failure. Five `WRONGTYPE` or similar data/command errors from one malformed Valkey key therefore open the client-wide breaker. Unrelated room metadata and notification member lookups then receive `ErrUnavailable` and fall back to their MongoDB queries at full request/message rate, turning one poisoned cache key into collection-wide MongoDB load. Exclude non-availability Redis command/data errors from breaker accounting while still returning them to callers; continue counting network, timeout, and availability errors. Add a test proving repeated `WRONGTYPE` errors do not short-circuit unrelated keys.

```

</details>

pkg/valkeyutil/breaker.go
```diff
@@ -0,0 +1,174 @@
...
+// a cold or sparse keyspace would trip the breaker on ordinary cache misses and
+// disable Valkey for a workload that is behaving perfectly — a self-inflicted
+// outage. A miss is a completed round trip and so unambiguously healthy; only
+// genuine transport failures may count toward the trip threshold.
+//
+// Errors that are evidence of nothing either way never reach here: isExcluded
+// removes them from the accounting first.
+func isSuccessful(err error) bool {
+	if err == nil {
+		return true
```
julianshen, line 69: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**The breaker treats server replies as cluster outages.**

isSuccessful counts every non-miss Redis error as a reachability failure, despite the breaker being shared by reads and writes. For example, five SETs rejected with a Valkey OOM/noeviction response open the breaker even though GETs remain healthy. All subsequent reads then short-circuit to MongoDB or Elasticsearch for five seconds; after recovery, further write errors can repeat this fallback-load spike. Redis server error replies prove that the server was reached and should be considered successful for breaker accounting while still being returned to the caller. Keep transport errors and deadlines as failures, and add coverage for OOM/READONLY or another redis.Error response.

<details>

```
In `pkg/valkeyutil/breaker.go` around line 69: isSuccessful counts every non-miss Redis error as a reachability failure, despite the breaker being shared by reads and writes. For example, five SETs rejected with a Valkey OOM/noeviction response open the breaker even though GETs remain healthy. All subsequent reads then short-circuit to MongoDB or Elasticsearch for five seconds; after recovery, further write errors can repeat this fallback-load spike. Redis server error replies prove that the server was reached and should be considered successful for breaker accounting while still being returned to the caller. Keep transport errors and deadlines as failures, and add coverage for OOM/READONLY or another redis.Error response.

```

</details>

pkg/valkeyutil/valkey.go
```diff
@@ -35,34 +36,112 @@ type clusterClient struct {
...
+	case pingErr == nil:
+		slog.Info("connected to Valkey cluster", "addrs", addrs)
+	case failFast:
+		// Close the half-constructed client so unreachable addrs don't leak internal go-redis pool state.
 		if closeErr := c.Close(); closeErr != nil {
 			slog.Warn("valkey cluster close after failed connect", "error", closeErr)
 		}
-		return nil, fmt.Errorf("valkey cluster connect: %w", err)
+		return nil, nil, fmt.Errorf("valkey cluster connect: %w", pingErr)
+	default:
```
julianshen, line 86: _🏷️ error-handling_ | _🟠 Warning_ | _`silent-failure`_

**Lazy connect masks permanent and canceled PING failures as success.**

The non-fail-fast branch treats every PING error as transient unavailability and returns `(client, nil)`. A reachable cluster rejecting credentials (`WRONGPASS`, `NOAUTH`, or `NOPERM`) therefore lets every service report successful startup while cache users permanently fall back, bot controls fail open, and presence requests keep failing. A canceled parent context is similarly logged as an outage and startup continues. Only transient reachability/availability errors should be demoted: close and return `ctx.Err()` when the parent is canceled, propagate authentication/ACL and other permanent configuration errors, and add tests for those paths.

<details>

```
In `pkg/valkeyutil/valkey.go` around line 86: The non-fail-fast branch treats every PING error as transient unavailability and returns `(client, nil)`. A reachable cluster rejecting credentials (`WRONGPASS`, `NOAUTH`, or `NOPERM`) therefore lets every service report successful startup while cache users permanently fall back, bot controls fail open, and presence requests keep failing. A canceled parent context is similarly logged as an outage and startup continues. Only transient reachability/availability errors should be demoted: close and return `ctx.Err()` when the parent is canceled, propagate authentication/ACL and other permanent configuration errors, and add tests for those paths.

```

</details>

julianshen (commented):
tGD inline review

hmchangw:
Cross-posting from a merge-risk review of **#188** (MongoDB outage survival + `unread-worker`), because these two PRs collide and the order matters.

## Why #188 needs this PR

Reviewing #188 for merge risk, the one item I judged blocking was Valkey: it moves seven L2 tiers onto the send and history paths while `ConnectCluster` sets only `Addrs` and `Password`. I costed the exposure at ~12s per command (go-redis v9.18.0: `DialTimeout` 5s, `ReadTimeout` 3s, `MaxRetries` 3) against a 10s `REQUEST_TIMEOUT` — a stalled command outliving the whole request budget.

**That estimate was optimistic, and this PR explains why.** With `ContextTimeoutEnabled: false` the caller's context deadline does not bound socket reads *at all*, so there is no ceiling to compute. Calling that field load-bearing is right — and the startup trap is arguably the worse of the two findings, since a `CrashLoopBackOff` that outlives the outage converts a degraded cache into a dead service.

So #188's Valkey exposure is designed away here, and I withdrew the blocking call on that basis. Worth stating plainly though: this PR is a **draft**, labelled **`do not merge`**, based on `109e4a5` — two commits behind current `main` (`e793090`). Designed away is not closed. If #188 merges while this parks, the exposure is live on exactly the paths #188 widens.

## The classification reasoning holds up independently

The three-way-outcome argument in this PR's description is correct, and #188 reached the same conclusion separately. `pkg/circuitbreaker/circuitbreaker.go:197-198`:

```go
IsSuccessful:  func(err error) bool { return err == nil },
IsExcluded:    func(err error) bool { return err != nil && !s.isFailure(err) },
```

An exempted error is **excluded**, never scored as success, so it cannot clear `ConsecutiveFailures` and hold the breaker closed through a real outage — the second of the two defects `dc8f5a2` fixed. Two independent arrivals at the same answer is decent evidence the answer is right.

## Concrete collisions

**1. `pkg/valkeyutil`.** This PR touches 12 files there; #188 touches 5. Three overlap:

| File | Both |
|---|---|
| `pkg/valkeyutil/valkey.go` | ✗ |
| `pkg/valkeyutil/valkey_test.go` | ✗ |
| `pkg/valkeyutil/integration_test.go` | ✗ |

#188 adds `tier.go` (the shared refresh/slide/invalidation policy the six read-through tiers follow); this PR restructures the client construction underneath it. Whichever lands second gets a real conflict in the file the other is reshaping — not a mechanical one.

**2. Both add `github.com/sony/gobreaker/v2 v2.4.0`** — same version, so no version fight, but a `go.mod`/`go.sum` conflict either way. This also collapses **open question 1**: the dependency needs sign-off once, not twice, and whichever merges first settles it.

**3. Two breakers wrapping the same library.** This PR's `pkg/valkeyutil/breaker.go` and #188's `pkg/circuitbreaker` both wrap gobreaker/v2 with the same three-way scheme. That is not duplication anyone chose — this PR is based on `main`, where `pkg/circuitbreaker` does not exist yet, so reuse was not available.

## Suggested order: #188 first

Not because it is more important, but because it strictly reduces the work here:

- `pkg/circuitbreaker` becomes available, so this PR's core breaker can likely become `circuitbreaker.New(threshold, cooldown, WithFailurePredicate(FailureExcept(redis.ErrPoolTimeout)))` — `FailureExcept` already exempts `context.Canceled` by default, and the defaults match what this PR needs (`MaxRequests: 1` for a single half-open probe, `Interval: 0` so the closed-state count never auto-resets). The Valkey-specific parts — profiles, `LogDegraded`, warn-flood suppression, the `Client` facade decoration — stay here regardless.
- Open question 1 is answered before you get to it.
- The `pkg/valkeyutil` conflict lands in this PR, where the author already has the client restructure in their head, rather than in #188, where it would be one unfamiliar conflict among many.

The cost of that order is the window where #188 is merged and this is not. Given this PR is already a draft with three open questions, that window is real and worth deciding about explicitly rather than by default.

## On open question 2

Drop `docs/superpowers/**`. #188 shipped its equivalents and I would not do that again — they are working notes, they inflate the review surface, and in this PR's case the plan still contains the superseded `isSuccessful` body verbatim, so shipping it commits a known-wrong transcript to the tree.

---

docs/superpowers/specs/2026-08-11-valkey-hardening-design.md
```diff
@@ -0,0 +1,264 @@
...
+**Rate limiting is suspended.** It is a protective control — it guards the
+platform, not the data — so availability wins during an outage. `botRateLimit`
+(`botplatform-service/middleware.go:128`) gains a fail-open path: on a Valkey
+error it logs WARN, emits a metric, and calls `c.Next()` instead of aborting with
+`errcode.Internal`.
+
+**Idempotency splits by endpoint.** It is a correctness control: the sentinel is
+what stops a retrying bot from double-executing within the 60s bucket. A
+duplicate message is visible but recoverable; a duplicate room creation or member
+add is expensive and awkward to undo. So:
```
julianshen, line 198: _📝 documentation_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**The design spec still documents the superseded posture split.**

This section says room-management idempotency remains fail-closed, but botIdempotency now admits every endpoint on SetNX errors and the PR description calls unconditional fail-open deliberate. The later risk statement also claims duplicates remain impossible on expensive room-management paths. Update the endpoint table, observability labels, and risk discussion so the retained design document matches production behavior.

<details>

```
In `docs/superpowers/specs/2026-08-11-valkey-hardening-design.md` around line 198: This section says room-management idempotency remains fail-closed, but botIdempotency now admits every endpoint on SetNX errors and the PR description calls unconditional fail-open deliberate. The later risk statement also claims duplicates remain impossible on expensive room-management paths. Update the endpoint table, observability labels, and risk discussion so the retained design document matches production behavior.

```

</details>

---

julianshen (commented):
### tGDBot review of `dc8f5a2` — 1 finding in 38 files

1 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/238#issuecomment-5333970621), which is authoritative for what this run found and where each finding went.
Rules: `cassandra`, `dead-code`, `distributed-system`, `nats`, `oom`, `performance-and-concurrency`, `silent-failure`, `tgd-review` · 3 rules failed (`elasticsearch`, `mongodb`, `nats-large-payload`) — reason in the summary.

<details>

</details>

<details>

One command per comment.

</details>

---

pkg/valkeyutil/profile.go
```diff
@@ -0,0 +1,70 @@
...
+var (
+	// CacheProfile serves consumers that have a backing store to fall back to
+	// (room meta, room subscriptions, search restricted-rooms). Failing fast
+	// matters more than succeeding slowly — the fallback is always available.
+	CacheProfile = Profile{
+		DialTimeout:  time.Second,
+		ReadTimeout:  150 * time.Millisecond,
+		WriteTimeout: 150 * time.Millisecond,
+		PoolTimeout:  250 * time.Millisecond,
+		MaxRetries:   1,
```
julianshen, line 39: _🎯 correctness_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`tgd-review`_

**Automatic retries can replay bot control writes.**

CacheProfile enables one retry, and botplatform uses it for SetNX and IncrEx. If Valkey applies a write but its reply times out, go-redis may retry it: SetNX then returns false and rejects the original request as a duplicate, while IncrEx increments twice and can reject the first request when the limit is 1. The breaker sees only the final result and cannot correct this. Use a no-retry profile for these controls or make the operations safely retryable.

<details>

```
In `pkg/valkeyutil/profile.go` around line 39: CacheProfile enables one retry, and botplatform uses it for SetNX and IncrEx. If Valkey applies a write but its reply times out, go-redis may retry it: SetNX then returns false and rejects the original request as a duplicate, while IncrEx increments twice and can reject the first request when the limit is 1. The breaker sees only the final result and cannot correct this. Use a no-retry profile for these controls or make the operations safely retryable.

```

</details>

---

botplatform-service/middleware.go
```diff
@@ -125,8 +129,8 @@ func botRateLimit(client incrExClient, perCaller, perGlobal int) gin.HandlerFunc
...
 				return
```
julianshen, line 134: _🏷️ partial-failure_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`_

**A caller-shard failure bypasses the healthy global limit**

When the caller key's shard is unavailable but the global key's shard is healthy, this branch runs downstream immediately and returns, so the global counter is never checked or incremented. Repeated traffic therefore escapes the service-wide ceiling during a single-shard failure. Record the caller-control bypass, continue to the global check, and call c.Next only after all available controls have run. Add a test where only the caller key errors and verify the global key is incremented.

<details>

> ‼️ **IMPORTANT**

```suggestion
				bypassControl(ctx, "rate limit", controlRateLimitCaller, err, "counter", "caller")
```

</details>

<details>

```
In `botplatform-service/middleware.go` around line 132: When the caller key's shard is unavailable but the global key's shard is healthy, this branch runs downstream immediately and returns, so the global counter is never checked or incremented. Repeated traffic therefore escapes the service-wide ceiling during a single-shard failure. Record the caller-control bypass, continue to the global check, and call c.Next only after all available controls have run. Add a test where only the caller key errors and verify the global key is incremented.

```

</details>

---

pkg/valkeyutil/breaker.go
```diff
@@ -0,0 +1,174 @@
...
+//   - Pool timeout: our own concurrency outran the pool, so Valkey was never
+//     asked.
+//
+// context.DeadlineExceeded is deliberately NOT excluded. Once
+// ContextTimeoutEnabled bounds socket reads, a deadline is exactly how a
+// blackholing Valkey surfaces — the degraded mode this package exists to catch —
+// so it must keep counting as a failure.
+func isExcluded(err error) bool {
+	return errors.Is(err, context.Canceled) || errors.Is(err, redis.ErrPoolTimeout)
+}
```
julianshen, line 96: _🏷️ cancellation_ | _🟠 Warning_ | _⚡ Quick fix_ | _`performance-and-concurrency`_

**Expired caller deadlines can open a healthy cluster's breaker.**

DeadlineExceeded is always counted as a Valkey failure, but ContextTimeoutEnabled can return it immediately when the incoming context has already expired without contacting Valkey. Five such requests during backlog open the shared breaker, forcing cache fallbacks and bot-control bypasses for the cooldown despite a healthy cluster. Pass the execution context into breaker classification and exclude DeadlineExceeded when ctx.Err() shows the caller deadline expired, while continuing to count socket/read timeouts where the caller context remains live.

<details>

```
In `pkg/valkeyutil/breaker.go` around line 96: DeadlineExceeded is always counted as a Valkey failure, but ContextTimeoutEnabled can return it immediately when the incoming context has already expired without contacting Valkey. Five such requests during backlog open the shared breaker, forcing cache fallbacks and bot-control bypasses for the cooldown despite a healthy cluster. Pass the execution context into breaker classification and exclude DeadlineExceeded when ctx.Err() shows the caller deadline expired, while continuing to count socket/read timeouts where the caller context remains live.

```

</details>

---

pkg/valkeyutil/breaker.go
```diff
@@ -0,0 +1,174 @@
...
+// genuine transport failures may count toward the trip threshold.
+//
+// Errors that are evidence of nothing either way never reach here: isExcluded
+// removes them from the accounting first.
+func isSuccessful(err error) bool {
+	if err == nil {
+		return true
+	}
+	return errors.Is(err, ErrCacheMiss)
+}
```
julianshen, line 72: _🏷️ failure-classification_ | _🟠 Warning_ | _⚡ Quick fix_ | _`performance-and-concurrency`_

**Redis server replies are counted as cluster outages.**

Every non-miss error currently counts as a reachability failure. Five fast WRONGTYPE, OOM, or other Redis error replies therefore open the client-wide breaker even though the server was reached; unrelated reads then short-circuit to MongoDB or Elasticsearch and can repeat this load spike every cooldown. Treat redis.Error replies as successful communication for breaker accounting while still returning the command error to the caller. Keep network errors and timeouts classified as failures.

<details>

> ‼️ **IMPORTANT**

```suggestion
	var redisErr redis.Error
	return errors.Is(err, ErrCacheMiss) || errors.As(err, &redisErr)
```

</details>

<details>

```
In `pkg/valkeyutil/breaker.go` around line 72: Every non-miss error currently counts as a reachability failure. Five fast WRONGTYPE, OOM, or other Redis error replies therefore open the client-wide breaker even though the server was reached; unrelated reads then short-circuit to MongoDB or Elasticsearch and can repeat this load spike every cooldown. Treat redis.Error replies as successful communication for breaker accounting while still returning the command error to the caller. Keep network errors and timeouts classified as failures.

```

</details>

---

julianshen (commented):
### tGDBot review of `dc8f5a2` — 4 findings in 38 files

1 🔴 blocking · 3 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/238#issuecomment-5333970621), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

hmchangw:
Closing in favour of #486.

This branch was written against a `main` that has since moved 130+ commits, and about half of it has been implemented upstream in the meantime — so rebasing it would have re-introduced work the repo already has, in worse form:

- **`pkg/circuitbreaker` now exists** and is the repo standard, used across 10+ files, with the same three-way classifier (`IsExcluded`) that `dc8f5a2` on this branch had converged on independently. This branch's `pkg/valkeyutil/breaker.go` would have landed as a second, competing breaker.
- **`valkeyutil` gained a `Config`/`Connect`/`ConnectRaw`/`ConnectOptional`/`Instrumented` API.** That subsumes this branch's `NewClusterClient` (it is `ConnectRaw`) and its presence-store commit — `user-presence-service` already dials through `ConnectRaw` with instrumentation on `main`.
- **`gobreaker/v2` is already a `main` dependency**, so the sign-off this PR asked for is moot.

What was still genuinely missing on `main` — verified rather than assumed — is carried forward in #486, rewritten against the current API:

- `dialCluster` still set **no timeouts at all** and left `ContextTimeoutEnabled` false;
- the startup `PING` still gated the dial, with ten call sites exiting on it;
- `IncrEx` still had the latent bug where a landed `INCR` with a failed `EXPIRE` strands a counter with no expiry, throttling that caller permanently;
- `botplatform-service` still returned 500 on a Valkey error rather than failing open.

#486 is a much smaller diff for the same outcome, and carries none of the spec/plan working notes that made up roughly 60% of this one.

Nothing here is lost that mattered: `dc8f5a2`'s correction to the breaker classification is the one piece of original thinking on this branch that `main` arrived at independently, and the reasoning behind it is preserved in #486's description.

---

PR #239, opened 2026-08-11
[NATS Tuning] Classify a missing upstream as unavailable, and take max_payload from the broker

hmchangw:
## The bug

Fourteen NATS request/reply call sites share one shape:

```go
msg, err := c.nc.Request(ctx, subject.RoomsInfoBatch(siteID), req, roomRPCTimeout)
if err != nil {
    return nil, fmt.Errorf("rooms-info rpc: %w", err)   // <- collapses to internal/500
}
if e, ok := errcode.Parse(msg.Data); ok {
    return nil, e                                       // <- careful: relays the remote's own classification
}
```

The third step is careful. The second is not. A raw wrapped error collapses to `internal` at the boundary, so `nats.ErrNoResponders` — *nothing is subscribed to this subject* — reaches the client as a 500 and reads to an operator as an application bug. It isn't one. It means an upstream service is down, not yet started, or not routed to this site.

The cost lands twice: a client can't tell "retry in a moment" from "this will never work", and an on-call engineer starts debugging application code for a deployment problem. The subject wasn't in the error either, so the log didn't say which upstream vanished.

## What changed

**`natsutil.RequestFailure(op string, err error) error`** classifies two failure families and leaves everything else exactly as it was:

| Input | Result |
|---|---|
| `nats.ErrNoResponders` | `errcode.Unavailable` (503), reason `no_responders` |
| `nats.ErrTimeout` / `context.DeadlineExceeded` | `errcode.Unavailable` (503), reason `upstream_timeout` |
| anything else | `fmt.Errorf("%s: %w", op, err)` — unchanged |

Matching is `errors.Is`, so a wrapped sentinel still resolves. The underlying error goes in `WithCause` — server-side only, logged once by `Classify` — deliberately **not** in `Metadata`, which is client-visible and would leak internal topology. A test asserts the cause never appears in the marshalled envelope.

`ErrTimeout` and `context.DeadlineExceeded` share one reason on purpose: the distinction survives in the cause and the server log, and the client's action is "retry" either way, so a second reason would ask clients to branch on something they can't act on. `context.Canceled` is not handled — only `admin-service` maps it, for a reason specific to that idempotent call.

**11 of the 14 sites adopt it. Three deliberately do not**, because they already do this better:

| Site | Why it stays |
|---|---|
| `room-service/reader_history.go` | Already returns `Unavailable` with the **domain** reason `RoomReadReceiptsUnavailable`, which clients are documented to branch on. The generic reason would tell them *less*. |
| `admin-service/room_onduty.go` | Already handles no-responders, timeout and deadline — **plus `context.Canceled`**, which the helper doesn't. Converting would drop a case. |
| `notification-worker/presence.go` | Fire-and-forget goroutine; logs and returns. No error path to classify. |

**Second, unrelated fix:** `notification-worker` enforced a push-batch size cap taken from `NATS_MAX_PAYLOAD_BYTES`, whose own comment said "must match broker max_payload" with nothing enforcing it. Drift *below* the broker's real limit made the worker reject batches the broker would have accepted — mobile notifications dropped by config drift, with no error at the broker and no alert. It now reads `nc.NatsConn().MaxPayload()`, as `room-service` already did, and the env var is deleted. The `int64`→`int` narrowing is clamped rather than cast (gosec G115).

## What a client can actually observe

Only **one** of the eleven sites changes client-visible behaviour: `subscription.setAppSubscription`. Subscribing to an app with no existing DM room creates a botDM room via room-service, and that error returns to the caller — so a room-service outage now gives 503 + a reason instead of a bare 500. Documented in `docs/client-api.md` and its derived view.

The other ten already degraded gracefully (per-site fan-out skips, placeholder quotes, best-effort enrichment) and never surfaced the errcode to a client. For them this is an internal typing and logging improvement only. Reviewers should not go looking for ten behaviour changes that aren't there.

## Deployment notes

- **`NATS_MAX_PAYLOAD_BYTES` is deleted.** Any environment still setting it will find it ignored — `caarlos0/env` only parses tags present on the struct, so there is no fail-fast risk in either deploy order, and ops can remove it whenever convenient.
- **500 → 503 on the one path above.** Intended: the old value was wrong.
- The payload cap is read **once at startup**. A reconnect to a cluster member advertising a *smaller* `max_payload` would leave a stale larger cap. That needs heterogeneous `max_payload` across members, and `room-service` has the identical one-shot pattern, so this is consistent with existing convention rather than new risk — noted for completeness.

## Verification

- `make test` — full repo, `-race`, all packages pass
- `make lint` — 0 issues
- `make build` — every touched service
- `RequestFailure` at 100% coverage; `pkg/natsutil` at 89.8%
- The no-responders path is proven against a **real embedded NATS server** (`natsserver.NewServer`, in-process), not a hand-made sentinel — a plain unit test, no Docker

**SAST is partially unverified in this environment.** `gosec` passes (confirming the G115 clamp), and semgrep's in-repo rules pass with the gate's real flags — 0 findings. `govulncheck` and semgrep's registry rulesets (`p/golang`, `p/security-audit`) could not run: the sandbox's egress policy returns 403 for `vuln.go.dev` and `semgrep.dev`. This branch touches no `go.mod`/`go.sum`, so `govulncheck`'s verdict should match `main`'s — but that is reasoning, not a result. **CI must run the full gate.**

`notification-worker` package coverage is 51.7%, below the 80% floor. That is pre-existing — `main` is 51.4% — and this branch improves it slightly. The shortfall is store and `main.go` wiring that CLAUDE.md scopes to integration tests.

## Notes for review

- The two new reasons are **mechanism**-named (`no_responders`, `upstream_timeout`) while the adjacent existing vocabulary is **outcome**-named (`upstream_unavailable`, `read_receipts_unavailable`). Deliberate — those describe an HTTP backend, these a NATS peer — but a client author skimming the catalog will see two philosophies.
- For the ten swallowed sites the **logged** error text changes from the raw NATS error to the errcode's static message, because `Error()` never includes the cause. The cause still reaches logs via `Classify`.
- `clampPayloadCap`'s clamp branch is unreachable on 64-bit (`math.MaxInt == math.MaxInt64`). It stays because it is correct on 32-bit and is what satisfies gosec; its test case is named and commented to say so rather than implying coverage it can't have.
- **Commit hygiene:** `e0975b4`'s message describes a plan-document change but the commit also contains the rewrite of `docs/notification-worker-downstream-contracts.md`. That was my staging error — a whole-directory `git add` while that file was being edited. The content is correct and reviewed; only the message under-describes it. Flagged rather than rewritten, since the branch was already pushed.

This is PR 2 of 2 from the NATS connection-handling review. PR 1 (#223, blocking drain + slow-consumer diagnostics) is independent — neither depends on the other and they may merge in either order.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EkBKmG55BngvTkgegDEQPq)_


julianshen:
**Actionable comments posted: 5** — 🟠 4 warning · 🔵 1 suggestion

<details>

* `docs/client-api.md`
* `docs/error-handling.md`
* `docs/superpowers/plans/2026-08-11-nats-payload-noresponders.md`
* `docs/superpowers/specs/2026-08-11-nats-payload-noresponders-design.md`
* `notification-worker/emit.go`
* `notification-worker/emit_test.go`
* `pkg/errcode/codes_platform.go`
* `pkg/natsutil/request_failure.go`
* `pkg/natsutil/request_failure_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `silent-failure`
* `tgd-review`

</details>

notification-worker/main.go
```diff
@@ -213,7 +212,10 @@ func main() {
 		os.Exit(1)
 	}
 
-	emitter := newMobileEmitter(&jsPublisher{js: otelJS}, wiring.PushSendSubject, cfg.NatsMaxPayloadBytes)
+	// The broker advertises max_payload in its INFO on connect, so this is
+	// always in step with the server. An env var was a second source of truth
+	// that silently dropped batches whenever it drifted below the real limit.
+	emitter := newMobileEmitter(&jsPublisher{js: otelJS}, wiring.PushSendSubject, clampPayloadCap(nc.NatsConn().MaxPayload()))
```
julianshen, line 218: _🏷️ failure behavior_ | _🟠 Warning_ | _`distributed-system`_

**The payload cap goes stale after a broker reconnect.**

The broker limit is sampled once when the emitter is constructed. If the connection later fails over during a rolling configuration change—from a 256 KiB server to a 1 MiB server, for example—the emitter permanently retains 256 KiB and rejects batches the current broker would accept. The inverse transition lets oversized batches reach PublishAsync and repeatedly fail with ErrMaxPayload. This recreates the notification loss/retry-loop behavior the change intends to remove. Pass the connection or a cap-supplier function into the emitter and read MaxPayload immediately before each publish; the publish error must remain the final authority if a reconnect races the check.

<details>

```
In `notification-worker/main.go` around line 218: The broker limit is sampled once when the emitter is constructed. If the connection later fails over during a rolling configuration change—from a 256 KiB server to a 1 MiB server, for example—the emitter permanently retains 256 KiB and rejects batches the current broker would accept. The inverse transition lets oversized batches reach PublishAsync and repeatedly fail with ErrMaxPayload. This recreates the notification loss/retry-loop behavior the change intends to remove. Pass the connection or a cap-supplier function into the emitter and read MaxPayload immediately before each publish; the publish error must remain the final authority if a reconnect races the check.

```

</details>

julianshen, line 218: _🏷️ nats-reconnect_ | _🟠 Warning_ | _`nats`_

**The payload cap becomes stale after a NATS reconnect.**

The emitter snapshots `MaxPayload()` only at startup, while the NATS connection can later reconnect to a server advertising a different `max_payload`. During a rolling broker configuration change, a cap captured from the old server can still reject batches accepted by the current server, or allow the preflight check only for the current connection to return `ErrMaxPayload`. This contradicts the stated guarantee that the guard always tracks the connected broker. Pass a live cap accessor to `mobileEmitter` and query it immediately before each size check, or atomically refresh the stored cap from the reconnect callback.

<details>

```
In `notification-worker/main.go` around line 218: The emitter snapshots `MaxPayload()` only at startup, while the NATS connection can later reconnect to a server advertising a different `max_payload`. During a rolling broker configuration change, a cap captured from the old server can still reject batches accepted by the current server, or allow the preflight check only for the current connection to return `ErrMaxPayload`. This contradicts the stated guarantee that the guard always tracks the connected broker. Pass a live cap accessor to `mobileEmitter` and query it immediately before each size check, or atomically refresh the stored cap from the reconnect callback.

```

</details>

julianshen, line 218: _🏷️ reconnect handling_ | _🟠 Warning_ | _`performance-and-concurrency`_

**The payload cap becomes stale after a NATS reconnect.**

`MaxPayload()` is sampled only when the emitter is constructed, but the NATS connection can reconnect to another server whose INFO advertises a different limit. For example, after starting against a 256 KiB server and reconnecting during a rolling configuration change to a 1 MiB server, every batch between those sizes remains locally rejected even though the active broker accepts it, dropping notifications under load. Reconnecting in the opposite direction bypasses the intended preflight guard and produces publish failures and retries. Read the connection's current `MaxPayload()` for each emission, or atomically refresh the stored cap from a reconnect callback; add a reconnect test using servers with different limits.

<details>

```
In `notification-worker/main.go` around line 218: `MaxPayload()` is sampled only when the emitter is constructed, but the NATS connection can reconnect to another server whose INFO advertises a different limit. For example, after starting against a 256 KiB server and reconnecting during a rolling configuration change to a 1 MiB server, every batch between those sizes remains locally rejected even though the active broker accepts it, dropping notifications under load. Reconnecting in the opposite direction bypasses the intended preflight guard and produces publish failures and retries. Read the connection's current `MaxPayload()` for each emission, or atomically refresh the stored cap from a reconnect callback; add a reconnect test using servers with different limits.

```

</details>

julianshen, line 218: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**The broker payload cap becomes stale after a reconnect.**

`MaxPayload()` is evaluated only once during startup and copied into `mobileEmitter.maxPayloadBytes`. NATS connections can reconnect to another server whose INFO advertises a different limit; after that, the emitter can still reject payloads the new server accepts or omit the intended preflight rejection for a smaller limit. This also contradicts the documentation's claim that the cap always tracks the connected broker. Query the connection's current cap when emitting, or update the stored cap from a reconnect callback using synchronization.

<details>

```
In `notification-worker/main.go` around line 218: `MaxPayload()` is evaluated only once during startup and copied into `mobileEmitter.maxPayloadBytes`. NATS connections can reconnect to another server whose INFO advertises a different limit; after that, the emitter can still reject payloads the new server accepts or omit the intended preflight rejection for a smaller limit. This also contradicts the documentation's claim that the cap always tracks the connected broker. Query the connection's current cap when emitting, or update the stored cap from a reconnect callback using synchronization.

```

</details>

docs/client-api.md
```diff
@@ -6710,6 +6712,8 @@ Every error response — NATS reply subjects, JetStream async results, and HTTP
 | `priority_contact_limit` | forbidden | user-service `settings.priorityContacts.add` (list already at the 30-entry cap and the contact is not already on it) |
 | `priority_contact_not_found` | not_found | user-service `settings.priorityContacts.add` (contact account does not resolve — inactive/unknown user, or app not found for a `.bot` account) |
 | `response_too_large` | internal | any RPC whose reply would exceed the transport `max_payload` (most often large history reads — retry with a smaller `limit`) |
+| `no_responders` | unavailable | any request/reply RPC whose upstream subject had no subscriber (`natsutil.RequestFailure`) — the upstream service is down, not yet started, or not routed to this site. Retry. |
+| `upstream_timeout` | unavailable | any request/reply RPC delivered to the upstream but not answered within the caller's timeout (`natsutil.RequestFailure`). Retry. |
```
julianshen, line 6716: _🏷️ observability_ | _🔵 Suggestion_ | _`distributed-system`_

**The timeout contract falsely asserts that the request was delivered.**

Neither nats.ErrTimeout nor context.DeadlineExceeded proves delivery. A parent context can expire before publication, while the same classification can also occur after the upstream committed a write and its reply was lost. Both timelines produce upstream_timeout, so describing it as “delivered but not answered” masks the operation's actual uncertainty. Define it as “no reply arrived before the deadline; delivery and processing are unknown” in this catalog and the corresponding code comments, and qualify retry guidance to idempotent operations.

<details>

```
In `docs/client-api.md` around line 6716: Neither nats.ErrTimeout nor context.DeadlineExceeded proves delivery. A parent context can expire before publication, while the same classification can also occur after the upstream committed a write and its reply was lost. Both timelines produce upstream_timeout, so describing it as “delivered but not answered” masks the operation's actual uncertainty. Define it as “no reply arrived before the deadline; delivery and processing are unknown” in this catalog and the corresponding code comments, and qualify retry guidance to idempotent operations.

```

</details>

mliu33 (approved):
Thanks!

PR #242, opened 2026-08-12
feat(broadcast-worker): emit new_thread_message event type for thread replies

chenjr0719:
## Summary

Newly-created thread replies now publish a distinct client event `type`,
`new_thread_message`, instead of reusing `new_message`. Normal (non-thread)
messages are unaffected.

## What's included

- `pkg/model/event.go`: new `RoomEventNewThreadMessage RoomEventType = "new_thread_message"` const.
- `broadcast-worker/handler.go`:
  - `handleThreadCreated` channel branch: sets `roomEvt.Type = model.RoomEventNewThreadMessage` after `buildRoomEvent`.
  - `handleThreadCreated` DM/BotDM branch: `publishDMEvents` now takes a `roomEventType model.RoomEventType` parameter, set per call site — `RoomEventNewMessage` from `handleCreated`, `RoomEventNewThreadMessage` from `handleThreadCreated`.
- `docs/client-api/events.md` + `docs/client-api.md`: doc-ratchet — added the `new_thread_message` event section, updated the event-type enumeration and the two other places that described thread replies as `new_message`.

## Breaking change

This is a wire-contract change. A thread reply's live client event now arrives with
`type: "new_thread_message"` rather than `"new_message"`. **Frontend clients need to
add a handler for the new type** (or widen their existing `new_message` handler to
also match `new_thread_message`) before/at the same time this ships, or thread replies
will stop rendering live for clients that only match on `"new_message"`.

Thread edit (`message_edited`) and delete (`message_deleted`) events are unchanged —
only the create event gets the new type.

## Verification

- `go test ./broadcast-worker/... ./pkg/model/...` — all green.
- `chat-lint.sh ./broadcast-worker/... ./pkg/model/...` — 0 issues.
- Updated unit tests assert `RoomEvent.type == "new_thread_message"` for both the
  channel-room and DM-room thread-reply paths; existing non-thread-create tests
  (channel + DM) continue to assert `"new_message"`, confirming the scoping.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated `new_thread_message` event for thread replies in channels and direct messages.
  * Regular messages continue to use the `new_message` event.
  * Documented event payloads, routing, and bot direct-message delivery behavior.

* **Bug Fixes**
  * Corrected thread-reply event classification for channel and direct-message notifications.
  * Improved bot direct-message fan-out to deliver events to the human participant only.


chenjr0719:
Thanks — all four addressed in the latest commit (delivery-path accuracy was the theme):

- **Canonical discriminator / type schema:** qualified the channel `type` row + the triggered-events note so a client following the canonical schema knows a thread reply arrives as `new_thread_message` (distinct type + delivery), not `new_message`.
- **Channel thread replies are per-subscriber:** documented that they fan out on `chat.user.{account}.event.room` (sender + parent author + thread followers + history-gated mentionees), **not** room-wide on `chat.room.{roomID}.event`. Fixed the subject table + the `new_thread_message` section.
- **botDM thread replies:** corrected — they **do** fan out per member; only an ordinary botDM `new_message` is suppressed. Fixed in both API views.
- **Coverage:** added `TestHandleThreadCreated_BotDMRoom_EmitsNewThreadMessage` asserting botDM thread fan-out carries `new_thread_message`.

julianshen:

<details>

* `docs/client-api.md`
* `docs/client-api/events.md`
* `pkg/model/event.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `silent-failure`
* `tgd-review`

</details>

broadcast-worker/handler.go
```diff
@@ -262,6 +262,7 @@ func (h *Handler) handleThreadCreated(ctx context.Context, evt *model.MessageEve
 		// in the main channel, so a room-level mention badge would appear with no
 		// visible message to explain it.
 		roomEvt := buildRoomEvent(&meta, clientMsg, evt.Timestamp)
+		roomEvt.Type = model.RoomEventNewThreadMessage
```
julianshen, line 265: _🏷️ protocol compatibility_ | _🟠 Warning_ | _`distributed-system`_

**The new discriminator breaks old clients during rollout**

Previously deployed clients were documented to handle thread replies as `new_message`. Once an upgraded worker processes a reply, it now publishes only `new_thread_message`; an old client that ignores unknown event types receives no live reply despite the send succeeding. During a rolling worker deployment, otherwise identical replies will also nondeterministically use old or new types depending on which replica processes them. Preserve `new_message` compatibility during a transition, version or feature-gate the protocol, or explicitly require clients to accept both forms and deploy that support before changing producers.

<details>

```
In `broadcast-worker/handler.go` around line 265: Previously deployed clients were documented to handle thread replies as `new_message`. Once an upgraded worker processes a reply, it now publishes only `new_thread_message`; an old client that ignores unknown event types receives no live reply despite the send succeeding. During a rolling worker deployment, otherwise identical replies will also nondeterministically use old or new types depending on which replica processes them. Preserve `new_message` compatibility during a transition, version or feature-gate the protocol, or explicitly require clients to accept both forms and deploy that support before changing producers.

```

</details>

docs/client-api.md
```diff
@@ -6317,15 +6317,15 @@ Delivered on `chat.user.{account}.response.{requestId}`. See [Error envelope](#6
 
 #### Triggered events — success path
```
julianshen, line 6319: _🏷️ delivery contract_ | _🟠 Warning_ | _`distributed-system`_

**The documented botDM suppression contradicts the producer branch**

This paragraph says an ordinary botDM message falls through the default branch and receives no fan-out, but `handleCreated` explicitly matches `RoomTypeDM, RoomTypeBotDM` and calls `publishDMEvents` with `RoomEventNewMessage`. Thus a non-thread botDM creation can publish per-member events contrary to the documented delivery contract. Correct the documentation to describe that fan-out, or remove/guard the botDM case if suppression is the intended contract, and add a non-thread botDM test to lock down the chosen behavior.

<details>

```
In `docs/client-api.md` around line 6319: This paragraph says an ordinary botDM message falls through the default branch and receives no fan-out, but `handleCreated` explicitly matches `RoomTypeDM, RoomTypeBotDM` and calls `publishDMEvents` with `RoomEventNewMessage`. Thus a non-thread botDM creation can publish per-member events contrary to the documented delivery contract. Correct the documentation to describe that fan-out, or remove/guard the botDM case if suppression is the intended contract, and add a non-thread botDM test to lock down the chosen behavior.

```

</details>

broadcast-worker/handler.go
```diff
@@ -202,7 +202,7 @@ func (h *Handler) handleCreated(ctx context.Context, evt *model.MessageEvent) er
 	case model.RoomTypeChannel:
 		return h.publishChannelEvent(ctx, &meta, clientMsg, evt.Timestamp, resolved.MentionAll, resolved.Participants)
 	case model.RoomTypeDM, model.RoomTypeBotDM:
-		return h.publishDMEvents(ctx, &meta, clientMsg, evt.Timestamp, resolved.Accounts)
+		return h.publishDMEvents(ctx, &meta, clientMsg, evt.Timestamp, resolved.Accounts, model.RoomEventNewMessage)
```
julianshen, line 205: _🏷️ NATS routing_ | _🟠 Warning_ | _`nats`_

**Ordinary botDM messages still fan out despite the documented contract.**

For a non-thread message in a `RoomTypeBotDM`, the preceding shared `RoomTypeDM, RoomTypeBotDM` case reaches this call, which lists room subscriptions and publishes `new_message` to each `chat.user.{account}.event.room` subject. This contradicts the changed documentation and new test commentary stating that ordinary botDM fan-out is suppressed, and can create an unexpected or duplicate client delivery alongside the separate bot integration path. Remove `RoomTypeBotDM` from the non-thread switch case (while retaining it in `handleThreadCreated`), and add a regression test asserting that an ordinary botDM message publishes no room event.

<details>

```
In `broadcast-worker/handler.go` around line 205: For a non-thread message in a `RoomTypeBotDM`, the preceding shared `RoomTypeDM, RoomTypeBotDM` case reaches this call, which lists room subscriptions and publishes `new_message` to each `chat.user.{account}.event.room` subject. This contradicts the changed documentation and new test commentary stating that ordinary botDM fan-out is suppressed, and can create an unexpected or duplicate client delivery alongside the separate bot integration path. Remove `RoomTypeBotDM` from the non-thread switch case (while retaining it in `handleThreadCreated`), and add a regression test asserting that an ordinary botDM message publishes no room event.

```

</details>

docs/client-api.md
julianshen, line 6319: _📝 documentation_ | _🟠 Warning_ | _`tgd-review`_

**The botDM delivery contract contradicts the implementation.**

This paragraph says ordinary botDM messages fall through and receive no fan-out, but `handleCreated` explicitly groups `RoomTypeDM` and `RoomTypeBotDM` and calls `publishDMEvents` for both. The same incorrect suppression claim also appears in `docs/client-api/events.md`. Update the documentation to describe the actual per-member botDM fan-out, or change the implementation if suppression is the intended contract.

<details>

```
In `docs/client-api.md` around line 6319: This paragraph says ordinary botDM messages fall through and receive no fan-out, but `handleCreated` explicitly groups `RoomTypeDM` and `RoomTypeBotDM` and calls `publishDMEvents` for both. The same incorrect suppression claim also appears in `docs/client-api/events.md`. Update the documentation to describe the actual per-member botDM fan-out, or change the implementation if suppression is the intended contract.

```

</details>

chenjr0719:
Good catch on the botDM contract — you're right, and I've fixed it (`79116d49`).

Verified against `TestHandleUpdated_BotDMRoom_SkipsBotAccount` (asserts `pub.records == 1`, subject `UserRoomEvent("alice")`, "only the human recipient gets the live event"): a botDM message is handled via the DM path (`publishDMEvents`), which delivers the `RoomEvent` to each **non-bot** member and skips only the **bot** account (`isBot`) — for both `new_message` and `new_thread_message`. So "botDM receives no fan-out / falls through the default branch" was wrong; corrected in `docs/client-api.md` and both places in `docs/client-api/events.md`.

On the rollout point (thread replies now `new_thread_message` vs the old `new_message`): that's an intentional breaking wire change — it's called out in the PR description, and the frontend cutover is being coordinated to land with the deploy so old clients don't miss replies.

broadcast-worker/handler_test.go
```diff
@@ -2490,11 +2490,51 @@ func TestHandleThreadCreated_DMRoom_FansOutToAllMembers(t *testing.T) {
...
+	ctrl := gomock.NewController(t)
+	store := NewMockStore(ctrl)
+	us := NewMockUserStore(ctrl)
+	pub := &mockPublisher{}
+	keyStore := NewMockRoomKeyProvider(ctrl)
+
+	msgTime := time.Date(2026, 4, 1, 11, 0, 0, 0, time.UTC)
+
+	// botDM thread replies fan out per member (unlike an ordinary botDM new_message,
+	// which is suppressed) — the production branch handles RoomTypeBotDM too.
```
julianshen: _🧹 readability_ | _🔵 Suggestion_ | _`tgd-review`_

**The botDM test comment contradicts the implemented behavior.**

`handleCreated` explicitly routes `RoomTypeBotDM` through `publishDMEvents`, which skips only bot subscriptions and still sends ordinary `new_message` events to human members. The updated documentation also states this. Correct the comment so it does not claim that ordinary botDM fan-out is suppressed.

<details>

> ‼️ **IMPORTANT**

```suggestion
	// The production branch handles RoomTypeBotDM through the DM fan-out path;
	// thread replies must carry the distinct new_thread_message event type.
```

</details>

<details>

```
In `broadcast-worker/handler_test.go` around line 2509: `handleCreated` explicitly routes `RoomTypeBotDM` through `publishDMEvents`, which skips only bot subscriptions and still sends ordinary `new_message` events to human members. The updated documentation also states this. Correct the comment so it does not claim that ordinary botDM fan-out is suppressed.

```

</details>

chenjr0719:
Fixed in `6d6b5b8e` — that test comment was a leftover of the same wrong "suppressed" claim; corrected to match the delivery (human member gets the event, bot account skipped).

broadcast-worker/handler.go
```diff
@@ -262,6 +262,7 @@ func (h *Handler) handleThreadCreated(ctx context.Context, evt *model.MessageEve
 		// in the main channel, so a room-level mention badge would appear with no
 		// visible message to explain it.
```
julianshen, line 263: _🏷️ protocol-compatibility_ | _🔴 Blocking_ | _`distributed-system`_

**The new event type breaks existing thread-reply consumers.**

Previously deployed clients receive thread replies as `new_message`, as documented by the old API contract. After this producer is deployed, the same subject carries only `new_thread_message`; a client dispatching on the documented `type` discriminator will not recognize it and will drop the live update. This creates a concrete mixed-version failure during rollout and for clients that cannot upgrade atomically, although the reply remains persisted. Preserve `new_message` for the existing protocol and expose thread semantics through the already-present `threadParentMessageId` or an additive field, or introduce a versioned/negotiated delivery path so each client receives exactly one compatible event. Blindly publishing both types is unsafe because upgraded clients could process the reply twice.

<details>

```
In `broadcast-worker/handler.go` around line 263: Previously deployed clients receive thread replies as `new_message`, as documented by the old API contract. After this producer is deployed, the same subject carries only `new_thread_message`; a client dispatching on the documented `type` discriminator will not recognize it and will drop the live update. This creates a concrete mixed-version failure during rollout and for clients that cannot upgrade atomically, although the reply remains persisted. Preserve `new_message` for the existing protocol and expose thread semantics through the already-present `threadParentMessageId` or an additive field, or introduce a versioned/negotiated delivery path so each client receives exactly one compatible event. Blindly publishing both types is unsafe because upgraded clients could process the reply twice.

```

</details>

mliu33 (approved):
Thanks

PR #253, opened 2026-08-12
feat(chat-frontend): warm-start the sidebar from a subscription cache

hmchangw:
## Why

The sidebar has no memory. Every login and reload starts from an empty `state.subscriptions`, and nothing renders until three paginated `subscription.list` buckets drain.

The blank sidebar is also sticky on failure. `fetchSidebarBuckets` degrades a failed bucket to `[]`, and `fetchAllPages` returns the pages it already has when a later page fails — so the caller cannot tell *"this user has no rooms"* from *"the fetch failed"*. It dispatched `BUCKETS_LOADED` either way, and that action replaces `subscriptions` wholesale. A dropped connection overwrote a populated sidebar with an empty one.

## What changed

**Warm start.** `RoomEventsProvider` hydrates from `lib/subscriptionCache` inside `useReducer`'s lazy initializer, so the first paint already has rooms and sections. Hydration replays the cached data through the *same* `BUCKETS_LOADED` + `CHATLIST_LOADED` actions the network path uses, so cached and fetched state are identical in shape by construction — no parallel hydration path to drift. `summaries` are derived via `subToRoom` rather than persisted, which also stops `unreadCount` (a session-local counter no fetch ever corrects) from persisting a permanently stale value.

localStorage rather than IndexedDB is load-bearing: the read must be synchronous to run during the first render. An async store would push hydration into an effect and leave the empty frame exactly where it is.

**Failures stop being destructive.** `fetchSidebarBuckets` now reports `failures: SidebarBucket[]`, covering all three causes — a rejected RPC, a rejected later page, and a server that never clears `hasMore`. `useRoomSubscriptions` branches on it:

| Failed buckets | Behavior |
|---|---|
| 0 | Replace wholesale — the only path that may delete rooms |
| 1–2 | `BUCKETS_LOADED` with `merge: true`, **omitting** the failed buckets' id arrays (the reducer reads presence as "this bucket is authoritative") |
| 3 | `ROOMS_FAILED` only — cached rooms stay on screen under the error banner |

**Cache contents.** Subscriptions, the three bucket id sets, and the chatlist overlay, under one key stamped with the owning `account`/`siteId` so another account's record is ignored rather than rendered. `room.privateKey` is stripped before writing, so room E2E keys stay memory-only as they are today. Write-through is a 1s trailing debounce keyed by reference on those slices, guarded by an identity check against `initialState` — `RESET` returns that object, making the check an exact "we've been torn down" test so logout and StrictMode remounts cannot blank a good cache. Quota failures retry once with the 300 most recently active rooms, then drop the record.

## Notes for reviewers

- **Encrypted rooms warm-paint without keys.** A cached encrypted room appears in the sidebar immediately, but its messages show the decryption placeholder until the bootstrap re-seeds keys. Deliberate — persisting key material was rejected.
- **Message previews are not cached.** The sidebar-preview feature landed after this design; warm-painted rows show a blank snippet line until the fetch lands. Persisting them would put message text on disk in a cache that deliberately survives logout, so it is left out pending a call on that trade-off.
- **The cache survives logout,** keyed per account, so re-login is warm. A second account on the same machine overwrites the record, which bounds disk use at the cost of a cold start for the first account.
- **One existing test file changed:** `RoomEventsContext.test.jsx` now clears `localStorage` between tests — without it one test's persisted sidebar warm-starts the next test's provider.
- The design doc is included at `docs/superpowers/specs/2026-08-12-chat-frontend-subscription-cache-design.md`.

## Testing

Written test-first throughout — every batch was run and watched fail before implementing. New coverage: `lib/subscriptionCache.test.js` (round trip, `privateKey` stripping, version/identity/malformed rejection, quota retry, hostile storage), `reducer.test.js` merge mode (upsert, omitted-bucket retention, preview seeding, replace-still-deletes), `fetchSidebarBuckets/index.test.ts` failure reporting, and `RoomEventsContext.cache.test.jsx` (first-render paint, per-bucket failure retention, channel subs for cached rooms, teardown guard).

`npm test` 986 passing across 87 files · `npm run typecheck` clean · `vite build` clean.

### DIFF
--- a/chat-frontend/src/api/fetchSidebarBuckets/index.ts
+++ b/chat-frontend/src/api/fetchSidebarBuckets/index.ts
@@ -33,10 +33,19 @@ interface SidebarBucketReply {
   hasMore: boolean
 }
 
+/** Which of the three `subscription.list` buckets a result came from. */
+export type SidebarBucket = 'favorites' | 'apps' | 'rooms'
+
 export interface SidebarBuckets {
   favoriteIds: string[]
   appIds: string[]
   channelDmIds: string[]
+  /** Buckets that did NOT drain cleanly — an RPC rejected, a later page
+   *  failed, or the server never cleared `hasMore`. Their contents are
+   *  whatever we managed to fetch, so the caller must treat them as
+   *  incomplete and refuse to delete rooms on their behalf. Empty means
+   *  the whole bootstrap is authoritative. */
+  failures: SidebarBucket[]
   /** Per-roomId map of the full subscription record (DM variant typing
    *  covers both kinds — see SidebarBucketReply above). The reducer
    *  stores this directly under `state.subscriptions` so components
@@ -82,12 +91,15 @@ export async function fetchSidebarBuckets({ user, request }: Nats): Promise<Side
     fetchAllPages(request, subject, { type: 'apps' }),
     fetchAllPages(request, subject, { type: 'rooms' }),
   ])
+  const failures: SidebarBucket[] = []
   const unwrap = (
-    result: PromiseSettledResult<DMSubscription[]>,
+    result: PromiseSettledResult<BucketPages>,
+    bucket: SidebarBucket,
     label: string,
   ): DMSubscription[] => {
     if (result.status === 'fulfilled') {
-      return result.value
+      if (result.value.degraded) failures.push(bucket)
+      return result.value.subs
     }
     const err = result.reason
     console.warn(
@@ -96,11 +108,12 @@ export async function fetchSidebarBuckets({ user, request }: Nats): Promise<Side
       'FAILED:',
       err?.message ?? err,
     )
+    failures.push(bucket)
     return []
   }
-  const favSubs = unwrap(results[0], `${subject} {type:current,favorite:true}`)
-  const appSubs = unwrap(results[1], `${subject} {type:apps}`)
-  const roomSubs = unwrap(results[2], `${subject} {type:rooms}`)
+  const favSubs = unwrap(results[0], 'favorites', `${subject} {type:current,favorite:true}`)
+  const appSubs = unwrap(results[1], 'apps', `${subject} {type:apps}`)
+  const roomSubs = unwrap(results[2], 'rooms', `${subject} {type:rooms}`)
 
   const subscriptions: Record<string, DMSubscription> = {}
   const rooms: Room[] = []
@@ -128,9 +141,17 @@ export async function fetchSidebarBuckets({ user, request }: Nats): Promise<Side
     favoriteIds: favSubs.map((s) => s.roomId),
     appIds: appSubs.map((s) => s.roomId),
     channelDmIds: roomSubs.map((s) => s.roomId),
+    failures,
     subscriptions,
     rooms,
   }
+}
+
+/** One bucket's drained pages plus whether the drain completed. `degraded`
+ *  means `subs` is a partial view of that bucket. */
+interface BucketPages {
+  subs: DMSubscription[]
+  degraded: boolean
 }
 
 /** Drain one subscription bucket to completion. Requests successive pages
@@ -150,7 +171,7 @@ async function fetchAllPages(
   request: Nats['request'],
   subject: string,
   filter: Record<string, unknown>,
-): Promise<DMSubscription[]> {
+): Promise<BucketPages> {
   const all: DMSubscription[] = []
   for (let page = 0; page < MAX_PAGES; page++) {
     let reply: SidebarBucketReply
@@ -167,17 +188,17 @@ async function fetchAllPages(
         'FAILED:',
         (err as Error)?.message ?? err,
       )
-      return all
+      return { subs: all, degraded: true }
     }
     if (reply?.subscriptions?.length) all.push(...reply.subscriptions)
-    if (!reply?.hasMore) return all
+    if (!reply?.hasMore) return { subs: all, degraded: false }
   }
   console.warn(
     '[sidebar-bootstrap]',
     `${subject} ${JSON.stringify(filter)}`,
     `stopped after MAX_PAGES=${MAX_PAGES}; server kept signalling hasMore — bucket may be truncated`,
   )
-  return all
+  return { subs: all, degraded: true }
 }
 
 /** Derive a `Room` from a subscription record — the ONE wire→Room mapper
--- a/chat-frontend/src/context/RoomEventsContext/reducer.js
+++ b/chat-frontend/src/context/RoomEventsContext/reducer.js
@@ -354,6 +354,47 @@ export function roomEventsReducer(state, action) {
     }
     case 'BUCKETS_LOADED': {
       const subs = action.subscriptions ?? {}
+      // Seed only rooms with no preview yet. A live message can land before
+      // fetchSidebarBuckets resolves (the DM subscription goes live first), and
+      // that message is NEWER than this list snapshot — overwriting it would show
+      // an older message in the sidebar than the room itself displays.
+      // Shared by both paths: a degraded bootstrap still carries real previews
+      // for the buckets it did reach.
+      let previews = state.previews
+      for (const [roomId, sub] of Object.entries(subs)) {
+        if (previews[roomId]) continue
+        const preview = previewFromWire(sub?.room?.previewMessage)
+        if (!preview) continue
+        if (previews === state.previews) previews = { ...state.previews }
+        previews[roomId] = preview
+      }
+      if (action.merge) {
+        // Degraded-bootstrap path: one or two `subscription.list` buckets
+        // failed, so this payload is INCOMPLETE and must not be read as
+        // "everything else is gone". Upsert what arrived and leave the
+        // rest — only a complete fetch (replace, below) may delete.
+        const byId = new Map(state.summaries.map((s) => [s.id, s]))
+        for (const r of action.rooms ?? []) {
+          const prev = byId.get(r.id)
+          // Room metadata is server-fresh; unread bookkeeping is
+          // session-local and the payload has no opinion on it.
+          const base = prev
+            ? { ...toSummary(r), unreadCount: prev.unreadCount, hasMention: prev.hasMention, mentionAll: prev.mentionAll }
+            : toSummary(r)
+          byId.set(r.id, mergeSubscriptionIntoSummary(base, subs[r.id]))
+        }
+        return {
+          ...state,
+          summaries: sortByLastMsgDesc([...byId.values()]),
+          // A failed bucket is OMITTED by the caller rather than sent
+          // empty, so presence alone marks a bucket authoritative.
+          favoriteIds: action.favoriteIds ? new Set(action.favoriteIds) : state.favoriteIds,
+          appIds: action.appIds ? new Set(action.appIds) : state.appIds,
+          channelDmIds: action.channelDmIds ? new Set(action.channelDmIds) : state.channelDmIds,
+          subscriptions: { ...state.subscriptions, ...subs },
+          previews,
+        }
+      }
       let summaries
       if (action.rooms) {
         // Cold-start path: rooms are derived from the three subscription
@@ -372,18 +413,6 @@ export function roomEventsReducer(state, action) {
         summaries = state.summaries.map((s) =>
           subs[s.id] ? mergeSubscriptionIntoSummary(s, subs[s.id]) : s
         )
-      }
-      // Seed only rooms with no preview yet. A live message can land before
-      // fetchSidebarBuckets resolves (the DM subscription goes live first), and
-      // that message is NEWER than this list snapshot — overwriting it would show
-      // an older message in the sidebar than the room itself displays.
-      let previews = state.previews
-      for (const [roomId, sub] of Object.entries(subs)) {
-        if (previews[roomId]) continue
-        const preview = previewFromWire(sub?.room?.previewMessage)
-        if (!preview) continue
-        if (previews === state.previews) previews = { ...state.previews }
-        previews[roomId] = preview
       }
       return {
         ...state,
--- a/chat-frontend/src/context/RoomEventsContext/useRoomSubscriptions.js
+++ b/chat-frontend/src/context/RoomEventsContext/useRoomSubscriptions.js
@@ -19,6 +19,27 @@ import {
  *  the server's lastSeenAt for the active user stays current. */
 const MARK_READ_DEBOUNCE_MS = 500
 
+/** Which `SidebarBuckets` id-array each bucket owns. A bucket that failed
+ *  is OMITTED from the merge action rather than sent empty — the reducer
+ *  reads presence as "this bucket is authoritative". */
+const BUCKET_ID_FIELD = {
+  favorites: 'favoriteIds',
+  apps: 'appIds',
+  rooms: 'channelDmIds',
+}
+
+const BUCKET_COUNT = Object.keys(BUCKET_ID_FIELD).length
+
+/** Build the BUCKETS_LOADED action for a bootstrap result. A clean fetch
+ *  replaces wholesale (so rooms left while away disappear); a partial one
+ *  merges and drops the failed buckets' id arrays. */
+function bucketsLoadedAction({ failures: _failures, ...buckets }, failures) {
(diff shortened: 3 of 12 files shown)


julianshen:
**Actionable comments posted: 9** — 🔴 1 blocking · 🟠 8 warning

<details>
<summary>📒 Files reviewed (12)</summary>

* `chat-frontend/CLAUDE.md`
* `chat-frontend/src/api/fetchSidebarBuckets/index.test.ts`
* `chat-frontend/src/api/fetchSidebarBuckets/index.ts`
* `chat-frontend/src/context/RoomEventsContext/RoomEventsContext.cache.test.jsx`
* `chat-frontend/src/context/RoomEventsContext/RoomEventsContext.test.jsx`
* `chat-frontend/src/context/RoomEventsContext/useRoomSubscriptions.js`
* `chat-frontend/src/lib/subscriptionCache.js`
* `chat-frontend/src/lib/subscriptionCache.test.js`
* `docs/superpowers/specs/2026-08-12-chat-frontend-subscription-cache-design.md`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `silent-failure`
* `tgd-review`

</details>

chat-frontend/src/api/fetchSidebarBuckets/index.ts
```diff
@@ -167,17 +188,17 @@ async function fetchAllPages(
         'FAILED:',
         (err as Error)?.message ?? err,
       )
-      return all
+      return { subs: all, degraded: true }
     }
```
julianshen, line 192: _🎯 correctness_ | _🔴 Blocking_ | _`nats`_

**Malformed replies are treated as authoritative empty pages.**

Because `!reply?.hasMore` is true for `undefined`, `null`, or an object missing `hasMore`, a malformed `subscription.list` reply returns `degraded: false`. All three buckets can consequently report no failures, causing `BUCKETS_LOADED` to replace cached subscriptions with empty data and report success. Runtime-validate that `subscriptions` is an array and `hasMore` is a boolean before consuming the reply; malformed replies must return `degraded: true` (or reject) and should have a regression test.

<details>

```
In `chat-frontend/src/api/fetchSidebarBuckets/index.ts` around line 192: Because `!reply?.hasMore` is true for `undefined`, `null`, or an object missing `hasMore`, a malformed `subscription.list` reply returns `degraded: false`. All three buckets can consequently report no failures, causing `BUCKETS_LOADED` to replace cached subscriptions with empty data and report success. Runtime-validate that `subscriptions` is an array and `hasMore` is a boolean before consuming the reply; malformed replies must return `degraded: true` (or reject) and should have a regression test.

```

</details>

chat-frontend/src/context/RoomEventsContext/useRoomSubscriptions.js
```diff
@@ -19,6 +19,27 @@ import {
...
+
+const BUCKET_COUNT = Object.keys(BUCKET_ID_FIELD).length
+
+/** Build the BUCKETS_LOADED action for a bootstrap result. A clean fetch
+ *  replaces wholesale (so rooms left while away disappear); a partial one
+ *  merges and drops the failed buckets' id arrays. */
+function bucketsLoadedAction({ failures: _failures, ...buckets }, failures) {
+  if (failures.length === 0) return { type: 'BUCKETS_LOADED', ...buckets }
+  const action = { type: 'BUCKETS_LOADED', merge: true, ...buckets }
+  for (const bucket of failures) delete action[BUCKET_ID_FIELD[bucket]]
```
julianshen, line 39: _🏷️ consistency_ | _🟠 Warning_ | _`distributed-system`_

**Partial-page IDs are discarded for a failed bucket.**

A later-page failure retains the pages already fetched, but this loop deletes that bucket's ID array from the action. For example, on a cold start the rooms bucket can return r1 on page 1 and time out on page 2; r1 is added to subscriptions and summaries, while channelDmIds remains empty. The persisted slices are then internally inconsistent, and consumers deriving placement from the bucket sets cannot classify the successfully fetched room. Preserve safety by unioning a degraded bucket's partial IDs with its previous set rather than omitting all of them; successful buckets should still replace their sets.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/useRoomSubscriptions.js` around line 39: A later-page failure retains the pages already fetched, but this loop deletes that bucket's ID array from the action. For example, on a cold start the rooms bucket can return r1 on page 1 and time out on page 2; r1 is added to subscriptions and summaries, while channelDmIds remains empty. The persisted slices are then internally inconsistent, and consumers deriving placement from the bucket sets cannot classify the successfully fetched room. Preserve safety by unioning a degraded bucket's partial IDs with its previous set rather than omitting all of them; successful buckets should still replace their sets.

```

</details>

chat-frontend/src/context/RoomEventsContext/useRoomSubscriptions.js
```diff
@@ -493,17 +514,37 @@ export function useRoomSubscriptions(
...
         // prior login must not seed keys or open subs into the new session.
         if (!isCurrent()) return
+
+        // A degraded bucket returns whatever it managed to fetch, which is
+        // indistinguishable from "the user has no rooms" — so an incomplete
+        // result must never be allowed to delete anything. Nothing at all
+        // came back: keep what's on screen (cache or a prior fetch) and
+        // report the failure instead of blanking the sidebar.
+        const failures = buckets.failures ?? []
+        if (failures.length === BUCKET_COUNT) {
```
julianshen, line 570: _🏷️ failure-handling_ | _🟠 Warning_ | _`distributed-system`_

**All degraded buckets discard successfully fetched pages.**

failures.length === BUCKET_COUNT does not mean that nothing was returned: fetchAllPages marks a bucket degraded after a later page fails while retaining earlier pages. If all three buckets fetch page 1 and then time out, this branch discards every partial result and returns. The bootstrap runs once per login cycle, so a cold client remains empty and a warm client ignores all newly fetched changes until reload. Apply useful partial data with merge semantics, surface the degraded status, and schedule a bounded, generation-checked retry with backoff to obtain the missing pages.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/useRoomSubscriptions.js` around line 544: failures.length === BUCKET_COUNT does not mean that nothing was returned: fetchAllPages marks a bucket degraded after a later page fails while retaining earlier pages. If all three buckets fetch page 1 and then time out, this branch discards every partial result and returns. The bootstrap runs once per login cycle, so a cold client remains empty and a warm client ignores all newly fetched changes until reload. Apply useful partial data with merge semantics, surface the degraded status, and schedule a bounded, generation-checked retry with backoff to obtain the missing pages.

```

</details>

chat-frontend/src/context/RoomEventsContext/reducer.js
```diff
@@ -264,6 +264,32 @@ export function roomEventsReducer(state, action) {
...
+          byId.set(r.id, mergeSubscriptionIntoSummary(base, subs[r.id]))
+        }
+        return {
+          ...state,
+          summaries: sortByLastMsgDesc([...byId.values()]),
+          // A failed bucket is OMITTED by the caller rather than sent
+          // empty, so presence alone marks a bucket authoritative.
+          favoriteIds: action.favoriteIds ? new Set(action.favoriteIds) : state.favoriteIds,
+          appIds: action.appIds ? new Set(action.appIds) : state.appIds,
+          channelDmIds: action.channelDmIds ? new Set(action.channelDmIds) : state.channelDmIds,
```
julianshen, line 393: _🏷️ consistency_ | _🟠 Warning_ | _`distributed-system`_

**An unrelated bucket failure preserves memberships proven stale.**

Merge mode unions every old subscription and summary even when their owning bucket completed authoritatively. For example, cached channel R was left while offline; on login the rooms bucket succeeds without R, but the unrelated apps bucket times out. Because the whole action uses merge mode, R remains in the canonical subscriptions map and summaries even though channelDmIds was authoritatively replaced without it. The warm-start path has also already opened R's channel subscription. Reconcile buckets independently: preserve old and partial members only for degraded buckets, remove records proven absent by successful membership buckets, and close any corresponding live channel subscription.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 289: Merge mode unions every old subscription and summary even when their owning bucket completed authoritatively. For example, cached channel R was left while offline; on login the rooms bucket succeeds without R, but the unrelated apps bucket times out. Because the whole action uses merge mode, R remains in the canonical subscriptions map and summaries even though channelDmIds was authoritatively replaced without it. The warm-start path has also already opened R's channel subscription. Reconcile buckets independently: preserve old and partial members only for degraded buckets, remove records proven absent by successful membership buckets, and close any corresponding live channel subscription.

```

</details>

julianshen, line 393: _🏷️ ordering_ | _🟠 Warning_ | _`distributed-system`_

**A delayed bootstrap snapshot can overwrite newer live updates.**

Live subscription handlers are installed before fetchSidebarBuckets resolves. If the RPC snapshot contains an old subscription, then a subscription.update event applies a newer role, name, or alert value before the RPC response arrives, this spread writes the older RPC value over the live update. The same race affects summaries and authoritative bucket sets, and no later event is guaranteed to repair the state. Buffer subscription deltas received during bootstrap and replay them after applying the snapshot, or attach monotonic revisions and reject snapshot records older than the current state.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 289: Live subscription handlers are installed before fetchSidebarBuckets resolves. If the RPC snapshot contains an old subscription, then a subscription.update event applies a newer role, name, or alert value before the RPC response arrives, this spread writes the older RPC value over the live update. The same race affects summaries and authoritative bucket sets, and no later event is guaranteed to repair the state. Buffer subscription deltas received during bootstrap and replay them after applying the snapshot, or attach monotonic revisions and reject snapshot records older than the current state.

```

</details>

chat-frontend/src/lib/subscriptionCache.js
```diff
@@ -0,0 +1,162 @@
...
+    ),
+    favoriteIds: toIdArray(favoriteIds),
+    appIds: toIdArray(appIds),
+    channelDmIds: toIdArray(channelDmIds),
+    chatlist,
+  }
+  try {
+    write(record)
+    return
+  } catch {
```
julianshen, line 109: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Failed cache writes repeatedly serialize and sort the full cache**

When storage is disabled or the full record exceeds quota, every subsequent write-through rebuilds and serializes the full subscription map, attempts the same synchronous failing localStorage write, then potentially sorts all subscriptions and serializes a trimmed record. A successful trimmed retry is not remembered, and non-quota errors take the same retry path. For large accounts receiving recurring subscription updates, this repeatedly blocks the browser main thread and creates substantial temporary allocations. Only retry quota errors, remember a quota-constrained mode for later writes, and disable further writes for the session after persistent/non-quota storage failures.

<details>

```
In `chat-frontend/src/lib/subscriptionCache.js` around line 109: When storage is disabled or the full record exceeds quota, every subsequent write-through rebuilds and serializes the full subscription map, attempts the same synchronous failing localStorage write, then potentially sorts all subscriptions and serializes a trimmed record. A successful trimmed retry is not remembered, and non-quota errors take the same retry path. For large accounts receiving recurring subscription updates, this repeatedly blocks the browser main thread and creates substantial temporary allocations. Only retry quota errors, remember a quota-constrained mode for later writes, and disable further writes for the session after persistent/non-quota storage failures.

```

</details>

chat-frontend/src/context/RoomEventsContext/useRoomSubscriptions.js
```diff
@@ -516,7 +557,7 @@ export function useRoomSubscriptions(
             if (buckets.subscriptions[roomId]) buckets.subscriptions[roomId].favorite = true
           }
         }
-        safeDispatch({ type: 'BUCKETS_LOADED', ...buckets })
+        safeDispatch(bucketsLoadedAction(buckets, failures))
```
julianshen, line 586: _🏷️ error-handling_ | _🟠 Warning_ | _`silent-failure`_

**Partial bucket failures are presented as a successful bootstrap.**

When one or two bucket requests fail, `fetchSidebarBuckets` records the failures but this path only dispatches the merged `BUCKETS_LOADED` action. The failures remain visible only through `console.warn`; a cold start can silently omit rooms from the failed bucket, while a warm start can show stale membership, with no error exposed through `useRoomSummaries()`. After applying the partial data, dispatch a non-destructive failure action so the degraded result is surfaced just as a total failure is.

<details>

> ‼️ **IMPORTANT**

```suggestion
        safeDispatch(bucketsLoadedAction(buckets, failures))
        if (failures.length > 0) {
          safeDispatch({ type: 'ROOMS_FAILED', error: 'Could not load all of your chats.' })
        }
```

</details>

<details>

```
In `chat-frontend/src/context/RoomEventsContext/useRoomSubscriptions.js` around line 560: When one or two bucket requests fail, `fetchSidebarBuckets` records the failures but this path only dispatches the merged `BUCKETS_LOADED` action. The failures remain visible only through `console.warn`; a cold start can silently omit rooms from the failed bucket, while a warm start can show stale membership, with no error exposed through `useRoomSummaries()`. After applying the partial data, dispatch a non-destructive failure action so the degraded result is surfaced just as a total failure is.

```

</details>

chat-frontend/src/context/RoomEventsContext/reducer.js
julianshen, line 393: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Merge-mode loads never clear the loading or error state.**

This early return bypasses the normal BUCKETS_LOADED status reset. Because a partial bootstrap terminates by dispatching this merge action, `roomsLoading` can remain true indefinitely, and an error from an earlier attempt can remain visible despite applying new data. Mark the load complete and clear the stale error here.

<details>

> ‼️ **IMPORTANT**

```suggestion
          subscriptions: { ...state.subscriptions, ...subs },
          roomsLoading: false,
          roomsError: null,
```

</details>

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 289: This early return bypasses the normal BUCKETS_LOADED status reset. Because a partial bootstrap terminates by dispatching this merge action, `roomsLoading` can remain true indefinitely, and an error from an earlier attempt can remain visible despite applying new data. Mark the load complete and clear the stale error here.

```

</details>

chat-frontend/src/lib/subscriptionCache.js
```diff
@@ -0,0 +1,162 @@
...
+  try {
+    const raw = window.localStorage.getItem(CACHE_KEY)
+    if (!raw) return null
+    record = JSON.parse(raw)
+  } catch {
+    return null
+  }
+  if (record?.v !== CACHE_VERSION) return null
+  if (record.account !== user.account || record.siteId !== user.siteId) return null
+  if (!record.subscriptions || typeof record.subscriptions !== 'object') return null
```
julianshen, line 143: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Valid JSON with malformed subscriptions can crash hydration.**

The only shape check accepts arrays and arbitrary objects, including entries such as `{ "r1": null }`; `chatlist` is not validated either. `hydrateFromCache` immediately passes every value to `subToRoom` and then feeds the cached structures into reducers during render, so malformed but valid localStorage content can throw or create invalid first-render state. Validate that `subscriptions` is a plain record whose values have the required room fields and validate `chatlist`, or catch the complete mapping/reducer hydration path and fall back to `initialState`. Add tests for malformed subscription entries and chatlist values.

<details>

```
In `chat-frontend/src/lib/subscriptionCache.js` around line 143: The only shape check accepts arrays and arbitrary objects, including entries such as `{ "r1": null }`; `chatlist` is not validated either. `hydrateFromCache` immediately passes every value to `subToRoom` and then feeds the cached structures into reducers during render, so malformed but valid localStorage content can throw or create invalid first-render state. Validate that `subscriptions` is a plain record whose values have the required room fields and validate `chatlist`, or catch the complete mapping/reducer hydration path and fall back to `initialState`. Add tests for malformed subscription entries and chatlist values.

```

</details>

yenta (approved):

PR #249, opened 2026-08-12
Cross-site failover: SP2–SP4 design cycles + SP4-core failover state (portal-service)

hmchangw:
## What this is

Downstream work on the cross-site failover ("lifeboat") program: the design cycles for SP2/SP3/SP4, plus the **first buildable slice implemented** — SP4-core in `portal-service`.

**Base branch:** this PR targets `claude/cross-site-failover-brainstorm-luol1d` (where the governing design + SP1 live) so the diff is *only* the downstream increment. SP1 is being implemented separately and is untouched here.

## Specs (design cycles)

- **SP2 rewritten to "World 1".** Production runs **one shared org-level NATS account** (confirmed), so the backup mints in that same account — not impersonation, no per-site keys. This **drops the entire key-custody / remote-signer / Vault design** (a World-2 concern) and shrinks SP2 to config/ops: the SP0-coupled `chat.local.room.>` grant + a backup `auth-service` deployment. No new minting code.
- **SP3 spec** — portal health-aware routing override (consumes SP4's `servingTarget` signal).
- **SP4 spec rewritten** — operator-confirmed posture; health probe deferred. SP4-core = the operator-driven `FailoverState` machine + control surface + `servingTarget` signal.
- **Reconciliation** — the governing design (§4.3, §11.4, failure-modes) and roadmap updated so the invalidated key-custody framing no longer misleads the next planner.

## Implemented — SP4-core (`portal-service`)

Operator-driven failover state; **no auto-detection** (deferred), no UI yet (a per-operator login/console is a planned follow-up — the auth is built as a middleware seam so that drops in without a rewrite).

- **State machine** — `healthy → failed_over → failing_back → healthy`, plus `resume` (false alarm). `servingTarget` (home|backup) is derived and is the entire SP3 contract.
- **Mongo CAS store** — one `failover_states` doc per site; portal is the sole writer, optimistic-concurrency guarded on `version` (the split-brain fence). Unknown sites synthesize a healthy default.
- **Control surface** — `GET`/`GET`/`POST /internal/v1/failover[/:siteId]` on an **internal-only listener** (kept off the public discovery server), gated by a dedicated ops bearer token (constant-time compare, never logged); `operator`/`reason` recorded for audit; `409` on illegal transition and version conflict.
- **Wiring** — internal server starts only when `FAILOVER_OPS_TOKEN` is set; fail-fast bind; graceful shutdown alongside the public server.

## Verification

- `make test SERVICE=portal-service` — **PASS** (race detector).
[...]

PR #256, opened 2026-08-12
Cross-site failover: SP3 — portal health-aware routing override

hmchangw:
## What this is

SP3 of the cross-site failover ("lifeboat") program — the routing brain that makes SP4 *act*: when a site is failed over, `portal-service` hands its SSO users the **backup's** connection coordinates so they reconnect there.

**Stacked PR:** targets `claude/cross-site-failover-downstream` (the SP4 branch, PR #249), so this diff is only the SP3 increment. SP3 consumes the `FailoverStore` / `servingTarget` that SP4 built.

## Status

Design settled via a collaborative brainstorm; **spec committed, implementation to follow on this branch.** Opening as draft now so the work is visible and durable.

## Decisions (locked in the brainstorm)

- **Thin override at portal's `resolve()` seam**, keyed entirely on SP4's `servingTarget` (home|backup). No new service, no new client protocol — the frontend's existing reconnect-on-failure path is the trigger.
- **Only the URLs swap; `siteId` stays home.** The backup materializes per-origin-`siteId` namespaces and `msg.send` subjects carry `{siteId}`, so the displaced client must keep reporting its home site or its messages land in the wrong namespace. This is the load-bearing correctness point.
- **Scoped to the SSO `/api/userInfo` path only.** The bot/admin `/api/v1/login` path is left unchanged.
- **Backup coordinates = a reserved entry** in `PORTAL_SITE_URLS` (`PORTAL_BACKUP_SITE_ID`), reusing existing boot-time validation.
- **Failback reconnect = no SP3 code** — the backup's drain on flip-to-healthy is itself the nudge.
- **TTL-cached `failoverReader`** built here over SP4's store (fail-safe `home` on miss/error); `main.go` refactored to construct the store unconditionally (routing needs it even where the control surface is off).

## ⚠️ Deferred but business-critical: bot failover

Bots are **critical**, and excluding them here is a deliberate deferral because the current lifeboat scope is already large — not a decision that they don't matter. The bot/admin `/api/v1/login` path forwards login to the home-site botplatform (down during an outage), so rerouting its URLs is moot until a **backup botplatform** exists. The real weight is in SP1 (materialize bot/botplatform state) + SP2 (run botplatform at the backup, wire the backup auth-service session branch); the SP3 side is then a single `servingURLs(...)` call added to `HandleLogin`. Recorded in the spec and roadmap so it is not lost.

## Spec

`docs/superpowers/specs/2026-08-11-sp3-portal-routing-override.md` (rewritten for these decisions).

https://claude.ai/code/session_01Bx1yBMCjo88ftnTB6ha5pA

---
_Generated by [Claude Code](https://claude.ai/code/session_01Bx1yBMCjo88ftnTB6ha5pA)_

### DIFF
--- a/docs/superpowers/plans/2026-08-12-sp6-ops-runbook.md
+++ b/docs/superpowers/plans/2026-08-12-sp6-ops-runbook.md
@@ -1,0 +1,231 @@
+# SP6 — Cross-Site Failover: Ops / IaC Runbook
+
+> **What this is.** SP6 of the cross-site failover ("lifeboat") program is a
+> **runbook coordinated with the platform/NATS/IaC team — not a TDD code plan**
+> (roadmap §SP6). This document consolidates every operational requirement, and
+> documents the operator-facing surface of the parts already shipped (SP3, SP4).
+>
+> - **Governing design:** `docs/superpowers/specs/2026-07-28-cross-site-failover-design.md` (§4, §6.5, §8, §9)
+> - **Roadmap:** `docs/superpowers/plans/2026-07-28-cross-site-failover-roadmap.md` (SP6)
+> - **Related shipped work:** SP4 (failover state + control surface, `portal-service`), SP3 (routing override), SP2 (identity — World 1).
+
+## Readiness legend
+
+Each item is tagged with what an operator can action **today** vs. what is gated:
+
+- **[READY]** — actionable now against shipped code.
+- **[GATED:SP1]** / **[GATED:SP1+SP2]** — needs the backup materialization / serving stack to exist first.
+- **[INFRA]** — external platform/IaC work (Kubernetes, NATS supercluster, cloud), owned by the platform team; this repo carries no code for it.
+
+---
+
+## 1. The failover control surface  **[READY]**
+
+Shipped in SP4 (`portal-service`). This is how an operator drives a failover and
+failback. Portal is the **single routing authority and split-brain fence** — it
+is the only place the decision is made.
+
+### 1.1 Configuration
+
+| Env | Where | Meaning | Ops action |
+|---|---|---|---|
+| `FAILOVER_OPS_TOKEN` | portal-service | Break-glass bearer token gating the control surface. **Empty disables the control surface entirely.** | Provision as a secret (below). Required to operate failover. |
+| `FAILOVER_INTERNAL_ADDR` | portal-service | Listen address for the internal-only control listener (default `:8090`), kept **off** the public `:8085` discovery API. | Expose only to the ops network (below). |
+| `PORTAL_BACKUP_SITE_ID` | portal-service | Reserved id in `PORTAL_SITE_URLS` whose `{baseUrl,natsUrl}` are served for a failed-over site (e.g. `_backup`). Empty in single-site deployments. | Set once the backup site exists; add its entry to `PORTAL_SITE_URLS`. |
+| `FAILOVER_STATE_TTL` | portal-service | How long portal caches a site's serving target before re-reading (default `5s`). | Leave default unless routing-propagation latency needs tuning. |
+| `PORTAL_SITE_URLS` | portal-service | Site registry; **must contain the `PORTAL_BACKUP_SITE_ID` entry** or a failover returns `500` (loud, by design). | Add the backup entry when the backup is deployed. |
+
+### 1.2 Secret provisioning — `FAILOVER_OPS_TOKEN`
+
+- Generate a high-entropy random token; store in the platform secret manager
+  (the same mechanism that delivers other portal secrets).
+- Deliver to portal-service as an env var / mounted secret. **Never** commit it;
+  the local `deploy/docker-compose.yml` value (`dev-failover-token`) is dev-only.
+- Rotation: replace the secret and restart/roll portal. There is no session
+  state to migrate — the token gates each request independently.
+- The token authenticates *possession*, not a person; the `operator` field in
+  each request is self-asserted for the audit trail. A per-operator login is a
+  planned follow-up (SP4 spec §8, "option 3") and is **not** in scope here.
+
+### 1.3 Network policy — internal listener
+
+- The control routes live on `FAILOVER_INTERNAL_ADDR` (a **separate** HTTP
+  listener from the public discovery API on `:8085`). Restrict it to the ops
+  network / an internal LB — it must **not** be internet-reachable.
+- Portal's public `:8085` surface is unchanged and carries no privileged writes.
+
+### 1.4 Operator procedures
+
+Control endpoints (all require `Authorization: Bearer $FAILOVER_OPS_TOKEN`):
+
+```
+GET  /internal/v1/failover            # list every site's failover state
+GET  /internal/v1/failover/{siteId}   # one site's state
+POST /internal/v1/failover/{siteId}   # {"action": "...", "operator": "...", "reason": "..."}
+```
+
+State machine (operator-driven; every write is CAS-guarded, so concurrent portal
+replicas cannot diverge):
+
+```
+healthy ──failover──► failed_over ──failback──► failing_back ──complete──► healthy
+   ▲                       │
+   └──────── resume ───────┘   (false alarm; skip the drain)
+```
+
+`servingTarget` is `backup` while `failed_over`/`failing_back`, else `home`.
+
+**Declare a site down (fail over):**
+```bash
+curl -sS -H "Authorization: Bearer $FAILOVER_OPS_TOKEN" \
+  -X POST https://portal-internal/internal/v1/failover/site-a \
+  -d '{"action":"failover","operator":"jane","reason":"site-a NATS unreachable"}'
+```
+Effect: `site-a` accounts' `/api/userInfo` now return the backup's coordinates;
+displaced clients reconnect there on their next connection failure.
+
+**Begin failback (start draining home):** `{"action":"failback",...}` → moves to
+`failing_back`. **`servingTarget` stays `backup`** — users keep writing to a
+stable backup while the home site drains. *(The drain/replay itself is SP5.)*
+
+**Complete failback (flip home):** `{"action":"complete",...}` → `healthy`;
+`servingTarget` flips to `home`; the backup drains its impersonation and clients
+reconnect home. **Gate this on SP5's lag≈0 signal** once SP5 exists; today it is a
+manual operator action.
+
+**False alarm (undo a failover):** `{"action":"resume",...}` → straight back to
+`healthy`, skipping the drain.
+
+**Inspect:** `GET /internal/v1/failover` for the whole fleet's state, including
+`operator`/`reason`/`since` for the audit trail.
+
+### 1.5 Split-brain guarantee
+
+A site is `home` **xor** `backup` because the decision is a single, single-valued
+`FailoverState` document that only portal writes (CAS-guarded). Do **not**
+attempt to route around portal or edit the `failover_states` collection by hand —
+that is the one thing that can break the fence.
+
+---
+
+## 2. Backup identity & auth-service deploy  **[READY (config) / GATED:SP1+SP2 (serving)]**
+
+Identity is **resolved (World 1)** — one shared org NATS account, so the backup
+mints with the *same* signing identity every site uses. No per-site keys, no KMS
+scheme (see `specs/2026-08-11-sp2-backup-identity-jwt-minting.md`).
+
+Backup `auth-service` deployment requirements:
+
+- **[READY]** Configure with the **shared** `AUTH_SCOPED_SIGNING_KEY` +
+  `AUTH_ACCOUNT_PUB_KEY` — identical to every site. No new key material.
+- **[READY]** The `chat.local.room.>` subscribe grant is **already on `main`** in
+  the shared scoped-signing template (SP0). Nothing to add; the backup inherits it.
+- **[READY]** Point OIDC at the central Keycloak (reachable from the backup).
+- **[READY]** Leave `BOTPLATFORM_URL` **unset** on the backup — bot/admin login is
+  out of lifeboat scope, so the session-token branch need not work there.
+- **[READY]** Size the backup `auth-service` (and Keycloak reachability) for **one
+  largest site's reconnect peak** — a whole site re-mints at once on failover.
+  Client reconnect backoff/jitter + proactive-refresh jitter already spread it
+  (`docs/superpowers/specs/2026-06-05-seamless-nats-jwt-refresh-design.md`).
+- **[INFRA]** The backup's NATS `resolver` must preload the same shared account
+  JWT so it validates the minted user JWTs (it does, being a supercluster peer).
+
+The serving *handlers* (send/receive/history against the materialized copy) are
+**[GATED:SP1+SP2]** — they need the materialized data to exist.
+
+---
+
+## 3. Backup platform footprint  **[INFRA]**
+
+Owned by the platform/NATS/IaC team; no code in this repo.
+
+- **HA within itself (multi-AZ).** The backup is a failover SPOF — if it is down
+  when a site fails, there is no lifeboat. It **must** be HA. Documented limit,
+  not silently degraded (design §9).
+- **Supercluster gateway peer.** The backup must be a full gateway peer so that,
+  during a failover, **global-room** delivery reaches members still on healthy
+  sites (design §7).
+- **Leaf-node `chat.local.>` deny on the backup's leaf**, same as every site, so
+  local-room subjects stay site-local. Displaced clients connect intra-cluster on
+  the backup, so this does not affect their local-room delivery (design §7).
+- **Per-`siteID` namespacing** of the backup's Mongo/Cassandra (the exact scheme —
+  DB-per-site vs prefixed collections/keyspaces — is decided in SP1a/SP1b).
+
+---
+
+## 4. Streams & retention  **[GATED:SP1]**
+
+- **Restore-log `MaxAge`.** The backup's canonical stream is the durable restore
+  log for failback. Size its `MaxAge` **≥ the maximum tolerated outage**
+  (days/weeks — it is 1× per message, cheap). Do **not** rely on the 72h
+  Cassandra read window for restore; a longer outage would age its earliest
+  messages out (design §6.5).
+- **DR feed streams** (`DR_MESSAGES_{siteID}`, `DR_OPLOG_{siteID}`, per SP1a/SP1b)
+  and the backup's INBOX — owned/bootstrapped per the `BOOTSTRAP_STREAMS`
+  convention; ops/IaC owns creation in production.
+- **Bucket-window parity.** The backup and every origin site **must** share
+  `MESSAGE_BUCKET_HOURS` (default 72). A mismatch silently mis-partitions writes
+  and reads (CLAUDE.md §Cassandra; design §6.4). Enforce this in the backup's
+  deploy config.
+
+---
+
+## 5. Monitoring & alerting  **[GATED:SP1]**
+
+- **Per-site replication lag — the RPO-decay signal (highest-value alert).** If
+  the backup cannot keep up with the N-site aggregate ingest, its lag grows and
+  the DR guarantee *quietly* weakens — **the lag is the live RPO** (design §8).
+  Alert per origin site. NOTE: there is **no existing lag metric to reuse today**
+  (the design's "reuse the membership-federation lag signal" is aspirational —
+  verified: no such metric in the repo). This must be **built as part of SP1**
+  (emit a per-site replication-lag gauge from the backup materializers) and wired
+  to alerting here.
+- **Backup capacity — two separate numbers.** Size/alert the **ingest** tier for
+  the N-site aggregate (cheap 1× plane) and the **serving** tier for one largest
+  site's peak with ×F fan-out (design §8). They are different capacity budgets.
+- **Multi-site-outage ceiling.** The backup is sized for **one site down at a
+  time**; alert on concurrent multi-site failure as a documented degradation
+  ceiling (design §9), not a silent failure.
+
+---
+
+## 6. Failback operational must-dos  **[GATED:SP1+SP2, wired by SP5]**
+
+When SP5 replays the outage log home, the platform config must ensure:
+
+- **Suppress push notifications** for replayed messages — they are stale
+  (`notification-worker` skips events flagged as replay; design §6.6).
+- **Suppress / client-dedup re-broadcast** — restore is about durability in the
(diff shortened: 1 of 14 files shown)


julianshen:
**17 findings · 10 unique issues · 10 inline comments posted** — 🔴 9 blocking · 🟠 8 warning

### Related work

- [PR #249](https://github.com/hmchangw/newchat/pull/249) — Cross-site failover: SP2–SP4 design cycles \+ SP4-core failover state \(portal-service\) (open)

<details>

* `docs/client-api.md`
* `docs/superpowers/plans/2026-07-28-cross-site-failover-roadmap.md`
* `docs/superpowers/plans/2026-08-12-sp3-portal-routing-override.md`
* `docs/superpowers/plans/2026-08-12-sp6-ops-runbook.md`
* `docs/superpowers/specs/2026-08-11-sp2-backup-identity-jwt-minting.md`
* `docs/superpowers/specs/2026-08-11-sp3-portal-routing-override.md`
* `portal-service/deploy/docker-compose.yml`
* `portal-service/failover_reader.go`
* `portal-service/failover_reader_integration_test.go`
* `portal-service/failover_reader_test.go`
* `portal-service/failover_userinfo_integration_test.go`
* `portal-service/handler.go`
* `portal-service/handler_test.go`
* `portal-service/main.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

portal-service/failover_reader.go
```diff
@@ -0,0 +1,60 @@
...
+	r.mu.Lock()
+	if c, ok := r.cache[siteID]; ok && r.now().Before(c.expires) {
+		r.mu.Unlock()
+		return c.target
+	}
+	r.mu.Unlock()
+
+	// Read outside the lock so a slow Mongo call does not serialize all readers.
+	st, err := r.store.Get(ctx, siteID)
+	if err != nil {
```
julianshen, line 49: _🏷️ nats-routing_ | _🔴 Blocking_ | _`distributed-system`, `mongodb`, `nats`, `tgd-review` (4 rules)_

**Store errors are reported as a valid home-site route.**

Every `FailoverStore.Get` error becomes `ServingHome`, which `servingURLs` treats as authoritative and returns with HTTP 200. If Mongo fails after a failed-over cache entry expires, or while a portal replica starts with an empty cache, discovery sends clients to the unavailable home NATS server. During `failing_back`, it can also move clients home before the routing fence permits it. Keep the last known target on refresh errors; when no authoritative cached value exists, propagate an error through `failoverTargeter` and return 5xx so clients retry instead of receiving a false-success NATS URL.

<details>

- **`tgd-review`**:51: After cache expiry or a cold start, any `FailoverStore.Get` error fabricates `ServingHome` without knowing the authoritative state. During `failed_over` or `failing_back`, existing clients and the backup can remain active while new `/api/userInfo` requests are directed home; during failback this can split writes across both sites and defeats the documented fence. Propagate an unavailable result on unknown state, or at minimum retain the last-known target on refresh errors and fail closed when no cached target exists. Add coverage for an expired backup entry followed by a store error.
- **`distributed-system`**:49: After a site is failed over, let its cached backup target expire and make Mongo temporarily unavailable. `ServingTarget` then ignores the previously known target and returns `ServingHome`, so `/api/userInfo` sends reconnecting users to the failed home site. If home has partially recovered while backup remains authoritative, both sides can accept writes. Return the last known target on read failure; if no prior value exists, propagate an unavailable error so discovery returns 503 rather than inventing a routing decision.
- **`mongodb`**:49: Once a cached backup target expires, any Mongo read error immediately returns home even though the authoritative document may still say `failed_over` or `failing_back`. This sends new discovery responses to the known-down or not-yet-restored home site. Because errors are uncached, a Mongo outage during a site-wide reconnect also causes every request to retry the failing query, amplifying pool pressure and log volume. Retain and serve the last known target on refresh errors with a bounded stale window or retry backoff; when no prior value exists, return an availability error rather than manufacturing a home decision.

</details>

<details>

```
In `portal-service/failover_reader.go` around line 49: Every `FailoverStore.Get` error becomes `ServingHome`, which `servingURLs` treats as authoritative and returns with HTTP 200. If Mongo fails after a failed-over cache entry expires, or while a portal replica starts with an empty cache, discovery sends clients to the unavailable home NATS server. During `failing_back`, it can also move clients home before the routing fence permits it. Keep the last known target on refresh errors; when no authoritative cached value exists, propagate an error through `failoverTargeter` and return 5xx so clients retry instead of receiving a false-success NATS URL.

```

</details>

portal-service/failover_reader.go
```diff
@@ -0,0 +1,60 @@
...
+func (r *failoverReader) ServingTarget(ctx context.Context, siteID string) ServingTarget {
+	r.mu.Lock()
+	if c, ok := r.cache[siteID]; ok && r.now().Before(c.expires) {
+		r.mu.Unlock()
+		return c.target
+	}
+	r.mu.Unlock()
+
+	// Read outside the lock so a slow Mongo call does not serialize all readers.
+	st, err := r.store.Get(ctx, siteID)
```
julianshen, line 48: _🎯 concurrency_ | _🔴 Blocking_ | _`distributed-system`, `mongodb`, `nats`, `performance-and-concurrency`, `tgd-review` (5 rules)_

**Concurrent cache misses can stampede Mongo and overwrite newer state**

On cold start or TTL expiry, every request can pass the unlocked cache miss path before any result is stored, so a site-wide reconnect burst issues one `store.Get` per login and can saturate Mongo's connection pool across every portal replica. The overlapping reads also create a lost-update race: request A can read version N, a failover transition occurs, request B reads and caches N+1, then A completes later and overwrites the cache with N for the full TTL, routing clients back to the failed home site. Store errors worsen the load because they are not cached and each request also emits a warning. Coalesce refreshes per site with singleflight or a per-key lock, retain the state version and reject older cache writes, and apply a short error retry-backoff/negative-cache interval with rate-limited logging. Add a concurrent test covering one store read per miss and out-of-order versions.

<details>

- **`distributed-system`**:57: Two refreshes can execute concurrently because reads occur outside the lock. One read can observe version 7/healthy before a transition but return late; another can observe version 8/failed_over and cache backup first. The late version-7 result then unconditionally overwrites it with home and a fresh TTL, regressing routing after this process already observed the newer state. Store the state version in `cachedTarget` and only replace an entry when the returned version is at least the cached version, or serialize refreshes per site.
- **`mongodb`**:48: Expired entries are read outside the mutex, and every caller later overwrites the cache without checking whether another refresh completed first. A Get that observed `healthy` before a Mongo transition can be delayed while a later Get observes `failed_over` and caches backup; when the first goroutine resumes, it replaces that newer value with home for another full TTL. The same pattern creates a Mongo query burst when many displaced clients reconnect after expiration. Coalesce refreshes per site (for example, with singleflight) and re-check the cache/generation before publishing a result so an older read cannot replace a completed refresh.
- **`nats`**:41: The cache lock is released before `Get`, and every completed read unconditionally replaces the cached target. Around a failover transition, request A can observe `home`, request B can later observe and cache `backup`, then A can complete last and overwrite it with `home` for another TTL. A reconnect burst on an empty or expired cache also issues one Mongo read per client, defeating the cache exactly during site-wide NATS migration. Coalesce refreshes per site with singleflight, or cache the state version and reject older completions; add a test that completes two refreshes out of order.
- **`tgd-review`**:58: The lock is released before `Get`, and the cache write has no version or generation check. If two refreshes overlap, one can read version 1 before a transition but return after another request has cached version 2, then overwrite the newer target for another full TTL. Concurrent misses also create a Mongo request burst. Coalesce refreshes per site or retain the state version and reject older cache writes; cover the out-of-order completion case with a concurrent test.

</details>

<details>

```
In `portal-service/failover_reader.go` around line 48: On cold start or TTL expiry, every request can pass the unlocked cache miss path before any result is stored, so a site-wide reconnect burst issues one `store.Get` per login and can saturate Mongo's connection pool across every portal replica. The overlapping reads also create a lost-update race: request A can read version N, a failover transition occurs, request B reads and caches N+1, then A completes later and overwrites the cache with N for the full TTL, routing clients back to the failed home site. Store errors worsen the load because they are not cached and each request also emits a warning. Coalesce refreshes per site with singleflight or a per-key lock, retain the state version and reject older cache writes, and apply a short error retry-backoff/negative-cache interval with rate-limited logging. Add a concurrent test covering one store read per miss and out-of-order versions.

```

</details>

docs/superpowers/plans/2026-08-12-sp6-ops-runbook.md
```diff
@@ -0,0 +1,231 @@
...
+manual operator action.
+
+**False alarm (undo a failover):** `{"action":"resume",...}` → straight back to
+`healthy`, skipping the drain.
+
+**Inspect:** `GET /internal/v1/failover` for the whole fleet's state, including
+`operator`/`reason`/`since` for the audit trail.
+
+### 1.5 Split-brain guarantee
+
```
julianshen, line 103: _🏷️ split-brain_ | _🔴 Blocking_ | _`distributed-system`_

**Portal routing does not fence the old site's writers.**

The claimed home-xor-backup guarantee does not follow from a single portal document. During a partial partition, some clients can lose home connectivity and rediscover the backup while other clients remain connected directly to home NATS and never consult portal again. TTL-stale or mixed-version portal replicas create the same interleaving. Since the data plane does not validate the serving state, both sites can accept writes despite the document being single-valued. Add a data-plane lease/serving epoch checked by writers and consumers, or require and verify that home writers are hard-fenced before changing routing; also document the mixed-version rollout gate.

<details>

```
In `docs/superpowers/plans/2026-08-12-sp6-ops-runbook.md` around line 103: The claimed home-xor-backup guarantee does not follow from a single portal document. During a partial partition, some clients can lose home connectivity and rediscover the backup while other clients remain connected directly to home NATS and never consult portal again. TTL-stale or mixed-version portal replicas create the same interleaving. Since the data plane does not validate the serving state, both sites can accept writes despite the document being single-valued. Add a data-plane lease/serving epoch checked by writers and consumers, or require and verify that home writers are hard-fenced before changing routing; also document the mixed-version rollout gate.

```

</details>

docs/superpowers/plans/2026-08-12-sp6-ops-runbook.md
```diff
@@ -0,0 +1,231 @@
...
+  -X POST https://portal-internal/internal/v1/failover/site-a \
+  -d '{"action":"failover","operator":"jane","reason":"site-a NATS unreachable"}'
+```
+Effect: `site-a` accounts' `/api/userInfo` now return the backup's coordinates;
+displaced clients reconnect there on their next connection failure.
+
+**Begin failback (start draining home):** `{"action":"failback",...}` → moves to
+`failing_back`. **`servingTarget` stays `backup`** — users keep writing to a
+stable backup while the home site drains. *(The drain/replay itself is SP5.)*
+
```
julianshen, line 90: _🏷️ atomicity_ | _🔴 Blocking_ | _`distributed-system`_

**Failback flips routing before backup writes are quiesced.**

A zero-lag observation is only a snapshot. After `complete` commits `healthy`, new discoveries begin using home while existing backup clients remain active until the later drain; portal replicas may also return cached backup targets for up to `FAILOVER_STATE_TTL`. A write accepted at backup after the final lag check can therefore race home writes or miss the replay entirely. Define a staged handoff that fences/quiesces backup writes, drains in-flight operations, replicates through a final checkpoint, advances a serving epoch, and only then routes and drains clients. The cache grace period must be included in that protocol.

<details>

```
In `docs/superpowers/plans/2026-08-12-sp6-ops-runbook.md` around line 90: A zero-lag observation is only a snapshot. After `complete` commits `healthy`, new discoveries begin using home while existing backup clients remain active until the later drain; portal replicas may also return cached backup targets for up to `FAILOVER_STATE_TTL`. A write accepted at backup after the final lag check can therefore race home writes or miss the replay entirely. Define a staged handoff that fences/quiesces backup writes, drains in-flight operations, replicates through a final checkpoint, advances a serving epoch, and only then routes and drains clients. The cache grace period must be included in that protocol.

```

</details>

docs/superpowers/plans/2026-08-12-sp6-ops-runbook.md
```diff
@@ -0,0 +1,231 @@
...
+```
+
+State machine (operator-driven; every write is CAS-guarded, so concurrent portal
+replicas cannot diverge):
+
+```
+healthy ──failover──► failed_over ──failback──► failing_back ──complete──► healthy
+   ▲                       │
+   └──────── resume ───────┘   (false alarm; skip the drain)
+```
```
julianshen, line 74: _🏷️ operational-safety_ | _🔴 Blocking_ | _`nats`_

**The runbook enables failover before the backup can serve NATS traffic.**

This failover command is presented under a `[READY]` procedure, but §2 explicitly says the backup serving handlers are gated on SP1+SP2. With this PR, invoking the command changes discovery to return the backup `natsUrl`; clients can therefore connect to a backup lacking materialized subscriptions and request/reply responders, producing timeouts or unavailable data during the outage. Mark failover activation as gated on SP1+SP2 and add an operational or technical readiness interlock; only deployment and inspection of the control surface should be labeled ready today.

<details>

```
In `docs/superpowers/plans/2026-08-12-sp6-ops-runbook.md` around line 74: This failover command is presented under a `[READY]` procedure, but §2 explicitly says the backup serving handlers are gated on SP1+SP2. With this PR, invoking the command changes discovery to return the backup `natsUrl`; clients can therefore connect to a backup lacking materialized subscriptions and request/reply responders, producing timeouts or unavailable data during the outage. Mark failover activation as gated on SP1+SP2 and add an operational or technical readiness interlock; only deployment and inspection of the control surface should be labeled ready today.

```

</details>

docs/superpowers/plans/2026-08-12-sp6-ops-runbook.md
```diff
@@ -0,0 +1,231 @@
...
+  -d '{"action":"failover","operator":"jane","reason":"site-a NATS unreachable"}'
+```
+Effect: `site-a` accounts' `/api/userInfo` now return the backup's coordinates;
+displaced clients reconnect there on their next connection failure.
+
+**Begin failback (start draining home):** `{"action":"failback",...}` → moves to
+`failing_back`. **`servingTarget` stays `backup`** — users keep writing to a
+stable backup while the home site drains. *(The drain/replay itself is SP5.)*
+
+**Complete failback (flip home):** `{"action":"complete",...}` → `healthy`;
```
julianshen, line 91: _🏷️ durability_ | _🔴 Blocking_ | _`silent-failure`_

**complete can acknowledge failback before replay is durable.**

This procedure says `complete` is available as a manual action today, but the same runbook states that replay/convergence is implemented by the still-gated SP5. The transition can therefore return success, mark the site healthy, and route clients home while outage writes exist only on the backup, exposing partial history or conflicting subsequent writes. The control surface should reject `complete` until durable replay and lag≈0 are verified; until that gate exists, the runbook must explicitly prohibit this transition rather than describe it as a current manual operation.

<details>

```
In `docs/superpowers/plans/2026-08-12-sp6-ops-runbook.md` around line 91: This procedure says `complete` is available as a manual action today, but the same runbook states that replay/convergence is implemented by the still-gated SP5. The transition can therefore return success, mark the site healthy, and route clients home while outage writes exist only on the backup, exposing partial history or conflicting subsequent writes. The control surface should reject `complete` until durable replay and lag≈0 are verified; until that gate exists, the runbook must explicitly prohibit this transition rather than describe it as a current manual operation.

```

</details>

portal-service/failover_reader.go
```diff
@@ -0,0 +1,60 @@
...
+// it returns the cached value; otherwise it reads the store, derives the target,
+// and caches it. A store error resolves to home (fail-safe) and is not cached.
+func (r *failoverReader) ServingTarget(ctx context.Context, siteID string) ServingTarget {
+	r.mu.Lock()
+	if c, ok := r.cache[siteID]; ok && r.now().Before(c.expires) {
+		r.mu.Unlock()
+		return c.target
+	}
+	r.mu.Unlock()
+
```
julianshen, line 46: _🏷️ backpressure_ | _🟠 Warning_ | _`distributed-system`_

**Cold-cache reconnects fan out one Mongo read per client.**

When a site outage disconnects many clients, concurrent discovery requests for the same site all observe the cache miss, release the lock, and call Mongo independently. An unavailable Mongo makes this repeat on every retry because errors are not cached. The resulting O(clients) read burst can overload the routing store precisely during failover. Coalesce in-flight refreshes per site and use the last known value or a short bounded error backoff to prevent retry amplification.

<details>

```
In `portal-service/failover_reader.go` around line 46: When a site outage disconnects many clients, concurrent discovery requests for the same site all observe the cache miss, release the lock, and call Mongo independently. An unavailable Mongo makes this repeat on every retry because errors are not cached. The resulting O(clients) read burst can overload the routing store precisely during failover. Coalesce in-flight refreshes per site and use the last known value or a short bounded error backoff to prevent retry amplification.

```

</details>

portal-service/main.go
```diff
@@ -65,6 +65,12 @@ type config struct {
 	// FailoverInternalAddr is the listen address for the internal-only control
 	// surface — kept off the public browser-facing server.
 	FailoverInternalAddr string `env:"FAILOVER_INTERNAL_ADDR" envDefault:":8090"`
+	// FailoverStateTTL bounds how long portal caches a site's serving target
+	// before re-reading it (routing freshness vs. Mongo load).
+	FailoverStateTTL time.Duration `env:"FAILOVER_STATE_TTL" envDefault:"5s"`
+	// BackupSiteID is the reserved PORTAL_SITE_URLS id served for a failed-over
+	// site. Empty in single-site/dev deployments (no failover occurs there).
+	BackupSiteID string `env:"PORTAL_BACKUP_SITE_ID" envDefault:""`
```
julianshen, line 73: _🏷️ configuration_ | _🟠 Warning_ | _`distributed-system`_

**Failover can commit without usable backup coordinates.**

`PORTAL_BACKUP_SITE_ID` defaults to empty and is not validated against `PORTAL_SITE_URLS`, while the operator handler can still commit a site to `failed_over`. The control request then succeeds and persists the transition, but every affected `/api/userInfo` request returns 500 when `servingURLs` cannot find the backup entry. Validate the backup ID and registry entry before enabling or accepting the `failover` transition, so an operator receives a failed transition rather than a successful state change that disables discovery.

<details>

```
In `portal-service/main.go` around line 73: `PORTAL_BACKUP_SITE_ID` defaults to empty and is not validated against `PORTAL_SITE_URLS`, while the operator handler can still commit a site to `failed_over`. The control request then succeeds and persists the transition, but every affected `/api/userInfo` request returns 500 when `servingURLs` cannot find the backup entry. Validate the backup ID and registry entry before enabling or accepting the `failover` transition, so an operator receives a failed transition rather than a successful state change that disables discovery.

```

</details>

docs/superpowers/plans/2026-08-12-sp6-ops-runbook.md
```diff
@@ -0,0 +1,231 @@
...
+   └──────── resume ───────┘   (false alarm; skip the drain)
+```
+
+`servingTarget` is `backup` while `failed_over`/`failing_back`, else `home`.
+
+**Declare a site down (fail over):**
+```bash
+curl -sS -H "Authorization: Bearer $FAILOVER_OPS_TOKEN" \
+  -X POST https://portal-internal/internal/v1/failover/site-a \
+  -d '{"action":"failover","operator":"jane","reason":"site-a NATS unreachable"}'
```
julianshen, line 82: _🏷️ reliability_ | _🟠 Warning_ | _`silent-failure`_

**The failover command exits successfully on HTTP failures.**

`curl -sS` returns exit status 0 for HTTP 4xx/5xx responses, so authentication, CAS, validation, or database failures can appear successful to an operator or script while users remain routed to the failed home site. Additionally, `-d` defaults to `application/x-www-form-urlencoded`, although this endpoint expects JSON. Make HTTP errors fail the command and send the correct content type.

<details>

> ‼️ **IMPORTANT**

```suggestion
curl --fail-with-body -sS \
  -H "Authorization: Bearer $FAILOVER_OPS_TOKEN" \
  -H "Content-Type: application/json" \
  -X POST https://portal-internal/internal/v1/failover/site-a \
  -d '{"action":"failover","operator":"jane","reason":"site-a NATS unreachable"}'
```

</details>

<details>

```
In `docs/superpowers/plans/2026-08-12-sp6-ops-runbook.md` around line 80: `curl -sS` returns exit status 0 for HTTP 4xx/5xx responses, so authentication, CAS, validation, or database failures can appear successful to an operator or script while users remain routed to the failed home site. Additionally, `-d` defaults to `application/x-www-form-urlencoded`, although this endpoint expects JSON. Make HTTP errors fail the command and send the correct content type.

```

</details>

julianshen (commented):
tGD inline review

portal-service/failover_userinfo_integration_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/hmchangw/chat/pkg/testutil"
+)
+
+func TestUserInfo_FailoverOverMongo(t *testing.T) {
+	db := testutil.MongoDB(t, "portal")
+	store := newMongoFailoverStore(db)
+	ctx := context.Background()
+
+	// Short TTL so a later flip-back would be observable within the test.
+	reader := newFailoverReader(store, 20*time.Millisecond)
```
julianshen, line 24: _🧪 tests_ | _🟠 Warning_ | _`tgd-review`_

**The integration test never exercises failback refresh.**

Although the reader uses a 20 ms TTL and the comment mentions a later flip-back, the test only transitions to `StatusFailedOver` and performs one discovery request. It does not verify that the composed handler and Mongo reader refresh after expiry and return home coordinates, which is a documented requirement in both directions. Extend the test to transition back through the valid state-machine path, wait or poll past the TTL, and assert the home URLs while preserving the home `siteId`.

<details>

```
In `portal-service/failover_userinfo_integration_test.go` around line 24: Although the reader uses a 20 ms TTL and the comment mentions a later flip-back, the test only transitions to `StatusFailedOver` and performs one discovery request. It does not verify that the composed handler and Mongo reader refresh after expiry and return home coordinates, which is a documented requirement in both directions. Extend the test to transition back through the valid state-machine path, wait or poll past the TTL, and assert the home URLs while preserving the home `siteId`.

```

</details>

hmchangw:
Superseded by #499, which applies the SP3 routing override (plus the SP4 work this was stacked on) directly on top of `main` and builds/tests/lints green there. This branch sits on a divergent history that can't be cleanly rebased onto `main`, so closing in favor of #499.

---

PR #259, opened 2026-08-13
refactor(obs): batch managed-baggage writes; document and gate msgraph body logging

flywindy:
Follow-up to #236, which merged before these review comments could be addressed. Four comments from @mliu33, all four handled here.

## `pkg/obs` — redundant clear + batching ([#discussion_r3768254900](https://github.com/hmchangw/newchat/pull/236#discussion_r3768254900), [#discussion_r3768272999](https://github.com/hmchangw/newchat/pull/236#discussion_r3768272999))

**The redundant clear was real for supplied keys only.** `ContextWithPublicIdentity` cleared all three managed keys, then `ContextWithIdentity` cleared each supplied one again via `contextWithBaggageAttribute`. But the pre-clear is *not* redundant for a key with no trusted value: `contextWithBaggageAttribute` returns at `if value == ""` before dropping anything, which is exactly the forgery case the function exists for. It is also not redundant when emission is off, where it is the only sanitization that runs.

So the clear is now scoped rather than removed: with emission on, only the keys no trusted value will overwrite are pre-cleared; with emission off, all of them still are.

**Batching.** `withoutBaggageAttributes` now takes keys variadically and does one `o11y.ContextWithoutBaggageValues` plus one span write for the whole set instead of one of each per key, skipping keys the baggage does not carry. `contextWithBaggageAttributes` materializes the accepted values in a single `SetAttributes`.

The baggage **set** path still runs one call per key — the SDK validates a single member per call (`ContextWithBaggageValue(ctx, key, value)`), and hand-rolling `baggage.New(members...)` to batch it would give up that validation plus the size-limit and reserved-key checks. Worth raising with the o11y SDK as a `ContextWithBaggageValues` request rather than working around here.

Net per call on the public path: up to 6 baggage rebuilds → at most 3, and up to 6 span writes → at most 3. This runs once per NATS request and once per message in `message-gatekeeper`.

## `pkg/msgraph` — request bodies are never logged ([#discussion_r3768154477](https://github.com/hmchangw/newchat/pull/236#discussion_r3768154477), [#discussion_r3768158742](https://github.com/hmchangw/newchat/pull/236#discussion_r3768158742))

To answer the question directly: no, `newExternalRequestWithContext` does not log the body. It strips baggage and hands `body` to the stdlib. Nothing else in the package logs one either — the only `slog` call is a throttling warning in `chats.go`, and both token paths deliberately surface just the HTTP status and the OAuth error code.

The finding is that a reviewer had to ask. Two changes:

- The doc comment now states what happens to `body`, and names the reason it matters: the client-credentials form carries `client_secret`, and the ROPC form in `presence.go` additionally carries `username` and `password`.
- `.semgrep/msgraph-secrets.yml` turns the convention into a CI gate. A form value, an `io.ReadAll` body, or a stringified body reaching `slog` or `fmt.Errorf` inside `pkg/msgraph` now fails the `sast` job.

## Validation

- `make fmt`, `make lint` — 0 issues
- `make test` (race detector, all packages)
- `make sast-gosec` — 0 findings
- Local semgrep rules with the CI excludes — 0 findings; the new rule verified **both ways**: nothing on the package as it stands, and all four violation shapes caught when introduced deliberately
- `make sast-vuln` and the registry half of `make sast-semgrep` (`p/golang`, `p/security-audit`) could not run here — `vuln.go.dev` and `semgrep.dev` are blocked by this environment's network policy. Covered by the CI `sast` job.

## Review notes

- No client-facing request/response or event schema changed, so `docs/client-api.md` and its derived views do not require updates.
- No store interface changed, so mock regeneration is not required.
- `TestUnsuppliedManagedKeys` pins the new helper against `ManagedBaggageKeys()`, so a fourth managed key cannot be registered for materialization and then silently left caller-controlled at public ingress.

https://claude.ai/code/session_019caNwQvoRXKxGeFPvkw8m7

---
_Generated by [Claude Code](https://claude.ai/code/session_019caNwQvoRXKxGeFPvkw8m7)_

## Summary by CodeRabbit

* **Security**
  * Added automated safeguards to detect potential logging of Microsoft Graph credentials, tokens, and response bodies.
  * Clarified that sensitive request bodies are transmitted without being read, logged, or attached to telemetry spans.

* **Bug Fixes**
  * Improved identity telemetry handling by clearing stale values and preventing untrusted identity data from persisting when telemetry is disabled or values are incomplete.
  * Ensured fully supplied trusted identity values consistently replace outdated data.

### DIFF
--- a/.semgrep/msgraph-secrets.yml
+++ b/.semgrep/msgraph-secrets.yml
@@ -18,13 +18,21 @@ rules:
         - "**/*_test.go"
     patterns:
       - pattern-either:
-          # A form/body/token variable reaching any slog or fmt sink.
+          # A form/body/token variable reaching any slog or fmt sink. Both sinks
+          # matter: an error carrying the form reaches the server log through
+          # errcode.Classify just as surely as a slog call does.
           - pattern: slog.$LEVEL(..., $FORM.Encode(), ...)
           - pattern: slog.$LEVEL(..., <... $FORM.Encode() ...>, ...)
+          - pattern: fmt.Errorf(..., $FORM.Encode(), ...)
+          - pattern: fmt.Errorf(..., <... $FORM.Encode() ...>, ...)
           - pattern: |
               $FORM := url.Values{}
               ...
               slog.$LEVEL(..., $FORM, ...)
+          - pattern: |
+              $FORM := url.Values{}
+              ...
+              fmt.Errorf(..., $FORM, ...)
           - pattern: |
               $BODY, $ERR := io.ReadAll(...)
               ...
--- a/.semgrep/msgraph-secrets.yml
+++ b/.semgrep/msgraph-secrets.yml
@@ -1,0 +1,39 @@
+rules:
+  - id: msgraph-no-credential-body-logging
+    languages: [go]
+    severity: ERROR
+    message: >
+      Do not log a Microsoft identity request/response body or its form values.
+      The client-credentials form in pkg/msgraph/msgraph.go carries
+      client_secret, and the ROPC form in pkg/msgraph/presence.go additionally
+      carries username and password; the token response carries access_token.
+      A single debug line here turns a credential into a log record that ships
+      to Loki. Surface the HTTP status and the OAuth error code (tr.Error) only
+      — see accessToken/ropcToken for the intended shape, and CLAUDE.md
+      "Never log tokens, passwords, or full message bodies".
+    paths:
+      include:
+        - "**/pkg/msgraph/"
+      exclude:
+        - "**/*_test.go"
+    patterns:
+      - pattern-either:
+          # A form/body/token variable reaching any slog or fmt sink.
+          - pattern: slog.$LEVEL(..., $FORM.Encode(), ...)
+          - pattern: slog.$LEVEL(..., <... $FORM.Encode() ...>, ...)
+          - pattern: |
+              $FORM := url.Values{}
+              ...
+              slog.$LEVEL(..., $FORM, ...)
+          - pattern: |
+              $BODY, $ERR := io.ReadAll(...)
+              ...
+              slog.$LEVEL(..., $BODY, ...)
+          - pattern: |
+              $BODY, $ERR := io.ReadAll(...)
+              ...
+              fmt.Errorf($MSG, ..., $BODY, ...)
+          - pattern: |
+              $BODY, $ERR := io.ReadAll(...)
+              ...
+              string($BODY)
--- a/.semgrep/msgraph-secrets.yml
+++ b/.semgrep/msgraph-secrets.yml
@@ -45,3 +45,27 @@ rules:
               $BODY, $ERR := io.ReadAll(...)
               ...
               string($BODY)
+          # The decoded token response: the raw-body patterns above stop at
+          # json.Unmarshal, but tr.AccessToken is the natural thing to log.
+          - pattern: slog.$LEVEL(..., $TR.AccessToken, ...)
+          - pattern: slog.$LEVEL(..., <... $TR.AccessToken ...>, ...)
+          - pattern: fmt.Errorf(..., $TR.AccessToken, ...)
+          - pattern: fmt.Errorf(..., <... $TR.AccessToken ...>, ...)
+          - pattern: |
+              var $TR tokenResponse
+              ...
+              slog.$LEVEL(..., $TR, ...)
+          - pattern: |
+              var $TR tokenResponse
+              ...
+              fmt.Errorf(..., $TR, ...)
+          # A single extracted form value. Typed to url.Values so an unrelated
+          # getter — resp.Header.Get("Retry-After"), which chats.go logs — does
+          # not trip the rule.
+          - patterns:
+              - pattern-either:
+                  - pattern: slog.$LEVEL(..., $FORM.Get(...), ...)
+                  - pattern: fmt.Errorf(..., $FORM.Get(...), ...)
+              - metavariable-type:
+                  metavariable: $FORM
+                  type: url.Values
--- a/pkg/obs/obs.go
+++ b/pkg/obs/obs.go
@@ -279,34 +279,18 @@ func ContextWithIdentity(ctx context.Context, account, roomID, siteID string) co
 // entry span's attributes, which the SDK's OnStart processor materialized from
 // the forged values before any handler ran.
 func ContextWithPublicIdentity(ctx context.Context, account, roomID, siteID string) context.Context {
-	if !contextAttributesEnabled.Load() {
-		// Emission is off — NATS tracing can still be forced on by
-		// OTEL_NATS_TRACING_ENABLED or the flag relay, so the caller's values can
-		// reach an observability-enabled downstream service. Nothing here will
-		// overwrite them, so every managed key has to go.
-		return withoutBaggageAttributes(ctx, managedBaggageKeys...)
-	}
-	// ContextWithIdentity drops each key it has a trusted value for, so only the
-	// unsupplied ones need clearing here.
-	ctx = withoutBaggageAttributes(ctx, unsuppliedManagedKeys(account, roomID, siteID)...)
+	// Sanitize unconditionally, before reading any toggle. Emission being off
+	// does not make the caller's values safe: NATS tracing can be forced on by
+	// OTEL_NATS_TRACING_ENABLED or the flag relay, carrying them to a service
+	// that does emit. Clearing here rather than leaving it to the trusted-value
+	// path also keeps this boundary's guarantee independent of what
+	// ContextWithIdentity decides — one decision, taken at ingress.
+	//
+	// This costs no more than clearing a subset would: the pass is batched, and
+	// the trusted-value path's own clear then finds nothing left to remove and
+	// skips its rebuild.
+	ctx = withoutBaggageAttributes(ctx, managedBaggageKeys...)
 	return ContextWithIdentity(ctx, account, roomID, siteID)
-}
-
-// unsuppliedManagedKeys returns the managed keys this boundary has no trusted
-// value for. TestUnsuppliedManagedKeys pins it against ManagedBaggageKeys so a
-// key added to one is not forgotten here.
-func unsuppliedManagedKeys(account, roomID, siteID string) []string {
-	keys := make([]string, 0, len(managedBaggageKeys))
-	if account == "" {
-		keys = append(keys, o11y.UserNameKey)
-	}
-	if roomID == "" {
-		keys = append(keys, RoomIDKey)
-	}
-	if siteID == "" {
-		keys = append(keys, SiteIDKey)
-	}
-	return keys
 }
 
 // contextWithBaggageAttributes sets each non-empty attribute as baggage and
--- a/pkg/obs/obs.go
+++ b/pkg/obs/obs.go
@@ -312,17 +312,30 @@ func unsuppliedManagedKeys(account, roomID, siteID string) []string {
 // contextWithBaggageAttributes sets each non-empty attribute as baggage and
 // materializes the accepted ones onto the current span in one write. Empty
 // values are skipped, which preserves trusted baggage from an internal hop.
-// The o11y SDK validates one member per call, so the baggage rebuilds cannot be
-// batched; the span write can, and is.
+//
+// Superseded values are dropped for the whole set up front — one rebuild rather
+// than one per key — which also strengthens the fail-safe: if any replacement
+// is rejected below, no stale value survives for any key. The o11y SDK
+// validates one member per call, so only the set loop remains per-key.
 func contextWithBaggageAttributes(ctx context.Context, kvs ...attribute.KeyValue) context.Context {
-	accepted := make([]attribute.KeyValue, 0, len(kvs))
+	supplied := make([]attribute.KeyValue, 0, len(kvs))
+	superseded := make([]string, 0, len(kvs))
 	for _, kv := range kvs {
-		key, value := string(kv.Key), kv.Value.AsString()
-		if value == "" {
+		if kv.Value.AsString() == "" {
 			continue
 		}
-		ctx = withoutBaggageAttributes(ctx, key)
-		next, err := o11y.ContextWithBaggageValue(ctx, key, value)
+		supplied = append(supplied, kv)
+		superseded = append(superseded, string(kv.Key))
+	}
+	if len(supplied) == 0 {
+		return ctx
+	}
+	ctx = withoutBaggageAttributes(ctx, superseded...)
+
+	accepted := make([]attribute.KeyValue, 0, len(supplied))
+	for _, kv := range supplied {
+		key := string(kv.Key)
+		next, err := o11y.ContextWithBaggageValue(ctx, key, kv.Value.AsString())
 		if err != nil {
 			slog.WarnContext(ctx, "telemetry identity attribute omitted", "attribute", key, "error", err)
 			continue
--- a/pkg/obs/obs.go
+++ b/pkg/obs/obs.go
@@ -249,7 +249,7 @@ func ContextWithIdentity(ctx context.Context, account, roomID, siteID string) co
 	}
 
 	if account != "" {
-		ctx = withoutBaggageAttribute(ctx, o11y.UserNameKey)
+		ctx = withoutBaggageAttributes(ctx, o11y.UserNameKey)
 		if userBaggageEnabled.Load() {
 			next, err := o11y.ContextWithUser(ctx, account)
 			if err != nil {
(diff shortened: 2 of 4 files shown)


.semgrep/msgraph-secrets.yml
```diff
@@ -0,0 +1,39 @@
...
+              ...
+              slog.$LEVEL(..., $FORM, ...)
+          - pattern: |
+              $BODY, $ERR := io.ReadAll(...)
+              ...
+              slog.$LEVEL(..., $BODY, ...)
+          - pattern: |
+              $BODY, $ERR := io.ReadAll(...)
+              ...
+              fmt.Errorf($MSG, ..., $BODY, ...)
```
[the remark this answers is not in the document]
flywindy: Correct, and it was the leak path that mattered most. Fixed in 3ee715c.

I reproduced the gap before changing anything: `fmt.Errorf("token request failed: %s", form.Encode())` in `pkg/msgraph` returned **0 findings** under the rule as written. The `$FORM` patterns were all anchored to `slog`, and the `fmt.Errorf` patterns were all anchored to an `io.ReadAll` body — so a form reaching an error slipped between them. That's worse than the slog case it did cover, because an error carrying the form reaches the server log through `errcode.Classify` without anyone writing a log line at all.

Added `fmt.Errorf(..., $FORM.Encode(), ...)`, its `<... ...>` deep-expression variant, and the `url.Values{} … fmt.Errorf(..., $FORM, ...)` taint shape. Re-verified in all three directions: 2/2 on the probe that previously scored 0, the earlier four `slog`/body shapes still caught, and still 0 findings on `pkg/msgraph` as it stands.

---

pkg/obs/obs_test.go
```diff
@@ -397,6 +397,18 @@ func mustBaggageMember(t *testing.T, key, value string) baggage.Member {
...
+// introduces: it clears only the keys no trusted value will overwrite, so a
+// managed key this helper forgets would stay caller-controlled.
+func TestUnsuppliedManagedKeys(t *testing.T) {
+	assert.ElementsMatch(t, ManagedBaggageKeys(), unsuppliedManagedKeys("", "", ""),
+		"with no trusted values every managed key must be cleared")
+	assert.Empty(t, unsuppliedManagedKeys("alice", "room-42", "site-a"),
+		"a fully supplied identity leaves nothing to pre-clear")
+	assert.Equal(t, []string{SiteIDKey}, unsuppliedManagedKeys("alice", "room-42", ""))
+	assert.Equal(t, []string{o11y.UserNameKey, RoomIDKey}, unsuppliedManagedKeys("", "", "site-a"))
+}
```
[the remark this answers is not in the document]
flywindy: Half of this already existed, half didn't — added the missing half in 3ee715c.

`TestContextWithPublicIdentity_MasterOffStillDropsManagedKeys` (from #236) already seeds every managed key with emission disabled and asserts they're cleared from **baggage**, plus that an unrelated key survives. What it never asserted was the **span** side, and that's the half my batching change here could have broken.

`TestContextWithPublicIdentity_MasterOffZeroesSpanAttributes` now seeds forged entry-span attributes as well and proves they're zeroed. It matters precisely because emission being off doesn't mean nothing was materialized — NATS tracing can be forced on independently via `OTEL_NATS_TRACING_ENABLED` or the flag relay, and OpenTelemetry can't delete an attribute once set, only overwrite it.

It passed on first run, so I checked it isn't vacuous: with the master-off clear stubbed out to `return ctx`, it fails with `Should be empty, but was forged` on all three keys.

---

julianshen:
**Actionable comments posted: 1** — 🟠 1 warning

<details>

* `.semgrep/msgraph-secrets.yml`
* `pkg/msgraph/msgraph.go`
* `pkg/obs/obs.go`
* `pkg/obs/obs_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `silent-failure`
* `tgd-review`

</details>

pkg/obs/obs.go
```diff
@@ -278,39 +279,86 @@ func ContextWithIdentity(ctx context.Context, account, roomID, siteID string) co
...
-		return ctx
+		// Emission is off — NATS tracing can still be forced on by
+		// OTEL_NATS_TRACING_ENABLED or the flag relay, so the caller's values can
+		// reach an observability-enabled downstream service. Nothing here will
+		// overwrite them, so every managed key has to go.
+		return withoutBaggageAttributes(ctx, managedBaggageKeys...)
 	}
+	// ContextWithIdentity drops each key it has a trusted value for, so only the
+	// unsupplied ones need clearing here.
+	ctx = withoutBaggageAttributes(ctx, unsuppliedManagedKeys(account, roomID, siteID)...)
```
julianshen: _🎯 distributed-correctness_ | _🔴 Blocking_ | _`distributed-system`_

**A concurrent flag change can propagate forged identity baggage.**

A public request can arrive with forged `user.name` baggage while supplying trusted `account="alice"`. If this function observes `contextAttributesEnabled` as true, `unsuppliedManagedKeys` leaves `user.name` intact. The flag relay can then disable emission before `ContextWithIdentity` performs its own load, causing it to return without replacing that key. Independently enabled NATS tracing can consequently propagate the forged identity to another service. Clear every managed key at ingress before delegating, as the previous implementation did, so correctness does not depend on two atomic flag reads agreeing.

<details>

> ‼️ **IMPORTANT**

```suggestion
	// Clear every caller-controlled key before delegating so a concurrent flag
	// change cannot make ContextWithIdentity return before replacing it.
	ctx = withoutBaggageAttributes(ctx, managedBaggageKeys...)
```

</details>

<details>

```
In `pkg/obs/obs.go` around line 289: A public request can arrive with forged `user.name` baggage while supplying trusted `account="alice"`. If this function observes `contextAttributesEnabled` as true, `unsuppliedManagedKeys` leaves `user.name` intact. The flag relay can then disable emission before `ContextWithIdentity` performs its own load, causing it to return without replacing that key. Independently enabled NATS tracing can consequently propagate the forged identity to another service. Clear every managed key at ingress before delegating, as the previous implementation did, so correctness does not depend on two atomic flag reads agreeing.

```

</details>

julianshen: _🔒 security_ | _🔴 Blocking_ | _`tgd-review`_

**A flag transition can preserve forged identity baggage.**

This function checks contextAttributesEnabled before clearing only unsupplied keys, but ContextWithIdentity checks the same atomic flag again. If the flag relay changes it from true to false between those loads, ContextWithIdentity returns without removing or replacing the supplied account, room, and site keys. Caller-controlled values can then propagate through forced NATS tracing. Clear every managed key before the second check, or refactor both functions to use one captured decision.

<details>

> ‼️ **IMPORTANT**

```suggestion
	ctx = withoutBaggageAttributes(ctx, managedBaggageKeys...)
```

</details>

<details>

```
In `pkg/obs/obs.go` around line 291: This function checks contextAttributesEnabled before clearing only unsupplied keys, but ContextWithIdentity checks the same atomic flag again. If the flag relay changes it from true to false between those loads, ContextWithIdentity returns without removing or replacing the supplied account, room, and site keys. Caller-controlled values can then propagate through forced NATS tracing. Clear every managed key before the second check, or refactor both functions to use one captured decision.

```

</details>

.semgrep/msgraph-secrets.yml
```diff
@@ -0,0 +1,39 @@
...
+      to Loki. Surface the HTTP status and the OAuth error code (tr.Error) only
+      — see accessToken/ropcToken for the intended shape, and CLAUDE.md
+      "Never log tokens, passwords, or full message bodies".
+    paths:
+      include:
+        - "**/pkg/msgraph/"
+      exclude:
+        - "**/*_test.go"
+    patterns:
+      - pattern-either:
```
julianshen, line 20: _🔒 security_ | _🟠 Warning_ | _`tgd-review`_

**The rule misses decoded tokens and extracted form secrets.**

Every alternative binds either the raw $BODY or $FORM, so obvious credential logging such as slog.Error(..., "token", tr.AccessToken) after JSON decoding or slog.Error(..., "secret", form.Get("client_secret")) matches none of them. This does not enforce the stated prohibition on logging token responses or form values. Model these values with taint sources and logging/error sinks, or add explicit patterns, and include positive rule fixtures for decoded access tokens and form getters.

<details>

```
In `.semgrep/msgraph-secrets.yml` around line 20: Every alternative binds either the raw $BODY or $FORM, so obvious credential logging such as slog.Error(..., "token", tr.AccessToken) after JSON decoding or slog.Error(..., "secret", form.Get("client_secret")) matches none of them. This does not enforce the stated prohibition on logging token responses or form values. Model these values with taint sources and logging/error sinks, or add explicit patterns, and include positive rule fixtures for decoded access tokens and form getters.

```

</details>

pkg/obs/obs.go
```diff
@@ -278,39 +279,99 @@ func ContextWithIdentity(ctx context.Context, account, roomID, siteID string) co
...
 	return ContextWithIdentity(ctx, account, roomID, siteID)
```
julianshen, line 293: _🔒 security_ | _🔴 Blocking_ | _`tgd-review`_

**A feature-flag race can preserve forged supplied identities.**

This now clears only unsupplied keys before calling ContextWithIdentity, which independently checks contextAttributesEnabled. Because this atomic flag can change at runtime, it can be true here and false when ContextWithIdentity checks it. With all identity arguments supplied, nothing is pre-cleared and ContextWithIdentity then returns without replacing the caller-controlled baggage, allowing forged identity values to propagate. Clear every managed key before the second check, as the previous implementation did, or pass one captured flag state into logic that does not reload it.

<details>

```
In `pkg/obs/obs.go` around line 292: This now clears only unsupplied keys before calling ContextWithIdentity, which independently checks contextAttributesEnabled. Because this atomic flag can change at runtime, it can be true here and false when ContextWithIdentity checks it. With all identity arguments supplied, nothing is pre-cleared and ContextWithIdentity then returns without replacing the caller-controlled baggage, allowing forged identity values to propagate. Clear every managed key before the second check, as the previous implementation did, or pass one captured flag state into logic that does not reload it.

```

</details>

.semgrep/msgraph-secrets.yml
```diff
@@ -0,0 +1,47 @@
...
+          # A form/body/token variable reaching any slog or fmt sink. Both sinks
```
julianshen, line 21: _🔒 security_ | _🟠 Warning_ | _`tgd-review`_

**The rule does not detect logging a parsed token response.**

The message says the rule protects access_token, but every response-body pattern requires the original io.ReadAll variable to reach a sink or be converted to string. After json.Unmarshal, code such as slog.Debug("token response", "response", tr) or logging tr.AccessToken matches none of these patterns. Add explicit patterns for token-response values and fields, or use suitable taint/dataflow rules, with positive and negative Semgrep test fixtures.

<details>

```
In `.semgrep/msgraph-secrets.yml` around line 21: The message says the rule protects access_token, but every response-body pattern requires the original io.ReadAll variable to reach a sink or be converted to string. After json.Unmarshal, code such as slog.Debug("token response", "response", tr) or logging tr.AccessToken matches none of these patterns. Add explicit patterns for token-response values and fields, or use suitable taint/dataflow rules, with positive and negative Semgrep test fixtures.

```

</details>

pkg/obs/obs.go
flywindy: Taken, though the stated mechanism doesn't hold — fixed in 18b07e4.

**The race isn't reachable as described.** `contextAttributesEnabled` is written in exactly two places, both inside `initSDK` (obs.go:204 and :220), i.e. at `obs.Init` during startup. The otel-nats flag relay controls whether the NATS wrapper traces; it never touches this variable. So in a running service the two reads always observe the same value, and the only way to flip it mid-flight is a second `obs.Init`, which happens in tests and not in production.

**But the fix is now free, so I took it anyway.** After the batching in this PR, clearing all three keys up front costs a single `ContextWithoutBaggageValues` — and the trusted-value path's own clear then finds the keys already absent, short-circuits on the `bag.Member(key).Key() == ""` check, and does no rebuild and no span write. So "clear everything at ingress" and "clear the unsupplied subset" now cost the same, and the former doesn't require anyone to reason about when a flag can change. One decision, taken at ingress.

`unsuppliedManagedKeys` and its drift-guard test go away with it: the clear derives from `managedBaggageKeys` directly, which is a stronger guarantee than a test asserting two lists agree.

Also added `TestContextWithPublicIdentity_FullySuppliedIdentityReplacesForgedValues` — the fully-supplied-plus-forged case you describe had no coverage either way.

---

.semgrep/msgraph-secrets.yml
flywindy: Correct — the rule claimed to protect `access_token` and didn't. Fixed in 18b07e4.

Confirmed all four shapes escaped before changing anything (0 findings): `slog.Error(..., "token", tr.AccessToken)`, `slog.Debug("token response", "response", tr)` on the decoded struct, `fmt.Errorf("token=%s", tr.AccessToken)`, and `form.Get("client_secret")`. Every alternative was anchored to the raw `$BODY` or `$FORM`, so anything that survived `json.Unmarshal` or a single-field extraction walked straight through.

Added: `$TR.AccessToken` at both sinks plus deep-expression variants, `var $TR tokenResponse … slog/fmt.Errorf(…, $TR, …)` for the whole decoded struct, and the form getter.

The getter needed care — a bare `$X.Get(...)` pattern would have flagged `resp.Header.Get("Retry-After")`, which `chats.go:153` legitimately logs. It's constrained with `metavariable-type: url.Values` so only the credential form matches. Verified 4/4 on the new probes, the earlier probes still caught, and 0 findings on `pkg/msgraph` including that throttle log.

On the fixtures: I tried `semgrep --test` with an annotated `.semgrep/msgraph-secrets.go`, and the pairing doesn't work for a path-scoped rule — `paths.include: **/pkg/msgraph/` excludes the fixture from the config's targets, so `--test` reports "No unit tests found" (and one invocation form crashes outright with an `IndexError` in `semgrep/test.py`). The `sast` target also excludes `testdata`, so a fixture there would never be scanned. Rather than ship a fixture that nothing runs, verification stays the probe runs recorded above and in the commit message. Happy to revisit if you'd rather have a `--test` harness wired up as its own change.

---

pkg/obs/obs.go
flywindy: Same finding as [#discussion_r3771967611](https://github.com/hmchangw/newchat/pull/259#discussion_r3771967611) — fixed in 18b07e4, full reasoning there. Short version: the flag is only written by `obs.Init` at startup (not by the relay), so the two-read window isn't reachable in a running service, but every managed key is now cleared at ingress before any toggle read anyway, because after this PR's batching that costs nothing.

---

pkg/obs/obs.go
flywindy: Same finding as [#discussion_r3771967611](https://github.com/hmchangw/newchat/pull/259#discussion_r3771967611) — fixed in 18b07e4, reasoning there.

---

.semgrep/msgraph-secrets.yml
flywindy: Same finding as [#discussion_r3771967624](https://github.com/hmchangw/newchat/pull/259#discussion_r3771967624) — fixed in 18b07e4. The decoded `tokenResponse` and its `AccessToken` field now match at both the `slog` and `fmt.Errorf` sinks, verified against probes covering exactly the `slog.Debug("token response", "response", tr)` and `tr.AccessToken` shapes you name.

---

pkg/obs/obs_test.go
```diff
@@ -390,6 +390,81 @@ func TestContextWithPublicIdentity_MasterOffStillDropsManagedKeys(t *testing.T)
...
+	clearEnv(t)
+	t.Setenv("OTEL_SERVICE_NAME", "public-identity-master-off-span-svc")
+
+	_, shutdown, err := Init(context.Background())
+	require.NoError(t, err)
+	t.Cleanup(func() { assert.NoError(t, shutdown(context.Background())) })
+
+	recorder := tracetest.NewSpanRecorder()
+	tp := sdktrace.NewTracerProvider(sdktrace.WithSpanProcessor(recorder))
+	t.Cleanup(func() { _ = tp.Shutdown(context.Background()) })
```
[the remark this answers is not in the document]
flywindy: Fixed in b180271. Fair catch — line 404 of the same test asserts the SDK shutdown error and line 408 discarded the tracer provider's, which is inconsistent within one test.

The provider's only processor is a `tracetest.SpanRecorder`, so the error is always nil today. That's the argument for asserting it rather than against: it costs nothing now, and a future non-nil error fails the test instead of vanishing.

Left the three older `_ = tp.Shutdown(...)` handlers at lines 247, 278 and 327 alone — they predate this PR and aren't in its diff.

---

.semgrep/msgraph-secrets.yml
```diff
@@ -0,0 +1,71 @@
+rules:
```
julianshen, line 1: _🧪 tests_ | _🟠 Warning_ | _`tgd-review`_

**The security rule has no executable detection tests.**

No Semgrep test fixture is added for this security-sensitive rule. Its behavior depends on nuanced sequence, deep-expression, and metavariable-type matching, so syntax validation alone cannot establish that every intended slog/fmt form, body, and token sink is detected without flagging safe cases such as Header.Get. Add positive and negative `semgrep --test` fixtures covering each pattern family and representative aliasing or formatting forms.

<details>

```
In `.semgrep/msgraph-secrets.yml` around line 1: No Semgrep test fixture is added for this security-sensitive rule. Its behavior depends on nuanced sequence, deep-expression, and metavariable-type matching, so syntax validation alone cannot establish that every intended slog/fmt form, body, and token sink is detected without flagging safe cases such as Header.Get. Add positive and negative `semgrep --test` fixtures covering each pattern family and representative aliasing or formatting forms.

```

</details>

mliu33 (approved):
Thanks!

PR #265, opened 2026-08-13
feat: user permission whitelist with cross-site sync and admin bulk operations

nalydadad:
## Summary

End-to-end user permission whitelist (first key: `external.image.view`), administered from the admin console and enforced across sites:

- **Write path (`admin-service`)** — `POST /v1/admin/permissions` grants/revokes a permission for any number of subject accounts in one call (request caps removed; bulk paste in the console). The materialized `PermissionState` on user docs and the `permission_grants` audit ledger are written in a single Mongo transaction, then fanned out cross-site.
- **Cross-site sync** — chunked fanout (5000 accounts/chunk) published directly to each remote site's INBOX stream, destinations in parallel under one shared 5s budget; peers whose publish did not land are reported as `syncFailures` in the response (the local write has already committed).
- **Resync (`POST /v1/admin/permissions/resync`)** — idempotent re-delivery of the current materialized state (read-only: no writes, no audit entries); stored `updatedAt` is preserved so remote watermark guards no-op anything already applied.
- **Apply path (`inbox-worker`)** — applies `user_permissions_updated` events behind a per-field `updatedAt` watermark guard, so replays and out-of-order deliveries cannot regress state.
- **Read path (`user-service`)** — `settings.get` now returns evaluated permissions alongside settings; the interim `permission.get` RPC stack is removed.
- **Admin console (`admin-frontend`)** — permissions page with bulk paste (dedup + offender highlighting), typeahead account pickers with keyboard/ARIA support, audit view, and a resync flow driven by `syncFailures`.

Design spec and implementation plan are under `docs/superpowers/specs/` and `docs/superpowers/plans/` (annotated as-shipped); the client-facing contract is in `docs/client-api.md` §9.13–§9.15 and the `settings.get` section.

## Testing

- `make lint` clean; `make test` (race) green; admin-service / inbox-worker / user-service integration suites green (testcontainers).
- Frontend: 210/210 tests, typecheck, production build.
- SAST: gosec / govulncheck / semgrep (97 rules) — 0 findings.
- Live docker-local E2E: grant → Mongo txn → cross-site fanout → remote apply → watermark rejection of stale events → `settings.get` composition → full UI resync arc (`syncFailures` banner → Resend → "Sync complete.").

## Deliberately out of scope (follow-ups)

- Unify the inbox-worker `user_*` handlers' poison-message handling on `errcode.Permanent` (all four handlers together, separate PR).
- `eslint-plugin-jsx-a11y` + Modal focus-trap/aria-live sweep in admin-frontend.
- `permission_grants` retention/TTL policy (needs a product retention decision).
- Pre-existing debounce-vs-pagination race in admin list views (now fixable once, in the shared `usePagedAdminList` hook).

### DIFF
--- a/admin-frontend/src/api/admin/index.ts
+++ b/admin-frontend/src/api/admin/index.ts
@@ -74,6 +74,64 @@ export interface AuditFilter {
   limit?: number
 }
 
+/** One row of the permission ledger (mirrors admin-service's `permissionGrantView`). */
+export interface PermissionGrantView {
+  id: string
+  permission: string
+  subjectAccount: string
+  granted: boolean
+  effectiveFrom?: string // "2026-09-01"
+  expiresAt?: string // "2026-12-31"
+  applicantAccount: string
+  approverAccount: string
+  reason: string
+  recordedBy: string
+  recordedAt: string
+}
+
+export interface CreatePermissionsRequest {
+  permission: string
+  subjectAccounts: string[]
+  granted: boolean
+  effectiveFrom?: string
+  expiresAt?: string
+  applicantAccount: string
+  approverAccount: string
+  reason?: string
+}
+
+/** `syncFailures` lists the sites whose whitelist sync failed — the ledger write still
+ * succeeded (201), so those sites need a `resyncPermissions` follow-up. */
+export interface CreatePermissionsResponse {
+  created: number
+  duplicatesIgnored: string[]
+  syncFailures?: string[]
+}
+
+export interface ResyncPermissionsRequest {
+  permission: string
+  accounts: string[]
+}
+
+export interface ResyncPermissionsResponse {
+  syncFailures?: string[]
+}
+
+export interface ListPermissionsResponse {
+  currentlyGranted?: boolean
+  entries: PermissionGrantView[]
+  total: number
+}
+
+/** `subjectAccount` and `permission` are independently optional and combinable; the server
+ * includes `currentlyGranted` in the response only when both are given. */
+export interface ListPermissionsParams {
+  subjectAccount?: string
+  permission?: string
+  page?: number
+  limit?: number
+}
+
 /** Raw shape of admin-service's `userView` as it appears on the wire — the
  * `omitempty` fields may be absent; `normalizeUser` fills the defaults. */
 interface UserViewWire {
@@ -224,4 +282,35 @@ export async function listAudit(
     limit: filter.limit,
   })
   return adminFetch<{ entries: AuditEntry[]; total: number }>(authToken, 'GET', `/audit${qs}`)
+}
+
+/** @throws {AsyncJobError} on a non-2xx response (e.g. `unknown_accounts`, `inactive_subject`). */
+export async function createPermissions(
+  authToken: string,
+  body: CreatePermissionsRequest,
+): Promise<CreatePermissionsResponse> {
+  return adminFetch<CreatePermissionsResponse>(authToken, 'POST', '/permissions', body)
+}
+
+/** Replays the current ledger state for `accounts` to every site's whitelist.
+ * @throws {AsyncJobError} on a non-2xx response (e.g. `unknown_accounts`). */
+export async function resyncPermissions(
+  authToken: string,
+  body: ResyncPermissionsRequest,
+): Promise<ResyncPermissionsResponse> {
+  return adminFetch<ResyncPermissionsResponse>(authToken, 'POST', '/permissions/resync', body)
+}
+
+/** @throws {AsyncJobError} on a non-2xx response. */
+export async function listPermissions(
+  authToken: string,
+  params: ListPermissionsParams = {},
+): Promise<ListPermissionsResponse> {
+  const qs = buildQuery({
+    subjectAccount: params.subjectAccount,
+    permission: params.permission,
+    page: params.page,
+    limit: params.limit,
+  })
+  return adminFetch<ListPermissionsResponse>(authToken, 'GET', `/permissions${qs}`)
 }
--- a/admin-frontend/src/components/shared/AccountPicker/AccountPicker.jsx
+++ b/admin-frontend/src/components/shared/AccountPicker/AccountPicker.jsx
@@ -1,0 +1,163 @@
+import { useState } from 'react'
+import { listUsers } from '@/api'
+import { useHandleAdminError } from '@/hooks/useHandleAdminError'
+import { useDebouncedSearch } from '@/hooks/useDebouncedSearch'
+import { useLatestRequest } from '@/hooks/useLatestRequest'
+import './style.css'
+
+const SEARCH_LIMIT = 10
+
+// Search-to-select account field: typing queries the admin users API (debounced) and only an
+// account present in the results can be selected — free-typed text is never committed via
+// `onChange`, so it can never reach a submit payload. `multiple` renders removable chips backed
+// by an array `value`; single mode holds one account string and hides the input once set.
+export default function AccountPicker({ id, label, authToken, multiple = false, value, onChange, disabled }) {
+  const [options, setOptions] = useState([])
+  const [open, setOpen] = useState(false)
+  // Index into `options`, not a selected account: this is the ARIA-1.2 combobox active
+  // descendant, so focus stays on the input and only `aria-activedescendant` moves.
+  // Reset to 0 wherever `options` is replaced, so it can never dangle past the end.
+  const [highlight, setHighlight] = useState(0)
+  const handleAdminError = useHandleAdminError()
+  const { begin, isCurrent } = useLatestRequest()
+
+  const selected = multiple ? value : value ? [value] : []
+
+  const search = async (q) => {
+    // Stamp every invocation — including the empty-query short-circuit below — so a
+    // still-in-flight older request (e.g. "al" resolving after "ali", or after the field was
+    // cleared) can never win a race and overwrite fresher/cleared state (see
+    // useLatestRequest's own doc comment).
+    const token = begin()
+    if (!q) {
+      setOptions([])
+      setOpen(false)
+      return
+    }
+    try {
+      const result = await listUsers(authToken, { q, page: 1, limit: SEARCH_LIMIT })
+      if (!isCurrent(token)) return // superseded by a newer query
+      setOptions(result.users.filter((u) => !selected.includes(u.account)))
+      setHighlight(0)
+      setOpen(true)
+    } catch (err) {
+      if (!isCurrent(token)) return
+      setOptions([])
+      setOpen(false)
+      handleAdminError(err) // invalid_token logs out; other errors just leave the dropdown empty
+    }
+  }
+
+  const { query, setQuery } = useDebouncedSearch({ onSearch: search })
+
+  const select = (account) => {
+    if (multiple) {
+      if (!value.includes(account)) onChange([...value, account])
+      // Keep the dropdown open (minus the just-picked match) so several accounts from the
+      // same search can be added in a row without re-typing the query each time.
+      setOptions((prev) => prev.filter((u) => u.account !== account))
+      setHighlight(0)
+    } else {
+      onChange(account)
+      setQuery('')
+      setOptions([])
+      setOpen(false)
+    }
+  }
+
+  const optionId = (i) => `${id}-option-${i}`
+  const listOpen = open && options.length > 0
+
+  const onKeyDown = (e) => {
+    if (e.key === 'Escape') {
+      setOpen(false)
+      return
+    }
+    if (!['ArrowDown', 'ArrowUp', 'Enter'].includes(e.key)) return
+    if (!listOpen) {
+      // ArrowDown re-opens a list closed by Escape; without it the only way back is
+      // re-typing the query. Enter/ArrowUp on a closed list fall through to the form.
+      if (e.key === 'ArrowDown' && options.length > 0) {
+        e.preventDefault()
+        setOpen(true)
+      }
+      return
+    }
+    // preventDefault on all three: the arrows would otherwise jump the caret to the
+    // start/end of the input, and Enter would submit the form this picker sits in.
+    e.preventDefault()
+    if (e.key === 'Enter') select(options[highlight].account)
+    else setHighlight((i) => (i + (e.key === 'ArrowDown' ? 1 : options.length - 1)) % options.length)
+  }
+
+  const remove = (account) => {
+    onChange(multiple ? value.filter((a) => a !== account) : '')
+  }
+
+  const showInput = multiple || selected.length === 0
+
+  return (
+    <div className="account-picker">
+      <label htmlFor={id}>{label}</label>
+      {selected.length > 0 && (
+        <ul className="account-picker-chips">
+          {selected.map((account) => (
+            <li key={account} className="account-picker-chip">
+              <span>{account}</span>
+              <button
+                type="button"
+                aria-label={`Remove ${account}`}
+                onClick={() => remove(account)}
+                disabled={disabled}
+              >
+                ×
+              </button>
+            </li>
+          ))}
+        </ul>
+      )}
+      {showInput && (
+        <div className="account-picker-input-wrap">
+          <input
+            id={id}
+            type="text"
+            autoComplete="off"
+            role="combobox"
+            aria-expanded={listOpen}
+            aria-controls={`${id}-listbox`}
+            aria-activedescendant={listOpen ? optionId(highlight) : undefined}
+            value={query}
+            onChange={(e) => setQuery(e.target.value)}
+            onFocus={() => options.length > 0 && setOpen(true)}
+            onBlur={() => setOpen(false)}
+            onKeyDown={onKeyDown}
+            disabled={disabled}
+          />
+          {listOpen && (
+            <ul
+              id={`${id}-listbox`}
+              className="account-picker-options"
+              role="listbox"
+              // Keeps focus on the input across the click so `onBlur` above doesn't close the
+              // dropdown before the option's own onClick has a chance to run.
+              onMouseDown={(e) => e.preventDefault()}
+            >
+              {options.map((u, i) => (
+                <li
+                  key={u.account}
+                  id={optionId(i)}
+                  role="option"
+                  aria-selected={i === highlight}
+                  onClick={() => select(u.account)}
+                >
+                  {u.account}
+                  {u.engName ? ` (${u.engName})` : ''}
+                </li>
+              ))}
+            </ul>
+          )}
+        </div>
+      )}
+    </div>
+  )
+}
--- a/admin-frontend/src/components/shared/AccountPicker/AccountPicker.test.jsx
+++ b/admin-frontend/src/components/shared/AccountPicker/AccountPicker.test.jsx
@@ -1,0 +1,396 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { act, render, screen, fireEvent, waitFor } from '@testing-library/react'
+
+vi.mock('@/context/AuthContext', () => ({ useAuth: vi.fn() }))
+vi.mock('@/api', async (importOriginal) => {
+  const actual = await importOriginal()
+  return { ...actual, listUsers: vi.fn() }
+})
+
+import AccountPicker from './AccountPicker'
+import { useAuth } from '@/context/AuthContext'
+import { listUsers, AsyncJobError } from '@/api'
+
+const ALICE = {
+  id: 'u-1',
+  account: 'alice',
+  siteId: 'site-1',
+  engName: 'Alice',
+  chineseName: '',
+  roles: [],
+  active: true,
+  requirePasswordChange: false,
+}
+const BOB = {
+  id: 'u-2',
+  account: 'bob',
+  siteId: 'site-1',
+  engName: 'Bob',
+  chineseName: '',
+  roles: [],
+  active: true,
+  requirePasswordChange: false,
+}
+
+let logout
+
+beforeEach(() => {
+  vi.clearAllMocks()
+  logout = vi.fn()
+  useAuth.mockReturnValue({ logout })
+  listUsers.mockResolvedValue({ users: [], total: 0 })
+})
+
+async function typeAndWait(input, text) {
+  fireEvent.change(input, { target: { value: text } })
+  await act(async () => {
+    await vi.advanceTimersByTimeAsync(300)
+  })
+}
+
+describe('AccountPicker — single select', () => {
+  it('does not query listUsers on mount', () => {
+    render(<AccountPicker id="p" label="Applicant" authToken="tok" value="" onChange={vi.fn()} />)
+    expect(listUsers).not.toHaveBeenCalled()
+  })
+
+  it('queries listUsers debounced as the admin types and lists matches', async () => {
+    listUsers.mockResolvedValue({ users: [ALICE], total: 1 })
+    render(<AccountPicker id="p" label="Applicant" authToken="tok" value="" onChange={vi.fn()} />)
+    vi.useFakeTimers({ shouldAdvanceTime: true })
+    try {
+      await typeAndWait(screen.getByLabelText(/applicant/i), 'ali')
+      expect(listUsers).toHaveBeenCalledWith('tok', { q: 'ali', page: 1, limit: 10 })
+      expect(await screen.findByRole('option', { name: /alice/i })).toBeInTheDocument()
+    } finally {
+      vi.useRealTimers()
+    }
+  })
+
+  it('selects a matched account on click, committing it via onChange', async () => {
+    listUsers.mockResolvedValue({ users: [ALICE], total: 1 })
+    const onChange = vi.fn()
+    render(<AccountPicker id="p" label="Applicant" authToken="tok" value="" onChange={onChange} />)
+    vi.useFakeTimers({ shouldAdvanceTime: true })
+    try {
+      await typeAndWait(screen.getByLabelText(/applicant/i), 'ali')
+      fireEvent.click(await screen.findByRole('option', { name: /alice/i }))
+    } finally {
+      vi.useRealTimers()
+    }
+    expect(onChange).toHaveBeenCalledWith('alice')
+  })
+
+  it('never commits free-typed text that was not selected from the results', async () => {
+    listUsers.mockResolvedValue({ users: [], total: 0 })
+    const onChange = vi.fn()
+    render(<AccountPicker id="p" label="Applicant" authToken="tok" value="" onChange={onChange} />)
+    vi.useFakeTimers({ shouldAdvanceTime: true })
+    try {
+      await typeAndWait(screen.getByLabelText(/applicant/i), 'zzz-not-a-match')
+    } finally {
+      vi.useRealTimers()
+    }
+    expect(onChange).not.toHaveBeenCalled()
+    expect(screen.queryByRole('option')).not.toBeInTheDocument()
+  })
+
+  it('renders the committed value as a removable chip and hides the search input', () => {
+    render(<AccountPicker id="p" label="Applicant" authToken="tok" value="alice" onChange={vi.fn()} />)
+    expect(screen.getByText('alice')).toBeInTheDocument()
+    expect(screen.queryByLabelText(/applicant/i)).not.toBeInTheDocument()
+  })
+
+  it('clears the selection and re-shows the input when the chip is removed', () => {
+    const onChange = vi.fn()
+    render(<AccountPicker id="p" label="Applicant" authToken="tok" value="alice" onChange={onChange} />)
+    fireEvent.click(screen.getByRole('button', { name: /remove alice/i }))
+    expect(onChange).toHaveBeenCalledWith('')
+  })
+
+  it('closes the dropdown on Escape without committing a selection', async () => {
+    listUsers.mockResolvedValue({ users: [ALICE], total: 1 })
+    const onChange = vi.fn()
+    render(<AccountPicker id="p" label="Applicant" authToken="tok" value="" onChange={onChange} />)
+    vi.useFakeTimers({ shouldAdvanceTime: true })
+    try {
+      const input = screen.getByLabelText(/applicant/i)
+      await typeAndWait(input, 'ali')
+      expect(await screen.findByRole('option', { name: /alice/i })).toBeInTheDocument()
+      fireEvent.keyDown(input, { key: 'Escape' })
+    } finally {
+      vi.useRealTimers()
+    }
+    expect(screen.queryByRole('option')).not.toBeInTheDocument()
+    expect(onChange).not.toHaveBeenCalled()
+  })
+})
+
+describe('AccountPicker — keyboard', () => {
+  it('moves the highlight with ArrowDown and commits it with Enter', async () => {
+    listUsers.mockResolvedValue({ users: [ALICE, BOB], total: 2 })
+    const onChange = vi.fn()
+    render(<AccountPicker id="p" label="Applicant" authToken="tok" value="" onChange={onChange} />)
+    vi.useFakeTimers({ shouldAdvanceTime: true })
+    try {
+      const input = screen.getByLabelText(/applicant/i)
+      await typeAndWait(input, 'a')
+      expect(await screen.findByRole('option', { name: /alice/i })).toBeInTheDocument()
+
+      // The highlight starts on the first match; one ArrowDown moves it to the second.
+      expect(input).toHaveAttribute('aria-activedescendant', 'p-option-0')
+      fireEvent.keyDown(input, { key: 'ArrowDown' })
+      expect(input).toHaveAttribute('aria-activedescendant', 'p-option-1')
+      expect(screen.getByRole('option', { name: /bob/i })).toHaveAttribute('aria-selected', 'true')
+      expect(screen.getByRole('option', { name: /alice/i })).toHaveAttribute('aria-selected', 'false')
+
+      // fireEvent returns false when the handler called preventDefault — Enter must not
+      // fall through and submit the form this picker sits in.
+      expect(fireEvent.keyDown(input, { key: 'Enter' })).toBe(false)
+    } finally {
+      vi.useRealTimers()
+    }
+    expect(onChange).toHaveBeenCalledWith('bob')
+  })
+
+  it('wraps the highlight from the first option to the last on ArrowUp', async () => {
+    listUsers.mockResolvedValue({ users: [ALICE, BOB], total: 2 })
+    const onChange = vi.fn()
+    render(<AccountPicker id="p" label="Applicant" authToken="tok" value="" onChange={onChange} />)
+    vi.useFakeTimers({ shouldAdvanceTime: true })
+    try {
+      const input = screen.getByLabelText(/applicant/i)
+      await typeAndWait(input, 'a')
+      expect(await screen.findByRole('option', { name: /alice/i })).toBeInTheDocument()
+      fireEvent.keyDown(input, { key: 'ArrowUp' })
+      expect(input).toHaveAttribute('aria-activedescendant', 'p-option-1')
+      fireEvent.keyDown(input, { key: 'Enter' })
+    } finally {
+      vi.useRealTimers()
+    }
+    expect(onChange).toHaveBeenCalledWith('bob')
+  })
+
+  it('commits the first match on Enter with no arrow keys pressed', async () => {
+    listUsers.mockResolvedValue({ users: [ALICE, BOB], total: 2 })
+    const onChange = vi.fn()
+    render(<AccountPicker id="p" label="Subjects" authToken="tok" multiple value={[]} onChange={onChange} />)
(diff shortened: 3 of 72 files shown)


admin-frontend/src/api/admin/index.ts
```diff
@@ -74,6 +74,64 @@ export interface AuditFilter {
...
+export interface CreatePermissionsRequest {
+  permission: string
+  subjectAccounts: string[]
+  granted: boolean
+  effectiveFrom?: string
+  expiresAt?: string
+  applicantAccount: string
+  approverAccount: string
+  reason: string
+}
```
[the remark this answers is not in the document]
nalydadad: Fixed — `CreatePermissionsRequest.reason` is now optional (`reason?: string`), matching the dialog's conditional omission and the server's optional handling. Folded into `feat(admin-frontend)` (3001d8eb).

admin-frontend/src/components/shared/AccountPicker/AccountPicker.test.jsx
```diff
@@ -0,0 +1,395 @@
...
+    )
+    vi.useFakeTimers({ shouldAdvanceTime: true })
+    try {
+      await typeAndWait(screen.getByLabelText(/subjects/i), 'a')
+      expect(await screen.findByRole('option', { name: /bob/i })).toBeInTheDocument()
+    } finally {
+      vi.useRealTimers()
+    }
+    expect(screen.queryByRole('option', { name: /^alice$/i })).not.toBeInTheDocument()
+  })
```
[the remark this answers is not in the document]
nalydadad: Fixed as suggested — options render as `account (engName)`, so the anchored `/^alice$/i` matcher could never fail. The test now asserts `getAllByRole('option')` has length 1 and uses the loose `/alice/i` matcher. Folded into `feat(admin-frontend)` (3001d8eb).

admin-frontend/src/hooks/usePagedAdminList.js
```diff
@@ -0,0 +1,98 @@
...
+      }
+    },
+    [authToken, pageSize, handleAdminError, begin, isCurrent],
+  )
+
+  useEffect(() => {
+    setPage(1)
+    fetchPage(defaultFilters, 1)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [authToken])
```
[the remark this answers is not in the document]
nalydadad: Declining this one. The only way to reach `session && !authToken` is a tampered/corrupted sessionStorage bundle — `login()` always persists the full `{authToken, account, siteId}` bundle. And if it does happen, the current flow self-heals: the request 401s with `invalid_token`, `handleAdminError` calls `logout()`, and the user lands back on the login page. Guarding the initial fetch would instead leave a stuck, empty console holding a phantom session, with no recovery path. We prefer the self-healing failure mode.

docs/client-api.md
```diff
@@ -6722,6 +6724,14 @@ Every error response — NATS reply subjects, JetStream async results, and HTTP
...
+| `invalid_reason` | bad_request | admin-service `POST /v1/admin/permissions` (§9.13) (`reason` over 1000 runes) |
+| `missing_permission_fields` | bad_request | admin-service `POST /v1/admin/permissions` (§9.13) (`granted` omitted, or `applicantAccount`/`approverAccount` empty) |
+| `invalid_permission_window` | bad_request | admin-service `POST /v1/admin/permissions` (§9.13) (grant only: malformed date, `effectiveFrom` after `expiresAt`, or `expiresAt` not in the future) |
+| `unexpected_permission_window` | bad_request | admin-service `POST /v1/admin/permissions` (§9.13) (revoke only: `effectiveFrom`/`expiresAt` present with a non-null value; an explicit `null` counts as absent) |
+| `inactive_subject` | bad_request | admin-service `POST /v1/admin/permissions` (§9.13) (a subject account is deactivated) |
+| `unknown_accounts` | not_found | admin-service `POST /v1/admin/permissions` (§9.13) (a subject, applicant, or approver account does not exist; the lookup is company-wide, not filtered by site) |
```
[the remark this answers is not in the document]
nalydadad: Fixed — added the missing `invalid_token` row (code `unauthenticated`) to the §6 present-today catalog, covering the admin-service bearer-token middleware and botplatform's session validation. The gap actually predates this PR (the middleware has emitted `invalid_token` since the sessions work), but the §9.13–§9.15 references made it visible. Folded into `docs(client-api)` (6669f6bf).

docs/client-api.md
```diff
@@ -7607,13 +7883,29 @@ Projected user record returned by all admin user endpoints. The `services` / bcr
...
+| `permission` | string | Permission key this row applies to. |
+| `subjectAccount` | string | The account this row grants or revokes the permission for. |
+| `granted` | boolean | What this row did — `true` for a grant, `false` for a revocation. Historical rows are never rewritten; this is not "does the subject currently hold the permission" (see `currentlyGranted`, §9.14). |
+| `effectiveFrom` | string | `YYYY-MM-DD`, decoded back from the stored instant at UTC+8. Omitted on revoke rows (no validity window). |
+| `expiresAt` | string | `YYYY-MM-DD`, the inclusive last valid day (decoded from the stored exclusive-end instant at UTC+8, minus one day). Omitted on revoke rows. |
+| `applicantAccount` | string | Who requested the change. |
+| `approverAccount` | string | Who approved the change. |
+| `reason` | string | Free-text justification, as submitted. |
+| `recordedBy` | string | The admin account that recorded this row (from the session token). |
+| `recordedAt` | string | RFC 3339. Server clock at write time — also the moment the change took effect. |
```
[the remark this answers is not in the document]
nalydadad: Fixed — `recordedAt` is now described as the server write timestamp only, explicitly independent of the `effectiveFrom`/`expiresAt` validity window (a grant can be back- or future-dated). Folded into `docs(client-api)` (6669f6bf).

docs/client-api/request-reply.md
```diff
@@ -269,6 +269,9 @@ matching `siteId`). Full schemas, examples, and error tables are in
...
 | `POST /v1/admin/rooms/:roomId/onduty` | synchronous HTTP | Toggle a channel's on-duty state: maps the boolean onto `restricted` + `externalAccess` via room-service's restrict RPC, with `ownerAccount` required when turning on. Emits a `room_restricted` room event; no system message, so nothing is displayed (§9.12). |
 | `POST /v1/password/change` | synchronous HTTP | Logged-in admin's self-service password change (§9.11). |
+| `POST /v1/admin/permissions` | synchronous HTTP | Grant or revoke a permission for one or more subject accounts; appends to the permission ledger, materializes the new state on each subject's user doc, and fans it out to every other site (unacknowledged destinations come back as `syncFailures`), with one slim audit entry per subject (§9.13). |
+| `GET /v1/admin/permissions?subjectAccount=<account>` | synchronous HTTP | List an account's permission ledger newest-first, plus the current decision read from the materialized state (§9.14). |
+| `POST /v1/admin/permissions/resync` | synchronous HTTP | Re-deliver the current materialized permission state for the given accounts to every other site; writes nothing, idempotent (§9.15). |
```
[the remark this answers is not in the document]
nalydadad: Fixed — the derived view now documents the full contract: company-wide listing by default, independent optional `subjectAccount`/`permission` filters, and `currentlyGranted` only when both are supplied, matching canonical §9.14. Folded into `docs(client-api)` (6669f6bf).

docs/client-api/request-reply.md
```diff
@@ -1621,8 +1626,9 @@ The stored settings object. All ten fields optional, present only when explicitl
 | `showNotificationsInCall` | boolean |
 | `initialChatScrollPosition` | string (`lastRead`\|`newest`) |
 | `priorityContacts` | string[] (raw accounts, read-only here — written only by `settings.priorityContacts.add`/`.remove`) |
+| `permissions` | map<permission key, boolean> (evaluated admin-managed permissions; every known key always present; read-only — `settings.set` cannot touch it) |
 
-`{ "fullWidth": true, "translateMessageInto": "en-US" }`
+`{ "fullWidth": true, "translateMessageInto": "en-US", "permissions": { "external.image.view": true } }`
```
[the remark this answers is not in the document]
nalydadad: The docs are correct as-is: the canonical wire constant is `PermissionExternalImageView = "external.image.view"` (`pkg/model/permission.go`), and the canonical and derived documents already use it consistently. The mismatch was in the PR description, which wrongly said `external_image_view` — the description has been corrected instead. No repo change. (The dotted key is also why the materialized state lives in a named struct field `externalImageView` rather than a Mongo map: dotted keys cannot be addressed by Mongo dot-path updates.)

pkg/model/permission.go
```diff
@@ -0,0 +1,97 @@
...
+
+// PermissionState is the latest admin decision for one permission key, materialized on
+// the user document and replicated to every site. UpdatedAt is the write's RecordedAt
+// and doubles as the per-key last-write-wins watermark.
+type PermissionState struct {
+	Granted       bool       `json:"granted"                 bson:"granted"`
+	EffectiveFrom *time.Time `json:"effectiveFrom,omitempty" bson:"effectiveFrom,omitempty"`
+	ExpiresAt     *time.Time `json:"expiresAt,omitempty"     bson:"expiresAt,omitempty"`
+	UpdatedAt     time.Time  `json:"updatedAt"               bson:"updatedAt"`
+}
```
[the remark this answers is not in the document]
nalydadad: This is a deliberate trade-off; staying with `$lte` on `updatedAt` for this PR:

- `$lte` (not `$lt`) is required by the resync path: resync re-delivers the *same* state carrying its stored `updatedAt`, and must pass the guard to converge a drifted site. A revision scheme needs the identical equal-is-apply semantics, so it would remove only the same-millisecond-tie window, not the mechanism.
- The tie window requires two *conflicting* admin decisions for the same account+key recorded within one millisecond, then delivered out of order. Permission writes are human-paced admin-console actions, and INBOX delivery is ordered per stream, so reordering additionally requires a redelivery. We judged that combination operationally implausible for this surface.
- If a site ever does diverge, `POST /v1/admin/permissions/resync` re-delivers the current origin state and converges it — that is the designed repair path, and the tie semantics are documented at the guard (`ApplyUserPermissions`, inbox-worker/main.go).

If permission writes ever become programmatic/high-frequency, a persisted per-key revision fanned out with the event is the right hardening — noted as a follow-up, out of scope here.

nalydadad:
Review addressed — all 8 actionable comments answered inline (5 fixed, 3 declined with reasoning). The fixes are folded into the branch's 10-commit narrative per repo convention, so the branch was force-pushed and the original comment anchors now show as outdated; new head is `6669f6bf`. Verification on the final tree: `make lint` clean, frontend 210/210 + typecheck, admin-service integration suite green.

On the 4 nitpicks:

- **`TestIntegration_` prefix (integration_test.go)** — Fixed: the three new `TestRecordPermissionChange_*` tests are renamed with the `TestIntegration_` prefix and verified under a real run (folded into `feat(admin-service)`, 8036a0c5). `TestLoginAndChangePasswordEndToEnd` also lacks the prefix but predates this PR, so it is left alone here.
- **Client-side date window validation (CreatePermissionsDialog)** — Declined: the server evaluates the validity window on UTC+8 business dates, while a client-side `toISOString()` check runs on the UTC date — near midnight the two disagree, so the `min`/window guard would block valid submissions or pass invalid ones. The server's `invalid_permission_window` already surfaces inline in the dialog; client checks stay structural by design.
- **`value ?? []` default (AccountPicker)** — Declined: every current caller passes `value`, and the component's behavior is pinned by tests. Happy to add the default if a caller without `value` ever appears.
- **Functional update in `updateFilter` (usePagedAdminList)** — Declined: the suggested form calls `setDebouncedFilters` inside the `setFilters` updater, and updaters must be pure (StrictMode double-invokes them, double-firing the debounce). No current call site can batch two `updateFilter` calls — each fires from a single input event — so the stale-closure hazard is unreachable today. If a multi-filter atomic update ever appears, we'll restructure then.

julianshen:
**Actionable comments posted: 17** — 🔴 5 blocking · 🟠 12 warning

### 💬 Additional comments (9)


**`admin-service/permissions.go`**

_🏷️ atomicity_ | _🔴 Blocking_ | _`mongodb`_

**Account validation is outside the permission transaction.**

createPermissions reads the users collection through FindAccountStates before RecordPermissionChange starts its transaction. A concurrent deactivation can therefore commit after validation but before RecordPermissionChange; its UpdateMany filters only by account and permission watermark, so it will grant the now-inactive subject and fan that state out. A concurrent removal can similarly leave a committed permission_grants row without a local user state. Repeat subject existence/active validation inside the same Mongo session transaction as the ledger insert and user update, propagate the session context to every operation, and abort the transaction when the transactional validation fails.

---

**`admin-service/store_mongo.go`**

_🎯 concurrency_ | _🔴 Blocking_ | _`performance-and-concurrency`_

**Equal-millisecond writes can leave sites with different permissions**

PermissionState.UpdatedAt is generated before the transaction, but MongoDB stores time.Time with millisecond precision. Two concurrent grant/revoke requests for the same account within one millisecond therefore have equal stored watermarks. Because RecordPermissionChange and ApplyUserPermissions both accept equality with $lte, commit order can select one state locally while per-destination fanout delivery order selects another state remotely. The divergence then persists indefinitely and can leave a revoked permission granted at some sites. Add a unique deterministic operation/version tie-breaker to PermissionState and compare the tuple (updatedAt, version) in both guarded update sites; cover opposite commit/delivery orders with equal-millisecond tests.

---

**`admin-service/store_mongo.go`**

_🏷️ consistency_ | _🟠 Warning_ | _`mongodb`_

**Millisecond-equal writes can leave sites with opposite permissions.**

The users permission update uses only PermissionState.UpdatedAt with $lte as its ordering guard. MongoDB stores time.Time as millisecond BSON datetimes, so concurrent grant and revoke requests created within one millisecond have equal stored watermarks. Both updates are accepted, and separate sites can receive the fanout events in different orders, permanently leaving one granted and another revoked. Add a shared per-change tie-breaker, such as a UUIDv7 change ID, to PermissionState and compare the tuple (updatedAt, changeID) in both the admin-service and inbox-worker guards so delivery order cannot change the winner.

---

**`admin-service/store_mongo.go`**

_🏷️ indexes_ | _🟠 Warning_ | _`mongodb`_

**Independent permission filters are not covered by the new indexes.**

ListPermissionGrants allows subjectAccount and permission independently, but the indexes added are {permission, subjectAccount, recordedAt, _id} and {recordedAt, _id}. A subjectAccount-only query cannot use the first index's equality prefix, so both CountDocuments and the sorted Find must scan the append-only ledger through the browse index and apply a residual filter. Permission-only queries likewise cannot obtain the requested sort from the compound index because subjectAccount lies between permission and recordedAt. Add {subjectAccount:1, recordedAt:-1, _id:-1} and {permission:1, recordedAt:-1, _id:-1} indexes, or narrow the supported filter combinations.

---

**`admin-service/store_mongo.go`**

_🏷️ bson-limits_ | _🟠 Warning_ | _`mongodb`_

**Unbounded account arrays can exceed MongoDB command limits.**

The API intentionally accepts an unbounded account list, but FindAccountStates, GetUserPermissionsForAccounts, and RecordPermissionChange each place the complete list into one {$in: accounts} predicate. A sufficiently large list, or one containing very long unknown account strings, makes the query/update BSON exceed MongoDB's command/document size limit and returns a 500 before the 5,000-account fanout chunking is reached. Chunk these reads and UpdateMany predicates to a BSON-safe size; keep all write chunks inside the existing transaction to preserve all-or-nothing behavior.

---

**`admin-service/store_mongo.go`**

_🏷️ pagination_ | _🟠 Warning_ | _`mongodb`_

**Offset pagination duplicates rows during concurrent ledger inserts.**

ListPermissionGrants uses skip=(page-1)*limit over an append-only, newest-first ledger. If a grant is inserted after page 1 is read, every existing row shifts by one and page 2 repeats the previous page's last row while delaying another row; large page numbers also require MongoDB to walk every skipped index entry. CountDocuments and Find are separate reads, so total and entries can also describe different snapshots. Use range pagination on the stable (recordedAt, _id) tuple, such as a before cursor; if an exact total must accompany a page, compute it in the same aggregation snapshot.

---

**`admin-service/permissions.go`**

_🏷️ resource-exhaustion_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Permission batches have no application-level size bound**

Both permission endpoints accept an unbounded account array, and createPermissions materializes several additional O(N) structures: dedup maps and slices, account-state results, grants, audit entries, InsertMany documents, and fanout payloads. A valid admin request containing an accidentally or maliciously huge list can therefore consume process memory and hold a large Mongo transaction until the service is exhausted. The new test explicitly establishes that there is no cap. Restore an HTTP body limit and a defensible maximum account count, or replace the endpoint with bounded streaming/batched processing that applies backpressure.

---

**`admin-service/store_mongo.go`**

_⚡ database-performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Subject-only ledger queries lack a supporting index**

ListPermissionGrants permits subjectAccount without permission, and the frontend exposes that filter. The added indexes start with either permission or recordedAt, so neither can efficiently satisfy a subjectAccount equality query. On the append-only ledger, CountDocuments must examine the collection/index broadly and the sorted Find may scan most of the recordedAt index to locate one subject. This cost repeats on filter changes and every page. Add an index beginning with subjectAccount and carrying the sort suffix, such as {subjectAccount: 1, recordedAt: -1, _id: -1}.

---

**`admin-service/permissions.go`**

_🏷️ backpressure_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Resync serializes every distinct permission state group**

resyncPermissions groups by the complete serialized state, including UpdatedAt, then waits for publishPermissionFanout to finish for each group before starting the next. Accounts changed by separate admin requests normally have distinct timestamps, so resyncing hundreds of such accounts performs hundreds of sequential JetStream acknowledgement rounds. With the single shared five-second deadline, healthy remote sites can be reported as failures simply because later groups start after the budget is consumed. Flatten group/destination/chunk publishes into a bounded worker pool under the shared context, aggregate failures per destination safely, and stop scheduling once the context expires.

<details>
<summary>📒 Files reviewed (73)</summary>

* `admin-frontend/src/api/_transport/httpEnvelope.test.ts`
* `admin-frontend/src/api/_transport/httpEnvelope.ts`
* `admin-frontend/src/api/admin/admin.test.ts`
* `admin-frontend/src/api/admin/index.ts`
* `admin-frontend/src/api/index.ts`
* `admin-frontend/src/components/AppShell/AppShell.jsx`
* `admin-frontend/src/components/AppShell/AppShell.test.jsx`
* `admin-frontend/src/components/AuditView/AuditView.jsx`
* `admin-frontend/src/components/AuditView/AuditView.test.jsx`
* `admin-frontend/src/components/AuditView/style.css`
* `admin-frontend/src/components/PermissionsView/CreatePermissionsDialog/CreatePermissionsDialog.jsx`
* `admin-frontend/src/components/PermissionsView/CreatePermissionsDialog/CreatePermissionsDialog.test.jsx`
* `admin-frontend/src/components/PermissionsView/CreatePermissionsDialog/index.jsx`
* `admin-frontend/src/components/PermissionsView/CreatePermissionsDialog/parseAccounts.js`
* `admin-frontend/src/components/PermissionsView/CreatePermissionsDialog/parseAccounts.test.js`
* `admin-frontend/src/components/PermissionsView/CreatePermissionsDialog/style.css`
* `admin-frontend/src/components/PermissionsView/PermissionsPage/PermissionsPage.jsx`
* `admin-frontend/src/components/PermissionsView/PermissionsPage/PermissionsPage.test.jsx`
* `admin-frontend/src/components/PermissionsView/PermissionsPage/index.jsx`
* `admin-frontend/src/components/PermissionsView/PermissionsPage/style.css`
* `admin-frontend/src/components/PermissionsView/PermissionsTable/PermissionsTable.jsx`
* `admin-frontend/src/components/PermissionsView/PermissionsTable/index.jsx`
* `admin-frontend/src/components/PermissionsView/PermissionsTable/style.css`
* `admin-frontend/src/components/PermissionsView/index.jsx`
* `admin-frontend/src/components/PermissionsView/permissionKey.js`
* `admin-frontend/src/components/shared/AccountPicker/AccountPicker.jsx`
* `admin-frontend/src/components/shared/AccountPicker/AccountPicker.test.jsx`
* `admin-frontend/src/components/shared/AccountPicker/index.jsx`
* `admin-frontend/src/components/shared/AccountPicker/style.css`
* `admin-frontend/src/hooks/usePagedAdminList.js`
* `admin-frontend/src/styles/index.css`
* `admin-service/config.go`
* `admin-service/deploy/docker-compose.yml`
* `admin-service/handler.go`
* `admin-service/handler_test.go`
* `admin-service/login_test.go`
* `admin-service/main.go`
* `admin-service/room_onduty_integration_test.go`
* `admin-service/room_onduty_test.go`
* `admin-service/routes.go`
* `admin-service/store.go`
* `docker-local/compose.deps.yaml`
* `docs/client-api.md`
* `docs/superpowers/plans/2026-08-13-permission-sync-and-bulk.md`
* `docs/superpowers/specs/2026-08-13-permission-bulk-frontend-design.md`
* `history-service/internal/cassrepo/utils.go`
* `inbox-worker/handler.go`
* `inbox-worker/main.go`
* `pkg/errcode/codes_permission.go`
* `pkg/errcode/codes_test.go`
* `pkg/model/event.go`
* `pkg/model/model_test.go`
* `pkg/model/permission.go`
* `pkg/model/permission_test.go`
* `pkg/model/user.go`
* `tools/jaeger/README.md`
* `tools/seed-sample-data/main_test.go`
* `user-service/models/settings.go`
* `user-service/mongorepo/settings_test.go`
* `user-service/mongorepo/users.go`
* `user-service/service/settings.go`
* `user-service/service/settings_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `silent-failure`
* `tgd-review`

</details>

admin-service/store_mongo.go
```diff
@@ -341,6 +366,194 @@ func (s *storeMongo) ListAudit(ctx context.Context, siteID string, f AuditFilter
...
+		"account": bson.M{"$in": accounts},
+		"$or": bson.A{
+			bson.M{path + ".updatedAt": bson.M{"$exists": false}},
+			bson.M{path + ".updatedAt": bson.M{"$lte": state.UpdatedAt}},
+		},
+	}
+	return s.withTransaction(ctx, func(ctx context.Context) error {
+		if _, err := s.permGrants.InsertMany(ctx, docs); err != nil {
+			return fmt.Errorf("insert permission grants: %w", err)
+		}
```
julianshen, line 415: _🏷️ ordering_ | _🔴 Blocking_ | _`distributed-system`_

**Wall-clock watermarks can silently discard successful revocations.**

A concrete failure occurs with two admin-service replicas: replica A's clock is ahead and records a grant with watermark T+1; replica B later handles a revoke with watermark T. The revoke ledger insert commits, but this guarded UpdateMany matches zero documents and its result is ignored. Fanout carries the older watermark, so other sites also reject it, while the caller still receives 201. BSON datetime millisecond truncation also makes opposite changes within one millisecond equal, allowing delivery order to leave sites with different states. Use a database-issued monotonic per-account/permission revision with a deterministic tie-breaker and propagate it in the event. Until then, treat an unexpected non-match as a conflict rather than reporting the change as created.

<details>

```
In `admin-service/store_mongo.go` around line 415: A concrete failure occurs with two admin-service replicas: replica A's clock is ahead and records a grant with watermark T+1; replica B later handles a revoke with watermark T. The revoke ledger insert commits, but this guarded UpdateMany matches zero documents and its result is ignored. Fanout carries the older watermark, so other sites also reject it, while the caller still receives 201. BSON datetime millisecond truncation also makes opposite changes within one millisecond equal, allowing delivery order to leave sites with different states. Use a database-issued monotonic per-account/permission revision with a deterministic tie-breaker and propagate it in the event. Until then, treat an unexpected non-match as a conflict rather than reporting the change as created.

```

</details>

admin-service/permissions.go
```diff
@@ -0,0 +1,570 @@
...
+	if err := h.store.RecordPermissionChange(ctx, grants, state); err != nil {
+		errhttp.Write(ctx, c, fmt.Errorf("record permission change: %w", err))
+		return
+	}
+
+	// Audit BEFORE the fanout, not after: the fanout can burn its full budget on an
+	// unreachable peer, and this runs on the request ctx — a client disconnect during
+	// that window would cancel the audit write and lose every entry for a ledger write
+	// that already committed. Best-effort either way: a failure is logged, never fatal.
+	action := auditActionPermissionRevoke
```
julianshen, line 275: _🏷️ atomicity_ | _🔴 Blocking_ | _`distributed-system`_

**A crash after commit permanently loses the cross-site update.**

RecordPermissionChange commits the ledger and origin state before any fanout is persisted. If the process is killed after this call returns but before the first JetStream publish, remote sites retain the previous decision indefinitely. No HTTP response or syncFailures value reaches the admin, and no pending work exists to recover after restart; for a revoke this can leave remote access enabled. Write an outbox record in the same Mongo transaction and publish it from a retrying worker, retaining failed work until acknowledged or dead-lettered.

<details>

```
In `admin-service/permissions.go` around line 275: RecordPermissionChange commits the ledger and origin state before any fanout is persisted. If the process is killed after this call returns but before the first JetStream publish, remote sites retain the previous decision indefinitely. No HTTP response or syncFailures value reaches the admin, and no pending work exists to recover after restart; for a revoke this can leave remote access enabled. Write an outbox record in the same Mongo transaction and publish it from a retrying worker, retaining failed work until acknowledged or dead-lettered.

```

</details>

admin-service/permissions.go
```diff
@@ -0,0 +1,570 @@
...
+	// UpdatedAt, so accounts granted across N admin submits yield N groups, and a
+	// per-group budget would stretch an unreachable peer's wall time to N × the
+	// budget — past httpWriteTimeout, which drops the connection before the admin
+	// can be told which peers to retry.
+	fanoutCtx, cancel := fanoutContext(ctx)
+	defer cancel()
+	seen := make(map[string]struct{})
+	var failures []string
+	for k, groupAccounts := range groups {
+		for _, dest := range h.publishPermissionFanout(fanoutCtx, key, groupAccounts, states[k]) {
```
julianshen, line 461: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**A blocked peer prevents later resync groups reaching healthy peers.**

Each state group calls publishPermissionFanout sequentially under the same five-second context, and that function waits for every destination goroutine. If one destination blocks until the deadline on the first group, subsequent groups receive an expired context and fail even for healthy destinations. Build all groups before a single per-destination fanout, or publish the groups concurrently under the shared deadline. Add a test with two groups, a blocked site-b, and a healthy site-c that asserts both groups reach site-c.

<details>

```
In `admin-service/permissions.go` around line 461: Each state group calls publishPermissionFanout sequentially under the same five-second context, and that function waits for every destination goroutine. If one destination blocks until the deadline on the first group, subsequent groups receive an expired context and fail even for healthy destinations. Build all groups before a single per-destination fanout, or publish the groups concurrently under the shared deadline. Add a test with two groups, a blocked site-b, and a healthy site-c that asserts both groups reach site-c.

```

</details>

inbox-worker/main.go
```diff
@@ -217,6 +217,32 @@ func (s *mongoInboxStore) UpdateUserSettings(ctx context.Context, account string
...
+	}
+	path := "permissions." + field
+	filter := bson.M{
+		"account": bson.M{"$in": accounts},
+		"$or": bson.A{
+			bson.M{path + ".updatedAt": bson.M{"$exists": false}},
+			bson.M{path + ".updatedAt": bson.M{"$lte": state.UpdatedAt}},
+		},
+	}
+	if _, err := s.userCol.UpdateMany(ctx, filter, bson.M{"$set": bson.M{path: state}}); err != nil {
```
julianshen, line 240: _🏷️ delivery_ | _🟠 Warning_ | _`distributed-system`_

**Missing remote users cause acknowledged permission events to be lost.**

If a user exists at the admin site but the remote site's HR-provisioned user document is delayed, UpdateMany matches zero for that account and still returns nil. The handler therefore acknowledges the event, the producer reports no sync failure because JetStream accepted it, and later user provisioning creates a document without the permission state. Persist or retry unapplied accounts until their user documents exist; distinguish missing documents from accounts rejected by a newer watermark so only genuinely pending work is retried.

<details>

```
In `inbox-worker/main.go` around line 240: If a user exists at the admin site but the remote site's HR-provisioned user document is delayed, UpdateMany matches zero for that account and still returns nil. The handler therefore acknowledges the event, the producer reports no sync failure because JetStream accepted it, and later user provisioning creates a document without the permission state. Persist or retry unapplied accounts until their user documents exist; distinguish missing documents from accounts rejected by a newer watermark so only genuinely pending work is retried.

```

</details>

inbox-worker/handler.go
```diff
@@ -429,6 +434,25 @@ func (h *Handler) handleUserSettingsUpdated(ctx context.Context, evt *model.Inbo
 	return nil
 }
 
+// handleUserPermissionsUpdated applies one chunk of an admin permission batch. A missing
+// user doc is a silent no-op (store-level guard), matching the other user events.
+func (h *Handler) handleUserPermissionsUpdated(ctx context.Context, evt *model.InboxEvent) error {
+	var e model.UserPermissionsUpdated
+	if err := json.Unmarshal(evt.Payload, &e); err != nil {
+		return fmt.Errorf("unmarshal user_permissions_updated payload: %w", err)
```
julianshen, line 442: _🏷️ compatibility_ | _🟠 Warning_ | _`distributed-system`_

**Unsupported permission keys are acknowledged and cannot replay after upgrade.**

During a rolling rollout of a future permission key, a new producer can publish that key while an older inbox-worker still runs this branch. The worker returns nil here, so the message is acknowledged and permanently removed; the producer already received a JetStream publish acknowledgement and reports no sync failure. Upgrading the worker cannot recover the state without a manual resync. Park unsupported-key events durably for replay, or gate writes on destination capabilities; do not silently acknowledge and discard them.

<details>

```
In `inbox-worker/handler.go` around line 442: During a rolling rollout of a future permission key, a new producer can publish that key while an older inbox-worker still runs this branch. The worker returns nil here, so the message is acknowledged and permanently removed; the producer already received a JetStream publish acknowledgement and reports no sync failure. Upgrading the worker cannot recover the state without a manual resync. Park unsupported-key events durably for replay, or gate writes on destination capabilities; do not silently acknowledge and discard them.

```

</details>

admin-service/main.go
```diff
@@ -41,6 +41,12 @@ func run() error {
 		return fmt.Errorf("init observability: %w", err)
 	}
 
+	// Logged after obs.Init so it lands in the JSON handler like every other line.
+	if len(remoteSites(cfg.AllSiteIDs, cfg.SiteID)) == 0 {
```
julianshen, line 45: _🏷️ nats-operational-safety_ | _🟠 Warning_ | _`nats`_

**Missing peer configuration silently disables permission replication.**

`ALL_SITE_IDS` defaults to empty, and startup only logs when `remoteSites` is empty. In a multi-site deployment where the variable is unset or contains only this site, `publishPermissionFanout` returns no failures without publishing anything; permission writes therefore return 201 with `syncFailures` omitted while every remote remains stale. Distinguish explicit single-site mode from missing federation configuration, and fail startup/readiness when a federated deployment has no validated remote destinations.

<details>

```
In `admin-service/main.go` around line 45: `ALL_SITE_IDS` defaults to empty, and startup only logs when `remoteSites` is empty. In a multi-site deployment where the variable is unset or contains only this site, `publishPermissionFanout` returns no failures without publishing anything; permission writes therefore return 201 with `syncFailures` omitted while every remote remains stale. Distinguish explicit single-site mode from missing federation configuration, and fail startup/readiness when a federated deployment has no validated remote destinations.

```

</details>

admin-frontend/src/components/shared/AccountPicker/AccountPicker.jsx
```diff
@@ -0,0 +1,163 @@
...
+    }
+    try {
+      const result = await listUsers(authToken, { q, page: 1, limit: SEARCH_LIMIT })
+      if (!isCurrent(token)) return // superseded by a newer query
+      setOptions(result.users.filter((u) => !selected.includes(u.account)))
+      setHighlight(0)
+      setOpen(true)
+    } catch (err) {
+      if (!isCurrent(token)) return
+      setOptions([])
```
julianshen, line 45: _🏷️ error-handling_ | _🟠 Warning_ | _`silent-failure`_

**Account lookup failures are presented as empty search results.**

When `listUsers` rejects for a non-auth failure, this catch clears and closes the options, then discards the message returned by `handleAdminError`. A network or server failure is therefore indistinguishable from a valid search with no matches. Because applicant, approver, and picker-mode subjects must be selected from these results, the form remains blocked without explaining why. Store the returned message in picker-level error state, render it beside the input, and clear it when a new search starts or succeeds; keep the existing logout behavior when the handler returns `null`.

<details>

```
In `admin-frontend/src/components/shared/AccountPicker/AccountPicker.jsx` around line 45: When `listUsers` rejects for a non-auth failure, this catch clears and closes the options, then discards the message returned by `handleAdminError`. A network or server failure is therefore indistinguishable from a valid search with no matches. Because applicant, approver, and picker-mode subjects must be selected from these results, the form remains blocked without explaining why. Store the returned message in picker-level error state, render it beside the input, and clear it when a new search starts or succeeds; keep the existing logout behavior when the handler returns `null`.

```

</details>

admin-frontend/src/components/shared/AccountPicker/AccountPicker.jsx
```diff
@@ -0,0 +1,163 @@
...
+    // still-in-flight older request (e.g. "al" resolving after "ali", or after the field was
+    // cleared) can never win a race and overwrite fresher/cleared state (see
+    // useLatestRequest's own doc comment).
+    const token = begin()
+    if (!q) {
+      setOptions([])
+      setOpen(false)
+      return
+    }
+    try {
```
julianshen, line 37: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**The picker cannot select valid accounts from remote sites.**

Every picker searches through listUsers, whose backend search path is scoped by siteId, while the new permission handlers deliberately validate subjects, applicants, and approvers company-wide. Paste mode works around this only for subjects; applicant and approver are picker-only and free-typed text is never committed. Consequently, valid remote-site applicants or approvers cannot be submitted from the console. Provide a company-wide account search for permission pickers or another validated way to enter those accounts.

<details>

```
In `admin-frontend/src/components/shared/AccountPicker/AccountPicker.jsx` around line 37: Every picker searches through listUsers, whose backend search path is scoped by siteId, while the new permission handlers deliberately validate subjects, applicants, and approvers company-wide. Paste mode works around this only for subjects; applicant and approver are picker-only and free-typed text is never committed. Consequently, valid remote-site applicants or approvers cannot be submitted from the console. Provide a company-wide account search for permission pickers or another validated way to enter those accounts.

```

</details>

admin-service/permissions.go
```diff
@@ -0,0 +1,570 @@
...
+	c.JSON(http.StatusCreated, createPermissionsResponse{
+		Created:           len(grants),
+		DuplicatesIgnored: duplicatesIgnored,
+		SyncFailures:      syncFailures,
+	})
+}
+
+// fanoutChunkSize bounds one UserPermissionsUpdated event's Accounts so a worst-case
+// event stays ~320KB — 3× margin under NATS's default 1MB max_payload.
+const fanoutChunkSize = 5000
```
mliu33, line 301: We only have 128kb max size. Use 1000 ?

admin-service/permissions.go
```diff
@@ -0,0 +1,570 @@
...
+	for i, dest := range dests {
+		if failed[i] {
+			failures = append(failures, dest)
+		}
+	}
+	return failures
+}
+
+// remoteSites filters allSiteIDs down to real remote destinations.
+func remoteSites(allSiteIDs []string, self string) []string {
```
mliu33, line 390: We can have this built up initially instead of loop it every time when needed

admin-service/permissions.go
```diff
@@ -0,0 +1,570 @@
...
+
+// fanoutChunkSize bounds one UserPermissionsUpdated event's Accounts so a worst-case
+// event stays ~320KB — 3× margin under NATS's default 1MB max_payload.
+const fanoutChunkSize = 5000
+
+// fanoutTimeout is the server-side budget for ONE request's whole cross-site fanout,
+// however many publishPermissionFanout calls that takes — resync makes one per state
+// group, and a per-call budget would multiply wall time by the group count until it
+// exceeded httpWriteTimeout and the admin lost the syncFailures reply.
+const fanoutTimeout = 5 * time.Second
```
mliu33, line 307: This is too short for whole site fanout. Let's use 30s

admin-service/permissions.go
mliu33: Good point, we should use separate fanoutCtx for each state group here c.c @nalydadad

mliu33 (approved):
Outstanding work! Thanks, some minor suggestions

Issue #23, opened 2026-07-14
Thread mention auto-creates `thread_subscriptions` for non-members — no room-membership check in mention resolution


hmchangw:
## Summary

When a **thread reply** contains `@account` mentions, `message-worker` resolves each `@account` via a **global** user lookup and then upserts a `thread_subscriptions` document plus an entry in `thread_rooms.replyAccounts` for every resolved mentionee — with no check that the mentionee is a member of the room. The record is then federated to the mentionee's home site via outbox/inbox. A non-member who is `@`-mentioned in a thread reply ends up with durable thread-follower state for a room they cannot access.

`notification-worker` does gate on membership before pushing, so non-members do not receive push. But the persisted state survives indefinitely, and `broadcast-worker`'s thread fan-out reads `replyAccounts` directly without a membership intersection — so a non-member in that list receives real-time thread events on `chat.user.{account}.event.room`.

Top-level mentions are not affected: `mention.Resolve` runs on them too, but the result is stored only in the message's `mentions` column. The `markThreadMentions` step that creates subscriptions runs only when `msg.ThreadParentMessageID != ""`.

## What's intentional vs the bug

- **Intentional:** Global mention resolution. `LookupFunc` is an abstract parameter, multiple callers pass different lookups (`message-worker` uses `FindUsersByAccounts`; `broadcast-worker` uses a pre-fetched map); tests pass mock lookups with no room context. The parser is a generic component.
- **Mechanically documented but intent unclear:** The `$setOnInsert` upsert in `MarkThreadSubscriptionMention` preserves existing subscription state on update (`store_mongo.go:94, 105-106`). Whether it was designed to handle "user joins room first vs user is mentioned first" or just adopted because the upsert primitive is convenient is not stated in code or comments.
- **The bug:** No membership gate between mention resolution and persistence. A `@bob` in a thread reply in a room `bob` is not in still produces a `thread_subscriptions` row, a `replyAccounts` entry, and (if `bob` is remote) a federated upsert on `bob`'s home site.

## Evidence

- Global mention resolution, no room/member param:
  - `pkg/mention/mention.go:48-49` — `type LookupFunc func(ctx, accounts) ([]model.User, error)`; godoc at `:58-61`.
  - `message-worker/handler.go:63` — `mention.Resolve(ctx, evt.Message.Content, h.userStore.FindUsersByAccounts)`.
- Thread-only persistence, no membership check:
[...]

PR #270, opened 2026-08-13
feat(mongoutil): let online services boot during a MongoDB outage

idiotech:
## Summary

Online services currently cannot **start** during a MongoDB outage. Two independent startup steps are fatal in every service:

1. `mongoutil.Connect` pings MongoDB and every caller treats failure as fatal (`mongo.Connect` is lazy in driver v2, so this ping *is* the reachability check).
2. `EnsureIndexes` at startup — `createIndexes` is a write, so it fails against a MongoDB that is down *or* degraded to read-only.

A pod that restarts mid-outage (deploy, node drain, OOM kill, HPA scale-up) therefore dies at boot and stays down for the rest of the outage plus the CrashLoopBackOff backoff after MongoDB returns. Across 22 services, a one-hour outage will almost certainly include at least one such restart — which means the request-path resilience work is worth nothing to a process that can't boot.

This PR lets a restarted pod boot, wire its stores, and reach a serving (degraded) state — **without** discarding the startup failures that no restart can fix.

## What changed

**`pkg/mongoutil`** — two additions, both leaving current behaviour as the default:

| Addition | Effect |
|---|---|
| `WithLazyConnect()` | Skips the startup ping so `Connect` returns a client without proving MongoDB is reachable. Lazy boots log at **WARN** since the connection is unproven. |
| `EnsureIndexes(ctx, steps...)` | Runs named index steps concurrently under one 30s budget, and splits failures by whether a restart can fix them (below). |

### Transient vs permanent index failures

This is the part worth reviewing. `EnsureIndexes` classifies:

- **Transient** — MongoDB unreachable or read-only. Logged at WARN with the step name and impact, reported as success, service boots. Self-heals on a later boot, since the step is idempotent and re-runs every start.
- **Permanent** — `E11000` (existing duplicates block a unique index), `85`/`86` (`IndexOptionsConflict` / `IndexKeySpecsConflict`). Returned to the caller, which exits as before.

Permanent failures have to stay fatal: `user-service/mongorepo/users.go` and `room-service/store_mongo.go` detect exactly these codes and return operator remediation text (*"drop the old non-unique `account_1` index … before starting this service"*). Nothing about a restart clears them, and the unique indexes involved are what room-worker's `$setOnInsert` retry-idempotency depends on — silently serving without them lets redelivered `member_added` events duplicate rows.

Steps run **concurrently** so a service with several stores stalls a boot once, not once per store — user-service has five, which would otherwise burn ~150s of an outage before its health server binds, long enough for a k8s startup probe to kill the pod and recreate the crash loop this PR removes.

**Call sites**

- `WithLazyConnect()` on **22 long-running online services** — the 21 online services plus `hr-sync-worker`, which has a health server, durable JetStream consumers and graceful shutdown, so it meets the documented eligibility bar.
- **18 index steps across 12 services** now go through `EnsureIndexes`. `media-service` is included: it had received `WithLazyConnect()` while keeping a fatal `EnsureEmojiIndexes`, so the change was a no-op there and it still crash-looped.
- `message-worker` and `inbox-worker` **create the new `thread_subscriptions` unique index before retiring the legacy one**. Dropping first was safe while a failed create was fatal; now that it isn't, that order could leave the collection with neither index while the service kept serving.
- `botplatform-service` no longer gates `/healthz` on a Mongo ping. A lazily-booted pod failed liveness and crash-looped through the outage anyway, and `docs/health-probes.md` already requires liveness to probe no dependency. The store's now-unused `Ping` is removed.

**Deliberately not opted in:** `data-migration/*`, `teams-*`, `tools/*`, `seed-sample-data`. Batch, migration and CLI jobs have no useful degraded mode — a silent lazy start would turn a clear startup failure into a confusing one deep into a run. They keep the ping and keep failing loudly.

## Behaviour notes

- **Lazy does not defer failure indefinitely.** The first operation still fails, bounded by the caller's context deadline or the driver's 30s server-selection timeout, whichever is shorter — reads return an error rather than hanging, so consumer loops can Nak and retry rather than wedge.
- **No background index retry, on purpose.** The step re-runs on every boot; a retry loop would add a goroutine and a termination path per service for a condition a restart fixes.
- **A recovered client needs no restart.** Once MongoDB returns, the same lazily-connected client works again (covered by an integration test that proxies an outage and then restores it).

## Risk

Unique indexes are the caveat that remains. If a service boots while MongoDB is unwritable, a *transient* failure means the index isn't created, and between MongoDB becoming writable again and the next restart, duplicates it would have rejected can be written. Permanent failures no longer fall into this gap — they stop the service as before.

The WARN line carries the step name and impact so it can be alerted on:

```
level=WARN msg="ensure indexes failed; continuing without them"
  indexes="user-service subscriptions" error="..."
  impact="degraded query performance; unique constraints unenforced until a later start succeeds"
```

## Testing

- `pkg/mongoutil/lazy_test.go` — option plumbing, ping-is-still-the-default, lazy succeeds against an unreachable port, client is usable, first operation fails in **bounded** time.
- `pkg/mongoutil/indexes_test.go` — success is quiet; transient failure warns and reports success; permanent failure (E11000, 85, 86) is returned and *not* downgraded; steps run concurrently; the helper owns a deadline and never loosens a tighter caller one.
- `pkg/mongoutil/lazy_integration_test.go` — an outage proxy boots a client against a dead endpoint, fails bounded while down, then **recovers with no restart** once MongoDB returns.
- `broadcast-worker/startup_resilience_integration_test.go` — walks the service's real Mongo boot sequence with MongoDB down (connect → build store → index step) and asserts all three complete.

## Follow-ups not in this PR

- Index bootstrap still writes from app code on every start. The repo's established pattern for ops/IaC-owned provisioning is `BOOTSTRAP_STREAMS`-style env gating (create in dev, verify-and-fail-fast in production); indexes arguably belong there too.
- Readiness probes NATS only, so a lazily-booted pod reports ready while unable to serve Mongo-backed requests. That follows the documented policy in `docs/health-probes.md`, but the doc's supporting argument predates pods that have never reached Mongo at all.
- Nothing enforces the service-vs-job choice; a `.semgrep` rule would stop service #23 silently picking the wrong one.

## Summary by CodeRabbit

- **New Features**
  - Services can start while MongoDB is temporarily unavailable, with database operations failing promptly and recovering when connectivity returns.
  - Health checks now report service liveness independently of MongoDB availability.
  - Database index setup is coordinated with time limits and clearer handling of temporary versus permanent failures.
- **Bug Fixes**
  - Improved startup resilience by preventing transient index or database connectivity issues from unnecessarily stopping background workers.
  - Updated thread subscription indexes safely, creating replacements before removing legacy indexes.


julianshen:
**Actionable comments posted: 10** — 🔴 4 blocking · 🟠 6 warning

<details>
<summary>📒 Files reviewed (29)</summary>

* `admin-service/main.go`
* `bot-room-service/main.go`
* `broadcast-worker/main.go`
* `broadcast-worker/startup_resilience_integration_test.go`
* `inbox-worker/main.go`
* `media-service/main.go`
* `message-worker/main.go`
* `pkg/mongoutil/indexes.go`
* `pkg/mongoutil/indexes_test.go`
* `pkg/mongoutil/lazy.go`
* `pkg/mongoutil/lazy_integration_test.go`
* `pkg/mongoutil/lazy_test.go`
* `pkg/mongoutil/mongo.go`
* `pkg/mongoutil/observability.go`
* `portal-service/main.go`
* `room-service/main.go`
* `room-worker/main.go`
* `search-service/main.go`
* `tcard-service/main.go`
* `upload-service/main.go`
* `user-service/main.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `silent-failure`
* `tgd-review`

</details>

pkg/mongoutil/indexes.go
```diff
@@ -0,0 +1,41 @@
...
+// restart, duplicate documents could be written that the index would have
+// rejected. Alert on this log line rather than letting it sit.
+//
+// name identifies the step in logs (e.g. "user-service subscriptions") so the
+// warning is actionable and alertable.
+func EnsureIndexesBestEffort(ctx context.Context, name string, ensure func(context.Context) error) {
+	if ensure == nil {
+		return
+	}
+	if err := ensure(ctx); err != nil {
```
julianshen: _🎯 distributed-correctness_ | _🔴 Blocking_ | _`distributed-system`_

**Unique invariants remain disabled after MongoDB recovers.**

During a new deployment or schema rollout, MongoDB can be unavailable or read-only when a newly required unique index is created. This helper swallows that failure, so callers continue initialization; when MongoDB later becomes writable without another restart, writes proceed without the constraint. Duplicate logical users, subscriptions, or tokens can then be committed, and a later index retry/restart may itself fail because duplicates now exist. The file explicitly acknowledges both the unique-index gap and the absence of recovery. Split correctness-critical unique indexes from performance-only indexes: retry critical creation and keep affected writers/consumers unready or paused until those indexes are verified; only non-unique performance indexes should be best-effort.

<details>

```
In `pkg/mongoutil/indexes.go` around line 34: During a new deployment or schema rollout, MongoDB can be unavailable or read-only when a newly required unique index is created. This helper swallows that failure, so callers continue initialization; when MongoDB later becomes writable without another restart, writes proceed without the constraint. Duplicate logical users, subscriptions, or tokens can then be committed, and a later index retry/restart may itself fail because duplicates now exist. The file explicitly acknowledges both the unique-index gap and the absence of recovery. Split correctness-critical unique indexes from performance-only indexes: retry critical creation and keep affected writers/consumers unready or paused until those indexes are verified; only non-unique performance indexes should be best-effort.

```

</details>

pkg/mongoutil/indexes.go
```diff
@@ -0,0 +1,41 @@
...
+		slog.Warn("ensure indexes failed; continuing without them",
```
julianshen: _🏷️ data-integrity_ | _🔴 Blocking_ | _`mongodb`_

**Unique constraints can remain absent after MongoDB recovers.**

This helper suppresses every createIndexes failure and performs no retry. On a fresh collection or a rollout adding a unique index, MongoDB can be unavailable during startup, recover while the same process remains running, and then accept duplicate writes until another restart. Those duplicates can make the next unique-index creation fail with E11000, so the documented claim that the next restart self-heals is not reliable. Separate correctness-critical unique indexes from optional performance indexes; keep the service unready or retry unique-index creation after recovery before permitting writes, while only non-unique indexes are best-effort.

<details>

```
In `pkg/mongoutil/indexes.go` around line 35: This helper suppresses every createIndexes failure and performs no retry. On a fresh collection or a rollout adding a unique index, MongoDB can be unavailable during startup, recover while the same process remains running, and then accept duplicate writes until another restart. Those duplicates can make the next unique-index creation fail with E11000, so the documented claim that the next restart self-heals is not reliable. Separate correctness-critical unique indexes from optional performance indexes; keep the service unready or retry unique-index creation after recovery before permitting writes, while only non-unique indexes are best-effort.

```

</details>

julianshen: _🏷️ error-handling_ | _🔴 Blocking_ | _`silent-failure`_

**Failed critical indexes are never retried or surfaced**

A transient startup outage makes `ensure` fail, after which this helper only logs and permanently forgets the failure. The lazy-client recovery tests explicitly expect the same process to continue after MongoDB returns, so the claimed next-start self-healing may never happen. Normal writes then resume without required indexes; unique constraints can admit duplicates, and those duplicates can make every later index-creation attempt fail permanently while this helper continues reporting startup success. Keep best-effort handling only for proven performance-only indexes. Required unique or other correctness-critical indexes must remain surfaced through readiness, or be retried after recovery until creation succeeds; an exhausted retry must become a health/startup failure.

<details>

```
In `pkg/mongoutil/indexes.go` around line 35: A transient startup outage makes `ensure` fail, after which this helper only logs and permanently forgets the failure. The lazy-client recovery tests explicitly expect the same process to continue after MongoDB returns, so the claimed next-start self-healing may never happen. Normal writes then resume without required indexes; unique constraints can admit duplicates, and those duplicates can make every later index-creation attempt fail permanently while this helper continues reporting startup success. Keep best-effort handling only for proven performance-only indexes. Required unique or other correctness-critical indexes must remain surfaced through readiness, or be retried after recovery until creation succeeds; an exhausted retry must become a health/startup failure.

```

</details>

pkg/mongoutil/indexes.go
julianshen: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**Permanent index errors are incorrectly treated as outages.**

This suppresses every EnsureIndexes error, including conflicting or invalid index specifications and duplicate-data failures while building unique indexes. Those errors do not self-heal, yet callers cannot distinguish them because the helper returns nothing, so services may run indefinitely without required constraints. Suppress only transient availability/read-only errors and return permanent errors so startup can fail; add tests for both classifications.

<details>

```
In `pkg/mongoutil/indexes.go` around line 34: This suppresses every EnsureIndexes error, including conflicting or invalid index specifications and duplicate-data failures while building unique indexes. Those errors do not self-heal, yet callers cannot distinguish them because the helper returns nothing, so services may run indefinitely without required constraints. Suppress only transient availability/read-only errors and return permanent errors so startup can fail; add tests for both classifications.

```

</details>

user-service/main.go
```diff
@@ -92,26 +93,11 @@ func main() {
...
-	}
-	if err := threadSubRepo.EnsureIndexes(ctx); err != nil {
-		slog.Error("ensure indexes failed", "error", err)
-		os.Exit(1)
-	}
-	if err := ssoTokenRepo.EnsureIndexes(ctx); err != nil {
-		slog.Error("ensure indexes failed", "error", err)
-		os.Exit(1)
-	}
+	mongoutil.EnsureIndexesBestEffort(ctx, "user-service subscriptions", subRepo.EnsureIndexes)
```
julianshen: _🏷️ failure-behavior_ | _🟠 Warning_ | _`distributed-system`_

**Sequential index checks can stall outage startup for 150 seconds.**

With MongoDB unreachable, lazy Connect returns immediately, but each of these five sequential createIndexes operations can independently consume the driver's documented 30-second server-selection timeout because they share the general service context rather than one bounded index-phase deadline. The user service can therefore remain in startup for roughly 150 seconds; a shorter startup or rollout budget will kill it and repeat the cycle, defeating the intended ability to boot during an outage. Use one short context deadline for the entire index phase, or stop attempting the remaining indexes after the first reachability failure and reconcile them asynchronously after recovery.

<details>

```
In `user-service/main.go` around line 96: With MongoDB unreachable, lazy Connect returns immediately, but each of these five sequential createIndexes operations can independently consume the driver's documented 30-second server-selection timeout because they share the general service context rather than one bounded index-phase deadline. The user service can therefore remain in startup for roughly 150 seconds; a shorter startup or rollout budget will kill it and repeat the cycle, defeating the intended ability to boot during an outage. Use one short context deadline for the entire index phase, or stop attempting the remaining indexes after the first reachability failure and reconcile them asynchronously after recovery.

```

</details>

user-service/main.go
```diff
@@ -92,26 +93,11 @@ func main() {
 	appRepo := mongorepo.NewAppRepo(db, readFromSecondary)
```
julianshen, line 93: _🏷️ lifecycle_ | _🟠 Warning_ | _`mongodb`_

**Sequential index attempts can still stall startup for minutes.**

These five synchronous calls use the shared startup context without a local deadline. With lazy connection and MongoDB unreachable, each createIndexes operation can consume the driver's documented default 30-second server-selection timeout, delaying user-service startup by roughly 150 seconds before it continues. That can exceed orchestration startup probes and preserve the CrashLoop behavior this change intends to prevent. Use one bounded context for the entire best-effort index phase, or make EnsureIndexesBestEffort impose a short configurable deadline; the same issue applies to other callers that pass an unbounded startup context.

<details>

```
In `user-service/main.go` around line 93: These five synchronous calls use the shared startup context without a local deadline. With lazy connection and MongoDB unreachable, each createIndexes operation can consume the driver's documented default 30-second server-selection timeout, delaying user-service startup by roughly 150 seconds before it continues. That can exceed orchestration startup probes and preserve the CrashLoop behavior this change intends to prevent. Use one bounded context for the entire best-effort index phase, or make EnsureIndexesBestEffort impose a short configurable deadline; the same issue applies to other callers that pass an unbounded startup context.

```

</details>

broadcast-worker/startup_resilience_integration_test.go
```diff
@@ -0,0 +1,93 @@
...
+	require.NotNil(t, store, "store must be constructible with MongoDB unreachable")
+
+	// Gate 2: the index step must warn and continue, not exit.
+	indexCtx, indexCancel := context.WithTimeout(ctx, 10*time.Second)
+	defer indexCancel()
+	require.NotPanics(t, func() {
+		mongoutil.EnsureIndexesBestEffort(indexCtx, "broadcast-worker store", store.EnsureIndexes)
+	}, "failed index creation must not stop startup")
+
+	// Serving state: the store answers requests with an error rather than
```
julianshen, line 57: _🧪 tests_ | _🟠 Warning_ | _`nats`_

**The startup test never exercises the JetStream consumer.**

The claimed serving-state check calls `store.GetRoom` directly; it never starts the worker's JetStream consumer or delivers a message. Consequently, the new Mongo-outage path could acknowledge database failures, hot-loop with immediate NAKs, or exhaust `MaxDeliver`, and this test would still pass. Exercise the actual consumer against test JetStream: publish an event requiring Mongo, verify bounded redelivery without acknowledgement while Mongo is unavailable, then verify successful processing after recovery.

<details>

```
In `broadcast-worker/startup_resilience_integration_test.go` around line 59: The claimed serving-state check calls `store.GetRoom` directly; it never starts the worker's JetStream consumer or delivers a message. Consequently, the new Mongo-outage path could acknowledge database failures, hot-loop with immediate NAKs, or exhaust `MaxDeliver`, and this test would still pass. Exercise the actual consumer against test JetStream: publish an event requiring Mongo, verify bounded redelivery without acknowledgement while Mongo is unavailable, then verify successful processing after recovery.

```

</details>

user-service/main.go
julianshen: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Serial index attempts can delay degraded startup by minutes.**

When MongoDB is unreachable and the startup context has no deadline, each synchronous EnsureIndexesBestEffort call can consume the driver's default 30-second server-selection timeout. These five calls run serially, so user-service may take roughly 150 seconds to reach its serving state; startup probes can repeatedly restart it, defeating lazy-connect resilience. Wrap the entire index batch in one short shared timeout, or perform one bounded reachability check and skip the batch when MongoDB is unavailable. The other multi-index startup call sites need the same aggregate bound.

<details>

```
In `user-service/main.go` around line 96: When MongoDB is unreachable and the startup context has no deadline, each synchronous EnsureIndexesBestEffort call can consume the driver's default 30-second server-selection timeout. These five calls run serially, so user-service may take roughly 150 seconds to reach its serving state; startup probes can repeatedly restart it, defeating lazy-connect resilience. Wrap the entire index batch in one short shared timeout, or perform one bounded reachability check and skip the batch when MongoDB is unavailable. The other multi-index startup call sites need the same aggregate bound.

```

</details>

julianshen: _🏷️ availability_ | _🟠 Warning_ | _`tgd-review`_

**Sequential index attempts can still hold startup for minutes.**

These five calls run synchronously using the shared startup context. With lazy connection, each EnsureIndexes call becomes a network operation, and lazy.go documents a default server-selection timeout of 30 seconds. During an outage this phase can therefore consume multiple selection windows before the service reaches its serving setup, undermining the intended startup resilience. Give the whole phase a short deadline or perform bounded index reconciliation asynchronously.

<details>

```
In `user-service/main.go` around line 96: These five calls run synchronously using the shared startup context. With lazy connection, each EnsureIndexes call becomes a network operation, and lazy.go documents a default server-selection timeout of 30 seconds. During an outage this phase can therefore consume multiple selection windows before the service reaches its serving setup, undermining the intended startup resilience. Give the whole phase a short deadline or perform bounded index reconciliation asynchronously.

```

</details>

pkg/mongoutil/indexes.go
```diff
@@ -0,0 +1,41 @@
...
+// the whole point is that callers must not treat this as a reason to exit.
+//
+// Rationale: index existence is an ops/IaC steady-state concern, in the same
+// family as stream bootstrap (which services already must not do in
+// production). createIndexes is a write, so it fails against a read-only
+// primary as well as a down one — refusing to boot on that is strictly worse
+// than booting without it. The step is idempotent and runs on every start, so
+// a boot that skips it self-heals on the next restart.
+//
+// There is no background retry, on purpose. The step already retries on every
```
julianshen: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**A recovered client never retries skipped index creation.**

The lazy client is designed to recover without a process restart, but index creation is attempted only once. If MongoDB is unavailable while a fresh or newly added index is missing, the service later resumes writes without creating it and may remain that way indefinitely. The logged impact explicitly acknowledges that unique constraints can be unenforced, allowing duplicates that may prevent later index creation. Add bounded background reconciliation after recovery, especially for required unique indexes, or provision and verify those indexes externally before enabling writes.

<details>

```
In `pkg/mongoutil/indexes.go` around line 20: The lazy client is designed to recover without a process restart, but index creation is attempted only once. If MongoDB is unavailable while a fresh or newly added index is missing, the service later resumes writes without creating it and may remain that way indefinitely. The logged impact explicitly acknowledges that unique constraints can be unenforced, allowing duplicates that may prevent later index creation. Add bounded background reconciliation after recovery, especially for required unique indexes, or provision and verify those indexes externally before enabling writes.

```

</details>

inbox-worker/main.go
```diff
@@ -399,18 +399,20 @@ func (s *mongoInboxStore) UpdateSubscriptionRead(ctx context.Context, roomID, ac
...
 		Keys:    bson.D{{Key: "threadRoomId", Value: 1}, {Key: "userAccount", Value: 1}},
 		Options: options.Index().SetUnique(true),
 	}); err != nil {
 		return fmt.Errorf("ensure thread_subscriptions (threadRoomId,userAccount) index: %w", err)
 	}
+
+	// Best-effort: retire the legacy (threadRoomId, userId) unique index. Mirrors
+	// message-worker's threadStoreMongo.EnsureIndexes; the index may not exist
+	// (fresh deploy / test container), so ignore all errors.
+	_ = s.threadSubCol.Indexes().DropOne(ctx, "threadRoomId_1_userId_1") //nolint:errcheck
```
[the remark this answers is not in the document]
hmchangw: Fixed in 1f00a8b, at both sites.

The retirement now goes through a shared `mongoutil.DropIndexIfExists`, which treats only `NamespaceNotFound` (26) and `IndexNotFound` (27) as "nothing to drop" and returns everything else wrapped.

This mattered more than it did before this PR: the `CreateOne` above it is now best-effort, so with the old `_ = ...DropOne(...)` a boot could report a completed retirement while the legacy `(threadRoomId, userId)` unique index was still in place — rejecting exactly the federated upserts the `(threadRoomId, userAccount)` replacement exists to accept. The returned error surfaces through `EnsureIndexes` as its transient WARN rather than as silence.

Covered by `TestIsIndexAbsentError` (classification) and `pkg/mongoutil/dropindex_integration_test.go` (real drop, absent collection/index, and an unreachable MongoDB reported rather than swallowed).

---

pkg/mongoutil/indexes_test.go
```diff
@@ -0,0 +1,222 @@
...
+			if active > peak {
+				peak = active
+			}
+			mu.Unlock()
+			wg.Done()
+			wg.Wait() // every step must be in flight before any returns
+			mu.Lock()
+			active--
+			mu.Unlock()
+			return nil
```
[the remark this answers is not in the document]
hmchangw: Fixed in 1f00a8b.

Steps now report themselves in flight over a buffered channel and park on a `select` that also watches `c.Done()`, and `EnsureIndexes` runs on its own goroutine while the test collects the start signals under a 5s bound. A serial implementation trips that bound and fails with `only 1 of 4 steps ran concurrently` instead of blocking to the test binary's timeout.

Verified by temporarily serializing `EnsureIndexes` (awaiting each `g.Go` before starting the next): the test failed in 5.00s with that message, rather than hanging.

---

broadcast-worker/startup_resilience_integration_test.go
hmchangw: Fixed in d222d24 — you're right that the old assertion couldn't see any of this.

`TestStartup_ConsumerRedeliversUntilMongoRecovers` now runs the worker's own `buildConsumerConfig`, `broadcastProcessor` and `consumeLoop` against a MongoDB that is down and then comes back (via an outage proxy promoted from `pkg/mongoutil`'s test file to `pkg/testutil`, so both packages share it). A canonical `message.created` event is published while Mongo is unreachable, and the test asserts:

- the event is **redelivered** (`deliveries >= 2`) — a failed store read that got Acked would never come back, and that's the silent event-loss case;
- nothing is broadcast from the failed read;
- `deliveries < MaxDeliver` at that point — a bare `Nak()` would have burned the budget in milliseconds, which is the other way the outage path loses the event;
- after `Restore`, in the same process with no restart, the next redelivery broadcasts exactly once on `subject.RoomEvent("r1", true)`;
- and the consumer ends with `NumAckPending == 0 && NumPending == 0`, i.e. the success was actually Acked.

The original `store.GetRoom` gate is kept as the first test — it still covers connect → build store → index step, which is the part that used to `os.Exit(1)`.

One caveat worth stating plainly: Docker isn't available in this session, so I could not execute the integration test locally — it compiles and vets under `-tags integration`, and CI's `test-integration (broadcast-worker)` job is the first real run.

---

user-service/main.go
hmchangw: This one is already addressed — it describes the pre-`abb71d3` shape of the code, where each store had its own `EnsureIndexesBestEffort(ctx, …)` call taking the unbounded startup context.

`user-service/main.go:96-102` is now a **single** `mongoutil.EnsureIndexes` call with the five steps passed as arguments:

```go
if err := mongoutil.EnsureIndexes(ctx,
    mongoutil.Step("user-service subscriptions", subRepo.EnsureIndexes),
    mongoutil.Step("user-service users", userRepo.EnsureIndexes),
    mongoutil.Step("user-service apps", appRepo.EnsureIndexes),
    mongoutil.Step("user-service thread_subscriptions", threadSubRepo.EnsureIndexes),
    mongoutil.Step("user-service sso_tokens", ssoTokenRepo.EnsureIndexes),
); err != nil {
```

The helper derives one `context.WithTimeout(ctx, EnsureIndexesTimeout)` (30s) covering the whole phase and runs the steps concurrently under an `errgroup`, so the worst case is one server-selection window for all five rather than five of them — ~30s, not ~150s. It also bounds callers that pass a deadline-less startup context, which was the other half of the concern.

Both properties are pinned by tests: `TestEnsureIndexes_RunsStepsConcurrently` (concurrency, and since d222d24 it fails cleanly rather than hanging if that regresses), `TestEnsureIndexes_BoundsTheSteps` (a deadline exists even when the caller sets none), and `TestEnsureIndexes_KeepsShorterCallerDeadline` (a tighter caller budget is not loosened).

No code change for this thread.

---

hmchangw:
Closing in favour of #284, which carries the same work rebased onto current `main`.

---

PR #258, opened 2026-08-12
docs: reorganize load-testing by test type and consolidate stray branches

flywindy:
## Summary

Reorganize `docs/load-testing/` by **test type**, and fold in load-testing content that was stranded on unmerged branches.

The directory had grown three competing organizing axes at the same level: by test type (`system/`, `failure-testing/`), by dependency (`cassandra/`, `*-failure-test-plan.md`), and by tool implementation (`loadgen-failure-*.md`, `dashboard-evidence-contract.md`). `main` also held a half-migrated state — flat failure plans alongside a `failure-testing/` directory opened by a later PR.

```
docs/load-testing/
├── common/       shared inputs and acceptance criteria
├── soak/         sustained realistic load, non-destructive
├── failure/      dependency fault campaigns under continuous load
├── performance/  end-to-end SLO validation and ramp-to-breach capacity
└── loadgen/      the load generator's own implementation contracts
```

`cassandra/` is dissolved: its soak plan exercises the whole `gatekeeper → canonical → worker → Cassandra` path and belongs in `soak/`, while `run-a-implementation-plan.md` is a loadgen engineering plan and belongs in `loadgen/`. `system/` becomes `performance/` now that `sli-slo.md` — which every program asserts against — has moved to `common/`.

## Content consolidated from unmerged branches

| Content | Was stranded on | Now |
|---|---|---|
| "MongoDB is exercised but not validated" section + corrected seeded-subscription estimate | `claude/cassandra-usage-load-test-6nd66i` (`483a70f`, `122e1b8`) | `soak/cassandra-soak-plan.md` |
| End-to-end plan — Run 1 / Run 2 structure, runbook, infra confirmations | `claude/stress-testing-section-rentq1` (`0712355`) | merged into `performance/end-to-end-plan.md` |
| End-to-end plan — scenario inventory, SLO relationship | this branch | merged into `performance/end-to-end-plan.md` |
| NATS metrics contract, observability deployment, core-message campaign runbook | this branch | `failure/` |
| `common/workload-model.md`, `common/environments-and-data-ownership.md` | this branch | `common/` |

The MongoDB section documents which MongoDB paths a soak run touches, which of them are cached, and why the run is not a valid MongoDB capacity test. It is the starting baseline for the upcoming MongoDB failure campaigns, and existed nowhere on `main`.

[...]

PR #272, opened 2026-08-13
feat: add application-side NATS failure metrics

flywindy:
## Summary

Implements Workstream B application-side NATS and JetStream failure telemetry for the first core-message campaign. The change provides one shared bounded-cardinality package, adopts it in `message-gatekeeper`, `message-worker`, `broadcast-worker`, and `notification-worker`, and adds opt-in NATS connection lifecycle evidence for those same four services.

The branch is synchronized with the current `main`. It does not change Ack/Nak/Term policy, introduce retries, add a telemetry provider, or instrument unrelated services.

## Service-by-service evidence

| Service | Consumer and disposition evidence | Publish/request evidence | Domain evidence |
|---|---|---|---|
| `message-gatekeeper` | `main.go` wires tracked MESSAGES deliveries; `handler.go` records Ack/Nak, invalid payload, retry, and final canonical-publish exhaustion | Canonical JetStream PubAck, client responses, and history request/reply | `message_gatekeeper_messages_total{result,reason}` |
| `message-worker` | `main.go` tracks canonical and Teams-batch deliveries; poison payloads are terminal; transient persistence failures Nak with existing backoff | Recipient/outbox publish attempts; all four Cassandra batches use the o11y batch seam | `message_worker_persistence_total{message_kind,result}` |
| `broadcast-worker` | `natsmetrics.Start` tracks the canonical loop and guarded handler disposition | Actual recipient Core NATS attempts and history requests | fan-out size, recipient success/failure, and `publish_exhausted` evidence for swallowed partial fan-out |
| `notification-worker` | Canonical delivery, redelivery, Ack/Nak, poison, and shutdown state | Notification/push publish attempts and presence/history requests | sent, suppressed, publish-failed, and failed outcomes |

## Shared metric contract

| Instrument | Labels | Bounded values |
|---|---|---|
| `chat.nats.consumer.loop.up` | `service_name,site,stream,consumer` | Deployment wiring only |
| `chat.nats.consumer.messages` | consumer labels + `event_type,outcome` | event: `created,updated,deleted,pinned,unpinned,reacted,thread_reply_added,send,teams_batch,unknown`; outcome: `ack,nak,term,left_pending,handler_cancelled` |
| `chat.nats.consumer.redeliveries` | consumer labels + `event_type` | Emitted only when JetStream `NumDelivered > 1` |
| `chat.nats.consumer.processing.duration` | consumer labels + `event_type,outcome` | Seconds with explicit o11y latency buckets |
| `chat.nats.publish.attempts` | `service_name,site,destination_kind,operation,outcome` | destination: `canonical,recipient_event,notification,push,outbox,inbox,client_response,user_sync,unknown`; outcome: `success,timeout,no_responders,disconnected,buffer_full,permission,payload_too_large,other_error` |
| `chat.nats.publish.retries` | `service_name,site,destination_kind,operation` | Declared; no producer because no target path has an application-managed publish loop |
| `chat.nats.terminal.failures` | consumer labels + `event_type,reason` | `max_deliver,permanent,publish_exhausted,consumer_deleted,stream_unavailable,invalid_payload,internal` |
| `chat.nats.requests` / `chat.nats.request.duration` | `service_name,site,operation,outcome` | Measured request/reply calls only; duration uses seconds and explicit latency buckets |
| `chat.nats.client.connected` | OTel resource only | Live connection count in the process |
| `chat.nats.client.connection.events` | `event` | `connected,disconnected,reconnected,closed,async_error` |

Connection metrics are opt-in through `natsutil.ConnectWithMetrics`; only the four campaign services call it. Each connection owns independent lifecycle state, while the shared up-down instrument aggregates the live connection count. A nil meter provider degrades to the ordinary connection path and never blocks startup.

## Domain metric contract

| Instrument | Labels and bounded values |
|---|---|
| `message_gatekeeper_messages_total` | result: `accepted,rejected,retry,failed`; reason: `none,invalid_subject,invalid_payload,not_subscribed,room_restricted,canonical_publish,dependency,unknown` |
| `message_worker_persistence_total` | message_kind: `user,system,thread_reply,teams_migration,unknown`; result: `success,error` |
| `broadcast_worker_fanout_recipients` | room_kind: `channel,dm,bot_dm,thread,unknown`; bounded `event_type` |
| `broadcast_worker_recipient_deliveries_total` | fan-out labels + result: `success,failed` |
| `notification_worker_outcomes_total` | kind: `push,notification,unknown`; result: `sent,suppressed,publish_failed,failed` |

No concrete subject, room ID, account, user ID, message ID, request ID, address, error text, or peer identity is used as a metric label.

## Ack, Nak, Term, and terminal semantics

- One `chat.nats.consumer.messages` disposition is emitted per JetStream delivery attempt.
- `Ack` and successful `DoubleAck` record `ack`; `Nak` and `NakWithDelay` record `nak`; `Term` records `term`; a failed disposition remains `left_pending`.
- A handler that returns without a disposition records `left_pending` while the loop and context are live, or `handler_cancelled` during shutdown.
- The loop gauge becomes 1 only after iterator creation and becomes 0 before terminal iterator/Next return and before graceful iterator stop.
- Redelivery is derived only from JetStream metadata where `NumDelivered > 1`.
- `max_deliver`, explicit permanent Term, poisoned payloads, loop failure, and swallowed partial fan-out use bounded terminal reasons. Marking is once-only per delivery.
- Gatekeeper validation/policy rejections are business outcomes, not terminal-loss evidence.
- Core NATS `outcome="success"` means accepted by the client/write or reconnect buffer, not recipient delivery. `ErrReconnectBufExceeded` is recorded at the publish boundary as `outcome="buffer_full"`.
- Metrics decorators preserve existing errors and business behavior. In particular, `broadcastMetricPublisher` returns the already-contextualized `natsPublisher.Publish` error unchanged instead of duplicating the subject wrapper.

## Tests added for the review follow-up

- Independent multi-connection state transitions and aggregate live count.
- Opt-in connection API and nil-provider no-op behavior.
- Gatekeeper accepted, rejected, retry, and exhausted canonical-publish outcomes.
- Message persistence success and Cassandra failure outcomes.
- Notification sent, suppressed, and publish-failed outcomes.
- Broadcast zero-recipient, full-success, partial-success, total-failure, and duplicate-recipient attempts, including partial-fan-out terminal evidence.
- Injected domain metrics for production handlers, with no redundant global-meter construction on the production path.

## Validation

- `make fmt` — completed
- `make test` — all unit tests pass with the race detector
- `make lint` — 0 issues
- `make sast-gosec` — pass
- `make sast-vuln` — no called vulnerabilities
- `make sast-semgrep` — 0 blocking findings
- `make test-integration SERVICE=broadcast-worker` — pass
- `make test-integration SERVICE=message-gatekeeper` — pass
- `make test-integration SERVICE=message-worker` — pass
- `make test-integration SERVICE=notification-worker` — pass
- `make test-integration SERVICE=pkg/natsutil` — pass
- `go.mod` / `go.sum` unchanged by this workstream

### Coverage evidence

| Scope | Statement coverage |
|---|---:|
| `pkg/natsmetrics` | 94.9% |
| `pkg/natsutil` | 88.0% |
| `broadcast-worker/handler.go` | 85.0% |
| `message-gatekeeper/handler.go` | 91.4% |
| `message-worker/handler.go` | 95.3% |
| `notification-worker/handler.go` | 95.0% |
| four service-local `nats_metrics.go` files | 93.3%–95.0% |

The existing flat `package main` services remain below 80% when process wiring and database adapters are included. Unit+integration package coverage is 70.0% (`broadcast-worker`), 65.2% (`message-gatekeeper`), 70.1% (`message-worker`), and 62.0% (`notification-worker`). This PR does not hide those legacy boundaries with coverage exclusions; the new shared instrumentation and the affected handler logic are covered at or above the scoped targets shown above.

## Remaining observability gaps

- Consumer terminate/recreate during a live campaign is not yet an embedded-JetStream integration scenario; the unit suite covers iterator creation/failure and loop-gauge transitions.
- Broker advisories/exporters, Prometheus recording rules, Grafana dashboards, fault annotations, and final rendered-name verification remain deployment/dashboard work.
- Workstream A owns the loadgen evidence ledger and recipient observers.
- `chat_jetstream_consumer_oldest_pending_age_seconds` remains unowned and is required by the pending-age settle gate.

### DIFF
--- a/broadcast-worker/handler.go
+++ b/broadcast-worker/handler.go
@@ -73,18 +73,34 @@ type Handler struct {
 	metrics       *broadcastMetrics
 }
 
-func NewHandler(store Store, userStore userstore.UserStore, pub Publisher, keyStore RoomKeyProvider, parentFetcher ParentFetcher, encrypt bool, routeMode subject.RoomRouteMode) *Handler {
-	metrics := newBroadcastMetrics(otel.Meter("broadcast-worker"))
+type handlerOption func(*handlerOptions)
+
+type handlerOptions struct {
+	metrics *broadcastMetrics
+}
+
+func withBroadcastMetrics(metrics *broadcastMetrics) handlerOption {
+	return func(opts *handlerOptions) { opts.metrics = metrics }
+}
+
+func NewHandler(store Store, userStore userstore.UserStore, pub Publisher, keyStore RoomKeyProvider, parentFetcher ParentFetcher, encrypt bool, routeMode subject.RoomRouteMode, options ...handlerOption) *Handler {
+	var opts handlerOptions
+	for _, option := range options {
+		option(&opts)
+	}
+	if opts.metrics == nil {
+		opts.metrics = newBroadcastMetrics(otel.Meter("broadcast-worker"))
+	}
 	return &Handler{
 		store:         store,
 		userStore:     userStore,
-		pub:           &broadcastMetricPublisher{next: pub, metrics: metrics},
+		pub:           &broadcastMetricPublisher{next: pub, metrics: opts.metrics},
 		keyStore:      keyStore,
 		parentFetcher: parentFetcher,
 		encrypt:       encrypt,
 		encoder:       roomcrypto.NewEncoder(),
 		routeMode:     routeMode,
-		metrics:       metrics,
+		metrics:       opts.metrics,
 	}
 }
 
--- a/broadcast-worker/handler.go
+++ b/broadcast-worker/handler.go
@@ -98,7 +98,7 @@ func (h *Handler) HandleMessage(ctx context.Context, data []byte) error {
 		return errcode.Permanent(errcode.BadRequest("malformed message event"))
 	}
 	ctx = obs.ContextWithIdentity(ctx, evt.Message.UserAccount, evt.Message.RoomID, evt.SiteID)
-	ctx = withBroadcastMetricLabels(ctx, "unknown", string(evt.Event))
+	ctx = withBroadcastMetricLabels(ctx, roomUnknown, natsmetrics.EventType(evt.Event))
 
 	switch evt.Event {
 	case model.EventCreated:
@@ -700,6 +700,10 @@ func (h *Handler) publishMutation(ctx context.Context, room *model.Room, roomEvt
 
 	switch room.Type {
 	case model.RoomTypeChannel:
+		// Record the intended audience here too: publishChannelEvent does it for
+		// new messages, and a mutation with no fanout sample would leave the
+		// channel lane blank for edit/delete/pin/react during a campaign.
+		h.metrics.Fanout(ctx, roomChannel, labels.eventType, room.UserCount)
 		return h.publishRoomEvent(ctx, room.ID, room.CrossSite, room.CrossSiteAt, payload, fmt.Sprintf("%s event (message %s)", roomEvtType, messageID))
 
 	case model.RoomTypeDM, model.RoomTypeBotDM:
@@ -853,7 +857,7 @@ func (h *Handler) publishChannelEvent(ctx context.Context, meta *roommetacache.M
 	slog.Log(ctx, logctx.LevelFlow, "broadcast fan-out", "phase", "fanout",
 		"request_id", natsutil.RequestIDFromContext(ctx), "room_id", meta.ID,
 		"type", string(meta.Type), "delivery", "room-stream", "audience", meta.UserCount)
-	h.metrics.Fanout(ctx, "channel", broadcastLabels(ctx).eventType, meta.UserCount)
+	h.metrics.Fanout(ctx, roomChannel, broadcastLabels(ctx).eventType, meta.UserCount)
 	return h.publishRoomEvent(ctx, meta.ID, meta.CrossSite, meta.CrossSiteAt, payload, "channel event")
 }
 
@@ -861,7 +865,7 @@ func (h *Handler) publishChannelEvent(ctx context.Context, meta *roommetacache.M
 // sanctioned path enforced by .semgrep room-subject-publish-must-route (never inline a subject).
 func (h *Handler) publishRoomEvent(ctx context.Context, roomID string, crossSite *bool, crossSiteAt *time.Time, payload []byte, op string) error {
 	labels := broadcastLabels(ctx)
-	ctx = withBroadcastMetricLabels(ctx, "channel", labels.eventType)
+	ctx = withBroadcastMetricLabels(ctx, roomChannel, labels.eventType)
 	now := time.Now().UTC()
 	var pubErr error
 	for _, subj := range subject.RoomEventTargets(roomID, crossSite, crossSiteAt, h.routeMode, now) {
@@ -998,8 +1002,8 @@ func buildClientMessage(msg *model.Message, userMap map[string]model.User) *mode
 // accounts that already received the event on the first attempt.
 func (h *Handler) publishToThreadAccounts(ctx context.Context, accounts []string, payload []byte, parentMsgID string) error {
 	labels := broadcastLabels(ctx)
-	ctx = withBroadcastMetricLabels(ctx, "thread", labels.eventType)
-	h.metrics.Fanout(ctx, "thread", labels.eventType, len(accounts))
+	ctx = withBroadcastMetricLabels(ctx, roomThread, labels.eventType)
+	h.metrics.Fanout(ctx, roomThread, labels.eventType, len(accounts))
 	if len(accounts) == 0 {
 		return nil
 	}
@@ -1033,16 +1037,16 @@ func (h *Handler) publishToThreadAccounts(ctx context.Context, accounts []string
 	return nil
 }
 
-func roomKind(roomType model.RoomType) string {
+func roomKind(roomType model.RoomType) roomKindLabel {
 	switch roomType {
 	case model.RoomTypeChannel:
-		return "channel"
+		return roomChannel
 	case model.RoomTypeDM:
-		return "dm"
+		return roomDM
 	case model.RoomTypeBotDM:
-		return "bot_dm"
+		return roomBotDM
 	default:
-		return "unknown"
+		return roomUnknown
 	}
 }
 
--- a/broadcast-worker/handler.go
+++ b/broadcast-worker/handler.go
@@ -608,6 +608,7 @@ func (h *Handler) handleReacted(ctx context.Context, evt *model.MessageEvent) er
 	msg := evt.Message
 	// Log-and-drop on malformed payloads: NAK would loop forever on a publisher contract violation.
 	if evt.ReactionDelta == nil {
+		natsmetrics.MarkTerminalFromContext(ctx, natsmetrics.TerminalInvalidPayload)
 		slog.ErrorContext(ctx, "reacted event missing ReactionDelta; dropping",
 			"messageID", msg.ID,
 			"roomID", msg.RoomID,
@@ -617,6 +618,7 @@ func (h *Handler) handleReacted(ctx context.Context, evt *model.MessageEvent) er
 		return nil
 	}
 	if msg.UpdatedAt == nil {
+		natsmetrics.MarkTerminalFromContext(ctx, natsmetrics.TerminalInvalidPayload)
 		slog.ErrorContext(ctx, "reacted event missing UpdatedAt; dropping",
 			"messageID", msg.ID,
 			"roomID", msg.RoomID,
--- a/broadcast-worker/handler.go
+++ b/broadcast-worker/handler.go
@@ -13,12 +13,14 @@ import (
 	"time"
 
 	"github.com/bytedance/sonic"
+	"go.opentelemetry.io/otel"
 
 	"github.com/hmchangw/chat/pkg/errcode"
 	"github.com/hmchangw/chat/pkg/logctx"
 	"github.com/hmchangw/chat/pkg/mention"
 	"github.com/hmchangw/chat/pkg/model"
 	"github.com/hmchangw/chat/pkg/model/cassandra"
+	"github.com/hmchangw/chat/pkg/natsmetrics"
 	"github.com/hmchangw/chat/pkg/natsutil"
 	"github.com/hmchangw/chat/pkg/obs"
 	"github.com/hmchangw/chat/pkg/roomcrypto"
@@ -68,18 +70,21 @@ type Handler struct {
 	encrypt       bool
 	encoder       *roomcrypto.Encoder
 	routeMode     subject.RoomRouteMode
+	metrics       *broadcastMetrics
 }
 
 func NewHandler(store Store, userStore userstore.UserStore, pub Publisher, keyStore RoomKeyProvider, parentFetcher ParentFetcher, encrypt bool, routeMode subject.RoomRouteMode) *Handler {
+	metrics := newBroadcastMetrics(otel.Meter("broadcast-worker"))
 	return &Handler{
 		store:         store,
 		userStore:     userStore,
-		pub:           pub,
+		pub:           &broadcastMetricPublisher{next: pub, metrics: metrics},
 		keyStore:      keyStore,
 		parentFetcher: parentFetcher,
 		encrypt:       encrypt,
 		encoder:       roomcrypto.NewEncoder(),
 		routeMode:     routeMode,
+		metrics:       metrics,
 	}
 }
 
@@ -87,11 +92,13 @@ func NewHandler(store Store, userStore userstore.UserStore, pub Publisher, keySt
 func (h *Handler) HandleMessage(ctx context.Context, data []byte) error {
 	var evt model.MessageEvent
 	if err := sonic.Unmarshal(data, &evt); err != nil {
+		natsmetrics.MarkTerminalFromContext(ctx, natsmetrics.TerminalInvalidPayload)
 		// Malformed payload — it will never parse on redelivery. Mark permanent
 		// so the caller Acks (drops) it instead of retrying until MaxDeliver.
 		return errcode.Permanent(errcode.BadRequest("malformed message event"))
 	}
 	ctx = obs.ContextWithIdentity(ctx, evt.Message.UserAccount, evt.Message.RoomID, evt.SiteID)
+	ctx = withBroadcastMetricLabels(ctx, "unknown", string(evt.Event))
 
 	switch evt.Event {
 	case model.EventCreated:
@@ -464,6 +471,8 @@ func (h *Handler) handleThreadTCountUpdated(ctx context.Context, evt *model.Mess
 
 func (h *Handler) publishThreadMetadata(ctx context.Context, room *model.Room, newTcount int, newTlm *time.Time,
 	parentMsgID, replyMsgID string, action model.ThreadAction, eventTimestamp int64) error {
+	labels := broadcastLabels(ctx)
+	ctx = withBroadcastMetricLabels(ctx, roomKind(room.Type), labels.eventType)
 	evt := model.ThreadMetadataUpdatedEvent{
 		Type:               model.RoomEventThreadMetadataUpdated,
 		RoomID:             room.ID,
@@ -665,6 +674,7 @@ func (h *Handler) handleReacted(ctx context.Context, evt *model.MessageEvent) er
 			"request_id", natsutil.RequestIDFromContext(ctx),
 		)
 	} else if pubErr := h.pub.Publish(ctx, subject.Notification(authorAccount), data); pubErr != nil {
+		natsmetrics.MarkTerminalFromContext(ctx, natsmetrics.TerminalPublishExhausted)
 		slog.ErrorContext(ctx, "publish reaction author notification failed",
 			"error", pubErr,
 			"author", authorAccount,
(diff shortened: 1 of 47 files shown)


flywindy:
Worked through all eight review findings. Six are fixed as described, one is fixed differently than suggested, and one is declined. Pushed in `8ae3a48`.

**Fixed as described**

| Finding | Change |
|---|---|
| Consumer loop not registered before shutdown waits | `natsmetrics.Start` does `wg.Add(1)` synchronously before launching the loop; all four services use it. The two hand-rolled loops in `message-worker` and `notification-worker` had the same shape and take the same guard. |
| `maxWorkers` not guarded | Clamped to at least 1 in `Consume`. Confirmed both failure modes first: `0` parks the dispatch send forever, `-1` panics with `makechan: size out of range`. |
| Malformed reaction events dropped without terminal evidence | Both nil-guard paths record `TerminalInvalidPayload`; the two existing drop tests now assert it. |
| Instrument creation errors discarded | All four constructors fall back to a no-op instrument. |
| History request metrics untested | Added a manual-reader publisher; both `success` and `no_responders` are asserted. The `errcode` envelope case asserts `success` — a replied-to request is a transport success even when the payload carries an error. |
| `request_id` missing from acknowledgement-failure logs | All five sites now use `ErrorContext` with an explicit `request_id`. One of them also had its error key as `"err"` rather than `"error"`; fixed while there. |

**Fixed differently: reconnect-buffer overflow**

The finding is right that `AsyncError` can never emit `event="buffer_full"` — `nats.go` returns `ErrReconnectBufExceeded` synchronously from `publish()` (`nats.go:4352`, unlock-and-return) and never routes it through `ErrorHandler`. Rather than add a publish-boundary hook into `connMetrics`, the dead value is removed from the connection-events family. `ClassifyPublishError` already maps that error, so every instrumented publish records it as `chat_nats_publish_attempts_total{outcome="buffer_full"}` with richer labels; a second recording path would duplicate an existing series. The metrics contract in #258 asserted the connection-events value was the loss signal, so that document is corrected in the same round.

**Declined: wrapping the publisher error with the subject**

`p.next.Publish` is `natsPublisher.Publish`, which already wraps as `fmt.Errorf("publish to %q: %w", subject, err)`. Adding another layer produces `publish to "X": publish to "X": …`. `broadcastMetricPublisher` is a metrics decorator and should stay transparent to the error it forwards.

**Docstring coverage check**: not adopting. Same position as #237 — the 20% figure counts unexported `package main` internals; exported identifiers carry doc comments per Go convention, and this is not a repository gate.

`make lint` is clean and the full `make test` suite passes.

---

broadcast-worker/nats_metrics.go
```diff
@@ -0,0 +1,160 @@
...
+type broadcastMetricPublisher struct {
+	next    Publisher
+	metrics *broadcastMetrics
+}
+
+func (p *broadcastMetricPublisher) Publish(ctx context.Context, subject string, data []byte) error {
+	err := p.next.Publish(ctx, subject, data)
+	labels := broadcastLabels(ctx)
+	p.metrics.Delivery(ctx, labels.roomKind, labels.eventType, err)
+	return err
```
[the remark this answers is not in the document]
flywindy: `natsPublisher.Publish` already wraps the underlying error with the subject at `broadcast-worker/main.go:310-315`. This metrics decorator intentionally returns that error unchanged; wrapping again here would produce duplicate context (`publish to X: publish to X: ...`). Resolving without a code change.

docs/specs/o11y/o11y-metrics-inventory.md
```diff
@@ -129,6 +131,85 @@ missing beyond shared cache/key counters.
...
+| `chat.nats.consumer.redeliveries` | counter | `pkg/natsmetrics` | stream, consumer, event_type |
+| `chat.nats.consumer.processing.duration` | histogram (s) | `pkg/natsmetrics` | stream, consumer, event_type, outcome |
+| `chat.nats.terminal.failures` | counter | `pkg/natsmetrics` | stream, consumer, event_type, reason |
+| `chat.nats.publish.attempts` | counter | `pkg/natsmetrics` | destination_kind, operation, outcome |
+| `chat.nats.publish.retries` | counter | `pkg/natsmetrics` | destination_kind, operation |
+| `chat.nats.requests` | counter | `pkg/natsmetrics` | operation, outcome |
+| `chat.nats.request.duration` | histogram (s) | `pkg/natsmetrics` | operation, outcome |
+| `chat.nats.client.connected` | up-down counter | `pkg/natsutil` | none — one series per process; value is the live connection count |
+| `chat.nats.client.connection.events` | counter | `pkg/natsutil` | event |
+| `nats_slow_consumer_events_total` | counter | `pkg/natsutil` | subject, queue |
```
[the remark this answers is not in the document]
flywindy: Verified against the locked exporter versions: `go.opentelemetry.io/otel/exporters/prometheus` v0.65.0 uses `github.com/prometheus/otlptranslator` v1.0.0, whose metric namer removes an existing `total` token before appending the Prometheus counter suffix. Its upstream test `metric name already contains total suffix` explicitly asserts `requests_total` -> `requests_total`, so the current instrument does not render as `_total_total`. This PR only inventories the pre-existing slow-consumer instrument; renaming its OTel identity here would be an unrelated compatibility change for non-Prometheus readers. Leaving the existing name unchanged.

broadcast-worker/handler.go
```diff
@@ -969,6 +1019,9 @@ func buildClientMessage(msg *model.Message, userMap map[string]model.User) *mode
 // publish fails — partial failure is tolerated to avoid duplicate delivery to
 // accounts that already received the event on the first attempt.
 func (h *Handler) publishToThreadAccounts(ctx context.Context, accounts []string, payload []byte, parentMsgID string) error {
+	labels := broadcastLabels(ctx)
+	ctx = withBroadcastMetricLabels(ctx, roomThread, labels.eventType)
+	h.metrics.Fanout(ctx, roomThread, labels.eventType, len(accounts))
```
mliu33, line 1024: This metrics should be fired after "wg.wait()"

broadcast-worker/main.go
```diff
@@ -105,6 +106,9 @@ func main() {
 		slog.Error("init observability failed", "error", err)
 		os.Exit(1)
 	}
+	sharedMetrics := natsmetrics.NewFromProvider(sdk.MeterProvider())
+	publishMetrics := sharedMetrics.Publisher("broadcast-worker", cfg.SiteID)
+	domainMetrics := newBroadcastMetrics(sdk.MeterProvider().Meter("broadcast-worker"))
```
mliu33: Why do we need three different metrics instance ?

Publish metrics and domain metrics could be combined ?

message-worker/main.go
```diff
@@ -90,8 +91,11 @@ func main() {
 		slog.Error("init observability failed", "error", err)
 		os.Exit(1)
 	}
+	sharedMetrics := natsmetrics.NewFromProvider(sdk.MeterProvider())
+	publishMetrics := sharedMetrics.Publisher("message-worker", cfg.SiteID)
+	domainMetrics := newPersistenceMetrics(sdk.MeterProvider().Meter("message-worker"))
```
mliu33, line 96: Can we have the name configured from env var? 

We will deploy two separate deployment using the same docker image. One is message-worker while the other is teams-message-worker

notification-worker/main.go
```diff
@@ -140,6 +141,9 @@ func main() {
 		slog.Error("init observability failed", "error", err)
 		os.Exit(1)
 	}
+	sharedMetrics := natsmetrics.NewFromProvider(sdk.MeterProvider())
+	publishMetrics := sharedMetrics.Publisher("notification-worker", cfg.SiteID)
+	domainMetrics := newNotificationMetrics(sdk.MeterProvider().Meter("notification-worker"))
```
mliu33, line 146: Can we make "notification-worker" read from env var? We will deploy two separate services using same codebase. One is "notification-wokrer" while another is "bot-notification-worker".

This apply to all broadcast-worker, message-worker, notification-worker.

notification-worker/nats_metrics.go
```diff
@@ -0,0 +1,77 @@
...
+	"go.opentelemetry.io/otel/metric/noop"
+)
+
+// notifyKind and notifyResult are closed enums so a typo is a compile error
+// rather than a silent `unknown` series at campaign time.
+type notifyKind string
+
+const (
+	notifyKindPush         notifyKind = "push"
+	notifyKindNotification notifyKind = "notification"
```
mliu33, line 17: notifyKindNotification is not used anywhere, right ?

pkg/natsmetrics/metrics.go
```diff
@@ -0,0 +1,575 @@
...
+
+	loop, err := meter.Int64UpDownCounter("chat.nats.consumer.loop.up", metric.WithDescription("Whether an application consumer loop can receive messages."))
+	if err != nil {
+		loop, _ = noopMeter.Int64UpDownCounter("chat.nats.consumer.loop.up")
+	}
+	messages, err := meter.Int64Counter("chat.nats.consumer.messages", metric.WithDescription("JetStream delivery attempts by terminal application disposition."))
+	if err != nil {
+		messages, _ = noopMeter.Int64Counter("chat.nats.consumer.messages")
+	}
+	redeliveries, err := meter.Int64Counter("chat.nats.consumer.redeliveries", metric.WithDescription("JetStream delivery attempts whose delivery count is greater than one."))
```
mliu33: In my opinion, this is redundant. We should use broker-side metrics which NATS server-side already has for these Jetstream related metrics.

pkg/natsmetrics/metrics.go
```diff
@@ -0,0 +1,575 @@
...
+	}
+	publishRetries, err := meter.Int64Counter("chat.nats.publish.retries", metric.WithDescription("Application-managed publish attempts after the first."))
+	if err != nil {
+		publishRetries, _ = noopMeter.Int64Counter("chat.nats.publish.retries")
+	}
+	terminal, err := meter.Int64Counter("chat.nats.terminal.failures", metric.WithDescription("Work that will receive no further application attempt."))
+	if err != nil {
+		terminal, _ = noopMeter.Int64Counter("chat.nats.terminal.failures")
+	}
+	requests, err := meter.Int64Counter("chat.nats.requests", metric.WithDescription("NATS request/reply results."))
```
mliu33: Is this metrics going to be used for calculating RPS? If yes, can we use like "chat.nats.requests.total"

pkg/natsmetrics/metrics.go
mliu33, line 172: In my option, we should use broker-side metrics to monitor redelivery which nats server already has this metrics. I am not sure if we really need this

pkg/natsmetrics/metrics.go
```diff
@@ -0,0 +1,575 @@
...
+	c.Terminal(ctx, EventUnknown, reason)
+}
+
+func (c *Consumer) Track(ctx context.Context, msg jetstream.Msg, eventType EventType, maxDeliver int) *Message {
+	eventType = NormalizeEventType(string(eventType))
+	tracked := &Message{Msg: msg, consumer: c, ctx: ctx, eventType: eventType, maxDeliver: maxDeliver, started: time.Now()}
+	if meta, err := msg.Metadata(); err == nil && meta != nil {
+		tracked.numDelivered = meta.NumDelivered
+		if meta.NumDelivered > 1 {
+			c.metrics.redeliveries.Add(ctx, 1, c.redeliv[eventType])
```
mliu33, line 322: Like above comment, every metrics and tracking also consume a bit of resource. If we can already monitor redelivery metrics using nats server-side metrics, I don't think we need to track redelivery on client side. Also wonder what's the cost to check delivery count on every message here?

pkg/natsmetrics/metrics.go
mliu33, line 192: Is this going to be used for RPS? If yes, can we use similar to standard http naming like "chat.nats.requests.total" ?

notification-worker/main.go
```diff
@@ -303,20 +314,31 @@ func main() {
...
 			}
 			sem <- struct{}{}
 			wg.Add(1)
-			go func() {
+			go func(msgCtx context.Context, msg jetstream.Msg) {
+				tracked := consumerMetrics.Track(msgCtx, msg, natsmetrics.EventTypeFromSubject(msg.Subject()), consumerCfg.MaxDeliver)
+				msg = tracked
+				msgCtx = tracked.Context(msgCtx)
 				defer func() {
+					tracked.Finish(msgCtx)
```
mliu33, line 341: The message Finish method is using OutcomeLeftPending ?

pkg/natsmetrics/metrics.go
```diff
@@ -0,0 +1,575 @@
...
+	disposeOnce  sync.Once
+	terminalOnce sync.Once
+}
+
+type messageContextKey struct{}
+
+// Context returns ctx carrying this delivery recorder so deeper fan-out code
+// can report a swallowed terminal failure without accepting dynamic labels.
+func (m *Message) Context(ctx context.Context) context.Context {
+	return context.WithValue(ctx, messageContextKey{}, m)
```
mliu33, line 352: I do have a bit concern about the increased memory consumption if we carry entire object in every context. If it's not absolutely necessary or important metrics, we should keep it simple and lightweight

pkg/natsmetrics/metrics.go
```diff
@@ -0,0 +1,575 @@
...
+	// advisory. The o11y SDK only registers bucket views for its own instrument
+	// names, so without this these two histograms fall back to the OTel default
+	// boundaries (0..10000) and every sub-second duration lands in bucket one.
+	latency := metric.WithExplicitBucketBoundaries(o11y.DefaultLatencyBuckets()...)
+
+	loop, err := meter.Int64UpDownCounter("chat.nats.consumer.loop.up", metric.WithDescription("Whether an application consumer loop can receive messages."))
+	if err != nil {
+		loop, _ = noopMeter.Int64UpDownCounter("chat.nats.consumer.loop.up")
+	}
+	messages, err := meter.Int64Counter("chat.nats.consumer.messages", metric.WithDescription("JetStream delivery attempts by terminal application disposition."))
```
mliu33, line 168: This is the total messages that consumer has processed, right ?

If yes, can we follow standard naming to be like "chat.nats.consumer.messages.total" ?

mliu33, line 168: This is the total messages that consumer has processed, right ?

If yes, can we follow standard naming to be like "chat.nats.consumer.messages.total" ?

mliu33 (approved):
Outstanding work! Just some comments, thanks a lot!!

broadcast-worker/handler.go
flywindy: Verified and kept as designed. Fanout records the intended recipient count, not publish completion, and the OTel histogram Record call is synchronous. Per-publish outcomes are recorded separately, while wg.Wait gates handler completion. Moving the fanout observation after the wait would change only its timestamp.

message-worker/main.go
flywindy: Addressed in #286. Shared publisher and consumer service_name labels now read OTEL_SERVICE_NAME. The fixed Meter name remains intentionally stable as instrumentation scope, while deployment identity comes from the OTel resource and shared metric label.

notification-worker/main.go
flywindy: Addressed across broadcast-worker, message-worker, and notification-worker in #286. Their shared NATS service_name labels now follow OTEL_SERVICE_NAME, while stable instrumentation-scope names remain unchanged.

notification-worker/nats_metrics.go
flywindy: Addressed in #286. The unused notification metric kind and its pre-warmed attribute set were removed.

pkg/natsmetrics/metrics.go
flywindy: Kept intentionally. Workstream B requires application-side redelivery evidence from NumDelivered greater than 1. Broker metrics remain the aggregate source, while this metric adds service, site, event type, and disposition correlation. Metadata is also required for final-delivery and max-deliver terminal evidence, so removing only this counter would not remove metadata parsing. #286 adds a focused benchmark target.

pkg/natsmetrics/metrics.go
flywindy: Benchmarked in #286 and kept. The repository-owned tracking path measures 38.82 ns and 1 allocation for first delivery, and 59.47 ns and 2 allocations for redelivery on the recorded Windows amd64 system. The benchmark uses pre-decoded metadata, so nats.go reply-subject parsing is explicitly outside the result.

pkg/natsmetrics/metrics.go
flywindy: No instrument rename is needed. chat.nats.requests is the OpenTelemetry instrument name; the Prometheus exporter renders it as chat_nats_requests_total. Adding .total to the OTel name would break the documented metric contract.

notification-worker/main.go
flywindy: Verified as intentional. Finish is the deferred fallback. Ack, Nak, and Term record their disposition first, and sync.Once prevents double counting. left_pending is emitted only when no disposition succeeds; stopped-loop or cancelled-context paths emit handler_cancelled instead.

pkg/natsmetrics/metrics.go
flywindy: Verified and benchmarked in #286. Context carries a pointer to the tracked Message rather than copying the message payload, and it lets deep handlers mark terminal invalid or publish-exhausted evidence. The measured wrapper cost is 25.01 ns, 48 B, and 1 allocation per Context call on the recorded system.

pkg/natsmetrics/metrics.go
flywindy: No instrument rename is needed. chat.nats.consumer.messages is the OpenTelemetry instrument name; the Prometheus exporter renders it as chat_nats_consumer_messages_total. Adding .total to the OTel name would violate the documented contract.

flywindy: Duplicate of the preceding naming comment. The OpenTelemetry name intentionally omits .total because the Prometheus exporter adds the _total suffix.

PR #240, opened 2026-08-12
Add CDC migration verification tool with live dashboard

general-lex:
## Summary

Introduces `tools/cdc-verify`, a live verification dashboard for the CDC migration pipeline. The tool passively observes the `MIGRATION-OPLOG-{siteID}` stream, re-reads source and destination state for each changed document, compares them field-by-field against a JSON-configurable mapping, and streams results to a browser dashboard. It is strictly read-only, uses an ephemeral JetStream consumer, and is deletable once the source is retired.

Design: `docs/superpowers/specs/2026-08-10-cdc-verify-tool-design.md` · Plan: `docs/superpowers/plans/2026-08-10-cdc-verify-tool.md`

## Key Changes

- **Core verifier engine** (`verifier.go`): Implements the check lifecycle — submits CDC events, polls source/destination stores with configurable timeout and retry, compares field values using normalized types, and tracks results in ring buffers. Supports field transformations (e.g., `unixMilli`, `msgBucket`) and resolver dependencies for cross-document lookups. Checks **freeze on match** (later updates start new checks); a newer event on the same key supersedes a pending check.

- **Mapping & configuration** (`mapping.go`, `mapping_validate.go`): Defines the JSON schema for source→destination field fan-out (one source field → many destination tables), operation actions (`verify`, `verify-absent`, `skip`), one-hop resolvers (`@user.username`), verbatim mode for direct-transfer collections, and per-target Cassandra/MongoDB key specs. Validates mapping structure fail-fast at startup; `mapping.example.json` ships schema-aligned to the real Cassandra tables.

- **Store adapters** (`lookup_mongo.go`, `lookup_cassandra.go`, `store.go`): Read-only lookups against MongoDB and Cassandra with point-select queries, projection support, ambiguity detection, and un-MapScannable CQL column handling (struct-keyed maps). Consumer-defined interfaces for mocking.

- **Field comparison** (`compare.go`): Normalizes driver scalar types (ints→float64, timestamps, bytes) for safe cross-store comparison, walks dot-paths through nested documents, implements the spec's absent-source/zero-value-dest matching rule, and reports field-level diffs.

- **Results storage** (`results.go`): Capped ring buffers for recent checks and accumulated failures, deep-copy isolation, once-per-check counters with symmetric eviction pruning.

[...]

PR #273, opened 2026-08-13
cdc-verify: review-wave hardening follow-up to #240

general-lex:
## Summary

Follow-up to #240 (merged), addressing the real findings from its 58 review threads. Fix-to-thread mapping and the defend/deferred rationale are recorded on #240 ([summary comment](https://github.com/hmchangw/newchat/pull/240#issuecomment-5284693139) + per-thread replies).

## Key Changes

- **Verify deadline is enforced end-to-end** (`verifier.go`): `VERIFY_TIMEOUT` now becomes a context deadline at Submit time — it bounds the semaphore wait (a check queued behind `MAX_CHECKS` expires as `failed` instead of parking forever) and every source/dest/resolver store call. Skipped newer events (`op-skip`, `sampled-out`) supersede a pending check on the same key; a supersede that races the terminal transition wins; `Inspect` shares `lookupDest`'s nil-Cassandra guard (removes a panic path); `Shutdown` reports drain-timeout as an error.
- **Subject-authoritative, failure-visible watcher** (`watcher.go`, `main.go`): the ordered consumer filters on `chat.migration.oplog.{site}.>`; routing collection/op comes from the subject tokens (payload disagreement logged, subject wins); undecodable payloads surface as skipped `decode-error` rows; `Run` observes `ConsumeContext.Closed()`; a watcher failure SIGTERMs the process through `pkg/shutdown` instead of `os.Exit(1)`.
- **Comparison semantics + JSON-safe diffs** (`compare.go`): explicit BSON `null` in the source counts as absent (spec §5.3); all diff values pass a sanitizer — non-string map keys stringified, `[]byte`→string, 2 KB truncation — so result marshaling can never fail and failure retention is byte-bounded per value.
- **Safer store access** (`bootstrap.go`, `lookup_cassandra.go`): stream bootstrap is create-only (an existing stream's config is never rewritten); Cassandra point-selects carry `LIMIT 2` (one row proves presence, a second proves ambiguity).
- **Stricter config** (`main.go`, `conninfo.go`): `STATS_INTERVAL` validated (ticker panic); new `SOURCE_READ_PREFERENCE` env, default `primary` (a stale secondary would report convergence lag that isn't there), surfaced in the connections panel.
- **Resilient dashboard** (`hub.go`, `handler.go`, `static/index.html`): the SSE hub disconnects a slow client instead of silently dropping frames; the UI refetches `/api/state` on every SSE reconnect; the failures table renders incrementally with a client-side cap; `/api/inspect` errors return a generic 500 body (detail server-log only).
- **Docs**: README run example includes the Cassandra env, fences get languages, `SOURCE_READ_PREFERENCE` documented. Plus a one-line goimports fix in `search-service/integration_index_test.go` from repo-wide `make fmt`.

## Testing

- TDD throughout — each fix landed red→green; `make test SERVICE=tools/cdc-verify` race-clean; `golangci-lint` 0 issues (pre-commit hook).
- Integration and demoharness builds vet-clean; integration suite still needs a Docker-capable machine (same pre-merge caveat as #240).

## Notes

- No `chat.user.*` handler surface → no `docs/client-api.md` changes.
- Deferred follow-ups agreed in the #240 threads: `system_schema`-based Cassandra key/type validation, `omittedFields` metadata on whole-row inspect, lossless int64 comparison if a future mapping needs it.

https://claude.ai/code/session_01RBUJtNXnwoeYSVWwUWrk74

---
_Generated by [Claude Code](https://claude.ai/code/session_01RBUJtNXnwoeYSVWwUWrk74)_

## Summary by CodeRabbit

- **New Features**
  - Added configurable MongoDB source read preferences via `SOURCE_READ_PREFERENCE`, defaulting to `primary`.
  - Added subject-filtered event processing for more accurate collection and operation routing.
  - Added clearer handling and reporting for undecodable events.

- **Bug Fixes**
  - Prevented sensitive error details from appearing in API responses.
  - Improved timeout, shutdown, and pending-check handling.
  - Limited Cassandra lookups to the rows needed for result validation.
  - Automatically disconnects slow event-stream clients.

- **User Interface**
  - Improved failure-list performance and capped stored failures at 1,000.
  - Reloads state after stream reconnections.

### DIFF
--- a/tools/cdc-verify/bootstrap.go
+++ b/tools/cdc-verify/bootstrap.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/nats-io/nats.go/jetstream"
@@ -17,12 +18,21 @@ type bootstrapConfig struct {
 
 // bootstrapStreams creates MIGRATION-OPLOG-{siteID} from schema (Name+Subjects) for
 // dev stacks without oplog-connector; disabled it no-ops and js.Stream fails fast.
+// Create-only: an existing stream is owned by the real pipeline and never updated —
+// a bare Name+Subjects update would strip its retention/replica config.
 func bootstrapStreams(ctx context.Context, js jetstream.JetStream, siteID string, enabled bool) error {
 	if !enabled {
 		return nil
 	}
 	cfg := stream.MigrationOplog(siteID)
-	if _, err := js.CreateOrUpdateStream(ctx, jetstream.StreamConfig{
+	_, err := js.Stream(ctx, cfg.Name)
+	if err == nil {
+		return nil
+	}
+	if !errors.Is(err, jetstream.ErrStreamNotFound) {
+		return fmt.Errorf("check MIGRATION-OPLOG stream: %w", err)
+	}
+	if _, err := js.CreateStream(ctx, jetstream.StreamConfig{
 		Name:     cfg.Name,
 		Subjects: cfg.Subjects,
 	}); err != nil {
--- a/tools/cdc-verify/compare.go
+++ b/tools/cdc-verify/compare.go
@@ -5,6 +5,7 @@ import (
 	"reflect"
 	"strings"
 	"time"
+	"unicode/utf8"
 )
 
 // getPath walks a dot-path through nested map[string]any. No array indexing.
@@ -91,6 +92,65 @@ func isZeroValue(v any) bool {
 	}
 }
 
+// diffValueMaxString caps one diff value's string length: a diff is a pointer
+// at a mismatch, not a document dump, and rows are retained + broadcast as JSON.
+const diffValueMaxString = 2048
+
+// sanitizeDiffValue makes a value safe to retain and JSON-encode in a diff:
+// non-string map keys are stringified (json.Marshal rejects e.g. map[bool]T
+// from a driver), []byte becomes string, and huge strings are truncated.
+func sanitizeDiffValue(v any) any {
+	switch t := v.(type) {
+	case nil:
+		return nil
+	case string:
+		return truncDiffString(t)
+	case []byte:
+		return truncDiffString(string(t))
+	case map[string]any:
+		out := make(map[string]any, len(t))
+		for k, vv := range t {
+			out[truncDiffString(k)] = sanitizeDiffValue(vv)
+		}
+		return out
+	case []any:
+		out := make([]any, len(t))
+		for i, vv := range t {
+			out[i] = sanitizeDiffValue(vv)
+		}
+		return out
+	}
+	rv := reflect.ValueOf(v)
+	switch rv.Kind() {
+	case reflect.Map:
+		out := make(map[string]any, rv.Len())
+		iter := rv.MapRange()
+		for iter.Next() {
+			out[truncDiffString(fmt.Sprint(iter.Key().Interface()))] = sanitizeDiffValue(iter.Value().Interface())
+		}
+		return out
+	case reflect.Slice, reflect.Array:
+		out := make([]any, rv.Len())
+		for i := range out {
+			out[i] = sanitizeDiffValue(rv.Index(i).Interface())
+		}
+		return out
+	default:
+		return v
+	}
+}
+
+func truncDiffString(s string) string {
+	if len(s) <= diffValueMaxString {
+		return s
+	}
+	cut := diffValueMaxString
+	for cut > 0 && !utf8.RuneStart(s[cut]) {
+		cut--
+	}
+	return s[:cut] + "…(truncated)"
+}
+
 // FieldDiff is one mismatched field in a failed sub-check.
 type FieldDiff struct {
 	SourcePath string `json:"sourcePath"`
@@ -116,7 +176,9 @@ func diffFields(src, dst map[string]any, pairs []fieldPair, reg transformRegistr
 		anyPresent := false
 		for _, sp := range p.SourcePaths {
 			v, ok := getPath(src, sp)
-			if ok {
+			// An explicit BSON null counts as absent: transformers drop null
+			// fields, so the dest legitimately holds nothing for them.
+			if ok && v != nil {
 				anyPresent = true
 			}
 			args = append(args, v)
@@ -130,7 +192,7 @@ func diffFields(src, dst map[string]any, pairs []fieldPair, reg transformRegistr
 			switch {
 			case !destZero:
 				diffs = append(diffs, FieldDiff{SourcePath: strings.Join(p.SourcePaths, ","),
-					DestField: p.DestField, Want: nil, Got: got, Cause: "absent in source, present in dest"})
+					DestField: p.DestField, Want: nil, Got: sanitizeDiffValue(got), Cause: "absent in source, present in dest"})
 			case p.Required:
 				diffs = append(diffs, FieldDiff{SourcePath: strings.Join(p.SourcePaths, ","),
 					DestField: p.DestField, Cause: "required field absent on both sides"})
@@ -146,7 +208,7 @@ func diffFields(src, dst map[string]any, pairs []fieldPair, reg transformRegistr
 		}
 		if !valuesEqual(want, got) {
 			diffs = append(diffs, FieldDiff{SourcePath: strings.Join(p.SourcePaths, ","),
-				DestField: p.DestField, Want: want, Got: got})
+				DestField: p.DestField, Want: sanitizeDiffValue(want), Got: sanitizeDiffValue(got)})
 		}
 	}
 	return diffs
@@ -165,7 +227,8 @@ func diffVerbatim(src, dst map[string]any, ignore []string) []FieldDiff {
 		}
 		got, ok := getPath(dst, k)
 		if !ok || !valuesEqual(want, got) {
-			diffs = append(diffs, FieldDiff{SourcePath: k, DestField: k, Want: want, Got: got})
+			diffs = append(diffs, FieldDiff{SourcePath: k, DestField: k,
+				Want: sanitizeDiffValue(want), Got: sanitizeDiffValue(got)})
 		}
 	}
 	for k, got := range dst {
@@ -173,7 +236,7 @@ func diffVerbatim(src, dst map[string]any, ignore []string) []FieldDiff {
 			continue
 		}
 		if _, ok := src[k]; !ok {
-			diffs = append(diffs, FieldDiff{DestField: k, Got: got, Cause: "present only in dest"})
+			diffs = append(diffs, FieldDiff{DestField: k, Got: sanitizeDiffValue(got), Cause: "present only in dest"})
 		}
 	}
 	return diffs
--- a/tools/cdc-verify/handler.go
+++ b/tools/cdc-verify/handler.go
@@ -65,7 +65,10 @@ func (h *handler) inspect(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusNotFound)
 		return
 	case err != nil:
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+		// Generic body only — a store error can carry URIs/hosts the browser
+		// must never see; the detail goes to the server log.
+		slog.Error("inspect failed", "collection", collection, "error", err)
+		http.Error(w, "inspect failed; see server log", http.StatusInternalServerError)
 		return
 	}
 	writeJSON(w, res)
@@ -117,7 +120,10 @@ func (h *handler) events(w http.ResponseWriter, r *http.Request) {
 		case <-keepalive.C:
 			fmt.Fprint(w, ": ping\n\n")
 			flusher.Flush()
-		case frame := <-ch:
+		case frame, ok := <-ch:
+			if !ok {
+				return // hub disconnected us as a slow client; the browser reconnects
+			}
 			fmt.Fprintf(w, "data: %s\n\n", frame)
 			flusher.Flush()
 		}
--- a/tools/cdc-verify/main.go
+++ b/tools/cdc-verify/main.go
@@ -241,7 +241,7 @@ func main() {
 
 	shutdown.Wait(context.Background(), 25*time.Second,
 		func(sctx context.Context) error { return srv.Shutdown(sctx) },
-		func(sctx context.Context) error { cancel(); v.Shutdown(sctx); return nil },
+		func(sctx context.Context) error { cancel(); return v.Shutdown(sctx) },
 		func(_ context.Context) error { return nc.Drain() },
 		func(sctx context.Context) error {
 			mongoutil.Disconnect(sctx, srcClient)
--- a/tools/cdc-verify/main.go
+++ b/tools/cdc-verify/main.go
@@ -13,7 +13,6 @@ import (
 	"github.com/gocql/gocql"
 	"github.com/nats-io/nats.go"
 	"github.com/nats-io/nats.go/jetstream"
-	"go.mongodb.org/mongo-driver/v2/mongo/readpref"
 
 	"github.com/hmchangw/chat/pkg/cassutil"
 	"github.com/hmchangw/chat/pkg/mongoutil"
@@ -32,6 +31,9 @@ type config struct {
 	SourceMongoUsername string `env:"SOURCE_MONGO_USERNAME" envDefault:""`
 	SourceMongoPassword string `env:"SOURCE_MONGO_PASSWORD" envDefault:""`
 	SourceDB            string `env:"SOURCE_DB" envDefault:"rocketchat"`
+	// primary by default: a verifier polling a stale secondary would report
+	// convergence lag that isn't there. Relax deliberately per deployment.
+	SourceReadPreference string `env:"SOURCE_READ_PREFERENCE" envDefault:"primary"`
 
 	TargetMongoURI      string `env:"TARGET_MONGO_URI,required"`
 	TargetMongoUsername string `env:"TARGET_MONGO_USERNAME" envDefault:""`
@@ -77,6 +79,12 @@ func (c *config) validate() error {
 	}
 	if c.MessageBucketHours <= 0 {
 		return fmt.Errorf("MESSAGE_BUCKET_HOURS must be positive, got %d", c.MessageBucketHours)
+	}
+	if c.StatsInterval <= 0 {
+		return fmt.Errorf("STATS_INTERVAL must be positive, got %s", c.StatsInterval)
+	}
+	if _, err := mongoutil.ParseReadPreference(c.SourceReadPreference); err != nil {
+		return fmt.Errorf("SOURCE_READ_PREFERENCE: %w", err)
 	}
 	if c.StartAtTime != "" {
 		if _, err := time.Parse(time.RFC3339, c.StartAtTime); err != nil {
@@ -150,8 +158,13 @@ func main() {
 		os.Exit(1)
 	}
 
+	srcReadPref, err := mongoutil.ParseReadPreference(cfg.SourceReadPreference) // validated already
+	if err != nil {
+		slog.Error("parse source read preference", "error", err)
+		os.Exit(1)
+	}
 	srcClient, err := mongoutil.Connect(ctx, cfg.SourceMongoURI, cfg.SourceMongoUsername, cfg.SourceMongoPassword,
-		mongoutil.WithReadPreference(readpref.PrimaryPreferred()))
+		mongoutil.WithReadPreference(srcReadPref))
 	if err != nil {
 		slog.Error("connect source mongo", "error", err)
 		os.Exit(1)
--- a/tools/cdc-verify/main.go
+++ b/tools/cdc-verify/main.go
@@ -6,6 +6,7 @@ import (
 	"log/slog"
 	"net/http"
 	"os"
+	"syscall"
 	"time"
 
 	"github.com/caarlos0/env/v11"
@@ -194,15 +195,18 @@ func main() {
 	if cfg.StartAtTime != "" {
 		startAt, _ = time.Parse(time.RFC3339, cfg.StartAtTime) // validated already
 	}
-	w := newWatcher(js, streamName, startAt, v)
+	filter := subject.MigrationOplogWildcard(cfg.SiteID)
+	w := newWatcher(js, streamName, filter, startAt, v)
 	go func() {
 		if err := w.Run(ctx); err != nil {
-			slog.Error("watcher stopped", "error", err)
-			os.Exit(1) // a dead feed makes the dashboard lie; die loudly
+			// A dead feed makes the dashboard lie; die loudly — but through the
+			// signal path so shutdown.Wait still drains connections cleanly.
+			slog.Error("watcher stopped; shutting down", "error", err)
+			if p, perr := os.FindProcess(os.Getpid()); perr == nil {
+				_ = p.Signal(syscall.SIGTERM)
+			}
 		}
 	}()
-
-	filter := subject.MigrationOplogWildcard(cfg.SiteID)
 	poller := newStatsPoller(streamName,
 		func(ctx context.Context) (*jetstream.StreamInfo, error) {
 			return s.Info(ctx, jetstream.WithSubjectFilter(filter))
--- a/tools/cdc-verify/static/index.html
+++ b/tools/cdc-verify/static/index.html
@@ -386,6 +386,7 @@
 
   var TERMINAL = { matched: true, failed: true, skipped: true, superseded: true };
   var FEED_CAP = 100;
+  var FAILURES_CAP = 1000; // client-side retention; matches the server's FAILED_CAP default
   var PAIR_CAP = 50;
 
   // pairKey(source, alias) -> {pair, rows: Map<id, r>, order: [ids newest first]}
@@ -779,6 +780,8 @@
     return tr;
   }
 
+  // Full rebuild — initial state application only; SSE updates go through
+  // upsertFailureRow so one event never re-renders the whole table.
   function renderFailures() {
     var tbody = document.querySelector('#failures tbody');
     tbody.replaceChildren();
@@ -791,6 +794,16 @@
       if (!r) return;
       tbody.appendChild(buildFailureRow(r));
     });
+  }
+
+  function upsertFailureRow(r, isNew) {
+    var tbody = document.querySelector('#failures tbody');
+    var existing = tbody.querySelector('tr[data-id="' + r.id + '"]');
+    if (existing) { existing.replaceWith(buildFailureRow(r)); return; }
+    if (!isNew) return; // update to a row already evicted from the table
+    if (failuresOrder.length === 1) tbody.replaceChildren(); // drop the 'No failures.' placeholder
+    tbody.prepend(buildFailureRow(r));
+    while (tbody.children.length > FAILURES_CAP) tbody.removeChild(tbody.lastChild);
   }
 
   // ── state application ──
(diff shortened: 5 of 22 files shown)


julianshen:
**Actionable comments posted: 16** — 🔴 1 blocking · 🟠 15 warning

<details>

* `search-service/integration_index_test.go`
* `tools/cdc-verify/README.md`
* `tools/cdc-verify/bootstrap.go`
* `tools/cdc-verify/bootstrap_test.go`
* `tools/cdc-verify/compare.go`
* `tools/cdc-verify/compare_test.go`
* `tools/cdc-verify/conninfo.go`
* `tools/cdc-verify/conninfo_test.go`
* `tools/cdc-verify/handler.go`
* `tools/cdc-verify/handler_test.go`
* `tools/cdc-verify/hub.go`
* `tools/cdc-verify/hub_test.go`
* `tools/cdc-verify/integration_test.go`
* `tools/cdc-verify/lookup_cassandra.go`
* `tools/cdc-verify/lookup_mongo_test.go`
* `tools/cdc-verify/main.go`
* `tools/cdc-verify/main_test.go`
* `tools/cdc-verify/static/index.html`
* `tools/cdc-verify/verifier.go`
* `tools/cdc-verify/verifier_test.go`
* `tools/cdc-verify/watcher.go`
* `tools/cdc-verify/watcher_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `silent-failure`
* `tgd-review`

</details>

tools/cdc-verify/main.go
```diff
@@ -194,15 +208,18 @@ func main() {
...
+	w := newWatcher(js, streamName, filter, startAt, v)
 	go func() {
 		if err := w.Run(ctx); err != nil {
-			slog.Error("watcher stopped", "error", err)
-			os.Exit(1) // a dead feed makes the dashboard lie; die loudly
+			// A dead feed makes the dashboard lie; die loudly — but through the
+			// signal path so shutdown.Wait still drains connections cleanly.
+			slog.Error("watcher stopped; shutting down", "error", err)
+			if p, perr := os.FindProcess(os.Getpid()); perr == nil {
+				_ = p.Signal(syscall.SIGTERM)
```
julianshen, line 219: _🏷️ error-handling_ | _🔴 Blocking_ | _`silent-failure`_

**Watcher failures are converted into a clean process exit.**

When w.Run fails, the error is only logged and SIGTERM is sent through the normal graceful-shutdown path. The watcher error is then discarded, so main can return with status 0; supervisors configured to restart only failed processes may leave the verifier stopped. The ignored Signal error can also leave the process running with a dead feed. Send the watcher error to main, initiate shutdown, and exit nonzero after shutdown completes; also surface any failure to initiate shutdown.

<details>

```
In `tools/cdc-verify/main.go` around line 219: When w.Run fails, the error is only logged and SIGTERM is sent through the normal graceful-shutdown path. The watcher error is then discarded, so main can return with status 0; supervisors configured to restart only failed processes may leave the verifier stopped. The ignored Signal error can also leave the process running with a dead feed. Send the watcher error to main, initiate shutdown, and exit nonzero after shutdown completes; also surface any failure to initiate shutdown.

```

</details>

tools/cdc-verify/bootstrap.go
```diff
@@ -17,12 +18,21 @@ type bootstrapConfig struct {
...
 	cfg := stream.MigrationOplog(siteID)
-	if _, err := js.CreateOrUpdateStream(ctx, jetstream.StreamConfig{
+	_, err := js.Stream(ctx, cfg.Name)
+	if err == nil {
+		return nil
+	}
+	if !errors.Is(err, jetstream.ErrStreamNotFound) {
+		return fmt.Errorf("check MIGRATION-OPLOG stream: %w", err)
+	}
+	if _, err := js.CreateStream(ctx, jetstream.StreamConfig{
```
julianshen, line 35: _🎯 distributed-correctness_ | _🟠 Warning_ | _`distributed-system`_

**Create-only bootstrap is not idempotent after an ambiguous create.**

The stream lookup and creation are separate operations. If CreateStream succeeds on NATS but its response is lost or times out, this function returns an error and startup fails although the desired stream now exists. The same failure occurs when another bootstrap or the real pipeline creates the stream between the lookup and create. After a create error, re-query the stream with a bounded usable context and treat an existing stream as success; only return the create failure when the stream is still absent.

<details>

```
In `tools/cdc-verify/bootstrap.go` around line 35: The stream lookup and creation are separate operations. If CreateStream succeeds on NATS but its response is lost or times out, this function returns an error and startup fails although the desired stream now exists. The same failure occurs when another bootstrap or the real pipeline creates the stream between the lookup and create. After a create error, re-query the stream with a bounded usable context and treat an existing stream as success; only return the create failure when the stream is still absent.

```

</details>

tools/cdc-verify/static/index.html
```diff
@@ -1113,16 +1129,26 @@ <h2>Failures <a href="/failures.json" download>Download JSON</a></h2>
...
+  }
+  loadState();
 
   var es = new EventSource('/api/events');
+  var esFirstOpen = true;
+  // A reconnect (network blip, or the hub dropped us as a slow client) means
+  // frames were missed — refetch full state instead of trusting the tail.
+  es.onopen = function () {
+    if (esFirstOpen) { esFirstOpen = false; return; }
+    loadState();
```
julianshen, line 1150: _🏷️ ordering_ | _🟠 Warning_ | _`distributed-system`_

**State reloads can overwrite SSE events received during the fetch.**

On reconnect, loadState starts asynchronously while the reopened EventSource already delivers frames. If /api/state snapshots state S, a newer event E is received and applied over SSE, and the slower snapshot response then calls applyInitialState(S), E is removed from the client state and will not be retransmitted. The dashboard remains stale until another reconnect. Buffer SSE frames while a state fetch is in flight and apply them after the snapshot, or attach a monotonic state revision and reject stale snapshots; also serialize overlapping reloads.

<details>

```
In `tools/cdc-verify/static/index.html` around line 1150: On reconnect, loadState starts asynchronously while the reopened EventSource already delivers frames. If /api/state snapshots state S, a newer event E is received and applied over SSE, and the slower snapshot response then calls applyInitialState(S), E is removed from the client state and will not be retransmitted. The dashboard remains stale until another reconnect. Buffer SSE frames while a state fetch is in flight and apply them after the snapshot, or attach a monotonic state revision and reject stale snapshots; also serialize overlapping reloads.

```

</details>

tools/cdc-verify/verifier.go
```diff
@@ -600,26 +629,36 @@ func snapshotTargets(states []targetState, withDiffs bool) []TargetResult {
 	return out
 }
 
-func (v *verifier) finish(row *CheckResult, state CheckState, states []targetState) {
+// finish publishes the terminal state; a supersede that raced the terminal
+// transition wins, so the row never shows matched/failed for a stale target.
+func (v *verifier) finish(h *checkHandle, row *CheckResult, state CheckState, states []targetState) {
+	if h.superseded.Load() {
```
julianshen, line 635: _🏷️ ordering_ | _🟠 Warning_ | _`distributed-system`_

**Terminal checks can publish after a newer event supersedes them.**

The atomic Load does not serialize the terminal transition with pending-check replacement. A newer event can install its handle via beginCheck, then the old check can read superseded=false and publish MATCHED before the newer Submit stores true on the displaced handle. Cancellation afterward cannot revise that already-terminal row, so a stale check remains matched despite the newer event. Serialize pending ownership replacement, superseding, and terminal classification under the same mutex; terminal completion should atomically remove its handle if still current, otherwise publish SUPERSEDED.

<details>

```
In `tools/cdc-verify/verifier.go` around line 635: The atomic Load does not serialize the terminal transition with pending-check replacement. A newer event can install its handle via beginCheck, then the old check can read superseded=false and publish MATCHED before the newer Submit stores true on the displaced handle. Cancellation afterward cannot revise that already-terminal row, so a stale check remains matched despite the newer event. Serialize pending ownership replacement, superseding, and terminal classification under the same mutex; terminal completion should atomically remove its handle if still current, otherwise publish SUPERSEDED.

```

</details>

tools/cdc-verify/verifier.go
```diff
@@ -277,17 +282,42 @@ func (v *verifier) Submit(ev CDCEvent) {
...
 	go func() {
 		defer v.wg.Done()
-		v.runCheck(ctx, h, key, &row, cs, action)
+		v.runCheck(ctx, h, key, &row, cs, action, deadline)
 	}()
 }
 
+// SubmitUndecodable records a skipped row for an event whose payload failed to
+// decode; collection/op come from the subject, so the row is still routable.
+func (v *verifier) SubmitUndecodable(collection, op string) {
```
julianshen, line 291: _🏷️ delivery-ordering_ | _🟠 Warning_ | _`distributed-system`_

**Undecodable events leave older checks active for an unknown document.**

SubmitUndecodable records the collection and operation but cannot supersede a pending check because it discards the document identity. For example, E1 for document d is pending, then a newer malformed E2 for d arrives; E1 remains active and can eventually compare the live post-E2 source and destination, become MATCHED, and freeze, even though its convergence target was invalidated by E2. Recover documentKey independently when possible and supersede that key. If no identity can be recovered, conservatively invalidate pending checks for the affected collection rather than allowing them to publish MATCHED.

<details>

```
In `tools/cdc-verify/verifier.go` around line 291: SubmitUndecodable records the collection and operation but cannot supersede a pending check because it discards the document identity. For example, E1 for document d is pending, then a newer malformed E2 for d arrives; E1 remains active and can eventually compare the live post-E2 source and destination, become MATCHED, and freeze, even though its convergence target was invalidated by E2. Recover documentKey independently when possible and supersede that key. If no identity can be recovered, conservatively invalidate pending checks for the affected collection rather than allowing them to publish MATCHED.

```

</details>

tools/cdc-verify/compare.go
```diff
@@ -91,6 +92,65 @@ func isZeroValue(v any) bool {
...
+		return truncDiffString(t)
+	case []byte:
+		return truncDiffString(string(t))
+	case map[string]any:
+		out := make(map[string]any, len(t))
+		for k, vv := range t {
+			out[truncDiffString(k)] = sanitizeDiffValue(vv)
+		}
+		return out
+	case []any:
```
julianshen, line 116: _🏷️ data-modeling_ | _🟠 Warning_ | _`mongodb`_

**Diff sanitization leaves BSON containers effectively unbounded.**

The Mongo lookup path can return a legal BSON document containing an array or embedded document near MongoDB's 16 MiB limit. Both the direct and reflection-based container branches recursively copy every element without an aggregate size or element limit. A mismatch therefore retains and broadcasts the entire container; with the default 1,000-failure ring, distinct large failures can consume many gigabytes and each SSE event can be megabytes. Apply a total byte/element/depth budget per sanitized value and replace overflow with an omitted-element summary.

<details>

```
In `tools/cdc-verify/compare.go` around line 116: The Mongo lookup path can return a legal BSON document containing an array or embedded document near MongoDB's 16 MiB limit. Both the direct and reflection-based container branches recursively copy every element without an aggregate size or element limit. A mismatch therefore retains and broadcasts the entire container; with the default 1,000-failure ring, distinct large failures can consume many gigabytes and each SSE event can be megabytes. Apply a total byte/element/depth budget per sanitized value and replace overflow with an omitted-element summary.

```

</details>

tools/cdc-verify/compare.go
```diff
@@ -91,6 +92,65 @@ func isZeroValue(v any) bool {
...
+			out[truncDiffString(fmt.Sprint(iter.Key().Interface()))] = sanitizeDiffValue(iter.Value().Interface())
+		}
+		return out
+	case reflect.Slice, reflect.Array:
+		out := make([]any, rv.Len())
+		for i := range out {
+			out[i] = sanitizeDiffValue(rv.Index(i).Interface())
+		}
+		return out
+	default:
```
julianshen, line 138: _🎯 correctness_ | _🟠 Warning_ | _`mongodb`_

**Legal non-finite BSON doubles remain unsafe for JSON.**

MongoDB permits BSON doubles containing NaN and positive or negative infinity. The default branch returns those float values unchanged, but encoding/json rejects them with an unsupported-value error. A mismatch containing one of these values causes hub.broadcast to drop the result and can also make state or failure responses unencodable, contradicting the sanitizer's JSON-safety contract. Detect non-finite float32/float64 values and convert them to explicit string or structured representations before retention.

<details>

```
In `tools/cdc-verify/compare.go` around line 138: MongoDB permits BSON doubles containing NaN and positive or negative infinity. The default branch returns those float values unchanged, but encoding/json rejects them with an unsupported-value error. A mismatch containing one of these values causes hub.broadcast to drop the result and can also make state or failure responses unencodable, contradicting the sanitizer's JSON-safety contract. Detect non-finite float32/float64 values and convert them to explicit string or structured representations before retention.

```

</details>

tools/cdc-verify/bootstrap.go
julianshen, line 35: _🏷️ nats-lifecycle_ | _🟠 Warning_ | _`nats`_

**Concurrent bootstrap can fail after another instance creates the stream.**

Two bootstrap-enabled instances can both observe ErrStreamNotFound, then race to CreateStream. The winner creates MIGRATION-OPLOG, while the loser receives a stream-name-in-use error and aborts startup even though the required stream now exists. On create failure, re-query cfg.Name and treat confirmed existence as success; propagate the original error only if the stream is still absent or the lookup also fails.

<details>

```
In `tools/cdc-verify/bootstrap.go` around line 35: Two bootstrap-enabled instances can both observe ErrStreamNotFound, then race to CreateStream. The winner creates MIGRATION-OPLOG, while the loser receives a stream-name-in-use error and aborts startup even though the required stream now exists. On create failure, re-query cfg.Name and treat confirmed existence as success; propagate the original error only if the stream is still absent or the lookup also fails.

```

</details>

tools/cdc-verify/verifier.go
```diff
@@ -261,7 +263,10 @@ func (v *verifier) Submit(ev CDCEvent) {
 		row.Targets[i] = TargetResult{Alias: alias}
 	}
 
-	ctx, cancel := context.WithCancel(v.baseCtx)
+	// The verify deadline counts from Submit — semaphore wait included — so a
+	// queued check expires instead of parking forever behind MAX_CHECKS.
+	deadline := v.now().Add(v.cfg.Timeout)
+	ctx, cancel := context.WithTimeout(v.baseCtx, v.cfg.Timeout)
```
julianshen, line 269: _🏷️ resource-exhaustion_ | _🟠 Warning_ | _`performance-and-concurrency`_

**MAX_CHECKS still permits an unbounded goroutine backlog.**

Every accepted event creates a timeout context and timer, increments the wait group, and launches a goroutine before runCheck waits on the semaphore. When distinct-document arrival rate exceeds verification throughput, as during migration bursts or downstream lag, up to arrival_rate × VERIFY_TIMEOUT goroutines and timers accumulate despite MAX_CHECKS; at production rates this can cause heavy GC pressure or OOM. Replace this with a fixed MAX_CHECKS worker pool and bounded queue, or acquire bounded admission before spawning and apply explicit backpressure or overload handling. Preserve the submission timestamp so queued work can still expire at VERIFY_TIMEOUT.

<details>

```
In `tools/cdc-verify/verifier.go` around line 269: Every accepted event creates a timeout context and timer, increments the wait group, and launches a goroutine before runCheck waits on the semaphore. When distinct-document arrival rate exceeds verification throughput, as during migration bursts or downstream lag, up to arrival_rate × VERIFY_TIMEOUT goroutines and timers accumulate despite MAX_CHECKS; at production rates this can cause heavy GC pressure or OOM. Replace this with a fixed MAX_CHECKS worker pool and bounded queue, or acquire bounded admission before spawning and apply explicit backpressure or overload handling. Preserve the submission timestamp so queued work can still expire at VERIFY_TIMEOUT.

```

</details>

tools/cdc-verify/static/index.html
```diff
@@ -1113,16 +1129,26 @@ <h2>Failures <a href="/failures.json" download>Download JSON</a></h2>
...
+      });
+  }
+  loadState();
 
   var es = new EventSource('/api/events');
+  var esFirstOpen = true;
+  // A reconnect (network blip, or the hub dropped us as a slow client) means
+  // frames were missed — refetch full state instead of trusting the tail.
+  es.onopen = function () {
+    if (esFirstOpen) { esFirstOpen = false; return; }
```
julianshen, line 1149: _🎯 concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**State snapshots race with SSE delivery and can lose updates.**

The state fetch and EventSource registration have no synchronization boundary. On initial load, an event occurring after /api/state takes its snapshot but before the SSE client registers is absent from both, and esFirstOpen explicitly skips the resync. During reconnects, an SSE frame can also be applied while loadState is pending and then be overwritten by an older snapshot response; multiple reconnect fetches may resolve out of order. The dashboard can therefore permanently omit or revert check results until another reconnect. Buffer SSE frames while loading state and replay them after applying the snapshot, ideally using a server sequence/version to discard frames already represented, and abort or ignore stale loadState responses. The first open must participate in the same synchronization.

<details>

```
In `tools/cdc-verify/static/index.html` around line 1149: The state fetch and EventSource registration have no synchronization boundary. On initial load, an event occurring after /api/state takes its snapshot but before the SSE client registers is absent from both, and esFirstOpen explicitly skips the resync. During reconnects, an SSE frame can also be applied while loadState is pending and then be overwritten by an older snapshot response; multiple reconnect fetches may resolve out of order. The dashboard can therefore permanently omit or revert check results until another reconnect. Buffer SSE frames while loading state and replay them after applying the snapshot, ideally using a server sequence/version to discard frames already represented, and abort or ignore stale loadState responses. The first open must participate in the same synchronization.

```

</details>

tools/cdc-verify/static/index.html
julianshen, line 1150: _🎯 async-correctness_ | _🟠 Warning_ | _`silent-failure`_

**Reconnect state loads can overwrite newer SSE updates.**

On reconnect, loadState starts an asynchronous snapshot fetch while es.onmessage continues applying live frames. If the server takes snapshot S, then a result frame is applied, and S arrives afterward, applyInitialState can rebuild the UI from the older snapshot and silently remove or regress that result. Rapid reconnects can similarly apply state responses out of order. Buffer and replay SSE frames while loading state, or add a monotonic sequence/generation protocol that prevents an older snapshot from replacing newer events.

<details>

```
In `tools/cdc-verify/static/index.html` around line 1150: On reconnect, loadState starts an asynchronous snapshot fetch while es.onmessage continues applying live frames. If the server takes snapshot S, then a result frame is applied, and S arrives afterward, applyInitialState can rebuild the UI from the older snapshot and silently remove or regress that result. Rapid reconnects can similarly apply state responses out of order. Buffer and replay SSE frames while loading state, or add a monotonic sequence/generation protocol that prevents an older snapshot from replacing newer events.

```

</details>

tools/cdc-verify/compare.go
```diff
@@ -91,6 +92,65 @@ func isZeroValue(v any) bool {
...
+		return v
```
julianshen, line 139: _🏷️ error-handling_ | _🟠 Warning_ | _`silent-failure`_

**The sanitizer still retains JSON-unsupported numeric values.**

BSON doubles can contain NaN or positive/negative infinity. sanitizeDiffValue returns those float values unchanged, but encoding/json rejects them as unsupported values. A mismatch containing such a value is stored, then hub.broadcast logs the marshal error and sends no SSE frame; state or failure exports containing the row can also fail to encode. Normalize non-finite float32/float64 values into an explicit JSON-safe representation and add nested map/slice tests for them.

<details>

```
In `tools/cdc-verify/compare.go` around line 139: BSON doubles can contain NaN or positive/negative infinity. sanitizeDiffValue returns those float values unchanged, but encoding/json rejects them as unsupported values. A mismatch containing such a value is stored, then hub.broadcast logs the marshal error and sends no SSE frame; state or failure exports containing the row can also fail to encode. Normalize non-finite float32/float64 values into an explicit JSON-safe representation and add nested map/slice tests for them.

```

</details>

julianshen, line 139: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Non-finite floats still make sanitized diffs fail JSON encoding.**

The default branch returns scalar values unchanged, but BSON doubles can contain NaN or positive/negative infinity. encoding/json rejects those values, so hub.broadcast will drop the entire event despite sanitizeDiffValue's JSON-safety guarantee. Normalize non-finite float32/float64 values to a JSON-safe representation and add coverage for them.

<details>

```
In `tools/cdc-verify/compare.go` around line 139: The default branch returns scalar values unchanged, but BSON doubles can contain NaN or positive/negative infinity. encoding/json rejects those values, so hub.broadcast will drop the entire event despite sanitizeDiffValue's JSON-safety guarantee. Normalize non-finite float32/float64 values to a JSON-safe representation and add coverage for them.

```

</details>

tools/cdc-verify/static/index.html
```diff
@@ -386,6 +386,7 @@ <h2>Failures <a href="/failures.json" download>Download JSON</a></h2>
 
   var TERMINAL = { matched: true, failed: true, skipped: true, superseded: true };
   var FEED_CAP = 100;
+  var FAILURES_CAP = 1000; // client-side retention; matches the server's FAILED_CAP default
```
julianshen, line 389: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**The hard-coded browser cap ignores FAILED_CAP.**

FAILED_CAP is configurable, but the client always trims to 1,000. With FAILED_CAP above 1,000, the initial state can render all retained failures, then the next SSE failure deletes rows and map entries until only 1,000 remain. Expose the effective server cap in the state/config response and use it here, or otherwise keep client retention consistent with the configured server retention.

<details>

```
In `tools/cdc-verify/static/index.html` around line 389: FAILED_CAP is configurable, but the client always trims to 1,000. With FAILED_CAP above 1,000, the initial state can render all retained failures, then the next SSE failure deletes rows and map entries until only 1,000 remain. Expose the effective server cap in the state/config response and use it here, or otherwise keep client retention consistent with the configured server retention.

```

</details>

tools/cdc-verify/static/index.html
julianshen, line 1149: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**State reloads can overwrite events received during the reload.**

On reconnect, loadState runs while SSE messages continue to be processed. If an event is applied after the server took the state snapshot but before that fetch resolves, applyInitialState subsequently replaces the newer client state with the older snapshot and the event remains missing. Skipping the first onopen also leaves a gap when the initial state fetch completes before the first SSE connection succeeds. Buffer SSE frames while synchronizing, apply the snapshot, then replay the buffered frames; initiate that synchronization on every open, including the first.

<details>

```
In `tools/cdc-verify/static/index.html` around line 1149: On reconnect, loadState runs while SSE messages continue to be processed. If an event is applied after the server took the state snapshot but before that fetch resolves, applyInitialState subsequently replaces the newer client state with the older snapshot and the event remains missing. Skipping the first onopen also leaves a gap when the initial state fetch completes before the first SSE connection succeeds. Buffer SSE frames while synchronizing, apply the snapshot, then replay the buffered frames; initiate that synchronization on every open, including the first.

```

</details>

tools/cdc-verify/bootstrap.go
```diff
@@ -17,12 +18,21 @@ type bootstrapConfig struct {
...
 	}
 	cfg := stream.MigrationOplog(siteID)
-	if _, err := js.CreateOrUpdateStream(ctx, jetstream.StreamConfig{
+	_, err := js.Stream(ctx, cfg.Name)
+	if err == nil {
+		return nil
+	}
+	if !errors.Is(err, jetstream.ErrStreamNotFound) {
+		return fmt.Errorf("check MIGRATION-OPLOG stream: %w", err)
+	}
```
julianshen, line 34: _🎯 concurrency_ | _🟠 Warning_ | _`tgd-review`_

**Concurrent bootstrappers can fail after the existence check.**

The Stream lookup followed by CreateStream is a check-then-create race. If another verifier or the real pipeline creates the stream between these calls, CreateStream reports that the name already exists and this instance fails startup, even though the intended create-only end state was reached. Treat the specific already-exists result as success, preferably after re-reading the stream, while continuing to propagate other creation errors.

<details>

```
In `tools/cdc-verify/bootstrap.go` around line 34: The Stream lookup followed by CreateStream is a check-then-create race. If another verifier or the real pipeline creates the stream between these calls, CreateStream reports that the name already exists and this instance fails startup, even though the intended create-only end state was reached. Treat the specific already-exists result as success, preferably after re-reading the stream, while continuing to propagate other creation errors.

```

</details>

mliu33 (approved):

general-lex:
All review findings addressed in **baec5fb**:

- **Non-finite floats + unbounded containers** — `sanitizeDiffValue` now stringifies NaN/±Inf (BSON permits them, `encoding/json` rejects them) and bounds every axis: 64 elements per map/slice with a truncation marker, nesting depth capped at 8. A mismatch can no longer retain or broadcast a near-16 MiB BSON container.
- **Supersede vs terminal serialization** — the displaced handle is marked inside `beginCheck`/`supersedeKey` under `v.mu`, and `finish()` decides-and-publishes under the same lock: either the supersede lands first and wins, or the terminal publish completes before the newer event displaces the check — never a stale terminal after displacement.
- **Bounded backlog** — new `MAX_PENDING` (default 10000, `0` = unbounded) caps total enrolled checks including those queued behind `MAX_CHECKS`; new-key events past the cap shed visibly as skipped `overload` rows, same-key replacement never sheds. This closes the arrival_rate × VERIFY_TIMEOUT goroutine-accumulation window.
- **Exit status** — a watcher failure is remembered across the graceful SIGTERM drain and `main` exits 1 afterward; if self-signaling fails, it exits 1 directly. Supervisors that restart only failed processes now see the failure.
- **Log hygiene** — inspect errors are scrubbed of URI userinfo (`scheme://user:pass@` → `scheme://***@`) before reaching the server log.
- **Bootstrap race** — losing the check-then-create race (`ErrStreamNameAlreadyInUse`) is treated as the create-only end state, not a startup failure.
- **UI state/SSE consistency** — SSE frames arriving during a state reload are buffered and replayed after the snapshot applies; the UI resyncs on *every* `EventSource` open (first included, closing the snapshot-vs-registration gap); the client failures cap now comes from the server's `failedCap` in `/api/state` instead of a hard-coded 1000.
- **Coverage** — new integration test deletes the stream under a live watcher and asserts `Run` surfaces the closed-unexpectedly error.

One thread I'm pushing back on (undecodable events superseding pending checks) has an individual reply.

---

tools/cdc-verify/handler.go
```diff
@@ -65,7 +65,10 @@ func (h *handler) inspect(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusNotFound)
 		return
 	case err != nil:
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+		// Generic body only — a store error can carry URIs/hosts the browser
+		// must never see; the detail goes to the server log.
+		slog.Error("inspect failed", "collection", collection, "error", err)
+		http.Error(w, "inspect failed; see server log", http.StatusInternalServerError)
```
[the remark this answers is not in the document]
general-lex: Fixed in baec5fb: the logged error now passes `scrubLogValue`, which redacts URI userinfo (`scheme://user:pass@` → `scheme://***@`) before it reaches the log — covered by table tests. On the request-ID part: this diagnostic tool deliberately has no request-ID middleware (same as `tools/nats-debug`); the CLAUDE.md request-tracing rules target services, and a single-operator dashboard's inspect calls are attributable from the `collection` field already logged. Skipping that half.

---

tools/cdc-verify/verifier.go
general-lex: Accepting this as a documented limitation rather than fixing: an undecodable payload's document identity is unrecoverable by definition — every decode-failure mode (invalid JSON, missing `documentKey`, non-string `_id`) leaves no trustworthy doc key to supersede with. The alternatives are worse: superseding *all* pending checks for the collection would let one malformed event cancel unrelated in-flight verifications, and guessing a doc id from a half-parsed payload reintroduces the payload-trust problem the subject-routing fix removed. The skipped `decode-error` row makes the condition visible immediately, and the stale-check window it leaves is the same one the convergence model already tolerates for any gap in event delivery. If malformed events ever occur in practice, that row count is the signal to fix the producer.

---

PR #276, opened 2026-08-14
fix(cassandra): close four startup traps in session config and readiness

hmchangw:
Audit of every Cassandra touchpoint (`pkg/cassutil`, the four services and three tools that open sessions, their configs, and the compose/Makefile bring-up) turned up seven startup traps. This PR fixes four of them; the other three are listed at the bottom as follow-ups.

### 1. `history-service` — `required:"true"` was an inert struct tag

`internal/config/config.go` marked `CASSANDRA_HOSTS`, `MONGO_URI`, and `NATS_URL` required with a standalone `required:"true"` tag. `caarlos0/env` only reads `required` as an option **inside** the `env` tag, so all three were optional. Verified before the fix: with all three unset, `Load()` returned `err=<nil>, hosts="", mongoURI="", natsURL=""`.

Empty Cassandra hosts at least died loudly (gocql `ErrNoHosts`), but an empty `NATS_URL` did not — nats.go falls back to `nats://127.0.0.1:4222`, so the pod would silently connect to a nonexistent local NATS.

Fixed to `env:"HOSTS,required,notEmpty"` (etc.). `notEmpty` matters: `required` alone accepts an explicitly-set empty string.

### 2. `bot-message-worker` — `MESSAGE_BUCKET_HOURS` was never validated

`msgbucket.Sizer.Of` divides by the window, and the package doc states callers must validate at startup. Every other caller does (`message-worker`, `history-service`, `es-index-migrator`, `cdc-verify`); this one did not, so `MESSAGE_BUCKET_HOURS=0` booted clean and panicked on the **first message**.

Added `validateConfig(*config)`, called immediately after parse.

### 3. No Cassandra readiness probe anywhere

`history-service`, `message-worker`, and `bot-message-worker` registered only `natsutil.HealthCheck(nc)`. A Cassandra outage after startup left them reporting `200 /readyz` while every read failed.

Added `pkg/cassutil.HealthCheck(session)` — a `health.Check` named `cassandra` running `SELECT key FROM system.local` at consistency ONE (coordinator-served, so it exercises the pool without touching application data or a user keyspace) — and wired it into all three services.

**Trade-off worth a reviewer's attention:** unlike the NATS check, this probes a *shared* datastore, so a cluster-wide Cassandra outage marks every pod not-ready at once. That is the intent for a hard dependency — readiness, never liveness, so traffic sheds without restart loops — but it does mean history-service's endpoints empty out during a full outage rather than serving errors. Documented at the function. Each probe also emits a query span on the instrumented sessions, so keep the readiness interval coarse.

### 4. `message-worker` defaulted its connection string

`CASSANDRA_HOSTS envDefault:"localhost"` meant a missing env in production dialled `localhost:9042` and crashlooped on a driver error rather than a config error (CLAUDE.md: never default connection strings). Now `required,notEmpty`. Both composes that run this binary already set it, so no deployment change was needed.

Also corrected `message-worker/README.md`, whose config table listed defaults for four vars that are all actually required — three of those rows (`NATS_URL`, `SITE_ID`, `MONGO_URI`) were already wrong before this change.

## Testing

TDD per CLAUDE.md — every test was confirmed failing before implementation (undefined `validateConfig`/`healthCheckFor`; "An error is expected but got nil" for both config changes).

- `history-service/internal/config`: `TestLoad_RejectsMissingConnectionEnv` — unset **and** empty, for each of the three vars
- `bot-message-worker`: table-driven `TestValidateConfig` — 72 / 1 / 0 / -1
- `pkg/cassutil`: unit tests for the error mapping and context propagation via a `prober` seam (mirroring `natsutil`'s `connStatus`), plus an integration test asserting live-session ready → closed-session not-ready

| Gate | Result |
|---|---|
| `make lint` | pass |
| `make test` (race, all packages) | pass, 0 failures |
| `go vet -tags integration` on touched packages | pass — integration tests compile |
| `make sast` | gosec **pass**; govulncheck and semgrep **could not run** |

Two verification gaps to be aware of:

- **`govulncheck` and `semgrep` did not run.** `vuln.go.dev:443` returns a gateway 403 under the sandbox's network policy, and semgrep isn't installed there. This diff adds no dependencies and doesn't touch `errcode`/`WithCause`, so both should be clean — but that is an expectation, not a result. CI gates on them.
- **The new Cassandra integration test was compiled, not executed** (needs Docker). Worth running `make test-integration SERVICE=pkg/cassutil` before merge.

## Not addressed (follow-ups from the same audit)

- **`MESSAGE_BUCKET_HOURS` compose divergence.** `message-worker`/`teams-message-worker`/`history-service` composes don't set it (image default 72); `bot-message-worker`/`es-index-migrator` hardcode 72; `loadgen` uses `${MESSAGE_BUCKET_HOURS:-72}` and the Makefile passes `--env-file docker-local/.env` to every service. Setting it in `.env` shifts only loadgen's bucket math — silent partition divergence, with no runtime cross-check since the window isn't persisted anywhere.
- **No schema verification at startup.** gocql validates hosts and keyspace, but nothing checks that the tables/UDTs from `docker-local/cassandra/init/*.cql` exist. A service pointed at a fresh-but-empty keyspace comes up healthy and fails per-message.
- **Local bring-up has no Cassandra gate.** `require-deps` inspects only the NATS container; no service compose has `depends_on` or `restart:`, and Cassandra's healthcheck has a 60s `start_period`, so `make up` too soon leaves permanently dead containers. Separately, `cassandra-init` sits in the `init` profile and only `make deps-up` runs it.

Lower-severity items also left alone: `MESSAGE_READ_MAX_BUCKETS` ↔ `MESSAGE_BUCKET_HOURS` coupling is unchecked (122 × 72h ≈ 366d matches the 365d floor by default only), writer/reader `ATREST_ENABLED` asymmetry isn't cross-checked at startup, and `bot-message-worker`'s `CASSANDRA_NUM_CONNS=4` is an unexplained outlier against cassutil's default of 8.

https://claude.ai/code/session_01LZxpmRCGEEYXW3wR76wRmh

---
_Generated by [Claude Code](https://claude.ai/code/session_01LZxpmRCGEEYXW3wR76wRmh)_

### DIFF
--- a/bot-message-worker/config_test.go
+++ b/bot-message-worker/config_test.go
@@ -1,0 +1,37 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestValidateConfig(t *testing.T) {
+	tests := []struct {
+		name       string
+		hours      int
+		wantErrSub string
+	}{
+		{name: "default window", hours: 72},
+		{name: "one hour window", hours: 1},
+		// msgbucket.Sizer divides by the window, so a non-positive value panics
+		// on the first message unless startup rejects it.
+		{name: "zero window", hours: 0, wantErrSub: "MESSAGE_BUCKET_HOURS"},
+		{name: "negative window", hours: -1, wantErrSub: "MESSAGE_BUCKET_HOURS"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			cfg := config{MessageBucketHours: tt.hours}
+
+			err := validateConfig(&cfg)
+
+			if tt.wantErrSub == "" {
+				require.NoError(t, err)
+				return
+			}
+			require.Error(t, err)
+			assert.Contains(t, err.Error(), tt.wantErrSub)
+		})
+	}
+}
--- a/bot-message-worker/main.go
+++ b/bot-message-worker/main.go
@@ -169,6 +169,7 @@ func run() error {
 
 	healthStop, err := health.ServeWithPprof(cfg.HealthAddr, 5*time.Second, cfg.PProfEnabled,
 		natsutil.HealthCheck(nc),
+		cassutil.HealthCheck(cassSess),
 	)
 	if err != nil {
 		return fmt.Errorf("health server: %w", err)
--- a/bot-message-worker/main.go
+++ b/bot-message-worker/main.go
@@ -58,11 +58,24 @@ func main() {
 	}
 }
 
+// validateConfig rejects values that would otherwise fail long after startup:
+// msgbucket.Sizer divides by the window, so a non-positive MESSAGE_BUCKET_HOURS
+// panics on the first message instead of failing the boot.
+func validateConfig(cfg *config) error {
+	if cfg.MessageBucketHours < 1 {
+		return fmt.Errorf("MESSAGE_BUCKET_HOURS must be >= 1, got %d", cfg.MessageBucketHours)
+	}
+	return nil
+}
+
 func run() error {
 	ctx := context.Background()
 	cfg, err := env.ParseAs[config]()
 	if err != nil {
 		return fmt.Errorf("parse config: %w", err)
+	}
+	if err := validateConfig(&cfg); err != nil {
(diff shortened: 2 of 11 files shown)


hmchangw:
Close for now. Not critical.

Issue #39, opened 2026-07-14
[Fix] Room-key cache TTL-safety validated against drift-prone per-service grace config


hmchangw:
## Summary
The room-key cache safety invariant ("cache TTL < key grace") is validated in `broadcast-worker` against **broadcast-worker's own** `ROOM_KEY_GRACE_PERIOD` env — but the **authoritative** grace is whatever `room-worker` stamps at rotation (`prevExpiresAt = now + room-worker's ROOM_KEY_GRACE_PERIOD`). Same env **name**, **independent per-service values**, nothing couples them. If they drift (room-worker grace < broadcast TTL), the guard still passes and broadcast-worker serves a rotated-out key past the real grace → **messages clients can no longer decrypt**, with no guardrail catching it.

## Root cause  (verified, file:line)
- `broadcast-worker/main.go:159` `keyCacheTTLSafe(ttl, grace)` = `ttl < grace`, called at `:167` with `cfg.RoomKeyCacheTTL` + **broadcast-worker's** `cfg.RoomKeyGracePeriod` (`main.go:49`, default 24h).
- Authoritative grace is set by the rotator: `room-worker/main.go:48` `ROOM_KEY_GRACE_PERIOD` (default 24h) → `roomkeystore.NewMongoStore(rooms, grace)` stamps `prevExpiresAt`. Read path honors `prevExpiresAt` (`pkg/roomkeystore/roomkeystore_mongo.go:245`).
- The two `ROOM_KEY_GRACE_PERIOD` values are configured independently per service; the safety check assumes—but does not enforce—that they match.

## Message-path trace
1–7: N/A except **workers (broadcast-worker config/validation)**. No data-path change; this is a config-correctness/guard issue.

## Fix
Make the grace value a single source of truth so the safety check uses the *authoritative* grace: e.g. read the effective grace from the room/keystore doc (the value actually stamped into `prevExpiresAt`) rather than broadcast-worker's local env, or centralize `ROOM_KEY_GRACE_PERIOD` so both services provably share it (+ assert/document the coupling and fail-closed on mismatch).

## Contracts & docs
- docs/client-api.md? N. Schema change? N (may add a documented invariant).

## Tests
- Unit: `keyCacheTTLSafe` against the authoritative grace; cache disabled (fail-closed) when TTL ≥ authoritative grace even if the local env disagrees.

## Acceptance criteria
[...]

Issue #34, opened 2026-07-14
[Fix] messageRead should publish a subscription.update (action "read") to the reader


hmchangw:
## Summary
When a user marks a room read, `messageRead` persists `lastSeenAt` and (optionally) publishes the shared read-floor event to other members — but never publishes a per-reader `subscription.update`. So the reader's **other devices** can't sync the cleared-unread state without polling. Fix: publish a `subscription.update` (action `"read"`) on `chat.user.{account}.event.subscription.update` for the reader, on both the same-site (room-service) and cross-site (inbox-worker) paths. Same multi-device shape as #323.

## Root cause  (verified, file:line — code is source of truth)
- `room-service/handler.go:1192` `messageRead`: after `store.UpdateSubscriptionRead` (line 1208) it does the floor fan-out + cross-site outbox (line 1263), but never calls `publishSubscriptionUpdate` for the reader's own account. Mute/favorite toggles DO (`handler.go:1827`, `1904`) — read is the gap.
- `inbox-worker/handler.go:234` `handleSubscriptionRead` (cross-site path) is write-only — same gap.

## Message-path trace  (mark change / no-change / N/A per segment)
1. Client request struct + docs/client-api.md — No request change (`messageRead` is no-body). Update the "Mark Messages Read" section (`docs/client-api.md:1244`) same PR to document the new triggered `subscription.update` event.
2. Gatekeeper mapping (request → canonical) — N/A — direct room-service handler.
3. Canonical model / event structs (pkg/model) — No new struct; `model.SubscriptionUpdateEvent` (`pkg/model/event.go:60`) already carries `Subscription` + `Action`. Add the `"read"` action value to the documented set (`event.go:63`).
4. Workers — **Change in room-service + inbox-worker.** room-service `messageRead`: after the read write, call `publishSubscriptionUpdate(ctx, account, "read", sub, now)` (pattern: `handler.go:1827/1904`). inbox-worker `handleSubscriptionRead`: publish the same event (needs a `publishCore` wired in — see open questions).
5. Storage tables (Cassandra / Mongo) — No change (`subscriptions` already updated by `UpdateSubscriptionRead`).
6. Read / history path (history-service) — No change (read-floor logic untouched).
7. Client-facing events (Leg-2 fan-out) — **Change.** Adds one `subscription.update` to the reader per `messageRead` (best-effort). The existing `message_read` floor event to other members is unchanged/orthogonal.

## Fix
[...]

PR #279, opened 2026-08-14
feat(federation): order remote rooms in the chat list

hmchangw:
## Problem

A subscription to a room owned by another site has **no local `rooms` document**, permanently. `inbox-worker.handleMemberAdded` creates a subscription and nothing else, and `room_sync` — the only full-room replication path — is published solely by `data-migration/oplog-collections-transformer`, never by room-worker or room-service in normal operation.

Chat-list ordering keys on `room.lastMsgAt`. With no room doc there is no key, so every remote room sorts below every local one, ordered only by the subscription&#39;s denormalized `name`. With a paginated client that caps how far it reads, **an active remote room can never appear** — and the more active it is, the more wrong its position. Under an `updatedWithinDays` window it is worse: the cutoff tests `lastMsgAt == nil`, so remote rooms are dropped from the list entirely.

The room&#39;s real `lastMsgAt` *is* fetched — `enrichCrossSite` RPCs each remote site — but in the service layer, **after** the repo has already sorted and paged. The data needed to order correctly arrives strictly after the ordering decision.

Pre-existing behaviour, not a regression: the old in-Mongo `$ifNull` sort buried missing rooms the same way.

## Approach

Give each site the ordering position for the remote rooms its users belong to, from two sources — seeded when a member joins, refreshed as the room stays active.

**1. Wire contract** (`90daf6e`). `MemberAddEvent.LastMsgAt` seeds; `RoomActivityEvent` refreshes; `RemoteRoom` is the stored row. `subject.RoomActivity` is deliberately outside `chat.inbox.&gt;` so the refresh never touches the INBOX JetStream stream — it is coalesced, idempotent and `$max`-guarded, needing neither persistence nor ordering, and a high-rate signal there would compete for retention and ack budget with membership events that need both.

**2. Producers** (`1fae56f`). Four call sites publish cross-site `member_added`. room-worker&#39;s add-members path must not read the position off `room` — that comes from `GetRoomMeta`, built from `roommetacache.Meta`, which structurally has no `lastMsgAt`, so it would be nil unconditionally; the full room is already re-read for the subscription fan-out and that value is reused. `bot-room-service` is the fourth producer, reached when a user joins a bot-owned channel that already has history; its DM-ensure path matters too, since the DM id is deterministic and ensure can land on an existing room. The two creation paths pass nil, which is correct.

**3. Storage and lifecycle** (`ea957cf`). `remote_rooms` (`model.RemoteRoom`), written by inbox-worker from both the seed and the refresh. Deliberately not the `rooms` collection: that one is room-service&#39;s and its readers assume complete documents (members, `encKey`, counts) a federated peer cannot supply. Migration&#39;s `room_sync` replicates whole rooms onto the same external lane, so a migrated room can hold both and a reader spanning the two must prefer `rooms`.

The write is a `$max`-guarded upsert — idempotent and order-independent, which is what lets the refresh ride a lossy transport. A duplicate key means two concurrent first-touches, not a stale event (unlike `UpsertRoom`, whose guard lives in the filter), so it retries rather than swallowing the newer value. Lifecycle is closed at both ends: seeding runs after the missing-user return so a room that gained no subscriber leaves no row, and `member_removed` drops the row once no member remains.

**4. Refresh publisher** (`6cbe4aa`). Rides the coalescer that already buffers `rooms.lastMsgAt` on a 250ms flush.

## Cost

The refresh is broadcast to every peer, which needed bounding in three places:

| | naive | shipped |
|---|---|---|
| Publishes/sec (1k active cross-site rooms, 5 sites) | ~4,000 | ~200 |
| Destination Mongo reads/sec | ~1,000 | ~0 |

- **Coalescing** collapses a flush window to one entry per room, so cost is independent of message rate.
- **`ROOM_ACTIVITY_REFRESH_INTERVAL`** (default 5s) then caps how often a room is announced at all, decoupling it from the Mongo flush — a ~20× cut. The interval can be that generous because the consumer cannot see the difference: `subscription.list` serves ordering from a sort-key cache whose own TTL is an order of magnitude longer.
- **`ROOM_SUB_CACHE_SIZE`/`_TTL`** memoize the destination&#39;s membership check, taking the per-message Mongo read to zero in the steady state. Both handlers keep entries in step rather than relying on expiry, and read-through fills never overwrite what a membership handler wrote — losing that race would pin a freshly-joined room at its seed for the whole TTL.

**One room per message, never a batch.** Payload size is then bounded by construction; a batched form would grow with active-room count and eventually exceed `max_payload` — a cliff appearing only at peak load, losing a whole flush window at once.

`ALL_SITE_IDS` unset disables the lane. Correct for a single site, but a warning says so, since a federated deployment missing it would look healthy while remote chat lists silently could not order this site&#39;s rooms.

## Considered and rejected

**Targeting peers from a `Room.MemberSites` list.** Removal is the sharp edge: dropping one member doesn&#39;t prove no others from that site remain. Refcounts break under the at-least-once, idempotent federation lanes (`$inc` is not idempotent, and a redelivered `member_removed` silently under-counts); recomputing is O(room size) per removal; sticky-never-shrink decays toward broadcasting anyway. `Room.CrossSite` already faced this and chose sticky over-approximation. After the throttle and the cache, this would save ~30 KB/s per peer — not worth a membership-tracking mechanism with a correctness hazard in it.

## Scope — what this does *not* do

- **Nothing reads `remote_rooms` yet.** The read-path fallback belongs in `resolveSortKeys`, which exists only on #197&#39;s branch; on `main`, `AggregateSubscriptions` is still the `$lookup` pipeline. It lands once #197 merges — that is the commit that makes this user-visible.
- **No backfill** for existing subscriptions. Cheapest route is write-behind from the `GetRoomsInfo` results the service layer already fetches per page, which doubles as ongoing repair.
- **The four producers are not consolidated.** &#34;Cross-site copies carry the position&#34; lives in four hand-maintained call sites across two services, and nothing catches a fifth appearing. A `model.NewCrossSiteMemberAdd(...)` constructor would make it a property of the type — worth doing, too large here.
- **No end-to-end integration test** across the wire (publisher → core NATS → consumer → `remote_rooms`). Each half is covered against fakes, but nothing yet proves they agree on the payload. `pkg/testutil.NATS(t)` would support it.

## Testing

TDD throughout; every test failed for the intended reason before implementation.

- **inbox-worker:** table over seed fires / nothing to seed / unknown user leaves no orphan row / seed failure doesn&#39;t fail the event; refresh applies, drops non-member rooms, rejects malformed and position-less payloads; cache serves from memory, caches negatives, promotes on `member_added`, re-reads after `member_removed`.
- **broadcast-worker:** cross-site rooms only; coalesces to one publish per room; throttles per room and independently across rooms; prunes watermarks; publish failure doesn&#39;t fail the flush; one message per peer with a bounded payload.
- **room-worker / bot-room-service:** the mock mirrors the real store&#39;s shape (`GetRoomMeta` returns no `lastMsgAt`) so a regression reading it from there fails; bot-room-service unwraps the published `OutboxEvent` to assert what the destination actually receives.
- **Integration:** the `$max` guard against newer (advances), older (no-op), and first-touch completeness.

`make lint` 0 issues; full-repo `make test` (race) passes across 112 packages; `gosec` clean. `govulncheck` and the semgrep registry rulesets are proxy-blocked in the sandbox — CI&#39;s `sast` job covers them. Integration tests need a Docker daemon this container lacks; CI covers them, and it caught a nil `remoteRoomCol` in test-only store literals that local runs could not.

## Note for reviewers

`inbox-worker` has no `//go:generate` directive and its `MockInboxStore` is orphaned (tests use a hand-written `stubInboxStore`), so `make generate` is a no-op there and the generated mock is stale against the interface. Unused, so nothing breaks; the installed `mockgen` is built with go1.24 and cannot load this go1.25 module, so I left it rather than hand-editing.

The duplicate-key retry path has no test — forcing a genuine concurrent-insert race deterministically needs fault injection the suite has no facility for.

https://claude.ai/code/session_01EfPa3pTMGaKVzV1xe1sqp7

### DIFF
--- a/bot-room-service/handler.go
+++ b/bot-room-service/handler.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hmchangw/chat/pkg/roomkeysender"
 	"github.com/hmchangw/chat/pkg/roomkeystore"
 	"github.com/hmchangw/chat/pkg/subject"
+	"github.com/hmchangw/chat/pkg/timeutil"
 )
 
 // Rocket.Chat-legacy single-char rooms.t values.
@@ -126,8 +127,19 @@ func (h *handler) handleDMEnsure(c *natsrouter.Context, req BotDMEnsureRequest) 
 		},
 		CreatedByBot: ident.ID,
 	}
-	if err := h.store.InsertRoom(c, room); err != nil && !errors.Is(err, ErrDuplicate) {
-		return nil, fmt.Errorf("insert dm room: %w", err)
+	if err := h.store.InsertRoom(c, room); err != nil {
+		if !errors.Is(err, ErrDuplicate) {
+			return nil, fmt.Errorf("insert dm room: %w", err)
+		}
+		// The DM id is deterministic, so a duplicate means this room already
+		// exists and may carry history — take its position rather than federating
+		// nil and leaving the target unable to order it.
+		if existing, findErr := h.store.FindRoom(c, roomID); findErr == nil {
+			room.LastMsgAt = existing.LastMsgAt
+		} else {
+			slog.WarnContext(c, "read existing dm room failed; federating without an activity position",
+				"room_id", roomID, "error", findErr)
+		}
 	}
 	if _, err := h.store.UpsertSubscription(c, &Subscription{
 		ID: h.newUUIDv7(), RoomID: roomID, UserID: ident.ID, Account: ident.Account,
@@ -147,7 +159,7 @@ func (h *handler) handleDMEnsure(c *natsrouter.Context, req BotDMEnsureRequest) 
 	} else {
 		// DM member_added carries roomType=botDM + RequesterAccount=bot so the target names the subscription after the counterparty; subscription-only, no rooms doc.
 		if err := h.federateMemberAdded(c, roomID, target.ID, target.Account, target.SiteID, createdAt,
-			model.RoomTypeBotDM, "", ident.Account); err != nil {
+			model.RoomTypeBotDM, "", ident.Account, room.LastMsgAt); err != nil {
 			return nil, err
 		}
 	}
@@ -238,7 +250,7 @@ func (h *handler) handleCreate(c *natsrouter.Context, req BotCreateRoomRequest) 
 		// Channel-shape event; RoomName names the subscription at the target.
 		if u.SiteID != "" && u.SiteID != h.siteID {
 			if err := h.federateMemberAdded(c, roomID, u.ID, u.Account, u.SiteID, createdAt,
-				model.RoomTypeChannel, req.Name, ident.Account); err != nil {
+				model.RoomTypeChannel, req.Name, ident.Account, nil); err != nil {
 				return nil, err
 			}
 		}
@@ -309,7 +321,7 @@ func (h *handler) handleAdd(c *natsrouter.Context, req BotMembersBatchRequest) (
 		if u.SiteID != "" && u.SiteID != h.siteID {
 			roomType := roomTypeToModel(room.Type)
 			if err := h.federateMemberAdded(c, roomID, u.ID, u.Account, u.SiteID, created,
-				roomType, room.Name, ident.Account); err != nil {
+				roomType, room.Name, ident.Account, room.LastMsgAt); err != nil {
 				return nil, err
 			}
 		}
@@ -530,8 +542,9 @@ func (h *handler) loadRoomAndAssertOwner(ctx context.Context, roomID string, ide
 
 // federateMemberAdded relays member_added to a remote site's inbox-worker.
 // For DM/botDM the target names the subscription from RequesterAccount; for channels, from RoomName.
+// lastMsgAt is nil for a room with no messages yet (both creation paths).
 func (h *handler) federateMemberAdded(ctx context.Context, roomID, userID, account, destSiteID string, at time.Time,
-	roomType model.RoomType, roomName, requesterAccount string,
+	roomType model.RoomType, roomName, requesterAccount string, lastMsgAt *time.Time,
 ) error {
 	payload, err := json.Marshal(model.MemberAddEvent{
 		Type:             "member_added",
@@ -542,6 +555,7 @@ func (h *handler) federateMemberAdded(ctx context.Context, roomID, userID, accou
 		SiteID:           h.siteID,
 		Accounts:         []string{account},
 		JoinedAt:         at.UnixMilli(),
+		LastMsgAt:        timeutil.TimeToMillis(lastMsgAt),
 		Timestamp:        at.UnixMilli(),
 	})
 	if err != nil {
--- a/broadcast-worker/coalescer.go
+++ b/broadcast-worker/coalescer.go
@@ -27,6 +27,13 @@ type bulkRoomLastMsgWriter interface {
 	BulkUpdateRoomLastMessage(ctx context.Context, updates map[string]roomLastMsgUpdate) error
 }
 
+// roomActivityRefresh is one room's coalesced position, handed to the publisher
+// so destination sites can order a room they hold no rooms doc for.
+type roomActivityRefresh struct {
+	roomID string
+	at     time.Time
+}
+
 // coalescingStore wraps an inner Store and intercepts UpdateRoomLastMessage,
 // buffering the latest (msgID, createdAt, mentionAll) per roomID in memory.
 // Flush periodically drains the buffer through a single Mongo BulkWrite.
@@ -44,8 +51,32 @@ type coalescingStore struct {
 	Store
 	bulk bulkRoomLastMsgWriter
 
+	// crossSite reports whether a room has members off this site; nil disables
+	// the refresh entirely. Backed by the room-meta cache, so the flush pays a
+	// cache hit rather than a read — the handler just resolved the same room.
+	crossSite func(ctx context.Context, roomID string) bool
+	// publishActivity emits one refresh; nil disables. Errors are logged, never
+	// propagated — see Flush.
+	publishActivity func(ctx context.Context, r roomActivityRefresh) error
+	// refreshInterval throttles refreshes to at most one per room per interval,
+	// independently of the (much shorter) Mongo flush cadence. Non-positive
+	// publishes on every flush. See refreshRemoteActivity for why this is safe.
+	refreshInterval time.Duration
+	// lastRefreshed is the per-room watermark backing that throttle. Pruned each
+	// flush, so it stays bounded by rooms active within the interval rather than
+	// growing with every room ever seen.
+	lastRefreshed map[string]time.Time
+	now           func() time.Time
+
 	mu      sync.Mutex
 	pending map[string]roomLastMsgUpdate
+}
+
+func (c *coalescingStore) clock() time.Time {
+	if c.now != nil {
+		return c.now()
+	}
+	return time.Now()
 }
 
 func newCoalescingStore(inner Store, bulk bulkRoomLastMsgWriter) *coalescingStore {
@@ -85,7 +116,69 @@ func (c *coalescingStore) Flush(ctx context.Context) error {
 	batch := c.pending
 	c.pending = make(map[string]roomLastMsgUpdate, len(batch))
 	c.mu.Unlock()
+	c.refreshRemoteActivity(ctx, batch)
 	return c.bulk.BulkUpdateRoomLastMessage(ctx, batch)
+}
+
+// refreshRemoteActivity emits at most one activity refresh per cross-site room
+// per refreshInterval. Cost scales with distinct active cross-site rooms per
+// interval, not with message rate — coalescing collapses the window, and the
+// throttle decouples the announce rate from the (much shorter) Mongo flush.
+//
+// The interval can be generous because the consumer cannot see the difference:
+// the subscription list serves ordering from a cache whose own TTL is an order
+// of magnitude longer, so a position a few seconds behind is indistinguishable
+// from a fresh one. The cost of the throttle is that a room going quiet right
+// after a burst keeps the position from its last announce — stale by at most
+// one interval, and repaired by its next message.
+//
+// Failures are logged, not returned: the position is decorative, the next
+// message re-establishes it, and the room batch this rides with is the write
+// that actually matters.
+func (c *coalescingStore) refreshRemoteActivity(ctx context.Context, batch map[string]roomLastMsgUpdate) {
+	if c.publishActivity == nil || c.crossSite == nil {
+		return
+	}
+	now := c.clock()
+	for roomID, u := range batch {
+		if !c.crossSite(ctx, roomID) {
+			continue
+		}
+		if c.throttled(roomID, now) {
+			continue
+		}
+		if err := c.publishActivity(ctx, roomActivityRefresh{roomID: roomID, at: u.at}); err != nil {
+			slog.WarnContext(ctx, "publish room activity refresh failed", "room_id", roomID, "error", err)
+		}
+	}
+	c.pruneWatermarks(now)
+}
+
+// throttled reports whether roomID was announced within refreshInterval, and
+// records the announce when it was not.
+func (c *coalescingStore) throttled(roomID string, now time.Time) bool {
+	if c.refreshInterval <= 0 {
+		return false
+	}
+	if last, ok := c.lastRefreshed[roomID]; ok && now.Sub(last) < c.refreshInterval {
+		return true
+	}
+	if c.lastRefreshed == nil {
+		c.lastRefreshed = make(map[string]time.Time)
+	}
+	c.lastRefreshed[roomID] = now
+	return false
+}
+
+// pruneWatermarks drops rooms that have gone quiet for longer than the
+// interval; their next message re-announces immediately, so keeping them would
+// only leak memory.
+func (c *coalescingStore) pruneWatermarks(now time.Time) {
+	for roomID, last := range c.lastRefreshed {
+		if now.Sub(last) >= c.refreshInterval {
+			delete(c.lastRefreshed, roomID)
+		}
+	}
 }
 
 // Run drives the periodic flush loop until ctx is cancelled. On cancellation a
--- a/broadcast-worker/roomactivity.go
+++ b/broadcast-worker/roomactivity.go
@@ -1,0 +1,81 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/hmchangw/chat/pkg/model"
+	"github.com/hmchangw/chat/pkg/roommetacache"
+	"github.com/hmchangw/chat/pkg/subject"
+)
+
+// metaReader is the slice of Store the cross-site check needs.
+type metaReader interface {
+	GetRoomMeta(ctx context.Context, roomID string) (roommetacache.Meta, error)
+}
+
+// remotePeers returns the sites to refresh: everything in allSiteIDs except this
+// one, blanks dropped and duplicates collapsed. Empty means single-site, which
+// disables the refresh.
+func remotePeers(siteID string, allSiteIDs []string) []string {
+	seen := make(map[string]struct{}, len(allSiteIDs))
+	var peers []string
+	for _, s := range allSiteIDs {
+		if s == "" || s == siteID {
+			continue
+		}
+		if _, dup := seen[s]; dup {
+			continue
+		}
+		seen[s] = struct{}{}
+		peers = append(peers, s)
+	}
+	return peers
+}
+
+// crossSiteChecker reports whether a room has members off this site, reading the
+// room-meta cache. An unclassified room (nil CrossSite) counts as cross-site:
+// the cost of guessing wrong is one wasted publish, and the destination ignores
+// rooms it holds no subscription for. A meta read failure skips the refresh —
+// the next message in the room retries it.
+func crossSiteChecker(s metaReader) func(ctx context.Context, roomID string) bool {
+	return func(ctx context.Context, roomID string) bool {
+		meta, err := s.GetRoomMeta(ctx, roomID)
+		if err != nil {
+			return false
+		}
+		return meta.CrossSite == nil || *meta.CrossSite
+	}
+}
+
+// roomActivityPublisher emits one refresh per peer site on the core-NATS
+// activity lane. One room per message keeps the payload bounded by
+// construction — a batched form would grow with the number of active rooms and
+// eventually exceed max_payload, silently losing a whole flush window.
+func roomActivityPublisher(p Publisher, siteID string, peers []string) func(context.Context, roomActivityRefresh) error {
+	return func(ctx context.Context, r roomActivityRefresh) error {
+		evt := model.RoomActivityEvent{
+			RoomID:    r.roomID,
+			SiteID:    siteID,
+			LastMsgAt: r.at.UTC().UnixMilli(),
+			Timestamp: time.Now().UTC().UnixMilli(),
+		}
+		data, err := json.Marshal(evt)
+		if err != nil {
+			return fmt.Errorf("marshal room activity event: %w", err)
+		}
+		// Every peer is attempted: one unreachable site must not deprive the
+		// others of the refresh, since a skipped peer keeps a stale position for
+		// the whole throttle interval.
+		var errs []error
+		for _, peer := range peers {
+			if err := p.Publish(ctx, subject.RoomActivity(peer), data); err != nil {
+				errs = append(errs, fmt.Errorf("publish room activity to %s: %w", peer, err))
+			}
+		}
+		return errors.Join(errs...)
+	}
+}
--- a/inbox-worker/handler.go
+++ b/inbox-worker/handler.go
@@ -28,6 +28,20 @@ type InboxStore interface {
 	// stored one is a silent no-op, so out-of-order federated delivery cannot
 	// regress room metadata.
 	UpsertRoom(ctx context.Context, room *model.Room) error
+	// UpsertRemoteRoomActivity records a remote room's ordering position under a
+	// $max guard, so a late or duplicate event is a no-op rather than a
+	// regression — what will let the activity-refresh publisher ride a lossy,
+	// unordered transport. Until it lands member_added is the only writer, so a
+	// dropped write is NOT self-healing.
+	UpsertRemoteRoomActivity(ctx context.Context, roomID, siteID string, lastMsgAt time.Time) error
+	// DeleteRemoteRoomActivity drops a remote room's ordering row once this site
+	// has no member left in it, so the collection does not accumulate rows for
+	// rooms nobody here can see.
+	DeleteRemoteRoomActivity(ctx context.Context, roomID string) error
+	// HasRoomSubscription reports whether this site holds any subscription for
+	// roomID. The activity refresh is broadcast to every peer, so this is what
+	// stops a site accumulating ordering rows for rooms none of its users are in.
+	HasRoomSubscription(ctx context.Context, roomID string) (bool, error)
 	// UpdateSubscriptionRoles applies roles guarded by rolesUpdatedAt (the source
 	// event's publish time): older/duplicate events are silent no-ops. A
 	// genuinely missing subscription still returns an error so the event is
@@ -122,11 +136,56 @@ type Handler struct {
 	// badge is the badge cache; nil (VALKEY_ADDRS unset) disables the
 	// invalidation hooks. Injected post-construction.
 	badge badgeCache
+	// roomSubs memoizes the "is this site a member of this room" check the
+	// activity refresh performs; nil disables it and every refresh reads through.
+	roomSubs *roomSubCache
+}
+
+// HandlerOption configures optional Handler behaviour.
+type HandlerOption func(*Handler)
+
+// WithRoomSubCache memoizes the room-membership check on the activity-refresh
+// lane. A non-positive size or ttl leaves it disabled.
+func WithRoomSubCache(size int, ttl time.Duration) HandlerOption {
+	return func(h *Handler) { h.roomSubs = newRoomSubCache(size, ttl) }
 }
 
 // NewHandler creates a Handler with the given store.
-func NewHandler(store InboxStore) *Handler {
-	return &Handler{store: store}
+func NewHandler(store InboxStore, opts ...HandlerOption) *Handler {
+	h := &Handler{store: store}
+	for _, opt := range opts {
+		opt(h)
+	}
+	return h
+}
+
+// HandleRoomActivity applies a core-NATS activity refresh for a room owned by
+// another site. Dropped when this site holds no subscription for the room: the
+// refresh is broadcast to every peer, and a row here would otherwise be an
+// orphan nothing deletes.
+//
+// Errors are returned for logging only — there is no ack on this lane, and the
+// $max guard makes the next refresh idempotent, so a loss self-heals on the
+// room's next message.
+func (h *Handler) HandleRoomActivity(ctx context.Context, data []byte) error {
+	var evt model.RoomActivityEvent
+	if err := json.Unmarshal(data, &evt); err != nil {
+		return fmt.Errorf("unmarshal room activity event: %w", err)
+	}
+	if evt.RoomID == "" || evt.LastMsgAt <= 0 {
+		return fmt.Errorf("room activity event missing roomId or lastMsgAt")
+	}
+	subscribed, err := h.hasRoomSubscription(ctx, evt.RoomID)
+	if err != nil {
+		return fmt.Errorf("check room subscription: %w", err)
+	}
+	if !subscribed {
+		return nil
+	}
+	if err := h.store.UpsertRemoteRoomActivity(ctx, evt.RoomID, evt.SiteID, time.UnixMilli(evt.LastMsgAt).UTC()); err != nil {
+		return fmt.Errorf("apply room activity: %w", err)
+	}
+	return nil
 }
 
 // HandleEvent processes a single JetStream message payload.
@@ -254,6 +313,18 @@ func (h *Handler) handleMemberAdded(ctx context.Context, evt *model.InboxEvent) 
 		return fmt.Errorf("member_added references unknown users %v in room %s", missing, event.RoomID)
 	}
 
+	// After the missing-user return, so a room that gained no subscriber leaves
+	// no orphan row (nothing deletes them). Best-effort — the subscriptions above
+	// must not be re-run because an ordering row failed — but not self-healing
+	// yet, so alert on this log line.
+	h.roomSubs.set(event.RoomID, true)
+	if event.LastMsgAt != nil {
+		if err := h.store.UpsertRemoteRoomActivity(ctx, event.RoomID, event.SiteID, time.UnixMilli(*event.LastMsgAt).UTC()); err != nil {
+			slog.WarnContext(ctx, "seed remote room activity failed",
+				"room_id", event.RoomID, "site", event.SiteID, "error", err)
+		}
+	}
+
 	// No SubscriptionUpdateEvent is published here — room-worker already publishes
 	// to the user's subject and the NATS supercluster routes it to the user's
 	// home site.
@@ -298,6 +369,20 @@ func (h *Handler) handleMemberRemoved(ctx context.Context, evt *model.InboxEvent
 	}
 	if err := h.store.DeleteSubscriptionsByAccounts(ctx, memberEvt.RoomID, memberEvt.Accounts); err != nil {
 		return fmt.Errorf("delete subscriptions for room %s: %w", memberEvt.RoomID, err)
+	}
+	// Other members may remain, so re-resolve rather than caching a guess. When
+	// none do, the ordering row has no reader left here — drop it, or it becomes
+	// the orphan the seed path is careful not to create. Best-effort: a stale row
+	// only costs space, and re-adding a member re-seeds it.
+	h.roomSubs.invalidate(memberEvt.RoomID)
+	if stillMember, err := h.hasRoomSubscription(ctx, memberEvt.RoomID); err != nil {
+		slog.WarnContext(ctx, "check remaining members failed; leaving ordering row",
+			"room_id", memberEvt.RoomID, "error", err)
+	} else if !stillMember {
+		if err := h.store.DeleteRemoteRoomActivity(ctx, memberEvt.RoomID); err != nil {
+			slog.WarnContext(ctx, "delete remote room activity failed",
+				"room_id", memberEvt.RoomID, "error", err)
+		}
 	}
 	// Scrub the removed accounts' thread read-state on this site too (#308).
 	if err := h.store.DeleteThreadSubscriptions(ctx, memberEvt.RoomID, memberEvt.Accounts); err != nil {
--- a/inbox-worker/main.go
+++ b/inbox-worker/main.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/caarlos0/env/v11"
 	o11yredis "github.com/flywindy/o11y/redis"
+	"github.com/nats-io/nats.go"
 	"github.com/nats-io/nats.go/jetstream"
 	"github.com/redis/go-redis/v9"
 	"go.mongodb.org/mongo-driver/v2/bson"
@@ -31,18 +32,23 @@ import (
 )
 
 type config struct {
-	NatsURL       string                  `env:"NATS_URL"        envDefault:"nats://localhost:4222"`
-	NatsCredsFile string                  `env:"NATS_CREDS_FILE" envDefault:""`
-	SiteID        string                  `env:"SITE_ID"         envDefault:"default"`
-	MongoURI      string                  `env:"MONGO_URI"       envDefault:"mongodb://localhost:27017"`
-	MongoDB       string                  `env:"MONGO_DB"        envDefault:"chat"`
-	MongoUsername string                  `env:"MONGO_USERNAME"  envDefault:""`
-	MongoPassword string                  `env:"MONGO_PASSWORD"  envDefault:""`
-	MaxWorkers    int                     `env:"MAX_WORKERS"     envDefault:"100"`
-	Consumer      stream.ConsumerSettings `envPrefix:"CONSUMER_"`
-	Bootstrap     bootstrapConfig         `envPrefix:"BOOTSTRAP_"`
-	HealthAddr    string                  `env:"HEALTH_ADDR" envDefault:":8081"`
-	PProfEnabled  bool                    `env:"PPROF_ENABLED" envDefault:"false"`
+	NatsURL       string `env:"NATS_URL"        envDefault:"nats://localhost:4222"`
+	NatsCredsFile string `env:"NATS_CREDS_FILE" envDefault:""`
+	SiteID        string `env:"SITE_ID"         envDefault:"default"`
+	MongoURI      string `env:"MONGO_URI"       envDefault:"mongodb://localhost:27017"`
+	MongoDB       string `env:"MONGO_DB"        envDefault:"chat"`
+	MongoUsername string `env:"MONGO_USERNAME"  envDefault:""`
+	MongoPassword string `env:"MONGO_PASSWORD"  envDefault:""`
+	MaxWorkers    int    `env:"MAX_WORKERS"     envDefault:"100"`
+	// RoomSubCache memoizes the room-membership check on the activity-refresh
+	// lane, which would otherwise cost a Mongo read per broadcast message.
+	// Either value non-positive disables it.
+	RoomSubCacheSize int                     `env:"ROOM_SUB_CACHE_SIZE" envDefault:"50000"`
+	RoomSubCacheTTL  time.Duration           `env:"ROOM_SUB_CACHE_TTL"  envDefault:"5m"`
+	Consumer         stream.ConsumerSettings `envPrefix:"CONSUMER_"`
+	Bootstrap        bootstrapConfig         `envPrefix:"BOOTSTRAP_"`
+	HealthAddr       string                  `env:"HEALTH_ADDR" envDefault:":8081"`
+	PProfEnabled     bool                    `env:"PPROF_ENABLED" envDefault:"false"`
 	// AdminAcctPrefix overrides the platform-admin account prefix (ADMIN_ACCT_PREFIX); keep it identical across services.
 	AdminAcctPrefix string `env:"ADMIN_ACCT_PREFIX" envDefault:"p_admin"`
 	// ValkeyAddrs seeds the Valkey cluster backing the badge cache
@@ -56,10 +62,67 @@ type config struct {
 
 // mongoInboxStore implements InboxStore using MongoDB.
 type mongoInboxStore struct {
-	subCol       *mongo.Collection
-	roomCol      *mongo.Collection
-	userCol      *mongo.Collection
-	threadSubCol *mongo.Collection
+	subCol        *mongo.Collection
+	roomCol       *mongo.Collection
+	userCol       *mongo.Collection
+	threadSubCol  *mongo.Collection
+	remoteRoomCol *mongo.Collection
+}
+
+// remoteRoomsCollection stores model.RemoteRoom. Deliberately NOT rooms: that
+// one is room-service's and its readers assume complete documents (members,
+// encKey, counts), which a federated peer cannot supply. Migration's room_sync
+// does replicate whole rooms here, so a migrated room can hold both and a
+// reader spanning the two must prefer rooms.
+const remoteRoomsCollection = "remote_rooms"
+
+// HasRoomSubscription reports whether any subscription for roomID exists here.
+// Served by the (roomId, u.account) unique index as a prefix scan; projected to
+// _id and capped at one doc since only existence matters.
+func (s *mongoInboxStore) HasRoomSubscription(ctx context.Context, roomID string) (bool, error) {
+	err := s.subCol.FindOne(ctx, bson.M{"roomId": roomID},
+		options.FindOne().SetProjection(bson.M{"_id": 1})).Err()
+	switch {
+	case err == nil:
+		return true, nil
+	case errors.Is(err, mongo.ErrNoDocuments):
+		return false, nil
+	default:
+		return false, fmt.Errorf("check room subscription %q: %w", roomID, err)
+	}
+}
+
+// DeleteRemoteRoomActivity drops a remote room's ordering row. Absent row is
+// not an error — the room may never have had one.
+func (s *mongoInboxStore) DeleteRemoteRoomActivity(ctx context.Context, roomID string) error {
+	if _, err := s.remoteRoomCol.DeleteOne(ctx, bson.M{"_id": roomID}); err != nil {
+		return fmt.Errorf("delete remote room activity %q: %w", roomID, err)
+	}
+	return nil
+}
+
+// UpsertRemoteRoomActivity advances a remote room's position under $max, so
+// out-of-order delivery can never regress it. $setOnInsert carries the
+// immutable identity so a first touch creates the row complete.
+func (s *mongoInboxStore) UpsertRemoteRoomActivity(ctx context.Context, roomID, siteID string, lastMsgAt time.Time) error {
+	filter := bson.M{"_id": roomID}
+	update := bson.M{
+		"$max":         bson.M{"lastMsgAt": lastMsgAt},
+		"$setOnInsert": bson.M{"siteId": siteID},
+	}
+	opts := options.UpdateOne().SetUpsert(true)
+	if _, err := s.remoteRoomCol.UpdateOne(ctx, filter, update, opts); err != nil {
+		if !mongo.IsDuplicateKeyError(err) {
+			return fmt.Errorf("upsert remote room activity %q: %w", roomID, err)
+		}
+		// Two concurrent first-touches. Unlike UpsertRoom, whose guard lives in
+		// the filter, this one filters on _id alone — so a duplicate says
+		// nothing about which value is newer. Retry; the row exists now.
+		if _, err := s.remoteRoomCol.UpdateOne(ctx, filter, update, opts); err != nil {
+			return fmt.Errorf("upsert remote room activity %q (retry after duplicate key): %w", roomID, err)
+		}
+	}
+	return nil
 }
 
 func (s *mongoInboxStore) CreateSubscription(ctx context.Context, sub *model.Subscription) error {
@@ -681,10 +744,11 @@ func main() {
 	}
 	db := mongoClient.Database(cfg.MongoDB)
 	store := &mongoInboxStore{
-		subCol:       db.Collection("subscriptions"),
-		roomCol:      db.Collection("rooms"),
-		userCol:      db.Collection("users"),
-		threadSubCol: db.Collection("thread_subscriptions"),
+		subCol:        db.Collection("subscriptions"),
+		roomCol:       db.Collection("rooms"),
+		userCol:       db.Collection("users"),
+		threadSubCol:  db.Collection("thread_subscriptions"),
+		remoteRoomCol: db.Collection(remoteRoomsCollection),
 	}
 	store.ensureIndexes(ctx)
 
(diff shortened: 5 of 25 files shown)


julianshen:
**11 findings · 8 unique issues · 8 inline comments posted** — 🔴 7 blocking · 🟠 4 warning

### Related work

- [Issue #39](https://github.com/hmchangw/newchat/issues/39) — \[Fix\] Room-key cache TTL-safety validated against drift-prone per-service grace config (open)
- [PR #197](https://github.com/hmchangw/newchat/pull/197) — perf\(user-service\): stop reading every room to serve one subscription.list page (open)
- [Issue #34](https://github.com/hmchangw/newchat/issues/34) — \[Fix\] messageRead should publish a subscription.update \(action "read"\) to the reader (closed)

### Local review memory

- `inbox-worker/handler.go:36` — @mliu33: This only updates lastMsgAt when there is a membership change. What is the plan ? How do we update this value on every new msg for remote site ? ([thread](https://github.com/hmchangw/newchat/pull/279#discussion_r3793155972))

<details>
<summary>📒 Files reviewed (25)</summary>

* `bot-room-service/handler_test.go`
* `bot-room-service/store.go`
* `bot-room-service/store_mongo.go`
* `broadcast-worker/deploy/bot/docker-compose.yml`
* `broadcast-worker/deploy/user/docker-compose.yml`
* `broadcast-worker/main.go`
* `broadcast-worker/roomactivity.go`
* `broadcast-worker/roomactivity_test.go`
* `docs/client-api.md`
* `docs/client-api/events.md`
* `inbox-worker/handler.go`
* `inbox-worker/main.go`
* `inbox-worker/roomsubcache.go`
* `inbox-worker/roomsubcache_test.go`
* `pkg/model/event.go`
* `pkg/model/model_test.go`
* `pkg/model/room.go`
* `pkg/subject/subject.go`
* `room-worker/handler.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

inbox-worker/handler.go
```diff
@@ -227,6 +233,17 @@ func (h *Handler) handleMemberAdded(ctx context.Context, evt *model.InboxEvent)
 		return fmt.Errorf("member_added references unknown users %v in room %s", missing, event.RoomID)
 	}
 
+	// After the missing-user return, so a room that gained no subscriber leaves
+	// no orphan row (nothing deletes them). Best-effort — the subscriptions above
+	// must not be re-run because an ordering row failed — but not self-healing
+	// yet, so alert on this log line.
+	if event.LastMsgAt != nil {
+		if err := h.store.UpsertRemoteRoomActivity(ctx, event.RoomID, event.SiteID, time.UnixMilli(*event.LastMsgAt).UTC()); err != nil {
```
julianshen, line 322: _🏷️ failure-handling_ | _🔴 Blocking_ | _`distributed-system`, `mongodb`, `nats`, `performance-and-concurrency`, `silent-failure`, `tgd-review` (6 rules)_

**A transient activity write failure is permanently acknowledged.**

After subscriptions are committed, a transient `remoteRoomCol.UpdateOne` failure reaches this branch. It only logs the failure and returns success, so the inbox delivery is ACKed. The new interface comment explicitly says `member_added` is currently the only writer and the loss is not self-healing; consequently the remote room has no ordering row indefinitely. `BulkCreateSubscriptions` is already an idempotent upsert, so redelivery can safely repeat it. Return the error to trigger retry, or durably enqueue the activity write before acknowledging the event.

<details>

- **`mongodb`**:241: When UpsertRemoteRoomActivity fails, the handler only logs the error and then successfully acknowledges the member_added event. The new interface comment explicitly states that member_added is currently the only writer and that a dropped write is not self-healing, so a timeout or primary outage can leave committed subscriptions with no remote_rooms ordering row indefinitely. Propagate the error so the event is redelivered through the existing idempotent subscription upsert path, or persist both effects through a transaction/durable retry mechanism.
- **`nats`**:233: For a cross-site `member_added`, subscription creation can succeed and `UpsertRemoteRoomActivity` can then fail transiently. The handler logs the error and returns success, so the NATS message is acknowledged even though the comment states this is currently the only writer and is not self-healing. The destination then permanently lacks the remote room's ordering position. Propagate the error so the message is retried; `BulkCreateSubscriptions` is explicitly idempotent, making redelivery safe.
- **`performance-and-concurrency`**:238: If subscription creation consumes most of the event deadline, this write can receive an expired context, fail, and then be swallowed. The handler consequently acknowledges the event despite never creating the ordering row. The new interface comment explicitly states that no self-healing publisher exists yet, so latency spikes or transient Mongo failures can leave newly joined remote rooms without an activity position indefinitely. Persist this write transactionally or through an outbox, or make the whole handler safely retryable and return the error.
- **`silent-failure`**:238: If UpsertRemoteRoomActivity fails, the handler only logs the error and returns success, so the inbox event can be acknowledged without persisting the remote room's ordering position. The interface comment confirms member_added is currently the only writer and the loss is not self-healing, leaving the room permanently absent from or incorrectly positioned in the remote chat list. Propagate the error so the event retries, or durably enqueue a separate retry before returning success; the subscription upsert path should remain idempotent.
- **`tgd-review`**:236: The handler logs this failure and acknowledges the event, even though the interface comment states that no activity-refresh publisher exists yet. Consequently, a transient MongoDB failure permanently leaves this remote room without an ordering position. Subscription creation is already tested as idempotent, and the activity upsert is guarded by $max, so returning the error allows safe redelivery of both writes.

</details>

<details>

```
In `inbox-worker/handler.go` around line 241: After subscriptions are committed, a transient `remoteRoomCol.UpdateOne` failure reaches this branch. It only logs the failure and returns success, so the inbox delivery is ACKed. The new interface comment explicitly says `member_added` is currently the only writer and the loss is not self-healing; consequently the remote room has no ordering row indefinitely. `BulkCreateSubscriptions` is already an idempotent upsert, so redelivery can safely repeat it. Return the error to trigger retry, or durably enqueue the activity write before acknowledging the event.

```

</details>

room-worker/handler.go
```diff
@@ -1271,6 +1275,7 @@ func (h *Handler) processAddMembers(ctx context.Context, data []byte) (err error
 			RequesterAccount:   req.RequesterAccount,
 			JoinedAt:           req.Timestamp,
 			HistorySharedSince: historySharedSince,
+			LastMsgAt:          timeutil.TimeToMillis(evtRoom.LastMsgAt),
```
julianshen, line 1309: _🏷️ failure-handling_ | _🔴 Blocking_ | _`distributed-system`, `mongodb`, `silent-failure` (3 rules)_

**A failed post-commit room read permanently drops the activity position.**

For a cross-site add to a room with history, subscriptions can commit and then the post-add `GetRoom` can fail transiently. `evtRoom` then remains the cache/meta projection, which the added comment confirms has no `LastMsgAt`; this line serializes nil and omits the field. The destination creates the subscription but performs no activity upsert, and there is currently no refresh writer to repair the missing row. Fetch and retain the full room activity position before committing subscriptions, then optionally re-read after commit for a fresher value, or durably stage the complete federation event with the subscription mutation. Simply returning the post-commit read error is insufficient because the existing comment notes redelivery recomputes no new subscriptions and drops the event.

<details>

- **`silent-failure`**:1278: The full-room reread failure is tolerated after subscriptions are committed, but the fallback meta view has no LastMsgAt. The federated event therefore serializes without lastMsgAt and still completes successfully. The destination then skips its only current opportunity to seed remote-room activity, with no retry or repair. Fetch the required activity position before the irreversible subscription writes, where a read failure can be returned safely, or persist the federation work durably so it can retry without depending on recomputed membership changes.
- **`mongodb`**:1278: LastMsgAt comes from evtRoom, but the full-room read only runs when len(subs) > 0. A pure org-to-individual upgrade emits member_added without creating a subscription, so evtRoom remains the metadata projection that the added test confirms has no LastMsgAt. A failed full-room re-read has the same result. The destination consequently skips the remote_rooms upsert even when the room has messages. Read the full room for every path that emits a cross-site member_added, and treat read failure as retryable or provide a durable backfill.

</details>

<details>

```
In `room-worker/handler.go` around line 1278: For a cross-site add to a room with history, subscriptions can commit and then the post-add `GetRoom` can fail transiently. `evtRoom` then remains the cache/meta projection, which the added comment confirms has no `LastMsgAt`; this line serializes nil and omits the field. The destination creates the subscription but performs no activity upsert, and there is currently no refresh writer to repair the missing row. Fetch and retain the full room activity position before committing subscriptions, then optionally re-read after commit for a fresher value, or durably stage the complete federation event with the subscription mutation. Simply returning the post-commit read error is insufficient because the existing comment notes redelivery recomputes no new subscriptions and drops the event.

```

</details>

room-worker/handler.go
```diff
@@ -1130,8 +1131,11 @@ func (h *Handler) processAddMembers(ctx context.Context, data []byte) (err error
 	// subs are already committed, so failing here would let a redelivery
 	// (which recomputes needSub as empty) silently drop the events — degrade
 	// to the pre-add meta view instead and let subscription.list reconcile.
+	// Post-add view: the meta view unless the re-read succeeds. Also supplies the
+	// cross-site MemberAddEvent's activity position below — the meta view carries
+	// no lastMsgAt, so a failed re-read ships without one.
+	evtRoom := room
```
julianshen, line 1169: _🏷️ NATS reliability_ | _🔴 Blocking_ | _`nats`_

**A failed room reread emits an unrecoverably incomplete event.**

When adding the first remote member to an active room, subscriptions are committed before the full-room reread. If that reread fails transiently, `evtRoom` remains the metadata view, which the added comment and test confirm cannot contain `LastMsgAt`. The cross-site `member_added` is then published without the activity position, the inbox consumer skips its only current activity write, and successful processing acknowledges the event with no self-healing path. Obtain the required full-room snapshot before committing subscriptions and fail before that write if it cannot be read, or persist the member event and activity position transactionally for retry.

<details>

```
In `room-worker/handler.go` around line 1137: When adding the first remote member to an active room, subscriptions are committed before the full-room reread. If that reread fails transiently, `evtRoom` remains the metadata view, which the added comment and test confirm cannot contain `LastMsgAt`. The cross-site `member_added` is then published without the activity position, the inbox consumer skips its only current activity write, and successful processing acknowledges the event with no self-healing path. Obtain the required full-room snapshot before committing subscriptions and fail before that write if it cannot be read, or persist the member event and activity position transactionally for retry.

```

</details>

inbox-worker/main.go
```diff
@@ -46,10 +46,42 @@ type config struct {
...
+}
+
+// remoteRoomsCollection stores model.RemoteRoom. Deliberately NOT rooms: that
+// one is room-service's and its readers assume complete documents (members,
+// encKey, counts), which a federated peer cannot supply. Migration's room_sync
+// does replicate whole rooms here, so a migrated room can hold both and a
+// reader spanning the two must prefer rooms.
+const remoteRoomsCollection = "remote_rooms"
+
+// UpsertRemoteRoomActivity advances a remote room's position under $max, so
```
julianshen, line 104: _🏷️ resource-growth_ | _🟠 Warning_ | _`mongodb`, `performance-and-concurrency` (2 rules)_

**Remote-room activity rows have no cleanup path**

Each first remote membership permanently upserts a row into `remote_rooms`, but the new API provides only an upsert and the handler comment confirms that nothing deletes orphan rows. When users join and later leave many remote rooms, collection and index cardinality therefore track every room ever seen rather than currently subscribed rooms, causing unbounded disk, index, and backup growth. Add race-safe cleanup when the last local subscription is removed or the room is deleted, or run periodic reconciliation against active subscriptions.

<details>

- **`mongodb`**:66: UpsertRemoteRoomActivity creates one remote_rooms document per remote room, while the handler comment confirms that nothing deletes these rows. After the site's final subscription to a remote room is removed, its activity document remains indefinitely. Repeated remote-room join/leave or room-creation/deletion churn therefore accumulates orphan documents and indexes. Add race-safe cleanup when the last local subscription is removed, or a reconciliation process that deletes rows with no corresponding subscriptions.

</details>

<details>

```
In `inbox-worker/main.go` around line 63: Each first remote membership permanently upserts a row into `remote_rooms`, but the new API provides only an upsert and the handler comment confirms that nothing deletes orphan rows. When users join and later leave many remote rooms, collection and index cardinality therefore track every room ever seen rather than currently subscribed rooms, causing unbounded disk, index, and backup growth. Add race-safe cleanup when the last local subscription is removed or the room is deleted, or run periodic reconciliation against active subscriptions.

```

</details>

room-worker/handler.go
julianshen, line 1169: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Some member_added events omit existing room activity.**

The fallback `room` is the metadata-cache view, which the accompanying test explicitly confirms has no LastMsgAt. Therefore a failed full-room read publishes nil for an active room. The read is also guarded by `len(subs) > 0`, so an emitted event without a newly created subscription, such as the documented org-to-individual upgrade, never attempts to load LastMsgAt. The destination interprets nil as no activity and skips seeding, with no self-healing publisher currently present. Obtain the activity position before the committed write or otherwise make omission retryable/self-healing.

<details>

```
In `room-worker/handler.go` around line 1137: The fallback `room` is the metadata-cache view, which the accompanying test explicitly confirms has no LastMsgAt. Therefore a failed full-room read publishes nil for an active room. The read is also guarded by `len(subs) > 0`, so an emitted event without a newly created subscription, such as the documented org-to-individual upgrade, never attempts to load LastMsgAt. The destination interprets nil as no activity and skips seeding, with no self-healing publisher currently present. Obtain the activity position before the committed write or otherwise make omission retryable/self-healing.

```

</details>

julianshen (commented):
tGD inline review

inbox-worker/handler.go
```diff
@@ -23,6 +23,12 @@ type InboxStore interface {
 	// stored one is a silent no-op, so out-of-order federated delivery cannot
 	// regress room metadata.
 	UpsertRoom(ctx context.Context, room *model.Room) error
+	// UpsertRemoteRoomActivity records a remote room's ordering position under a
+	// $max guard, so a late or duplicate event is a no-op rather than a
+	// regression — what will let the activity-refresh publisher ride a lossy,
+	// unordered transport. Until it lands member_added is the only writer, so a
+	// dropped write is NOT self-healing.
+	UpsertRemoteRoomActivity(ctx context.Context, roomID, siteID string, lastMsgAt time.Time) error
```
mliu33, line 36: This only updates lastMsgAt when there is a membership change. What is the plan ? How do we update this value on every new msg for remote site ?

broadcast-worker/coalescer.go
```diff
@@ -85,9 +116,71 @@ func (c *coalescingStore) Flush(ctx context.Context) error {
...
+	for roomID, u := range batch {
+		if !c.crossSite(ctx, roomID) {
+			continue
+		}
+		if c.throttled(roomID, now) {
+			continue
+		}
+		if err := c.publishActivity(ctx, roomActivityRefresh{roomID: roomID, at: u.at}); err != nil {
+			slog.WarnContext(ctx, "publish room activity refresh failed", "room_id", roomID, "error", err)
+		}
```
julianshen, line 152: _🏷️ coalescing_ | _🔴 Blocking_ | _`distributed-system`, `performance-and-concurrency` (2 rules)_

**The throttle permanently discards the latest room position.**

The leading-edge throttle drops updates instead of retaining them for a trailing publish. For example, a room publishes at t=0, receives its final message at t=4.9s, and is skipped because it is still throttled. If the room then goes quiet, later empty Flush calls return before refreshRemoteActivity, so that final position is never announced. The destination remains stale indefinitely, not merely for one interval. Retain the latest suppressed refresh per room and publish it when the watermark expires; keep this pending state bounded by active throttled rooms.

<details>

- **`distributed-system`**:145: A room can publish position A, receive position B during the five-second throttle window, and then go quiet. B is skipped here and is not retained for a trailing publish, so the destination remains at A indefinitely rather than being stale by at most one interval. The same happens after a partial publish failure because the watermark is recorded before publication succeeds. Retain each room's latest unsent position and publish it when the interval expires, tracking failures per peer, or move these updates to a durable compacted/retryable lane.

</details>

<details>

```
In `broadcast-worker/coalescer.go` around line 152: The leading-edge throttle drops updates instead of retaining them for a trailing publish. For example, a room publishes at t=0, receives its final message at t=4.9s, and is skipped because it is still throttled. If the room then goes quiet, later empty Flush calls return before refreshRemoteActivity, so that final position is never announced. The destination remains stale indefinitely, not merely for one interval. Retain the latest suppressed refresh per room and publish it when the watermark expires; keep this pending state bounded by active throttled rooms.

```

</details>

bot-room-service/handler.go
```diff
@@ -126,8 +127,19 @@ func (h *handler) handleDMEnsure(c *natsrouter.Context, req BotDMEnsureRequest)
...
+	if err := h.store.InsertRoom(c, room); err != nil {
+		if !errors.Is(err, ErrDuplicate) {
+			return nil, fmt.Errorf("insert dm room: %w", err)
+		}
+		// The DM id is deterministic, so a duplicate means this room already
+		// exists and may carry history — take its position rather than federating
+		// nil and leaving the target unable to order it.
+		if existing, findErr := h.store.FindRoom(c, roomID); findErr == nil {
+			room.LastMsgAt = existing.LastMsgAt
+		} else {
```
julianshen, line 139: _🏷️ failure-handling_ | _🔴 Blocking_ | _`distributed-system`, `silent-failure`, `tgd-review` (3 rules)_

**A failed duplicate-room read is converted into a successful ensure.**

After InsertRoom reports a duplicate, a transient FindRoom failure is only logged. The handler then creates the subscription and federates member_added without LastMsgAt, while still returning success. Because the existing DM may remain quiet, no later activity refresh is guaranteed to repair the missing ordering position, leaving the target unable to order its historical room. Return the read error before creating the subscription so the request can retry safely.

<details>

- **`distributed-system`**:139: A retry after a prior partial DM creation can encounter ErrDuplicate while the existing room already has history. If FindRoom then fails transiently, this branch continues, commits the subscription, and federates a nil LastMsgAt. With no subsequent message, the destination never receives a correcting refresh. This read occurs before the subscription write, so fail the request and let the caller retry.
- **`tgd-review`**:137: On a transient FindRoom failure, ensure succeeds and sends nil LastMsgAt even though the duplicate room may contain history. Its earlier activity refreshes occurred before the target subscription existed, so a quiet DM may never receive an ordering row. No mutation requiring best-effort handling has occurred yet; return the lookup error so the ensure request can retry.

</details>

<details>

```
In `bot-room-service/handler.go` around line 139: After InsertRoom reports a duplicate, a transient FindRoom failure is only logged. The handler then creates the subscription and federates member_added without LastMsgAt, while still returning success. Because the existing DM may remain quiet, no later activity refresh is guaranteed to repair the missing ordering position, leaving the target unable to order its historical room. Return the read error before creating the subscription so the request can retry safely.

```

</details>

julianshen (commented):
tGD inline review

inbox-worker/handler.go
```diff
@@ -272,6 +343,20 @@ func (h *Handler) handleMemberRemoved(ctx context.Context, evt *model.InboxEvent
...
+	// none do, the ordering row has no reader left here — drop it, or it becomes
+	// the orphan the seed path is careful not to create. Best-effort: a stale row
+	// only costs space, and re-adding a member re-seeds it.
+	h.roomSubs.invalidate(memberEvt.RoomID)
+	if stillMember, err := h.hasRoomSubscription(ctx, memberEvt.RoomID); err != nil {
+		slog.WarnContext(ctx, "check remaining members failed; leaving ordering row",
+			"room_id", memberEvt.RoomID, "error", err)
+	} else if !stillMember {
+		if err := h.store.DeleteRemoteRoomActivity(ctx, memberEvt.RoomID); err != nil {
+			slog.WarnContext(ctx, "delete remote room activity failed",
```
julianshen, line 383: _🎯 concurrency_ | _🔴 Blocking_ | _`distributed-system`_

**Activity cleanup races with concurrent refreshes and re-adds.**

The subscription check and remote-room deletion are not atomic with either activity upserts or membership creation. For example, removal can observe no subscriptions, then a concurrent re-add can create a subscription and seed remote_rooms, after which this delete removes the fresh seed. The final state has a valid subscription but no ordering row until another message arrives. Conversely, an activity handler that passed its membership check before this delete can upsert afterward and recreate an orphan. These lanes run concurrently across workers and replicas. Use a transactionally guarded delete, or delayed periodic reconciliation that rechecks membership and can also restore missing rows.

<details>

```
In `inbox-worker/handler.go` around line 356: The subscription check and remote-room deletion are not atomic with either activity upserts or membership creation. For example, removal can observe no subscriptions, then a concurrent re-add can create a subscription and seed remote_rooms, after which this delete removes the fresh seed. The final state has a valid subscription but no ordering row until another message arrives. Conversely, an activity handler that passed its membership check before this delete can upsert afterward and recreate an orphan. These lanes run concurrently across workers and replicas. Use a transactionally guarded delete, or delayed periodic reconciliation that rechecks membership and can also restore missing rows.

```

</details>

inbox-worker/handler.go
```diff
@@ -249,6 +308,18 @@ func (h *Handler) handleMemberAdded(ctx context.Context, evt *model.InboxEvent)
 		return fmt.Errorf("member_added references unknown users %v in room %s", missing, event.RoomID)
 	}
 
+	// After the missing-user return, so a room that gained no subscriber leaves
+	// no orphan row (nothing deletes them). Best-effort — the subscriptions above
+	// must not be re-run because an ordering row failed — but not self-healing
+	// yet, so alert on this log line.
+	h.roomSubs.set(event.RoomID, true)
```
julianshen, line 320: _🏷️ durability_ | _🔴 Blocking_ | _`mongodb`_

**A failed activity seed is permanently acknowledged.**

After subscriptions are committed, a transient remote_rooms update failure is only logged, so the JetStream member_added event is acknowledged without its ordering row. The new activity lane repairs this only if the room receives another message; an existing room that remains quiet stays unpositioned indefinitely. BulkCreateSubscriptions is an idempotent upsert and this write is $max-guarded, so return the error and let the event retry.

<details>

```
In `inbox-worker/handler.go` around line 315: After subscriptions are committed, a transient remote_rooms update failure is only logged, so the JetStream member_added event is acknowledged without its ordering row. The new activity lane repairs this only if the room receives another message; an existing room that remains quiet stays unpositioned indefinitely. BulkCreateSubscriptions is an idempotent upsert and this write is $max-guarded, so return the error and let the event retry.

```

</details>

inbox-worker/roomsubcache.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/hmchangw/chat/pkg/cachemetrics"
+)
+
+// roomSubCache answers "does this site hold a subscription for this room" from
+// memory. The activity refresh is broadcast to every peer, so without it each
+// arriving message costs a Mongo read just to discard most of them — and the
+// answer changes only on member_added/member_removed, which are rare against
+// message traffic, so the hit rate approaches 1 after warmup.
+//
+// Both polarities are cached. Negative entries are the point: rooms this site
```
julianshen, line 19: _🏷️ consistency_ | _🟠 Warning_ | _`mongodb`_

**Process-local membership caching is incoherent across queue replicas.**

The cache assumes member_added/member_removed handlers keep entries current in the same process, but the activity subscription is a queue group and deployments can have multiple inbox-worker instances. If instance A handles member_added while instance B holds a cached negative, activity messages routed to B are discarded for up to the default five-minute TTL. Stale positives can likewise recreate remote_rooms rows after the final subscription is removed. Disable this cache when multiple replicas run, or move membership state/invalidation to a shared cache with atomic updates.

<details>

```
In `inbox-worker/roomsubcache.go` around line 19: The cache assumes member_added/member_removed handlers keep entries current in the same process, but the activity subscription is a queue group and deployments can have multiple inbox-worker instances. If instance A handles member_added while instance B holds a cached negative, activity messages routed to B are discarded for up to the default five-minute TTL. Stale positives can likewise recreate remote_rooms rows after the final subscription is removed. Disable this cache when multiple replicas run, or move membership state/invalidation to a shared cache with atomic updates.

```

</details>

inbox-worker/handler.go
```diff
@@ -272,6 +343,20 @@ func (h *Handler) handleMemberRemoved(ctx context.Context, evt *model.InboxEvent
...
+	// Other members may remain, so re-resolve rather than caching a guess. When
+	// none do, the ordering row has no reader left here — drop it, or it becomes
+	// the orphan the seed path is careful not to create. Best-effort: a stale row
+	// only costs space, and re-adding a member re-seeds it.
+	h.roomSubs.invalidate(memberEvt.RoomID)
+	if stillMember, err := h.hasRoomSubscription(ctx, memberEvt.RoomID); err != nil {
+		slog.WarnContext(ctx, "check remaining members failed; leaving ordering row",
+			"room_id", memberEvt.RoomID, "error", err)
+	} else if !stillMember {
+		if err := h.store.DeleteRemoteRoomActivity(ctx, memberEvt.RoomID); err != nil {
```
julianshen, line 382: _🏷️ atomicity_ | _🟠 Warning_ | _`mongodb`_

**Subscription checks and activity cleanup race across handlers.**

DeleteSubscriptionsByAccounts, HasRoomSubscription, and DeleteRemoteRoomActivity are separate Mongo operations, while HandleRoomActivity separately checks membership before upserting. A refresh can observe a subscription, then the removal path can delete the subscription and activity row, after which the refresh recreates an orphan row. Concurrent add/remove processing can also delete a newly seeded row. Use a shared per-room serialization/guard that works across replicas, or add durable reconciliation that removes rows without subscriptions and restores rows for active subscriptions.

<details>

```
In `inbox-worker/handler.go` around line 355: DeleteSubscriptionsByAccounts, HasRoomSubscription, and DeleteRemoteRoomActivity are separate Mongo operations, while HandleRoomActivity separately checks membership before upserting. A refresh can observe a subscription, then the removal path can delete the subscription and activity row, after which the refresh recreates an orphan row. Concurrent add/remove processing can also delete a newly seeded row. Use a shared per-room serialization/guard that works across replicas, or add durable reconciliation that removes rows without subscriptions and restores rows for active subscriptions.

```

</details>

inbox-worker/handler.go
```diff
@@ -119,11 +133,56 @@ type Handler struct {
...
+		return fmt.Errorf("unmarshal room activity event: %w", err)
+	}
+	if evt.RoomID == "" || evt.LastMsgAt <= 0 {
+		return fmt.Errorf("room activity event missing roomId or lastMsgAt")
+	}
+	subscribed, err := h.hasRoomSubscription(ctx, evt.RoomID)
+	if err != nil {
+		return fmt.Errorf("check room subscription: %w", err)
+	}
+	if !subscribed {
```
julianshen, line 182: _🏷️ ordering_ | _🟠 Warning_ | _`nats`_

**Cross-lane reordering can permanently discard the newest position.**

Room activity uses core NATS while member_added uses JetStream, so there is no ordering between them. A message sent immediately after a remote member is added can have its activity refresh arrive first; HasRoomSubscription then returns false and this branch discards the newer position. The later member_added only seeds the older add-time position, and if the room becomes quiet it is never repaired. Preserve the latest pre-membership refresh in a bounded pending cache and merge it on member_added, or carry both signals through an ordered/persistent lane.

<details>

```
In `inbox-worker/handler.go` around line 179: Room activity uses core NATS while member_added uses JetStream, so there is no ordering between them. A message sent immediately after a remote member is added can have its activity refresh arrive first; HasRoomSubscription then returns false and this branch discards the newer position. The later member_added only seeds the older add-time position, and if the room becomes quiet it is never repaired. Preserve the latest pre-membership refresh in a bounded pending cache and merge it on member_added, or carry both signals through an ordered/persistent lane.

```

</details>

inbox-worker/handler.go
julianshen, line 382: _🏷️ nats-reliability_ | _🟠 Warning_ | _`nats`_

**A failed activity cleanup is ACKed and leaves a permanent row.**

If deleting remote_rooms fails after the last local subscription is removed, the error is logged and member_removed completes successfully, acknowledging the JetStream event. With no remaining member there may never be another removal to retry cleanup, so the orphan persists indefinitely. Return the error; subscription deletion and deleting an absent activity row are both safe to repeat.

<details>

```
In `inbox-worker/handler.go` around line 355: If deleting remote_rooms fails after the last local subscription is removed, the error is logged and member_removed completes successfully, acknowledging the JetStream event. With no remaining member there may never be another removal to retry cleanup, so the orphan persists indefinitely. Return the error; subscription deletion and deleting an absent activity row are both safe to repeat.

```

</details>

inbox-worker/main.go
```diff
@@ -849,6 +931,8 @@ func main() {
 				return fmt.Errorf("worker drain timed out: %w", ctx.Err())
 			}
 		},
+		// Unsubscribe before the drain so no refresh arrives after the store closes.
+		func(_ context.Context) error { return activitySub.Unsubscribe() },
```
julianshen, line 928: _🏷️ lifecycle_ | _🟠 Warning_ | _`nats`_

**Unsubscribe can outlive an in-flight activity callback.**

Subscription.Unsubscribe stops new delivery but does not wait for an asynchronous callback already executing. Because the subscription is removed before the subsequent connection drain, that drain may no longer wait for its callback; Mongo can then be disconnected while HandleRoomActivity is still writing, losing the final refresh. Leave the subscription registered and let natsutil.Drain drain all subscriptions and await callbacks, or explicitly drain and await this subscription before closing Mongo.

<details>

```
In `inbox-worker/main.go` around line 935: Subscription.Unsubscribe stops new delivery but does not wait for an asynchronous callback already executing. Because the subscription is removed before the subsequent connection drain, that drain may no longer wait for its callback; Mongo can then be disconnected while HandleRoomActivity is still writing, losing the final refresh. Leave the subscription registered and let natsutil.Drain drain all subscriptions and await callbacks, or explicitly drain and await this subscription before closing Mongo.

```

</details>

broadcast-worker/coalescer.go
```diff
@@ -85,9 +116,71 @@ func (c *coalescingStore) Flush(ctx context.Context) error {
 	batch := c.pending
 	c.pending = make(map[string]roomLastMsgUpdate, len(batch))
 	c.mu.Unlock()
+	c.refreshRemoteActivity(ctx, batch)
```
julianshen, line 119: _⚡ memory-exhaustion_ | _🟠 Warning_ | _`oom`_

**Synchronous refresh fan-out can make the pending map grow without bound.**

Each flush now synchronously scans every distinct room in `batch`, performs a cross-site lookup, and publishes once per peer before returning to the periodic flush loop. Under sustained traffic across many rooms, especially with cache misses or many peers, this work can take longer than the 250ms cadence. `UpdateRoomLastMessage` continues accepting new room IDs into the replacement `pending` map without a size limit or backpressure, so successive batches grow, take still longer to refresh, and retain both the old batch and the growing pending map concurrently. This reachable overload loop can cause GC thrashing or process OOM. Keep activity refresh work outside the critical flush cadence using a bounded queue/coalescing map with dropping or replacement, or impose a hard per-flush room/time budget; the decorative refresh must not delay draining the unbounded pending map.

<details>

```
In `broadcast-worker/coalescer.go` around line 119: Each flush now synchronously scans every distinct room in `batch`, performs a cross-site lookup, and publishes once per peer before returning to the periodic flush loop. Under sustained traffic across many rooms, especially with cache misses or many peers, this work can take longer than the 250ms cadence. `UpdateRoomLastMessage` continues accepting new room IDs into the replacement `pending` map without a size limit or backpressure, so successive batches grow, take still longer to refresh, and retain both the old batch and the growing pending map concurrently. This reachable overload loop can cause GC thrashing or process OOM. Keep activity refresh work outside the critical flush cadence using a bounded queue/coalescing map with dropping or replacement, or impose a hard per-flush room/time budget; the decorative refresh must not delay draining the unbounded pending map.

```

</details>

broadcast-worker/coalescer.go
```diff
@@ -85,9 +116,71 @@ func (c *coalescingStore) Flush(ctx context.Context) error {
...
 	return c.bulk.BulkUpdateRoomLastMessage(ctx, batch)
 }
 
+// refreshRemoteActivity emits at most one activity refresh per cross-site room
+// per refreshInterval. Cost scales with distinct active cross-site rooms per
+// interval, not with message rate — coalescing collapses the window, and the
```
julianshen, line 125: _🏷️ backpressure_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Activity fan-out blocks the critical Mongo flush.**

Flush performs all activity work before BulkUpdateRoomLastMessage. A batch of N distinct rooms therefore executes N synchronous metadata lookups and up to N×peer-count publishes before persisting rooms.lastMsgAt. With cache churn or more active rooms than the 10,000-entry default cache, lookup misses can become serial Mongo reads; a large peer list further amplifies the publish loop. During that delay, message handlers continue adding distinct rooms to pending while the single flush loop cannot advance, causing memory growth and stale Mongo positions. Persist the bulk batch first, then hand activity updates to a separately bounded, coalescing worker with cancellation and explicit backpressure rather than blocking or spawning per-room goroutines.

<details>

```
In `broadcast-worker/coalescer.go` around line 125: Flush performs all activity work before BulkUpdateRoomLastMessage. A batch of N distinct rooms therefore executes N synchronous metadata lookups and up to N×peer-count publishes before persisting rooms.lastMsgAt. With cache churn or more active rooms than the 10,000-entry default cache, lookup misses can become serial Mongo reads; a large peer list further amplifies the publish loop. During that delay, message handlers continue adding distinct rooms to pending while the single flush loop cannot advance, causing memory growth and stale Mongo positions. Persist the bulk batch first, then hand activity updates to a separately bounded, coalescing worker with cancellation and explicit backpressure rather than blocking or spawning per-room goroutines.

```

</details>

inbox-worker/handler.go
```diff
@@ -119,11 +133,56 @@ type Handler struct {
...
+		return nil
+	}
+	if err := h.store.UpsertRemoteRoomActivity(ctx, evt.RoomID, evt.SiteID, time.UnixMilli(evt.LastMsgAt).UTC()); err != nil {
+		return fmt.Errorf("apply room activity: %w", err)
```
julianshen, line 186: _🎯 concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Membership cleanup races with activity upserts.**

The subscription existence check and remote-room upsert are separate operations on a core-NATS callback that runs concurrently with member_removed processing. An activity handler can observe a subscription, pause, let the final removal delete both the subscription and remote row, then resume and recreate an orphan row. No later removal cleans it, so this interleaving leaks remote_rooms documents and index entries under room churn. Conversely, removal can decide no subscription remains and then delete a row concurrently seeded by a new member_added, losing the position until another message arrives. The cache mutex does not cover these store mutations, and per-process locking would not protect multiple replicas. Make the membership decision and row mutation atomic through MongoDB coordination, and provide durable or periodic reconciliation so failed or raced cleanup is eventually retried.

<details>

```
In `inbox-worker/handler.go` around line 183: The subscription existence check and remote-room upsert are separate operations on a core-NATS callback that runs concurrently with member_removed processing. An activity handler can observe a subscription, pause, let the final removal delete both the subscription and remote row, then resume and recreate an orphan row. No later removal cleans it, so this interleaving leaks remote_rooms documents and index entries under room churn. Conversely, removal can decide no subscription remains and then delete a row concurrently seeded by a new member_added, losing the position until another message arrives. The cache mutex does not cover these store mutations, and per-process locking would not protect multiple replicas. Make the membership decision and row mutation atomic through MongoDB coordination, and provide durable or periodic reconciliation so failed or raced cleanup is eventually retried.

```

</details>

broadcast-worker/coalescer.go
julianshen, line 119: _🏷️ failure-handling_ | _🟠 Warning_ | _`silent-failure`_

**Activity is published before the authoritative room update succeeds.**

Flush sends remote activity before BulkUpdateRoomLastMessage. If that bulk write then fails, peers have already advanced their ordering position even though the origin's authoritative room projection did not, and the drained batch is no longer pending. This creates persistent cross-site ordering disagreement when the room receives no later message. Execute the bulk write first and call refreshRemoteActivity only after it succeeds; preserve or requeue the batch on failure if Run does not already retry it.

<details>

```
In `broadcast-worker/coalescer.go` around line 119: Flush sends remote activity before BulkUpdateRoomLastMessage. If that bulk write then fails, peers have already advanced their ordering position even though the origin's authoritative room projection did not, and the drained batch is no longer pending. This creates persistent cross-site ordering disagreement when the room receives no later message. Execute the bulk write first and call refreshRemoteActivity only after it succeeds; preserve or requeue the batch on failure if Run does not already retry it.

```

</details>

inbox-worker/roomsubcache.go
```diff
@@ -0,0 +1,110 @@
...
+
+// fill records a read-through result without overwriting an entry a membership
+// handler wrote while the read was in flight. Those are authoritative; a
+// concurrent read only ever observed a snapshot that may already be stale.
+func (c *roomSubCache) fill(roomID string, subscribed bool) {
+	if c == nil {
+		return
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
julianshen, line 79: _🎯 concurrency_ | _🟠 Warning_ | _`tgd-review`_

**An in-flight read can undo member-removal invalidation.**

The cache mutex is released during HasRoomSubscription, and fill only checks whether an entry currently exists. A refresh can therefore read true, then the last subscription can be deleted and the cache invalidated, after which the refresh fills the stale true value. The removal handler can hit that value and skip DeleteRemoteRoomActivity; cache expiry later does not delete the orphan row. Track a per-room generation across miss/fill and increment it on set/invalidate, or serialize the read-through with membership mutation and cleanup.

<details>

```
In `inbox-worker/roomsubcache.go` around line 79: The cache mutex is released during HasRoomSubscription, and fill only checks whether an entry currently exists. A refresh can therefore read true, then the last subscription can be deleted and the cache invalidated, after which the refresh fills the stale true value. The removal handler can hit that value and skip DeleteRemoteRoomActivity; cache expiry later does not delete the orphan row. Track a per-room generation across miss/fill and increment it on set/invalidate, or serialize the read-through with membership mutation and cleanup.

```

</details>

broadcast-worker/roomactivity.go
```diff
@@ -0,0 +1,81 @@
...
+			return false
+		}
+		return meta.CrossSite == nil || *meta.CrossSite
+	}
+}
+
+// roomActivityPublisher emits one refresh per peer site on the core-NATS
+// activity lane. One room per message keeps the payload bounded by
+// construction — a batched form would grow with the number of active rooms and
+// eventually exceed max_payload, silently losing a whole flush window.
```
julianshen, line 57: _🏷️ delivery_ | _🔴 Blocking_ | _`distributed-system`_

**Core NATS can permanently lose a room's final activity.**

Suppose the destination subscriber is restarting or the supercluster route is partitioned when the origin publishes a room's final position B. Core NATS has no persistence or end-to-end acknowledgement, so Publish may succeed locally while the destination receives nothing. If the room then remains quiet, no later flush repairs position A, leaving the remote chat list stale indefinitely. Use a durable per-destination stream or periodic anti-entropy that retries independently of subsequent room messages.

<details>

```
In `broadcast-worker/roomactivity.go` around line 57: Suppose the destination subscriber is restarting or the supercluster route is partitioned when the origin publishes a room's final position B. Core NATS has no persistence or end-to-end acknowledgement, so Publish may succeed locally while the destination receives nothing. If the room then remains quiet, no later flush repairs position A, leaving the remote chat list stale indefinitely. Use a durable per-destination stream or periodic anti-entropy that retries independently of subsequent room messages.

```

</details>

broadcast-worker/coalescer.go
```diff
@@ -85,9 +116,71 @@ func (c *coalescingStore) Flush(ctx context.Context) error {
...
+// message re-establishes it, and the room batch this rides with is the write
+// that actually matters.
+func (c *coalescingStore) refreshRemoteActivity(ctx context.Context, batch map[string]roomLastMsgUpdate) {
+	if c.publishActivity == nil || c.crossSite == nil {
+		return
+	}
+	now := c.clock()
+	for roomID, u := range batch {
+		if !c.crossSite(ctx, roomID) {
+			continue
```
julianshen, line 145: _🏷️ delivery_ | _🔴 Blocking_ | _`distributed-system`_

**The throttle permanently discards the latest refresh.**

A room can publish A at t=0, receive its final message B at t=4.9s, and have B skipped as throttled. The skipped value is not retained, and later empty flushes do not revisit it, so B is never announced. The same loss occurs when one peer publish fails because the watermark is recorded before publication and no per-peer retry is retained. Keep the latest unsent position per room and peer, process due entries on the flush timer even without new messages, and mark a peer refreshed only after its publish succeeds.

<details>

```
In `broadcast-worker/coalescer.go` around line 145: A room can publish A at t=0, receive its final message B at t=4.9s, and have B skipped as throttled. The skipped value is not retained, and later empty flushes do not revisit it, so B is never announced. The same loss occurs when one peer publish fails because the watermark is recorded before publication and no per-peer retry is retained. Keep the latest unsent position per room and peer, process due entries on the flush timer even without new messages, and mark a peer refreshed only after its publish succeeds.

```

</details>

inbox-worker/handler.go
```diff
@@ -254,6 +313,18 @@ func (h *Handler) handleMemberAdded(ctx context.Context, evt *model.InboxEvent)
 		return fmt.Errorf("member_added references unknown users %v in room %s", missing, event.RoomID)
 	}
 
+	// After the missing-user return, so a room that gained no subscriber leaves
+	// no orphan row (nothing deletes them). Best-effort — the subscriptions above
```
julianshen, line 317: _🏷️ atomicity_ | _🔴 Blocking_ | _`distributed-system`_

**A failed activity seed is permanently acknowledged.**

After subscriptions are committed, a transient remote_rooms write failure is only logged, so the JetStream member_added event is ACKed. Earlier refreshes were discarded because this site had no subscription, and a quiet historical room may emit no later refresh, leaving its ordering row absent indefinitely. Both subscription creation and the $max activity write are idempotent, so return the error and let the event retry.

<details>

```
In `inbox-worker/handler.go` around line 317: After subscriptions are committed, a transient remote_rooms write failure is only logged, so the JetStream member_added event is ACKed. Earlier refreshes were discarded because this site had no subscription, and a quiet historical room may emit no later refresh, leaving its ordering row absent indefinitely. Both subscription creation and the $max activity write are idempotent, so return the error and let the event retry.

```

</details>

bot-room-service/handler.go
```diff
@@ -126,8 +127,19 @@ func (h *handler) handleDMEnsure(c *natsrouter.Context, req BotDMEnsureRequest)
 		},
 		CreatedByBot: ident.ID,
 	}
-	if err := h.store.InsertRoom(c, room); err != nil && !errors.Is(err, ErrDuplicate) {
-		return nil, fmt.Errorf("insert dm room: %w", err)
+	if err := h.store.InsertRoom(c, room); err != nil {
+		if !errors.Is(err, ErrDuplicate) {
+			return nil, fmt.Errorf("insert dm room: %w", err)
+		}
+		// The DM id is deterministic, so a duplicate means this room already
```
julianshen, line 134: _🏷️ failure-handling_ | _🔴 Blocking_ | _`distributed-system`, `nats`, `silent-failure`, `tgd-review` (4 rules)_

**A failed duplicate-room read is converted into success.**

After ErrDuplicate, a transient FindRoom failure is logged and processing continues. The handler then commits the subscription and federates nil LastMsgAt for an existing DM that may contain history. Its earlier refreshes predate the target subscription, and a quiet DM provides no subsequent repair. No new mutation has occurred at this point, so fail the request and allow a safe retry.

<details>

- **`silent-failure`**:137: When InsertRoom returns ErrDuplicate, a transient FindRoom failure is only logged. The handler then creates the subscription and federates member_added without the existing DM's LastMsgAt. If that historical DM remains quiet, no subsequent refresh repairs the missing ordering position. No mutation has occurred at this point, so return the lookup error and let the ensure request retry.
- **`tgd-review`**:137: After ErrDuplicate, a transient FindRoom failure is logged and processing continues. The subscription is then created and member_added is federated without the historical DM's LastMsgAt. If that DM remains quiet, no later refresh is guaranteed to repair its ordering position. No subscription mutation has occurred yet, so return the lookup error and let the ensure request retry.
- **`nats`**:137: After deterministic DM insertion reports a duplicate, a transient room lookup failure is only logged. The handler then creates the subscription and federates `member_added` without the historical activity position. Since no mutation requiring best-effort handling has occurred yet, return the lookup error so the request can retry before publishing an incomplete event.

</details>

<details>

```
In `bot-room-service/handler.go` around line 134: After ErrDuplicate, a transient FindRoom failure is logged and processing continues. The handler then commits the subscription and federates nil LastMsgAt for an existing DM that may contain history. Its earlier refreshes predate the target subscription, and a quiet DM provides no subsequent repair. No new mutation has occurred at this point, so fail the request and allow a safe retry.

```

</details>

broadcast-worker/roomactivity.go
```diff
@@ -0,0 +1,81 @@
...
+		if err != nil {
+			return fmt.Errorf("marshal room activity event: %w", err)
+		}
+		// Every peer is attempted: one unreachable site must not deprive the
+		// others of the refresh, since a skipped peer keeps a stale position for
+		// the whole throttle interval.
+		var errs []error
+		for _, peer := range peers {
+			if err := p.Publish(ctx, subject.RoomActivity(peer), data); err != nil {
+				errs = append(errs, fmt.Errorf("publish room activity to %s: %w", peer, err))
```
julianshen, line 76: _🏷️ nats-reliability_ | _🟠 Warning_ | _`nats`_

**Core NATS cannot confirm that the destination received the refresh.**

`Publish` returning nil only confirms local acceptance; it does not report that `chat.roomactivity.<peer>` had a subscriber or that the callback completed. If the destination inbox-worker is stopped or reconnecting when a room's final refresh is sent, the message is lost and a quiet room never repairs its ordering position. Use a dedicated JetStream lane with a durable consumer, or request/reply confirmation with bounded retry and retained latest-per-room state.

<details>

```
In `broadcast-worker/roomactivity.go` around line 76: `Publish` returning nil only confirms local acceptance; it does not report that `chat.roomactivity.<peer>` had a subscriber or that the callback completed. If the destination inbox-worker is stopped or reconnecting when a room's final refresh is sent, the message is lost and a quiet room never repairs its ordering position. Use a dedicated JetStream lane with a durable consumer, or request/reply confirmation with bounded retry and retained latest-per-room state.

```

</details>

julianshen (commented):
tGD inline review

pkg/subject/subject.go
```diff
@@ -309,6 +309,20 @@ func InboxExternalAll(siteID string) string {
...
+// position to a destination site: `chat.roomactivity.{destSiteID}`.
+//
+// Deliberately outside `chat.inbox.>` so it is never captured by the INBOX
+// JetStream stream. The payload is a decorative ordering hint — coalesced,
+// idempotent, and applied under a $max guard — so it needs neither persistence
+// nor ordering, and putting it on INBOX would make a high-rate signal compete
+// for retention and ack budget with membership events that do need both.
+// Interest-routed across the supercluster: a site with no subscriber simply
+// does not receive it.
+func RoomActivity(destSiteID string) string {
```
julianshen, line 322: _🏷️ subject-construction_ | _🟠 Warning_ | _`nats`_

**Raw site IDs can change NATS subject token semantics.**

`destSiteID` is interpolated directly into the subject, and the added configuration path only removes blanks and duplicates. A configured site ID of `*` or `>` makes the inbox-worker subscribe across destination sites, while publishing to such an ID fails because publish subjects cannot contain wildcards; dots also create additional tokens. Validate `SITE_ID` and every `ALL_SITE_IDS` entry at startup as one nonempty NATS token excluding `.`, `*`, `>`, and whitespace.

<details>

```
In `pkg/subject/subject.go` around line 322: `destSiteID` is interpolated directly into the subject, and the added configuration path only removes blanks and duplicates. A configured site ID of `*` or `>` makes the inbox-worker subscribe across destination sites, while publishing to such an ID fails because publish subjects cannot contain wildcards; dots also create additional tokens. Validate `SITE_ID` and every `ALL_SITE_IDS` entry at startup as one nonempty NATS token excluding `.`, `*`, `>`, and whitespace.

```

</details>

inbox-worker/main.go
```diff
@@ -742,8 +806,26 @@ func main() {
...
 
-	handler := NewHandler(store)
+	handler := NewHandler(store, WithRoomSubCache(cfg.RoomSubCacheSize, cfg.RoomSubCacheTTL))
 	handler.badge = badge
+	slog.Info("room-sub cache configured", "size", cfg.RoomSubCacheSize, "ttl", cfg.RoomSubCacheTTL)
+
+	// Core-NATS queue subscriber for the cross-site room-activity refresh. Not on
+	// INBOX by design: the signal is coalesced, idempotent and $max-guarded, so it
+	// needs no persistence or ordering, and keeping it off the stream stops a
+	// high-rate hint competing for retention with membership events that do need
```
julianshen, line 816: _🏷️ throughput_ | _🟠 Warning_ | _`performance-and-concurrency`_

**The activity subscriber serializes one Mongo operation per room.**

NATS asynchronous callbacks are serialized per subscription, while each accepted activity event performs an individual remote_rooms UpdateOne and cold or evicted entries also perform a subscription lookup. With many active remote rooms, the one-event-per-room fan-out can outpace this callback, fill the NATS pending buffer, and drop final positions while applying sustained Mongo pressure. Enqueue into a bounded per-room coalescer and periodically BulkWrite the latest positions, or otherwise use bounded parallelism and batching.

<details>

```
In `inbox-worker/main.go` around line 816: NATS asynchronous callbacks are serialized per subscription, while each accepted activity event performs an individual remote_rooms UpdateOne and cold or evicted entries also perform a subscription lookup. With many active remote rooms, the one-event-per-room fan-out can outpace this callback, fill the NATS pending buffer, and drop final positions while applying sustained Mongo pressure. Enqueue into a bounded per-room coalescer and periodically BulkWrite the latest positions, or otherwise use bounded parallelism and batching.

```

</details>

inbox-worker/handler.go
```diff
@@ -299,6 +370,20 @@ func (h *Handler) handleMemberRemoved(ctx context.Context, evt *model.InboxEvent
 	if err := h.store.DeleteSubscriptionsByAccounts(ctx, memberEvt.RoomID, memberEvt.Accounts); err != nil {
 		return fmt.Errorf("delete subscriptions for room %s: %w", memberEvt.RoomID, err)
 	}
+	// Other members may remain, so re-resolve rather than caching a guess. When
+	// none do, the ordering row has no reader left here — drop it, or it becomes
+	// the orphan the seed path is careful not to create. Best-effort: a stale row
+	// only costs space, and re-adding a member re-seeds it.
+	h.roomSubs.invalidate(memberEvt.RoomID)
+	if stillMember, err := h.hasRoomSubscription(ctx, memberEvt.RoomID); err != nil {
```
julianshen, line 378: _🧹 cleanup_ | _🟠 Warning_ | _`silent-failure`_

**Cleanup failures are acknowledged after the final member leaves.**

If the remaining-subscription check or DeleteRemoteRoomActivity fails, member_removed still completes and is acknowledged. After the final member leaves there may be no future removal to retry, leaving the activity row orphaned permanently. Subscription deletion and deletion of an absent activity row are repeatable, so return either error and let JetStream redeliver.

<details>

```
In `inbox-worker/handler.go` around line 378: If the remaining-subscription check or DeleteRemoteRoomActivity fails, member_removed still completes and is acknowledged. After the final member leaves there may be no future removal to retry, leaving the activity row orphaned permanently. Subscription deletion and deletion of an absent activity row are repeatable, so return either error and let JetStream redeliver.

```

</details>

broadcast-worker/coalescer.go
mliu33: We should address this issue in a separate PR. Never have unbound map. And we could initiate a separate go routine to do remote room activity work. Thanks

c.c @hmchangw

inbox-worker/main.go
```diff
@@ -742,8 +806,26 @@ func main() {
...
+	// both. Fire-and-forget — a failure self-heals on the room's next message.
+	activitySub, err := nc.QueueSubscribe(ctx, subject.RoomActivity(cfg.SiteID), "inbox-worker",
```
mliu33, line 818: Why not reuse existing mechanism as updageUserStatus which publishes directly to remote inbox Jetstream ?

inbox-worker/roomsubcache.go
```diff
@@ -0,0 +1,110 @@
...
+		entries: lru.NewLRU[string, bool](size, nil, ttl),
+		metrics: cachemetrics.For("room_sub", "l1"),
+	}
+}
+
+func (c *roomSubCache) get(ctx context.Context, roomID string) (bool, bool) {
+	if c == nil {
+		return false, false
+	}
+	c.mu.Lock()
```
mliu33, line 49: Use RLock. Don't lock for read operation

pkg/subject/subject.go
```diff
@@ -309,6 +309,20 @@ func InboxExternalAll(siteID string) string {
...
+	return fmt.Sprintf("chat.roomactivity.%s", destSiteID)
```
mliu33, line 323: Our convention for server-to-server is using prefix "chat.server.>"

mliu33 (approved):
Excellent work! Some suggestions, thanks!

PR #280, opened 2026-08-15
feat(chat-frontend): show the last message as a sidebar preview

hmchangw:
## What

Each sidebar room now renders its most recent message as a second line — `Alice: hey are we still on` — kept live as messages arrive, are edited, and are deleted.

**No backend change and no wire change.** `PreviewMessage` has been arriving on `subscription.list` rows (`sub.room.previewMessage`) and on `message_edited` / `message_deleted` events all along; the frontend never declared the field, so the data was being thrown away.

## How

- **`state.previews`** — a new `Record<roomId, { messageId, senderName, text, createdAt, encrypted? }>` in `roomEventsReducer`. A keyed map rather than a field on `RoomSummary`, because summaries are rebuilt wholesale from `Room` records that mirror `pkg/model.Room`, and the backend hangs the preview off `SubscriptionRoom` instead.
- **`lib/previewText.js`** — flattens a body to one plain-text line over the existing `parseMessageContent` tokenizer: mentions resolve to `@Display Name`, links to their visible text, markup and fences dropped, whitespace collapsed, capped at 140 chars on a surrogate-safe boundary.
- **`ROOM_PREVIEW_UPDATED`** — its own action rather than a field threaded onto `MESSAGE_EDITED` / `MESSAGE_DELETED`. Both of those bail when a room has no message buffer, which is precisely the normal case for a sidebar row of a room the user hasn't opened.
- **`useSidebarSections`** joins the map into each room; **`RoomItem`** renders the second line.

## Behavior worth knowing

| Case | Behavior |
|---|---|
| No preview available | Row keeps full height with a blank snippet line. The sidebar bootstraps over three paginated RPCs, so a row that grew a second line as its preview landed would reflow the list in stages. |
| DM / botDM | Sender prefix omitted — the row title is already the counterpart's name. |
| Attachment-only message | `Photo` / `Audio` / `Video` / the file's title. |
| System messages | Excluded, mirroring the server's eligibility rule. Note `important` is *not* a system type and previews normally. |
| Thread replies | Excluded. This is broader than the server's rule, which excludes only hidden replies — correct only while this frontend has no `tshow` support, and flagged in the code for whoever adds it. |
| Logout | `RESET` clears the map, so one account's preview can't survive into the next session on a shared device. |
| Concurrent edit/delete | Rejected when strictly older than the stored preview; `broadcast-worker` processes canonical messages concurrently, so a stale resolve can arrive after a newer message. |
| Bootstrap race | `BUCKETS_LOADED` seeds only rooms with no preview yet — the DM subscription goes live before `fetchSidebarBuckets` resolves, so a message arriving in that window is newer than the list snapshot. |

## Testing

949 tests across 85 files, `tsc --noEmit` clean, production build clean.

The design spec and implementation plan are included under `docs/superpowers/`.

## Follow-ups (none blocking)

- **Backend.** `history-service`'s `previewAfterMutation` returns `nil` both for "no eligible message remains" and for a read error. The client cannot distinguish them, so a transient Cassandra failure during a delete blanks a valid snippet until the next `subscription.list`, and an encrypted room still shows server plaintext on cold load. Both need an explicit cleared-signal on the event; no honest client-side mitigation exists, since never-clearing would strand a deleted message's text in the sidebar permanently.
- Undecryptable E2E messages preview as `Unknown: [encrypted message]` — untested.
- A fourth inline copy of the `dm || botDM` predicate now exists; candidate for a shared helper in `lib/roomFormat`.
- `previewText` parses the full body to keep 140 chars. A bounded pre-parse was tried and reverted: markup shrinks up to 7:1 when flattened, so any slice can truncate below the cap. If this ever shows in a profile, the safe optimization is skipping the parse entirely for bodies with no markup characters.

### DIFF
--- a/chat-frontend/src/api/types.ts
+++ b/chat-frontend/src/api/types.ts
@@ -48,6 +48,11 @@ export interface SubscriptionRoom {
    *  for initial key bootstrap (same payload as the room.key.get RPC). */
   privateKey?: string
   keyVersion?: number
+  /** The room's latest eligible message. Omitted when the room has no
+   *  message, that site's enrichment degraded, or the request set
+   *  `includeLastMessage: false`. Never present on live
+   *  `subscription.update` events — only on `subscription.list` rows. */
+  previewMessage?: PreviewMessage
 }
 
 /** Mirrors pkg/model.Subscription — the per-user record linking a user
@@ -172,6 +177,10 @@ export interface Participant {
   engName?: string
   chineseName?: string
   siteId?: string
+  /** Server-composed render-ready name (engName + chineseName + account
+   *  fallback; a bot sender's is its app name). Prefer it over the raw
+   *  fields. Mirrors model.Participant.DisplayName. */
+  displayName?: string
 }
 
 /** One reactor on a message reaction. Mirrors the wire `reactionUser`
@@ -214,6 +223,27 @@ export interface Attachment {
   /** Client-only: a local object URL for an optimistic just-sent image.
    *  Never arrives from the server; preferred as the <img> src when present. */
   localUrl?: string
+}
+
+/** Mirrors model.PreviewMessage — a room's most-recent eligible message,
+ *  resolved server-side at read time for room-list rendering. Eligible means
+ *  not soft-deleted and not a system message; the server walks back past an
+ *  ineligible tail, so the client never re-implements that rule.
+ *
+ *  Delivered on `subscription.list` rows (`SubscriptionRoom.previewMessage`)
+ *  and refreshed on `message_edited` / `message_deleted` events. */
+export interface PreviewMessage {
+  messageId: string
+  sender: Participant
+  /** The full message body; the client truncates for display. */
+  content: string
+  /** RFC3339. */
+  createdAt: string
+  attachments?: Attachment[]
+  mentions?: Participant[]
+  /** Always empty today — the server-side write path has not landed.
+   *  Declared for forward-compat; nothing reads it. */
+  visibleTo?: string
 }
 
 /** Cassandra's QuotedParentMessage shape — what gets embedded on a
--- a/chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx
+++ b/chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx
@@ -473,20 +473,22 @@ export interface SidebarSection {
  */
 export function useSidebarSections(): SidebarSection[] {
   const { state } = useRoomEventsInternal()
-  const { summaries, subscriptions, chatlist } = state
+  const { summaries, subscriptions, chatlist, previews } = state
   return useMemo(() => {
     const enrich = (room: RoomSummary): RoomSummary => {
       const sub = subscriptions[room.id]
-      if (!sub) return room
+      const preview = previews[room.id]
+      if (!sub && !preview) return room
       return {
         ...room,
-        subscriptionName: sub.name ?? room.subscriptionName,
-        hrInfo: sub.hrInfo ?? room.hrInfo,
+        subscriptionName: sub?.name ?? room.subscriptionName,
+        hrInfo: sub?.hrInfo ?? room.hrInfo,
+        preview,
       }
     }
     const sections = deriveSidebarSections(summaries, subscriptions, chatlist) as SidebarSection[]
     return sections.map((s) => ({ ...s, rooms: s.rooms.map(enrich) }))
-  }, [summaries, subscriptions, chatlist])
+  }, [summaries, subscriptions, chatlist, previews])
 }
 
 /** Raw overlay section order (the full list the backend stores — built-ins +
--- a/chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx
+++ b/chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx
@@ -68,6 +68,25 @@ interface RoomPreview {
   senderName: string
   /** Single-line, flattened, capped at PREVIEW_MAX_LENGTH. */
   text: string
+  /** RFC3339, from the previewed message's createdAt. Powers
+   *  ROOM_PREVIEW_UPDATED's recency guard: broadcast-worker processes
+   *  canonical messages concurrently, so an edit/delete's server-resolved
+   *  preview can arrive after a genuinely newer message's live preview —
+   *  a strictly older incoming createdAt is rejected rather than regressing
+   *  the sidebar. Optional because older code paths / test fixtures may not
+   *  set it; a missing timestamp on either side of the comparison is
+   *  treated as "accept the write". */
+  createdAt?: string
+  /** True when this preview was built from a live message the client
+   *  couldn't decrypt — the reducer's "[encrypted message]" placeholder
+   *  branch. Guards ROOM_PREVIEW_UPDATED from overwriting the placeholder
+   *  with the server's plaintext previewMessage body for the SAME message
+   *  (history-service / broadcast-worker relay that body unencrypted). A
+   *  wire preview for a DIFFERENT, newer message still overwrites normally.
+   *  Known residual: after a reload there's no placeholder in state to
+   *  compare against, so bootstrap still shows the server's plaintext —
+   *  fully closing that gap needs a backend change, not a client one. */
+  encrypted?: boolean
 }
 
 /** Sidebar summary — derived from `model.Room` + the user's
--- a/chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx
+++ b/chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx
@@ -58,6 +58,18 @@ interface RoomBufferState {
  *  returned page of exactly this size means older messages may follow. */
 const HISTORY_PAGE_SIZE = 50
 
+/** One room's stored sidebar preview. Flattened and name-resolved at write
+ *  time so the render layer never branches on whether it came from a wire
+ *  PreviewMessage or a live Message. */
+interface RoomPreview {
+  /** The previewed message's id — guards the delete-clears rule in the
+   *  reducer's ROOM_PREVIEW_UPDATED case. */
+  messageId: string
+  senderName: string
+  /** Single-line, flattened, capped at PREVIEW_MAX_LENGTH. */
+  text: string
+}
+
 /** Sidebar summary — derived from `model.Room` + the user's
  *  Subscription. Only the fields the sidebar / chat header read. */
 interface RoomSummary {
@@ -75,6 +87,10 @@ interface RoomSummary {
    *  for DM rooms whose subscription carried it. Plain channels stay
    *  undefined here. */
   hrInfo?: SubscriptionHRInfo
+  /** Joined in by useSidebarSections from state.previews. Undefined when the
+   *  room has no preview — the row still renders at full height with a blank
+   *  snippet line. */
+  preview?: RoomPreview
 }
 
 /** Top-level state shape returned by `roomEventsReducer`. */
@@ -96,6 +112,11 @@ interface RoomEventsState {
    *  so consumers reading the map for either channel or DM rooms see
    *  hrInfo as optional without narrowing. */
   subscriptions: Record<string, DMSubscription>
+  /** Keyed by roomId. Absent key = no preview to show. Deliberately NOT a
+   *  field on RoomSummary: summaries are rebuilt from `Room` records, which
+   *  mirror pkg/model.Room — and the backend hangs previewMessage off
+   *  SubscriptionRoom instead, so a summary field would break the mirror. */
+  previews: Record<string, RoomPreview>
   /** Chatlist section-definition overlay (names, order, sortMode). Seeded by
    *  CHATLIST_LOADED, replaced by CHATLIST_UPDATED (LWW). Membership rides the
    *  subscriptions; this is O(sections). */
@@ -530,4 +551,4 @@ export function useSubscription(roomId: string | null | undefined): DMSubscripti
   return roomId ? state.subscriptions[roomId] : undefined
 }
 
-export type { RoomEventsState, RoomSummary, RoomBufferState, RoomEventsContextValue }
+export type { RoomEventsState, RoomSummary, RoomPreview, RoomBufferState, RoomEventsContextValue }
--- a/chat-frontend/src/context/RoomEventsContext/reducer.js
+++ b/chat-frontend/src/context/RoomEventsContext/reducer.js
@@ -1,7 +1,8 @@
 import { appendBounded, mergeById, MAX_CACHED } from '@/lib/messageBuffer'
 import { defaultChatlistState, sortByLastMsgDesc } from '@/lib/chatlist'
-import { previewSnippet, previewText } from '@/lib/previewText'
+import { previewSnippet } from '@/lib/previewText'
 import { participantDisplayName, messageSenderName } from '@/lib/participantName'
+import { isSystemMessageType } from '@/lib/messageType'
 
 export { MAX_CACHED }
 
@@ -134,6 +135,17 @@ function toSummary(room) {
 // empty sentinel.
 function previewFromMessage(msg, fallbackMentions) {
   if (!msg || !msg.id) return null
+  // Fix 1: a system message (room_renamed, members_added, …) is published to
+  // MESSAGES-CANONICAL with non-empty content and fans out as an ordinary
+  // new_message event, but it must never become the room's sidebar snippet —
+  // history-service's previewMessage resolution already excludes it, so
+  // treating it as eligible here makes a reload flip the snippet back.
+  // isSystemMessageType is a set-membership test, NOT `type !== ''`: the
+  // client-settable "important" type is deliberately absent from that set
+  // and previews like a normal message. sysMsgData is checked belt-and-
+  // braces because that's the field MessageList actually keys off to route
+  // a message to the SystemMessage renderer.
+  if (isSystemMessageType(msg.type) || msg.sysMsgData != null) return null
   return {
     messageId: msg.id,
     senderName: messageSenderName(msg),
@@ -141,13 +153,25 @@ function previewFromMessage(msg, fallbackMentions) {
     // Message.Mentions. Without the fallback a live preview shows the raw
     // @account while the same message shows @Display Name after a reload.
     text: previewSnippet(msg.content, msg.mentions ?? fallbackMentions, msg.attachments),
+    createdAt: msg.createdAt,
+    // Fix 3: marks a preview built from the "[encrypted message]" placeholder
+    // (see the MESSAGE_RECEIVED encrypted-message branch below) so
+    // ROOM_PREVIEW_UPDATED can avoid clobbering it with the server's
+    // unencrypted previewMessage body for the same message.
+    ...(msg.encrypted ? { encrypted: true } : {}),
   }
 }
 
-// Two stored previews are interchangeable when all three rendered fields match.
-// Used to keep the previews reference stable on a same-content write (e.g. the
-// server echo of a message this client already stored optimistically) — a fresh
-// object would invalidate useSidebarSections' memo for every room in the sidebar.
+// Two stored previews are interchangeable when all three RENDERED fields
+// match (messageId/senderName/text — what RoomList actually displays). Used
+// to keep the previews reference stable on a same-content write (e.g. the
+// server echo of a message this client already stored optimistically) — a
+// fresh object would invalidate useSidebarSections' memo for every room in
+// the sidebar. createdAt is deliberately NOT part of this comparison: it's
+// metadata for the ROOM_PREVIEW_UPDATED recency guard, not something the UI
+// renders, and two writes sharing the same messageId always share the same
+// createdAt (it's immutable per message) so including it could never change
+// the outcome — it would just be a redundant field-by-field check.
 function samePreview(a, b) {
   return !!a && !!b && a.messageId === b.messageId && a.senderName === b.senderName && a.text === b.text
 }
@@ -160,7 +184,21 @@ function previewFromWire(previewMessage) {
     messageId: previewMessage.messageId,
     senderName: participantDisplayName(previewMessage.sender),
     text: previewSnippet(previewMessage.content, previewMessage.mentions, previewMessage.attachments),
+    createdAt: previewMessage.createdAt,
   }
+}
+
+// Fix 7: true only when `nextCreatedAt` is a STRICTLY older, parseable
+// timestamp than `curCreatedAt`. Tolerant by design — a missing or
+// unparseable timestamp on either side is treated as "accept the write"
+// rather than dropping a legitimate update just because older stored data
+// (or a malformed wire payload) lacks a comparable createdAt.
+function isOlderPreview(nextCreatedAt, curCreatedAt) {
+  if (!nextCreatedAt || !curCreatedAt) return false
+  const nextT = Date.parse(nextCreatedAt)
+  const curT = Date.parse(curCreatedAt)
+  if (Number.isNaN(nextT) || Number.isNaN(curT)) return false
+  return nextT < curT
 }
 
 /**
@@ -506,7 +544,13 @@ export function roomEventsReducer(state, action) {
           prev.messages.some((m) => m.id === msg.id) ||
           prev.pendingLiveMessages.some((m) => m.id === msg.id)
         ) {
-          return state
+          // Fix 2: previews is computed once above and applied at every
+          // return point in this action — this duplicate-message guard used
+          // to skip that, discarding an optimistic preview's upgrade from
+          // the server echo whenever the room sat in historical buffer mode.
+          // Only allocate when the write actually changed something, so a
+          // true no-op still returns the identical state reference.
+          return previews === state.previews ? state : { ...state, previews }
         }
         const pendingLiveMessages = [...prev.pendingLiveMessages, msg]
         const nextRoomState = {
@@ -833,10 +877,21 @@ export function roomEventsReducer(state, action) {
       // action carries no mentions, so mention tokens flatten literally
       // here; the server's message_edited event follows with the
       // authoritative preview a moment later.
+      // Fix 4: use previewSnippet (not previewText) so an edit that blanks
+      // the body falls back to the attachment label instead of leaving a
+      // dangling "Sender: " row — attachments aren't on the action (edits
+      // never touch them), so read them off the buffered message being
+      // edited (updatedMsg, which carries them forward from prev.messages[idx]).
       const cur = state.previews[action.roomId]
       const previews =
         cur && cur.messageId === action.messageId
-          ? { ...state.previews, [action.roomId]: { ...cur, text: previewText(action.content) } }
+          ? {
+              ...state.previews,
+              [action.roomId]: {
+                ...cur,
+                text: previewSnippet(action.content, updatedMsg.mentions, updatedMsg.attachments),
+              },
+            }
           : state.previews
       return {
         ...state,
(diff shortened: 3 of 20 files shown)


julianshen:
**12 findings · 7 unique issues · 7 inline comments posted** — 🔴 4 blocking · 🟠 8 warning

<details>
<summary>📒 Files reviewed (20)</summary>

* `chat-frontend/src/api/index.ts`
* `chat-frontend/src/components/MainApp/Sidebar/RoomList/RoomList.jsx`
* `chat-frontend/src/components/MainApp/Sidebar/RoomList/RoomList.test.jsx`
* `chat-frontend/src/components/MainApp/Sidebar/RoomList/style.css`
* `chat-frontend/src/components/shared/MessageList/MessageRow/MessageRow.jsx`
* `chat-frontend/src/components/shared/MessageList/MessageRow/MessageRow.test.jsx`
* `chat-frontend/src/context/RoomEventsContext/RoomEventsContext.test.jsx`
* `chat-frontend/src/context/RoomEventsContext/useRoomSubscriptions.js`
* `chat-frontend/src/lib/messageType.js`
* `chat-frontend/src/lib/messageType.test.js`
* `chat-frontend/src/lib/participantName.js`
* `chat-frontend/src/lib/participantName.test.js`
* `chat-frontend/src/lib/previewText.js`
* `chat-frontend/src/lib/previewText.test.js`
* `docs/superpowers/plans/2026-08-14-chat-frontend-message-preview.md`
* `docs/superpowers/specs/2026-08-14-chat-frontend-message-preview-design.md`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

chat-frontend/src/context/RoomEventsContext/reducer.js
```diff
@@ -419,6 +484,21 @@ export function roomEventsReducer(state, action) {
...
+      // applied at every return point below.
+      //
+      // Note: excluding EVERY thread reply this way is broader than the
+      // server's rule — the server excludes only hidden (tshow: false)
+      // replies, and a tshow: true reply can legitimately be a room's
+      // preview. That's fine only because this frontend has no tshow
+      // support at all, so no thread reply ever reaches the room timeline
+      // here. Anyone adding tshow must revisit this.
+      const nextPreview = previewFromMessage(msg, evt.mentions)
+      const previews =
```
julianshen, line 536: [inline permission test] tGDBot inline-comment permission test for PR #280.

chat-frontend/src/context/RoomEventsContext/reducer.js
```diff
@@ -283,13 +335,26 @@ export function roomEventsReducer(state, action) {
           subs[s.id] ? mergeSubscriptionIntoSummary(s, subs[s.id]) : s
         )
       }
+      // Seed only rooms with no preview yet. A live message can land before
+      // fetchSidebarBuckets resolves (the DM subscription goes live first), and
+      // that message is NEWER than this list snapshot — overwriting it would show
+      // an older message in the sidebar than the room itself displays.
+      let previews = state.previews
```
julianshen, line 380: _🏷️ ordering_ | _🟠 Warning_ | _`distributed-system`_

**Bootstrap arrival order can preserve a stale preview**

`BUCKETS_LOADED` skips the authoritative `sub.room.previewMessage` whenever any live preview entry already exists, assuming that an event delivered before the HTTP response is newer than the snapshot. A delayed or replayed `MESSAGE_RECEIVED` for m1 can arrive before a list snapshot containing newer m2, causing m1 to block seeding; if it arrives afterward, the live writer also overwrites m2 without any ordering check. The sidebar can therefore remain stale until another refresh. Carry a server-monotonic room revision/sequence (or another authoritative ordering value) in the preview state and accept only newer updates; do not use map presence as a freshness test.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 342: `BUCKETS_LOADED` skips the authoritative `sub.room.previewMessage` whenever any live preview entry already exists, assuming that an event delivered before the HTTP response is newer than the snapshot. A delayed or replayed `MESSAGE_RECEIVED` for m1 can arrive before a list snapshot containing newer m2, causing m1 to block seeding; if it arrives afterward, the live writer also overwrites m2 without any ordering check. The sidebar can therefore remain stale until another refresh. Carry a server-monotonic room revision/sequence (or another authoritative ordering value) in the preview state and accept only newer updates; do not use map presence as a freshness test.

```

</details>

chat-frontend/src/context/RoomEventsContext/reducer.js
```diff
@@ -117,6 +129,40 @@ function toSummary(room) {
   }
 }
 
+// Build a stored preview from a live message. Returns null when there is
```
julianshen, line 133: _🏷️ consistency_ | _🟠 Warning_ | _`distributed-system`_

**Live system messages can overwrite the eligible room preview**

`previewFromMessage` accepts every message with an id, and `MESSAGE_RECEIVED` writes its result into `state.previews`. The documented eligibility rule excludes system messages, but this live path has no equivalent filter. If a system message is delivered after an eligible message, the sidebar shows the system message even though the next `subscription.list` snapshot would omit it; no later preview correction is triggered for ordinary new-message delivery. Reuse the timeline's system-message predicate before deriving a live preview and add a regression test.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 132: `previewFromMessage` accepts every message with an id, and `MESSAGE_RECEIVED` writes its result into `state.previews`. The documented eligibility rule excludes system messages, but this live path has no equivalent filter. If a system message is delivered after an eligible message, the sidebar shows the system message even though the next `subscription.list` snapshot would omit it; no later preview correction is triggered for ordinary new-message delivery. Reuse the timeline's system-message predicate before deriving a live preview and add a regression test.

```

</details>

chat-frontend/src/context/RoomEventsContext/useRoomSubscriptions.js
```diff
@@ -198,6 +198,23 @@ export function useRoomSubscriptions(
...
+        // previewMessage the server sends. Separately, this frontend's own
+        // preview computation (reducer.js) excludes EVERY thread reply from
+        // being a preview candidate, which is broader than the server's rule
+        // (hidden/tshow: false only) — correct only because this frontend has
+        // no tshow support, so no shown reply ever reaches the room timeline
+        // here. Anyone adding tshow must revisit that exclusion.
+        if (evt.previewMessage) {
+          safeDispatch({
+            type: 'ROOM_PREVIEW_UPDATED',
+            roomId: evt.roomId,
```
julianshen, line 214: _🏷️ routing_ | _🟠 Warning_ | _`nats`_

**Room preview events trust payload roomId across room subscriptions**

This handler is attached to room-specific subjects such as `chat.room.g1.event`, but the new preview action routes using `evt.roomId` without checking it against the subscribed room. A malformed or misrouted NATS payload received on g1 with `roomId: g2` can overwrite or clear g2's sidebar preview, including rooms with no message buffer. Capture the expected room ID in the subscription closure and reject mismatches (or route using that bound ID) before dispatching preview updates.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/useRoomSubscriptions.js` around line 214: This handler is attached to room-specific subjects such as `chat.room.g1.event`, but the new preview action routes using `evt.roomId` without checking it against the subscribed room. A malformed or misrouted NATS payload received on g1 with `roomId: g2` can overwrite or clear g2's sidebar preview, including rooms with no message buffer. Capture the expected room ID in the subscription closure and reject mismatches (or route using that bound ID) before dispatching preview updates.

```

</details>

chat-frontend/src/lib/previewText.js
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Flatten a message body to a single line of plain text.
+ *
+ * @param {string | null | undefined} content
+ * @param {{ account?: string, engName?: string }[]} [mentions]
+ * @returns {string} '' when there is nothing to show
+ */
+export function previewText(content, mentions = []) {
+  if (!content) return ''
```
julianshen, line 23: _⚡ memory_ | _🟠 Warning_ | _`oom`_

**The 140-character cap is applied after unbounded parsing and copying**

A user-controlled message body is fully parsed by `parseMessageContent`, concatenated by `flattenNodes`, and copied again by whitespace normalization before `slice` limits the result. Large message payloads therefore cause transient allocations proportional to the entire body on every live message and preview update, despite the nominal cap, leading to GC thrashing or process OOM. Enforce a bounded input/traversal while flattening (and preserve the 140-character output limit).

<details>

```
In `chat-frontend/src/lib/previewText.js` around line 23: A user-controlled message body is fully parsed by `parseMessageContent`, concatenated by `flattenNodes`, and copied again by whitespace normalization before `slice` limits the result. Large message payloads therefore cause transient allocations proportional to the entire body on every live message and preview update, despite the nominal cap, leading to GC thrashing or process OOM. Enforce a bounded input/traversal while flattening (and preserve the 140-character output limit).

```

</details>

chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx
```diff
@@ -452,20 +473,22 @@ export interface SidebarSection {
...
       return {
         ...room,
-        subscriptionName: sub.name ?? room.subscriptionName,
-        hrInfo: sub.hrInfo ?? room.hrInfo,
+        subscriptionName: sub?.name ?? room.subscriptionName,
+        hrInfo: sub?.hrInfo ?? room.hrInfo,
+        preview,
       }
     }
     const sections = deriveSidebarSections(summaries, subscriptions, chatlist) as SidebarSection[]
```
julianshen, line 508: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Every preview update recreates and rerenders the entire sidebar**

Each live message changes `previews`, invalidating this memo. `sections.map(... rooms.map(enrich))` then creates a new summary object for every subscribed room, even when its preview did not change. With many rooms and a busy workspace, every message causes O(number of rooms) allocations and rerenders all RoomItems, creating avoidable main-thread and React commit load. Preserve unchanged room identities and update only the affected room (for example with per-room memoized selectors or a memoized row-level preview join).

<details>

```
In `chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx` around line 489: Each live message changes `previews`, invalidating this memo. `sections.map(... rooms.map(enrich))` then creates a new summary object for every subscribed room, even when its preview did not change. With many rooms and a busy workspace, every message causes O(number of rooms) allocations and rerenders all RoomItems, creating avoidable main-thread and React commit load. Preserve unchanged room identities and update only the affected room (for example with per-room memoized selectors or a memoized row-level preview join).

```

</details>

chat-frontend/src/api/types.ts
```diff
@@ -48,6 +48,11 @@ export interface SubscriptionRoom {
    *  for initial key bootstrap (same payload as the room.key.get RPC). */
   privateKey?: string
   keyVersion?: number
+  /** The room's latest eligible message. Omitted when the room has no
+   *  message, that site's enrichment degraded, or the request set
+   *  `includeLastMessage: false`. Never present on live
```
julianshen, line 53: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**The preview wire field is declared at a different nesting than it is read.**

`SubscriptionRoom.previewMessage` is declared at the subscription root, but `reducer.js` seeds previews from `sub.room.previewMessage`, and the integration fixtures use that nested shape. Align the type declaration with the actual subscription.list payload (or change the reducer if the wire field is actually top-level), otherwise the API mirror is incorrect and can hide a cold-start regression.

<details>

```
In `chat-frontend/src/api/types.ts` around line 53: `SubscriptionRoom.previewMessage` is declared at the subscription root, but `reducer.js` seeds previews from `sub.room.previewMessage`, and the integration fixtures use that nested shape. Align the type declaration with the actual subscription.list payload (or change the reducer if the wire field is actually top-level), otherwise the API mirror is incorrect and can hide a cold-start regression.

```

</details>

chat-frontend/src/context/RoomEventsContext/reducer.js
```diff
@@ -117,6 +129,40 @@ function toSummary(room) {
...
+    messageId: msg.id,
+    senderName: messageSenderName(msg),
+    // Mentions ride the event, not the message — the gatekeeper never populates
+    // Message.Mentions. Without the fallback a live preview shows the raw
+    // @account while the same message shows @Display Name after a reload.
+    text: previewSnippet(msg.content, msg.mentions ?? fallbackMentions, msg.attachments),
+  }
+}
+
+// Two stored previews are interchangeable when all three rendered fields match.
```
julianshen: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Live previews do not apply the server's system-message eligibility rule.**

`previewFromMessage` only checks for an id, but it is used for `MESSAGE_RECEIVED` and therefore can replace the sidebar preview with a system or already-deleted message. The documented server eligibility excludes system and soft-deleted messages and walks back to the previous eligible message; live message-derived previews need the corresponding client-side exclusion or an authoritative preview update for those events.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 147: `previewFromMessage` only checks for an id, but it is used for `MESSAGE_RECEIVED` and therefore can replace the sidebar preview with a system or already-deleted message. The documented server eligibility excludes system and soft-deleted messages and walks back to the previous eligible message; live message-derived previews need the corresponding client-side exclusion or an authoritative preview update for those events.

```

</details>

julianshen (commented):
tGD inline review

mliu33 (approved):

julianshen (commented):
tGD inline review

chat-frontend/src/context/RoomEventsContext/reducer.js
```diff
@@ -791,6 +958,45 @@ export function roomEventsReducer(state, action) {
...
+        // compare against (BUCKETS_LOADED seeds straight from the wire), so
+        // bootstrap still shows the server's plaintext — fully closing that
+        // needs a backend change (withholding plaintext for encrypted rooms
+        // server-side), not something this guard can fix client-side.
+        if (cur?.encrypted && cur.messageId === next.messageId) return state
+        // Fix 7: broadcast-worker processes canonical messages concurrently,
+        // so an edit/delete's server-resolved preview can be computed before
+        // a genuinely newer message and still arrive after it — reject only
+        // a STRICTLY older write; missing/unparseable timestamps are
+        // accepted rather than dropped (see isOlderPreview).
```
julianshen, line 987: _🏷️ silent-failure_ | _🔴 Blocking_ | _`distributed-system`, `mongodb`, `nats`, `performance-and-concurrency`, `silent-failure`, `tgd-review` (6 rules)_

**Deleting the latest message silently preserves its stale preview.**

When deleting the currently displayed latest message, the server returns the previous eligible message as `previewMessage`; its `createdAt` is necessarily older than the deleted message. This guard therefore returns the unchanged state before considering `deletedMessageId`, while the separate deletion action still succeeds. The sidebar continues exposing the deleted message text until another message or reload. Bypass the recency guard only when the event deletes the currently stored preview, while retaining it for delayed deletes after a newer preview has arrived. Add a regression test for a delete carrying an older replacement preview.

<details>

- **`mongodb`**:985: After soft-deleting the displayed message m2, the server's latest-eligible-message lookup normally returns the preceding message m1. Because m1.createdAt is necessarily older than m2.createdAt, this guard rejects the authoritative result and leaves deleted m2 content in the sidebar indefinitely. Accept an older preview when deletedMessageId identifies the currently displayed preview; continue rejecting it when a stale delete targets another message. Add a regression test where deleting m2 returns m1.
- **`distributed-system`**:989: When the displayed latest message m2 is deleted, the server correctly returns the preceding eligible message m1 as `previewMessage`. Because m1 necessarily has an older `createdAt` than m2, this guard rejects the update and leaves the deleted m2 visible indefinitely. Bypass the recency rejection when `deletedMessageId` identifies the currently displayed preview; retain it when a delayed delete concerns some other message.
- **`tgd-review`**:984: A delete event can legitimately return the previous eligible message, whose `createdAt` is older than the deleted preview. The unconditional recency check rejects that replacement, leaving the deleted message displayed indefinitely. Bypass the older-preview guard when `deletedMessageId` matches the currently displayed preview; retain it for stale delete events targeting another message.
- **`nats`**:985: When the newest message m2 is deleted, the `message_deleted` event legitimately carries the previous eligible message m1 as `previewMessage`. Because m1's `createdAt` is older than the currently displayed m2, this guard rejects the server-authoritative walk-back and leaves the deleted message in the sidebar. Allow an older replacement when `deletedMessageId` matches the current preview, while retaining the guard for a delayed deletion of some other message.
- **`performance-and-concurrency`**:985: Deleting the displayed latest message m2 normally returns previous message m1 as `previewMessage`; because m1 necessarily has an older `createdAt`, `isOlderPreview` rejects the authoritative fallback and leaves deleted m2 visible indefinitely. Permit an older preview when `deletedMessageId` matches the currently displayed preview, while retaining the guard when the current preview has already advanced to another message.

</details>

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 987: When deleting the currently displayed latest message, the server returns the previous eligible message as `previewMessage`; its `createdAt` is necessarily older than the deleted message. This guard therefore returns the unchanged state before considering `deletedMessageId`, while the separate deletion action still succeeds. The sidebar continues exposing the deleted message text until another message or reload. Bypass the recency guard only when the event deletes the currently stored preview, while retaining it for delayed deletes after a newer preview has arrived. Add a regression test for a delete carrying an older replacement preview.

```

</details>

chat-frontend/src/context/RoomEventsContext/reducer.js
```diff
@@ -283,13 +373,26 @@ export function roomEventsReducer(state, action) {
           subs[s.id] ? mergeSubscriptionIntoSummary(s, subs[s.id]) : s
         )
       }
+      // Seed only rooms with no preview yet. A live message can land before
+      // fetchSidebarBuckets resolves (the DM subscription goes live first), and
+      // that message is NEWER than this list snapshot — overwriting it would show
+      // an older message in the sidebar than the room itself displays.
+      let previews = state.previews
+      for (const [roomId, sub] of Object.entries(subs)) {
+        if (previews[roomId]) continue
```
julianshen, line 382: _🏷️ ordering_ | _🟠 Warning_ | _`distributed-system`_

**Preview updates still lack a monotonic room order.**

The prior ordering concern remains. A delayed or replayed m1 can populate `previews` before a `BUCKETS_LOADED` snapshot containing m2, causing this presence check to discard the newer snapshot; if m1 arrives afterward, `MESSAGE_RECEIVED` overwrites m2 without a recency check. The new guard only covers `ROOM_PREVIEW_UPDATED`, and its immutable message `createdAt` also cannot order two edits of the same message. With concurrent broadcast workers or a lost event followed by snapshot reconciliation, the sidebar can remain stale indefinitely. Carry a server-monotonic room revision/sequence on snapshots and all live mutations, and apply every preview writer only when its revision is newer.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 382: The prior ordering concern remains. A delayed or replayed m1 can populate `previews` before a `BUCKETS_LOADED` snapshot containing m2, causing this presence check to discard the newer snapshot; if m1 arrives afterward, `MESSAGE_RECEIVED` overwrites m2 without a recency check. The new guard only covers `ROOM_PREVIEW_UPDATED`, and its immutable message `createdAt` also cannot order two edits of the same message. With concurrent broadcast workers or a lost event followed by snapshot reconciliation, the sidebar can remain stale indefinitely. Carry a server-monotonic room revision/sequence on snapshots and all live mutations, and apply every preview writer only when its revision is newer.

```

</details>

chat-frontend/src/context/RoomEventsContext/reducer.js
```diff
@@ -117,6 +130,77 @@ function toSummary(room) {
...
+// to keep the previews reference stable on a same-content write (e.g. the
+// server echo of a message this client already stored optimistically) — a
+// fresh object would invalidate useSidebarSections' memo for every room in
+// the sidebar. createdAt is deliberately NOT part of this comparison: it's
+// metadata for the ROOM_PREVIEW_UPDATED recency guard, not something the UI
+// renders, and two writes sharing the same messageId always share the same
+// createdAt (it's immutable per message) so including it could never change
+// the outcome — it would just be a redundant field-by-field check.
+function samePreview(a, b) {
+  return !!a && !!b && a.messageId === b.messageId && a.senderName === b.senderName && a.text === b.text
```
julianshen, line 176: _🏷️ time-consistency_ | _🟠 Warning_ | _`distributed-system`_

**The server echo does not correct an optimistic client timestamp.**

`MESSAGE_SENT_LOCAL` stores the client's wall-clock `createdAt`, while the server echo carries the canonical timestamp. If the rendered fields match, `samePreview` suppresses that update and retains the client timestamp. A client clock ahead of the server then makes a later authoritative edit or fallback preview appear older to `isOlderPreview`, so the update is rejected. Include `createdAt` in equality so the echo replaces optimistic ordering metadata.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 176: `MESSAGE_SENT_LOCAL` stores the client's wall-clock `createdAt`, while the server echo carries the canonical timestamp. If the rendered fields match, `samePreview` suppresses that update and retains the client timestamp. A client clock ahead of the server then makes a later authoritative edit or fallback preview appear older to `isOlderPreview`, so the update is rejected. Include `createdAt` in equality so the echo replaces optimistic ordering metadata.

```

</details>

chat-frontend/src/context/RoomEventsContext/reducer.js
```diff
@@ -718,9 +845,25 @@ export function roomEventsReducer(state, action) {
...
+      // broader than the server's rule — the server excludes only hidden
+      // (tshow: false) replies; a tshow: true reply can legitimately be a
+      // room's preview. Correct only because this frontend has no tshow
+      // support: no thread reply ever appears in the room timeline here.
+      // Anyone adding tshow must revisit this.
+      // A local optimistic send has no event, so there's no fallback mentions
+      // source — msg.mentions is always undefined for these and the raw
+      // @account renders until the server echo (MESSAGE_RECEIVED) arrives.
+      const nextPreview = msg.threadParentMessageId ? null : previewFromMessage(msg)
+      const previews =
```
julianshen, line 859: _🏷️ failure-handling_ | _🟠 Warning_ | _`distributed-system`_

**A rejected optimistic send remains the room preview.**

`MESSAGE_SENT_LOCAL` promotes an unacknowledged message to the canonical-looking sidebar preview. If the send is definitively rejected, or times out before the server accepts it, no `MESSAGE_RECEIVED` event follows and no failure action restores the previous preview; the row can therefore display a message that was never accepted, without the failed status visible in the timeline. Update the preview only after the server echo/ack, or retain the prior preview and restore or refetch it when the optimistic send fails.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 859: `MESSAGE_SENT_LOCAL` promotes an unacknowledged message to the canonical-looking sidebar preview. If the send is definitively rejected, or times out before the server accepts it, no `MESSAGE_RECEIVED` event follows and no failure action restores the previous preview; the row can therefore display a message that was never accepted, without the failed status visible in the timeline. Update the preview only after the server echo/ack, or retain the prior preview and restore or refetch it when the optimistic send fails.

```

</details>

chat-frontend/src/context/RoomEventsContext/reducer.js
```diff
@@ -117,6 +130,77 @@ function toSummary(room) {
...
+// match (messageId/senderName/text — what RoomList actually displays). Used
+// to keep the previews reference stable on a same-content write (e.g. the
+// server echo of a message this client already stored optimistically) — a
+// fresh object would invalidate useSidebarSections' memo for every room in
+// the sidebar. createdAt is deliberately NOT part of this comparison: it's
+// metadata for the ROOM_PREVIEW_UPDATED recency guard, not something the UI
+// renders, and two writes sharing the same messageId always share the same
+// createdAt (it's immutable per message) so including it could never change
+// the outcome — it would just be a redundant field-by-field check.
+function samePreview(a, b) {
```
julianshen, line 175: _🏷️ mongodb-consistency_ | _🟠 Warning_ | _`mongodb`_

**Rendered equality preserves an unauthoritative client timestamp.**

MESSAGE_SENT_LOCAL stores the client's createdAt, while the server echo carries the authoritative timestamp, but samePreview treats those records as equal and discards the server metadata. With a fast client clock, later Mongo-resolved edit previews compare as older and are rejected even for the same message. Include createdAt and encrypted in semantic equality, or merge authoritative metadata on the server echo; test a same-content local send and echo with different timestamps.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 175: MESSAGE_SENT_LOCAL stores the client's createdAt, while the server echo carries the authoritative timestamp, but samePreview treats those records as equal and discards the server metadata. With a fast client clock, later Mongo-resolved edit previews compare as older and are rejected even for the same message. Include createdAt and encrypted in semantic equality, or merge authoritative metadata on the server echo; test a same-content local send and echo with different timestamps.

```

</details>

chat-frontend/src/context/RoomEventsContext/reducer.js
```diff
@@ -283,13 +373,26 @@ export function roomEventsReducer(state, action) {
           subs[s.id] ? mergeSubscriptionIntoSummary(s, subs[s.id]) : s
         )
       }
+      // Seed only rooms with no preview yet. A live message can land before
+      // fetchSidebarBuckets resolves (the DM subscription goes live first), and
+      // that message is NEWER than this list snapshot — overwriting it would show
+      // an older message in the sidebar than the room itself displays.
```
julianshen, line 379: _🏷️ mongodb-consistency_ | _🟠 Warning_ | _`mongodb`_

**Map presence still substitutes for authoritative preview ordering.**

The prior ordering concern remains: BUCKETS_LOADED ignores a newer subscription.list snapshot whenever any event-derived preview exists, while MESSAGE_RECEIVED can later overwrite that snapshot without comparing ordering metadata. A delayed or concurrently processed room event can therefore preserve or restore an older message than the Mongo latest-message query returned. Carry a server-monotonic room/message revision through both carriers and compare it in bootstrap, live-message, and mutation writers; createdAt alone cannot order edits of the same message.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 379: The prior ordering concern remains: BUCKETS_LOADED ignores a newer subscription.list snapshot whenever any event-derived preview exists, while MESSAGE_RECEIVED can later overwrite that snapshot without comparing ordering metadata. A delayed or concurrently processed room event can therefore preserve or restore an older message than the Mongo latest-message query returned. Carry a server-monotonic room/message revision through both carriers and compare it in bootstrap, live-message, and mutation writers; createdAt alone cannot order edits of the same message.

```

</details>

chat-frontend/src/context/RoomEventsContext/reducer.js
```diff
@@ -117,6 +130,77 @@ function toSummary(room) {
...
+
+// Build a stored preview from a wire PreviewMessage (subscription.list rows
+// and the refreshed preview on edit/delete events).
+function previewFromWire(previewMessage) {
+  if (!previewMessage || !previewMessage.messageId) return null
+  return {
+    messageId: previewMessage.messageId,
+    senderName: participantDisplayName(previewMessage.sender),
+    text: previewSnippet(previewMessage.content, previewMessage.mentions, previewMessage.attachments),
+    createdAt: previewMessage.createdAt,
```
julianshen, line 187: _🔒 security_ | _🟠 Warning_ | _`tgd-review`_

**Cold-start previews expose plaintext for undecryptable E2E messages.**

`previewFromWire` renders `previewMessage.content` directly, and `BUCKETS_LOADED` invokes it without encryption context. The new `RoomPreview.encrypted` documentation explicitly acknowledges that after reload this displays server-provided plaintext even when the timeline would show `[encrypted message]`. Fix the backend contract to encrypt or withhold E2E preview bodies, and avoid seeding such wire previews until the client can decrypt them.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 187: `previewFromWire` renders `previewMessage.content` directly, and `BUCKETS_LOADED` invokes it without encryption context. The new `RoomPreview.encrypted` documentation explicitly acknowledges that after reload this displays server-provided plaintext even when the timeline would show `[encrypted message]`. Fix the backend contract to encrypt or withhold E2E preview bodies, and avoid seeding such wire previews until the client can decrypt them.

```

</details>

PR #283, opened 2026-08-15
feat: harden single-site NATS failure readiness

flywindy:
## Summary

- expand the application-side NATS/JetStream failure evidence from the original four workers to the first single-site room/member campaign: `message-gatekeeper`, `message-worker`, `broadcast-worker`, `notification-worker`, `history-service`, `room-service`, and `room-worker`
- supervise every long-running pull iterator, including notification invalidation, with bounded lookup-only recovery backoff, single-iterator ownership, cancellation-aware shutdown, and recovery metrics
- add bounded inbound request/reply result and duration metrics at the opt-in `natsrouter` boundary, plus bounded outbound request and Core NATS/JetStream publish-attempt evidence for history and room paths
- derive metric `service_name` from `OTEL_SERVICE_NAME`, with distinct normal, bot, and Teams deployment identities and stable instrumentation scopes
- document the new metric families, closed label sets, and consumer-recovery semantics in the o11y inventory

This combines P0-2 and P0-4 for the first single-site failure-readiness round. It intentionally does not add load-generator room/member workloads, dashboards, federation traffic, push traffic, WSS simulation, broker/exporter metrics, or new business retry policies.

Consumer supervision observes and recovers transport failures without changing handler-owned Ack, Nak, Term, redelivery, or publish-retry behavior. A recoverable terminal iterator failure marks the loop down and records bounded terminal evidence; recovery then looks up the existing durable and recreates only its iterator with capped exponential backoff that grows across iterator generations which fail before receiving a message. A successful delivery resets the backoff. A missing or deleted durable is terminal and is never auto-created during recovery because its server-side cursor no longer exists; this prevents `DeliverAllPolicy` replay and `DeliverNewPolicy` data loss even when a concurrent connection outage masks the deletion from the old iterator. Recovery does not intentionally trigger a liveness restart. Shutdown cancels retries, stops the active iterator, and preserves the existing in-flight wait and NATS drain order.

## TDD evidence

Red was captured before each implementation slice:

- supervisor tests initially failed to compile because `NewSupervisor`, `SupervisorConfig`, and recovery-result metrics did not exist
- identity tests initially failed because the affected configs did not expose `OTEL_SERVICE_NAME`
- bounded room/history classification tests initially failed because the closed operations and subject classifiers did not exist
- notification invalidation tests initially failed before the handler was extracted into a testable function
- router tests initially failed because the opt-in metrics boundary did not exist
- history publisher and room outbound-client tests initially failed before their opt-in metrics options existed
- room-key publish tests initially failed before `WithMetrics` and `SendDataContext` existed

Green/refactor coverage includes terminal iterator failure, recreation failure and capped backoff, successful recovery, cancellation, duplicate-start rejection, in-flight shutdown waiting, invalidation dispositions, service identity, and bounded dynamic-subject labels.

Review follow-up Red tests reproduced deleted-durable recreation, asynchronous startup failure, floor-only iterator-flap backoff, the missing `msg.thread.parent` classification, and dynamic `msg`/`teams` token collisions before the fixes. A second Red test reproduced the residual transport-loss race: the initial iterator returned `ErrConnectionClosed`, recovery lookup returned wrapped `ErrConsumerNotFound`, and the test required the supervisor to stop without invoking the create-or-update factory or retrying the missing durable.

## Review follow-up (2026-08-16)

- wrapped the message-gatekeeper delivery callback with `jobguard.Run`, with a regression test proving a panicking poison message is Acked instead of crashing the process
- moved room-service member-list and history-reader request metrics to deferred final-result recording, covering transport, remote errcode, and decode failures without false-success series
- changed room/inbox publish classification to fixed event-token positions so dynamic site IDs cannot mint the wrong bounded operation
- made void-handler logs context-aware and made the asynchronous router metric test wait for deferred recording
- added invalidation Ack-error, room-key publish-error, exact Compose identity, and generated request-recorder mock coverage
- clarified that `service_name` and `site` are deployment identity dimensions and documented the per-recovery `js.Consumer` lookup that precedes iterator creation

Red evidence reproduced the missing gatekeeper guard as a compile failure, the dynamic-site classifier returning `member_publish` for room events, and the stale generated request-recorder mock before implementation and regeneration. Focused race-enabled tests passed after the fixes.
## Validation

- rebased onto `origin/main` at `c2de0ba2`
- `make fmt`
- `make lint`
- `make test`
- focused race-enabled `make test SERVICE=...` runs for all seven services plus `pkg/natsmetrics`, `pkg/natsrouter`, and `pkg/roomkeysender`
- integration-tagged tests for message-gatekeeper, message-worker, broadcast-worker, notification-worker, room-worker, history-service, room-service, and pkg/natsrouter
- `make sast` (`gosec`, `govulncheck`, and Semgrep all passed; no reachable vulnerabilities or findings)
- `make benchmark-natsmetrics`

Benchmark snapshot on Windows/amd64, AMD Ryzen 9 9950X:

```text
BenchmarkConsumer_Track/first_delivery-32    39.70 ns/op    128 B/op    1 allocs/op
BenchmarkConsumer_Track/redelivery-32        55.53 ns/op    144 B/op    2 allocs/op
BenchmarkMessage_Context-32                  22.81 ns/op     48 B/op    1 allocs/op
```

## Coverage status

The changed reusable packages satisfy the repository threshold:

- `pkg/natsmetrics`: 89.43%
- `pkg/natsrouter`: 91.59%
- `pkg/roomkeysender`: 91.30%
- `history-service/internal/publisher`: 83.3%
- `history-service/internal/config`: 93.1%

The integration-tagged full-service package baselines are still below the nominal 80% policy in several pre-existing `package main`/adapter surfaces (`history-service/cmd` 0.0%, `message-gatekeeper` 64.2%, `message-worker` 71.7%, `broadcast-worker` 70.0%, `notification-worker` 63.6%, `room-worker` 77.0%, and `room-service` 79.1%). `history-service/internal/mongorepo` is also 74.0%. The implementation and tests pass, but this policy gap remains explicit; the PR stays draft pending a decision to expand test scope or accept separate baseline remediation.

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive NATS and request metrics, including publish outcomes, request results, recovery activity, and processing durations.
  * Added configurable OpenTelemetry service names through `OTEL_SERVICE_NAME`.
  * Added automatic consumer recovery, coordinated shutdown, and context-aware publishing.
  * Added room invalidation processing with malformed-message handling.
* **Bug Fixes**
  * Improved handling of deleted consumers, interrupted deliveries, failed requests, and publish errors.
  * Added safer processing when handlers panic or invalid messages are received.
* **Documentation**
  * Expanded observability metrics coverage and deployment guidance.

### DIFF
--- a/message-gatekeeper/main.go
+++ b/message-gatekeeper/main.go
@@ -173,7 +173,7 @@ func main() {
 	})
 	var wg sync.WaitGroup
 	supervisor := natsmetrics.NewSupervisor(natsmetrics.SupervisorConfig{})
-	iteratorFactory := func(factoryCtx context.Context) (natsmetrics.Iterator, error) {
+	initialIteratorFactory := func(factoryCtx context.Context) (natsmetrics.Iterator, error) {
 		cons, err := js.CreateOrUpdateConsumer(factoryCtx, messagesCfg.Name, consumerCfg)
 		if err != nil {
 			return nil, fmt.Errorf("create consumer: %w", err)
@@ -184,7 +184,19 @@ func main() {
 		}
 		return iter, nil
 	}
-	if err := supervisor.Start(ctx, iteratorFactory, consumerMetrics, cfg.MaxWorkers, consumerCfg.MaxDeliver, &wg,
+	recoveryIteratorFactory := func(factoryCtx context.Context) (natsmetrics.Iterator, error) {
+		cons, err := js.Consumer(factoryCtx, messagesCfg.Name, consumerCfg.Durable)
+		if err != nil {
+			return nil, fmt.Errorf("lookup consumer: %w", err)
+		}
+		iter, err := cons.Messages(factoryCtx, jetstream.PullMaxMessages(2*cfg.MaxWorkers))
+		if err != nil {
+			return nil, fmt.Errorf("create recovery iterator: %w", err)
+		}
+		return iter, nil
+	}
+	if err := supervisor.StartWithRecovery(ctx, initialIteratorFactory, recoveryIteratorFactory,
+		consumerMetrics, cfg.MaxWorkers, consumerCfg.MaxDeliver, &wg,
 		func(msg jetstream.Msg) natsmetrics.EventType { return natsmetrics.EventTypeFromSubject(msg.Subject()) },
 		func(msgCtx context.Context, msg *natsmetrics.Message) {
 			handlerCtx, _ := natsutil.StampRequestID(msgCtx, msg.Headers(), msg.Subject())
--- a/message-gatekeeper/main.go
+++ b/message-gatekeeper/main.go
@@ -26,6 +26,7 @@ import (
 )
 
 type config struct {
+	ServiceName        string                  `env:"OTEL_SERVICE_NAME" envDefault:"unknown-service"`
 	NatsURL            string                  `env:"NATS_URL,required"`
 	NatsCredsFile      string                  `env:"NATS_CREDS_FILE" envDefault:""`
 	SiteID             string                  `env:"SITE_ID,required"`
@@ -84,7 +85,7 @@ func main() {
 		os.Exit(1)
 	}
 	sharedMetrics := natsmetrics.NewFromProvider(sdk.MeterProvider())
-	publishMetrics := sharedMetrics.Publisher("message-gatekeeper", cfg.SiteID)
+	publishMetrics := sharedMetrics.Publisher(cfg.ServiceName, cfg.SiteID)
 	domainMetrics := newGatekeeperMetrics(sdk.MeterProvider().Meter("message-gatekeeper"))
 
 	nc, err := natsutil.ConnectWithMetrics(ctx, cfg.NatsURL, cfg.NatsCredsFile, sdk.TracerProvider(), sdk.Propagator, sdk.Toggles.Trace, sdk.MeterProvider())
@@ -167,33 +168,33 @@ func main() {
 	messagesCfg := stream.Messages(cfg.SiteID)
 	consumerCfg := buildConsumerConfig(cfg.Consumer)
 	consumerMetrics := sharedMetrics.Consumer(natsmetrics.ConsumerConfig{
-		ServiceName: "message-gatekeeper", Site: cfg.SiteID,
+		ServiceName: cfg.ServiceName, Site: cfg.SiteID,
 		Stream: messagesCfg.Name, Consumer: consumerCfg.Durable,
 	})
-	consumerMetrics.LoopStopped(ctx)
-	cons, err := js.CreateOrUpdateConsumer(ctx, messagesCfg.Name, consumerCfg)
-	if err != nil {
-		slog.Error("create consumer failed", "error", err)
-		os.Exit(1)
-	}
-
-	iter, err := cons.Messages(ctx, jetstream.PullMaxMessages(2*cfg.MaxWorkers))
-	if err != nil {
-		slog.Error("messages failed", "error", err)
-		os.Exit(1)
-	}
-	consumerMetrics.LoopStarted(ctx)
-
 	var wg sync.WaitGroup
-
-	natsmetrics.Start(ctx, iter, consumerMetrics, cfg.MaxWorkers, consumerCfg.MaxDeliver, &wg,
+	supervisor := natsmetrics.NewSupervisor(natsmetrics.SupervisorConfig{})
+	iteratorFactory := func(factoryCtx context.Context) (natsmetrics.Iterator, error) {
+		cons, err := js.CreateOrUpdateConsumer(factoryCtx, messagesCfg.Name, consumerCfg)
+		if err != nil {
+			return nil, fmt.Errorf("create consumer: %w", err)
+		}
+		iter, err := cons.Messages(factoryCtx, jetstream.PullMaxMessages(2*cfg.MaxWorkers))
+		if err != nil {
+			return nil, fmt.Errorf("create iterator: %w", err)
+		}
+		return iter, nil
+	}
+	if err := supervisor.Start(ctx, iteratorFactory, consumerMetrics, cfg.MaxWorkers, consumerCfg.MaxDeliver, &wg,
 		func(msg jetstream.Msg) natsmetrics.EventType { return natsmetrics.EventTypeFromSubject(msg.Subject()) },
 		func(msgCtx context.Context, msg *natsmetrics.Message) {
 			handlerCtx, _ := natsutil.StampRequestID(msgCtx, msg.Headers(), msg.Subject())
 			handlerCtx = logctx.Admit(handlerCtx, msg.Headers())
 			logctx.CapturePayload(handlerCtx, "consumed", msg.Subject(), msg.Data())
 			handler.HandleJetStreamMsg(handlerCtx, msg)
-		})
+		}); err != nil {
+		slog.Error("start consumer supervisor failed", "error", err)
+		os.Exit(1)
+	}
 
 	healthStop, err := health.ServeWithPprof(cfg.HealthAddr, 5*time.Second, cfg.PProfEnabled,
 		natsutil.HealthCheck(nc),
@@ -206,9 +207,8 @@ func main() {
 	slog.Info("message-gatekeeper running", "site", cfg.SiteID)
 
 	shutdown.Wait(ctx, 25*time.Second,
-		func(ctx context.Context) error {
-			consumerMetrics.LoopStopped(ctx)
-			iter.Stop()
+		func(context.Context) error {
+			supervisor.Stop()
 			return nil
 		},
 		func(ctx context.Context) error {
--- a/message-gatekeeper/main.go
+++ b/message-gatekeeper/main.go
@@ -13,6 +13,7 @@ import (
 	"github.com/nats-io/nats.go/jetstream"
 
 	"github.com/hmchangw/chat/pkg/health"
+	"github.com/hmchangw/chat/pkg/jobguard"
 	"github.com/hmchangw/chat/pkg/logctx"
 	"github.com/hmchangw/chat/pkg/model"
 	"github.com/hmchangw/chat/pkg/mongoutil"
@@ -198,12 +199,12 @@ func main() {
 	if err := supervisor.StartWithRecovery(ctx, initialIteratorFactory, recoveryIteratorFactory,
 		consumerMetrics, cfg.MaxWorkers, consumerCfg.MaxDeliver, &wg,
 		func(msg jetstream.Msg) natsmetrics.EventType { return natsmetrics.EventTypeFromSubject(msg.Subject()) },
-		func(msgCtx context.Context, msg *natsmetrics.Message) {
+		guardedGatekeeperProcessor(func(msgCtx context.Context, msg *natsmetrics.Message) {
 			handlerCtx, _ := natsutil.StampRequestID(msgCtx, msg.Headers(), msg.Subject())
 			handlerCtx = logctx.Admit(handlerCtx, msg.Headers())
 			logctx.CapturePayload(handlerCtx, "consumed", msg.Subject(), msg.Data())
 			handler.HandleJetStreamMsg(handlerCtx, msg)
-		}); err != nil {
+		})); err != nil {
 		slog.Error("start consumer supervisor failed", "error", err)
 		os.Exit(1)
 	}
@@ -239,6 +240,12 @@ func main() {
 		func(_ context.Context) error { valkeyutil.Disconnect(metaValkey); return nil },
 		func(ctx context.Context) error { return obsShutdown(ctx) },
 	)
+}
+
+func guardedGatekeeperProcessor(process natsmetrics.ProcessMessage) natsmetrics.ProcessMessage {
+	return func(msgCtx context.Context, msg *natsmetrics.Message) {
+		jobguard.Run(msg, func() { process(msgCtx, msg) })
+	}
 }
 
 // buildConsumerConfig returns the durable consumer config for
--- a/notification-worker/invalidation.go
+++ b/notification-worker/invalidation.go
@@ -8,11 +8,16 @@ import (
 	"github.com/nats-io/nats.go/jetstream"
 
 	"github.com/hmchangw/chat/pkg/model"
+	"github.com/hmchangw/chat/pkg/natsmetrics"
 )
 
 func processInvalidationMessage(ctx context.Context, msg jetstream.Msg, queue chan<- string) {
 	var evt model.CanonicalMemberEvent
 	if err := sonic.Unmarshal(msg.Data(), &evt); err != nil {
+		// Acking a payload that will never decode drops it permanently, and Ack
+		// records a success disposition. Mark the terminal reason so the drop is
+		// visible as loss rather than as normal processing.
+		natsmetrics.MarkTerminalFromContext(ctx, natsmetrics.TerminalInvalidPayload)
 		slog.WarnContext(ctx, "canonical member event decode failed", "error", err)
 		ackInvalidation(ctx, msg)
 		return
--- a/notification-worker/invalidation.go
+++ b/notification-worker/invalidation.go
@@ -1,0 +1,34 @@
+package main
+
+import (
+	"context"
+	"log/slog"
+
+	"github.com/bytedance/sonic"
+	"github.com/nats-io/nats.go/jetstream"
+
+	"github.com/hmchangw/chat/pkg/model"
+)
+
+func processInvalidationMessage(ctx context.Context, msg jetstream.Msg, queue chan<- string) {
+	var evt model.CanonicalMemberEvent
+	if err := sonic.Unmarshal(msg.Data(), &evt); err != nil {
+		slog.WarnContext(ctx, "canonical member event decode failed", "error", err)
+		ackInvalidation(ctx, msg)
+		return
+	}
+	if evt.RoomID != "" {
+		select {
+		case queue <- evt.RoomID:
+		default:
+			slog.WarnContext(ctx, "invalidation queue full, dropping (TTL will reconcile)", "roomId", evt.RoomID)
+		}
+	}
+	ackInvalidation(ctx, msg)
+}
+
+func ackInvalidation(ctx context.Context, msg jetstream.Msg) {
+	if err := msg.Ack(); err != nil {
+		slog.ErrorContext(ctx, "canonical member event ack failed", "error", err)
+	}
+}
--- a/pkg/natsmetrics/loop.go
+++ b/pkg/natsmetrics/loop.go
@@ -113,9 +113,10 @@ type SupervisorConfig struct {
 	wait       waitBackoff
 }
 
-// IteratorFactory recreates both the durable consumer handle and its pull
-// iterator. Callers keep stream, durable, and pull settings closed over in the
-// factory so recovery cannot drift from startup configuration.
+// IteratorFactory creates a durable consumer handle and its pull iterator.
+// Startup factories may create or update the durable; recovery factories must
+// only look up the existing durable so recovery cannot silently replace a lost
+// server-side delivery cursor.
 type IteratorFactory func(context.Context) (Iterator, error)
 
 // Supervisor owns exactly one iterator at a time and recreates it after a
@@ -160,6 +161,14 @@ func NewSupervisor(cfg SupervisorConfig) *Supervisor {
 // process cannot create duplicate pull iterators for the same durable through
 // this supervisor.
 func (s *Supervisor) Start(ctx context.Context, factory IteratorFactory, consumer *Consumer, maxWorkers, maxDeliver int, wg *sync.WaitGroup, classify ClassifyEvent, process ProcessMessage) error {
+	return s.StartWithRecovery(ctx, factory, factory, consumer, maxWorkers, maxDeliver, wg, classify, process)
+}
+
+// StartWithRecovery separates fail-fast startup creation from lookup-only
+// recovery. The recovery factory must never create a missing durable: after a
+// transport failure the iterator cannot reliably report that the durable was
+// concurrently deleted, and recreating it would replay or skip retained work.
+func (s *Supervisor) StartWithRecovery(ctx context.Context, initialFactory, recoveryFactory IteratorFactory, consumer *Consumer, maxWorkers, maxDeliver int, wg *sync.WaitGroup, classify ClassifyEvent, process ProcessMessage) error {
 	if !s.running.CompareAndSwap(false, true) {
 		return ErrSupervisorRunning
 	}
@@ -168,7 +177,7 @@ func (s *Supervisor) Start(ctx context.Context, factory IteratorFactory, consume
 	s.cancel = cancel
 	s.mu.Unlock()
 	consumer.LoopStopped(runCtx)
-	iter, err := factory(runCtx)
+	iter, err := initialFactory(runCtx)
 	if err != nil {
 		s.abortStart(cancel)
 		return fmt.Errorf("create initial consumer iterator: %w", err)
@@ -186,12 +195,12 @@ func (s *Supervisor) Start(ctx context.Context, factory IteratorFactory, consume
 	go func() {
 		defer wg.Done()
 		defer s.finish(runCtx, consumer)
-		s.run(runCtx, iter, factory, consumer, maxDeliver, wg, sem, classify, process)
+		s.run(runCtx, iter, recoveryFactory, consumer, maxDeliver, wg, sem, classify, process)
 	}()
 	return nil
 }
 
-func (s *Supervisor) run(ctx context.Context, iter Iterator, factory IteratorFactory, consumer *Consumer, maxDeliver int, wg *sync.WaitGroup, sem chan struct{}, classify ClassifyEvent, process ProcessMessage) {
+func (s *Supervisor) run(ctx context.Context, iter Iterator, recoveryFactory IteratorFactory, consumer *Consumer, maxDeliver int, wg *sync.WaitGroup, sem chan struct{}, classify ClassifyEvent, process ProcessMessage) {
 	backoff := s.minBackoff
 	recovered := false
 	for {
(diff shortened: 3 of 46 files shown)


flywindy:
## Review

Reviewed at `939ae31` against base `c9f1351` (44 files). Built the repo and ran `go vet` plus `go test -race` on `pkg/natsmetrics`, `pkg/natsrouter`, and `pkg/roomkeysender` — all clean, so everything below is behavioural rather than a build or test failure.

Two findings I'd treat as blockers. Both have the same shape: the new self-healing machinery introduces a data-correctness risk that did not exist when the loop simply stopped.

### Consumer recreation replays the entire stream

`Supervisor.run` recreates a lost consumer through each service's factory, and every factory calls `js.CreateOrUpdateConsumer`. When the durable still exists this is an update and the cursor is untouched. When the durable has been **deleted** — `jetstream.ErrConsumerDeleted`, the exact error the supervisor classifies as recoverable — it is a fresh creation, and `pkg/stream/consumer.go` documents what that means:

> `DeliverPolicy=All` so a freshly-created durable (new deploy, new site, **or a deleted-and-recreated durable**) **replays the stream from the start**.

So on the one failure this supervisor was built to survive, the four workers built on `DurableConsumerDefaults` replay their whole retention window: `broadcast-worker` re-fans-out to connected clients, `notification-worker` re-notifies, `message-worker` re-persists, `message-gatekeeper` re-validates.

Before this PR that path ended with `loop.up=0` — an alert and a human decision. Now it silently self-heals into a full replay.

Worth noting the mirror image on the same mechanism: `notification-worker`'s invalidation consumer (`main.go:271`) is `DeliverNewPolicy`, and this PR puts it under the same supervisor. On a deleted-and-recreated durable that one **silently skips** everything pending instead. One supervisor, two opposite correctness failures.

Changing `DeliverPolicy` is probably the wrong fix — `search-sync-worker`'s MV rebuild and `inbox-worker`'s federated catch-up depend on `All` at creation. Not auto-recreating on `ErrConsumerDeleted`, and keeping the stop-and-alert behaviour for that specific cause, preserves both.

### Consumer startup no longer fails fast

The `os.Exit(1)` at each call site is still present (e.g. `broadcast-worker/main.go:251-256`) — the issue is that it has become unreachable for creation failures.

`Start` returns an error only for a duplicate start (`ErrSupervisorRunning`). The factory now runs inside the supervisor goroutine, so a consumer or iterator creation failure records `RecoveryAttempt(RecoveryFailure)`, logs a warning, and retries forever. Since `/healthz` only checks the NATS connection, a production deployment with `BOOTSTRAP_STREAMS=false` against a missing stream stays **green and idle indefinitely** rather than crash-looping.

The fix belongs in `Start`'s error semantics rather than in restoring the exit: the first creation, before any loop has run, is a startup failure and should be reported as one. It is also currently recorded as a *recovery* failure, which it isn't — nothing had recovered yet.

### Smaller findings

- **`pkg/natsmetrics/subject.go:68` — `msg.thread.parent` classifies as `operation="unknown"`.** `history-service/internal/service/service.go:198` registers `MsgThreadParentPattern`. The `.msg.` branch's read-suffix list contains `.msg.thread`, which is not a suffix of `…msg.thread.parent`. Since a matched outer `case` does not fall through, control leaves the switch entirely and hits the trailing `return OperationUnknown`, while every sibling history route is labelled `history_read`.
- **`pkg/natsmetrics/subject.go:71` — unanchored `Contains` on `.msg.` and `.teams.`.** Any dynamic token equal to `msg` or `teams` hijacks classification; an account literally named `teams` would report all of its `member.add` calls as `operation="teams_room"`. The relative order of the two cases is also load-bearing without saying so.
- **`pkg/natsmetrics/loop.go:212` — backoff resets on every iterator loss.** Backoff only grows across consecutive `factory()` failures; `backoff = s.minBackoff` runs after a loop failure. A create-succeeds-then-`Next`-fails-immediately condition (raft leadership flap) recreates at the 250 ms floor indefinitely, hammering the JetStream API during exactly the degradation the supervisor exists to ride out.

### Checked and clear

Supervisor iterator ownership and `Stop` races (no double-`Stop`, no deadlock, no semaphore leak on the `sem`-versus-`ctx.Done` select); defer ordering in `natsrouter.addRoute` (the metrics defer runs after the panic backstop and before `releaseContext`); `Publisher`/`Consumer` attribute-slice construction (no aliasing between precomputed sets); nil-guards on the zero-value `Publisher`; `PublishLabelsFromSubject` against every subject `room-service`, `room-worker`, and `history-service` actually publish; `OTEL_SERVICE_NAME` present in all seven services' compose files.

---

flywindy:
Addressed all five findings in `d9e83296`.

- Initial consumer/iterator creation is synchronous again. `Start` now returns the initial factory error, so existing startup paths fail fast; initial startup failures do not emit recovery attempts.
- `jetstream.ErrConsumerDeleted` is terminal and is not auto-recreated. The loop still records terminal failure/down evidence without risking a `DeliverAll` replay or a `DeliverNew` gap.
- Recovery backoff now grows across iterator generations that fail before delivering a message, and resets only after a successful delivery.
- `msg.thread.parent` is classified as `history_read`.
- Request classification now uses fixed subject-token positions, so dynamic account or room tokens named `msg` or `teams` cannot hijack bounded labels.

The tests were written and observed failing before the implementation changes. Validation completed with `make fmt`, `make lint`, full `make test`, focused race tests, relevant integration tests for all affected services plus `pkg/natsrouter`, `make sast`, and `make benchmark-natsmetrics`. Changed-package coverage for `pkg/natsmetrics` is 89.28%.

The PR remains draft because the already-documented pre-existing full-service package coverage baselines still require an explicit disposition before merge readiness.

flywindy:
## Re-review at `d9e8329`

All five findings are addressed, and the two blockers are fixed the way I'd have picked. Verified with `make lint` (clean), `go build ./...`, and `go test -race` on `pkg/natsmetrics`, `pkg/natsrouter`, and all five workers — all passing.

**Consumer recreation.** `ErrConsumerDeleted` is now terminal and the loop stays down. I checked that this is the right error to key on rather than a guess: `jetstream/pull.go:720` is where the pull iterator turns the server's status message into a terminal error, and `ErrConsumerDeleted` is exactly what it surfaces. The gauge and terminal metric both land correctly — `consume` calls `LoopFailed` (gauge to 0, `consumer_deleted` recorded) and the deferred `finish` is a no-op second `LoopStopped` rather than a double decrement. The doc paragraph explaining *why* it isn't recreated, covering both the `DeliverAllPolicy` replay and the `DeliverNewPolicy` skip, is the right thing to leave behind.

**Fail-fast startup.** `Start` now creates the first iterator synchronously and returns `create initial consumer iterator: %w`, which makes the pre-existing `os.Exit(1)` at each call site reachable again. Leaving the five `main.go` files untouched is correct — they were already checking the error. The spurious `RecoveryAttempt(RecoveryFailure)` on a first-ever creation failure is gone with it.

**Subject classification.** `.msg.thread.parent` resolves to `history_read` now, and the token anchoring replaces both unanchored `Contains` calls. I checked the anchoring against the real builders rather than the tests, since that is where a regression would hide — `isRequestFamily` covers `chat.user.{account}.request.room.{roomID}.{siteID}.msg.thread.parent` and both teams shapes (`…request.room.{roomID}.{siteID}.teams.call` via the `tokens[4]=="room" && tokens[7]==family` arm, `…request.teams.{siteID}.call.user` via `tokens[4]==family`), and `isMigrationFamily` covers `chat.migration.internal.{siteID}.msg.edit`/`.delete`. Nothing that previously classified has been dropped. As a side effect the publish-side `chat.teams.room.canonical.…` subjects no longer false-positive as `teams_room`, which the old `Contains` did match.

**Backoff.** Growth across consecutive generations that fail before receiving anything, reset on delivery, is the right policy and closes the flap case. The "one message then die" pathology would still reset to the floor, but treating actual progress as a reset is the conventional trade and I would not change it.

### One residual gap

The guard keys on the error the iterator reports, so it only fires when the durable disappears *while the loop is reading it*. If a durable is deleted during a connection outage, the iterator fails with `ErrConnectionClosed` instead, the supervisor treats it as recoverable, and the factory's `CreateOrUpdateConsumer` then silently **creates** the missing durable — replaying the retention window, which is the outcome this commit set out to prevent.

Narrow (it needs a deletion concurrent with a transport failure) and not something I'd block on. If you want it closed, the check belongs in the factory rather than the supervisor: look the consumer up first and treat "absent" as terminal, so recreation can never turn into creation regardless of which error triggered it.

---

flywindy:
## Re-review at `92c41de`, plus two fixes pushed in `8324126`

`make lint` clean, full `make test` green, and `-race` on every package I touched.

### The recovery fix is right, and closes the gap I raised

`StartWithRecovery` splitting the initial factory from the recovery factory is the correct layer for this — the constraint belongs where the consumer handle is obtained, not in the supervisor's error classification. I checked the six supervised consumers rather than the helper alone: all five workers plus notification's invalidation consumer use `js.Consumer(...)` on the recovery path, so recovery can look up but never create. `isMissingConsumer` covering `ErrConsumerNotFound` as well as `ErrConsumerDeleted` is what makes the connection-outage case safe, and the `fmt.Errorf("lookup consumer: %w", err)` wrap keeps `errors.Is` working through it.

The token-anchored member detection in `PublishLabelsFromSubject` I verified against the builders rather than the tests, since that is where an anchoring regression hides: index 4 for `chat.room.canonical.{site}.event.member.…`, index 5 for `chat.teams.room.canonical.{site}.…`, and the `member_` prefix at index 4 for `chat.inbox.{site}.external.{eventType}`. All correct.

### Two evidence defects I fixed

**`room-service` counted a business rejection as a failed request.** `ListMembers` recorded `errNotRoomMember` through the new named-return defer. That is a complete answer from a healthy peer — the request/reply exchange worked — and `chat.nats.requests` carries a transport-shaped outcome enum, so the rejection landed in `other_error` and left the family non-zero at baseline. That is the same trap the metrics contract calls out for terminal failures, one family over. `GetMessageReadMeta` in the same commit already excludes its `CodeNotFound` for exactly this reason, so the two call sites disagreed; `ListMembers` now follows the same rule and `other_error` keeps meaning "the transport failed in a way we could not classify".

To be explicit about what this does *not* revert: transport failures, decode failures, and unexpected remote errcodes still count as failures, so the masking problem the earlier fix addressed stays fixed.

**`notification-worker` dropped poison payloads invisibly.** `processInvalidationMessage` acks an undecodable canonical member event, and `Message.Ack` records `OutcomeAck` — a success. A permanently discarded payload was therefore indistinguishable from normal processing. It now marks `invalid_payload` first, matching the drop paths in `broadcast-worker`.

I deliberately left the queue-full drop unmarked. That one is designed backpressure with the `roomsubcache` TTL as the reconciler, not a terminal failure, and marking it would put expected load-shedding into the loss signal.

### On the coverage question

I measured it rather than reasoning about it, comparing the same method on both sides:

| package | base | this branch | delta |
|---|---|---|---|
| `message-gatekeeper` | 63.7% | 63.7% | — |
| `message-worker` | 49.7% | 50.8% | +1.1 |
| `broadcast-worker` | 62.2% | 62.3% | +0.1 |
| `notification-worker` | 53.1% | 56.2% | +3.1 |
| `room-worker` | 63.5% | 63.2% | −0.3 |

This branch does not regress the baseline; it improves it slightly. The sub-threshold numbers are genuinely pre-existing, so the disposition this PR was waiting on is a standing question about the repository's baseline debt rather than anything this change introduced. Holding a non-regressing PR against debt it did not create only makes that debt harder to pay down, so I've moved this out of draft and would suggest tracking the baselines separately.

---

flywindy:
CI's `test` job went red on `8324126` — `TestSupervisor_RecreationFailuresUseCappedBackoff` expected two recovery datapoints and saw only the `failure` one. Fixed in `7c40831`.

It is a timing flake, not a bad assertion. `blockingSupervisorIterator` signals `created` from *inside* the factory, so the test's `<-recovered` can wake and call `Stop()` before the supervisor reaches `activate` — which then finds the context cancelled and returns, and `RecoverySuccess` is never recorded. I could not reproduce it locally in 150 runs including `GOMAXPROCS=1`, so I confirmed it by elimination instead: the `waits` assertion passed, which means the full 100/200/400/400 sequence ran and the fifth factory call succeeded, and `failure=3` accounts for calls two through four. Between "factory returned an iterator" and "success recorded" the only path that skips the record is that `activate` return.

Rather than loosen the assertion, I moved where the metric is recorded: `RecoverySuccess` now fires where the factory returns, symmetrically with the `RecoveryFailure` recorded where it errors. That makes the counter describe the recreation attempt itself, which is what the inventory already documents ("Each failed or successful recreation increments…"), and it no longer depends on the loop surviving long enough to start another generation. Steady-state counts are unchanged — every generation after the first is entered through exactly one successful recreation.

The regression test cancels inside the factory so the window is hit deterministically instead of by scheduling luck. I verified it is not vacuous: it fails with the fix reverted.

`make lint` clean, full `make test` green, and `pkg/natsmetrics` run 20× under `-race`.

---

flywindy:
## Parked as draft — this PR is being narrowed to the supervisor

This change is two things at once, and only one of them is metrics. The observation half has been split out into #286; this PR keeps the half that changes runtime behaviour:

- the consumer supervisor and automatic iterator recovery (`pkg/natsmetrics/loop.go`, `StartWithRecovery`, the recovery factories in five workers plus notification invalidation)
- `chat.nats.consumer.recovery.attempts`, which only exists because the supervisor does
- `jobguard.Run` on the message-gatekeeper callback — a handler panic previously crashed the process; it is now recovered and the message Acked
- the notification-worker invalidation restructure
- `room-worker`'s NATS metrics adoption and the Teams deployment identity change

That grouping is not arbitrary: every substantive finding on this PR — full-stream replay on recreation, startup no longer failing fast, a durable deleted behind a transport error, the flaky recovery counter — came from the supervisor. The review history here is worth keeping attached to it, which is why the supervisor stayed with this number rather than moving to the new one.

**No action needed on the branch right now.** It still contains both halves; once #286 merges, rebasing onto `main` collapses this diff to the supervisor alone with no manual surgery.

Two things carried over deliberately rather than being pulled into #286:

`room-worker` has no NATS metrics in `main` and adopts them here *through* the supervisor. Instrumenting it in #286 would have meant rewriting its hand-rolled consume loop onto `natsmetrics.Start` — code in neither PR, which would put new risk into the half meant to be low-risk. `pkg/roomkeysender` follows it, since `room-worker` is its only instrumented caller.

That also means the Teams `room-worker` service-name change stays here, so #286 does not move `service_name` on any currently-scraped series — the deployment-identity change lands together with the rest of room-worker's onboarding.

---

julianshen:
**9 findings · 6 unique issues · 6 inline comments posted** — 🔴 2 blocking · 🟠 7 warning

<details>
<summary>📒 Files reviewed (46)</summary>

* `Makefile`
* `broadcast-worker/consumeloop_test.go`
* `broadcast-worker/main.go`
* `history-service/internal/publisher/publisher_test.go`
* `message-gatekeeper/config_test.go`
* `message-gatekeeper/nats_metrics_test.go`
* `message-worker/config_test.go`
* `message-worker/main.go`
* `notification-worker/invalidation.go`
* `notification-worker/invalidation_test.go`
* `pkg/natsmetrics/disposition_test.go`
* `pkg/natsmetrics/loop.go`
* `pkg/natsmetrics/loop_test.go`
* `pkg/natsmetrics/metrics_benchmark_test.go`
* `pkg/natsmetrics/metrics_test.go`
* `pkg/natsmetrics/supervisor_test.go`
* `pkg/natsrouter/context.go`
* `pkg/natsrouter/router.go`
* `pkg/roomkeysender/roomkeysender.go`
* `pkg/roomkeysender/roomkeysender_test.go`
* `room-service/main.go`
* `room-worker/deploy/teams/docker-compose.yml`
* `room-worker/handler.go`
* `room-worker/main.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

message-gatekeeper/main.go
```diff
@@ -229,6 +242,12 @@ func main() {
 	)
 }
 
+func guardedGatekeeperProcessor(process natsmetrics.ProcessMessage) natsmetrics.ProcessMessage {
+	return func(msgCtx context.Context, msg *natsmetrics.Message) {
+		jobguard.Run(msg, func() { process(msgCtx, msg) })
```
julianshen, line 247: _🏷️ delivery-guarantees_ | _🔴 Blocking_ | _`distributed-system`, `nats`, `silent-failure` (3 rules)_

**Handler panics acknowledge and permanently discard valid messages.**

The added guard acknowledges every message whose handler panics; the new test explicitly verifies this behavior. A valid send can be delivered, panic before `HandleJetStreamMsg` completes, and then be Acked, so JetStream will not redeliver it and the user operation is lost. A panic is not evidence of a deterministic poison payload and may also occur after partial side effects. Recover by Nak-ing with bounded backoff, relying on idempotency for partial retries, and only Ack/Term after explicitly classifying a payload as permanently invalid.

<details>

- **`silent-failure`**:244: The new guard passes every handler panic to jobguard.Run, and TestGuardedGatekeeperProcessor_AcksPanickingMessage confirms that this Acks the JetStream message. A panic can occur before publishing or after only partial processing, so the broker permanently removes an uncompleted send event instead of redelivering it. Recover panics by Nak/NakWithDelay and let MaxDeliver enforce a terminal poison-message policy; only Ack failures proven to be deterministic invalid input.
- **`nats`**:246: `guardedGatekeeperProcessor` passes every panic to `jobguard.Run`, and the added test confirms that this Acks the message. A panic before gatekeeper processing or downstream publication completes therefore removes the message from JetStream with no redelivery. Panics are not necessarily deterministic poison payloads. Recover generic panics with NAK/NakWithDelay and let MaxDeliver bound retries; Ack or Term only errors proven permanent, such as payload decoding failures.

</details>

<details>

```
In `message-gatekeeper/main.go` around line 247: The added guard acknowledges every message whose handler panics; the new test explicitly verifies this behavior. A valid send can be delivered, panic before `HandleJetStreamMsg` completes, and then be Acked, so JetStream will not redeliver it and the user operation is lost. A panic is not evidence of a deterministic poison payload and may also occur after partial side effects. Recover by Nak-ing with bounded backoff, relying on idempotency for partial retries, and only Ack/Term after explicitly classifying a payload as permanently invalid.

```

</details>

pkg/natsmetrics/loop.go
```diff
@@ -67,3 +95,236 @@ func Consume(ctx context.Context, iter Iterator, consumer *Consumer, maxWorkers,
...
+		wait = waitForBackoff
+	}
+	return &Supervisor{minBackoff: minBackoff, maxBackoff: maxBackoff, wait: wait}
+}
+
+// Start creates the first consumer iterator synchronously so startup remains
+// fail-fast. After that first successful creation it registers the supervisor
+// loop with wg before returning. A second concurrent Start is rejected so one
+// process cannot create duplicate pull iterators for the same durable through
+// this supervisor.
```
julianshen, line 162: _🏷️ delivery-guarantees_ | _🟠 Warning_ | _`distributed-system`, `tgd-review` (2 rules)_

**Start reuses a create-capable factory during recovery.**

`Start` passes the same factory as both the initial and recovery factory, although `IteratorFactory` explicitly permits startup factories to create or update a durable. If the iterator reports a transport error while the durable is concurrently deleted, recovery can call `CreateOrUpdateConsumer`, silently create a new cursor, and either replay retained messages or skip pending messages depending on the delivery policy. Remove this unsafe convenience method or require a separate lookup-only recovery factory, as the production `StartWithRecovery` call sites already do.

<details>

- **`tgd-review`**:164: This aliases the startup factory as the recovery factory, despite the adjacent contract requiring recovery to only look up an existing durable. If a caller supplies the same CreateOrUpdate-style startup factory used by the services, a transport error masking concurrent durable deletion will invoke that factory during recovery and recreate the durable with a lost cursor. Remove this convenience method or require separate initial and lookup-only recovery factories, as StartWithRecovery does.

</details>

<details>

```
In `pkg/natsmetrics/loop.go` around line 162: `Start` passes the same factory as both the initial and recovery factory, although `IteratorFactory` explicitly permits startup factories to create or update a durable. If the iterator reports a transport error while the durable is concurrently deleted, recovery can call `CreateOrUpdateConsumer`, silently create a new cursor, and either replay retained messages or skip pending messages depending on the delivery policy. Remove this unsafe convenience method or require a separate lookup-only recovery factory, as the production `StartWithRecovery` call sites already do.

```

</details>

notification-worker/invalidation.go
```diff
@@ -0,0 +1,39 @@
...
+		// visible as loss rather than as normal processing.
+		natsmetrics.MarkTerminalFromContext(ctx, natsmetrics.TerminalInvalidPayload)
+		slog.WarnContext(ctx, "canonical member event decode failed", "error", err)
+		ackInvalidation(ctx, msg)
+		return
+	}
+	if evt.RoomID != "" {
+		select {
+		case queue <- evt.RoomID:
+		default:
```
julianshen, line 28: _🏷️ observability_ | _🟠 Warning_ | _`distributed-system`_

**Queue-full invalidations are acknowledged as successful processing.**

When the invalidation queue is full, the default branch drops the room ID and then falls through to `Ack`. In the new tracked consumer path this records a successful Ack, unlike malformed payloads which receive an explicit terminal marker. During cache-refresher slowness, a valid mute update is therefore permanently removed from JetStream, the cache remains stale until TTL reconciliation, and consumer metrics hide the drop. If TTL-based shedding is intentional, mark a bounded backpressure-drop terminal reason before Ack; otherwise Nak with delay and Ack only after enqueue succeeds.

<details>

```
In `notification-worker/invalidation.go` around line 28: When the invalidation queue is full, the default branch drops the room ID and then falls through to `Ack`. In the new tracked consumer path this records a successful Ack, unlike malformed payloads which receive an explicit terminal marker. During cache-refresher slowness, a valid mute update is therefore permanently removed from JetStream, the cache remains stale until TTL reconciliation, and consumer metrics hide the drop. If TTL-based shedding is intentional, mark a bounded backpressure-drop terminal reason before Ack; otherwise Nak with delay and Ack only after enqueue succeeds.

```

</details>

pkg/natsmetrics/loop.go
```diff
@@ -36,21 +42,43 @@ func Start(ctx context.Context, iter Iterator, consumer *Consumer, maxWorkers, m
...
 		if err != nil {
+			if ctx.Err() != nil {
+				consumer.LoopStopped(ctx)
+				return received, ctx.Err()
+			}
 			consumer.LoopFailed(ctx, err)
-			return
+			return received, err
+		}
+		received = true
```
julianshen, line 69: _🏷️ JetStream disposition_ | _🟠 Warning_ | _`nats`_

**Shutdown leaves a delivered message pending until AckWait.**

When all worker slots are occupied, `Next` can deliver another message before dispatch blocks on `sem`. If shutdown then cancels the context, this branch calls `Finish` without Ack, NAK, or Term. The durable therefore retains the message as ack-pending until AckWait expires, delaying failover to another instance. NAK the undispatched message while the NATS connection is still available.

<details>

```
In `pkg/natsmetrics/loop.go` around line 69: When all worker slots are occupied, `Next` can deliver another message before dispatch blocks on `sem`. If shutdown then cancels the context, this branch calls `Finish` without Ack, NAK, or Term. The durable therefore retains the message as ack-pending until AckWait expires, delaying failover to another instance. NAK the undispatched message while the NATS connection is still available.

```

</details>

pkg/natsmetrics/loop.go
```diff
@@ -67,3 +95,236 @@ func Consume(ctx context.Context, iter Iterator, consumer *Consumer, maxWorkers,
...
+
+		for {
+			if err := s.wait(ctx, backoff); err != nil {
+				return
+			}
+			iter, err = recoveryFactory(ctx)
+			if err == nil {
+				// Recorded here, symmetrically with the RecoveryFailure below, so
+				// the counter describes the recreation attempt itself. Deferring it
+				// until the next generation starts would drop the success whenever
```
julianshen, line 232: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Deterministic recovery backoff synchronizes all consumer replicas.**

A NATS outage makes every supervisor use the same exact 250ms, 500ms, 1s, 2s, 4s, then 5s schedule. Replicas that lose their iterators together will therefore call the recovery factory together, concentrating JetStream consumer lookups and iterator creation on a broker that is already recovering. At production replica counts this thundering herd can increase broker load and delay recovery. Apply bounded jitter to each context-cancellable wait, such as choosing a delay in [backoff/2, backoff], while retaining the existing exponential cap.

<details>

```
In `pkg/natsmetrics/loop.go` around line 232: A NATS outage makes every supervisor use the same exact 250ms, 500ms, 1s, 2s, 4s, then 5s schedule. Replicas that lose their iterators together will therefore call the recovery factory together, concentrating JetStream consumer lookups and iterator creation on a broker that is already recovering. At production replica counts this thundering herd can increase broker load and delay recovery. Apply bounded jitter to each context-cancellable wait, such as choosing a delay in [backoff/2, backoff], while retaining the existing exponential cap.

```

</details>

notification-worker/invalidation.go
```diff
@@ -0,0 +1,39 @@
...
+		// Acking a payload that will never decode drops it permanently, and Ack
+		// records a success disposition. Mark the terminal reason so the drop is
+		// visible as loss rather than as normal processing.
+		natsmetrics.MarkTerminalFromContext(ctx, natsmetrics.TerminalInvalidPayload)
+		slog.WarnContext(ctx, "canonical member event decode failed", "error", err)
+		ackInvalidation(ctx, msg)
+		return
+	}
+	if evt.RoomID != "" {
+		select {
```
julianshen, line 26: _🏷️ reliability_ | _🟠 Warning_ | _`silent-failure`_

**Missing room IDs are silently acknowledged without invalidation.**

A payload such as `{}` decodes successfully, leaves RoomID empty, skips the invalidation queue, and is then Acked as normal success. This permanently hides a malformed canonical event and leaves cached mute state stale until TTL reconciliation. Treat an empty room ID like the already-handled decode failure: record invalid_payload, log it, and Ack-drop it explicitly.

<details>

```
In `notification-worker/invalidation.go` around line 26: A payload such as `{}` decodes successfully, leaves RoomID empty, skips the invalidation queue, and is then Acked as normal success. This permanently hides a malformed canonical event and leaves cached mute state stale until TTL reconciliation. Treat an empty room ID like the already-handled decode failure: record invalid_payload, log it, and Ack-drop it explicitly.

```

</details>

julianshen (commented):
tGD inline review

PR #284, opened 2026-08-16
feat(mongoutil): let online services boot during a MongoDB outage

hmchangw:
## Summary

Online services currently cannot **start** during a MongoDB outage. Two independent startup steps are fatal in every service:

1. `mongoutil.Connect` pings MongoDB and every caller treats failure as fatal (`mongo.Connect` is lazy in driver v2, so this ping *is* the reachability check).
2. `EnsureIndexes` at startup — `createIndexes` is a write, so it fails against a MongoDB that is down *or* degraded to read-only.

A pod that restarts mid-outage (deploy, node drain, OOM kill, HPA scale-up) therefore dies at boot and stays down for the rest of the outage plus the CrashLoopBackOff backoff after MongoDB returns. Across 22 services, a one-hour outage will almost certainly include at least one such restart — which means the request-path resilience work is worth nothing to a process that can't boot.

This PR lets a restarted pod boot, wire its stores, and reach a serving (degraded) state — **without** discarding the startup failures that no restart can fix.

## What changed

**`pkg/mongoutil`** — two additions, both leaving current behaviour as the default:

| Addition | Effect |
|---|---|
| `WithLazyConnect()` | Skips the startup ping so `Connect` returns a client without proving MongoDB is reachable. Lazy boots log at **WARN** since the connection is unproven. |
| `EnsureIndexes(ctx, steps...)` | Runs named index steps concurrently under one 30s budget, and splits failures by whether a restart can fix them (below). |

### Transient vs permanent index failures

This is the part worth reviewing. `EnsureIndexes` classifies:

- **Transient** — MongoDB unreachable or read-only. Logged at WARN with the step name and impact, reported as success, service boots. Self-heals on a later boot, since the step is idempotent and re-runs every start.
- **Permanent** — `E11000` (existing duplicates block a unique index), `85`/`86` (`IndexOptionsConflict` / `IndexKeySpecsConflict`). Returned to the caller, which exits as before.

Permanent failures have to stay fatal: `user-service/mongorepo/users.go` and `room-service/store_mongo.go` detect exactly these codes and return operator remediation text (*"drop the old non-unique `account_1` index … before starting this service"*). Nothing about a restart clears them, and the unique indexes involved are what room-worker's `$setOnInsert` retry-idempotency depends on — silently serving without them lets redelivered `member_added` events duplicate rows.

Steps run **concurrently** so a service with several stores stalls a boot once, not once per store — user-service has five, which would otherwise burn ~150s of an outage before its health server binds, long enough for a k8s startup probe to kill the pod and recreate the crash loop this PR removes.

**Call sites**

- `WithLazyConnect()` on **22 long-running online services** — the 21 online services plus `hr-sync-worker`, which has a health server, durable JetStream consumers and graceful shutdown, so it meets the documented eligibility bar.
- **18 index steps across 12 services** now go through `EnsureIndexes`. `media-service` is included: it had received `WithLazyConnect()` while keeping a fatal `EnsureEmojiIndexes`, so the change was a no-op there and it still crash-looped.
- `message-worker` and `inbox-worker` **create the new `thread_subscriptions` unique index before retiring the legacy one**. Dropping first was safe while a failed create was fatal; now that it isn't, that order could leave the collection with neither index while the service kept serving.
- `botplatform-service` no longer gates `/healthz` on a Mongo ping. A lazily-booted pod failed liveness and crash-looped through the outage anyway, and `docs/health-probes.md` already requires liveness to probe no dependency. The store's now-unused `Ping` is removed.

**Deliberately not opted in:** `data-migration/*`, `teams-*`, `tools/*`, `seed-sample-data`. Batch, migration and CLI jobs have no useful degraded mode — a silent lazy start would turn a clear startup failure into a confusing one deep into a run. They keep the ping and keep failing loudly.

## Behaviour notes

- **Lazy does not defer failure indefinitely.** The first operation still fails, bounded by the caller's context deadline or the driver's 30s server-selection timeout, whichever is shorter — reads return an error rather than hanging, so consumer loops can Nak and retry rather than wedge.
- **No background index retry, on purpose.** The step re-runs on every boot; a retry loop would add a goroutine and a termination path per service for a condition a restart fixes.
- **A recovered client needs no restart.** Once MongoDB returns, the same lazily-connected client works again (covered by an integration test that proxies an outage and then restores it).

## Risk

Unique indexes are the caveat that remains. If a service boots while MongoDB is unwritable, a *transient* failure means the index isn't created, and between MongoDB becoming writable again and the next restart, duplicates it would have rejected can be written. Permanent failures no longer fall into this gap — they stop the service as before.

The WARN line carries the step name and impact so it can be alerted on:

```
level=WARN msg="ensure indexes failed; continuing without them"
  indexes="user-service subscriptions" error="..."
  impact="degraded query performance; unique constraints unenforced until a later start succeeds"
```

## Testing

- `pkg/mongoutil/lazy_test.go` — option plumbing, ping-is-still-the-default, lazy succeeds against an unreachable port, client is usable, first operation fails in **bounded** time.
- `pkg/mongoutil/indexes_test.go` — success is quiet; transient failure warns and reports success; permanent failure (E11000, 85, 86) is returned and *not* downgraded; steps run concurrently; the helper owns a deadline and never loosens a tighter caller one.
- `pkg/mongoutil/lazy_integration_test.go` — an outage proxy boots a client against a dead endpoint, fails bounded while down, then **recovers with no restart** once MongoDB returns.
- `broadcast-worker/startup_resilience_integration_test.go` — walks the service's real Mongo boot sequence with MongoDB down (connect → build store → index step) and asserts all three complete.

Locally: `go build ./...` clean, `make lint` → 0 issues, `make test` → 108 packages ok. Integration suites not run (need Docker).

## Follow-ups not in this PR

- Index bootstrap still writes from app code on every start. The repo's established pattern for ops/IaC-owned provisioning is `BOOTSTRAP_STREAMS`-style env gating (create in dev, verify-and-fail-fast in production); indexes arguably belong there too.
- Readiness probes NATS only, so a lazily-booted pod reports ready while unable to serve Mongo-backed requests. That follows the documented policy in `docs/health-probes.md`, but the doc's supporting argument predates pods that have never reached Mongo at all.
- Workers that boot during an outage now consume during it. `inbox-worker`'s bare `msg.Nak()` against `MaxDeliver=5` burns all redeliveries in milliseconds, and `broadcast-worker`'s backoff gives ~36s of retry, so a longer outage drops those events where previously the pod wasn't running to receive them.
- Nothing enforces the service-vs-job choice; a `.semgrep` rule would stop service #23 silently picking the wrong one.

---

Supersedes #270 (same commits, same content).

---
_Generated by [Claude Code](https://claude.ai/code/session_01KDzjqaoGFtNaRdyTLBDSMf)_

## Summary by CodeRabbit

* **New Features**
  * Services can now start while MongoDB is temporarily unavailable and reconnect automatically when it recovers.
  * Added standardized index initialization with bounded execution and improved handling of temporary failures.
  * Added safer index migration support, including reliable removal of legacy indexes.

* **Bug Fixes**
  * Health checks now report service availability without requiring MongoDB to be reachable.
  * Workers can recover database operations and retry message processing without restarting.

### DIFF
--- a/bot-room-service/main.go
+++ b/bot-room-service/main.go
@@ -68,7 +68,8 @@ func run() error {
 		return fmt.Errorf("init jetstream: %w", err)
 	}
 
-	mc, err := mongoutil.Connect(ctx, cfg.MongoURI, cfg.MongoUsername, cfg.MongoPassword, mongoutil.WithObservability(sdk))
+	mc, err := mongoutil.Connect(ctx, cfg.MongoURI, cfg.MongoUsername, cfg.MongoPassword,
+		mongoutil.WithObservability(sdk), mongoutil.WithLazyConnect())
 	if err != nil {
 		return fmt.Errorf("connect mongo: %w", err)
 	}
--- a/botplatform-service/handler.go
+++ b/botplatform-service/handler.go
@@ -52,12 +52,13 @@ func newHandler(s BotplatformStore, cfg *config) *handler {
 	}
 }
 
+// HandleHealth answers the liveness probe, which asks "is this process alive",
+// not "can it reach everything it needs". It deliberately checks nothing:
+// docs/health-probes.md says a dependency being down must never get the pod
+// restarted. That matters more now the service starts without waiting for
+// MongoDB — a health check that failed when MongoDB was down would have
+// Kubernetes restarting the pod through the very outage it can now survive.
 func (h *handler) HandleHealth(c *gin.Context) {
-	if err := h.store.Ping(c.Request.Context()); err != nil {
-		slog.WarnContext(c.Request.Context(), "healthz ping failed", "error", err)
-		c.JSON(http.StatusServiceUnavailable, gin.H{"status": "down"})
-		return
-	}
 	c.JSON(http.StatusOK, gin.H{"status": "ok"})
 }
 
--- a/broadcast-worker/main.go
+++ b/broadcast-worker/main.go
@@ -136,7 +136,7 @@ func main() {
 		slog.Info("room-meta L2 cache enabled", "ttl", cfg.RoomMetaL2TTL)
 	}
 	store := NewMongoStore(db.Collection("rooms"), db.Collection("subscriptions"), db.Collection("thread_rooms"), metaValkey, cfg.RoomMetaL2TTL)
-	if err := store.EnsureIndexes(ctx); err != nil {
+	if err := mongoutil.EnsureIndexes(ctx, mongoutil.Step("broadcast-worker store", store.EnsureIndexes)); err != nil {
 		slog.Error("ensure indexes failed", "error", err)
 		os.Exit(1)
 	}
--- a/broadcast-worker/main.go
+++ b/broadcast-worker/main.go
@@ -116,7 +116,7 @@ func main() {
 		os.Exit(1)
 	}
 	mongoClient, err := mongoutil.Connect(ctx, cfg.MongoURI, cfg.MongoUsername, cfg.MongoPassword,
-		mongoutil.WithObservability(sdk), mongoutil.WithReadPreference(readPref))
+		mongoutil.WithObservability(sdk), mongoutil.WithReadPreference(readPref), mongoutil.WithLazyConnect())
 	if err != nil {
 		slog.Error("mongo connect failed", "error", err)
 		os.Exit(1)
--- a/broadcast-worker/startup_resilience_integration_test.go
+++ b/broadcast-worker/startup_resilience_integration_test.go
@@ -1,0 +1,216 @@
+//go:build integration
+
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/nats-io/nats.go"
+	"github.com/nats-io/nats.go/jetstream"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/hmchangw/chat/pkg/model"
+	"github.com/hmchangw/chat/pkg/mongoutil"
+	"github.com/hmchangw/chat/pkg/natsmetrics"
+	"github.com/hmchangw/chat/pkg/stream"
+	"github.com/hmchangw/chat/pkg/subject"
+	"github.com/hmchangw/chat/pkg/testutil"
+	"github.com/hmchangw/chat/pkg/userstore"
+)
+
+// TestStartup_ReachesServingStateWithMongoUnreachable walks the service's real
+// Mongo boot sequence with MongoDB down: connect, build the store, run the
+// index step. Before this change each of those steps killed the process; all
+// three must now complete so the worker can go on to consume from JetStream.
+func TestStartup_ReachesServingStateWithMongoUnreachable(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	// Gate 1: connect must not fail.
+	client, err := mongoutil.Connect(ctx, testutil.UnreachableMongoURI(t), "", "", mongoutil.WithLazyConnect())
+	require.NoError(t, err, "service must connect lazily while MongoDB is unreachable")
+	t.Cleanup(func() { mongoutil.Disconnect(context.Background(), client) })
+
+	db := client.Database("broadcast_worker_startup_test")
+	store := NewMongoStore(
+		db.Collection("rooms"),
+		db.Collection("subscriptions"),
+		db.Collection("thread_rooms"),
+		nil, // no Valkey L2 tier
+		time.Minute,
+	)
+	require.NotNil(t, store, "store must be constructible with MongoDB unreachable")
+
+	// Gate 2: an unreachable MongoDB is transient, so the index step must warn
+	// and report success, not hand the caller a reason to exit.
+	require.NoError(t,
+		mongoutil.EnsureIndexes(ctx, mongoutil.Step("broadcast-worker store", store.EnsureIndexes)),
+		"failed index creation must not stop startup")
+
+	// Serving state: the store answers requests with an error rather than
+	// hanging, so the consumer loop can Nak and retry instead of wedging.
+	opCtx, opCancel := context.WithTimeout(ctx, 5*time.Second)
+	defer opCancel()
+	start := time.Now()
+	_, err = store.GetRoom(opCtx, "room-1")
+	elapsed := time.Since(start)
+
+	assert.Error(t, err, "reads must fail while MongoDB is unreachable")
+	assert.Less(t, elapsed, 8*time.Second, "reads must fail bounded, not hang")
+}
+
+// TestStartup_ConsumerRedeliversUntilMongoRecovers is the half that booting
+// during an outage is for. Reaching a serving state is worth nothing if the
+// consumer then mishandles the events it pulls, and the two ways to get that
+// wrong are both silent: Ack a message the store never processed (the event is
+// gone for good), or hot-loop Naks that burn MaxDeliver before MongoDB returns
+// (same outcome, slower). So this drives the worker's real consumer — the
+// production consumer config, broadcastProcessor, consumeLoop — against a
+// MongoDB that is down and then comes back, with no restart in between.
+func TestStartup_ConsumerRedeliversUntilMongoRecovers(t *testing.T) {
+	const siteID = "site-outage"
+	ctx := context.Background()
+
+	// Seed through a direct connection: the room has to exist for the post-
+	// recovery delivery to succeed, and the worker's own client cannot write it
+	// while the outage is on.
+	seed := testutil.MongoDB(t, "broadcast_worker_outage_test")
+	_, err := seed.Collection("rooms").InsertOne(ctx, model.Room{
+		ID: "r1", Name: "general", Type: model.RoomTypeChannel, UserCount: 2, SiteID: siteID,
+	})
+	require.NoError(t, err)
+	_, err = seed.Collection("subscriptions").InsertMany(ctx, []interface{}{
+		model.Subscription{ID: "s1", User: model.SubscriptionUser{ID: "u1", Account: "alice"}, RoomID: "r1"},
+		model.Subscription{ID: "s2", User: model.SubscriptionUser{ID: "u2", Account: "bob"}, RoomID: "r1"},
+	})
+	require.NoError(t, err)
+	seedUsers(t, seed)
+
+	// The worker reaches the same database through an endpoint that is down.
+	outage := testutil.NewMongoOutage(t, testutil.MongoURI(t))
+	client, err := mongoutil.Connect(ctx, outage.URI(), "", "", mongoutil.WithLazyConnect())
+	require.NoError(t, err, "worker must boot while MongoDB is unreachable")
+	t.Cleanup(func() { mongoutil.Disconnect(context.Background(), client) })
+
+	db := client.Database(seed.Name())
+	store := NewMongoStore(db.Collection("rooms"), db.Collection("subscriptions"), db.Collection("thread_rooms"), nil, 0)
+	pub := &recordingPublisher{}
+	handler := NewHandler(store, userstore.NewMongoStore(db.Collection("users")), pub, nil,
+		defaultParentFetcher, false, subject.RouteGlobal)
+
+	js, cons, iter, subj := startCanonicalConsumer(t, siteID)
+
+	// Count deliveries around the real processor rather than replacing it, so
+	// the Ack/Nak decision under test stays broadcastProcessor's.
+	var deliveries atomic.Int32
+	process := broadcastProcessor(handler)
+	counted := func(msgCtx context.Context, msg jetstream.Msg) {
+		deliveries.Add(1)
+		process(msgCtx, msg)
+	}
+
+	// The production composition: natsmetrics.Consume driving
+	// guardedProcessor(broadcastProcessor(…)), same as main.go wires.
+	metrics, _ := newTestBroadcastMetrics(t)
+	consumer := metrics.Consumer(natsmetrics.ConsumerConfig{
+		ServiceName: "broadcast-worker", Site: siteID,
+		Stream: stream.MessagesCanonical(siteID).Name, Consumer: "broadcast-worker",
+	})
+	consumer.LoopStarted(ctx)
+
+	var wg sync.WaitGroup
+	go natsmetrics.Consume(ctx, iter, consumer, 4, consumerMaxDeliver, &wg,
+		func(msg jetstream.Msg) natsmetrics.EventType { return natsmetrics.EventTypeFromSubject(msg.Subject()) },
+		guardedProcessor(counted))
+
+	evt := model.MessageEvent{
+		Event:  model.EventCreated,
+		SiteID: siteID,
+		Message: model.Message{
+			ID: "m1", RoomID: "r1", UserID: "u1", UserAccount: "alice", Content: "hello",
+			CreatedAt: time.Now().UTC().Truncate(time.Millisecond),
+		},
+	}
+	data, err := json.Marshal(evt)
+	require.NoError(t, err)
+	_, err = js.Publish(ctx, subj, data)
+	require.NoError(t, err)
+
+	// While MongoDB is down the event must come back. A second delivery proves
+	// the store failure was Nak'd rather than Acked away.
+	require.Eventually(t, func() bool { return deliveries.Load() >= 2 }, 30*time.Second, 100*time.Millisecond,
+		"the event was never redelivered — a failed Mongo read was Acked, losing the event")
+	assert.Empty(t, pub.getRecords(), "nothing may be broadcast from a read that failed")
+
+	// The retries are spaced, not a hot loop: a Nak() without delay would burn
+	// the consumer's MaxDeliver within milliseconds of the first failure.
+	assert.Less(t, deliveries.Load(), int32(consumerMaxDeliver),
+		"retries are spaced by jsretry backoff, so MaxDeliver must not be exhausted this early")
+
+	outage.Restore(t)
+
+	// Same process, same consumer: the next redelivery must now succeed.
+	require.Eventually(t, func() bool { return len(pub.getRecords()) > 0 }, 60*time.Second, 200*time.Millisecond,
+		"the event was never processed after MongoDB returned")
+
+	records := pub.getRecords()
+	require.Len(t, records, 1, "the recovered delivery must broadcast exactly once")
+	assert.Equal(t, subject.RoomEvent("r1", true), records[0].subject)
+
+	// And it is Acked: an unacked success would sit in ack-pending until AckWait
+	// expired and then broadcast the same event a second time.
+	require.Eventually(t, func() bool {
+		info, infoErr := cons.Info(ctx)
+		return infoErr == nil && info.NumAckPending == 0 && info.NumPending == 0
+	}, 10*time.Second, 200*time.Millisecond, "the processed event was never Acked")
+}
+
+// Consumer settings for the outage test. AckWait must exceed the time one
+// delivery spends failing against an unreachable MongoDB, or JetStream
+// redelivers underneath a handler that is still running and the delivery count
+// stops meaning what this test reads it to mean.
+const (
+	consumerAckWait    = 15 * time.Second
+	consumerMaxDeliver = 10
+)
+
+// startCanonicalConsumer builds the worker's real durable consumer over
+// MESSAGES-CANONICAL on the shared test NATS, returning it both directly (for
+// ack-state assertions) and as a Consume iterator, plus the subject to publish
+// canonical events on.
+func startCanonicalConsumer(t *testing.T, siteID string) (jetstream.JetStream, jetstream.Consumer, natsmetrics.Iterator, string) {
+	t.Helper()
+
+	nc, err := nats.Connect(testutil.NATS(t))
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = nc.Drain() })
+
+	js, err := jetstream.New(nc)
+	require.NoError(t, err)
+
+	sc := stream.MessagesCanonical(siteID)
+	_, err = js.CreateOrUpdateStream(context.Background(), jetstream.StreamConfig{Name: sc.Name, Subjects: sc.Subjects})
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = js.DeleteStream(context.Background(), sc.Name) })
+
+	cons, err := js.CreateOrUpdateConsumer(context.Background(), sc.Name, buildConsumerConfig(stream.ConsumerSettings{
+		AckWait:       consumerAckWait,
+		MaxDeliver:    consumerMaxDeliver,
+		MaxWaiting:    512,
+		MaxAckPending: 1000,
+	}, "broadcast-worker", sc.Subjects[0]))
+	require.NoError(t, err)
+
+	iter, err := cons.Messages()
+	require.NoError(t, err)
+	t.Cleanup(iter.Stop)
+
+	return js, cons, plainIterAdapter{inner: iter}, fmt.Sprintf("chat.msg.canonical.%s.created", siteID)
+}
--- a/pkg/mongoutil/indexes.go
+++ b/pkg/mongoutil/indexes.go
@@ -1,0 +1,132 @@
+package mongoutil
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"time"
+
+	"go.mongodb.org/mongo-driver/v2/mongo"
+	"golang.org/x/sync/errgroup"
+)
+
+// EnsureIndexesTimeout is how long one EnsureIndexes call gets in total. Its
+// steps run at the same time as each other, so this is the whole index budget
+// for a service's startup, not a cost per store: with MongoDB unreachable a
+// service waits this long once, however many stores it has.
+//
+// It is 30s rather than something shorter because asking MongoDB to create an
+// index does not return until the index is built, and building one over a
+// collection that already holds a lot of data genuinely takes a while.
+const EnsureIndexesTimeout = 30 * time.Second
+
+// IndexStep is one index-creation step to run at startup. The name is what
+// shows up in the log if it fails, so make it say which service and which
+// collection — for example "user-service subscriptions".
+type IndexStep struct {
+	Name   string
+	Ensure func(context.Context) error
+}
+
+// Step builds an IndexStep.
+func Step(name string, ensure func(context.Context) error) IndexStep {
+	return IndexStep{Name: name, Ensure: ensure}
+}
+
+// EnsureIndexes creates a service's indexes at startup. It sorts failures into
+// two kinds: ones a restart will clear, and ones only a person can.
+//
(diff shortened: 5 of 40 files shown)


julianshen:
**3 findings · 3 inline comments posted** — 🔴 2 blocking · 🟠 1 warning

### Related work

- [Issue #23](https://github.com/hmchangw/newchat/issues/23) — Thread mention auto-creates \`thread\_subscriptions\` for non-members — no room-membership check in mention resolution (open)
- [PR #270](https://github.com/hmchangw/newchat/pull/270) — feat\(mongoutil\): let online services boot during a MongoDB outage (closed)

### Local review memory

- `bot-room-service/main.go:72` — @coderabbitai: _🩺 Stability & Availability_ _🟠 Major_ _⚡ Quick win_ **Route index initialization through mongoutil.EnsureIndexes .** WithLazyConnect() is followed by a direct store.EnsureIndexes() call at Lines [80-81]. When MongoDB is unavail… ([thread](https://github.com/hmchangw/newchat/pull/284#discussion_r3820007029))
- `broadcast-worker/startup_resilience_integration_test.go:131` — @coderabbitai: _🩺 Stability & Availability_ _🟡 Minor_ _⚡ Quick win_ **Drain the consumer goroutines before the test tears MongoDB down.** wg is passed to natsmetrics.Consume but nothing waits on it, and ctx at line 79 is context.Background()… ([thread](https://github.com/hmchangw/newchat/pull/284#discussion_r3820007044))

<details>
<summary>📒 Files reviewed (40)</summary>

* `admin-service/main.go`
* `bot-room-service/main.go`
* `botplatform-service/handler.go`
* `botplatform-service/handler_test.go`
* `botplatform-service/mock_store_test.go`
* `botplatform-service/store.go`
* `botplatform-service/store_mongo.go`
* `broadcast-worker/main.go`
* `broadcast-worker/startup_resilience_integration_test.go`
* `hr-sync-worker/main.go`
* `inbox-worker/main.go`
* `media-service/main.go`
* `message-worker/main.go`
* `pkg/mongoutil/dropindex_integration_test.go`
* `pkg/mongoutil/dropindex_test.go`
* `pkg/mongoutil/indexes.go`
* `pkg/mongoutil/indexes_test.go`
* `pkg/mongoutil/lazy.go`
* `pkg/mongoutil/lazy_integration_test.go`
* `pkg/mongoutil/lazy_test.go`
* `pkg/mongoutil/mongo.go`
* `pkg/mongoutil/observability.go`
* `pkg/testutil/mongooutage.go`
* `pkg/testutil/mongouri.go`
* `portal-service/main.go`
* `room-service/main.go`
* `room-worker/main.go`
* `search-service/main.go`
* `tcard-service/main.go`
* `upload-service/main.go`
* `user-service/main.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

julianshen (commented):
tGD inline review

pkg/mongoutil/indexes.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+		g.Go(func() error {
+			err := step.Ensure(ctx)
+			if err == nil {
+				return nil
+			}
+			if isPermanentIndexError(err) {
+				return fmt.Errorf("ensure indexes %s: %w", step.Name, err)
+			}
+			slog.Warn("ensure indexes failed; continuing without them",
```
julianshen, line 74: _🏷️ consistency_ | _🔴 Blocking_ | _`distributed-system`_

**Failed unique indexes remain absent for the pod's lifetime.**

Every non-conflict index error is downgraded to a warning, and the failed step is never retried. For example, if an inbox or message worker starts while MongoDB is unreachable, creation of the new unique `(threadRoomId,userAccount)` index fails, startup continues, and recovery of MongoDB does not install the constraint. Concurrent upserts for the same account can then both insert, leaving duplicate subscription state for the lifetime of every such pod. Treat unique-index steps as startup-critical, or retry and gate the consumers that rely on them until reconciliation succeeds; only ordinary performance indexes are safe to skip indefinitely.

<details>

```
In `pkg/mongoutil/indexes.go` around line 74: Every non-conflict index error is downgraded to a warning, and the failed step is never retried. For example, if an inbox or message worker starts while MongoDB is unreachable, creation of the new unique `(threadRoomId,userAccount)` index fails, startup continues, and recovery of MongoDB does not install the constraint. Concurrent upserts for the same account can then both insert, leaving duplicate subscription state for the lifetime of every such pod. Treat unique-index steps as startup-critical, or retry and gate the consumers that rely on them until reconciliation succeeds; only ordinary performance indexes are safe to skip indefinitely.

```

</details>

broadcast-worker/startup_resilience_integration_test.go
```diff
@@ -0,0 +1,216 @@
...
+	require.NoError(t, err)
+	_, err = js.Publish(ctx, subj, data)
+	require.NoError(t, err)
+
+	// While MongoDB is down the event must come back. A second delivery proves
+	// the store failure was Nak'd rather than Acked away.
+	require.Eventually(t, func() bool { return deliveries.Load() >= 2 }, 30*time.Second, 100*time.Millisecond,
+		"the event was never redelivered — a failed Mongo read was Acked, losing the event")
+	assert.Empty(t, pub.getRecords(), "nothing may be broadcast from a read that failed")
+
```
julianshen, line 151: _🏷️ delivery-guarantees_ | _🔴 Blocking_ | _`distributed-system`_

**A prolonged Mongo outage can exhaust all message deliveries.**

The consumer has `MaxDeliver` set to 10, but the test restores MongoDB immediately after observing only two failures. With the new lazy startup, a worker can begin consuming while MongoDB remains down; if the outage lasts through all ten delayed attempts, JetStream stops redelivering, so restoring MongoDB later cannot produce the broadcast. The test comments themselves identify exhausting `MaxDeliver` as permanent loss, but this timeline is not covered. Dependency failures need unlimited bounded-backoff retries or a durable retry/dead-letter path that is replayed after recovery, and the integration test should keep MongoDB down through the current delivery limit before restoring it.

<details>

```
In `broadcast-worker/startup_resilience_integration_test.go` around line 151: The consumer has `MaxDeliver` set to 10, but the test restores MongoDB immediately after observing only two failures. With the new lazy startup, a worker can begin consuming while MongoDB remains down; if the outage lasts through all ten delayed attempts, JetStream stops redelivering, so restoring MongoDB later cannot produce the broadcast. The test comments themselves identify exhausting `MaxDeliver` as permanent loss, but this timeline is not covered. Dependency failures need unlimited bounded-backoff retries or a durable retry/dead-letter path that is replayed after recovery, and the integration test should keep MongoDB down through the current delivery limit before restoring it.

```

</details>

pkg/mongoutil/indexes.go
```diff
@@ -0,0 +1,132 @@
...
+	return IndexStep{Name: name, Ensure: ensure}
+}
+
+// EnsureIndexes creates a service's indexes at startup. It sorts failures into
+// two kinds: ones a restart will clear, and ones only a person can.
+//
+// If MongoDB is unreachable, or is up but refusing writes, that is the first
+// kind. Creating an index is a write, so both look the same here. EnsureIndexes
+// logs a warning and reports success, so the service still starts. Nothing
+// retries in the background: the whole step runs again on the next start
```
julianshen, line 43: _🏷️ data-integrity_ | _🔴 Blocking_ | _`mongodb`_

**Transient index failures leave constraints missing indefinitely**

The helper deliberately performs no background retry after returning success for a transient failure. On a fresh deployment that starts during a MongoDB outage, the inbox-worker and message-worker `thread_subscriptions` migration can therefore remain without the unique `(threadRoomId,userAccount)` index after MongoDB recovers; concurrent upserts can then insert duplicate logical subscriptions. If creation succeeds but dropping the legacy `(threadRoomId,userId)` index transiently fails, that stale constraint instead continues rejecting valid federated writes. Retry idempotent index steps with backoff after recovery, or keep consumers/readiness gated until correctness-critical unique-index migrations complete.

<details>

```
In `pkg/mongoutil/indexes.go` around line 43: The helper deliberately performs no background retry after returning success for a transient failure. On a fresh deployment that starts during a MongoDB outage, the inbox-worker and message-worker `thread_subscriptions` migration can therefore remain without the unique `(threadRoomId,userAccount)` index after MongoDB recovers; concurrent upserts can then insert duplicate logical subscriptions. If creation succeeds but dropping the legacy `(threadRoomId,userId)` index transiently fails, that stale constraint instead continues rejecting valid federated writes. Retry idempotent index steps with backoff after recovery, or keep consumers/readiness gated until correctness-critical unique-index migrations complete.

```

</details>

pkg/mongoutil/indexes.go
```diff
@@ -0,0 +1,132 @@
...
+			}
+			if isPermanentIndexError(err) {
+				return fmt.Errorf("ensure indexes %s: %w", step.Name, err)
+			}
+			slog.Warn("ensure indexes failed; continuing without them",
+				"indexes", step.Name,
+				"error", err.Error(),
+				"impact", "degraded query performance; unique constraints unenforced until a later start succeeds",
+			)
+			return nil
```
julianshen, line 79: _🎯 concurrency_ | _🔴 Blocking_ | _`performance-and-concurrency`, `tgd-review` (2 rules)_

**Transient index failures are never retried after recovery**

When a pod starts during a MongoDB outage, every failed step reaches this return and `EnsureIndexes` reports success. If MongoDB then recovers without another restart, the indexes remain absent indefinitely. Under load, queries intended to use those indexes can continue scanning collections, and concurrent upserts relying on a missing unique index can both observe no match and insert duplicates. Keep startup resilient, but retain transiently failed steps and retry them from one lifecycle-bound worker using capped backoff and jitter until they succeed; continue returning permanent index conflicts as fatal errors.

<details>

- **`tgd-review`**:79: After warning, the helper returns success and never reruns the failed step. If this release starts during an outage, the thread-subscription migration in message-worker and inbox-worker leaves the new index absent and the legacy unique index present even after MongoDB recovers. Cross-site writes can therefore continue failing until another process restart, contradicting the no-restart recovery goal. Retry transient index steps in the background after recovery, or otherwise prevent affected writes until migration succeeds.

</details>

<details>

```
In `pkg/mongoutil/indexes.go` around line 79: When a pod starts during a MongoDB outage, every failed step reaches this return and `EnsureIndexes` reports success. If MongoDB then recovers without another restart, the indexes remain absent indefinitely. Under load, queries intended to use those indexes can continue scanning collections, and concurrent upserts relying on a missing unique index can both observe no match and insert duplicates. Keep startup resilient, but retain transiently failed steps and retry them from one lifecycle-bound worker using capped backoff and jitter until they succeed; continue returning permanent index conflicts as fatal errors.

```

</details>

pkg/mongoutil/indexes.go
```diff
@@ -0,0 +1,132 @@
...
+	var g errgroup.Group
+	for _, step := range steps {
+		if step.Ensure == nil {
+			continue
+		}
+		g.Go(func() error {
+			err := step.Ensure(ctx)
+			if err == nil {
+				return nil
+			}
```
julianshen, line 70: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**Unknown permanent index errors are downgraded to warnings.**

`isPermanentIndexError` recognizes only duplicate keys and codes 85/86, so every other failure reaches the warning branch and is reported as success. Permanent failures such as Unauthorized (13), BadValue (2), CannotCreateIndex (67), InvalidOptions (72), or an invalid index specification will therefore let services start with a broken schema. This also masks failures from `DropIndexIfExists`, potentially leaving the legacy unique index active. Default unknown server errors to fatal and explicitly whitelist genuinely transient network, selection, retryable-write, and context timeout failures.

<details>

```
In `pkg/mongoutil/indexes.go` around line 70: `isPermanentIndexError` recognizes only duplicate keys and codes 85/86, so every other failure reaches the warning branch and is reported as success. Permanent failures such as Unauthorized (13), BadValue (2), CannotCreateIndex (67), InvalidOptions (72), or an invalid index specification will therefore let services start with a broken schema. This also masks failures from `DropIndexIfExists`, potentially leaving the legacy unique index active. Default unknown server errors to fatal and explicitly whitelist genuinely transient network, selection, retryable-write, and context timeout failures.

```

</details>

pkg/mongoutil/indexes.go
```diff
@@ -0,0 +1,132 @@
...
+//     reject, so the duplicates have to be sorted out first.
+//   - 85 and 86: an index on those fields already exists with different
+//     settings. MongoDB will not change one in place, so the old one has to be
+//     dropped by hand.
+func isPermanentIndexError(err error) bool {
+	if mongo.IsDuplicateKeyError(err) {
+		return true
+	}
+	if se := mongo.ServerError(nil); errors.As(err, &se) {
+		return se.HasErrorCode(85) || se.HasErrorCode(86)
```
julianshen, line 129: _🎯 correctness_ | _🟠 Warning_ | _`mongodb`_

**Authentication failures are misclassified as transient index errors**

Because lazy connection skips the startup ping, `CreateIndexes` may be the first authenticated operation. MongoDB authentication failure (code 18) or missing index-management privilege (code 13) currently falls through this classifier, so `EnsureIndexes` logs a warning and reports successful startup. Invalid credentials leave every MongoDB operation failing indefinitely, while insufficient DDL privileges can leave unique or TTL indexes absent. Treat these persistent configuration errors as fatal and add corresponding classifier tests.

<details>

```
In `pkg/mongoutil/indexes.go` around line 129: Because lazy connection skips the startup ping, `CreateIndexes` may be the first authenticated operation. MongoDB authentication failure (code 18) or missing index-management privilege (code 13) currently falls through this classifier, so `EnsureIndexes` logs a warning and reports successful startup. Invalid credentials leave every MongoDB operation failing indefinitely, while insufficient DDL privileges can leave unique or TTL indexes absent. Treat these persistent configuration errors as fatal and add corresponding classifier tests.

```

</details>

broadcast-worker/startup_resilience_integration_test.go
```diff
@@ -0,0 +1,216 @@
...
+	// The production composition: natsmetrics.Consume driving
+	// guardedProcessor(broadcastProcessor(…)), same as main.go wires.
+	metrics, _ := newTestBroadcastMetrics(t)
+	consumer := metrics.Consumer(natsmetrics.ConsumerConfig{
+		ServiceName: "broadcast-worker", Site: siteID,
+		Stream: stream.MessagesCanonical(siteID).Name, Consumer: "broadcast-worker",
+	})
+	consumer.LoopStarted(ctx)
+
+	var wg sync.WaitGroup
```
julianshen, line 128: _🏷️ NATS lifecycle_ | _🟠 Warning_ | _`nats`_

**The test leaves the JetStream consumer running during teardown.**

The consume loop uses a non-cancellable context and the WaitGroup passed to natsmetrics.Consume is never awaited. When the test finishes, iterator, NATS, and MongoDB cleanups can run while an in-flight delivery is still accessing those resources, causing goroutine leaks or failures after the test completes. Cancel the consumer and wait for pending handlers before teardown.

<details>

```
In `broadcast-worker/startup_resilience_integration_test.go` around line 128: The consume loop uses a non-cancellable context and the WaitGroup passed to natsmetrics.Consume is never awaited. When the test finishes, iterator, NATS, and MongoDB cleanups can run while an in-flight delivery is still accessing those resources, causing goroutine leaks or failures after the test completes. Cancel the consumer and wait for pending handlers before teardown.

```

</details>

botplatform-service/handler.go
```diff
@@ -52,12 +52,13 @@ func newHandler(s BotplatformStore, cfg *config) *handler {
 	}
 }
 
+// HandleHealth answers the liveness probe, which asks "is this process alive",
```
julianshen, line 55: _⚡ memory-exhaustion_ | _🟠 Warning_ | _`oom`_

**Mongo outages can leave unbounded requests resident in memory.**

`WithLazyConnect` now lets this service accept traffic while MongoDB is unavailable, while this change removes the only dependency-aware health check (`Ping` is also removed from the store interface). During an outage, sustained login/API traffic is therefore admitted while each MongoDB operation can wait roughly 30 seconds for server selection, as documented in `pkg/mongoutil/lazy.go`. Each request keeps its handler goroutine and request-scoped state alive, so in-flight memory grows approximately with request rate multiplied by that timeout; an external flood or ordinary traffic spike can cause GC thrashing, container eviction, or process OOM. Keep `/healthz` dependency-free for liveness, but add a separate Mongo-backed readiness endpoint and configure Kubernetes to use it, or enforce a hard in-flight request limit with short database-operation deadlines so outage traffic receives immediate backpressure.

<details>

```
In `botplatform-service/handler.go` around line 55: `WithLazyConnect` now lets this service accept traffic while MongoDB is unavailable, while this change removes the only dependency-aware health check (`Ping` is also removed from the store interface). During an outage, sustained login/API traffic is therefore admitted while each MongoDB operation can wait roughly 30 seconds for server selection, as documented in `pkg/mongoutil/lazy.go`. Each request keeps its handler goroutine and request-scoped state alive, so in-flight memory grows approximately with request rate multiplied by that timeout; an external flood or ordinary traffic spike can cause GC thrashing, container eviction, or process OOM. Keep `/healthz` dependency-free for liveness, but add a separate Mongo-backed readiness endpoint and configure Kubernetes to use it, or enforce a hard in-flight request limit with short database-operation deadlines so outage traffic receives immediate backpressure.

```

</details>

pkg/mongoutil/indexes.go
julianshen, line 129: _🏷️ error-handling_ | _🟠 Warning_ | _`silent-failure`_

**Authorization failures are converted into successful startup**

MongoDB error codes 13 (Unauthorized) and 18 (AuthenticationFailed) fall through as transient failures, so EnsureIndexes logs a warning and returns nil. With lazy connection, this may be the first operation that detects invalid credentials or insufficient index privileges; the service then appears to start successfully without required indexes, potentially leaving unique constraints unenforced. Treat these person-actionable errors as permanent and add corresponding tests.

<details>

```
In `pkg/mongoutil/indexes.go` around line 129: MongoDB error codes 13 (Unauthorized) and 18 (AuthenticationFailed) fall through as transient failures, so EnsureIndexes logs a warning and returns nil. With lazy connection, this may be the first operation that detects invalid credentials or insufficient index privileges; the service then appears to start successfully without required indexes, potentially leaving unique constraints unenforced. Treat these person-actionable errors as permanent and add corresponding tests.

```

</details>

broadcast-worker/startup_resilience_integration_test.go
julianshen, line 128: _🧪 tests_ | _🟠 Warning_ | _`tgd-review`_

**The consumer goroutines outlive the integration test.**

The consumer uses `context.Background`, and the supplied wait group is never waited on. Test cleanup can consequently stop NATS and disconnect MongoDB while handlers are still running, causing leaks or failures after the test returns. Cancel and drain the consumer before the previously registered resource cleanups execute.

<details>

```
In `broadcast-worker/startup_resilience_integration_test.go` around line 128: The consumer uses `context.Background`, and the supplied wait group is never waited on. Test cleanup can consequently stop NATS and disconnect MongoDB while handlers are still running, causing leaks or failures after the test returns. Cancel and drain the consumer before the previously registered resource cleanups execute.

```

</details>

broadcast-worker/main.go
```diff
@@ -136,7 +136,7 @@ func main() {
 		slog.Info("room-meta L2 cache enabled", "ttl", cfg.RoomMetaL2TTL)
 	}
 	store := NewMongoStore(db.Collection("rooms"), db.Collection("subscriptions"), db.Collection("thread_rooms"), metaValkey, cfg.RoomMetaL2TTL)
```
julianshen, line 138: _🏷️ delivery_ | _🔴 Blocking_ | _`performance-and-concurrency`_

**A long Mongo outage exhausts the broadcast consumer's deliveries.**

Lazy connection plus this best-effort index gate lets the worker consume while MongoDB is unavailable. The added integration scenario uses `MaxDeliver=10` and shows each failed Mongo read is negatively acknowledged, but restores Mongo after only two attempts. If the outage lasts through all ten delayed deliveries, JetStream stops delivering the event to this durable consumer, so Mongo recovery cannot produce the broadcast. Use unlimited bounded-backoff retries for dependency failures or route exhausted events to a durable replayed dead-letter path, and test recovery after the current delivery limit is exceeded.

<details>

```
In `broadcast-worker/main.go` around line 138: Lazy connection plus this best-effort index gate lets the worker consume while MongoDB is unavailable. The added integration scenario uses `MaxDeliver=10` and shows each failed Mongo read is negatively acknowledged, but restores Mongo after only two attempts. If the outage lasts through all ten delayed deliveries, JetStream stops delivering the event to this durable consumer, so Mongo recovery cannot produce the broadcast. Use unlimited bounded-backoff retries for dependency failures or route exhausted events to a durable replayed dead-letter path, and test recovery after the current delivery limit is exceeded.

```

</details>

bot-room-service/main.go
```diff
@@ -68,7 +68,8 @@ func run() error {
 		return fmt.Errorf("init jetstream: %w", err)
 	}
```
julianshen, line 69: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**A direct index check still prevents startup during Mongo outages.**

After enabling lazy connection, this service still calls store.EnsureIndexes directly under a timeout and returns its error. An unavailable MongoDB therefore still makes bot-room-service exit instead of reaching its serving state. Route that call through mongoutil.EnsureIndexes, as the other changed services do, while preserving fatal handling for permanent conflicts.

<details>

```
In `bot-room-service/main.go` around line 69: After enabling lazy connection, this service still calls store.EnsureIndexes directly under a timeout and returns its error. An unavailable MongoDB therefore still makes bot-room-service exit instead of reaching its serving state. Route that call through mongoutil.EnsureIndexes, as the other changed services do, while preserving fatal handling for permanent conflicts.

```

</details>

pkg/mongoutil/mongo.go
```diff
@@ -59,15 +59,29 @@ func connect(ctx context.Context, clientOpts *options.ClientOptions, uri string,
...
 	}
-	if err := client.Ping(ctx, nil); err != nil {
-		_ = client.Disconnect(context.Background())
-		runCleanup(cleanup)
-		return nil, fmt.Errorf("mongo ping: %w", err)
+	// mongo.Connect above did not talk to MongoDB — it only built a client.
+	// This ping is the one thing that proves the database is reachable, which
+	// is why skipping it is what lets a service start during an outage. See
+	// WithLazyConnect for who should skip it and why.
+	if cfg.lazy {
```
julianshen, line 66: _🏷️ configuration_ | _🟠 Warning_ | _`tgd-review`_

**Lazy startup also hides permanently invalid Mongo credentials.**

Skipping Ping avoids outage failures, but it also removes the only authentication check for services that do not immediately execute an index operation. A deployment with invalid credentials can therefore report successful startup even though the fixed client credentials can never perform useful work. Add an eventual dependency validation/readiness mechanism that distinguishes authentication and authorization failures from transient reachability failures.

<details>

```
In `pkg/mongoutil/mongo.go` around line 66: Skipping Ping avoids outage failures, but it also removes the only authentication check for services that do not immediately execute an index operation. A deployment with invalid credentials can therefore report successful startup even though the fixed client credentials can never perform useful work. Add an eventual dependency validation/readiness mechanism that distinguishes authentication and authorization failures from transient reachability failures.

```

</details>

julianshen (commented):
tGD inline review

mliu33 (approved):
Fabulous! Thanks!

mliu33:
Require rebase and resolve conflicts

mliu33:
@hmchangw btw, we have made a change before to decide not to crash the pod for any ensureIndex error

Reference: https://github.com/hmchangw/newchat/pull/333/files

Please help to merge with this change. Thanks!

hmchangw:
## Superseded by #484 — proposing we close this rather than rebase it

Both requests on this PR are answered, but by a successor rather than by a rebase here. Laying out the evidence so the decision is reviewable, and listing what #484 does **not** carry over so nothing is dropped silently.

### The two open requests

1. **"Require rebase and resolve conflicts."** Still true — this PR is `mergeable_state: dirty` and no longer merges into `main`; its base `433531d` is old enough that the merge base has aged out. #484 is rebased on current `main`, eight signed commits, CI 15/15 green.
2. **"We decided not to crash the pod for any ensureIndex error (#333). Please help to merge with this change."** #484 is built on #333 rather than around it: every `EnsureIndexes` call site warns and continues, it uses #333's `EnsureIndexWithRepair` for owners, and it adds `WarnMissingUniqueIndexes` as a strengthened `WarnMissingIndexes` that checks the `unique` **option** rather than the name alone — which closes the case where an index exists but silently lost its constraint.

### The three blocking findings here are what #484 implements

This is the substantive reason for a successor instead of a rebase. The review on this PR named the fixes, and they are structural, not patches:

| Finding here | Recommended fix | In #484 |
|---|---|---|
| "Failed unique indexes remain absent for the pod's lifetime" (🔴) | "retry and gate the consumers that rely on them until reconciliation succeeds" | `mongoutil.IndexGate` — confirms the index before each dependent write and NAKs until it exists |
| "A long Mongo outage exhausts the broadcast consumer's deliveries" (🔴) | "unlimited bounded-backoff retries for dependency failures" | `stream.WithOutageRetryBudget` — `MaxDeliver` 6 → ~17, so a degraded pod parks through the outage |
| "Lazy startup also hides permanently invalid Mongo credentials" (🟠) | "distinguish authentication and authorization failures from transient reachability failures" | the ping is kept, not skipped; rejection is read from the ping **and** the pool monitor, and a post-start rejection exits non-zero |

That last row is the mechanism difference. `WithLazyConnect()` skips the ping, which is the only startup authentication check, so a pod with a wrong password boots and reports healthy. `WithDegradedStart()` keeps the ping and only downgrades a *reachability* failure, so bad credentials stay fatal.

Similarly, this PR's own Risk section accepts that "between MongoDB becoming writable again and the next restart, duplicates it would have rejected can be written". `IndexGate` closes exactly that window.

### What #484 does NOT carry over

Real gaps, worth tracking separately rather than losing:

- **Service coverage: 22 here vs 8 in #484.** Of the 14 not in #484, two (`room-service`, `bot-room-service`) are *deliberately* excluded under the documented rule that MongoDB is their primary datastore, so they must keep failing fast. The other twelve — `admin-service`, `bot-message-handler`, `bot-message-worker`, `botplatform-service`, `hr-sync-worker`, `inbox-worker`, `media-service`, `portal-service`, `room-worker`, `search-sync-worker`, `upload-service`, `user-service` — are simply outside #484's scope and need a per-service decision.
- **`botplatform-service` `/healthz` no longer pinging Mongo.** Still pings on `main` (`handler.go:56`), and `docs/health-probes.md` says liveness should probe no dependency, so this remains a valid independent fix.
- **The outage test harness** — `pkg/testutil/mongooutage.go` and `mongouri.go` (~210 lines) are absent from `main`. #484 uses an in-process fake mongod instead, so these are still unclaimed and would be reusable.
- **The `dropIndex` helper tests**, and the create-before-drop ordering for the legacy `thread_subscriptions` index.

### Proposal

Close this PR, land #484, and open a follow-up for the twelve remaining services plus the `botplatform-service` liveness fix. Happy to be argued out of it if the broader service coverage here is the more urgent half — in that case the cleaner path is to rebase this one as a coverage-only change on top of #484 rather than re-landing the `mongoutil` core twice.

---

hmchangw:
Close in favor of #484

PR #285, opened 2026-08-16
feat(history-service): hasNext in Load History + fix empty history for rooms missing lastMsgAt

mliu33:
## Summary

Two changes to history-service's read path:

1. **`hasNext` in the Load History response** — the frontend can lazy-load older pages on scroll-up instead of guessing from page fullness.
2. **Bugfix: rooms whose Mongo doc has no `lastMsgAt` always returned empty history** — loadHistory (and msg.next / msg.surrounding) now work for legacy/migrated rooms.

## 1. `hasNext` in Load History

- `LoadHistoryResponse` gains `HasNext bool` (`json:"hasNext"`, always present) — the Cassandra bucket walker already computed this on every backward page read; `LoadHistory` previously discarded it.
- Request body and `before`-based pagination contract are unchanged; `nextCursor`/`cursor` plumbing (as in `msg.next`) was deliberately left out of scope.
- Semantics: `hasNext=false` exactly when the walk reached the caller's boundary (room start, history floor, or access window). Conservative — rarely `true` when nothing older remains; the next fetch then returns an empty page with `hasNext=false`.
- Guard on the raw walker flag: **an empty page always reports `hasNext=false`** (`page.HasNext && len(page.Data) > 0`). The walker caps each walk at `MESSAGE_READ_MAX_BUCKETS` (default 122 ≈ 366 days at 72h buckets) and can return an empty *resumable* page; without a cursor the client has nothing to advance `before` with, so the raw flag would livelock it on an identical request.
- `docs/client-api.md` + derived `docs/client-api/request-reply.md` updated.

## 2. Fix: missing `lastMsgAt` → permanently empty history

**Root cause:** `resolveRoomTimes`' consistency collapse rewrote a zero (unset) `lastMsgAt` to `createdAt`, conflating "not recorded" with "empty room". `LoadHistory`'s dead-room optimization then capped `before` at `createdAt+1ms`, so the walk only ever scanned pre-creation time — always empty. The same collapsed value fed `walkBounds`' ceiling, breaking `msg.next` (client never receives newer messages) and `msg.surrounding` (no messages after the pivot) identically. A room doc can lack `lastMsgAt` while holding messages: the field is only written by `message-worker` on persist, so legacy/migrated docs or failed updates hit this.

**Fix:** guard the collapse — `if !last.IsZero() && created.After(*last)`. Zero `lastMsgAt` now survives as "unknown", which every caller already handles (`LoadHistory` skips its `before`-cap; `walkBounds` ceilings at `now + 1h` skew). The hint-consistency refetch and the non-zero inverted-pair collapse keep today's behavior.

## Tests

- TDD throughout (red-first): resolver tests (`missing lastMsgAt stays zero`, hint-refetch variant), a new `walkBounds` table test, service-level regressions (`LoadHistory` returns messages for a no-`lastMsgAt` room with the `before` bound provably uncapped; `hasNext` true/false/empty-page cases).
- Full unit suite (`make test`, race detector): 108 packages pass. Lint clean.
- Integration (`make test-integration`, real Cassandra/Mongo/NATS via testcontainers): history-service packages all pass.

https://claude.ai/code/session_01JKhzZLnkNwBs7SuJ5VnPrb

### DIFF
--- a/data-migration/es-index-migrator/deploy/docker-compose.yml
+++ b/data-migration/es-index-migrator/deploy/docker-compose.yml
@@ -25,7 +25,7 @@ services:
       - USER_ROOM_INDEX=user-room-mv-site-local
       - MIGRATION_START_AT=2025-07-01T00:00:00Z
       - MIGRATION_END_AT=2026-07-01T00:00:00Z
-      - MESSAGE_BUCKET_HOURS=72
+      - MESSAGE_BUCKET_HOURS=360
       - MONGO_URI=mongodb://mongodb:27017
       - MONGO_DB=chat
       - CASSANDRA_HOSTS=cassandra
--- a/docker-local/cassandra/init/10-table-messages_by_room.cql
+++ b/docker-local/cassandra/init/10-table-messages_by_room.cql
@@ -38,7 +38,7 @@ CREATE TABLE IF NOT EXISTS chat.messages_by_room (
   AND compaction = {
     'class': 'TimeWindowCompactionStrategy',
     'compaction_window_unit': 'HOURS',
-    'compaction_window_size': '72'
+    'compaction_window_size': '360'
   };
 
 ALTER TABLE chat.messages_by_room ADD enc_payload BLOB;
--- a/docs/client-api.md
+++ b/docs/client-api.md
@@ -3040,6 +3040,7 @@ Live reaction events (`MessageReactedPayload`) carry a single-actor delta (`{sho
 | Field | Type | Notes |
 |---|---|---|
 | `messages` | array<Message> | Most-recent first. See [Message schema](#message-schema). |
+| `hasNext` | boolean | `true` if older messages exist beyond this page — fetch the next page with `before` = the oldest returned message's `createdAt`. `false` once the caller's history boundary (room start, history floor, or access window) is reached. |
 | `minUserLastSeenAt` | number | Optional. UTC milliseconds since Unix epoch. The room's **strict read floor** — `MIN(lastSeenAt)` across all subscribers, present **only when every member has read** the room. Omitted (the key is absent, never `null`) when any member has not read yet (so botDM rooms, where the bot never reads, never set it), when the most recent read is already past `room.lastMsgAt` (recompute is skipped), or when the value cannot be retrieved (best-effort; messages still load). See the Message Read RPC for how this floor is recomputed. |
 
 ```json
@@ -3057,6 +3058,7 @@ Live reaction events (`MessageReactedPayload`) carry a single-actor delta (`{sho
       "msg": "morning team"
     }
   ],
+  "hasNext": true,
   "minUserLastSeenAt": 1746518100000
 }
 ```
--- a/docs/client-api.md
+++ b/docs/client-api.md
@@ -3040,7 +3040,7 @@ Live reaction events (`MessageReactedPayload`) carry a single-actor delta (`{sho
 | Field | Type | Notes |
 |---|---|---|
 | `messages` | array<Message> | Most-recent first. See [Message schema](#message-schema). |
-| `hasNext` | boolean | `true` if older messages exist beyond this page — fetch the next page with `before` = the oldest returned message's `createdAt`. `false` once the caller's history boundary (room start, history floor, or access window) is reached. |
+| `hasNext` | boolean | `true` if older messages may exist beyond this page — fetch the next page with `before` = the oldest returned message's `createdAt`. `false` once the caller's history boundary (room start, history floor, or access window) is reached. Conservative: occasionally `true` when nothing older remains; the following fetch then returns an empty page with `hasNext=false`. An empty page always has `hasNext=false`. |
 | `minUserLastSeenAt` | number | Optional. UTC milliseconds since Unix epoch. The room's **strict read floor** — `MIN(lastSeenAt)` across all subscribers, present **only when every member has read** the room. Omitted (the key is absent, never `null`) when any member has not read yet (so botDM rooms, where the bot never reads, never set it), when the most recent read is already past `room.lastMsgAt` (recompute is skipped), or when the value cannot be retrieved (best-effort; messages still load). See the Message Read RPC for how this floor is recomputed. |
 
 ```json
--- a/docs/superpowers/specs/2026-08-14-history-service-hasnext-design.md
+++ b/docs/superpowers/specs/2026-08-14-history-service-hasnext-design.md
@@ -1,0 +1,87 @@
+# history-service: `hasNext` in Load History response
+
+**Date:** 2026-08-14
+**Status:** Approved
+
+## Problem
+
+The Load History RPC (`chat.user.{account}.request.room.{roomID}.{siteID}.msg.history`,
+handled by `HistoryService.LoadHistory`) returns a page of messages but gives the
+frontend no signal whether older messages exist. The frontend needs a `hasNext`
+field to drive lazy loading when the user scrolls up in a room — today it can only
+guess (e.g. "page came back full") and must issue a wasted fetch to discover the
+end of history.
+
+## Decision
+
+Expose the pagination signal the Cassandra layer already computes. Every backward
+page read (`GetMessagesBefore` / `GetMessagesBetweenDesc`) returns
+`cassrepo.Page[models.Message]` with a populated `HasNext`; `LoadHistory`
+currently discards it. Plumb it through to the wire.
+
+Scope is deliberately `hasNext` only. `nextCursor`/`cursor` (full cursor
+pagination as in `msg.next`) is out of scope: the frontend paginates Load History
+by `before` = oldest received `createdAt`, and that contract stays unchanged.
+A cursor can be added later without breaking this change.
+
+## Semantics
+
+`hasNext=true` means the page walk did not reach its terminal boundary. The
+boundary is whichever applies to the caller:
+
+- the room's effective start (room `createdAt`, clamped by the configured
+  history floor), for callers with full history access, or
+- the caller's access window (`accessSince`), for windowed access — so a
+  limited-history user sees `hasNext=false` at *their* boundary, not the room's
+  true beginning.
+
+The signal is conservative: `true` can rarely occur when the remaining buckets
+happen to be empty (e.g. the page filled exactly at the last message and buckets
+below are empty). The frontend's next fetch then returns an empty page with
+`hasNext=false`. This is the identical contract already shipped by
+`msg.next` (Load Next Messages), thread replies, and pinned messages.
+
+## Changes
+
+1. **Model** — `history-service/internal/models/message.go`:
+   `LoadHistoryResponse` gains `HasNext bool` with tag `json:"hasNext"`
+   (always present, no `omitempty` — matching `LoadNextMessagesResponse`).
+   Request body unchanged.
+
+2. **Handler** — `history-service/internal/service/messages.go`,
+   `LoadHistory`: set `HasNext: page.HasNext` in the returned
+   `LoadHistoryResponse`. Both read paths (open walk and access-windowed walk)
+   already produce it.
+
+3. **Docs** — client-facing RPC, so in the same PR:
+   - `docs/client-api.md` § Load History: add the `hasNext` row to the success
+     response table and to the JSON example, worded like the Load Next
+     Messages section.
+   - `docs/client-api/request-reply.md` § Load History: mirror the same change
+     (derived view must not drift).
+   - `docs/client-api/events.md`: untouched (no event change).
+
+## Testing (TDD)
+
+Extend the `LoadHistory` unit tests in
+`history-service/internal/service/messages_test.go`, following the existing
+`Test<Type>_<Method>_<Scenario>` per-scenario pattern, with mocked reader pages:
+
+- reader returns `Page{HasNext: true}` → response `hasNext=true`;
+- reader returns a terminal page (`HasNext: false`) → response `hasNext=false`;
+- both on the open-walk path and the access-windowed path.
+
+Red first (assert the new field before implementing), then the one-line
+implementation. No `cassrepo` integration-test changes: the walker's `HasNext`
+behavior is already covered there.
+
+## Error handling
+
+Unchanged. No new error paths; failures still surface via the existing
+`errcode`/`errnats` boundary.
+
+## Out of scope
+
+- `nextCursor` in the Load History response / `cursor` in its request.
+- Any frontend implementation.
+- Changes to `LoadNextMessages`, surrounding-messages, threads, or pins.
--- a/docs/superpowers/specs/2026-08-14-history-service-hasnext-design.md
+++ b/docs/superpowers/specs/2026-08-14-history-service-hasnext-design.md
@@ -41,6 +41,16 @@ below are empty). The frontend's next fetch then returns an empty page with
 `hasNext=false`. This is the identical contract already shipped by
 `msg.next` (Load Next Messages), thread replies, and pinned messages.
 
+One divergence from the raw walker flag (added during review): an **empty page
+always reports `hasNext=false`**, i.e. the handler returns
+`page.HasNext && len(page.Data) > 0`. The walker can return an empty
+*resumable* page when its bucket budget is exhausted crossing a long silent
+gap; cursor RPCs resume via `nextCursor`, but this RPC pages by `before` =
+oldest returned `createdAt` — an empty page gives the client nothing to advance
+with, so passing the raw flag through would livelock the client on an
+identical request. Any history hidden by this guard is unreachable via the
+`before` contract regardless.
+
 ## Changes
 
 1. **Model** — `history-service/internal/models/message.go`:
--- a/history-service/internal/config/config.go
+++ b/history-service/internal/config/config.go
@@ -58,9 +58,9 @@ type Config struct {
 	Cassandra               CassandraConfig `envPrefix:"CASSANDRA_"`
 	Mongo                   MongoConfig     `envPrefix:"MONGO_"`
 	NATS                    NATSConfig      `envPrefix:"NATS_"`
-	MessageBucketHours      int             `env:"MESSAGE_BUCKET_HOURS"        envDefault:"72"`
+	MessageBucketHours      int             `env:"MESSAGE_BUCKET_HOURS"        envDefault:"360"`
 	MessageReadMaxBuckets   int             `env:"MESSAGE_READ_MAX_BUCKETS"    envDefault:"122"`
-	MessageHistoryFloorDays int             `env:"MESSAGE_HISTORY_FLOOR_DAYS"  envDefault:"365"`
+	MessageHistoryFloorDays int             `env:"MESSAGE_HISTORY_FLOOR_DAYS"  envDefault:"730"`
 	LargeRoomThreshold      int             `env:"LARGE_ROOM_THRESHOLD"        envDefault:"500"`
 	MaxPinnedPerRoom        int             `env:"MAX_PINNED_PER_ROOM"         envDefault:"10"`
 	PinEnabled              bool            `env:"PIN_ENABLED"                 envDefault:"true"`
--- a/history-service/internal/service/messages.go
+++ b/history-service/internal/service/messages.go
@@ -95,6 +95,7 @@ func (s *HistoryService) LoadHistory(c *natsrouter.Context, req models.LoadHisto
 	setDecodedAttachments(c, page.Data)
 	return &models.LoadHistoryResponse{
 		Messages:          page.Data,
+		HasNext:           page.HasNext,
 		MinUserLastSeenAt: minMs,
 	}, nil
 }
--- a/history-service/internal/service/messages.go
+++ b/history-service/internal/service/messages.go
@@ -93,9 +93,12 @@ func (s *HistoryService) LoadHistory(c *natsrouter.Context, req models.LoadHisto
 
 	redactUnavailableQuotes(page.Data, accessSince)
 	setDecodedAttachments(c, page.Data)
+	// An empty page must never claim hasNext: this RPC pages by before = oldest
+	// returned createdAt, so an empty resumable page (budget-exhausted walk over
+	// a long silent gap) would leave the client no way to advance.
 	return &models.LoadHistoryResponse{
 		Messages:          page.Data,
-		HasNext:           page.HasNext,
+		HasNext:           page.HasNext && len(page.Data) > 0,
 		MinUserLastSeenAt: minMs,
 	}, nil
 }
--- a/history-service/internal/service/room_times.go
+++ b/history-service/internal/service/room_times.go
@@ -101,7 +101,7 @@ func (s *HistoryService) resolveRoomTimes(
 	}
 
 	// A merged hint+Mongo pair can be internally inconsistent (created > last): refetch from
-	// Mongo when a hint was involved; if still inverted the room is genuinely empty — normalise.
+	// Mongo when a hint was involved; if still inverted with a real lastMsgAt, normalise.
 	if created.After(*last) {
 		if metaLast || metaCreated {
 			l, c, gerr := s.rooms.GetRoomTimes(ctx, roomID)
@@ -111,8 +111,12 @@ func (s *HistoryService) resolveRoomTimes(
 			last = &l
 			created = &c
 		}
-		// Empty room or hint-refetch still inconsistent — collapse the range.
-		if created.After(*last) {
+		// Still inverted with a real lastMsgAt — corrupt pair; collapse the
+		// range. A zero lastMsgAt stays zero: "not recorded" means UNKNOWN,
+		// not "empty room" — the room may hold messages (legacy docs, failed
+		// lastMsgAt update); callers treat zero as unknown (LoadHistory skips
+		// its before-cap, walkBounds ceilings at now+skew).
+		if !last.IsZero() && created.After(*last) {
 			last = created
 		}
 	}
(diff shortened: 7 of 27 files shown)


julianshen:
**8 findings · 5 unique issues · 5 inline comments posted** — 🔴 7 blocking · 🟠 1 warning

<details>
<summary>📒 Files reviewed (27)</summary>

* `CLAUDE.md`
* `bot-message-worker/deploy/docker-compose.yml`
* `data-migration/es-index-migrator/deploy/docker-compose.yml`
* `docker-local/cassandra/init/10-table-messages_by_room.cql`
* `docs/cassandra_message_model.md`
* `docs/client-api.md`
* `docs/superpowers/plans/2026-08-14-history-service-hasnext.md`
* `docs/superpowers/plans/2026-08-16-loadhistory-missing-lastmsgat.md`
* `docs/superpowers/specs/2026-08-14-history-service-hasnext-design.md`
* `docs/superpowers/specs/2026-08-16-loadhistory-missing-lastmsgat-design.md`
* `history-service/internal/mongorepo/room_test.go`
* `history-service/internal/service/messages_lastmsgat_test.go`
* `history-service/internal/service/room_times.go`
* `history-service/internal/service/room_times_test.go`
* `message-worker/main.go`
* `tools/cdc-verify/main.go`
* `tools/cdc-verify/main_test.go`
* `tools/loadgen/deploy/k8s/values-local.yaml`
* `tools/loadgen/main.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

history-service/internal/service/messages.go
```diff
@@ -93,8 +93,12 @@ func (s *HistoryService) LoadHistory(c *natsrouter.Context, req models.LoadHisto
 
 	redactUnavailableQuotes(page.Data, accessSince)
 	setDecodedAttachments(c, page.Data)
+	// An empty page must never claim hasNext: this RPC pages by before = oldest
+	// returned createdAt, so an empty resumable page (budget-exhausted walk over
+	// a long silent gap) would leave the client no way to advance.
 	return &models.LoadHistoryResponse{
 		Messages:          page.Data,
+		HasNext:           page.HasNext && len(page.Data) > 0,
```
julianshen, line 101: _🎯 pagination-correctness_ | _🔴 Blocking_ | _`cassandra`, `distributed-system` (2 rules)_

**Empty resumable pages incorrectly terminate history pagination.**

The preceding comment identifies a reachable failure: after a long silent gap, the bucket walker can exhaust its budget and return an empty page with `page.HasNext=true` while older messages still exist. This assignment converts that state to `false`, so the client treats the response as the history boundary and permanently stops before those messages. Returning the raw flag would instead cause an identical-request loop because no message supplies a new `before` value. Preserve progress by exposing the walker's opaque continuation cursor in the response/request, or continue the walk server-side until it finds data or reaches the true boundary within an explicit overall work limit.

<details>

- **`cassandra`**:100: Both GetMessagesBefore and GetMessagesBetweenDesc can return Page{Data: nil, HasNext: true} after exhausting the Cassandra bucket-walk budget across a long silent gap. Masking that state to false tells clients the history boundary was reached, so they stop even though older messages exist. This directly contradicts the documented hasNext contract and makes that older history unreachable. Provide a progress token/bucket-boundary cursor that the next request can submit (or make the repository continue across empty resumable pages until it finds data or reaches the true boundary); then preserve HasNext until the actual terminal boundary.

</details>

<details>

```
In `history-service/internal/service/messages.go` around line 101: The preceding comment identifies a reachable failure: after a long silent gap, the bucket walker can exhaust its budget and return an empty page with `page.HasNext=true` while older messages still exist. This assignment converts that state to `false`, so the client treats the response as the history boundary and permanently stops before those messages. Returning the raw flag would instead cause an identical-request loop because no message supplies a new `before` value. Preserve progress by exposing the walker's opaque continuation cursor in the response/request, or continue the walk server-side until it finds data or reaches the true boundary within an explicit overall work limit.

```

</details>

julianshen, line 101: _🎯 correctness_ | _🔴 Blocking_ | _`silent-failure`_

**Bucket-budget exhaustion is reported as the end of history.**

The walker can return an empty page with `page.HasNext=true` after exhausting its bucket budget across a long silent gap. Converting that state to `HasNext=false` returns an apparently terminal success even though older messages remain, so clients stop pagination and silently lose access to that history. Preserve a continuation mechanism for this state: add cursor support, continue walking server-side, or return a retriable error instead of reporting a terminal page.

<details>

```
In `history-service/internal/service/messages.go` around line 101: The walker can return an empty page with `page.HasNext=true` after exhausting its bucket budget across a long silent gap. Converting that state to `HasNext=false` returns an apparently terminal success even though older messages remain, so clients stop pagination and silently lose access to that history. Preserve a continuation mechanism for this state: add cursor support, continue walking server-side, or return a retriable error instead of reporting a terminal page.

```

</details>

docs/client-api.md
```diff
@@ -3011,6 +3011,7 @@ Live reaction events (`MessageReactedPayload`) carry a single-actor delta (`{sho
 | Field | Type | Notes |
 |---|---|---|
 | `messages` | array<Message> | Most-recent first. See [Message schema](#message-schema). |
+| `hasNext` | boolean | `true` if older messages may exist beyond this page — fetch the next page with `before` = the oldest returned message's `createdAt`. `false` once the caller's history boundary (room start, history floor, or access window) is reached. Conservative: occasionally `true` when nothing older remains; the following fetch then returns an empty page with `hasNext=false`. An empty page always has `hasNext=false`. |
```
julianshen, line 3043: _🏷️ ordering_ | _🟠 Warning_ | _`distributed-system`_

**A timestamp-only continuation cannot represent messages with equal timestamps.**

The new pagination instruction uses only the oldest message's `createdAt` as the continuation position. If a page boundary splits concurrent messages sharing that timestamp, a strict `before` query skips unseen ties, while an inclusive query repeats ties; `hasNext=true` cannot prevent either outcome. Define a total order and return an opaque cursor containing the complete Cassandra clustering position, such as `(createdAt, messageId)`, while retaining the existing request form for compatibility.

<details>

```
In `docs/client-api.md` around line 3014: The new pagination instruction uses only the oldest message's `createdAt` as the continuation position. If a page boundary splits concurrent messages sharing that timestamp, a strict `before` query skips unseen ties, while an inclusive query repeats ties; `hasNext=true` cannot prevent either outcome. Define a total order and return an opaque cursor containing the complete Cassandra clustering position, such as `(createdAt, messageId)`, while retaining the existing request form for compatibility.

```

</details>

docs/superpowers/specs/2026-08-14-history-service-hasnext-design.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Changes
+
+1. **Model** — `history-service/internal/models/message.go`:
+   `LoadHistoryResponse` gains `HasNext bool` with tag `json:"hasNext"`
+   (always present, no `omitempty` — matching `LoadNextMessagesResponse`).
+   Request body unchanged.
+
+2. **Handler** — `history-service/internal/service/messages.go`,
+   `LoadHistory`: set `HasNext: page.HasNext` in the returned
```
julianshen, line 62: _📝 documentation_ | _🟠 Warning_ | _`tgd-review`_

**The approved spec documents the wrong handler assignment.**

The Semantics section and implementation correctly suppress `hasNext` for empty pages, but this Changes item still says to return the raw `page.HasNext`. The implementation plan's Step 6 contains the same stale assignment. Update both documents so future work does not remove the empty-page guard.

<details>

```
In `docs/superpowers/specs/2026-08-14-history-service-hasnext-design.md` around line 62: The Semantics section and implementation correctly suppress `hasNext` for empty pages, but this Changes item still says to return the raw `page.HasNext`. The implementation plan's Step 6 contains the same stale assignment. Update both documents so future work does not remove the empty-page guard.

```

</details>

julianshen (commented):
tGD inline review

kimberly-WY (commented):
Good improvement overall. One edge case: empty string vs null aren't treated the same downstream — might need a normalization step.

history-service/internal/service/messages.go
julianshen, line 101: _🏷️ pagination_ | _🔴 Blocking_ | _`cassandra`_

**Empty resumable pages are reported as terminal**

The Cassandra bucket walker can return `Page{Data: nil, HasNext: true}` after exhausting its bucket budget across a long silent gap. This assignment converts that resource-limit condition into a terminal response. A client then stops pagination even when `GetMessagesBefore` or `GetMessagesBetweenDesc` has older rows within the history/access floor. This is especially reachable for the newly supported missing-`lastMsgAt` rooms because their walk starts near `now`; enough empty buckets before an older message will silently hide that message. Returning raw `HasNext` would instead repeat the same `before` request, so the minimal correct design is to return an opaque continuation cursor/bucket boundary and accept it on the next request, or resume empty pages internally under an explicit total-work limit. Do not represent budget exhaustion as history-boundary exhaustion.

<details>

```
In `history-service/internal/service/messages.go` around line 101: The Cassandra bucket walker can return `Page{Data: nil, HasNext: true}` after exhausting its bucket budget across a long silent gap. This assignment converts that resource-limit condition into a terminal response. A client then stops pagination even when `GetMessagesBefore` or `GetMessagesBetweenDesc` has older rows within the history/access floor. This is especially reachable for the newly supported missing-`lastMsgAt` rooms because their walk starts near `now`; enough empty buckets before an older message will silently hide that message. Returning raw `HasNext` would instead repeat the same `before` request, so the minimal correct design is to return an opaque continuation cursor/bucket boundary and accept it on the next request, or resume empty pages internally under an explicit total-work limit. Do not represent budget exhaustion as history-boundary exhaustion.

```

</details>

history-service/internal/service/room_times.go
```diff
@@ -111,8 +111,12 @@ func (s *HistoryService) resolveRoomTimes(
 			last = &l
 			created = &c
 		}
-		// Empty room or hint-refetch still inconsistent — collapse the range.
-		if created.After(*last) {
+		// Still inverted with a real lastMsgAt — corrupt pair; collapse the
+		// range. A zero lastMsgAt stays zero: "not recorded" means UNKNOWN,
+		// not "empty room" — the room may hold messages (legacy docs, failed
```
julianshen, line 116: _🏷️ partial-failure_ | _🔴 Blocking_ | _`distributed-system`_

**A stale nonzero lastMsgAt still hides persisted messages.**

The guard only treats a missing timestamp as unknown. Concrete failure: m1 is persisted and Mongo records t1; m2 is then persisted to Cassandra, but the worker crashes or loses connectivity before updating Mongo. GetRoomTimes returns the nonzero but stale t1, this resolver accepts it, LoadHistory caps `before` at t1+1ms, and `walkBounds` uses t1 as its ceiling. The successfully persisted m2 at t2 is therefore absent from history, next-message, and surrounding-message reads. Remove asynchronously maintained `lastMsgAt` as a hard Cassandra read ceiling (use the request bound or now plus skew); retain it only as a soft optimization after verifying that no newer Cassandra data exists.

<details>

```
In `history-service/internal/service/room_times.go` around line 116: The guard only treats a missing timestamp as unknown. Concrete failure: m1 is persisted and Mongo records t1; m2 is then persisted to Cassandra, but the worker crashes or loses connectivity before updating Mongo. GetRoomTimes returns the nonzero but stale t1, this resolver accepts it, LoadHistory caps `before` at t1+1ms, and `walkBounds` uses t1 as its ceiling. The successfully persisted m2 at t2 is therefore absent from history, next-message, and surrounding-message reads. Remove asynchronously maintained `lastMsgAt` as a hard Cassandra read ceiling (use the request bound or now plus skew); retain it only as a soft optimization after verifying that no newer Cassandra data exists.

```

</details>

history-service/internal/service/messages.go
```diff
@@ -93,8 +93,12 @@ func (s *HistoryService) LoadHistory(c *natsrouter.Context, req models.LoadHisto
 
 	redactUnavailableQuotes(page.Data, accessSince)
 	setDecodedAttachments(c, page.Data)
+	// An empty page must never claim hasNext: this RPC pages by before = oldest
+	// returned createdAt, so an empty resumable page (budget-exhausted walk over
+	// a long silent gap) would leave the client no way to advance.
 	return &models.LoadHistoryResponse{
 		Messages:          page.Data,
```
julianshen, line 100: _🏷️ pagination_ | _🔴 Blocking_ | _`distributed-system`, `silent-failure` (2 rules)_

**Budget exhaustion is reported as the end of history.**

The walker can return `Data=[]` and `HasNext=true` after exhausting its bucket budget while crossing a long silent gap. This line converts that resumable state to `hasNext=false`; a client following the documented contract stops permanently even when older messages exist beyond the scanned buckets. Returning raw true would instead repeat the identical `before` request, so the protocol needs an advance mechanism. Minimally, expose and accept the walker's continuation cursor for empty resumable pages, or have the service continue from that cursor until it finds data or reaches the real boundary before constructing this response.

<details>

- **`silent-failure`**:102: The walker can return `HasNext=true` with no data when it exhausts its bucket budget while crossing a long silent gap. This assignment converts that reachable partial walk into `hasNext=false`, which the client interprets as reaching the history boundary, permanently hiding older messages. Do not acknowledge this state as terminal success: continue walking server-side until data or the true boundary is reached, add a resumable cursor, or surface an error when an empty resumable page cannot be represented by this RPC.

</details>

<details>

```
In `history-service/internal/service/messages.go` around line 100: The walker can return `Data=[]` and `HasNext=true` after exhausting its bucket budget while crossing a long silent gap. This line converts that resumable state to `hasNext=false`; a client following the documented contract stops permanently even when older messages exist beyond the scanned buckets. Returning raw true would instead repeat the identical `before` request, so the protocol needs an advance mechanism. Minimally, expose and accept the walker's continuation cursor for empty resumable pages, or have the service continue from that cursor until it finds data or reaches the real boundary before constructing this response.

```

</details>

history-service/internal/service/messages.go
julianshen, line 101: _⚡ pagination/performance_ | _🔴 Blocking_ | _`performance-and-concurrency`_

**Empty resumable walks repeatedly exhaust the bucket budget**

Preserving a zero `lastMsgAt` makes `walkBounds` start at `now+clockSkewTolerance`. For an inactive affected room whose newest message is beyond the walker's per-request bucket budget, the repository returns the documented empty resumable page (`page.HasNext=true`, no data). This line discards that continuation and reports a terminal page. The client cannot advance its timestamp-based cursor, and every later room load repeats the same maximum-budget Cassandra scan without ever reaching the messages. Propagate a walker cursor through the Load History request/response, as the cursor RPCs do, or resume empty pages server-side under cancellation and a bounded overall deadline until data or the true boundary is reached. Repairing/backfilling missing `lastMsgAt` would also prevent these repeated scans.

<details>

```
In `history-service/internal/service/messages.go` around line 101: Preserving a zero `lastMsgAt` makes `walkBounds` start at `now+clockSkewTolerance`. For an inactive affected room whose newest message is beyond the walker's per-request bucket budget, the repository returns the documented empty resumable page (`page.HasNext=true`, no data). This line discards that continuation and reports a terminal page. The client cannot advance its timestamp-based cursor, and every later room load repeats the same maximum-budget Cassandra scan without ever reaching the messages. Propagate a walker cursor through the Load History request/response, as the cursor RPCs do, or resume empty pages server-side under cancellation and a bounded overall deadline until data or the true boundary is reached. Repairing/backfilling missing `lastMsgAt` would also prevent these repeated scans.

```

</details>

julianshen (commented):
tGD inline review

hmchangw (approved):
Thanks!!

message-worker/main.go
```diff
@@ -44,7 +44,7 @@ type config struct {
 	CassandraPassword  string                  `env:"CASSANDRA_PASSWORD"   envDefault:""`
 	CassandraNumConns  int                     `env:"CASSANDRA_NUM_CONNS"  envDefault:"8"`
 	MaxWorkers         int                     `env:"MAX_WORKERS"          envDefault:"100"`
-	MessageBucketHours int                     `env:"MESSAGE_BUCKET_HOURS" envDefault:"72"`
+	MessageBucketHours int                     `env:"MESSAGE_BUCKET_HOURS" envDefault:"360"`
```
julianshen, line 47: _🏷️ cassandra-data-modeling_ | _🔴 Blocking_ | _`cassandra`_

**Changing the bucket width makes existing message partitions unreadable.**

Existing `messages_by_room` rows were written under `(room_id, floor(created_at/72h)*72h)`, while updated readers and writers use 360-hour boundaries. Four of every five old 72-hour partition keys do not coincide with the new key, so historical rows become invisible; during a rolling deployment, old and new service versions also cannot reliably read each other's writes. Changing TWCS does not relocate partition-key values. Keep all services at 72 hours until implementing a versioned table or coordinated dual-read/dual-write and backfill strategy, followed by a cutover.

<details>

```
In `message-worker/main.go` around line 47: Existing `messages_by_room` rows were written under `(room_id, floor(created_at/72h)*72h)`, while updated readers and writers use 360-hour boundaries. Four of every five old 72-hour partition keys do not coincide with the new key, so historical rows become invisible; during a rolling deployment, old and new service versions also cannot reliably read each other's writes. Changing TWCS does not relocate partition-key values. Keep all services at 72 hours until implementing a versioned table or coordinated dual-read/dual-write and backfill strategy, followed by a cutover.

```

</details>

history-service/internal/config/config.go
```diff
@@ -58,9 +58,9 @@ type Config struct {
 	Cassandra               CassandraConfig `envPrefix:"CASSANDRA_"`
 	Mongo                   MongoConfig     `envPrefix:"MONGO_"`
 	NATS                    NATSConfig      `envPrefix:"NATS_"`
-	MessageBucketHours      int             `env:"MESSAGE_BUCKET_HOURS"        envDefault:"72"`
+	MessageBucketHours      int             `env:"MESSAGE_BUCKET_HOURS"        envDefault:"360"`
```
julianshen, line 61: _🏷️ rolling-deploy_ | _🔴 Blocking_ | _`distributed-system`, `silent-failure`, `tgd-review` (3 rules)_

**Changing bucket width makes existing and mixed-version messages unreadable.**

Existing rows were written under partition key `(room_id, floor(createdAt/72h)*72h)`. After this change, history-service queries only 360-hour bucket keys, which differ for most timestamps, so historical rows remain committed in Cassandra but silently disappear. During a rolling deployment, old and new workers/readers likewise cannot reliably see each other's writes. Keep the 72-hour derivation, or introduce a versioned migration with dual reads/writes and backfill before cutting every producer and consumer over; startup validation should also reject mixed bucket settings.

<details>

- **`tgd-review`**:61: Existing rows were partitioned with `floor(timestamp / 72h) * 72h`, while the updated reader will query only keys computed with 360-hour windows. Within each 360-hour interval, four of the five old 72-hour partition keys therefore will not be queried. A rolling deployment also splits new reads and writes between incompatible partitions, exactly the silent-loss condition documented in CLAUDE.md. Keep 72 hours until a coordinated migration exists, or implement dual-window reads/backfill and an atomic cutover.
- **`silent-failure`**:61: Existing rows were partitioned using 72-hour bucket values, while the new reader computes 360-hour values. For most timestamps those values differ, so even a coordinated rollout will query different partitions and return apparently successful but incomplete history; a rolling deployment can also hide newly written messages while services disagree. Keep the 72-hour setting until a dual-read/backfill migration is implemented, then switch writers and readers through a staged migration.

</details>

<details>

```
In `history-service/internal/config/config.go` around line 61: Existing rows were written under partition key `(room_id, floor(createdAt/72h)*72h)`. After this change, history-service queries only 360-hour bucket keys, which differ for most timestamps, so historical rows remain committed in Cassandra but silently disappear. During a rolling deployment, old and new workers/readers likewise cannot reliably see each other's writes. Keep the 72-hour derivation, or introduce a versioned migration with dual reads/writes and backfill before cutting every producer and consumer over; startup validation should also reject mixed bucket settings.

```

</details>

history-service/internal/service/room_times.go
```diff
@@ -111,8 +111,12 @@ func (s *HistoryService) resolveRoomTimes(
...
+		// lastMsgAt update); callers treat zero as unknown (LoadHistory skips
+		// its before-cap, walkBounds ceilings at now+skew).
+		if !last.IsZero() && created.After(*last) {
```
julianshen, line 119: _🏷️ partial-failure_ | _🔴 Blocking_ | _`distributed-system`, `silent-failure` (2 rules)_

**A stale nonzero lastMsgAt still hides committed messages.**

The new guard handles only an absent timestamp. A reachable partial failure remains: m1 is persisted and Mongo records t1; m2 is then persisted to Cassandra, but the worker crashes or loses Mongo connectivity before updating `lastMsgAt`. The resolver accepts stale nonzero t1, after which LoadHistory and `walkBounds` use it as a hard ceiling and omit m2 at t2. Do not use the asynchronously maintained Mongo timestamp as an authoritative Cassandra upper bound; read through the request bound or `now + clockSkewTolerance`, retaining `lastMsgAt` only as a soft optimization after checking for newer Cassandra data. Add a regression case with a stale nonzero timestamp.

<details>

- **`silent-failure`**:120: The new guard handles only a missing timestamp. If Cassandra persists a later message but the asynchronous Mongo lastMsgAt update fails, Mongo retains a nonzero stale timestamp. LoadHistory and walkBounds then use that stale value as a hard ceiling, silently omitting the persisted message from history, next-message, and surrounding-message reads. Do not use asynchronously maintained lastMsgAt as a hard Cassandra boundary unless newer data has been ruled out; use the request bound or now plus skew instead.

</details>

<details>

```
In `history-service/internal/service/room_times.go` around line 119: The new guard handles only an absent timestamp. A reachable partial failure remains: m1 is persisted and Mongo records t1; m2 is then persisted to Cassandra, but the worker crashes or loses Mongo connectivity before updating `lastMsgAt`. The resolver accepts stale nonzero t1, after which LoadHistory and `walkBounds` use it as a hard ceiling and omit m2 at t2. Do not use the asynchronously maintained Mongo timestamp as an authoritative Cassandra upper bound; read through the request bound or `now + clockSkewTolerance`, retaining `lastMsgAt` only as a soft optimization after checking for newer Cassandra data. Add a regression case with a stale nonzero timestamp.

```

</details>

data-migration/es-index-migrator/deploy/docker-compose.yml
```diff
@@ -25,7 +25,7 @@ services:
       - USER_ROOM_INDEX=user-room-mv-site-local
       - MIGRATION_START_AT=2025-07-01T00:00:00Z
       - MIGRATION_END_AT=2026-07-01T00:00:00Z
-      - MESSAGE_BUCKET_HOURS=72
+      - MESSAGE_BUCKET_HOURS=360
```
julianshen, line 28: _🏷️ elasticsearch-migration_ | _🔴 Blocking_ | _`elasticsearch`_

**The migrator skips rows stored with the previous bucket width.**

The Elasticsearch migrator now scans Cassandra using 360-hour bucket keys, but existing messages were partitioned using 72-hour keys. Because changing this setting does not relocate existing rows, a 360-hour walk will not address four of the five historical 72-hour partitions in each 360-hour interval, so those messages can be silently omitted from the Elasticsearch index. Coordinate this change with a Cassandra re-bucketing migration or make the migrator read both layouts during transition; do not switch its bucket width until historical and concurrent writes are covered.

<details>

```
In `data-migration/es-index-migrator/deploy/docker-compose.yml` around line 28: The Elasticsearch migrator now scans Cassandra using 360-hour bucket keys, but existing messages were partitioned using 72-hour keys. Because changing this setting does not relocate existing rows, a 360-hour walk will not address four of the five historical 72-hour partitions in each 360-hour interval, so those messages can be silently omitted from the Elasticsearch index. Coordinate this change with a Cassandra re-bucketing migration or make the migrator read both layouts during transition; do not switch its bucket width until historical and concurrent writes are covered.

```

</details>

julianshen (commented):
tGD inline review

docker-local/cassandra/init/10-table-messages_by_room.cql
```diff
@@ -38,7 +38,7 @@ CREATE TABLE IF NOT EXISTS chat.messages_by_room (
   AND compaction = {
     'class': 'TimeWindowCompactionStrategy',
     'compaction_window_unit': 'HOURS',
-    'compaction_window_size': '72'
+    'compaction_window_size': '360'
```
julianshen, line 41: _🏷️ schema-migration_ | _🟠 Warning_ | _`cassandra`_

**Existing tables will not receive the new TWCS window.**

The compaction option is inside `CREATE TABLE IF NOT EXISTS`, which is a no-op when `chat.messages_by_room` already exists. Fresh environments therefore use a 360-hour TWCS window while existing clusters retain 72 hours, producing environment-dependent compaction, SSTable, and tombstone behavior. Add a versioned `ALTER TABLE chat.messages_by_room WITH compaction = ...` migration for every deployed environment; retain this init definition for fresh clusters.

<details>

```
In `docker-local/cassandra/init/10-table-messages_by_room.cql` around line 41: The compaction option is inside `CREATE TABLE IF NOT EXISTS`, which is a no-op when `chat.messages_by_room` already exists. Fresh environments therefore use a 360-hour TWCS window while existing clusters retain 72 hours, producing environment-dependent compaction, SSTable, and tombstone behavior. Add a versioned `ALTER TABLE chat.messages_by_room WITH compaction = ...` migration for every deployed environment; retain this init definition for fresh clusters.

```

</details>

PR #286, opened 2026-08-16
feat: instrument inbound NATS requests and three more services

flywindy:
## Summary

#272 instrumented the **asynchronous** half: the JetStream consumers in `message-gatekeeper`, `message-worker`, `broadcast-worker`, and `notification-worker` — loop liveness, per-delivery disposition, redeliveries, publish attempts.

This PR lights up the **synchronous** half, and the services #272 did not reach.

| | After #272 | After this PR |
|---|---|---|
| Async consumer paths in the four workers | ✅ | ✅ |
| Requests those workers **send** | ✅ | ✅ |
| Requests any service **receives** | ❌ | ✅ |
| `room-service`, `history-service` | ❌ | ✅ |
| `room-worker` | ❌ | ✅ |

The gap that motivated it: you could see *"gatekeeper asked history-service and timed out"*, but never *"history-service took 500 requests this minute, answered 40% `forbidden`, p99 800 ms"*. Only the caller's side was visible, never the answerer's.

That completes the seven services in the NATS failure-test scope (`docs/load-testing/nats-jetstream-failure-test-plan.md`).

## What each addition buys you

| Addition | What was invisible without it |
|---|---|
| **`pkg/natsrouter`** | How many requests a service **receives**, how fast it answers, and with what result — plus reply-publish failures, which at the client look identical to a slow handler. Opt-in via `WithMetrics`; a router built without it is unchanged. |
| **`pkg/natsmetrics` classifiers** | (Enabler.) Subjects carry account, room, and site tokens. Every label is derived from **fixed token positions** so none of those can become one; an unrecognised subject normalises to `unknown` rather than minting a label. |
| **`history-service`** | JetStream publish failure rate. |
| **`room-service`** | Cross-site call health — the first place a federation outage shows up. |
| **`room-worker`, `pkg/roomkeysender`** | Whether the ROOMS consumer is alive, and whether membership side effects are being dropped silently. |

## Review comments from #272 closed here

| Comment | Resolution |
|---|---|
| `service_name` should come from an env var — the same image deploys as `teams-message-worker` / `bot-notification-worker` / `bot-broadcast-worker` | All four #272 services plus the three added here read `OTEL_SERVICE_NAME`, defaulting to `unknown-service`. Each config test covers an explicit deployment name and the unset default. |
| `notifyKindNotification` is unused | Removed, with its pre-warmed attribute set. `allNotifyKinds` is now `push` and `unknown`. |
| Redelivery tracking costs resources; the broker already has this | Kept, now with a benchmark: `make benchmark-natsmetrics`. First delivery 119.2 ns / 1 alloc, redelivery 168.3 ns / 2 allocs on Linux amd64. Broker metrics stay the aggregate source; this adds the per-service/site/event dimension. |
| Carrying the tracked object in every context may cost memory | Benchmarked: 46.10 ns / 48 B / 1 alloc. The context holds a pointer, not a copy of the payload. |
| `chat.nats.*` should follow `_total` naming | No instrument rename — the OTel name omits `.total` and the Prometheus exporter appends it. The rule is now written into the metrics inventory, verified against a real registry: `chat.nats.requests` exports as `chat_nats_requests_total`, while `nats_slow_consumer_events_total` is left unchanged. |

## Three changes that are not pure observation

Called out because everything else here is strictly additive.

1. **`room-worker`'s loop goroutine now registers with the `WaitGroup`.** Shutdown stops the iterator and then waits on `wg`; an uncounted loop lets that wait pass through while a message `Next` already returned is still on its way to a worker. Every other instrumented worker already counts it.
2. **`room-worker`'s shutdown hook marks the loop down before stopping the iterator.** `LoopFailed` only reports a terminal cause while the loop still reads as up, so without this every clean restart posts a `terminal{reason=internal}` — the exact signal used to detect a consumer that died behind a green process. This mirrors `message-worker`.
3. **The Teams `room-worker` deploy takes its own `service_name`** (`teams-room-worker`). Both deploys of that binary set `OTEL_SERVICE_NAME=room-worker` today and differ only by `MODE`, so their series are indistinguishable. This is the one existing label this PR changes — dashboards or alerts pinning `service_name="room-worker"` and expecting to cover the Teams pod need updating.

## One semantic decision worth a reviewer's attention

The outbound clients record the **final** result, not just the transport error, so a decode failure or an unexpected remote errcode is no longer invisible.

A remote *"you are not a room member"* is deliberately **not** counted as a failure. The request/reply exchange worked and the peer answered; `chat.nats.requests` carries a transport-shaped outcome enum, so counting a business rejection there lands it in `other_error` and leaves the family non-zero at baseline — useless as a failure signal. `GetMessageReadMeta` already excluded its `CodeNotFound` for the same reason; `ListMembers` now matches.

## Validation

- `make lint` — clean
- `make test` — full suite green with `-race`
- `make sast` — `gosec` clean (`govulncheck` and `semgrep` are unavailable in the dev sandbox; both run in CI)
- Merged `origin/main` after #271; the only conflict was the `Makefile` `.PHONY` line, resolved as the union of both sides' targets


flywindy:
Four of the five findings are fixed in `4dbd439`; one is declined.

**Suffix classification wasn't anchored.** Correct, and worth more than the "quick win" label — it is the one finding here that could produce wrong data. A Red test with four probe subjects misclassified three of them (`chat.dynamic.member.list`, `chat.room.canonical.site-a.member.add`, `chat.inbox.site-a.external.create`). The suffix cases now sit behind the same request-shape check the rest of the file already used, so the anchoring rule holds throughout instead of in half the function.

**Default service identity.** Added to all three config tests. An empty `service_name` silently orphans every series joined through it, so the default deserves its own case rather than riding on the explicit one.

**Generated mock.** The metrics seam is this package's own interface, so it now uses a generated `MockattemptRecorder`, matching the request recorder in room-service. I kept the JetStream double handwritten and said so in the code: `o11ynats.JetStream` is an external interface this package does not own, only `PublishMsg` is exercised, and the guidance to define interfaces at the consumer cuts against generating a mock for someone else's surface.

**`_total` suffix.** Fixed, and I verified it rather than taking the learning on faith — a real `prometheus.Registry` with the OTel exporter shows `nats_slow_consumer_events_total` exporting unchanged while `chat.nats.requests` becomes `chat_nats_requests_total`. The doc now states that the exporter strips the existing token.

### Declining the `New` → `NewPublisher` rename

The premise is wrong: `New` is not a constructor changed in this PR. It exists in `main` at `history-service/internal/publisher/publisher.go:19`; this PR only adds a variadic option to it. Renaming would also churn `user-service/main.go`, which this PR does not otherwise touch.

On the guideline itself — in a package named `publisher`, `New` returning `*Publisher` *is* the `New<Type>` pattern. `publisher.NewPublisher(...)` stutters at the call site, which is what the Go naming conventions the guideline derives from exist to prevent. Same shape as `errors.New`.

`make lint` clean, full `make test` green with `-race`.

---

flywindy:
@coderabbitai review

Scope changed since your last completed review: `room-worker` and `pkg/roomkeysender` are now in this PR rather than #283, so the七 services in the NATS failure-test scope are all instrumented here. The PR description has been rewritten to match — in particular the "Three changes that are not pure observation" section, which is where I would focus.

---

julianshen:

### Related work

- [PR #272](https://github.com/hmchangw/newchat/pull/272) — feat: add application-side NATS failure metrics (merged)
- [PR #271](https://github.com/hmchangw/newchat/pull/271) — feat\(loadgen\): track per-message delivery outcomes with a durable evidence ledger (merged)

### Local review memory

- `history-service/internal/publisher/publisher.go:45` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟠 Major_ _⚡ Quick win_ **Rename the constructor to NewPublisher .** New is a constructor changed in this PR. The Go guidelines require New<Type> constructor names. Update callers with the rena… ([thread](https://github.com/hmchangw/newchat/pull/286#discussion_r3790975283))

<details>
<summary>📒 Files reviewed (39)</summary>

* `Makefile`
* `broadcast-worker/main.go`
* `history-service/internal/publisher/mock_publisher_test.go`
* `history-service/internal/publisher/publisher_test.go`
* `message-gatekeeper/config_test.go`
* `message-worker/config_test.go`
* `message-worker/main.go`
* `pkg/natsmetrics/metrics_benchmark_test.go`
* `pkg/natsrouter/context.go`
* `pkg/natsrouter/router.go`
* `pkg/roomkeysender/roomkeysender.go`
* `pkg/roomkeysender/roomkeysender_test.go`
* `room-service/main.go`
* `room-worker/consumer_config_test.go`
* `room-worker/deploy/teams/docker-compose.yml`
* `room-worker/handler.go`
* `room-worker/main.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

julianshen (commented):
tGD inline review

room-worker/main.go
```diff
@@ -258,25 +278,35 @@ func main() {
...
 
+	wg.Add(1)
 	go func() {
+		// The loop itself is counted so shutdown, which stops the iterator and
+		// then waits on wg, cannot pass through while a message Next already
+		// returned is still on its way to a worker.
+		defer wg.Done()
 		for {
 			msgCtx, msg, err := iter.Next()
 			if err != nil {
```
julianshen, line 291: _🏷️ reliability_ | _🔴 Blocking_ | _`silent-failure`_

**An unexpected iterator failure silently stops message consumption.**

When iter.Next() returns an unexpected error, LoopFailed only records a metric and the goroutine returns. The failure is not sent to main, logged, retried, or used to cancel the process, so the worker can remain running while permanently consuming no room events and allowing the stream backlog to grow. Propagate the error through an error channel or shared cancellation path so the service shuts down and restarts (while preserving the existing clean-shutdown handling after iter.Stop()).

<details>

```
In `room-worker/main.go` around line 291: When iter.Next() returns an unexpected error, LoopFailed only records a metric and the goroutine returns. The failure is not sent to main, logged, retried, or used to cancel the process, so the worker can remain running while permanently consuming no room events and allowing the stream backlog to grow. Propagate the error through an error channel or shared cancellation path so the service shuts down and restarts (while preserving the existing clean-shutdown handling after iter.Stop()).

```

</details>

room-service/main.go
```diff
@@ -27,6 +28,7 @@ import (
 )
 
 type config struct {
+	ServiceName   string `env:"OTEL_SERVICE_NAME"          envDefault:"unknown-service"`
```
julianshen, line 31: _🏷️ observability_ | _🟠 Warning_ | _`distributed-system`_

**A shared fallback service name merges failures across services.**

The supported unset path assigns `unknown-service` to every adopter. If room-service and room-worker, for example, run at the same site without `OTEL_SERVICE_NAME`, their NATS metric series share the same `service_name` and `site`; successes from one can then mask request or publish failures from the other, and the documented `target_info` join is no longer uniquely attributable. This regresses the previous service-specific labels during configuration omissions. Use each binary's established service name as its default (`room-service`, `history-service`, etc.), while retaining explicit names for bot/Teams deployments, and update the corresponding default-value tests.

<details>

```
In `room-service/main.go` around line 31: The supported unset path assigns `unknown-service` to every adopter. If room-service and room-worker, for example, run at the same site without `OTEL_SERVICE_NAME`, their NATS metric series share the same `service_name` and `site`; successes from one can then mask request or publish failures from the other, and the documented `target_info` join is no longer uniquely attributable. This regresses the previous service-specific labels during configuration omissions. Use each binary's established service name as its default (`room-service`, `history-service`, etc.), while retaining explicit names for bot/Teams deployments, and update the corresponding default-value tests.

```

</details>

flywindy:
@julianshen — on the `unknown-service` fallback finding, I'd push back. The series-merge mechanism is real, but the proposed fix would break the thing this change was made to fix.

**`unknown-service` is the existing repo convention, with a documented rationale.** `pkg/obs/obs.go:77-81` already uses exactly this default for `service.name` on every span, metric, and log:

```go
// ServiceName drives service.name on every span/metric/log. It defaults to a
// visible placeholder rather than being required so a missing env degrades to
// mislabeled telemetry, not a startup crash-loop; production deploys MUST set
// OTEL_SERVICE_NAME (a "unknown-service" service map entry is the signal it wasn't).
ServiceName string `env:"OTEL_SERVICE_NAME" envDefault:"unknown-service"`
```

The seven services here read the same variable with the same default precisely so the metric label and the OTel resource agree.

**Defaulting per-binary would make them disagree in exactly the case it targets.** With a per-binary fallback, a pod that never sets `OTEL_SERVICE_NAME` would emit `service_name="room-worker"` on `chat.nats.*` while its OTel resource still resolves `service.name` to `unknown-service` — so joining the NATS metrics to traces/logs through `target_info` would silently fail. That join is the reason #272 asked for the label to come from the env var rather than a literal in the first place.

**The merge scenario requires a deployment whose telemetry is already broken.** All seven adopters set the variable explicitly, including the bot and Teams identities:

| Service | Values in `deploy/` |
|---|---|
| message-gatekeeper | `message-gatekeeper` |
| message-worker | `message-worker`, `teams-message-worker` |
| broadcast-worker | `broadcast-worker`, `bot-broadcast-worker` |
| notification-worker | `notification-worker`, `bot-notification-worker` |
| history-service | `history-service` |
| room-service | `room-service` |
| room-worker | `room-worker`, `teams-room-worker` |

A pod that reaches `unknown-service` has broken traces and logs too, not just merged metric series — and `unknown-service` appearing in `sum by (service_name) (...)` is the designed signal for that, which a per-binary default would hide.

Happy to change it if you'd rather have per-binary fallbacks, but it should then change in `pkg/obs` at the same time so the label and the resource stay consistent — that's a wider change than this PR.

On the `room-worker/main.go` `iter.Next()` path: agreed it isn't a regression here. Consumer recovery supervision is deliberately sequenced *after* the NATS failure campaign runs — the campaign's scenario F11 accepts either "self-heals" or "clearly requires restart" as an outcome, and landing recovery first would make F11 measure the mitigation instead of the current behaviour.

---

pkg/natsmetrics/subject.go
```diff
@@ -23,3 +23,165 @@ func EventTypeFromSubject(subj string) EventType {
...
+		return EventRoomCreate
+	default:
+		return EventUnknown
+	}
+}
+
+// RequestOperationFromSubject maps concrete request subjects to the bounded
+// operation vocabulary used by natsrouter. Dynamic account, room, and site
+// tokens are inspected only for shape and never become labels.
+func RequestOperationFromSubject(subj string) Operation {
```
mliu33, line 58: Is this metrics used for client-facing request/reply ? If yes, could we have label for nats "subject" ? Just like http metrics "uri". It is necessary to draw dashboard on RPS and Latency and debug for each endpoint, especially when we need to focus on some of the important ones.

If we just put "subject" in the label, I think there is no need for this operation label. In addition, client-facing and server-to-server (chat.server.>) should be distinguishable.

Thanks

pkg/natsrouter/router.go
```diff
@@ -187,15 +210,19 @@ func (r *Router) addRoute(pattern string, handlers []HandlerFunc) {
 	all = append(all, handlers...)
 
 	natsHandler := func(msgCtx context.Context, m *nats.Msg) {
+		started := time.Now()
+		operation := natsmetrics.RequestOperationFromSubject(m.Subject)
```
mliu33, line 214: Like previous comment, I think we should just put subject as the label. What do u think ?

pkg/natsmetrics/subject.go
```diff
@@ -23,3 +23,165 @@ func EventTypeFromSubject(subj string) EventType {
...
+
+func isMigrationFamily(tokens []string, family string) bool {
+	return len(tokens) >= 5 && tokens[0] == "chat" && tokens[1] == "migration" &&
+		tokens[2] == "internal" && tokens[4] == family
+}
+
+// PublishLabelsFromSubject maps a publish destination to closed labels. It is
+// intentionally conservative: a new subject family remains unknown until it
+// receives an explicit bounded mapping.
+func PublishLabelsFromSubject(subj string) (DestinationKind, Operation) {
```
mliu33, line 125: Same for previous comment, maybe it's better just to use subject as label.

mliu33 (approved):
Awesome! Thanks! I have some suggestions on metrics labels that we can discuss further

PR #288, opened 2026-08-16
feat: per-cache Valkey memory attribution

hmchangw:
## Why

Valkey reports one `used_memory` per node and nothing decomposes it by application cache. A node sitting at 6 GiB gives no indication which of our caches owns it.

The failure mode depends on `maxmemory-policy`. Under `noeviction`, a cache growing unboundedly makes *writes fail* while the memory graph stays flat and unattributed. Under `allkeys-lru` it's quieter and worse: one fat cache silently evicts another's hot entries, visible only as an unexplained rise in `cache_misses_total`.

## What

Four commits, reviewable independently.

### 1. `refactor(cachekeys)` — centralize Valkey key construction

Key patterns were string literals at their call sites, so nothing could enumerate them. That blocks attribution outright: a scanner cannot classify a key back to its owning cache without a registry.

`pkg/cachekeys` now holds every Valkey keyspace as a `Keyspace` (literal prefix, opaque variable segment, literal suffix) plus the builder that constructs it. Registration is a precondition of building a key, mirroring how `pkg/outbox.Publish` rejects an event type in no consumer filter set.

All keys are byte-identical to the literals they replace — the builder tests pin each one, since changing a key orphans its live cached entries on deploy.

Matching is on prefix/suffix rather than a glob, so a key containing glob metacharacters cannot be misclassified. The tests assert no two patterns overlap and that each builder's output classifies back to its own name.

### 2. `feat(cache-metrics-worker)` — measure and export

`pkg/cachescan` walks the keyspace with `SCAN`, classifies each key, measures a sample with `MEMORY USAGE`, and extrapolates per-cache bytes. `cache-metrics-worker` runs it on a timer and exports:

```
valkey_cache_keys{cache="roomsubs"}
valkey_cache_bytes{cache="roomsubs"}
valkey_cache_keys_total / valkey_cache_bytes_total
```

The `cache` label matches the existing `cache_hits_total` series, so one panel can show hit rate and memory share together — which is what makes the metric actionable ("60% of memory, 20% hit rate" is a decision).

It stands alone rather than riding in an existing service because the scan is O(keyspace) and must not share a process with a request path.

### 3. `refactor(cachekeys)` — register the badge cache keyspaces

`pkg/badgecache` built its two keys from string literals, so neither was in the registry and both would have reported as `unclassified`. The badge set is a per-account SET that grows with unread rooms — exactly the kind of cache this attribution exists to size. It also stood as a live counterexample to the rule commit 1 adds to CLAUDE.md.

Registered as two keyspaces rather than one label: the marker is a fixed-size sentinel per account while the set grows, so folding them together would hide which of the two drives the memory. `Key`/`MarkerKey` keep their exported signatures and delegate to the builders — byte-identical output.

### 4. `Merge origin/main` — absorb #188's much larger keyspace

#188 landed an L2 tier for `roommetacache`, a schema-version bump plus a legacy lane for `roomsubcache`, and five new caches (`roomtimescache`, `sessioncache`, `subauthcache`, `userstore`'s L2, `atrest`'s DEK store). The registry was incomplete the moment that merged — nine keyspaces would have reported as `unclassified`, and this PR's own CLAUDE.md rule would have shipped with nine counterexamples in tree.

**Final tally: twenty-two keyspaces registered, twenty-three call sites across fourteen files.** Three conflicts, all additive: `roommetacache/valkey.go` was deleted upstream and replaced by `l2.go` (took the deletion, routed its two builders through the registry); `roomsubcache` kept main's v4 bump and legacy v3 lane with both builders delegating; `valkeyutil/valkey_test.go` had tests appended at the same spot on both sides.

Two gates main added since caught real work, both fixed here: the instrument-documentation gate (#337) required the four `valkey_cache_*` gauges in `docs/specs/o11y/nats-metrics-contract.md`, and semgrep's new `metrics-no-per-call-attribute-set` rule correctly flagged `cachescan` building an `attribute.NewSet` per cache per tick — the label space is closed by construction, so the sets are now precomputed once from the registry.

## Design notes worth reviewing

- **`unclassified` is its own series.** A cache added without a `cachekeys` entry surfaces as unclassified rather than vanishing from the breakdown. Detection is automatic; naming stays manual — deriving labels from key structure would give unbounded cardinality on `idem:`.
- **Versioned caches register both generations under separate labels** (`roommeta` / `roommeta.legacy`, `roomsubs` / `roomsubs.legacy`). A legacy lane exists so a rolling deploy can still bust it; a legacy lane you cannot see on the breakdown is one you cannot confirm has drained before deleting it.
- **Every registered cache is seeded at zero.** A cache dropping to zero keys reports zero instead of leaving a stale gauge that reads as a healthy steady state.
- **`MinSamples`.** At a 1-in-100 sample rate, a cache holding 4 keys would never be measured and would report zero bytes despite holding data.
- **A failed scan does not end the loop.** Ending it would freeze every gauge at its last value — indistinguishable from a healthy steady state.
- **The badge marker nests under the badge set's prefix** (`badge:fresh:{acct}` vs `badge:{acct}`) yet the patterns cannot overlap: the set's prefix ends in `{`, which the marker's literal `fresh:` segment never matches. A dedicated collision test pins that, alongside the room hash-tag and versioned/legacy ones.
- **`valkeyutil.Cluster`** unwraps the raw cluster client for `SCAN`/`MEMORY USAGE`. Widening `valkeyutil.Client` instead would force every implementation and test fake in the repo to grow methods it has no use for.

## Reported bytes are attributed bytes

**They deliberately do not sum to `used_memory`.** Measured on Valkey 7.2 with ~75-byte string values, `MEMORY USAGE` reported **160 B/key against a true `used_memory` delta of 196 B/key** (~82%) — it excludes each key's share of the dict bucket array and the cluster slot index, and a node carries a fixed baseline belonging to no cache.

Read the gauges as per-cache share and trend; read `used_memory` for what the node holds. Documented on the package, in CLAUDE.md, and in the metrics contract.

## Two things the extraction surfaced

Both preserved as-is — this PR documents the keyspace, it does not migrate it:

- `roommetacache` hash-tags its room ID (`room:{id}:meta:v3`) while `roomsubcache` does not (`room:v4:id:subs`), putting the same room in different cluster slots. The stated house convention isn't actually held.
- `tools/seed-sample-data` carried its own copy of search-service's `searchservice:restrictedrooms:%s` literal. Two literals, one keyspace, nothing connecting them. Now impossible.

## CLAUDE.md

New **Valkey** section: every key is built by a `pkg/cachekeys` builder, a new cache must register a `Keyspace`, what `unclassified` means, and the attributed-vs-`used_memory` caveat. Same rule shape as the existing OUTBOX event-type partition rule.

## Verification

Re-run in full after the merge:

| Check | Result |
|---|---|
| `make lint` | 0 issues |
| `make test` (race) | pass, full suite, no exceptions |
| `gosec` | PASS |
| `semgrep` (repo-local `.semgrep/` rules) | **0 findings**, 19 rules over 1807 files |
| `semgrep` (registry rulesets `p/golang`, `p/security-audit`) | **not run** — `semgrep.dev` unreachable from the dev environment (proxy returns 403); CI must run these |
| `govulncheck` | **not run** — `vuln.go.dev` unreachable from the dev environment |
| Coverage | `cachekeys` 100%, `cachescan` 81.8%, worker 35.1% |
| Real-Valkey scan | verified end to end (below) |

Verified against a live Valkey 7.2.12 cluster, not only fakes:

```
cache               keys  sampled   est bytes    share
roommeta            1200       29      192000    81.4%
roomsubs             300       11       43200    18.3%
botidempotency         4        4         320     0.1%
presence.index         1        1         112     0.0%
unclassified           1        1          96     0.0%
```

Sampling landed exactly on its budget (5 floor + 1200/50 = 29), the 4-key cache was measured because of `MinSamples`, and an unregistered key surfaced as unclassified. That run predates commits 3 and 4 — it is what the `unclassified` row was catching, twice over.

Notes on the table: `gosec` initially flagged a G115 narrowing conversion in a test fake I wrote — the conversion was fixed rather than suppressed. The worker's 35.1% is `main()` being untestable wiring; its logic (`runner`, `validate`, `scanOptions`) is 100%, and comparable services sit in the same band (`outbox-worker` 37.3%, `message-worker` 47.9%).

**`pkg/cachescan/integration_test.go` compiles under `-tags integration` but has never executed** — no Docker in the dev environment. That's the one piece of this change resting on CI rather than on local evidence.

## Not included

A semgrep rule banning raw key literals against a `valkeyutil.Client` would make the registry enforceable rather than conventional. Left out deliberately: a rule precise enough to avoid false positives in test files needs its own careful pass, and a noisy blocking gate is worse than none. Now that `.semgrep/` has a tested-fixture convention (`make sast-semgrep-test`), this is more tractable than it was.


julianshen:
**11 findings · 9 unique issues · 9 inline comments posted** — 🟠 10 warning · 🔵 1 suggestion

> [!NOTE]

### Related work

- [PR #188](https://github.com/hmchangw/newchat/pull/188) — MongoDB outage survival for send \+ history load, and extract broadcast-worker's room-level writes into roomlist-worker (merged)
- [PR #337](https://github.com/hmchangw/newchat/pull/337) — perf\(o11y\): cut the NATS metrics down to what is read, and gate the next ones (merged)

- **CACHE_SCAN_COUNT can materialize an arbitrarily large SCAN page**
  - `cache-metrics-worker/main.go:39` (`oom`) — CACHE_SCAN_COUNT can materialize an arbitrarily large SCAN page
  - `cache-metrics-worker/runner.go:48` (`silent-failure`) — Persistent scan failures leave apparently healthy stale gauges
  - `cache-metrics-worker/runner.go:29` (`distributed-system`) — Failed scans leave healthy-looking gauges stale indefinitely
  - `pkg/cachescan/cachescan.go:226` (`performance-and-concurrency`) — Sampled MEMORY USAGE calls are serialized per primary.
  - `cache-metrics-worker/main.go:64` (`silent-failure`) — Invalid scan tuning silently falls back to misleading values

<details>

* `CLAUDE.md`
* `cache-metrics-worker/deploy/Dockerfile`
* `cache-metrics-worker/deploy/azure-pipelines.yml`
* `cache-metrics-worker/deploy/docker-compose.yml`
* `cache-metrics-worker/main.go`
* `cache-metrics-worker/runner.go`
* `cache-metrics-worker/runner_test.go`
* `pkg/atrest/dek_store_l2.go`
* `pkg/cachekeys/cachekeys.go`
* `pkg/cachekeys/cachekeys_test.go`
* `pkg/cachescan/cachescan.go`
* `pkg/cachescan/cachescan_test.go`
* `pkg/cachescan/integration_test.go`
* `pkg/cachescan/valkey.go`
* `pkg/roommetacache/l2.go`
* `pkg/roomtimescache/roomtimescache.go`
* `pkg/sessioncache/sessioncache.go`
* `pkg/subauthcache/subauthcache.go`
* `pkg/userstore/l2.go`
* `pkg/valkeyutil/valkey.go`
* `search-service/store_valkey.go`
* `tools/seed-sample-data/sidestores.go`
* `user-presence-service/presencestore/store.go`
* `user-presence-service/sync/valkey.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

pkg/roommetacache/l2.go
```diff
@@ -33,7 +35,7 @@ const cacheKeySchemaVersion = "v3"
 // key (pkg/roomkeystore), matching house convention for room-scoped keys.
 // The version segment trails the key so the hash tag keeps that colocation.
 func MetaKey(roomID string) string {
-	return "room:{" + roomID + "}:meta:" + cacheKeySchemaVersion
+	return cachekeys.RoomMeta(roomID)
```
julianshen, line 27: _🏷️ dead-code_ | _🟠 Warning_ | _⚡ Quick fix_ | _`dead-code`_

**The schema-version constant is now vestigial.**

`MetaKey` now delegates key construction and versioning to `cachekeys.RoomMeta`, leaving the package's `cacheKeySchemaVersion` declaration disconnected from the generated key. Keeping a second version marker here is misleading because changing it has no effect. Remove the stale constant; the authoritative `v3` suffix now lives in `pkg/cachekeys`.

<details>

```
In `pkg/roommetacache/l2.go` around line 38: `MetaKey` now delegates key construction and versioning to `cachekeys.RoomMeta`, leaving the package's `cacheKeySchemaVersion` declaration disconnected from the generated key. Keeping a second version marker here is misleading because changing it has no effect. Remove the stale constant; the authoritative `v3` suffix now lives in `pkg/cachekeys`.

```

</details>

---

pkg/cachescan/cachescan.go
```diff
@@ -0,0 +1,322 @@
...
+	return out
+}
+
+// Scan walks every primary in c and returns per-cache usage. It fails on the
+// first node error rather than reporting a partial keyspace as if it were
+// whole — a silently short report would understate a cache's share.
+func Scan(ctx context.Context, c Cluster, opts Options) (Report, error) {
+	opts = opts.normalize()
+	acc := newAccumulator(opts)
+
```
julianshen, line 208: _🎯 distributed correctness_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`_

**Cluster-wide aggregation hides the shard exhausting maxmemory**

`ForEachPrimary` feeds every primary into one shared accumulator, and `Report` carries no node identity. If cache A is concentrated on shard 1 and exhausts that shard's per-node maxmemory while a larger cache B resides on shard 2, the exported cluster totals can make B appear responsible. Preserve node or shard identity through the callback and emit per-node cache gauges, with separately named cluster totals if needed, so attribution can be joined to each node's `used_memory` and policy.

<details>

```
In `pkg/cachescan/cachescan.go` around line 208: `ForEachPrimary` feeds every primary into one shared accumulator, and `Report` carries no node identity. If cache A is concentrated on shard 1 and exhausts that shard's per-node maxmemory while a larger cache B resides on shard 2, the exported cluster totals can make B appear responsible. Preserve node or shard identity through the callback and emit per-node cache gauges, with separately named cluster totals if needed, so attribution can be joined to each node's `used_memory` and policy.

```

</details>

---

cache-metrics-worker/runner.go
```diff
@@ -0,0 +1,64 @@
...
+	return nil
+}
+
+// run scans immediately, then on every tick until ctx is done. A failed scan is
+// logged and the loop continues: ending it would freeze every gauge at its last
+// value, which reads as a healthy steady state rather than an outage.
+func (r *runner) run(ctx context.Context, interval time.Duration) {
+	scan := func() {
+		if err := r.scanOnce(ctx); err != nil {
+			slog.ErrorContext(ctx, "cache keyspace scan failed", "error", err)
```
julianshen, line 48: _🏷️ observability_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`, `silent-failure` (2 rules)_

**Persistent scan failures leave apparently healthy stale gauges**

After a successful scan, any persistent SCAN or MEMORY error returns before `Record`; `run` only logs the error while the generic health endpoint remains available. Existing gauge values therefore remain indistinguishable from a current successful scan—for example, revoked MEMORY permissions freeze attribution indefinitely and can mask cache growth or eviction pressure. Keep retrying, but expose attempt success or last-success time and fail readiness or alert once the report is stale.

<details>

- **`distributed-system`**:29: `scanOnce` returns before `Record` on any node, SCAN, or MEMORY error, while the loop only logs the failure. After one successful report, an unreachable primary can make every later scan fail while the gauges and health endpoint continue presenting the old allocation as current. Retrying does not expose this staleness until recovery. Export a scan-success status or failure counter plus last-success timestamp, and alert on its age while retaining the last complete report if desired.

</details>

<details>

```
In `cache-metrics-worker/runner.go` around line 48: After a successful scan, any persistent SCAN or MEMORY error returns before `Record`; `run` only logs the error while the generic health endpoint remains available. Existing gauge values therefore remain indistinguishable from a current successful scan—for example, revoked MEMORY permissions freeze attribution indefinitely and can mask cache growth or eviction pressure. Keep retrying, but expose attempt success or last-success time and fail readiness or alert once the report is stale.

```

</details>

---

cache-metrics-worker/main.go
```diff
@@ -0,0 +1,161 @@
...
+type config struct {
+	ValkeyAddrs    []string `env:"VALKEY_ADDRS,required" envSeparator:","`
+	ValkeyPassword string   `env:"VALKEY_PASSWORD"       envDefault:""`
+	SiteID         string   `env:"SITE_ID"               envDefault:"site-local"`
+
+	// ScanInterval is how often the keyspace is walked. Memory shares move
+	// slowly; a short interval buys nothing and costs a full walk each time.
+	ScanInterval time.Duration `env:"CACHE_SCAN_INTERVAL" envDefault:"5m"`
+	ScanTimeout  time.Duration `env:"CACHE_SCAN_TIMEOUT"  envDefault:"2m"`
+	// ScanCount is the SCAN COUNT hint (page size).
```
julianshen, line 39: _⚡ memory-exhaustion_ | _🟠 Warning_ | _⚡ Quick fix_ | _`oom`, `silent-failure` (2 rules)_

**CACHE_SCAN_COUNT can materialize an arbitrarily large SCAN page**

ScanCount has no upper bound and flows through scanOptions into go-redis Scan. On a large keyspace, an accidentally huge CACHE_SCAN_COUNT can make Valkey build and return a correspondingly huge response; scanNode retains the resulting []string while issuing MEMORY USAGE calls for that page. Concurrent primary scans multiply this peak, risking Valkey or worker OOM and container eviction. Reject values above a conservative operational maximum, rather than relying on COUNT being only a hint.

<details>

- **`silent-failure`**:64: `validate` accepts invalid scan count, sample rate, and minimum-sample values. `scanOptions` passes them through, after which `Options.normalize` silently substitutes defaults or converts a negative minimum to zero. With `CACHE_SCAN_MIN_SAMPLES=-1` and the default sample rate, a cache containing fewer than 100 keys receives no MEMORY USAGE sample and reports zero bytes while the worker starts successfully. Reject these values during startup.

<details>

```text
	if c.ScanCount <= 0 {
		return fmt.Errorf("CACHE_SCAN_COUNT must be positive, got %d", c.ScanCount)
	}
	if c.ScanSampleRate <= 0 {
		return fmt.Errorf("CACHE_SCAN_SAMPLE_RATE must be positive, got %d", c.ScanSampleRate)
	}
	if c.ScanMinSamples < 0 {
		return fmt.Errorf("CACHE_SCAN_MIN_SAMPLES must not be negative, got %d", c.ScanMinSamples)
	}
	return nil
```

</details>


</details>

<details>

```
In `cache-metrics-worker/main.go` around line 39: ScanCount has no upper bound and flows through scanOptions into go-redis Scan. On a large keyspace, an accidentally huge CACHE_SCAN_COUNT can make Valkey build and return a correspondingly huge response; scanNode retains the resulting []string while issuing MEMORY USAGE calls for that page. Concurrent primary scans multiply this peak, risking Valkey or worker OOM and container eviction. Reject values above a conservative operational maximum, rather than relying on COUNT being only a hint.

```

</details>

---

pkg/cachescan/cachescan.go
```diff
@@ -0,0 +1,322 @@
...
+// scanNode walks one primary's keyspace to cursor exhaustion.
+func scanNode(ctx context.Context, n Node, acc *accumulator, opts Options) error {
+	var cursor uint64
+	for {
+		keys, next, err := n.Scan(ctx, cursor, opts.ScanCount)
+		if err != nil {
+			return fmt.Errorf("scan at cursor %d: %w", cursor, err)
+		}
+		for _, key := range keys {
+			if !acc.count(key) {
```
julianshen, line 226: _⚡ performance_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`performance-and-concurrency`_

**Sampled MEMORY USAGE calls are serialized per primary.**

With N keys on one primary, this loop performs roughly N/100 `MEMORY USAGE` requests serially, plus N/1000 sequential `SCAN` requests. At 10 million keys—within the documented multi-million-key target—that is about 110,000 network round trips; at 2 ms each it exceeds the default two-minute timeout, discarding every report while Valkey still handles the workload. Batch sampled keys from each SCAN page into a bounded pipeline, then record successful results in the accumulator.

<details>

```
In `pkg/cachescan/cachescan.go` around line 226: With N keys on one primary, this loop performs roughly N/100 `MEMORY USAGE` requests serially, plus N/1000 sequential `SCAN` requests. At 10 million keys—within the documented multi-million-key target—that is about 110,000 network round trips; at 2 ms each it exceeds the default two-minute timeout, discarding every report while Valkey still handles the workload. Batch sampled keys from each SCAN page into a bounded pipeline, then record successful results in the accumulator.

```

</details>

---

cache-metrics-worker/runner.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+	}
+
+	scan()
+
```
julianshen, line 53: _🎯 concurrency_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`performance-and-concurrency`_

**Every worker instance duplicates the full cluster scan.**

Every process scans immediately and then starts its own ticker without a cluster-wide lease. During a rolling update where old and new instances overlap, or any multi-replica deployment, all instances concurrently walk every primary and issue the same sampled `MEMORY USAGE` workload. On a multi-million-key cluster this multiplies command load precisely during deployment or startup. Ensure only one active scanner per Valkey cluster, such as with a renewable distributed lease or guaranteed non-overlapping singleton deployment, and jitter retry timing for standby instances.

<details>

```
In `cache-metrics-worker/runner.go` around line 53: Every process scans immediately and then starts its own ticker without a cluster-wide lease. During a rolling update where old and new instances overlap, or any multi-replica deployment, all instances concurrently walk every primary and issue the same sampled `MEMORY USAGE` workload. On a multi-million-key cluster this multiplies command load precisely during deployment or startup. Ensure only one active scanner per Valkey cluster, such as with a renewable distributed lease or guaranteed non-overlapping singleton deployment, and jitter retry timing for standby instances.

```

</details>

---

botplatform-service/middleware.go
```diff
@@ -131,7 +132,7 @@ func botRateLimit(client incrExClient, perCaller, perGlobal int) gin.HandlerFunc
 		}
```
julianshen, line 132: _🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**The global bot rate-limit key remains unclassified.**

`botRateLimit` has caller and global Valkey branches, but only the caller key was moved into the registry. The global branch still increments `botrl:global`, while the registry only matches `botrl:caller:*`; consequently every scan attributes the global counter to `unclassified`. Add a fixed global-rate-limit Keyspace and builder, then use that builder in the `perGlobal` branch.

<details>

```
In `botplatform-service/middleware.go` around line 132: `botRateLimit` has caller and global Valkey branches, but only the caller key was moved into the registry. The global branch still increments `botrl:global`, while the registry only matches `botrl:caller:*`; consequently every scan attributes the global counter to `unclassified`. Add a fixed global-rate-limit Keyspace and builder, then use that builder in the `perGlobal` branch.

```

</details>

---

pkg/cachekeys/cachekeys_test.go
```diff
@@ -0,0 +1,244 @@
...
+			assert.NotEqual(t, Unclassified, Classify(tc.current))
+			assert.NotEqual(t, Unclassified, Classify(tc.legacy))
+		})
+	}
+}
+
+// TestKeyspaces_SampleMatchesExactlyOne is the invariant that keeps Classify
+// unambiguous as keyspaces are added: no sample key may match two keyspaces.
+// A new keyspace whose pattern overlaps an existing one fails here.
+func TestKeyspaces_SampleMatchesExactlyOne(t *testing.T) {
```
julianshen, line 152: _🧪 tests_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**Sample checks do not prove keyspace patterns cannot overlap.**

This test only checks each hand-picked sample against every matcher. Two variable patterns can intersect while both samples avoid the intersection: `a*yz` with sample `a1yz` and `ab*z` with sample `ab1z` both match `ab1yz`, yet this test passes. `Classify` would then silently choose whichever appears first. Add a pairwise pattern-intersection check for compatible prefixes and suffixes; retain the sample checks for builder wiring.

<details>

```
In `pkg/cachekeys/cachekeys_test.go` around line 150: This test only checks each hand-picked sample against every matcher. Two variable patterns can intersect while both samples avoid the intersection: `a*yz` with sample `a1yz` and `ab*z` with sample `ab1z` both match `ab1yz`, yet this test passes. `Classify` would then silently choose whichever appears first. Add a pairwise pattern-intersection check for compatible prefixes and suffixes; retain the sample checks for builder wiring.

```

</details>

---

julianshen (commented):
### tGDBot review of `d89edf4` — 11 findings in 30 files

10 🟠 warning · 1 🔵 suggestion
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/288#issuecomment-5581503116), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

pkg/cachekeys/cachekeys.go
```diff
@@ -0,0 +1,314 @@
...
+		Name: "roommeta", Prefix: "room:{", Suffix: "}:meta:v3", Variable: true,
+		Sample: "room:{r1}:meta:v3",
+	}
+	roomMetaLegacy = Keyspace{
+		Name: "roommeta.legacy", Prefix: "room:{", Suffix: "}:meta", Variable: true,
+		Sample: "room:{r1}:meta",
+	}
+	// roomSubs is deliberately un-hash-tagged, unlike roomMeta. The two are
+	// therefore in different cluster slots for the same room. Preserved as-is:
+	// this package documents the keyspace, it does not migrate it.
```
julianshen, line 98: _📝 documentation_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`tgd-review`_

**The room-subscription schema-version comment is stale.**

The registered current prefix is `room:v4:`, but this comment calls it the `v3` segment. That conflicts with the implementation and the exact-string test and could direct a future schema bump at the wrong generation.

<details>

> ‼️ **IMPORTANT**

```suggestion
	// The "v4" segment is pkg/roomsubcache's Member-wire-shape schema version:
```

</details>

<details>

```
In `pkg/cachekeys/cachekeys.go` around line 89: The registered current prefix is `room:v4:`, but this comment calls it the `v3` segment. That conflicts with the implementation and the exact-string test and could direct a future schema bump at the wrong generation.

```

</details>

---

PR #295, opened 2026-08-16
feat(loadgen): expand the continuous soak with room, member, presence, read-receipt, user and search lanes

flywindy:
## What this does

Expands the **existing** continuous soak (`phase: soak`) with eight new lanes. No new Deployment, no new phase — the lanes ride the same `seed → soak → stopped` lifecycle.

| Lane | Drives | Ledger |
|---|---|---|
| `member_mutation` | paired add/remove over a reusable candidate ring | `admission` + `room_state` |
| `room_mutation` | rename + mute/unmute over a fixed room pool | `admission` + `room_state` |
| `room_create` | low rate, capped by a per-run budget | `admission` + `room_state` |
| `read_receipt` | `messageRead` | `admission` + `room_state` |
| `room_read` | member list, rooms-info, subscription list, read receipts | none |
| `user_read` | 14 user-service reads | none |
| `search_read` | message and room search | none |
| `presence` | hello/ping/activity/bye + batch query | none (by design) |

## Why the async mutations are ledger-tracked

Every room/member/read-receipt/create mutation journals a durable WAL intent **before** it is sent, records its expected final state and a deadline, and is reconciled through the room-service RPC with a **MongoDB primary read** as the authoritative arbiter (the shared soak client is `secondaryPreferred`; a lagging secondary would report a completed write as missing and turn replication lag into a false data-loss claim during exactly the failures being measured).

Rules the implementation enforces:

- **Mutations are never resent.** They are not idempotent — a replayed remove drops a member the first attempt already removed. Ambiguity is settled by reconciliation, not retry.
- **`not_sent` only for proven local failures** — a body that could not be encoded, or a request never attempted. A NATS timeout, a dropped connection, *or a reply that failed to decode* all stay ambiguous (`unverified`).
- **`missing_after_deadline` requires admission to have recorded `good`.** An ambiguous request that left no state is `unverified`, never a data-loss claim.
- **Reconciliation resumes from the PVC/WAL after pod replacement**, including the pool reservations and the create budget.

## Three things deliberately kept *out* of the ledger

**Presence.** Its signals are core NATS fire-and-forget publishes that the client library buffers during an outage and flushes on reconnect. A successful publish proves nothing about delivery and a failed one proves nothing about loss. Only the batch query is evidence, and only outside the settle window and inside the connection TTL.

**Read-only lanes.** A read has no expected side effect to reconcile, so they carry latency/error/result metrics only.

**Read receipts are the exception that proves the rule** — `messageRead` is a synchronous write, and the cursor is monotonic, so verification compares two *server-written* timestamps and loadgen's clock never enters the verdict.

## The search-index observer is present but refused at startup

`search-sync-worker` Acks and drops a message whose payload it cannot decode or turn into a bulk action. That loss leaves its consumer at **zero pending, no Nak, no redelivery** — invisible from JetStream by construction. Search is the only one of MESSAGES-CANONICAL's downstreams with no end-to-end check, which is why the observer exists.

**It cannot be enabled yet, and startup rejects it.** The probe locates a message by full-text search, but every soak body is a single run of one character differing only in length and containing no whitespace — it analyzes to one huge token, so the query distinguishes nothing and would report *every* message lost. Payloads need a per-message searchable marker first. Rather than leave that as a note in the docs, `SOAK_SEARCH_OBSERVER_ENABLED=true` fails validation.

The machinery around it is complete and tested for when the marker lands: three-valued probe results, settle-boundary rescheduling, the capacity rule, and an additive observer contract (with the flag off, the contract is byte-identical to one built before the observer existed, so an existing `SOAK_LEDGER_EPOCH` stays valid).

## Ledger epoch

The run ID owns the seeded topology; the epoch owns the evidence journal. WAL path is `{SOAK_LEDGER_DIR}/{runId}.{epoch}.wal`.

A contract change makes an existing journal for the same run ID incompatible — a hard startup failure, never a silent downgrade. Bumping `SOAK_LEDGER_EPOCH` starts a fresh journal on an unchanged topology, so upgrading the image no longer forces a re-seed of ~10k rooms / ~310k subscriptions. Earlier journals stay on disk as evidence, are never replayed, and are counted in `loadgen_failure_abandoned_journals`.

## Defects found and fixed while building this

Worth reading, because several would have made round 1 report success from a run that tested nothing.

**The room lanes never ran — two independent reasons.** `LoadTopology` never populated `BorrowedUsers`, and the member lane draws every candidate from it, so a real `seed → soak` run built a pool with zero candidates and exited before sending anything. Separately, `soak_main.go` built its `soakWorkloadConfig` as an inline literal that never set six of the lane rates, and `lanes()` skips any lane with `rate <= 0` — so the configured target was reported and nothing was sent. Nothing covered the `seed → LoadTopology → pool` path that production actually takes; there is now an integration test that does.

**An executed mutation could settle as `not_sent`.** A body that failed to encode and a reply that failed to decode shared one error class, and the mutation lane treated that class as proof the request never left. They sit on opposite sides of the wire — a decode failure means the server answered. Split into `request_encode` / `response_decode`.

**`room_create` journaled after sending.** A timeout or lost reply could leave a created room with no operation, no reconciliation and no ownership record — unreachable by teardown, and a direct contradiction of the before-send rule. The room ID does not exist until the server answers, so the client-generated *name* is journaled instead; verification and ownership both resolve through it.

**Restart could score real loss as `good`.** The pending map was rebuilt empty, so a replacement process could issue a second mutation against a target whose first was still unresolved — and that second mutation can supply the very effect the first one lost. The create budget also reset per process, turning a per-run cap into an unbounded one across a crash loop.

**A restarted process expected the opposite mute state.** The subscription projection dropped `muted` and `lastSeenAt`, which the pool uses to seed its next toggle and its read-receipt baseline.

**Enabling the search observer would have saturated reconciliation ~6×.** A message inside the settle window returned a generic `unknown`, re-queued one retry interval later — ~20 s of once-a-second polling per message that cannot succeed, ~2000 reconcile slots/s against a 350/s budget. The probe now answers `too_early` distinctly and reschedules to the settle boundary.

**The integration harness never matched a request.** It subscribed with `subject.MemberAddPattern` — the `{account}` documentation template, a literal token on the wire.

## Observability

Consumer lag sampling went from 4 to **9 durables** — the room and member lanes publish into ROOMS and INBOX, and `room-worker`, `notification-worker-room-event-invalidate`, `message-sync`, `spotlight-sync` and `user-room-sync` all had traffic with no backlog evidence. Both the durable and the stream stay allowlisted, so a typo fails at construction rather than minting an unbounded Prometheus label. A consumer that is not deployed is reversible: `loadgen_consumer_up` goes to 0 and a bounded sample error is counted — it never invalidates the ledger.

**Traffic-validity gate corrected.** `loadgen_soak_dispatched_total` counts scheduler slots and is consumed even when a lane finds no target, so a lane idling on an exhausted pool looked fully loaded. Moving that counter would break the pacing identity it belongs to, so a separate `loadgen_soak_lane_attempts_total{lane,outcome}` distinguishes offered load from an idle lane, and the evidence contract now points at it.

New metric families are all bounded. **Room IDs, accounts and message IDs remain WAL and log content only, never Prometheus labels** — asserted by a test.

Quarantine overflow is reported as a **reversible** degradation, never a ledger invalidation: invalidation is a one-way, process-lifetime latch, so using it there would make every later fault window in a multi-day run inconclusive.

Six dashboard panels added.

## Coverage this brings

| Service | Handlers | Before | After |
|---|---:|---:|---:|
| room-service | 19 | 7 | 8 |
| user-service | 28 | 1 | 15 |
| history-service | 13 | 9 | 9 |
| search-service | 5 | 0 | 2 |
| user-presence-service | 6 | 5 | 5 |
| **Total** | **71** | **22 (31%)** | **39 (55%)** |

JetStream consumers with lag visibility: 4/9 → **9/9**.

## Config

Helm gains lane rates only — no new deployment, no new phase. New env vars: `SOAK_MEMBER_MUTATION_RATE`, `SOAK_ROOM_MUTATION_RATE`, `SOAK_ROOM_READ_RATE`, `SOAK_ROOM_CREATE_RATE`, `SOAK_ROOM_CREATE_BUDGET`, `SOAK_ROOM_CREATE_SIZE`, `SOAK_ROOM_RECONCILE_READ_SHARE`, `SOAK_MEMBER_QUARANTINE_MAX`, `SOAK_READ_RECEIPT_RATE`, `SOAK_USER_READ_RATE`, `SOAK_SEARCH_READ_RATE`, `SOAK_SEARCH_SETTLE`, `SOAK_SEARCH_OBSERVER_ENABLED`, `SOAK_LEDGER_EPOCH`, and the seven `SOAK_PRESENCE_*` knobs.

`docs/client-api.md` is untouched: this adds no client-facing handler and changes no `pkg/model` request/reply field.

## Verification

CI is green on `200e958`: `lint`, `test`, `test-integration (tools)`, `sast`.

Locally: `make lint` 0 issues, `go test -race ./tools/loadgen/...` pass, `go vet -tags integration` pass, `make coverage-loadgen-soak` **90.38%** (gate 90%), `make sast-gosec` clean.

**Not verified in this environment**, worth a reviewer's eye:

- `make validate-loadgen-k8s` — `helm` is not installed here and `get.helm.sh` is blocked (403) by the outbound proxy
- `make sast-vuln` — `vuln.go.dev` is blocked

## Known gap left open

The `search_index` observer is unusable until soak payloads carry a per-message searchable marker. That change touches the send path, the `SOAK_PAYLOAD_*` byte budgets, and the sonic wire-compat tests, so it is deliberately deferred rather than rushed — startup refuses the flag in the meantime.

## Before running

Use a **new `SOAK_RUN_ID`** for this image, and set `SOAK_LEDGER_EPOCH` deliberately.

## Summary by CodeRabbit

* **New Features**
  * Expanded soak testing with room/member mutations, room creation, reads, read receipts, user-service reads, search, and presence activity.
  * Added state reconciliation, lane-specific failure tracking, recovery, quarantine handling, consumer monitoring, and Prometheus metrics.
  * Added configuration and deployment controls for all workload areas.

* **Documentation**
  * Updated load-testing guides and plans with configuration, verification, reconciliation, metrics, and operational guidance.

* **Bug Fixes**
  * Improved request-encoding and response-decoding error classification.
  * Added validation for workload capacity, presence settings, ledger epochs, and currently unsupported search observation.


tools/loadgen/soak_wire_test.go
```diff
@@ -246,3 +247,287 @@ func TestSoakWireRequests_OmitOptionalFieldsLikeHistoryService(t *testing.T) {
...
+	require.NoError(t, err)
+
+	appsJSON, err := json.Marshal(usermodels.AppsListRequest{Limit: 20, Offset: 40})
+	require.NoError(t, err)
+	assert.JSONEq(t, string(appsJSON), string(soakJSON))
+
+	channelsJSON, err := json.Marshal(usermodels.GetChannelsRequest{Limit: 20, Offset: 40})
+	require.NoError(t, err)
+	assert.JSONEq(t, string(channelsJSON), string(soakJSON))
+}
```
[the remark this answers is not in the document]
flywindy: Fixed in 501e345, with one deviation worth naming.

The gap is real — the comment claimed three paged readers and only two were asserted. But the suggested shape (the same `Limit` and `Offset` across all three cases) would not hold: `model.ThreadListRequest` pages by `Cursor`, not `Offset`, so a `{Limit: 20, Offset: 40}` body would assert a field that type does not have and the case would fail. The lane never sends an offset there either — `ThreadList` sends `soakUserPageRequest{Limit: r.cfg.PageLimit}`.

So the thread-list case carries limit alone, which is the body the lane actually sends, and still pins the `limit` key against drift. Apps and channels keep both fields.

---

flywindy:
@coderabbitai review

The last incremental review was rate limited on `501e345`, which carries the table-driven `thread.list` paging fix. CI is green on that commit.

---

julianshen:
**22 findings · 21 unique issues · 17 inline comments posted** — 🔴 15 blocking · 🟠 7 warning

### 💬 Additional comments (4)


**`tools/loadgen/soak_roommember.go`**

_🏷️ atomicity_ | _🔴 Blocking_ | _`distributed-system`_

**Room ownership is recorded after the operation becomes terminal.**

A confirmed create is first finalized by `ledger.ObserveWithReason`; only afterward does `settle` call `finishCreate`, which resolves the room and invokes `AppendOwnedRooms`. If the process crashes after the terminal WAL record but before the ownership marker/chunk is written, or if MongoDB fails during `AppendOwnedRooms`, replay excludes the terminal operation and ownership is never retried. The room is then absent from teardown's ownership chunks and can leak permanently. Make ownership registration an idempotent prerequisite to the terminal `good` observation: keep or release the ledger claim on ownership failure, and finalize only after a durable marker/chunk record succeeds. The marker and chunk should also be transactional or keyed idempotently so a crash between them is recoverable.

---

**`tools/loadgen/soak_roommember.go`**

_🏷️ mongodb-atomicity_ | _🔴 Blocking_ | _`mongodb`_

**Ownership can be lost after the create operation is finalized.**

The room-create reconciliation records the terminal ledger observation before settle calls finishCreate. finishCreate then resolves the room and calls AppendOwnedRooms, but any lookup/append failure is only counted and is not retried because the ledger operation and pending entry are already gone. A crash in the same window has the same result. Additionally, AppendOwnedRooms updates rooms.soakRunId and inserts the ownership chunk as separate MongoDB writes, so a failure between them leaves a marked room that teardown still cannot reach. The concrete impact is a permanently leaked created room. Keep the ledger operation claimable until ownership succeeds, make the ownership append idempotent, and transactionally update the room marker and ownership record while propagating the Mongo session context to both operations; release the ledger claim for retry on failure.

---

**`tools/loadgen/soak_roommember.go`**

_🏷️ durability_ | _🔴 Blocking_ | _`silent-failure`_

**Room creation succeeds before teardown ownership is durable.**

Reconciliation records the create operation as terminal `good` before `settle` calls `finishCreate`. If `RoomIDByName` or `AppendOwnedRooms` then fails, `finishCreate` only logs/counts the problem and returns; the operation cannot be retried. The room may therefore be reported successful while absent from an ownership chunk, causing teardown to leak it. An unmarked room is also omitted by `CountCreatedRooms` after restart, allowing the run to exceed its creation budget. Make ownership registration a durable, retryable prerequisite for terminal success. `AppendOwnedRooms` must also surface a short `MatchedCount` rather than treating partial ownership as success.

---

**`tools/loadgen/soak_roommember.go`**

_🎯 correctness_ | _🔴 Blocking_ | _`silent-failure`_

**A failed cursor read silently weakens later receipt verification.**

After a read receipt is verified and finalized as `good`, `settle` performs another `SubscriptionLastSeen` call through `observedReadCursor`. Any error or missing result is silently converted to `time.Time{}`. `SettleRead` then marks the baseline unknown, so the next receipt operation accepts any existing cursor as `good`; an unchanged cursor from a later lost mark-read is therefore masked. Preserve the cursor obtained during verification and pass it into settlement, or surface the follow-up read failure and keep the operation retryable instead of discarding the baseline.

<details>
<summary>📒 Files reviewed (60)</summary>

* `docs/load-testing/loadgen-failure-observation.md`
* `docs/superpowers/plans/2026-08-16-room-member-soak-expansion.md`
* `tools/loadgen/README.md`
* `tools/loadgen/consumerlag.go`
* `tools/loadgen/consumerlag_test.go`
* `tools/loadgen/deploy/grafana/dashboards/loadtest.json`
* `tools/loadgen/deploy/k8s/templates/_helpers.tpl`
* `tools/loadgen/failure_dashboard_contract_test.go`
* `tools/loadgen/failure_evidence_integration_test.go`
* `tools/loadgen/failure_ledger_test.go`
* `tools/loadgen/failure_observation_runtime.go`
* `tools/loadgen/failure_observation_runtime_test.go`
* `tools/loadgen/failure_observer.go`
* `tools/loadgen/failure_wal_characterization_test.go`
* `tools/loadgen/metrics.go`
* `tools/loadgen/metrics_test.go`
* `tools/loadgen/soak_config_test.go`
* `tools/loadgen/soak_failure_test.go`
* `tools/loadgen/soak_main_test.go`
* `tools/loadgen/soak_presence.go`
* `tools/loadgen/soak_presence_integration_test.go`
* `tools/loadgen/soak_presence_test.go`
* `tools/loadgen/soak_roommember.go`
* `tools/loadgen/soak_roommember_integration_test.go`
* `tools/loadgen/soak_roomops.go`
* `tools/loadgen/soak_roomops_test.go`
* `tools/loadgen/soak_roomstate.go`
* `tools/loadgen/soak_roomstate_test.go`
* `tools/loadgen/soak_roomverify_test.go`
* `tools/loadgen/soak_rpc.go`
* `tools/loadgen/soak_rpc_test.go`
* `tools/loadgen/soak_search.go`
* `tools/loadgen/soak_search_test.go`
* `tools/loadgen/soak_send.go`
* `tools/loadgen/soak_send_test.go`
* `tools/loadgen/soak_verify_test.go`
* `tools/loadgen/soak_wire_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

tools/loadgen/soak_failure.go
```diff
@@ -79,12 +123,18 @@ func openSoakFailureLedger(
 		return nil, fmt.Errorf("soak configuration is required")
 	}
 	var journal failureJournal
-	contract := newFailureObserverContract(cfg.RecipientObserverEnabled)
+	contract := newFailureObserverContract(
+		cfg.RecipientObserverEnabled, cfg.SearchObserverEnabled,
```
julianshen, line 140: _🏷️ cassandra-migration_ | _🔴 Blocking_ | _`cassandra`_

**Pre-epoch Cassandra evidence journals are silently skipped**

The `cassandra_history` ledger previously opened `{runID}.wal`, but this now unconditionally opens `{runID}.{epoch}.wal`. On an upgrade using the default `v1`, unresolved accepted `message_send` operations in the existing WAL are never replayed and therefore never reconciled against Cassandra. The old file is also invisible to `recordAbandonedFailureJournals`, whose `runID + ".*.wal"` glob does not match `{runID}.wal`, so operators receive no abandoned-journal signal. Detect the legacy path when the epoch path is absent and either migrate/adopt it after contract validation or fail startup until the operator explicitly acknowledges the epoch transition; also include legacy journals in the abandoned count.

<details>

```
In `tools/loadgen/soak_failure.go` around line 127: The `cassandra_history` ledger previously opened `{runID}.wal`, but this now unconditionally opens `{runID}.{epoch}.wal`. On an upgrade using the default `v1`, unresolved accepted `message_send` operations in the existing WAL are never replayed and therefore never reconciled against Cassandra. The old file is also invisible to `recordAbandonedFailureJournals`, whose `runID + ".*.wal"` glob does not match `{runID}.wal`, so operators receive no abandoned-journal signal. Detect the legacy path when the epoch path is absent and either migrate/adopt it after contract validation or fail startup until the operator explicitly acknowledges the epoch transition; also include legacy journals in the abandoned count.

```

</details>

tools/loadgen/soak_main.go
```diff
@@ -743,6 +769,105 @@ func runSoakWorkload(
...
+	// created, or the new process races its predecessor's in-flight mutations
+	// and restarts the create cap from zero.
+	if recovered := ledger.ActiveOperations(); len(recovered) > 0 {
+		if inFlight := roomLanes.Rehydrate(recovered); inFlight > 0 {
+			// Counted for visibility only. An in-flight create whose room was
+			// made is already included in the created count below, and one
+			// whose room was not made occupies no allowance, so charging it
+			// here would double-spend the budget.
+			slog.Info("recovered in-flight soak room creates", "count", inFlight)
+		}
```
julianshen, line 839: _🏷️ resource-bounds_ | _🔴 Blocking_ | _`distributed-system`_

**Recovered creates do not consume the per-run room budget.**

On restart, `Rehydrate` counts active room-create operations, but the count is only logged. The new lane starts with the full in-memory budget and subtracts only `CountCreatedRooms`, whose query requires `soakRunId`; active creates are not marked until reconciliation finishes. Thus a pod can send many accepted creates, crash before reconciliation, restart with those operations active but a full budget, and send another full budget. Repeated failures make the documented per-run bound effectively per-process and allow unbounded MongoDB growth. Conservatively charge every recovered create intent against the restarted budget, or persist a durable per-run spent counter; under the current ordering, subtracting both the marked-room count and recovered active-create count is the minimal safe correction.

<details>

```
In `tools/loadgen/soak_main.go` around line 837: On restart, `Rehydrate` counts active room-create operations, but the count is only logged. The new lane starts with the full in-memory budget and subtracts only `CountCreatedRooms`, whose query requires `soakRunId`; active creates are not marked until reconciliation finishes. Thus a pod can send many accepted creates, crash before reconciliation, restart with those operations active but a full budget, and send another full budget. Repeated failures make the documented per-run bound effectively per-process and allow unbounded MongoDB growth. Conservatively charge every recovered create intent against the restarted budget, or persist a durable per-run spent counter; under the current ordering, subtracting both the marked-room count and recovered active-create count is the minimal safe correction.

```

</details>

julianshen (commented):
tGD inline review

tools/loadgen/soak_roomstate.go
```diff
@@ -0,0 +1,700 @@
...
+		}
+		room.memberLease = true
+		return true
+	case failureOperationRoomRename:
+		if room.renameLease {
+			return false
+		}
+		room.renameLease = true
+		return true
+	case failureOperationMuteToggle:
```
julianshen, line 368: _🏷️ recovery_ | _🔴 Blocking_ | _`distributed-system`_

**Rehydration leaves recovered member targets in the available ring.**

`ReserveInFlight` retakes only the room lease and deliberately leaves candidate slices unchanged. Consider an add that times out and then the process crashes. If the restarted topology currently shows the account absent, the pool places it in `available`; rehydration reserves the room but leaves that entry there. If reconciliation later ends `unverified` because both sources are unavailable, `SettleMember` marks the account quarantined but does not remove the stale `available` entry. `NextMemberIntent` consumes slices without checking the state map and sends another add even though the first request may still land, violating the no-resend/unknown-state contract and potentially creating duplicate or conflicting mutations. When reserving a recovered member operation, remove all occurrences of its account from `available` and `members` and transition it to the corresponding in-flight state before later settlement or quarantine.

<details>

```
In `tools/loadgen/soak_roomstate.go` around line 280: `ReserveInFlight` retakes only the room lease and deliberately leaves candidate slices unchanged. Consider an add that times out and then the process crashes. If the restarted topology currently shows the account absent, the pool places it in `available`; rehydration reserves the room but leaves that entry there. If reconciliation later ends `unverified` because both sources are unavailable, `SettleMember` marks the account quarantined but does not remove the stale `available` entry. `NextMemberIntent` consumes slices without checking the state map and sends another add even though the first request may still land, violating the no-resend/unknown-state contract and potentially creating duplicate or conflicting mutations. When reserving a recovered member operation, remove all occurrences of its account from `available` and `members` and transition it to the corresponding in-flight state before later settlement or quarantine.

```

</details>

tools/loadgen/soak_store.go
```diff
@@ -762,3 +810,312 @@ type soakOwnershipChunk struct {
...
+// run already spent rather than from a full allowance.
+//
+// It counts by name prefix rather than by ownership, because the seeded rooms
+// carry the same ownership marker and outnumber the budget many times over —
+// counting those would zero the allowance the moment any process restarted.
+func (s *mongoSoakStore) CountCreatedRooms(ctx context.Context, runID string) (int, error) {
+	if runID == "" {
+		return 0, fmt.Errorf("count created soak rooms requires a run ID")
+	}
+	total, err := s.primary("rooms").CountDocuments(ctx, bson.D{
```
julianshen, line 849: _🏷️ mongodb-bounded-growth_ | _🔴 Blocking_ | _`mongodb`_

**Crash-recovered rooms do not count toward the creation budget.**

CountCreatedRooms requires both the run-specific name prefix and soakRunId, but room-service creates the room without soakRunId; that marker is added later by AppendOwnedRooms. If the process stops after creation but before ownership reconciliation, Rehydrate deliberately does not charge the recovered create and this query also excludes the room. Repeated restarts can therefore exceed SOAK_ROOM_CREATE_BUDGET and grow the rooms collection without the intended per-run bound. Count every room carrying the generated run-specific creation prefix, including rooms not yet claimed.

<details>

```
In `tools/loadgen/soak_store.go` around line 843: CountCreatedRooms requires both the run-specific name prefix and soakRunId, but room-service creates the room without soakRunId; that marker is added later by AppendOwnedRooms. If the process stops after creation but before ownership reconciliation, Rehydrate deliberately does not charge the recovered create and this query also excludes the room. Repeated restarts can therefore exceed SOAK_ROOM_CREATE_BUDGET and grow the rooms collection without the intended per-run bound. Count every room carrying the generated run-specific creation prefix, including rooms not yet claimed.

```

</details>

tools/loadgen/soak_roomops.go
```diff
@@ -0,0 +1,195 @@
...
+			fmt.Errorf("mute toggle requires an account and room")
+	}
+	var reply soakMuteToggleReply
+	return m.call(ctx, soakRPCRequest{
+		Action:  soakRPCMuteToggle,
+		Subject: subject.MuteToggle(account, roomID, m.siteID),
+		Timeout: m.timeout, RetryMode: soakRetryNever,
+	}, &reply, roomID, func(outcome *soakRoomMutationOutcome) {
+		// Mute is applied inline, so a reply is proof of the stored state.
+		outcome.Accepted = true
```
julianshen: _🏷️ nats-request-reply_ | _🔴 Blocking_ | _`nats`_

**Malformed mutation replies produce definitive ledger verdicts.**

The mute-toggle and mark-read handlers set `Accepted` to true after any JSON-decodable reply, so `{}` records admission as good. Conversely, add/remove/rename/create leave `Accepted` false for an unrecognized or missing status, and `observeAdmission` treats that as an explicit rejection. In both cases the NATS request reached a responder and may have committed before returning a malformed or version-skewed reply. This can create false `missing_after_deadline` claims or release reservations while an applied mutation remains unreconciled. Validate each reply's required status and fields; classify missing, malformed, or unknown statuses as ambiguous admission, keep the operation active for reconciliation, and reserve `bad` for explicit bounded rejection envelopes.

<details>

```
In `tools/loadgen/soak_roomops.go` around line 123: The mute-toggle and mark-read handlers set `Accepted` to true after any JSON-decodable reply, so `{}` records admission as good. Conversely, add/remove/rename/create leave `Accepted` false for an unrecognized or missing status, and `observeAdmission` treats that as an explicit rejection. In both cases the NATS request reached a responder and may have committed before returning a malformed or version-skewed reply. This can create false `missing_after_deadline` claims or release reservations while an applied mutation remains unreconciled. Validate each reply's required status and fields; classify missing, malformed, or unknown statuses as ambiguous admission, keep the operation active for reconciliation, and reserve `bad` for explicit bounded rejection envelopes.

```

</details>

tools/loadgen/failure_ledger.go
```diff
@@ -750,15 +848,46 @@ func (l *failureLedger) Expire(now time.Time) (int, error) {
 }
 
 func (l *failureLedger) ClaimDue(now time.Time) (failureOperation, bool) {
+	return l.claimDue(now, nil)
```
julianshen, line 851: _🎯 concurrency_ | _🔴 Blocking_ | _`performance-and-concurrency`, `tgd-review` (2 rules)_

**The message reconciler can claim room-lane operations.**

ClaimDue now passes nil to claimDue, which scans every lane queue, while the existing message reconciler still uses ClaimDue rather than ClaimDueLanes. Under the default rates, message reconciliation has far more dispatch opportunities than room reconciliation, so it can repeatedly pop room/member operations and send them through a verifier that cannot observe room_state. Those claims are then delayed, retried, or expired unverified instead of reaching the correct verifier. Scope the message reconciler to soakFailureLaneMessageSend with ClaimDueLanes, as the room reconciler already does.

<details>

- **`tgd-review`**:853: ClaimDue now scans every per-lane queue when passed nil. The existing soakFailureReconciler.Try still uses ClaimDue and only understands message history, recipient, and search effects, while the room reconciler correctly uses ClaimDueLanes. Consequently the high-rate message read lane can repeatedly claim room/read-receipt operations, postpone them, or eventually record room_state as unverified without invoking soakRoomStateVerifier. Scope the message reconciler's claim to soakFailureLaneMessageSend.

</details>

<details>

```
In `tools/loadgen/failure_ledger.go` around line 851: ClaimDue now passes nil to claimDue, which scans every lane queue, while the existing message reconciler still uses ClaimDue rather than ClaimDueLanes. Under the default rates, message reconciliation has far more dispatch opportunities than room reconciliation, so it can repeatedly pop room/member operations and send them through a verifier that cannot observe room_state. Those claims are then delayed, retried, or expired unverified instead of reaching the correct verifier. Scope the message reconciler to soakFailureLaneMessageSend with ClaimDueLanes, as the room reconciler already does.

```

</details>

tools/loadgen/soak_roomstate.go
```diff
@@ -0,0 +1,700 @@
...
+		return nil, fmt.Errorf(
+			"soak room state pool requires a channel room with an owner and at least one candidate",
+		)
+	}
+	pool.refreshGaugesLocked()
+	return pool, nil
+}
+
+// ReserveInFlight re-takes the lease an operation held before the process was
+// replaced. Without it a replacement process can issue a second mutation
```
julianshen, line 316: _🎯 concurrency_ | _🔴 Blocking_ | _`performance-and-concurrency`_

**Recovered member reservations leave candidates in runnable queues.**

ReserveInFlight retakes only memberLease and deliberately leaves the account in available or members. After a restart, an unresolved add that is currently absent is therefore still in available. If reconciliation ends unverified, SettleMember marks it quarantined but does not remove that existing slice entry; once the room lease is released, NextMemberIntent ignores the state map and immediately issues another add against the unknown target. A recovered successful remove can similarly append a second available entry. This defeats restart-time single-flight protection and can produce duplicate conflicting mutations. When reserving a recovered member operation, remove the account from both runnable slices and place it in the matching in-flight state before taking the lease.

<details>

```
In `tools/loadgen/soak_roomstate.go` around line 250: ReserveInFlight retakes only memberLease and deliberately leaves the account in available or members. After a restart, an unresolved add that is currently absent is therefore still in available. If reconciliation ends unverified, SettleMember marks it quarantined but does not remove that existing slice entry; once the room lease is released, NextMemberIntent ignores the state map and immediately issues another add against the unknown target. A recovered successful remove can similarly append a second available entry. This defeats restart-time single-flight protection and can produce duplicate conflicting mutations. When reserving a recovered member operation, remove the account from both runnable slices and place it in the matching in-flight state before taking the lease.

```

</details>

tools/loadgen/soak_workload.go
```diff
@@ -11,28 +11,74 @@ import (
...
+		SendRate:           cfg.SendRate,
+		ReadRate:           cfg.ReadRate,
+		MutationRate:       cfg.MutationRate,
+		ReactionRate:       cfg.ReactionRate,
+		PinnedListRate:     cfg.PinnedListRate,
+		VerifyRate:         cfg.VerifyRate,
+		MemberMutationRate: cfg.MemberMutationRate,
+		RoomMutationRate:   cfg.RoomMutationRate,
+		RoomReadRate:       cfg.RoomReadRate,
+		UserReadRate:       cfg.UserReadRate,
```
julianshen, line 75: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**Search reads are never scheduled**

soakWorkloadConfigFrom copies every new lane rate except SearchReadRate. Production therefore constructs the workload with a zero search-read rate, and lanes() skips that lane despite SOAK_SEARCH_READ_RATE defaulting to 5. The new enables-every-lane test should also expose this omission.

<details>

```
In `tools/loadgen/soak_workload.go` around line 75: soakWorkloadConfigFrom copies every new lane rate except SearchReadRate. Production therefore constructs the workload with a zero search-read rate, and lanes() skips that lane despite SOAK_SEARCH_READ_RATE defaulting to 5. The new enables-every-lane test should also expose this omission.

```

</details>

tools/loadgen/soak_roommember.go
```diff
@@ -0,0 +1,791 @@
...
+	observation failureObservation,
+	reason failureReason,
+	now time.Time,
+) error {
+	finalized, err := l.ledger.ObserveWithReason(
+		operationID, failureObserverRoomState, observation, reason, now,
+	)
+	if err != nil {
+		_ = l.ledger.ReleaseClaim(operationID, now.Add(l.cfg.RetryInterval))
+		return fmt.Errorf("record soak room state observation: %w", err)
```
julianshen, line 620: _🏷️ durability_ | _🔴 Blocking_ | _`tgd-review`_

**Room ownership is persisted after the ledger operation closes**

The room-state observation is committed and the operation finalized before settle calls finishCreate and AppendOwnedRooms. A crash in that gap, or any ownership append error, leaves no active WAL operation to retry. The created room then lacks both the ownership marker and chunk, is omitted by CountCreatedRooms, and cannot be removed by teardown. Make ownership registration a durable prerequisite to finalizing a successful create, or persist a separate retryable ownership task.

<details>

```
In `tools/loadgen/soak_roommember.go` around line 590: The room-state observation is committed and the operation finalized before settle calls finishCreate and AppendOwnedRooms. A crash in that gap, or any ownership append error, leaves no active WAL operation to retry. The created room then lacks both the ownership marker and chunk, is omitted by CountCreatedRooms, and cannot be removed by teardown. Make ownership registration a durable prerequisite to finalizing a successful create, or persist a separate retryable ownership task.

```

</details>

tools/loadgen/soak_roommember.go
```diff
@@ -0,0 +1,791 @@
...
+			l.mu.Lock()
+			l.pending[operation.ID] = soakRoomPending{
+				kind:     failureOperationRoomCreate,
+				roomName: operation.Targets["roomName"],
+			}
+			l.mu.Unlock()
+			continue
+		}
+		roomID := operation.Targets["roomId"]
+		account := operation.Targets["account"]
```
julianshen, line 156: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**Recovered member targets can bypass quarantine**

Rehydrate calls ReserveInFlight, but that method only takes the room lease and deliberately leaves the target in the rebuilt available/member slices. If the recovered operation later resolves unverified or missing, SettleMember marks the map entry quarantined without removing the stale slice entry; NextMemberIntent reads those slices and can immediately mutate the unknown target again. Retaking a recovered member reservation must remove the account from both candidate slices and establish the corresponding in-flight state before settlement.

<details>

```
In `tools/loadgen/soak_roommember.go` around line 156: Rehydrate calls ReserveInFlight, but that method only takes the room lease and deliberately leaves the target in the rebuilt available/member slices. If the recovered operation later resolves unverified or missing, SettleMember marks the map entry quarantined without removing the stale slice entry; NextMemberIntent reads those slices and can immediately mutate the unknown target again. Retaking a recovered member reservation must remove the account from both candidate slices and establish the corresponding in-flight state before settlement.

```

</details>

tools/loadgen/soak_presence.go
```diff
@@ -0,0 +1,457 @@
...
+	matched, mismatched, unknown := 0, 0, 0
+	for account, expected := range expectations {
+		status, ok := reported[account]
+		switch {
+		case !ok:
+			unknown++
+		case status == expected:
+			matched++
+		default:
+			mismatched++
```
julianshen, line 222: _🏷️ nats-delivery_ | _🟠 Warning_ | _`nats`_

**Lost presence publishes are counted as unknown rather than mismatches.**

Core NATS can drop a fire-and-forget hello while presence-service has no subscriber, yet `PublishMsg` still succeeds and the lane records the connection as online. After service recovery and the settle window, a batch query can omit that account; this branch records `unknown`, so the exact delivery loss the lane is meant to observe never increments `mismatch`. The integration spy also omits accounts for which no signal arrived, making this path concrete. Treat an omitted live expectation as a mismatch before TTL, or interpret omission according to the query API as offline and compare that value; retain `unknown` only for an incomplete/malformed reply whose completeness cannot be established.

<details>

```
In `tools/loadgen/soak_presence.go` around line 222: Core NATS can drop a fire-and-forget hello while presence-service has no subscriber, yet `PublishMsg` still succeeds and the lane records the connection as online. After service recovery and the settle window, a batch query can omit that account; this branch records `unknown`, so the exact delivery loss the lane is meant to observe never increments `mismatch`. The integration spy also omits accounts for which no signal arrived, making this path concrete. Treat an omitted live expectation as a mismatch before TTL, or interpret omission according to the query API as offline and compare that value; retain `unknown` only for an incomplete/malformed reply whose completeness cannot be established.

```

</details>

tools/loadgen/soak_config.go
```diff
@@ -264,6 +299,79 @@ func validateSoakConfig(cfg *soakConfig, cassandraKeyspace string) error {
 	return nil
 }
 
+func validateSoakRoomLaneConfig(cfg *soakConfig) error {
+	if !isFinite(cfg.RoomReconcileReadShare) ||
+		cfg.RoomReconcileReadShare <= 0 || cfg.RoomReconcileReadShare > 1 {
+		return fmt.Errorf(
+			"SOAK_ROOM_RECONCILE_READ_SHARE must be greater than zero and at most 1",
```
julianshen, line 310: _⚡ memory-exhaustion_ | _🟠 Warning_ | _`oom`_

**Room creation budget has no effective upper bound.**

`SOAK_ROOM_CREATE_BUDGET` accepts any non-negative `int`. Each successful create reaches `RegisterCreatedRoom`, which permanently appends the room ID to `soakRoomReader.created` and adds it to `accounts`; neither collection evicts entries during a continuous run. A large allowed budget combined with a correspondingly high create/read rate therefore grows retained heap linearly until container OOM or severe GC thrashing. Add a practical hard maximum here and in the Helm schema. If larger database creation budgets are required, keep only a bounded sample of created rooms in the in-memory read target set.

<details>

```
In `tools/loadgen/soak_config.go` around line 306: `SOAK_ROOM_CREATE_BUDGET` accepts any non-negative `int`. Each successful create reaches `RegisterCreatedRoom`, which permanently appends the room ID to `soakRoomReader.created` and adds it to `accounts`; neither collection evicts entries during a continuous run. A large allowed budget combined with a correspondingly high create/read rate therefore grows retained heap linearly until container OOM or severe GC thrashing. Add a practical hard maximum here and in the Helm schema. If larger database creation budgets are required, keep only a bounded sample of created rooms in the in-memory read target set.

```

</details>

tools/loadgen/soak_roomverify.go
```diff
@@ -0,0 +1,462 @@
...
+	case member == expectMember:
+		authoritative = soakRoomStateMatched
+		v.setHealth(true, "member_read")
+	default:
+		authoritative = soakRoomStateAbsent
+		v.setHealth(true, "member_read")
+	}
+	v.countSource(soakRoomStateSourceStore, authoritative)
+
+	result := resolveSoakRoomState(rpc, authoritative)
```
julianshen, line 173: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Successful RPC verification still blocks on MongoDB.**

verifyMember, verifyRename, verifyMute, and verifyRead all query the primary even after room-service has already returned the expected state. The documented contract says that positive RPC evidence settles the operation alone. During primary latency or an outage, each otherwise-complete reconciliation therefore occupies a room-read and global in-flight slot until the Mongo timeout; concurrent slots can accumulate and starve unrelated workload actions. Healthy runs also add an avoidable primary query per mutation. Return immediately when the RPC result is matched, and consult Mongo only for unknown, absent, or contradictory RPC results.

<details>

```
In `tools/loadgen/soak_roomverify.go` around line 155: verifyMember, verifyRename, verifyMute, and verifyRead all query the primary even after room-service has already returned the expected state. The documented contract says that positive RPC evidence settles the operation alone. During primary latency or an outage, each otherwise-complete reconciliation therefore occupies a room-read and global in-flight slot until the Mongo timeout; concurrent slots can accumulate and starve unrelated workload actions. Healthy runs also add an avoidable primary query per mutation. Return immediately when the RPC result is matched, and consult Mongo only for unknown, absent, or contradictory RPC results.

```

</details>

tools/loadgen/soak_roomstate.go
```diff
@@ -0,0 +1,700 @@
...
+}
+
+func (p *soakRoomStatePool) refreshGaugesLocked() {
+	if p.metrics == nil {
+		return
+	}
+	counts := make(map[soakMemberCandidateState]int, len(soakMemberCandidateStates))
+	degraded := len(p.quarantine)+p.inProbe >= p.quarantineMax
+	for _, room := range p.rooms {
+		for _, state := range room.states {
```
julianshen: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Candidate metrics rescan the entire pool while holding its mutex.**

refreshGaugesLocked iterates every candidate in every room and is called on each member dispatch and settlement, as well as mute and quarantine transitions. With the default 10,000 rooms and a 64-candidate ring, one call scans roughly 640,000 map entries while holding p.mu; several calls per second serialize member, mute, read-receipt, create, and probe target selection. Higher configured rates amplify this into lane saturation. Maintain per-state and degraded-room counters incrementally during transitions, updating the gauges in O(1), or snapshot counters and publish metrics outside the pool lock.

<details>

```
In `tools/loadgen/soak_roomstate.go` around line 625: refreshGaugesLocked iterates every candidate in every room and is called on each member dispatch and settlement, as well as mute and quarantine transitions. With the default 10,000 rooms and a 64-candidate ring, one call scans roughly 640,000 map entries while holding p.mu; several calls per second serialize member, mute, read-receipt, create, and probe target selection. Higher configured rates amplify this into lane saturation. Maintain per-state and degraded-room counters incrementally during transitions, updating the gauges in O(1), or snapshot counters and publish metrics outside the pool lock.

```

</details>

tools/loadgen/soak_presence.go
```diff
@@ -0,0 +1,457 @@
...
+		}
```
julianshen, line 223: _🎯 correctness_ | _🟠 Warning_ | _`silent-failure`_

**Missing online accounts are classified as unknown, not mismatched.**

Once settle has elapsed and the connection remains inside its TTL, a successful batch response that omits an account expected to be online or away is the concrete failure this observer is meant to detect: the announced presence was not applied or was dropped early. The `!ok` branch instead increments `unknown`, so that failure never reaches the mismatch metric. Classify omitted online/away expectations as mismatches; omission may count as a match when the expected state is offline.

<details>

```
In `tools/loadgen/soak_presence.go` around line 223: Once settle has elapsed and the connection remains inside its TTL, a successful batch response that omits an account expected to be online or away is the concrete failure this observer is meant to detect: the announced presence was not applied or was dropped early. The `!ok` branch instead increments `unknown`, so that failure never reaches the mismatch metric. Classify omitted online/away expectations as mismatches; omission may count as a match when the expected state is offline.

```

</details>

tools/loadgen/soak_failure.go
```diff
@@ -70,6 +76,44 @@ func runSoakFailureExpiry(
...
+func recordAbandonedFailureJournals(metrics *Metrics, dir, runID, epoch string) {
+	if metrics == nil {
+		return
+	}
+	active := failureWALPath(dir, runID, epoch)
+	matches, err := filepath.Glob(filepath.Join(dir, runID+".*.wal"))
+	if err != nil {
+		slog.Error("scan retained failure journals", "runId", runID, "error", err)
+		return
+	}
```
julianshen, line 109: _🏷️ durability_ | _🟠 Warning_ | _`tgd-review`_

**Pre-epoch WALs are silently abandoned without being counted**

The previous implementation wrote {runID}.wal, but the abandoned-journal scan only matches {runID}.*.wal. After upgrading, openSoakFailureLedger opens {runID}.v1.wal and neither replays nor reports the old journal, so unresolved evidence is silently hidden and loadgen_failure_abandoned_journals remains zero. Include the exact legacy {runID}.wal path in the retained-journal scan and surface it as abandoned or require an explicit migration decision.

<details>

```
In `tools/loadgen/soak_failure.go` around line 103: The previous implementation wrote {runID}.wal, but the abandoned-journal scan only matches {runID}.*.wal. After upgrading, openSoakFailureLedger opens {runID}.v1.wal and neither replays nor reports the old journal, so unresolved evidence is silently hidden and loadgen_failure_abandoned_journals remains zero. Include the exact legacy {runID}.wal path in the retained-journal scan and surface it as abandoned or require an explicit migration decision.

```

</details>

tools/loadgen/soak_roommember.go
```diff
@@ -0,0 +1,791 @@
...
+			"runId", l.cfg.RunID, "roomName", roomName, "error", err)
+		l.countUntracked("ownership")
+		return
+	}
+	if err := l.store.AppendOwnedRooms(ctx, l.cfg.RunID, []string{roomID}); err != nil {
+		l.countUntracked("ownership")
+		return
+	}
+	if l.reader != nil {
+		if requester, ok := l.pool.AnyOwner(); ok {
```
julianshen: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Created rooms are registered with an unrelated account**

finishCreate registers the room reader using pool.AnyOwner rather than the requester that created the room. With multiple seeded rooms this is usually an account that is not a member of the created room, so member-list and read-receipt traffic against created rooms will measure authorization failures instead of valid reads. Preserve the create requester in soakRoomPending, restore it from the WAL during Rehydrate, and pass that account to RegisterCreatedRoom.

<details>

```
In `tools/loadgen/soak_roommember.go` around line 656: finishCreate registers the room reader using pool.AnyOwner rather than the requester that created the room. With multiple seeded rooms this is usually an account that is not a member of the created room, so member-list and read-receipt traffic against created rooms will measure authorization failures instead of valid reads. Preserve the create requester in soakRoomPending, restore it from the WAL during Rehydrate, and pass that account to RegisterCreatedRoom.

```

</details>

tools/loadgen/soak_workload.go
flywindy: Not reproducible — `SearchReadRate` is wired.

`soakWorkloadConfigFrom` copies it at `soak_workload.go:76`:

```go
RoomReadRate:   cfg.RoomReadRate,
UserReadRate:   cfg.UserReadRate,   // line 75
SearchReadRate: cfg.SearchReadRate, // line 76
RoomCreateRate: cfg.RoomCreateRate,
```

The finding anchors on line 75, which is `UserReadRate`. `lanes()` therefore sees the configured `SOAK_SEARCH_READ_RATE` and schedules the lane; `TestSoakWorkload_EnablesEveryLane` covers it, and `allSoakWorkloadActions` fills the `SearchRead` slot (added in 4804743 after an earlier review round flagged that gap).

No change made.

---

tools/loadgen/soak_presence.go
flywindy: The premise doesn't hold against the service, so I've left this as `unknown`.

`QueryBatch` — the subject this lane calls — emits one `PresenceState` per **requested** account, substituting offline for anything it cannot resolve (`user-presence-service/handler.go:224-236`):

```go
states := make([]model.PresenceState, 0, len(req.Accounts))
for _, account := range req.Accounts {
    status, ok := statusByAccount[account]
    if !ok || status == model.StatusNone {
        status = model.StatusOffline
    }
    ...
}
```

`QueryBatchPeer` does the same at `:144-153`, and the doc comment states it outright: "a peer failure or an unknown account degrades to StatusOffline".

So the failure this finding describes — a fire-and-forget hello dropped while presence-service had no subscriber — does **not** arrive as an omission. It arrives as `status == offline` against an `online` expectation, which the `default:` branch already counts as `mismatched`. Reclassifying omissions as mismatches would not add that detection; it would only mislabel replies that violated the contract, which is precisely what `unknown` is reserved for.

One part of the report is worth acting on separately: if the integration spy omits accounts for which no signal arrived, the spy is unfaithful to the real handler. That's a test-double fidelity issue rather than a production classification one, and I'd rather fix it as such than change the production semantics to match an inaccurate double.

No change made.

---

tools/loadgen/failure_ledger.go
flywindy: Confirmed and fixed in 1792eef.

Reproduced first: a room-lane operation started in the ledger was claimed by `soakFailureReconciler.Try`. `claimDue(now, nil)` enumerates every lane queue, and a room operation has no `history` observation, so it fell through to `r.verifier.Verify` — the Cassandra message lookup — and past the deadline `r.observe` recorded the verdict against `failureObserverHistory`, an observer the operation never declared, while `room_state` stayed unresolved.

`Try` now claims `[]string{soakFailureLaneMessageSend}` only. `TestSoakFailureReconciler_LeavesRoomLaneOperationsAlone` fails on the old code and passes on the new.

---

tools/loadgen/soak_roomstate.go
flywindy: Confirmed and fixed in 1792eef.

`ReserveInFlight` now removes the account from both `available` and `members` and sets the matching in-flight state before taking the lease.

Worth recording how bad it was: the reproduction showed the good path was broken too, not just the quarantine path. `SettleMember` **appends** on every terminal result, so a rehydrated operation that resolved cleanly duplicated the candidate rather than returning it. In a 64-dispatch run the same account came back **32 times** — the ring was being flooded, not just leaked from.

Two tests, both failing on the old code:
- `TestSoakRoomStatePool_ReserveInFlightRemovesTheCandidateFromTheRing` — a quarantined pair is never dispatched again
- `TestSoakRoomStatePool_ReserveInFlightDoesNotDuplicateTheCandidate` — exactly one entry across the room's slices after settle

---

tools/loadgen/soak_roomstate.go
flywindy: Same defect as the `soak_roomstate.go:280` thread — confirmed and fixed in 1792eef. `ReserveInFlight` now clears the account from both runnable slices and sets the in-flight state before taking the lease.

---

tools/loadgen/soak_roommember.go
flywindy: Same defect as the two `soak_roomstate.go` threads — confirmed and fixed in 1792eef.

---

tools/loadgen/soak_store.go
flywindy: Confirmed and fixed in 1792eef. `CountCreatedRooms` now matches on the name prefix alone.

The `soakRunId` clause was excluding exactly the rooms that needed counting, as described. The prefix `soak-{runId}-created-` already embeds the run ID, so it scopes the query on its own and the extra filter bought nothing.

This also settles the `soak_main.go:837` thread: the comment there claims a recovered in-flight create is "already included in the created count below", which was false while this filter stood and is true now. The comment has been rewritten to name that dependency so the filter is not reinstated.

`TestCountCreatedRooms_ExcludesTheSeededPopulation` gained a case for a room carrying the prefix with no ownership marker.

---

tools/loadgen/soak_main.go
flywindy: Confirmed, and fixed at the other end in 1792eef.

The root cause was `CountCreatedRooms` requiring `soakRunId`, which `AppendOwnedRooms` only writes afterwards — so an accepted-but-unclaimed create really was invisible to the count, exactly as described. That filter is gone; the name prefix already embeds the run ID.

With that, charging `Rehydrate`'s in-flight creates a second time would double-spend: a create whose room was made is now counted below whether or not ownership landed, and one whose room was never made occupies no allowance. The count stays the single source of truth, and the comment here has been rewritten to name that dependency so the filter is not reinstated.

---

tools/loadgen/soak_roomops.go
flywindy: Confirmed and fixed in 1792eef, both halves.

`observeAdmission` gained a `case err == nil:` branch that records `unverified` and keeps the operation reconcilable. Only a bounded error envelope (`err != nil`, non-transient class) reaches the rejection path now, so a responder that answered with a status this run does not recognise no longer closes the operation and hands the candidate back as untouched.

Mute and mark-read now check their reply instead of setting `Accepted = true` unconditionally. That change immediately caught a wire mismatch in our own tests: the fixtures had mark-read replying `{"status":"ok"}`, but `room-service` returns `{"status":"accepted"}` (`handler.go:1394`). The unconditional accept had been hiding it. Fixtures corrected, and `TestSoakRoomStatusConstants_MatchRoomServiceReplies` now pins both strings against `model.StatusReply` and `model.MuteToggleResponse`.

`TestSoakRoomLanes_UnrecognisedReplyStaysAmbiguous` covers the classification.

---

tools/loadgen/soak_roommember.go
flywindy: Confirmed and fixed in 1792eef.

Ownership is now a prerequisite of finalizing the create rather than a step after it. `Reconcile` calls `claimCreatedRoom` before `observe` on the matched path; a failure before the deadline releases the claim so the operation stays open and retries, and only past the deadline does it log and count `untracked`. `settle` no longer does ownership at all.

`TestSoakRoomLanes_RoomCreateHoldsTheOperationUntilOwnershipLands` drives it: with `AppendOwnedRooms` failing the operation is still active and nothing is scored `good`; once the store recovers, the same operation completes and the room is claimed.

`TestSoakRoomLanes_OwnershipFailureIsCountedAsUntracked` was updated to advance past the deadline, since before it the correct behaviour is now to retry rather than give up.

---

tools/loadgen/soak_roommember.go
flywindy: Confirmed and fixed in 1792eef. `claimCreatedRoom` now registers the room under `operation.Attributes[requester]` — the account that actually created it — and `pool.AnyOwner()` has been deleted, since that was its only caller.

Taking the requester from the journaled attributes rather than `soakRoomPending` also makes it survive a replacement process, which was the other half of the suggestion.

`TestSoakRoomLanes_RoomCreateRegistersTheRequesterAsReader` asserts the reader resolves the created room to the operation's requester.

---

tools/loadgen/soak_roomstate.go
flywindy: Confirmed and fixed in 1792eef, with a follow-up in 097cab4.

The gauges are now maintained incrementally: `stateCounts` and `emptyRooms` on the pool, adjusted by `setCandidateStateLocked` / `trackRoomFillLocked` at each transition, so `refreshGaugesLocked` is O(states) instead of O(rooms × candidates) and the pool mutex is no longer held across the scan.

Incremental counters only hold if every transition adjusts them, so `TestSoakRoomStatePool_CandidateGaugesTrackEveryTransition` walks startup → dispatch → settle → quarantine → restart reservation and compares the published gauges against the pool's own `room.states` — the totals a recomputing implementation would have produced.

That test did not cover quarantine overflow, which deletes a state rather than moving between two, and the direct map delete left the count behind. Caught in review of 1792eef and fixed in 097cab4 with `dropCandidateStateLocked` plus a second test that drives a pool past its quarantine capacity.

---

tools/loadgen/soak_roomverify.go
flywindy: Agreed and fixed in 1792eef. This was the documented contract disagreeing with the code — the doc already said a positive RPC settles alone, and all four verify paths queried the primary anyway.

`settledByRPC` returns immediately when the RPC result is `matched`; unknown, absent and contradictory results still go to the store, so absence claims are unchanged.

One thing I checked before doing it: skipping the store read also skips `setHealth`. That is safe because the observer health is edge-triggered (`Set` records only on transitions) and the room-state health is not consumed by any `HealthyThroughout` gate — only the recipient observer uses that. The store is still consulted on every path where its reachability actually matters.

`TestSoakRoomStateVerifier_MemberAddIsSettledByThePositiveRPC` asserts the primary is not read at all on that path.

---

tools/loadgen/soak_config.go
flywindy: Fixed in 1792eef. `SOAK_ROOM_CREATE_BUDGET` is now validated against `soakRoomCreateBudgetMax` (100000) rather than only `>= 0`, alongside the existing bounds on `SOAK_ROOM_CREATE_SIZE` and `SOAK_MEMBER_QUARANTINE_MAX`.

Kept the bound in Go rather than the Helm schema: the chart passes these as env-var strings, so the numeric range is not expressible there and the validator is the single place it can be enforced.

The retained read-target set is the reason, as described — `RegisterCreatedRoom` never evicts.

---

tools/loadgen/soak_failure.go
flywindy: Fixed in 1792eef. `recordAbandonedFailureJournals` now stats the exact legacy path in addition to the epoch glob, so `{runId}.wal` is counted and surfaced in the warning like any other retained journal.

Your framing is the right one: bumping the epoch while keeping the run ID is the documented upgrade path — that is the whole point of splitting topology identity from evidence identity — so the legacy file is exactly what an in-place upgrade inherits, not a corner case.

`TestFailureObservationRuntime_PreEpochJournalIsCountedAsAbandoned` writes a `{runId}.wal` before opening the ledger and asserts both the count and that the file is retained.

The migrate-or-fail half is discussed on the sibling `soak_failure.go:140` thread, which I have left open.

---

tools/loadgen/soak_presence.go
flywindy: Same finding as the `soak_presence.go:223` thread, and I've left the classification as `unknown` for the reason set out there: `QueryBatch` returns an entry for **every** requested account, defaulting to `StatusOffline` (`user-presence-service/handler.go:224-236`), so a dropped hello arrives as `offline` against an `online` expectation and the `default:` branch already counts it as `mismatched`. An omission means the reply broke its own contract, which is what `unknown` is for.

The one part I do take from this: you note the integration spy omits accounts for which no signal arrived. If so the spy is unfaithful to the real handler, and that is worth fixing as a test-double issue rather than by bending production semantics to match it.

No change made.

---

tools/loadgen/soak_failure.go
flywindy: Partially addressed — flagging that rather than closing it, because I did one of the two things you asked for.

**Done (1792eef):** the legacy `{runId}.wal` is now counted as an abandoned journal and surfaced in the warning, so an in-place upgrade no longer orphans it silently. `TestFailureObservationRuntime_PreEpochJournalIsCountedAsAbandoned` covers it.

**Not done:** migrate/adopt the legacy journal, or fail startup until the operator acknowledges the transition.

My reasoning for stopping there, which is a judgement call worth your review:

- Adopting it means replaying `message_send` records written under observer contract v1 into a v2 ledger. The epoch exists precisely because that contract changed; the PR already treats a contract mismatch as a hard startup failure rather than a silent downgrade, and replaying incompatible records would be the silent downgrade in another form.
- Failing startup would block the upgrade path the epoch was introduced to enable, over a file the design already classifies as retained-not-replayed. Every earlier-epoch journal gets exactly this treatment; the legacy file is now consistent with them rather than special.

So the position is: the legacy journal is treated like any other abandoned one, and its unresolved Cassandra operations are genuinely lost as evidence for that run. If you would rather that be a hard stop — or want the adoption path built — say so and I will implement it; it is a deliberate default, not an oversight.

---

flywindy:
@coderabbitai review

The incremental review was rate limited on `097cab4`, which carries the fix for the quarantine-overflow counter drift you found in `1792eef` (`dropCandidateStateLocked`, plus a regression test that drives a pool past its quarantine capacity). CI is green on that commit.

---

yenta (approved):

PR #297, opened 2026-08-17
Substitute @mentions with display names in push notification bodies

evelynchang5525:
## Summary
Adds mention name resolution to push notifications so lock screens display human-readable names instead of account handles. Messages like "hey @bob" become "hey Bob Chen" in the push body, with special handling for @all/@here literals and graceful degradation on lookup failures.

### Core mention substitution logic (`pkg/mention/mention.go`)
- Added `LookupAccountsFromParsed()` to extract accounts needing user lookup (excluding @all/@here literals), taking an already-parsed result so the hot path runs the mention regexp once rather than twice
- Added `ReplaceAccounts()` to rewrite @mention tokens: @all→"All", @here→"here", known accounts→display names, unknown accounts→kept verbatim with @marker
- Defined `literalMentions` map for the two non-account tokens that resolve from constants, never hitting the users collection

### User mention name resolver (`notification-worker/mentionnames.go`)
- Implemented `userMentionNames` struct that queries the users collection for display names
- Composes `EngName` and `ChineseName` using the same rule as message-gatekeeper's `UserDisplayName`
- Omits users with no names so unresolved accounts keep their @marker instead of degrading to bare text
- Bounds lookups with a 2-second timeout so slow user collection cannot hold up fan-out
- Returns partial results on error so a failed lookup substitutes what it can — a contract `userstore.Cache` actually exhibits, documented on the `UserAccountFinder` interface

### Handler integration (`notification-worker/handler.go`)
- Added `MentionNames` field to `HandlerDeps` (optional; nil → only @all/@here are substituted)
- Added `resolveBody()` method that:
  - Skips lookup if no mentions or no survivors (after DND/muting filter)
  - Calls resolver only for accounts needing lookup (not @all/@here)
  - Caps resolution at 50 distinct accounts per message — message content is user-controlled and the `$in` was otherwise unbounded
  - Fails open: resolver errors substitute whatever names came back, leaves rest as raw @tokens
  - Applies substitution to every batch so all recipients see the same body

### Configuration (`notification-worker/main.go`, `config_test.go`)
- Added `UserCacheSize`, `UserCacheTTL`, `MentionNamesTimeout` config fields with sensible defaults
- Added `MENTION_NAMES_ENABLED` (default `true`) as a kill switch for a sick users collection, matching `USER_SETTINGS_ENABLED` / `BADGE_COUNT_RPC_ENABLED` / `PRESENCE_RPC_ENABLED`
- A bad user-cache config degrades to no substitution rather than exiting — losing display names is not worth refusing to deliver notifications

### Observability
- Added `notification_worker_mention_resolution_total{result="resolved|unresolved|failed"}`, counting one point per looked-up token. Without it a users-collection outage left the existing outcome metric fully green while every push body shipped raw @tokens. A failed lookup's remainder is attributed to `failed` rather than `unresolved`, so an outage is distinguishable from senders mentioning unknown accounts.
- The fail-open warn uses `slog.WarnContext` so it carries traceId/spanId and the handler's identity baggage, and logs the resolved count alongside the requested one

### Testing
- Added unit tests in `handler_test.go`: body replacement, literal mentions, unresolved preservation, resolver call deduplication, graceful degradation, batch consistency, the lookup cap, partial resolution on error, and mentions coexisting with per-recipient badge counts
- Added resolver unit tests in `mentionnames_test.go` covering name composition, edge cases, error handling, the timeout default, and the timeout actually bounding the call
- Added mention package tests in `pkg/mention/mention_test.go` for `ReplaceAccounts()` and `LookupAccountsFromParsed()`, pinning that substituted text is never rescanned and that regexp metacharacters in a display name survive verbatim
- Added integration tests with real MongoDB for end-to-end substitution

### Documentation
- Updated `docs/notification-worker-downstream-contracts.md` with the substituted example body, the 50-account cap, the cache staleness window, the new env var, and the operator guidance for the new metric

## Implementation Details

**Mention resolution pipeline:**
1. Parse message content for @mentions (already done)
2. Filter to accounts needing lookup (exclude @all/@here), capped at 50
3. Query users collection for display names (with timeout)
4. Rewrite body: literals from constants, known accounts from map, unknown accounts kept verbatim
5. Apply same substituted body to every batch

**Failure modes:**
- No mentions → skip lookup entirely
- No survivors → skip lookup (message won't be sent anyway)
- Resolver error → use partial results, leave unresolved as @tokens, send push anyway
- Nil resolver (`MENTION_NAMES_ENABLED=false`, or a bad cache config) → only @all/@here are substituted, accounts kept verbatim
- Empty display name → account kept verbatim (preserves today's behavior for unnamed users)

**Performance:**
- Lookup runs once per message, not per batch, and only after the no-survivors early return
- Deduplicates repeated mentions in same message
- Bounded by 2-second timeout so slow collection cannot stall fan-out
- Messages with no mentions short-circuit before `ReplaceAccounts` at zero allocations

**Known behavior:** mentions are resolved from message content, so mentioning someone who is not a room member renders their display name in that room's push body. Display names are directory-sync data; flagging it as a product call rather than a defect.

https://claude.ai/code/session_01HfHNt5BXCQ9fG5bzAKXruh


julianshen:
**No actionable comments.** ✅

### Disputed

- `notification-worker/handler.go:266` (`mongodb`) — The body is recomputed from the users collection/cache on every delivery attempt, but each batch retains the stable Nats-Msg-Id `{messageId}-b{batchIndex}`. If one batch publishes and a later batch fails, a redelivery after a rename, replica-lag change, cache expiry, or lookup failure can produce a different body; NATS then suppresses the earlier batch while accepting the later one, so recipients of the same message see different renderings. Preserve the chosen rendering for the deduplication window or include deterministic mention display data in the canonical message.
- `notification-worker/handler.go:430` (`mongodb`) — Every message containing a cache-miss mention synchronously awaits the users `$in` query before any push batch is emitted. Distinct unknown accounts bypass the cache, so a sender can repeatedly force this path; when Mongo or its network path is unavailable, each attempt occupies a worker for up to `MENTION_NAMES_TIMEOUT`. With user-settings reads earlier in the same path, the timeouts can also accumulate sequentially and grow the JetStream backlog. Add automatic failure isolation—such as a breaker with bounded probes that temporarily bypasses mention lookups after timeouts/errors—so optional enrichment degrades immediately to raw mentions rather than depending on a manual kill switch.
- `notification-worker/handler.go:266` (`nats`) — The body is recomputed from a cache- and replica-dependent lookup on every source delivery, while each batch retains the stable `Nats-Msg-Id` `{messageId}-b{batchIndex}`. If batch 0 succeeds and batch 1 fails, a redelivery after a rename, cache change, or lookup failure can produce a different body; JetStream suppresses batch 0 by ID but accepts batch 1, so recipients receive different renderings of the same message. Existing title nondeterminism does not remove this newly added path. Persist or memoize the selected rendering by message ID for at least the deduplication window, or resolve it into immutable canonical data upstream.
- `notification-worker/handler.go:266` (`nats`) — `ReplaceAccounts` can expand every repeated occurrence of a short token into a much longer display name; the 50-account cap limits distinct lookups, not output size, and the resulting body is explicitly untruncated. Once a serialized batch exceeds the NATS `max_payload`, publishing cannot succeed, and replaying the source message recomputes the same oversized event, causing repeated delivery attempts until consumer termination limits intervene. The concrete risk is a poison-message/redelivery loop rather than an async-buffer OOM. Apply a UTF-8-safe byte cap after substitution and before batching, or classify oversized publishes as permanent and terminate them with explicit metrics.
- `notification-worker/handler.go:430` (`nats`) — Mention resolution is awaited before any push publish or source acknowledgement. Unknown accounts are not negatively cached, so messages containing fresh or repeated unknown mentions force backing-store reads; when that dependency blackholes, each delivery occupies a worker until `MENTION_NAMES_TIMEOUT`. Enough such deliveries can consume the shared worker pool and backlog unrelated stream messages despite enrichment being optional. The timeout and manual kill switch bound individual calls but do not provide automatic failure isolation. Add a bounded circuit breaker that bypasses mention resolution after lookup failures/timeouts, or move this optional enrichment outside the acknowledgement-critical path.
- `notification-worker/handler.go:266` (`oom`) — The async-queue mechanism from the earlier finding is incorrect, but the changed path still permits substantial attacker-driven heap amplification. As confirmed in discussion, ingress accepts about 20 KB of content; packing it with repeated `@a` tokens and resolving `a` to a 100-byte display name expands one body to about 689 KB. With the default 100 concurrent workers, roughly 69 MB can remain referenced by rendered bodies alone, plus transient regexp-building and JSON-marshalling buffers. Large rooms prolong this retention while the same body is serialized for each sequential recipient batch. A burst of such messages can therefore cause GC thrashing or container eviction/OOM. Apply a hard UTF-8-safe byte limit after mention substitution and before assigning `PushNotificationEvent.Body`—preferably the downstream payload budget—and add a repeated-token test covering expansion and multibyte truncation.
- `notification-worker/handler.go:430` (`performance-and-concurrency`) — The lookup cap bounds one query but not its rate. The changed code notes that concurrent misses are not deduplicated and unknown accounts are not negatively cached, so a burst of messages mentioning the same uncached account can issue one Mongo query per worker. With the default 100 workers, a blackholed read path can hold every slot for the two-second timeout; repeated unknown mentions sustain that state and grow the JetStream backlog even though enrichment eventually fails open. This lookup can also fail independently of the existing primary-pinned settings read. Coalesce concurrent misses, briefly cache successful negative results, and open a short-lived breaker after lookup errors/timeouts so subsequent messages immediately retain raw mentions while bounded probes test recovery.
- `notification-worker/handler.go:266` (`performance-and-concurrency`) — The 50-account cap does not limit repeated occurrences: `ReplaceAccounts` rewrites every `@a`, so a maximum-size message containing one short token can expand many-fold when its display name is long. The resulting string is then marshaled and published in every recipient batch; for a large room this produces work and broker traffic proportional to batch count times the expanded size. For example, a roughly 20 KB accepted body can become about 690 KB with a 100-byte replacement. Longer names can exceed the broker payload limit, and the current non-permanent publish failure then causes repeated redelivery of the same expensive message. Synchronous publishing avoids retaining all encodings concurrently but does not prevent this amplification or retry loop. Apply a UTF-8-safe byte cap after substitution and before batching, and treat any remaining broker-oversize condition as permanent.
- `notification-worker/handler.go:266` (`tgd-review`) — The body is recomputed from a cache- and replica-dependent lookup on every attempt, while batch IDs remain stable. If one attempt publishes only some batches, a retry after a rename, cache change, or lookup failure can have its earlier batches deduplicated and its later batches accepted with different body text. Existing title nondeterminism does not remove this newly added body inconsistency. Preserve the selected rendering for the deduplication window or make body rendering deterministic from canonical-message data.
- `notification-worker/handler.go:266` (`tgd-review`) — The 50-account cap bounds distinct lookups, not replacements: one short mention may occur throughout the content and every occurrence is replaced by a much longer display name. The unbounded expanded body is then serialized into each batch before downstream truncation, so otherwise acceptable source content can exceed the NATS payload limit and repeatedly fail publication. Bound the post-substitution size with UTF-8-safe handling, or fail open to the original raw body when substitution would exceed a safe transport budget.

### Local review memory

- `notification-worker/handler.go:266` — @evelynchang5525: The stated mechanism doesn't hold, but the underlying concern is partly real — splitting them: **Refuted: there is no async pending buffer.** PUSH_ASYNC_MAX_PENDING does not exist in this repo (no hit across *.go / *.yml ), and the em… ([thread](https://github.com/hmchangw/newchat/pull/297#discussion_r3796815786))
- `notification-worker/handler.go:266` — @evelynchang5525: The mechanism is accurate — the body is recomputed per delivery attempt from a cache- and replica-dependent lookup while Nats-Msg-Id stays {messageId}-b{batchIndex} ( emit.go:54 ), so a partial-publish failure plus a rename or replica… ([thread](https://github.com/hmchangw/newchat/pull/297#discussion_r3796815758))
- `notification-worker/handler.go:430` — @evelynchang5525: Agreed on the mechanism, with one correction to the throughput math and one reason for not scoping the fix to mentions. **Correction.** "Each worker is constrained to roughly one attempt per timeout interval" understates the concurrency: t… ([thread](https://github.com/hmchangw/newchat/pull/297#discussion_r3796815782))

<details>

* `notification-worker/mentionnames.go`
* `notification-worker/mentionnames_test.go`
* `notification-worker/nats_metrics_test.go`
* `pkg/mention/mention.go`
* `pkg/mention/mention_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

notification-worker/handler.go
```diff
@@ -260,7 +266,7 @@ func (h *Handler) HandleMessage(ctx context.Context, data []byte) (retErr error)
 	pushEvt := model.PushNotificationEvent{
```
julianshen, line 266: _🏷️ idempotency_ | _🟠 Warning_ | _`distributed-system`_

**Redelivery can produce different bodies under the same deduplication ID.**

The body is recomputed from a cache/replica-dependent lookup on every handling attempt, while each batch retains the stable Nats-Msg-Id `{messageId}-b{batchIndex}`. Concrete timeline: attempt A resolves `@bob`, publishes batch 0, then fails publishing batch 1; after a rename, replica change, or transient lookup failure, attempt B computes a different body. NATS suppresses batch 0 as a duplicate but accepts batch 1, so recipients of the same message receive different body representations. Rolling between old workers that emit raw content and new workers has the same interleaving. Make the rendered body deterministic for the canonical message—preferably resolve and store it at canonical-message write time—or persist the chosen rendering by message ID for the entire redelivery/deduplication window.

<details>

```
In `notification-worker/handler.go` around line 266: The body is recomputed from a cache/replica-dependent lookup on every handling attempt, while each batch retains the stable Nats-Msg-Id `{messageId}-b{batchIndex}`. Concrete timeline: attempt A resolves `@bob`, publishes batch 0, then fails publishing batch 1; after a rename, replica change, or transient lookup failure, attempt B computes a different body. NATS suppresses batch 0 as a duplicate but accepts batch 1, so recipients of the same message receive different body representations. Rolling between old workers that emit raw content and new workers has the same interleaving. Make the rendered body deterministic for the canonical message—preferably resolve and store it at canonical-message write time—or persist the chosen rendering by message ID for the entire redelivery/deduplication window.

```

</details>

notification-worker/handler.go
```diff
@@ -405,6 +411,59 @@ func shortRoomType(t model.RoomType) string {
...
+		return content
+	}
+	var names map[string]string
+	if h.deps.MentionNames != nil {
+		if lookup := mention.LookupAccountsFromParsed(parsed); len(lookup) > 0 {
+			// Message content is user-controlled, so cap the fan-out: the $in is
+			// unbounded otherwise, and userstore's batch path neither dedups
+			// concurrent misses nor negatively-caches them, so a message spamming
+			// unknown @tokens would re-query Mongo on every redelivery. Tokens past
+			// the cap keep their raw @token — a push body can't render 50 names anyway.
```
julianshen, line 430: _🏷️ backpressure_ | _🟠 Warning_ | _`distributed-system`_

**A slow users database stalls notification delivery before fail-open.**

Mention enrichment is default-enabled and awaited synchronously before any push batch is emitted. If Mongo or its network path blackholes requests, every message containing a cache-miss mention waits up to `MENTION_NAMES_TIMEOUT` (2 seconds); senders can force misses with distinct unknown accounts, and the cache deliberately does not negative-cache them. Repeated messages therefore constrain each worker to roughly one attempt per timeout interval and grow the JetStream backlog even though display-name enrichment is optional. Add automatic failure isolation, such as a circuit breaker that temporarily bypasses lookups after timeouts/errors (with bounded probes), so an unhealthy users collection degrades immediately to raw mentions rather than requiring an operator to flip the kill switch.

<details>

```
In `notification-worker/handler.go` around line 430: Mention enrichment is default-enabled and awaited synchronously before any push batch is emitted. If Mongo or its network path blackholes requests, every message containing a cache-miss mention waits up to `MENTION_NAMES_TIMEOUT` (2 seconds); senders can force misses with distinct unknown accounts, and the cache deliberately does not negative-cache them. Repeated messages therefore constrain each worker to roughly one attempt per timeout interval and grow the JetStream backlog even though display-name enrichment is optional. Add automatic failure isolation, such as a circuit breaker that temporarily bypasses lookups after timeouts/errors (with bounded probes), so an unhealthy users collection degrades immediately to raw mentions rather than requiring an operator to flip the kill switch.

```

</details>

notification-worker/handler.go
julianshen, line 266: _⚡ memory-exhaustion_ | _🟠 Warning_ | _`oom`_

**Mention replacement can amplify every queued batch payload without a bound**

The transformed body is not size-limited after replacing mentions. A sender can repeat one short token such as `@a` throughout a large message; replacing every occurrence with a longer display name can expand the body many-fold despite the 50-distinct-account lookup cap. The expanded body is then included in every recipient batch, and up to `PUSH_ASYNC_MAX_PENDING` (default 1024) separately encoded NATS events may be retained concurrently. A large-room message can therefore multiply one attacker-controlled body across hundreds or thousands of pending payloads, causing GC thrashing, container eviction, or process OOM. Apply a hard byte limit, ideally the documented APNs/FCM payload budget, to the substituted body before batching/serialization; truncation must occur after substitution and preserve valid UTF-8.

<details>

```
In `notification-worker/handler.go` around line 266: The transformed body is not size-limited after replacing mentions. A sender can repeat one short token such as `@a` throughout a large message; replacing every occurrence with a longer display name can expand the body many-fold despite the 50-distinct-account lookup cap. The expanded body is then included in every recipient batch, and up to `PUSH_ASYNC_MAX_PENDING` (default 1024) separately encoded NATS events may be retained concurrently. A large-room message can therefore multiply one attacker-controlled body across hundreds or thousands of pending payloads, causing GC thrashing, container eviction, or process OOM. Apply a hard byte limit, ideally the documented APNs/FCM payload budget, to the substituted body before batching/serialization; truncation must occur after substitution and preserve valid UTF-8.

```

</details>

julianshen (commented):
tGD inline review

notification-worker/handler.go
evelynchang5525: The stated mechanism doesn't hold, but the underlying concern is partly real — splitting them:

**Refuted: there is no async pending buffer.** `PUSH_ASYNC_MAX_PENDING` does not exist in this repo (no hit across `*.go` / `*.yml`), and the emitter publishes **synchronously**: `jsPublisher.PublishMsg` (`emit.go:82-86`) calls `js.PublishMsg` and blocks for the `PubAck`. There is no `PublishAsync` anywhere in the service. So batches are marshaled, published, and released one at a time — 1024 concurrently-retained encoded payloads is not a state this code can reach, and the OOM conclusion drawn from it doesn't follow.

**Confirmed: the amplification itself is real.** Content is capped at 20 KB at ingress (`message-gatekeeper/handler.go:31`, `maxContentBytes`), and I measured the worst case by packing that budget with the shortest repeatable token:

| display name | content | substituted body | ratio |
|---|---|---|---|
| 20 B | 20478 B | 143346 B | 7.0x |
| 40 B | 20478 B | 279866 B | 13.7x |
| 100 B | 20478 B | 689426 B | 33.7x |

So a single sender can turn 20 KB into a ~689 KB body, and in a large room that body is re-encoded per batch (sequentially, not concurrently). That is real marshal churn and GC pressure, and past the broker's `max_payload` the pre-flight guard at `emit.go:44-46` returns a **non-permanent** error, so the batch NAKs and redelivers rather than being dropped — a redelivery loop is the more likely failure than OOM.

**Why I'm not truncating in this PR.** Worker-side truncation is an explicit, already-tracked decision in the contract doc: `body` is specified as **untruncated**, push-service truncates to the APNs/FCM budget, and "truncation/PII-scrubbing on the worker side is tracked as follow-up" (`docs/notification-worker-downstream-contracts.md` §1). Changing that here would silently alter a published downstream contract in a PR about mention rendering, and the cutoff point and any ellipsis marker are a product call, not something to pick unilaterally. Raising it with the PR owner; happy to add a UTF-8-safe byte cap in a follow-up if they want it now.

Worth noting the pre-existing shape: the 20 KB → oversized-body path exists today for any long message, mentions or not. This change widens the ratio, it does not create the failure mode.

---

evelynchang5525: The mechanism is accurate — the body is recomputed per delivery attempt from a cache- and replica-dependent lookup while `Nats-Msg-Id` stays `{messageId}-b{batchIndex}` (`emit.go:54`), so a partial-publish failure plus a rename or replica change between attempts can leave batch 0 deduped with the old rendering and batch 1 accepted with the new one. I'm not disputing the timeline.

Where I'd push back is on it being **new**, and on the proposed remedy.

**The envelope is already non-deterministic across redeliveries on an already-merged path.** `resolveTitle` (`handler.go:468-472`) reads the room name from `pkg/roommetacache` — an LRU+TTL cache with the same replica-lag and staleness properties. A room renamed between attempt A and attempt B yields a different `title` under exactly the same stable dedup ID, with the same partial-publish interleaving. So "recipients of one message can see two renderings of the envelope" is a property this worker already had before mention substitution existed; this change extends it from `title` to `body`. That reframes it as a pre-existing design characteristic to fix deliberately, not a regression to block this PR on.

**On the suggested fix.** Resolving at canonical-message write time was considered and deliberately rejected: `message-gatekeeper` already composes `sender.displayName` there, but the mention rewrite is needed *only* by push, and moving it upstream would make every consumer of the canonical message (Cassandra history, search-sync, broadcast) carry a push-shaped rendering of the content. The alternative — persisting the chosen rendering per message ID for the dedup window — adds a new store write on the per-message hot path to fix a cosmetic divergence in a notification.

The realistic blast radius is that some recipients of one message see `Bob Chen` and others see `@bob`, in a push preview, only when a publish partially fails *and* the underlying name changed in the gap. Message delivery, ordering, and content are unaffected.

I'd rather not take either remedy inside this PR. If the inconsistency is judged worth closing, the coherent fix covers `title` and `body` together — that's a design change for its own PR, and I'll open an issue for it if the owner agrees.

---

notification-worker/handler.go
evelynchang5525: Agreed on the mechanism, with one correction to the throughput math and one reason for not scoping the fix to mentions.

**Correction.** "Each worker is constrained to roughly one attempt per timeout interval" understates the concurrency: the consumer runs `MAX_WORKERS` (default 100, `main.go:50`) in-flight message goroutines, so a fully blackholed users collection caps a pod at ~100 messages per 2 s, not one. Still a severe degradation, but two orders of magnitude off the stated figure, and it only affects messages that *contain* mentions with cache misses — mention-free messages short-circuit before any lookup (`handler.go:420`) and are unaffected.

**The negative-caching point is correct and is the sharp edge here.** `userstore.Cache` doesn't negatively-cache misses, so repeated unknown accounts stay permanent misses. That's why this PR caps the per-message lookup at 50 distinct accounts — it bounds the fan-out per message, though it doesn't bound the *rate* across messages, which is your actual point.

**Why not a mention-specific breaker.** Mention enrichment isn't the only synchronous Mongo/RPC dependency in this handler — the same per-message path already awaits user settings (2 s, Mongo), presence (2 s, RPC), and the badge RPC, all before emit. A blackholed Mongo stalls this worker through **user settings** first, and that one gates delivery decisions rather than being optional, so a breaker on mentions alone would leave the dominant stall path untouched while adding a second, inconsistent failure-isolation mechanism to the service.

The right shape is one breaker policy applied across all four dependencies, with the optional ones (mentions, presence) failing open and the delivery-gating one (settings) failing closed. That's a service-wide change with its own state, probe policy, and tests — larger than this PR and not something to bolt on mid-review.

What this PR does ship toward the same goal: `MENTION_NAMES_ENABLED` (an operator kill switch that didn't exist before), and `notification_worker_mention_resolution_total{result="failed"}`, which is the signal that would drive either an alert or an automatic breaker. Manual today, but the observability to justify and tune a breaker now exists. I'll propose the service-wide breaker as a follow-up.

---

notification-worker/handler.go
```diff
@@ -405,6 +411,59 @@ func shortRoomType(t model.RoomType) string {
...
+			if len(lookup) > maxMentionLookups {
+				lookup = lookup[:maxMentionLookups]
+			}
+			resolved, err := h.deps.MentionNames.Resolve(ctx, lookup)
+			names = resolved
+			h.recordMentionResults(ctx, lookup, resolved, err)
```
mliu33, line 436: Let's remove this recordMentionResults as I think this is unnecessary. Let's only record metrics for important failures for now since too many redundant metrics will have some impact to the performance. Thanks

mliu33 (approved):
Excellent, thanks a lot

PR #301, opened 2026-08-17
fix(admin-service): permission fanout review fixes — per-destination lanes, 1000-account chunks, FANOUT_TIMEOUT 30s

nalydadad:
## Summary

Post-merge fixes to the `admin-service` cross-site **permission fanout**, answering the review threads on #265. Scope is the four items accepted for this round; the rest are recorded as deferred below.

| Commit | Review thread | Change |
|---|---|---|
| `fix: fan out resync per destination…` | [julianshen — a blocked peer prevents later resync groups reaching healthy peers](https://github.com/hmchangw/newchat/pull/265#discussion_r3776656870) | Resync hands every `(accounts, state)` group to **one** fanout call; each destination walks all groups' chunks in its own goroutine lane under the shared budget, so a stalled peer burns only its own lane and can no longer hand a later group an expired ctx. Once the budget is gone a lane stops and is reported once. The budget moves from a package constant to **`FANOUT_TIMEOUT`** config, validated at startup below the HTTP write timeout the same way `ROOM_RPC_TIMEOUT` is (this is also what lets the stalled-peer regression test shrink it via `cfg`). |
| `fix: cap fanout chunks at 1000 accounts…` | [mliu33 — we only have 128kb max size, use 1000?](https://github.com/hmchangw/newchat/pull/265#discussion_r3779998931) | 5000 → 1000 accounts/event. The INBOX envelope base64-encodes the payload (`InboxEvent.Payload` is `[]byte`, ~4/3 inflation); `TestFanoutChunk_FitsBrokerMaxPayload` pushes a worst-case chunk (1000 × 64-char accounts, every state bound set) through the real publisher and asserts the envelope stays under 128KB (~90KB; the old 5000 fails it). |
| `fix: raise the fanout budget to 30s and the HTTP write timeout to 40s` | [mliu33 — too short for whole-site fanout, use 30s](https://github.com/hmchangw/newchat/pull/265#discussion_r3780070538) | `FANOUT_TIMEOUT` default 5s → 30s. `net/http`'s `WriteTimeout` counts in-handler time, so it moves 30s → 40s to keep the budget strictly below it — otherwise the connection drops before the admin can read `syncFailures`. `loadConfig` rejects any `FANOUT_TIMEOUT` at or above the write timeout. |
| `refactor: precompute remote fanout destinations at construction, deduped` | [mliu33 — build this up initially instead of looping every time](https://github.com/hmchangw/newchat/pull/265#discussion_r3780061667) | `remoteSites(cfg.AllSiteIDs, cfg.SiteID)` computed once in `newHandler`. Also collapses a peer listed twice in `ALL_SITE_IDS` (it used to get two lanes and could appear twice in `syncFailures`). |
| `docs(specs): amend…` | — | Design specs under `docs/superpowers/specs/` amended for this round (chunk size, lanes, `FANOUT_TIMEOUT`, single-call resync). |

Wire shape is unchanged (`inbox-worker` needs nothing); the only client-visible difference is that `POST /v1/admin/permissions` and `/resync` may now take up to 30s to answer while a remote peer is unreachable.

### New config

`FANOUT_TIMEOUT` (default `30s`, must be `> 0` and `< 40s`) — added to `admin-service/deploy/docker-compose.yml`. Fail-fast at startup like `ROOM_RPC_TIMEOUT`.

### Deploy note

With a stalled peer every permission POST/resync now runs for up to 30s. Any ingress / reverse proxy in front of admin-service needs an upstream read timeout ≥ 40s, or it will return 502/504 before the `syncFailures` reply arrives — the reply this round exists to deliver.

## Testing

- `make lint` clean; `make test SERVICE=admin-service` (race) green; every commit in the range is individually green.
- New/extended tests: stalled-peer lane isolation (deterministic, ctx-driven, no `time.Sleep`), single shared deadline across resync groups, worst-case chunk envelope under 128KB, empty-batch resync, `FANOUT_TIMEOUT` load/validation/override, `remoteSites` dedupe.
- Final whole-branch review: 0 critical; the one important finding (unpinned chunk-size assumption) is fixed in this branch.
- SAST: gosec and govulncheck clean locally; semgrep is not installed on this machine and runs in the CI `sast` job.

## Deferred from the #265 review (owner decision, not in this PR)

- [julianshen — wall-clock watermarks can discard revocations across replicas (monotonic revision)](https://github.com/hmchangw/newchat/pull/265#discussion_r3776656851) — needs a per-user revision design; tracked as follow-up.
- [julianshen — a crash after commit loses the cross-site update (durable OUTBOX for the fanout)](https://github.com/hmchangw/newchat/pull/265#discussion_r3776656858) — the in-process best-effort fanout + `syncFailures` + manual resync stays for now; the OUTBOX-lite migration is the recorded upgrade path.
- [julianshen — apply is a silent no-op when the remote site has no user doc](https://github.com/hmchangw/newchat/pull/265#discussion_r3776656879), [unknown-key runbook note](https://github.com/hmchangw/newchat/pull/265#discussion_r3776656887), [`ALL_SITE_IDS` should be required](https://github.com/hmchangw/newchat/pull/265#discussion_r3776656900), [AccountPicker error state](https://github.com/hmchangw/newchat/pull/265#discussion_r3776656914), [AccountPicker search](https://github.com/hmchangw/newchat/pull/265#discussion_r3776656921) — follow-ups.
- Deriving the chunk size from `nc.NatsConn().MaxPayload()` at runtime instead of the pinned constant.

## Self-Review Updates
- **Fanout deadline capped by the whole-request budget** (`b649e115`, answers the tGD blocking thread): `createPermissions`/`resyncPermissions` pin one absolute deadline at handler entry (`requestBudget = httpWriteTimeout − 2s`) and the fanout takes the earlier of that and `now + FANOUT_TIMEOUT` before severing client cancellation, so local work that already spent part of the write window can no longer let a stalled peer push the response past net/http's WriteTimeout. An already-spent budget reports every destination without publishing.
- Declined/deferred with replies on-thread: resync prebuffer OOM (bounded by real accounts with a stored state), site-ID token validation (repo-wide follow-up in `pkg/subject`), fanout 30s vs shutdown 25s (documented trade-off), account-ID length bound (owner decision).

### DIFF
--- a/admin-service/config.go
+++ b/admin-service/config.go
@@ -12,6 +12,11 @@ import (
 // the handler always wins the race against net/http closing the connection.
 const httpWriteTimeout = 40 * time.Second
 
+// requestBudget is the absolute per-request deadline the permission handlers pin at
+// entry (withRequestBudget): httpWriteTimeout minus a margin to write the response
+// after the last in-handler wait, so syncFailures always reaches the caller.
+const requestBudget = httpWriteTimeout - 2*time.Second
+
 type Config struct {
 	Port                  string `env:"PORT" envDefault:"8082"`
 	SiteID                string `env:"SITE_ID,required"`
--- a/admin-service/config.go
+++ b/admin-service/config.go
@@ -10,7 +10,7 @@ import (
 // httpWriteTimeout bounds a response write measured from the request, so in-handler
 // waits count against it: RoomRPCTimeout and FanoutTimeout must stay under it so
 // the handler always wins the race against net/http closing the connection.
-const httpWriteTimeout = 30 * time.Second
+const httpWriteTimeout = 40 * time.Second
 
 type Config struct {
 	Port                  string `env:"PORT" envDefault:"8082"`
@@ -29,10 +29,13 @@ type Config struct {
 	// closes the connection before the handler can answer.
 	RoomRPCTimeout time.Duration `env:"ROOM_RPC_TIMEOUT" envDefault:"5s"`
 	// FanoutTimeout is the server-side budget for ONE request's whole cross-site
-	// permission fanout — every destination lane, every batch, every chunk. Like
-	// RoomRPCTimeout it must stay below the HTTP write timeout, or net/http drops the
-	// connection before the admin can read syncFailures.
-	FanoutTimeout time.Duration `env:"FANOUT_TIMEOUT" envDefault:"5s"`
+	// permission fanout — every destination lane, every batch, every chunk. Sized for a
+	// whole-site batch over cross-site gateways. Like RoomRPCTimeout it must stay below
+	// the HTTP write timeout, or net/http drops the connection before the admin can read
+	// syncFailures. The default exceeds main.go's 25s graceful-shutdown budget: a SIGTERM
+	// mid-fanout can kill the request before the admin sees syncFailures. Accepted — the
+	// ledger write already committed, and resync re-delivers anything cut short.
+	FanoutTimeout time.Duration `env:"FANOUT_TIMEOUT" envDefault:"30s"`
 	// AllSiteIDs lists every site in the federation (including this one); empty means
 	// no cross-site fanout — correct for single-site dev.
 	AllSiteIDs []string `env:"ALL_SITE_IDS" envSeparator:"," envDefault:""`
--- a/admin-service/config.go
+++ b/admin-service/config.go
@@ -7,7 +7,8 @@ import (
 	"github.com/caarlos0/env/v11"
 )
 
-// httpWriteTimeout bounds a response write; RoomRPCTimeout must stay under it so
+// httpWriteTimeout bounds a response write measured from the request, so in-handler
+// waits count against it: RoomRPCTimeout and FanoutTimeout must stay under it so
 // the handler always wins the race against net/http closing the connection.
 const httpWriteTimeout = 30 * time.Second
 
@@ -27,6 +28,11 @@ type Config struct {
 	// RoomRPCTimeout must stay below the HTTP server's WriteTimeout, or net/http
 	// closes the connection before the handler can answer.
 	RoomRPCTimeout time.Duration `env:"ROOM_RPC_TIMEOUT" envDefault:"5s"`
+	// FanoutTimeout is the server-side budget for ONE request's whole cross-site
+	// permission fanout — every destination lane, every batch, every chunk. Like
+	// RoomRPCTimeout it must stay below the HTTP write timeout, or net/http drops the
+	// connection before the admin can read syncFailures.
+	FanoutTimeout time.Duration `env:"FANOUT_TIMEOUT" envDefault:"5s"`
 	// AllSiteIDs lists every site in the federation (including this one); empty means
 	// no cross-site fanout — correct for single-site dev.
 	AllSiteIDs []string `env:"ALL_SITE_IDS" envSeparator:"," envDefault:""`
@@ -37,11 +43,24 @@ func loadConfig() (Config, error) {
 	if err := env.Parse(&c); err != nil {
 		return Config{}, err
 	}
-	if c.RoomRPCTimeout <= 0 {
-		return Config{}, fmt.Errorf("invalid ROOM_RPC_TIMEOUT %s: must be > 0", c.RoomRPCTimeout)
+	if err := checkHandlerTimeout("ROOM_RPC_TIMEOUT", c.RoomRPCTimeout); err != nil {
+		return Config{}, err
 	}
-	if c.RoomRPCTimeout >= httpWriteTimeout {
-		return Config{}, fmt.Errorf("invalid ROOM_RPC_TIMEOUT %s: must be below the %s HTTP write timeout", c.RoomRPCTimeout, httpWriteTimeout)
+	if err := checkHandlerTimeout("FANOUT_TIMEOUT", c.FanoutTimeout); err != nil {
+		return Config{}, err
 	}
 	return c, nil
+}
+
+// checkHandlerTimeout rejects a per-request budget the handler cannot honour: a
+// non-positive value yields an already-expired context, and one at or above
+// httpWriteTimeout lets net/http close the connection before the handler answers.
+func checkHandlerTimeout(name string, d time.Duration) error {
+	if d <= 0 {
+		return fmt.Errorf("invalid %s %s: must be > 0", name, d)
+	}
+	if d >= httpWriteTimeout {
+		return fmt.Errorf("invalid %s %s: must be below the %s HTTP write timeout", name, d, httpWriteTimeout)
+	}
+	return nil
 }
--- a/admin-service/permissions.go
+++ b/admin-service/permissions.go
@@ -136,7 +136,8 @@ func displayUntilDate(t time.Time) string {
 // a permission for one or more subject accounts, all-or-nothing, with one
 // audit entry per created row. Validation order matches spec §4.4 exactly.
 func (h *Handler) createPermissions(c *gin.Context) {
-	ctx := c.Request.Context()
+	ctx, cancel := withRequestBudget(c)
+	defer cancel()
 
 	var req permissionRequest
 	if err := c.ShouldBindJSON(&req); err != nil {
@@ -322,12 +323,19 @@ func (h *Handler) publishPermissionFanout(ctx context.Context, permission model.
 	if h.publishInbox == nil || len(dests) == 0 {
 		return nil
 	}
-	// The request ctx has no deadline and dies if the client disconnects — neither fits:
-	// the caller's local work is done (create's ledger write already committed; resync
-	// writes nothing), so re-delivery runs to completion on its own budget
-	// (cfg.FanoutTimeout), shared by every lane. A destination that doesn't land within
-	// it is reported like any other unacknowledged publish.
-	ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), h.cfg.FanoutTimeout)
+	// The request ctx dies if the client disconnects, which doesn't fit: the caller's
+	// local work is done (create's ledger write already committed; resync writes
+	// nothing), so re-delivery runs to completion on its own budget, shared by every
+	// lane. That budget is cfg.FanoutTimeout capped by the request's absolute deadline
+	// (withRequestBudget): the local work already spent part of net/http's write window,
+	// and a fresh full FanoutTimeout on top could outlive it — the connection would
+	// close before the caller reads syncFailures. A destination that doesn't land in
+	// time is reported like any other unacknowledged publish.
+	deadline := time.Now().Add(h.cfg.FanoutTimeout)
+	if reqDeadline, ok := ctx.Deadline(); ok && reqDeadline.Before(deadline) {
+		deadline = reqDeadline
+	}
+	ctx, cancel := context.WithDeadline(context.WithoutCancel(ctx), deadline)
 	defer cancel()
 
 	now := time.Now().UTC().UnixMilli()
@@ -403,6 +411,13 @@ func (h *Handler) publishPermissionFanout(ctx context.Context, permission model.
 	return failures
 }
 
+// withRequestBudget pins one absolute deadline for the whole request, measured from
+// handler entry: net/http's WriteTimeout runs from the request, so the fanout can
+// only spend what the local work left over — see publishPermissionFanout.
+func withRequestBudget(c *gin.Context) (context.Context, context.CancelFunc) {
+	return context.WithTimeout(c.Request.Context(), requestBudget)
+}
+
 // remoteSites filters allSiteIDs down to real remote destinations: no self, no blanks,
 // no repeats — a peer listed twice in ALL_SITE_IDS would otherwise get a second lane,
 // a duplicate publish, and a duplicate syncFailures entry.
@@ -431,7 +446,8 @@ type resyncPermissionsResponse struct {
 // and is idempotent (delivered states keep their stored watermarks, so remote guards
 // no-op anything already applied).
 func (h *Handler) resyncPermissions(c *gin.Context) {
-	ctx := c.Request.Context()
+	ctx, cancel := withRequestBudget(c)
+	defer cancel()
 	var req resyncPermissionsRequest
 	if err := c.ShouldBindJSON(&req); err != nil {
 		errhttp.Write(ctx, c, errcode.BadRequest("invalid request body", errcode.WithReason(errcode.AuthMissingFields)))
--- a/admin-service/permissions.go
+++ b/admin-service/permissions.go
@@ -318,7 +318,7 @@ type fanoutBatch struct {
 // aggregated per destination and never abort the fanout. Returns the destinations whose
 // publish was not acknowledged (nil when all landed).
 func (h *Handler) publishPermissionFanout(ctx context.Context, permission model.PermissionKey, batches []fanoutBatch) []string {
-	dests := remoteSites(h.cfg.AllSiteIDs, h.cfg.SiteID)
+	dests := h.remoteDests
 	if h.publishInbox == nil || len(dests) == 0 {
 		return nil
 	}
@@ -344,7 +344,7 @@ func (h *Handler) publishPermissionFanout(ctx context.Context, permission model.
 			if err != nil {
 				// Cannot serialize the event at all: every destination is out of sync.
 				slog.ErrorContext(ctx, "marshal user permissions event", "error", err)
-				return dests
+				return slices.Clone(dests)
 			}
 			chunks = append(chunks, payload)
 		}
@@ -403,10 +403,13 @@ func (h *Handler) publishPermissionFanout(ctx context.Context, permission model.
 	return failures
 }
 
-// remoteSites filters allSiteIDs down to real remote destinations.
+// remoteSites filters allSiteIDs down to real remote destinations: no self, no blanks,
+// no repeats — a peer listed twice in ALL_SITE_IDS would otherwise get a second lane,
+// a duplicate publish, and a duplicate syncFailures entry.
 func remoteSites(allSiteIDs []string, self string) []string {
 	var out []string
-	for _, dest := range allSiteIDs {
+	deduped, _ := dedupPreserveOrder(allSiteIDs)
+	for _, dest := range deduped {
 		if dest != "" && dest != self {
 			out = append(out, dest)
 		}
--- a/admin-service/permissions.go
+++ b/admin-service/permissions.go
@@ -295,8 +295,14 @@ func (h *Handler) createPermissions(c *gin.Context) {
 }
 
 // fanoutChunkSize bounds one UserPermissionsUpdated event's Accounts so a worst-case
-// event stays ~320KB — 3× margin under NATS's default 1MB max_payload.
-const fanoutChunkSize = 5000
+// chunk's INBOX envelope stays under the 128KB max_payload our NATS brokers are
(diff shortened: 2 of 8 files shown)


julianshen:
## Needs clarification

What server-side validation guarantees account IDs are bounded, unescaped ASCII?
Reply in this thread, or post: `answer clar_yr4pdp74mriffkprxtdxjav3o6: <your answer>`

julianshen:
**10 findings · 7 unique issues · 7 inline comments posted** — 🔴 3 blocking · 🟠 7 warning

### Needs clarification

What server-side validation guarantees account IDs are bounded, unescaped ASCII?

`answer clar_yr4pdp74mriffkprxtdxjav3o6: <your answer>`

[Open the question](https://github.com/hmchangw/newchat/pull/301#issuecomment-5321939331)

### Related work

- [PR #265](https://github.com/hmchangw/newchat/pull/265) — feat: user permission whitelist with cross-site sync and admin bulk operations (merged)

<details>

* `admin-service/config.go`
* `admin-service/config_test.go`
* `admin-service/deploy/docker-compose.yml`
* `admin-service/handler.go`
* `docs/superpowers/specs/2026-08-13-permission-bulk-frontend-design.md`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

julianshen (commented):
tGD inline review

admin-service/permissions.go
```diff
@@ -297,63 +295,82 @@ func (h *Handler) createPermissions(c *gin.Context) {
...
+	if h.publishInbox == nil || len(dests) == 0 {
 		return nil
 	}
+	// The request ctx has no deadline and dies if the client disconnects — neither fits:
+	// the caller's local work is done (create's ledger write already committed; resync
+	// writes nothing), so re-delivery runs to completion on its own budget
+	// (cfg.FanoutTimeout), shared by every lane. A destination that doesn't land within
+	// it is reported like any other unacknowledged publish.
+	ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), h.cfg.FanoutTimeout)
+	defer cancel()
```
julianshen, line 340: _🏷️ failure-behavior_ | _🔴 Blocking_ | _`distributed-system`, `nats`, `performance-and-concurrency`, `silent-failure` (4 rules)_

**A fresh fanout timeout can outlive the HTTP request.**

The timeout starts only after the local work has completed, while the HTTP WriteTimeout is measured from request start. For example, the ledger can commit after 11s, then a stalled destination can consume the default 30s fanout budget; at 40s net/http closes the connection, and the handler reaches its response around 41s. Configuration also permits values arbitrarily close to 40s, making even minimal pre-fanout work sufficient. The caller therefore observes failure after the ledger committed and may not receive syncFailures, while remote sites have only partially applied the update. The same ambiguity occurs during shutdown because the detached default 30s fanout exceeds the documented 25s shutdown budget. Establish one absolute per-request deadline at ingress that reserves response time and fits within graceful shutdown, then detach client cancellation without discarding that deadline and let fanout use only its remaining budget. Alternatively, persist fanout work in a durable outbox before acknowledging the local write.

<details>

- **`silent-failure`**:328: The 30s timeout starts only after request parsing and the create transaction/resync read have completed, while net/http's 40s WriteTimeout starts at request arrival. If that earlier work takes more than 10s and a peer then stalls for the fanout budget, net/http closes the connection before c.JSON can deliver syncFailures, silently hiding the failed delivery. Establish an end-to-end request deadline at ingress and bound the detached fanout deadline by its remaining time with a response-writing margin, rather than always granting a fresh FanoutTimeout.
- **`nats`**:330: The FANOUT_TIMEOUT timer starts only after the Mongo work, while the 40s HTTP WriteTimeout starts with the request. Thus a create taking more than 10s before the default 30s fanout, or almost any delay with a configured timeout near 40s, lets a stalled NATS publish consume the HTTP deadline before syncFailures can be returned. The default also exceeds the documented 25s graceful-shutdown budget; because WithoutCancel detaches fanout from request cancellation, SIGTERM can terminate the process before the publish lanes finish. Establish an end-to-end handler deadline at request entry and give fanout only the remaining time, then ensure graceful shutdown exceeds that deadline plus drain margin.
- **`performance-and-concurrency`**:328: The full FanoutTimeout starts only after the preceding database, audit, and grouping work, while net/http's 40s write deadline has already been running. With the defaults, if pre-fanout work takes 11s under load, this detached context may run another 30s and c.JSON executes after the connection's write deadline, losing syncFailures while continuing work that can no longer produce a response. Establish an absolute whole-request deadline below httpWriteTimeout at handler entry, then give fanout the earlier of that deadline and now+FanoutTimeout; if client cancellation must remain detached, apply that absolute deadline to context.WithoutCancel(ctx).

</details>

<details>

```
In `admin-service/permissions.go` around line 332: The timeout starts only after the local work has completed, while the HTTP WriteTimeout is measured from request start. For example, the ledger can commit after 11s, then a stalled destination can consume the default 30s fanout budget; at 40s net/http closes the connection, and the handler reaches its response around 41s. Configuration also permits values arbitrarily close to 40s, making even minimal pre-fanout work sufficient. The caller therefore observes failure after the ledger committed and may not receive syncFailures, while remote sites have only partially applied the update. The same ambiguity occurs during shutdown because the detached default 30s fanout exceeds the documented 25s shutdown budget. Establish one absolute per-request deadline at ingress that reserves response time and fits within graceful shutdown, then detach client cancellation without discarding that deadline and let fanout use only its remaining budget. Alternatively, persist fanout work in a durable outbox before acknowledging the local write.

```

</details>

admin-service/permissions.go
```diff
@@ -297,63 +295,82 @@ func (h *Handler) createPermissions(c *gin.Context) {
...
+				Accounts:   accounts,
+				State:      b.state,
+				Timestamp:  now,
+			})
+			if err != nil {
+				// Cannot serialize the event at all: every destination is out of sync.
+				slog.ErrorContext(ctx, "marshal user permissions event", "error", err)
+				return slices.Clone(dests)
+			}
+			chunks = append(chunks, payload)
```
julianshen, line 357: _⚡ memory-exhaustion_ | _🔴 Blocking_ | _`oom`_

**Resync prebuffers every state group without a size bound.**

The resync endpoint has no subject-count or request-body limit, and stored states can form one group per account because grouping includes UpdatedAt. The new single fanout call marshals every group's chunks and appends all payloads to `chunks` before publishing begins. This retains the request, query/group data, and all serialized copies until every destination finishes; a stalled destination can keep them reachable for the 30-second timeout even after client disconnect because the context uses WithoutCancel. Large or concurrent resyncs can therefore cause GC thrashing or process OOM. The previous per-group calls released each group's serialized chunks before processing the next group. Preserve the shared deadline while generating payloads on demand with bounded buffering, or enforce a hard request/body limit.

<details>

```
In `admin-service/permissions.go` around line 349: The resync endpoint has no subject-count or request-body limit, and stored states can form one group per account because grouping includes UpdatedAt. The new single fanout call marshals every group's chunks and appends all payloads to `chunks` before publishing begins. This retains the request, query/group data, and all serialized copies until every destination finishes; a stalled destination can keep them reachable for the 30-second timeout even after client disconnect because the context uses WithoutCancel. Large or concurrent resyncs can therefore cause GC thrashing or process OOM. The previous per-group calls released each group's serialized chunks before processing the next group. Preserve the shared deadline while generating payloads on demand with bounded buffering, or enforce a hard request/body limit.

```

</details>

julianshen, line 357: _🏷️ backpressure_ | _🟠 Warning_ | _`nats`_

**Resync buffers every event before publishing any of them.**

All batches are marshaled into chunks before the destination lanes start. Resync can now combine every distinct stored state into this one call, and the request has no subject-count/body cap, so many state groups create an unbounded number of retained payloads. The timeout does not interrupt this loop because neither json.Marshal nor the loop checks ctx; a large resync can exhaust memory or pass the HTTP deadline before making one NATS publish. Stream or bounded-buffer the chunks to each destination lane and check ctx between chunks instead of materializing the complete fanout first.

<details>

```
In `admin-service/permissions.go` around line 349: All batches are marshaled into chunks before the destination lanes start. Resync can now combine every distinct stored state into this one call, and the request has no subject-count/body cap, so many state groups create an unbounded number of retained payloads. The timeout does not interrupt this loop because neither json.Marshal nor the loop checks ctx; a large resync can exhaust memory or pass the HTTP deadline before making one NATS publish. Stream or bounded-buffer the chunks to each destination lane and check ctx between chunks instead of materializing the complete fanout first.

```

</details>

admin-service/permissions.go
```diff
@@ -297,63 +295,82 @@ func (h *Handler) createPermissions(c *gin.Context) {
...
+	// the reported order stays the configured destination order.
 	failed := make([]bool, len(dests))
 	var g errgroup.Group
 	for i, dest := range dests {
 		g.Go(func() error {
-			for _, payload := range chunks {
+			subj := subject.InboxExternal(dest, model.InboxUserPermissionsUpdated)
+			for n, payload := range chunks {
+				if err := ctx.Err(); err != nil {
+					// The budget is gone: every remaining chunk would fail the same way, so
```
julianshen, line 376: _🏷️ subjects_ | _🟠 Warning_ | _`nats`_

**Destination IDs are interpolated into subjects without token validation.**

remoteSites only removes blanks, self, and exact duplicates before dest is inserted into chat.inbox.{dest}.external.user_permissions_updated. An ALL_SITE_IDS entry containing '*', '>', whitespace, or an empty token produces an invalid publish subject, while an ID containing '.' changes the subject token boundaries. The service still starts and repeatedly reports that destination as unsynchronized; depending on stream filters, dotted IDs may also route to an unintended stream. Validate SITE_ID and every ALL_SITE_IDS entry at startup as exactly one nonempty literal NATS token with no '.', '*', '>', or whitespace, and reject invalid configuration before constructing remoteDests.

<details>

```
In `admin-service/permissions.go` around line 368: remoteSites only removes blanks, self, and exact duplicates before dest is inserted into chat.inbox.{dest}.external.user_permissions_updated. An ALL_SITE_IDS entry containing '*', '>', whitespace, or an empty token produces an invalid publish subject, while an ID containing '.' changes the subject token boundaries. The service still starts and repeatedly reports that destination as unsynchronized; depending on stream filters, dotted IDs may also route to an unintended stream. Validate SITE_ID and every ALL_SITE_IDS entry at startup as exactly one nonempty literal NATS token with no '.', '*', '>', or whitespace, and reject invalid configuration before constructing remoteDests.

```

</details>

admin-service/config.go
```diff
@@ -27,6 +28,14 @@ type Config struct {
...
 	// closes the connection before the handler can answer.
 	RoomRPCTimeout time.Duration `env:"ROOM_RPC_TIMEOUT" envDefault:"5s"`
+	// FanoutTimeout is the server-side budget for ONE request's whole cross-site
+	// permission fanout — every destination lane, every batch, every chunk. Sized for a
+	// whole-site batch over cross-site gateways. Like RoomRPCTimeout it must stay below
+	// the HTTP write timeout, or net/http drops the connection before the admin can read
+	// syncFailures. The default exceeds main.go's 25s graceful-shutdown budget: a SIGTERM
+	// mid-fanout can kill the request before the admin sees syncFailures. Accepted — the
+	// ledger write already committed, and resync re-delivers anything cut short.
+	FanoutTimeout time.Duration `env:"FANOUT_TIMEOUT" envDefault:"30s"`
```
julianshen, line 43: _🏷️ shutdown_ | _🟠 Warning_ | _`performance-and-concurrency`_

**The default fanout exceeds the graceful-shutdown budget.**

The new 30s default is longer than main.go's documented 25s shutdown budget. If SIGTERM arrives near the start of fanout, graceful shutdown times out first and process exit kills the handler and destination lanes before they can report failures or finish delivery. This makes routine rolling deployments truncate committed permission fanout. Increase the graceful-shutdown budget beyond the maximum HTTP request lifetime, or validate/cap FanoutTimeout so fanout and response completion fit within shutdown.

<details>

```
In `admin-service/config.go` around line 38: The new 30s default is longer than main.go's documented 25s shutdown budget. If SIGTERM arrives near the start of fanout, graceful shutdown times out first and process exit kills the handler and destination lanes before they can report failures or finish delivery. This makes routine rolling deployments truncate committed permission fanout. Increase the graceful-shutdown budget beyond the maximum HTTP request lifetime, or validate/cap FanoutTimeout so fanout and response completion fit within shutdown.

```

</details>

admin-service/permissions.go
```diff
@@ -297,63 +295,82 @@ func (h *Handler) createPermissions(c *gin.Context) {
...
 	}
+	// The request ctx has no deadline and dies if the client disconnects — neither fits:
+	// the caller's local work is done (create's ledger write already committed; resync
+	// writes nothing), so re-delivery runs to completion on its own budget
+	// (cfg.FanoutTimeout), shared by every lane. A destination that doesn't land within
+	// it is reported like any other unacknowledged publish.
+	ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), h.cfg.FanoutTimeout)
+	defer cancel()
 
 	now := time.Now().UTC().UnixMilli()
```
julianshen, line 341: _⚡ memory_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Resync materializes every serialized group before publishing.**

The resync path now passes all state groups into one call, and this loop retains every marshaled payload in chunks before any destination starts publishing. Because resync has no account-count or request-body bound, a request containing many accounts with distinct UpdatedAt values creates one payload allocation per account and retains all of them concurrently; concurrent large resyncs can cause substantial heap growth or OOM. Previously each group's payloads became reclaimable after that group's fanout call. Keep the single shared deadline and independent destination lanes, but have each lane iterate batches and marshal only its current chunk, or otherwise use bounded per-lane buffering.

<details>

```
In `admin-service/permissions.go` around line 333: The resync path now passes all state groups into one call, and this loop retains every marshaled payload in chunks before any destination starts publishing. Because resync has no account-count or request-body bound, a request containing many accounts with distinct UpdatedAt values creates one payload allocation per account and retains all of them concurrently; concurrent large resyncs can cause substantial heap growth or OOM. Previously each group's payloads became reclaimable after that group's fanout call. Keep the single shared deadline and independent destination lanes, but have each lane iterate batches and marshal only its current chunk, or otherwise use bounded per-lane buffering.

```

</details>

admin-service/permissions.go
```diff
@@ -297,63 +295,82 @@ func (h *Handler) createPermissions(c *gin.Context) {
...
+	dests := h.remoteDests
+	if h.publishInbox == nil || len(dests) == 0 {
 		return nil
 	}
+	// The request ctx has no deadline and dies if the client disconnects — neither fits:
+	// the caller's local work is done (create's ledger write already committed; resync
+	// writes nothing), so re-delivery runs to completion on its own budget
+	// (cfg.FanoutTimeout), shared by every lane. A destination that doesn't land within
+	// it is reported like any other unacknowledged publish.
+	ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), h.cfg.FanoutTimeout)
```
julianshen: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**The fanout deadline does not account for earlier request work.**

The 30s fanout budget starts only after the create transaction/audit work or resync read/grouping has completed, while the 40s HTTP WriteTimeout is measured from the request. Any earlier work taking more than 10s—especially an uncapped bulk request—allows fanout to finish after net/http has already closed the connection, contradicting the guarantee that syncFailures remains deliverable. Cap fanout by the remaining whole-request write budget, with response-writing margin, rather than validating only that FANOUT_TIMEOUT is individually below WriteTimeout.

<details>

```
In `admin-service/permissions.go` around line 330: The 30s fanout budget starts only after the create transaction/audit work or resync read/grouping has completed, while the 40s HTTP WriteTimeout is measured from the request. Any earlier work taking more than 10s—especially an uncapped bulk request—allows fanout to finish after net/http has already closed the connection, contradicting the guarantee that syncFailures remains deliverable. Cap fanout by the remaining whole-request write budget, with response-writing margin, rather than validating only that FANOUT_TIMEOUT is individually below WriteTimeout.

```

</details>

admin-service/permissions.go
nalydadad: Fixed in b649e115 — createPermissions/resyncPermissions now pin one absolute per-request deadline at handler entry (`requestBudget = httpWriteTimeout - 2s` response margin) and publishPermissionFanout takes the earlier of that and `now + FANOUT_TIMEOUT` before severing client cancellation, so the fanout can only spend what the local work left of the write window; an already-spent budget reports every destination without publishing. Tests: "fanout deadline is capped by the whole-request budget" and "an already-spent request budget reports every destination". The shutdown half of this thread is answered on the config.go thread.

admin-service/permissions.go
nalydadad: Fixed in b649e115 — same change as the thread on line 332: fanout deadline = min(request deadline pinned at handler entry, now + FANOUT_TIMEOUT).

admin-service/permissions.go
nalydadad: Not changing this round. The retained set is bounded by real accounts, not by the caller: resync only emits accounts that exist with a stored state (`st == nil` is skipped) and create rejects unknown accounts. Each payload is ~200–300 B, so even 100k distinct-state accounts is ~30 MB transient — the same order as the `perms` map the handler already holds for those accounts. The endpoint is behind `requireAdmin`. Happy to add a hard cap on resync `accounts` as a follow-up if you want a belt-and-braces bound.

nalydadad: See the reply on the OOM thread at this line: bounded by real accounts. The marshal loop is CPU-only and finishes in ms even at 100k accounts, so a ctx check between marshals doesn't change the outcome.

admin-service/permissions.go
nalydadad: Duplicate of the OOM thread at line 349 — same reasoning: bounded by real accounts with a stored state, same order of memory as the `perms` map already held.

admin-service/permissions.go
nalydadad: Agreed it's a hole, but it isn't introduced here — `ALL_SITE_IDS`/`SITE_ID` are consumed unvalidated by room-service, room-worker and outbox-worker too. Validating only in admin-service would be inconsistent; the right fix is a shared `pkg/subject` token validator applied at every service's config load. Out of scope for this PR — follow-up.

admin-service/config.go
nalydadad: Known and documented on the field. Fixing it means either raising graceful shutdown above the request lifetime (needs the K8s grace period > 30s — an ops decision — and breaks the repo rule shutdown 25s < grace 30s) or capping FANOUT_TIMEOUT below ~23s, which reverses the 30s explicitly requested on #265. The failure mode is safe: the ledger already committed, and resync re-delivers anything a SIGTERM cut short. Leaving as-is unless you'd rather move the K8s grace period.

nalydadad:
answer clar_yr4pdp74mriffkprxtdxjav3o6: None today — createUser only requires a non-empty account, and SSO-migrated accounts carry whatever the old stack allowed. The 64-char figure in TestFanoutChunk_FitsBrokerMaxPayload is a sizing assumption, not an enforced bound; a chunk only breaches 128KB at >~110-char average account IDs. Not changing in this PR (owner decision); if we want it enforced, the cheap fix is a length cap in createUser rather than byte-sized chunking.

mliu33 (approved):

PR #308, opened 2026-08-17
fix(loadgen): repair five soak read lanes that were rejected on every request

flywindy:
## What this fixes

Five soak lanes shipped in #295 never measured what they claimed. All five surfaced in `loadgen_soak_errors_total` as what looked like service faults, but the fault was on the load generator's side in every case — and not always the same kind of fault, which the fixes have to respect.

**Three were rejected outright**, on every request:

| Lane | What it sent | Why the handler refused it |
|---|---|---|
| `subscription_list` | no body | `natsrouter.Register` decodes one; `null` → empty `type` → *unknown subscription type* |
| `room_state_read` | no body | same subject, same rejection — and this one is the room-state observer's RPC source |
| `user_subscription_channels` | page fields only | requires exactly one of `membersContain` / `accountNames` |

**Two were served, and could only ever answer not-found:**

| Lane | What it asked | Why the answer was empty |
|---|---|---|
| `read_receipt_list` | as the room's held account | room-service serves receipts only to the message's own sender, so anything but a chance collision was refused |
| `user_subscription_dm` | two random accounts | a topology has `dmCount` DM rooms among `C(active,2)+` possible pairs, so a random pair almost never shares one |

## Why the observer fix is not the obvious one

Repairing `room_state_read` by adding the body would have made things **worse**.

`subscription.list` is paged at `SUBSCRIPTION_DEFAULT_LIMIT` (40). `verifyMute` and `verifyRead` both initialise `rpc = mismatch` and only leave that state if the room appears in the response. While the request was being rejected outright the path was unreachable — `err != nil` kept the verdict at `unknown`. Making the request valid makes it live, and a room that merely fell off the first page is then recorded as `room_service / mismatch`: the impossible-state bucket, invented by the observer itself.

At `channelMembers: 500` (supported — `largeRoomThreshold` is 500) an account holds ~75 subscriptions, so every mute and read verification for the whole run would have emitted fabricated corruption evidence.

It now uses `subscription.getByRoomID` — 0-or-1 for the room asked about, no page to fall off, and an empty answer genuinely means absent. `muted` and `lastSeenAt` both survive that response's assembly path (`buildListItems` wraps the embedded `*model.Subscription`, so the base fields flatten at marshal time), and its match predicate is the same `{u.account, roomId}` the loadgen's authoritative primary read uses — so the RPC source and the store fallback can no longer disagree structurally, which was the point.

## Why `getChannels` needed more than a body

Two ways to make this lane green while it measures nothing:

Passing the requester as its own `membersContain` is accepted, so the errors stop. But `FindChannelsByMembers` computes `dedupeStrings(members + account)`, which collapses a self-reference to a single element, and the `$size` gate then matches every channel the requester belongs to. You could delete the `$size` clause — a real correctness guard — and this lane would not move.

Naming a *random* other account is also accepted, but at `channelMembers: 100` two random accounts share a channel roughly 7% of the time, so an empty page becomes the lane's normal answer — indistinguishable from a broken query. That is the same argument this PR uses to justify indexing real DM rooms, so the channel read gets the same treatment: both co-members now come from one index over the topology's own rooms, keyed by room type. Rows with `isSubscribed=false` are excluded — a left room keeps its subscription record, and pairing on it would name a counterpart who no longer shares the room.

## Encryption preflight: its own gate and budget

The preflight's timeout was `max(30s, SOAK_PERSIST_GRACE+10s)`. `persistGrace` is not a local knob: it also decides when a message becomes eligible for mutation and read-back, when thread replies become readable, and when **both ledgers start verifying**. Buying preflight headroom there delays every verification in the run by the same amount, so a slow environment ends up proving less rather than waiting longer.

- `SOAK_ENCRYPTION_PREFLIGHT_TIMEOUT` (default `30s`) budgets the check on its own.
- `SOAK_ENCRYPTION_PREFLIGHT` (default `true`) turns it off where at-rest encryption is not what is being measured. **Refused when `SOAK_ENVIRONMENT` is `staging` or `production`** — in the Go validation and again in the chart, so `helm template` fails too.
- Skipping is not silent: startup warns, and `loadgen_soak_encryption_preflight` reports 0 so the artifact a reader sees afterwards still says the encrypted write path was never proven. `0` means *not proven* generally — skipped, in flight, failed, or a loadgen mode that never runs the check — so alert on a sustained 0 during soak, not on any 0. The passing path raises it to 1, asserted by the integration test rather than left to the one code path nothing exercised.

## Two contract tests, and what they do not cover

**`TestSoakRPCCallSites_CarryABodyWhenTheHandlerDecodesOne`** walks the repo for `natsrouter.Register` / `RegisterNoBody`, maps each subject builder to whether its handler decodes a body, then brace-matches every `soakRPCRequest` literal and asserts a `Register`-ed subject is never addressed without one.

It checks that a body is **present**, not that its shape satisfies the handler — that covers two of the five bugs above. The other three need their own assertions and have them. The header says so rather than overclaiming.

**`TestSoakConfigMap_SetsEveryConfiguredKnobAndNoOthers`** pins the chart's `SOAK_` keys and `soakConfig`'s env tags to each other in both directions. It is what caught the missing chart wiring for both new knobs — the same class of defect as the six lane rates that were once configured, reported as the run's target, and never sent.

Both fail loudly if their own scan comes back empty, and the RPC sweep anchors on two named builders (one Pattern-suffixed, one not) so a matcher that quietly stops resolving cannot pass as a clean sweep.

## Verification

Green locally: `make lint` 0 issues, `go test -race ./tools/loadgen/...`, `go vet -tags integration`, `make sast-gosec`, scoped soak coverage **90.37%** (gate 90).

Chart, checked statically because `helm` is unavailable in this environment: `values.schema.json` validates both `values-validation.yaml` and `values-local.yaml` with 0 errors, all 7 templates parse, and the new staging/production guard evaluates correctly against both value files.

**Not verified here**, worth a reviewer's eye:

- `make validate-loadgen-k8s` — `helm` is not installed and `get.helm.sh` is blocked (403) by the outbound proxy
- `make coverage-loadgen-soak`'s ≥80% aggregate gate — needs Docker; the ≥90% core gate above is a lower bound measured without integration tests
- `make sast-vuln` — `vuln.go.dev` is blocked (no dependency changes in this PR)

## Not fixed here

`docker-compose.yml` gains the two new keys, but seven pre-existing `SOAK_` knobs remain absent from it (`SOAK_RUN_MODE`, the heartbeat pair, the teardown batch trio, `SOAK_LARGE_ROOM_THRESHOLD`). The new configmap test pins only the Helm surface, so compose can still drift. Left out to keep this PR to the defects it is about.

**Replication lag can now reach the RPC verdict.** user-service defaults to `MONGO_READ_PREFERENCE=secondaryPreferred`, so `getByRoomID` reads a secondary while the loadgen's authoritative fallback reads the primary. In `verifyRead` a lagging secondary can classify as `mismatch` in the `room_service` source counter. The final verdict is still the primary's, and `verifyName`/`verifyMember` already had this exposure — this PR only makes it reachable for mute and read. Worth knowing before anyone reads that bucket as gospel.

`user_profile_get` / `user_status_get` returning `not_found` against a live environment is still open. The request contract is correct — the body sends `name` as an account, the repo looks it up by account, and both sides filter active users the same way — so it is being investigated as a data or deployment question rather than patched here.

## Summary by CodeRabbit

* **New Features**
  * Added an encryption preflight check before load generation, enabled by default with a configurable timeout.
  * Added monitoring to report successful encryption verification.
  * Added deployment configuration for controlling preflight behavior.

* **Bug Fixes**
  * Improved read receipts, subscription checks, and direct-message/channel request selection.
  * Added safeguards against invalid or incomplete room and membership data.

* **Validation**
  * Encryption preflight can only be disabled outside staging and production environments.
  * Added validation for timeout settings and deployment configuration consistency.


flywindy:
@coderabbitai review

Out of draft. CI is green on `52a07f8` across all five checks.

Two things worth your attention specifically:

1. `SubscriptionFor` in `tools/loadgen/soak_roomread.go` moved from the paged `subscription.list` to `subscription.getByRoomID`. The verifiers in `soak_roomverify.go` treat "room absent from the response" as a `mismatch`, so a paged list would have let a paged-out room be recorded as corruption evidence. Please check I have not missed a field the verifiers read that the new response assembles differently.

2. `soakUserRoomPairs` in `tools/loadgen/soak_userread.go` indexes DM and channel co-members from the topology, restricted to active requesters. It is built once in the constructor and read under the same mutex as the rng — worth confirming that holds for the lane's concurrent callers.

---

tools/loadgen/soak_userread.go
```diff
@@ -81,10 +93,76 @@ func newSoakUserReader(
...
+	}
+	members := make(map[string][]string)
+	for i := range topology.Subscriptions {
+		subscription := &topology.Subscriptions[i]
+		if subscription.RoomType != roomType || subscription.User.Account == "" {
+			continue
+		}
+		members[subscription.RoomID] = append(
+			members[subscription.RoomID], subscription.User.Account,
+		)
```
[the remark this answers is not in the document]
flywindy: Good catch — fixed in `ac07cdd`.

This was a real defect and the reasoning is exactly right: a left room keeps its subscription row with `isSubscribed=false`, so pairing on it names a counterpart who no longer shares the room, and the read comes back legitimately empty. That is the same "an empty answer is indistinguishable from a broken query" failure the index was added to prevent, so it defeated its own purpose.

It also disagreed with `soak_roomstate.go:243`, which has always filtered this field when building the member pool. Two indexes over the same topology should not have different ideas of who is in a room.

Added a red test first (`TestSoakUserReader_RoomPairsIgnoreUnsubscribedMemberships`) — a DM whose second participant has `IsSubscribed: false` produced a request before the fix and now correctly skips.

---

tools/loadgen/soak_integration_test.go
```diff
@@ -131,14 +132,23 @@ func TestSoakRunA_SeedFrontDoorReadBackAndTeardown(t *testing.T) {
...
 		sender,
 		selector,
 		replies,
-		0,
 	))
+	// The gauge is the only artifact a reader sees after the run, so the passing
+	// path has to actually raise it — not merely avoid an error.
+	assert.Equal(t, float64(1),
+		prommetrics.ToFloat64(preflightMetrics.SoakEncryptionPreflight),
+		"a verified encrypted write path must be recorded, not just logged")
```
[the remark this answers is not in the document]
flywindy: This one doesn't hold — the constraint is already there, on line 1 of the file:

```
$ head -3 tools/loadgen/soak_integration_test.go
//go:build integration

package main
```

The finding is anchored at lines 135–151, which is where my diff hunk is; the build tag is well above it and outside the hunk, so it looks absent from the diff alone.

Confirming it is actually in effect rather than just present: `go test ./tools/loadgen/` runs 0 of this file's tests and needs no MongoDB or NATS, while `go vet -tags integration ./tools/loadgen/` does compile it. In CI the two are separate jobs — `test` and `test-integration (tools)` — and only the latter provisions services.

Resolving as not applicable.

---

tools/loadgen/soak_rpc_registration_contract_test.go
```diff
@@ -0,0 +1,199 @@
...
+				return fs.SkipDir
+			}
+			return nil
+		}
+		if !strings.HasSuffix(path, ".go") || strings.HasSuffix(path, "_test.go") {
+			return nil
+		}
+		source, readErr := os.ReadFile(path)
+		if readErr != nil {
+			return readErr
```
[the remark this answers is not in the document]
flywindy: Taken — fixed in `ac07cdd`. Both the `WalkDir` traversal error and the `os.ReadFile` error now name the path and the scan step:

```go
return fmt.Errorf("walk %q while scanning registrations: %w", path, err)
return fmt.Errorf("read %q while scanning registrations: %w", path, readErr)
```

Worth doing here specifically because this test walks the whole repo. A bare `permission denied` from `require.NoError` gives no way to tell which file broke the sweep, and a broken sweep is the one failure mode this test cannot afford to be vague about — it is the guard against a matcher that silently stops resolving.

---

tools/loadgen/soak_config.go
```diff
@@ -236,6 +247,9 @@ func validateSoakConfig(cfg *soakConfig, cassandraKeyspace string) error {
 	if err := validateSoakRoomLaneConfig(cfg); err != nil {
 		return err
 	}
+	if err := validateSoakEncryptionPreflight(cfg); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
flywindy: Not taking this one, and I want to be explicit about why rather than quietly skipping it.

`validateSoakConfig` dispatches to six sub-validators and returns every one of them bare — `validatePositiveRate` (×2, lines 172/175), `validateNonNegativeRate` (194), `validateSoakRoomLaneConfig` (247), `validateSoakTeardownConfig` (309), and mine (251). Wrapping only the line this PR happens to touch makes it the single inconsistent one in the function.

The guideline's stated purpose is "describe what the current function was doing". These validators already return complete, self-describing messages that name the offending variable:

```
SOAK_ENCRYPTION_PREFLIGHT_TIMEOUT must be greater than zero
SOAK_ENCRYPTION_PREFLIGHT cannot be disabled when SOAK_ENVIRONMENT is "staging"
```

Wrapping produces `validate encryption preflight configuration: SOAK_ENCRYPTION_PREFLIGHT_TIMEOUT must be greater than zero`, which restates in prose what the variable name already says. This error is printed once at startup and read by a human; the prefix costs clarity rather than adding a trail.

Contrast with the other error finding in this review, which I did take: `SubscriptionFor` returns a bare transport error (`request timeout`) from a call shared by two verifiers, so nothing identifies *which room's* state could not be read. That wrap adds information a reader does not otherwise have. This one doesn't.

If the team would rather have literal compliance across the board, that is a reasonable call — but it should be one sweep over all six, not one line, and it belongs in its own change rather than buried in a lane-repair PR. Happy to open that separately if you want it.

---

flywindy:
@coderabbitai review

`ac07cdd` addresses the last round. CI is green on it across all five checks.

What changed since `52a07f8`:

- `soakUserRoomPairs` now skips `IsSubscribed == false` rows, with `TestSoakUserReader_RoomPairsIgnoreUnsubscribedMemberships` as the regression guard — it was written first and did fail before the filter.
- `SubscriptionFor` wraps its RPC error with the room ID, since the mute and read verifiers share that call.
- The registration scan wraps traversal and read failures with the path.

The two findings I did not act on are answered in their threads: the `//go:build integration` tag was already on line 1 of the file, and the bare `err` in `validateSoakConfig` matches all six sibling sub-validators whose messages already name the offending `SOAK_*` variable.

---

julianshen:
**6 findings · 3 unique issues · 3 inline comments posted** — 🟠 6 warning

### Related work

- [PR #295](https://github.com/hmchangw/newchat/pull/295) — feat\(loadgen\): expand the continuous soak with room, member, presence, read-receipt, user and search lanes (merged)

<details>

* `tools/loadgen/deploy/k8s/templates/_helpers.tpl`
* `tools/loadgen/metrics.go`
* `tools/loadgen/metrics_test.go`
* `tools/loadgen/soak_configmap_contract_test.go`
* `tools/loadgen/soak_integration_test.go`
* `tools/loadgen/soak_preflight_config_test.go`
* `tools/loadgen/soak_rpc_registration_contract_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

julianshen (commented):
tGD inline review

tools/loadgen/soak_main.go
```diff
@@ -1127,21 +1132,44 @@ func warmSoakPinnedCatalog(
...
+			"encryptionPreflight", false,
+			"consequence", "this run does not verify the encrypted write path",
+		)
+		return nil
+	}
 	target, content := selector.nextSend()
 	pending, err := sender.Publish(ctx, target, content)
 	if err != nil {
 		return fmt.Errorf("publish encrypted front-door probe: %w", err)
 	}
```
julianshen, line 1176: _🏷️ timeout_ | _🟠 Warning_ | _`cassandra`, `performance-and-concurrency`, `silent-failure`, `tgd-review` (4 rules)_

**The preflight timeout does not cover the network publish.**

The configured timeout is created only after `sender.Publish(ctx, ...)` returns, and the publish receives the long-lived workload context instead. Under NATS/front-door backpressure or an outage, publishing can therefore block beyond `SOAK_ENCRYPTION_PREFLIGHT_TIMEOUT`—potentially until the 72-hour run context expires—so the soak workload never starts. Create `probeCtx` before publishing, pass it to `sender.Publish`, and use that same context while awaiting persistence.

<details>

- **`cassandra`**:1172: The configured SOAK_ENCRYPTION_PREFLIGHT_TIMEOUT context is created only after sender.Publish returns. During a front-door or NATS outage, that publish therefore runs under the workload's parent context and can consume far more than the advertised preflight budget before the Cassandra verification even starts. Create the timeout context before publishing and pass probeCtx to sender.Publish as well as the subsequent persistence check.
- **`silent-failure`**:1171: The probe is published with the parent `ctx` before `probeCtx` is created. If `sender.Publish` stalls, `SOAK_ENCRYPTION_PREFLIGHT_TIMEOUT` never expires that operation; startup can remain blocked until the much longer run context is canceled instead of surfacing a preflight timeout and exiting. Create `probeCtx` immediately after the enabled check and pass it to `sender.Publish` as well as the reply and persistence checks.
- **`tgd-review`**:1173: The timeout context is created only after sender.Publish returns, and Publish receives the outer workload context. If publishing blocks, SOAK_ENCRYPTION_PREFLIGHT_TIMEOUT does not bound the preflight as documented, potentially delaying workload startup until the much longer outer context expires. Create probeCtx before Publish and pass it to both Publish and the observation loop.

</details>

<details>

```
In `tools/loadgen/soak_main.go` around line 1171: The configured timeout is created only after `sender.Publish(ctx, ...)` returns, and the publish receives the long-lived workload context instead. Under NATS/front-door backpressure or an outage, publishing can therefore block beyond `SOAK_ENCRYPTION_PREFLIGHT_TIMEOUT`—potentially until the 72-hour run context expires—so the soak workload never starts. Create `probeCtx` before publishing, pass it to `sender.Publish`, and use that same context while awaiting persistence.

```

</details>

tools/loadgen/soak_userread.go
```diff
@@ -81,10 +93,80 @@ func newSoakUserReader(
 	for i := range topology.Rooms {
 		reader.rooms = append(reader.rooms, topology.Rooms[i].ID)
 	}
+	reader.dmPairs = soakUserRoomPairs(topology, reader.accounts, model.RoomTypeDM)
```
julianshen, line 96: _🏷️ consistency_ | _🟠 Warning_ | _`distributed-system`_

**Channel co-member pairs become stale after membership mutations.**

`channelPairs` is built once from the startup topology and is never refreshed, while the configured soak workload mutates memberships. A reachable timeline is: accounts A and B initially share their only channel, this index records (A,B), the member-mutation lane removes B, and subsequent `SubscriptionChannels` calls continue sending B as `membersContain`. User-service then legitimately returns an empty page, which the read lane records as a successful request, recreating the false-health behavior this change intends to prevent. Store the originating room with each pair and invalidate/update pairs after successful membership mutations, or select only from membership state maintained dynamically by the workload; add a test covering removal after reader construction.

<details>

```
In `tools/loadgen/soak_userread.go` around line 96: `channelPairs` is built once from the startup topology and is never refreshed, while the configured soak workload mutates memberships. A reachable timeline is: accounts A and B initially share their only channel, this index records (A,B), the member-mutation lane removes B, and subsequent `SubscriptionChannels` calls continue sending B as `membersContain`. User-service then legitimately returns an empty page, which the read lane records as a successful request, recreating the false-health behavior this change intends to prevent. Store the originating room with each pair and invalidate/update pairs after successful membership mutations, or select only from membership state maintained dynamically by the workload; add a test covering removal after reader construction.

```

</details>

tools/loadgen/soak_userread.go
```diff
@@ -81,10 +93,80 @@ func newSoakUserReader(
...
+		participants := members[topology.Rooms[i].ID]
+		// A DM room is always exactly two accounts; for a channel the first two
+		// are enough. Fewer than two cannot name a counterpart at all.
+		if roomType == model.RoomTypeDM && len(participants) != 2 {
+			continue
+		}
+		if len(participants) < 2 || participants[0] == participants[1] {
+			continue
+		}
+		for side, requester := range participants[:2] {
```
julianshen: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Channel pairing can ignore active members after the first two.**

The code slices participants to the first two before filtering requesters against the active-account set. A channel ordered as [borrowed-a, borrowed-b, active-a] therefore contributes no pair, even though active-a has valid co-members; if this ordering occurs across the topology, SubscriptionChannels skips every request. Iterate participants until up to the desired number of active requesters have been paired with a distinct member, rather than limiting the candidate requesters first.

<details>

```
In `tools/loadgen/soak_userread.go` around line 158: The code slices participants to the first two before filtering requesters against the active-account set. A channel ordered as [borrowed-a, borrowed-b, active-a] therefore contributes no pair, even though active-a has valid co-members; if this ordering occurs across the topology, SubscriptionChannels skips every request. Iterate participants until up to the desired number of active requesters have been paired with a distinct member, rather than limiting the candidate requesters first.

```

</details>

tools/loadgen/soak_main.go
flywindy: Taken — fixed in `263599a`.

The contract was untrue as written: `SOAK_ENCRYPTION_PREFLIGHT_TIMEOUT` bounded only the wait for evidence, and the send it is named for ran under the run-scoped context. The budget now starts before the publish:

```go
probeCtx, cancel := context.WithTimeout(ctx, cfg.Timeout)
defer cancel()
target, content := selector.nextSend()
pending, err := sender.Publish(probeCtx, target, content)
```

One correction to the severity, for the record rather than to argue the fix. In the soak wiring the publisher is `InjectFrontdoor`, which lands on core `nc.Publish` — that buffers into the pending-write queue and returns rather than blocking on the wire, so the 72-hour stall the finding describes is not reachable through today's sender. What made this worth fixing anyway is that the guarantee is stated in the config's own doc comment and enforced by nothing: the publisher is an interface, and the next implementation behind it (a JetStream publish that waits for the ack, say) inherits the unbounded context silently.

Red-first, and the test asserts the property rather than the line ordering — a stub publisher records `ctx.Deadline()`, which is empty under the old code:

```
--- FAIL: TestRunSoakEncryptionPreflight_PublishesUnderTheConfiguredTimeout
    Messages: the probe publish ran under an unbounded context, so a stalled
              publish would outlive SOAK_ENCRYPTION_PREFLIGHT_TIMEOUT
```

The persistence check already used `probeCtx`, so it needed no change.

---

tools/loadgen/soak_userread.go
flywindy: Taken — fixed in `263599a`. The analysis is right, including the reason it matters: which members a room lists first is an ordering accident, so truncating before the active filter made lane coverage depend on load order.

Worth adding why it was not visible in practice, because that is the part that made it dangerous. `soak_topology.go:131` seeds every channel with an anchor drawn from the active set *first*, so at head the active member is usually participant 0 — but `LoadTopology` re-sorts subscriptions by `_id`, so on the restart path that placement is not preserved. The truncation was therefore an implicit dependency on a property one of the two construction paths does not maintain.

The scan now looks for active requesters anywhere in the list and pairs each with a distinct co-member, capped at `soakUserPairsPerRoom = 2` so the index stays linear:

```go
paired := 0
for _, requester := range participants {
    if paired == soakUserPairsPerRoom { break }
    if _, ok := active[requester]; !ok { continue }
    peer, ok := soakUserPeerFor(participants, requester)
    ...
}
```

Red-first with exactly your fixture — a channel ordered `[borrowed-a, borrowed-b, active-a]` — which sent no request before the change and now sends one from `active-a`.

One thing your suggestion does not mention that the rewrite had to preserve: the old code's `participants[0] == participants[1]` guard was the only thing stopping a room whose rows all name one account from pairing that account with itself, which is the same degenerate `membersContain: <self>` query that broke `getChannels` in the first place. `soakUserPeerFor` returning `false` covers it now, and `TestSoakUserReader_ChannelPairsSkipRoomsWithASingleDistinctMember` locks it — verified by stubbing the distinctness check to `true` and watching the test fail.

---

tools/loadgen/soak_userread.go
flywindy: Not taking this one — the timeline is unreachable, because the member lane and the pair index draw from disjoint sets of accounts by construction. Evidence, since this is the kind of claim that deserves line numbers rather than assertion.

**The member lane can only remove accounts it added.** `nextMemberIntent` pops its target from one of two slices (`soak_roomstate.go:430-436`):

```go
case len(room.members) > 0:
    account, room.members = room.members[0], room.members[1:]
case len(room.available) > 0:
    account, room.available = room.available[0], room.available[1:]
```

`room.members` has exactly three append sites — `soak_roomstate.go:463`, `:475`, and `:656` — and all three append an account that came out of `room.available` earlier in the cycle (a completed add, a rolled-back remove, and a resolved probe respectively). Nothing else ever puts an account into `room.members`.

**`room.available` excludes everyone the pair index can see.** It is built only from `topology.BorrowedUsers`, and skips any account already subscribed to that room (`soak_roomstate.go:284-287`):

```go
if _, member := subscribed[state.id][user.Account]; member {
    continue
}
state.available = append(state.available, user.Account)
```

where `subscribed` is populated only from rows with `IsSubscribed == true` (`:243`).

**`channelPairs` is built only from those same seeded rows** — `soakUserRoomPairs` skips `!subscription.IsSubscribed`, so every account it can name is in `subscribed[room]` and therefore was never eligible for `room.available`.

So a paired account is a seeded member, and the member lane's universe is exactly the borrowed accounts that were *not* seeded members. There is no account in both, and the "the member-mutation lane removes B" step cannot happen for any B the index holds.

That said, one part of the finding is worth writing down rather than dismissing: the safety here rests on an invariant that lives in two files and is asserted in neither. If a future change lets the member lane seed from subscribed accounts, or lets `soakUserRoomPairs` index borrowed ones, staleness becomes real and nothing fails. I have recorded the coupling in the `soakUserRoomPairs` doc comment; a test pinning the disjointness would be better still, but it belongs with whichever change breaks the assumption rather than pre-emptively here — I would rather not add a test that passes for a reason the code does not state.

Happy to reopen if you can point to a path into `room.members` I have missed.

---

tools/loadgen/soak_preflight_config_test.go
```diff
@@ -0,0 +1,135 @@
...
+		selector,
+		make(chan soakSendObservation),
+	)
+	require.Error(t, err)
+
+	require.True(t, publisher.hasDeadline,
+		"the probe publish ran under an unbounded context, so a stalled publish "+
+			"would outlive SOAK_ENCRYPTION_PREFLIGHT_TIMEOUT")
+	assert.WithinDuration(t, time.Now().Add(40*time.Millisecond), publisher.deadline,
+		2*time.Second, "the publish must share the preflight's budget")
```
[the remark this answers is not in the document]
flywindy: Taken — fixed in `c7b9fc0`. This one was worse than "minor": the assertion was vacuous.

`runSoakEncryptionPreflight` waits out its entire budget before returning, so by the time `time.Now()` is evaluated the deadline has essentially arrived. `deadline - now` ≈ `timeout - elapsed` ≈ 0 for *any* finite timeout, and a 2s tolerance swallows the rest. It proved a deadline existed — which the `hasDeadline` require one line above already proved — and nothing about whose budget it was.

Fixed along the lines you suggested, with the anchor recorded inside `Publish`, and bounded from both sides so a budget that is too small fails too:

```go
budget := publisher.deadline.Sub(publisher.publishedAt)
assert.LessOrEqual(t, budget, soakPreflightTestTimeout, ...)
assert.Greater(t, budget, soakPreflightTestTimeout/2, ...)
```

I used `LessOrEqual`/`Greater` rather than `InDelta` because the upper bound here is exact, not approximate: the deadline is `t0 + timeout` and the publish starts at `t0 + ε`, so the budget it sees can only be ≤ the configured value. A symmetric delta would allow overshoot that the code cannot legitimately produce. The timeout also moved to a named `soakPreflightTestTimeout = 200ms` so the assertions and the config cannot drift apart, and so the wiring cost between creating the deadline and entering `Publish` stays comfortably inside the lower bound.

Verified the new form discriminates by giving the preflight ten times its configured budget:

```
Error: "1.999935466s" is not less than or equal to "200ms"
Messages: the publish was given a deadline from some other budget, not the preflight's own
```

The previous assertion accepted exactly that mutation, which is the point.

---

flywindy:
@coderabbitai review

The last automatic run hit the hourly limit before it could read the current head. Two commits landed since the range you last reviewed (`263599a`):

- `b4784f2` — comment only: records why `soakUserRoomPairs` needs no invalidation on membership mutations (the member lane and the pair index draw from disjoint account sets).
- `c7b9fc0` — strengthens the preflight deadline assertion you flagged, measuring the budget from the instant `Publish` is entered.

CI is green on `c7b9fc0` (lint, test, sast, test-integration, prewarm).

---

tools/loadgen/soak_userread.go
flywindy: Independent confirmation, so this is no longer just my reading of the code.

CodeRabbit reviewed the current head with its own repo scan and reached the same conclusion without being pointed at my argument:

> `soakUserRoomPairs` uses only subscribed memberships. `newSoakRoomStatePool` builds `room.available` from borrowed accounts that are not subscribed to that same room. The member lane only removes a candidate after it previously added that candidate. It cannot remove an account in the pair index for that room.

Two independent analyses of the same code agreeing on the disjointness is as much as this finding can be settled by, so I am resolving it. If you can point to a path into `room.members` that neither of us traced, reopen it and I will take it.

The one thing I do agree with in the original finding survives in the code: the invariant spans two files and nothing asserts it, so it is now written down at the pair index (`b4784f2`) where a future change that breaks it would be read.

---

yenta (approved):

PR #310, opened 2026-08-18
fix(history-service): decrypt pinned messages so edited pins aren't empty

nalydadad:
## Problem

`chat.user.{account}.request.room.{roomID}.{siteID}.msg.pinned.list` returned `msg: ""` for any pinned message that had been edited.

`pinned_messages_by_room` carries `enc_payload`/`enc_meta`, and the at-rest edit path (`write.go`'s `editPinnedMsgEncrypted`) moves the body into `enc_payload` while nulling `msg`, `attachments`, `card` and `card_action` on the pinned mirror. But the pinned-list read omitted both encrypted columns from its projection and never called `decryptIfNeeded` — so an edited pin came back with an empty body and no attachments. `ATREST_ENABLED` defaults to `true`, so this reproduces on any default deployment.

Sibling readers (`messages_by_room.go`, `messages_by_id.go`, `thread_messages.go`) already select and decrypt these columns; the pinned read was the only one that didn't.

## Fix

`history-service/internal/cassrepo/pin.go` only:

- add `enc_payload, enc_meta` to `pinnedColumns`
- `GetPinnedMessages` now scans through the shared `scanMessagesUpTo` helper, which structScans and decrypts each row (the same helper `GetThreadMessages` uses with this exact QueryBuilder/Fetch shape)
- a decrypt failure fails the whole request, matching the message-history readers; the error wrap is renamed `scan …` → `read …` since it now covers decrypt too
- `GetAllPinnedMessages` deliberately keeps reading undecrypted — its only caller (`enforcePinLimit`) reads just `MessageID`/`PinnedAt`

No change to any function signature, the `MessageReader` interface, generated mocks, the wire schema, or the Cassandra DDL — so no `make generate` and no `docs/client-api.md` schema edit.

## Behavior change reviewers should know about

Routing the pinned read through `scanMessagesUpTo` also makes it **honor `limit`**. The old loop drained the gocql iterator, and gocql auto-pages by default (`conn.go:1450` sets `iter.next` whenever more pages exist; `DisableAutoPage` is never called here) — so a request with `limit: 3` against a 10-pin room previously returned all 10 messages with `hasNext: false`. It now returns 3 plus a `nextCursor`.

That is the contract `docs/client-api.md` already documents (`limit` = max per page, `hasNext` = more pages exist) and matches `GetThreadMessages`, so this corrects a latent pagination bug rather than introducing one. Pins are capped at `MAX_PINNED_PER_ROOM` (default 10) and the default limit is far above that, so only a caller explicitly sending a small `limit` is affected. The loadgen soak lane already follows the cursor correctly.

One more consequence of decrypting here: if `ATREST_ENABLED` is turned off while encrypted pinned rows exist, `msg.pinned.list` now fails outright instead of returning blank bodies — same as message history already does in that situation.

## Tests

New integration test `TestRepository_GetPinnedMessages_EditedEncryptedPin` (real Cassandra + Vault-backed cipher): seed an encrypted message → `GetMessageByID` → `PinMessage` → `GetMessageByID` → `UpdateMessageContent` → `GetPinnedMessages`, asserting the edited body, the restored attachments, and that no ciphertext escapes the repository.

Written test-first and watched fail on the unfixed code:

```
--- FAIL: TestRepository_GetPinnedMessages_EditedEncryptedPin
    expected: "edited body"          actual: ""
    expected: [][]uint8{"att-1.bin"} actual: [][]uint8(nil)
```

Verification after the fix:

- `make lint` → `0 issues.`
- `make test SERVICE=history-service` → all packages ok
- `make test-integration SERVICE=history-service` → all packages ok (`cassrepo` 150s)

## Known gap, deliberately out of scope

`insertPinnedMsg` still writes the pin mirror in **plaintext** at pin time even when at-rest encryption is on — so a pinned message's body is stored unencrypted in `pinned_messages_by_room` until it is edited. That is the inconsistency that produced this bug (plaintext on pin, ciphertext on edit). Fixing it means changing the pin write path and is left for a separate change.

### DIFF
--- a/history-service/internal/cassrepo/pin.go
+++ b/history-service/internal/cassrepo/pin.go
@@ -33,7 +33,10 @@ const (
 	pinnedColumns = "room_id, pinned_at, message_id, sender, " +
 		"msg, mentions, attachments, card, card_action, quoted_parent_message, " +
 		"visible_to, deleted, type, sys_msg_data, site_id, edited_at, " +
-		"updated_at, pinned_by, created_at, tshow, thread_parent_id, thread_parent_created_at"
+		"updated_at, pinned_by, created_at, tshow, thread_parent_id, thread_parent_created_at, " +
+		// An at-rest edit rewrites the pinned row as ciphertext (msg nulled), so
+		// the read must carry these or an edited pin comes back empty.
+		"enc_payload, enc_meta"
 
 	pinnedByRoomQuery = "SELECT " + pinnedColumns + " FROM pinned_messages_by_room WHERE room_id = ?"
 )
@@ -120,11 +123,15 @@ func (r *Repository) GetPinnedMessages(ctx context.Context, roomID string, pageR
 			if !found {
 				break
 			}
+			if dErr := r.decryptIfNeeded(ctx, &m); dErr != nil {
+				scanErr = dErr
+				return
+			}
 			rows = append(rows, m)
 		}
 	})
 	if scanErr != nil {
-		return Page[models.Message]{}, fmt.Errorf("scan pinned message row for room %s: %w", roomID, scanErr)
+		return Page[models.Message]{}, fmt.Errorf("read pinned message row for room %s: %w", roomID, scanErr)
 	}
 	if err != nil {
 		return Page[models.Message]{}, fmt.Errorf("query pinned messages for room %s: %w", roomID, err)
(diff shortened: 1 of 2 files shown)


mliu33 (approved):
Thanks a lot

PR #311, opened 2026-08-18
search-sync-worker: pipeline ES flushes, pace failed-write retries, close ordering gaps

hmchangw:
Started as an investigation into search-sync-worker running slowly with a growing backlog of pending messages. Four things came out of it: one throughput fix, one retry-amplification fix, and two correctness gaps found along the way.

> **Rebased onto `de777a1`.** `#344` landed on main in the meantime and independently moved this worker's flush path onto `pkg/jsretry`, so the retry commit below is now narrower than when it was written — see §2.

## ⚠️ Action required before this reaches production

Depth 2 at `BULK_BATCH_SIZE=500` needs `CONSUMER_MAX_ACK_PENDING=1500`, and the shared default is **1000**. `CONSUMER_MAX_ACK_PENDING` lives in `stream.ConsumerSettings`, so this service can&#39;t carry its own default via struct tags — the dev compose sets 1500, production has to set it too.

If it isn&#39;t set, the worker logs the coupling warning at startup and the size-based flush can never fire, so every batch falls back to the 5s timer undersized — **worse than before this PR**. Two ways to close it:

| Option | Effect |
|---|---|
| `CONSUMER_MAX_ACK_PENDING=1500` | keeps 500-doc batches, more messages in flight |
| `BULK_BATCH_SIZE=250` (no ack change) | 750 in flight (fewer than today&#39;s 1000), two concurrent 250-doc requests, halves the nak-all blast radius |

The `BULK_BATCH_SIZE` default is deliberately left at 500 rather than lowered here, so existing measurements stay comparable.

## 1. The consumer loop was fully serial

`runConsumer` did `Fetch → build → Bulk` with no overlap: nothing was fetched during the ES round-trip, nothing indexed during the fetch. It was the only JetStream worker in the repo with no overlap at all — the others run `cons.Messages()` with a `MAX_WORKERS` semaphore.

`Handler.Flush` is split into `Take` (detach the buffer, no I/O) + `FlushBatch`, so a batch can be handed to a background flusher while the loop keeps buffering. `PIPELINE_DEPTH` (default **2**) sets how many bulk requests stay in flight.

Batch time goes from `fetch + build + ES` to `max(fetch + build, ES)`, then depth multiplies the ES side.

**On the expected gain:** measured `fetch+build` is ~10–30ms against ES bulk latency of 50–500ms, so `b/a ≈ 4–10`. Depth 1 alone is worth roughly 10–25%, not 2× — the structural win is real but modest when ES dominates, which is why depth is configurable. Depth 2 rather than higher because it multiplies **per collection and per pod** (five collections at depth 2 is already ten concurrent bulk requests per replica), and because it&#39;s the most informative step: if doubling concurrency doesn&#39;t move throughput, deeper won&#39;t either.

Rollback is `PIPELINE_DEPTH=1`, which reproduces the pre-PR behaviour exactly.

## 2. Failed writes retried instantly, amplifying the failure

A failed flush nakked with a bare `Nak()`, which asks JetStream for **immediate** redelivery. The consumer&#39;s `BackOff` schedule does not pace that path — it governs AckWait-expiry redelivery only (confirmed in the nats.go docs and in the server&#39;s `processNak`, which puts a bare NAK straight on the redelivery queue and signals).

Worst at exactly the case it handled worst: an overloaded Elasticsearch reports `es_rejected_execution_exception` as a per-item **429 inside an otherwise-200** `_bulk` response. That&#39;s correctly a failure, so the batch was re-submitted at once — adding load to a cluster that had just shed it.

**What&#39;s left of this after the rebase:** `#344` fixed the bare-nak half on main — the whole flush path now paces through `pkg/jsretry`, which is where this PR was taking it. What remains here is the half `#344` didn&#39;t address: **classification**. `jsretry` picks a delay by *delivery count*, and a delivery count can&#39;t express &#34;the backend just told me it is full&#34;. New `searchengine.IsBulkItemBackpressure` recognises saturation (429, or the error type, since a circuit breaker can surface under a 503) and routes those items to `jsretry.BackpressureBackoff`, which skips the 1s/5s rungs that feed the overload.

The rebase also surfaced an inversion worth a commit of its own: `#344` gave `DefaultBackoff` a 10m tail, which left `BackpressureBackoff` with a **smaller** total budget than an ordinary failure (3m30s vs 12m36s) — reporting saturation would have bought a message *less* time to land. Its tail is extended to match, and a test now pins the dominance in both directions so the two schedules can&#39;t drift apart again.

Retry window before a message is dropped (`MaxDeliver=6`, so five retries; `jsretry` applies AWS equal jitter, so each delay lands in `[base/2, base]` and the window is a range):

| | Before | After |
|---|---|---|
| General failure | ~0s | 6m18s – 12m36s (`jsretry.DefaultBackoff`) |
| Backpressure | ~0s | 9m15s – 18m30s (`jsretry.BackpressureBackoff`) |

Still bounded: saturation lasting longer than that drops messages, and `CONSUMER_MAX_DELIVER` is the ops lever for it.

## 3. Poison documents burned the whole retry budget

A bulk item failing with **400** (`mapper_parsing_exception`, an unparseable field) can never succeed — the backend rejected the document itself. Every item failure was settled as transient, so a bad document burned all its deliveries and held an ack-pending slot for the full retry window before being dropped anyway. Same end state, twelve minutes and five wasted round-trips later.

400s now settle through `errcode.Permanent`, so `jsretry` Ack-drops them. Deliberately narrow: 429/503 are backpressure, a 409 on an update is a live conflict a retry can win, and a 404 `index_not_found_exception` clears once the index exists. All templates set `dynamic:false`, so an unmapped field is ignored rather than rejected — a 400 can&#39;t be a transient mapping race during a rolling deploy.

**Behaviour change:** malformed documents are now discarded immediately instead of after the full retry window. Because the settle path uses `SettleQuiet`, `jsretry`&#39;s own drop warning is suppressed, so the bulk-item log carries an explicit `disposition` field (`drop` | `retry`) — otherwise this would have been a silent drop.

## 4. spotlight-org had no ordering guard at all

Its writes were plain `doc_as_upsert` updates with no version and no timestamp compare, so whichever landed last won. Two ways that goes wrong: a batch nakked and redelivered after a later one applied, and two pods sharing the durable consumer processing adjacent batches concurrently. Either leaves the older org row stored until HR happens to republish that section.

Every other collection had a token to guard on and this one didn&#39;t — messages/spotlight use the event timestamp as an ES external version, user-room compares `params.ts` in painless. The HR `employees.upsert` payload is a bare array with no envelope, no timestamp, and no version field on the employee record (`employees.quit` has one; upsert doesn&#39;t).

So the guard keys on the message&#39;s **JetStream stream sequence** — a strict total order assigned server-side at publish, and the only ordering token the payload path offers. Reaching it needs the message rather than just its bytes, so `Collection` gains a narrow optional extension, `sequencedCollection`, that `Handler` routes through. Only spotlight-org implements it; a compile-time assertion keeps the wiring from being lost to a rename. Widening `BuildAction` for every collection would have churned ~65 call sites to hand three of them a parameter they have no use for.

The write becomes a `scripted_upsert` so the script owns both insert and merge and a new `orgSeq` field records its sequence. Merging `params.doc` key-by-key preserves the doc-merge semantics the `omitempty` row relies on — the existing integration assertion for partial publishes is unchanged and still passes. Unreadable metadata yields sequence 0, which `BuildAction` rejects as an error rather than letting a sequence-less write reach ES.

**Operational caveat**, documented on the script: the comparison is only meaningful within one stream. Recreating the HR stream restarts sequences at 1, which would lose to stored values and silently drop every later update — clear `orgSeq` or reindex as part of any such rebuild.

This is also what unblocks depth &gt; 1. Concurrent flushes can reach ES out of order, which is only safe now that all five collections carry a guard; the pipeline doc records that a collection added without one must run at depth 1.

## Commits

Four logical commits plus the rebase follow-up, each independently building and green so the branch stays bisectable:

1. `chore:` ignore the stray build binary
2. `fix:` settle failed writes through `pkg/jsretry` (now: backpressure/poison classification — §2, §3)
3. `fix:` guard spotlight-org against out-of-order writes (§4)
4. `perf:` pipeline the ES bulk flush, depth 2 by default (§1)
5. `fix(jsretry):` give `BackpressureBackoff` at least the default retry budget

## Testing

Every behaviour was written test-first and watched fail for the right reason — the pipelining tests failed with *&#34;consumer must fetch the next batch while the previous bulk request is in flight&#34;*, the backpressure-budget test with *&#34;3m30s is not greater than or equal to 12m36s&#34;*.

- `make test` (full monorepo, `-race`): passes
- Each of the five commits verified independently: `go build ./...` clean, package tests green
- `make lint`: 0 issues · `gosec`: clean
- Delay assertions are ranges, not exact values — `#344`&#39;s jitter deliberately randomizes within `[base/2, base]`, and pinning an exact duration would be flaky
- New functions at 100% coverage (`Take`, `bulkItemError`, `itemBackoff`, `buildActions`, `streamSequence`, `newFlushPipeline`, `run`, `wait`, `checkBatchAckCoupling`); `runConsumer` went from **0%** to 74%; worker package 55.7% → 63.1%
- The concurrency tests are bounded in both directions: they assert the pipeline *reaches* depth 2 and that no third request starts while both slots are held

**Not verified locally:** Docker was unavailable, so the extended spotlight-org integration test — which is the only thing that actually executes the new painless script — was compile-checked with `go vet -tags integration` but never run. A syntax error in that script would pass everything above and fail only in CI. Worth watching. `govulncheck` and `semgrep` also couldn&#39;t run (blocked egress / not installed); `gosec` passed.

## Not in scope

Found during the investigation, deliberately left out:

- **Synchronous per-message ES parent lookup** — a `Search` across every monthly index (~96 shard queries) per thread reply, on the consumer goroutine. Compounded by `refresh_interval: 30s`, which makes it systematically miss parents younger than 30s. Can exceed `AckWait` on reply-heavy replays, and can eat this PR&#39;s entire benefit by making the loop build-bound. **Now fixed separately in #313.**
- **No deadline on the ES bulk request** — the transport clones `http.DefaultTransport` (no `ResponseHeaderTimeout`) and the context is `context.Background()`, so a hung write can stall a collection indefinitely.
- Redundant per-account body marshalling in the fan-out collections; no `pkg/natsmetrics` instrumentation on this worker.

https://claude.ai/code/session_01NnEWtiYrtp72HsX78yjDLy

### DIFF
--- a/pkg/jsretry/jsretry_test.go
+++ b/pkg/jsretry/jsretry_test.go
@@ -181,6 +181,29 @@ func TestDefaultBackoff_Shape(t *testing.T) {
 	}, DefaultBackoff)
 }
 
+// BackpressureBackoff exists to be strictly gentler than DefaultBackoff on a
+// backend that just said it is full: never retry sooner, never give up earlier.
+// The second half matters as much as the first — a schedule that skipped the
+// fast rungs but capped its tail below DefaultBackoff would swap one failure
+// mode (amplifying the overload) for another (dropping writes while the cluster
+// is still draining).
+func TestBackpressureBackoff_DominatesDefault(t *testing.T) {
+	assert.GreaterOrEqual(t, BackpressureBackoff[0], DefaultBackoff[2],
+		"the first backpressure retry must skip the fast rungs that feed the overload")
+
+	sum := func(s []time.Duration) time.Duration {
+		var total time.Duration
+		for _, d := range s {
+			total += d
+		}
+		return total
+	}
+	assert.GreaterOrEqual(t, sum(BackpressureBackoff), sum(DefaultBackoff),
+		"a saturated cluster gets at least the retry budget an ordinary failure gets")
+	assert.GreaterOrEqual(t, BackpressureBackoff[len(BackpressureBackoff)-1], DefaultBackoff[len(DefaultBackoff)-1],
+		"the reused tail entry paces the long outage; it must not be shorter than the default tail")
+}
+
 // Deliberately not extended: broadcast fan-out is user-visible and a
 // 15-minute-late broadcast is worthless.
 func TestLowLatencyBackoff_StaysShort(t *testing.T) {
--- a/pkg/searchengine/classify.go
+++ b/pkg/searchengine/classify.go
@@ -47,3 +47,44 @@ func IsBulkItemSuccess(action ActionType, result BulkResult) bool {
 	}
 	return false
 }
+
+// ErrRejectedExecution / ErrCircuitBreaking are the ES error types that mean
+// "the cluster is saturated", not "this write is invalid". They arrive as
+// per-item failures inside an otherwise-200 _bulk response.
+const (
+	ErrRejectedExecution = "es_rejected_execution_exception"
+	ErrCircuitBreaking   = "circuit_breaking_exception"
+)
+
+// IsBulkItemBackpressure reports whether a failed bulk item was rejected
+// because the search backend is overloaded rather than because the write
+// itself was bad. Callers use it to retry on a longer delay: retrying
+// promptly against a cluster that just shed load is what turns a transient
+// overload into a sustained one.
+//
+// Matched on status (429) or on the error type, since a circuit breaker can
+// surface under a 503 as well.
+func IsBulkItemBackpressure(result BulkResult) bool {
+	if result.Status == 429 {
+		return true
+	}
+	switch result.ErrorType {
+	case ErrRejectedExecution, ErrCircuitBreaking:
+		return true
+	}
+	return false
+}
+
+// IsBulkItemPermanent reports whether a failed bulk item can never succeed on
+// retry. A 400 means the backend rejected the document itself — a mapping
+// conflict, an unparseable field, an illegal argument — so redelivering
+// identical bytes can only fail identically. Callers drop it rather than
+// burning the consumer's redelivery budget on a doomed retry.
+//
+// Deliberately narrow: 429/503 are backpressure (IsBulkItemBackpressure), a 409
+// on an update is a live conflict a retry can win, and a 404
+// index_not_found_exception clears once the index or template exists. None of
+// those are permanent.
+func IsBulkItemPermanent(result BulkResult) bool {
+	return result.Status == 400
+}
--- a/search-sync-worker/handler.go
+++ b/search-sync-worker/handler.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"log/slog"
 	"sync"
 	"time"
@@ -11,6 +12,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 
+	"github.com/hmchangw/chat/pkg/errcode"
 	"github.com/hmchangw/chat/pkg/jsretry"
 	"github.com/hmchangw/chat/pkg/natsutil"
 	"github.com/hmchangw/chat/pkg/searchengine"
@@ -146,6 +148,30 @@ func (h *Handler) AddWithContext(ctx context.Context, msg jetstream.Msg) {
 	h.mu.Unlock()
 }
 
+// bulkItemError builds the settle error for one failed bulk item. A permanent failure
+// is wrapped with errcode.Permanent so jsretry Ack-drops it instead of retrying — the
+// backend rejected the document itself, so redelivering identical bytes can only fail
+// identically, and retrying would just hold an ack-pending slot for the full window.
+//
+// Only the status and error type go into the message; the document body never belongs
+// in an error that reaches the server log.
+func bulkItemError(result searchengine.BulkResult) error {
+	if searchengine.IsBulkItemPermanent(result) {
+		return errcode.Permanent(errcode.BadRequest(
+			fmt.Sprintf("search backend rejected the document: status %d %q", result.Status, result.ErrorType)))
+	}
+	return fmt.Errorf("bulk item failed: status %d %q", result.Status, result.ErrorType)
+}
+
+// itemBackoff picks the retry schedule a failed bulk item calls for. Backpressure gets
+// the slow curve; everything else rides the shared default.
+func itemBackoff(result searchengine.BulkResult) []time.Duration {
+	if searchengine.IsBulkItemBackpressure(result) {
+		return jsretry.BackpressureBackoff
+	}
+	return jsretry.DefaultBackoff
+}
+
 // Flush sends all buffered actions to ES and acks/naks per source message.
 func (h *Handler) Flush(ctx context.Context) {
 	h.mu.Lock()
@@ -194,35 +220,57 @@ func (h *Handler) Flush(ctx context.Context) {
 
 	h.metrics.recordFlush(bulkCtx, flushSuccess, elapsed, len(actions))
 
-	var acked, nakked int
+	var acked, poisoned, nakked int
 	for _, p := range pending {
-		allOK := true
+		var itemErr error
+		backoff := jsretry.DefaultBackoff
+		failed, permanent := false, false
 		for i := p.actionStart; i < p.actionStart+p.actionCount; i++ {
 			if searchengine.IsBulkItemSuccess(actions[i].Action, results[i]) {
 				continue
 			}
 			h.metrics.recordItemFailure(bulkCtx, string(actions[i].Action), results[i].Status)
-			if allOK {
-				// Log only the first failed item per message; the counter above still
-				// sees every failed action.
-				slog.Error("bulk item failed",
-					"status", results[i].Status,
-					"error", results[i].Error,
-					"docID", actions[i].DocID,
-					"index", actions[i].Index,
-				)
+			if failed {
+				continue // the counter above sees every failed action; the rest is decided once
 			}
-			allOK = false
+			failed = true
+			// The first failure decides the message's fate. disposition spells out what that
+			// was, because SettleQuiet leaves the Ack-drop of a permanent failure otherwise
+			// unlogged; it reads from the result, not the error, so the log cannot disagree
+			// with the settle decision.
+			permanent = searchengine.IsBulkItemPermanent(results[i])
+			disposition := "retry"
+			if permanent {
+				disposition = "drop"
+			}
+			slog.ErrorContext(p.ctx, "bulk item failed",
+				"status", results[i].Status,
+				"error", results[i].Error,
+				"errorType", results[i].ErrorType,
+				"backpressure", searchengine.IsBulkItemBackpressure(results[i]),
+				"disposition", disposition,
+				"docID", actions[i].DocID,
+				"index", actions[i].Index,
+			)
+			itemErr, backoff = bulkItemError(results[i]), itemBackoff(results[i])
 		}
-		if allOK {
+		// A permanent failure is Acked, not nakked, so it counts as a poison drop rather
+		// than inflating the nakked series with messages that were never retried.
+		switch {
+		case !failed:
 			acked++
-			natsutil.Ack(p.jsMsg, "bulk actions succeeded")
-		} else {
+		case permanent:
+			poisoned++
+		default:
 			nakked++
-			jsretry.Nak(bulkCtx, p.jsMsg, jsretry.DefaultBackoff, "bulk action failed")
 		}
+		// One SettleQuiet covers all three dispositions: a nil error acks, a permanent
+		// error Ack-drops, anything else naks on `backoff`. Quiet because the failure is
+		// already logged above with its ES detail.
+		jsretry.SettleQuiet(p.ctx, p.jsMsg, backoff, itemErr)
 	}
 	h.metrics.recordMessages(bulkCtx, dispAckedSuccess, acked)
+	h.metrics.recordMessages(bulkCtx, dispAckedPoison, poisoned)
 	h.metrics.recordMessages(bulkCtx, dispNakkedItemFailed, nakked)
 }
 
--- a/search-sync-worker/handler.go
+++ b/search-sync-worker/handler.go
@@ -193,18 +193,44 @@ func streamSequence(msg jetstream.Msg) uint64 {
 	return md.Sequence.Stream
 }
 
-// Flush sends all buffered actions to ES and acks/naks per source message.
-func (h *Handler) Flush(ctx context.Context) {
+// flushBatch is one detached unit of work: the buffered source messages and
+// the ES bulk actions they produced. Take hands it off so the consumer loop can
+// keep buffering the next batch while this one's bulk request is still in
+// flight.
+type flushBatch struct {
+	pending []pendingMsg
+	actions []searchengine.BulkAction
+}
+
+// Take detaches everything currently buffered and resets the buffer, returning
+// nil when there is nothing to flush. It never touches the search engine, so
+// the caller can hand the returned batch to a background flusher and keep
+// buffering immediately — ActionCount() reads 0 as soon as this returns.
+func (h *Handler) Take() *flushBatch {
 	h.mu.Lock()
+	defer h.mu.Unlock()
 	if len(h.pending) == 0 {
-		h.mu.Unlock()
-		return
+		return nil
 	}
-	pending := h.pending
-	actions := h.actions
+	batch := &flushBatch{pending: h.pending, actions: h.actions}
 	h.pending = make([]pendingMsg, 0, h.bulkSize)
 	h.actions = make([]searchengine.BulkAction, 0, h.bulkSize)
-	h.mu.Unlock()
+	return batch
+}
+
+// Flush detaches the buffer and sends it to ES. Callers that want the bulk
+// request to overlap with the next batch use Take + FlushBatch instead.
+func (h *Handler) Flush(ctx context.Context) {
+	h.FlushBatch(ctx, h.Take())
+}
+
+// FlushBatch sends one detached batch to ES and acks/naks per source message.
+// A nil batch is a no-op.
+func (h *Handler) FlushBatch(ctx context.Context, batch *flushBatch) {
+	if batch == nil {
+		return
+	}
+	pending, actions := batch.pending, batch.actions
 
 	bulkCtx, span := h.startFlushSpan(ctx, pending, len(actions))
 	defer span.End()
--- a/search-sync-worker/handler.go
+++ b/search-sync-worker/handler.go
@@ -121,7 +121,7 @@ func (h *Handler) AddWithContext(ctx context.Context, msg jetstream.Msg) {
 		}
 	}
 
-	actions, err := h.collection.BuildAction(data)
+	actions, err := h.buildActions(msg, data)
 	if err != nil {
 		// Every BuildAction error is parse/validation poison — Ack drops it for
 		// good, so this line is the only trace; keep it identifying the message.
@@ -170,6 +170,27 @@ func itemBackoff(result searchengine.BulkResult) []time.Duration {
 		return jsretry.BackpressureBackoff
 	}
 	return jsretry.DefaultBackoff
+}
+
+// buildActions converts one payload, handing collections that guard on stream order
+// the message's sequence (see sequencedCollection).
+func (h *Handler) buildActions(msg jetstream.Msg, data []byte) ([]searchengine.BulkAction, error) {
+	sc, ok := h.collection.(sequencedCollection)
+	if !ok {
+		return h.collection.BuildAction(data)
+	}
+	return sc.BuildActionSeq(data, streamSequence(msg))
+}
+
+// streamSequence returns the message's position in its stream, or 0 when the metadata
+// can't be read. Zero is the fail-closed value: it loses to any stored sequence, so a
+// message we can't order is skipped rather than allowed to overwrite a newer row.
+func streamSequence(msg jetstream.Msg) uint64 {
+	md, err := msg.Metadata()
+	if err != nil || md == nil {
+		return 0
+	}
+	return md.Sequence.Stream
 }
 
 // Flush sends all buffered actions to ES and acks/naks per source message.
--- a/search-sync-worker/main.go
+++ b/search-sync-worker/main.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/caarlos0/env/v11"
@@ -77,6 +78,13 @@ type config struct {
 	// the time-based counterpart to the size trigger, bounding write latency during idle periods.
 	BulkFlushInterval int `env:"BULK_FLUSH_INTERVAL" envDefault:"5"`
 
+	// PipelineDepth is how many ES bulk requests one collection keeps in flight while later
+	// batches fetch and build. 1 serializes fetch/build behind every round-trip; higher trades
+	// ack-pending headroom for throughput when ES latency dominates. Costs are per collection
+	// AND per pod, so the cluster sees depth x collections x replicas concurrent bulk requests.
+	// Needs (PipelineDepth+1) * BulkBatchSize <= CONSUMER_MAX_ACK_PENDING — see checkBatchAckCoupling.
+	PipelineDepth int `env:"PIPELINE_DEPTH" envDefault:"2"`
+
 	Consumer  stream.ConsumerSettings `envPrefix:"CONSUMER_"`
 	Bootstrap bootstrapConfig         `envPrefix:"BOOTSTRAP_"`
 
@@ -115,6 +123,12 @@ func main() {
 		slog.Error("invalid config", "name", "BULK_FLUSH_INTERVAL", "value", cfg.BulkFlushInterval, "reason", "must be > 0")
 		os.Exit(1)
 	}
+	// A non-positive depth would make the pipeline's slot channel unbuffered and park the
+	// consumer on its first flush forever.
+	if cfg.PipelineDepth <= 0 {
+		slog.Error("invalid config", "name", "PIPELINE_DEPTH", "value", cfg.PipelineDepth, "reason", "must be > 0")
+		os.Exit(1)
+	}
 	if _, _, ok := searchindex.StripVersion(cfg.MsgIndexPrefix); !ok {
 		slog.Error("invalid config", "name", "MSG_INDEX_PREFIX", "value", cfg.MsgIndexPrefix, "reason", "must end with -v<N>, e.g. messages-site-a-v1")
 		os.Exit(1)
@@ -134,10 +148,11 @@ func main() {
 	}
 
 	// Warn (don't fail) if the bulk batch size can't be reached under the consumer's ack-pending ceiling — see checkBatchAckCoupling.
-	if warning := checkBatchAckCoupling(cfg.BulkBatchSize, cfg.Consumer.MaxAckPending); warning != "" {
+	if warning := checkBatchAckCoupling(cfg.BulkBatchSize, cfg.Consumer.MaxAckPending, cfg.PipelineDepth); warning != "" {
 		slog.Warn("batch/ack-pending config coupling",
 			"bulkBatchSize", cfg.BulkBatchSize,
 			"maxAckPending", cfg.Consumer.MaxAckPending,
+			"pipelineDepth", cfg.PipelineDepth,
 			"detail", warning,
 		)
 	}
@@ -334,7 +349,11 @@ func main() {
 			"filters", consumerCfg.FilterSubjects,
 		)
 
-		go runConsumer(ctx, fetcher, handler, cfg.FetchBatchSize, cfg.BulkBatchSize, bulkFlushInterval, stopCh, doneCh)
+		go runConsumer(ctx, fetcher, handler, consumerTuning{
+			fetchBatchSize:    cfg.FetchBatchSize,
+			bulkFlushInterval: bulkFlushInterval,
+			pipelineDepth:     cfg.PipelineDepth,
+		}, stopCh, doneCh)
 	}
 
 	healthStop, err := health.ServeWithPprof(cfg.HealthAddr, 5*time.Second, cfg.PProfEnabled,
@@ -397,23 +416,90 @@ func pushMappings(ctx context.Context, engine searchengine.SearchEngine, collect
 	return nil
 }
 
-// runConsumer is the batch-flush consumer loop: fetchBatchSize bounds JetStream Fetch() pulls
-// (client-tuning only), bulkBatchSize caps buffered ES actions (flushed on stopCh, size, or bulkFlushInterval; the loop clamps fetch to remaining bulk capacity, but a fan-out message can still overshoot mid-loop and triggers its own flush).
+// flushPipeline overlaps up to `depth` ES bulk requests with the fetch and build of the
+// batches behind them. The slot channel is the cap: unbounded concurrency would queue
+// batches against ES without limit and blow through the consumer's ack-pending budget.
+//
+// Concurrent flushes can reach ES out of order, which is safe only because every
+// collection carries an ordering guard — external versioning on messages and spotlight,
+// a painless timestamp compare on user-room, a stream-sequence compare on spotlight-org.
+// A collection added without one MUST run at depth 1.
+//
+// Only the consumer goroutine calls run/wait, so the struct needs no lock of its own.
+type flushPipeline struct {
+	slots chan struct{} // buffered to depth; a token is held for each in-flight flush
+	wg    sync.WaitGroup
+}
+
+func newFlushPipeline(depth int) *flushPipeline {
+	return &flushPipeline{slots: make(chan struct{}, depth)}
+}
+
+// wait blocks until every in-flight flush has finished.
+func (p *flushPipeline) wait() {
+	p.wg.Wait()
+}
+
+// run detaches whatever the handler has buffered and flushes it in the background,
+// blocking for a slot once `depth` flushes are already in flight. Taking the batch first
+// frees the buffer immediately, so ActionCount() reads 0 on return.
+func (p *flushPipeline) run(ctx context.Context, h *Handler) {
+	batch := h.Take()
+	if batch == nil {
+		return
+	}
+	p.slots <- struct{}{}
+	p.wg.Add(1)
+	go func() {
+		defer p.wg.Done()
+		defer func() { <-p.slots }()
+		// jobguard recovers panics from the batch handler so a poison message or malformed
+		// bulk response can't crash this collection's consumer; the batch's messages stay
+		// un-acked and JetStream redelivers after AckWait.
+		jobguard.Guard("search-sync flush", func() { h.FlushBatch(ctx, batch) })
+	}()
+}
+
+// consumerTuning groups runConsumer's throughput knobs so call sites name them instead
+// of passing a run of bare numbers.
+type consumerTuning struct {
+	fetchBatchSize    int
+	bulkFlushInterval time.Duration
+	pipelineDepth     int
+}
+
+// runConsumer is the batch-flush consumer loop: tune.fetchBatchSize bounds JetStream Fetch()
+// pulls (client-tuning only), while the handler's own bulk size caps buffered ES actions
+// (flushed on stopCh, size, or tune.bulkFlushInterval; the loop clamps fetch to remaining bulk
+// capacity, but a fan-out message can still overshoot mid-loop and triggers its own flush).
+//
+// Each flush is handed to a flushPipeline holding up to pipelineDepth requests in flight, so
+// ES round-trips overlap with the fetch and build behind them rather than serializing; the
+// loop only stalls once every slot is busy, which is the backpressure we want.
 func runConsumer(
 	ctx context.Context,
 	cons msgFetcher,
 	handler *Handler,
-	fetchBatchSize, bulkBatchSize int,
-	bulkFlushInterval time.Duration,
+	tune consumerTuning,
 	stopCh <-chan struct{},
 	doneCh chan<- struct{},
 ) {
 	defer close(doneCh)
 	lastFlush := time.Now()
 
-	// jobguard recovers panics from the batch handler so a poison message or malformed bulk response
-	// can't crash this collection's consumer goroutine; in-flight messages stay un-acked and JetStream redelivers after AckWait.
-	flush := func() { jobguard.Guard("search-sync flush", func() { handler.Flush(ctx) }) }
+	// bulkBatchSize is the handler's own cap — read it from there rather than passing the
+	// same number twice and letting the two drift.
+	bulkBatchSize := handler.bulkSize
+
+	pipe := newFlushPipeline(tune.pipelineDepth)
+	// flushNow hands the buffered batch to the pipeline and restarts the interval clock.
+	flushNow := func() {
+		pipe.run(ctx, handler)
+		lastFlush = time.Now()
+	}
+	// drain also waits on the in-flight bulk requests, so shutdown never abandons a write
+	// that is already on its way to ES.
+	drain := func() { flushNow(); pipe.wait() }
 	add := func(msgCtx context.Context, m jetstream.Msg) {
 		jobguard.Guard("search-sync add: "+m.Subject(), func() { handler.AddWithContext(msgCtx, m) })
 	}
@@ -421,35 +507,25 @@ func runConsumer(
 	for {
 		select {
 		case <-stopCh:
-			flush()
+			drain()
 			return
 		default:
 		}
 
 		// Bound the next Fetch by remaining bulk capacity so a steady stream of 1:1 messages can't
 		// overshoot bulkBatchSize; fan-out messages may still push us over, handled mid-loop below.
-		remaining := bulkBatchSize - handler.ActionCount()
-		if remaining <= 0 {
-			flush()
-			lastFlush = time.Now()
-			continue
-		}
-		fetchCount := fetchBatchSize
-		if fetchCount > remaining {
-			fetchCount = remaining
-		}
+		fetchCount := min(tune.fetchBatchSize, bulkBatchSize-handler.ActionCount())
 
 		batch, err := cons.Fetch(ctx, fetchCount, jetstream.FetchMaxWait(time.Second))
 		if err != nil {
 			select {
 			case <-stopCh:
-				flush()
+				drain()
 				return
 			default:
 			}
-			if handler.ActionCount() > 0 && time.Since(lastFlush) >= bulkFlushInterval {
-				flush()
-				lastFlush = time.Now()
+			if handler.ActionCount() > 0 && time.Since(lastFlush) >= tune.bulkFlushInterval {
+				flushNow()
 			}
 			continue
 		}
@@ -458,34 +534,35 @@ func runConsumer(
 		// goroutine blocked on an unbuffered send; an early break would leak it and stall shutdown.
 		for msg := range batch.Messages() {
 			add(msg.Ctx, msg.Msg)
-			// Mid-batch flush: if a fan-out message just pushed the buffer over the bulk cap, flush
-			// immediately — otherwise the next message's actions add to an already-oversized request.
+			// Size trigger, checked per message: a fan-out message can cross the cap on its own, and
+			// flushing here keeps the next message's actions out of an already-oversized request.
+			// Take() empties the buffer synchronously, so ActionCount stays under the cap after this.
 			if handler.ActionCount() >= bulkBatchSize {
-				flush()
-				lastFlush = time.Now()
+				flushNow()
 			}
 		}
 
-		if handler.ActionCount() >= bulkBatchSize {
-			flush()
-			lastFlush = time.Now()
-		} else if handler.ActionCount() > 0 && time.Since(lastFlush) >= bulkFlushInterval {
-			flush()
-			lastFlush = time.Now()
+		// Interval trigger: the size trigger above already fired for anything at the cap.
+		if handler.ActionCount() > 0 && time.Since(lastFlush) >= tune.bulkFlushInterval {
+			flushNow()
 		}
 	}
 }
 
-// checkBatchAckCoupling warns when bulkBatchSize exceeds maxAckPending: a 1:1 collection then
-// stalls at maxAckPending un-acked messages before the size-based flush can fire, undersizing every batch. Fan-out collections are unaffected. Empty string = no issue.
-func checkBatchAckCoupling(bulkBatchSize, maxAckPending int) string {
-	if bulkBatchSize > maxAckPending {
+// checkBatchAckCoupling warns when the pipelined batches can't fit under maxAckPending.
+// depth batches can be in flight while one more fills, so a 1:1 collection needs headroom
+// for (depth+1)*bulkBatchSize un-acked messages; below that it stalls before the size-based
+// flush can fire, undersizing every batch. Fan-out collections are unaffected. "" = no issue.
+func checkBatchAckCoupling(bulkBatchSize, maxAckPending, pipelineDepth int) string {
+	needed := (pipelineDepth + 1) * bulkBatchSize
+	if needed > maxAckPending {
 		return fmt.Sprintf(
-			"BULK_BATCH_SIZE (%d) exceeds CONSUMER_MAX_ACK_PENDING (%d): "+
-				"the size-based flush can never fire for 1:1 collections, so flushes "+
-				"will wait the full BULK_FLUSH_INTERVAL and batches stay undersized. "+
-				"Lower BULK_BATCH_SIZE to <= MAX_ACK_PENDING or raise MAX_ACK_PENDING.",
-			bulkBatchSize, maxAckPending,
+			"BULK_BATCH_SIZE (%[1]d) at PIPELINE_DEPTH %[2]d needs CONSUMER_MAX_ACK_PENDING >= "+
+				"%[3]d (%[2]d in flight plus one filling) but it is %[4]d: the size-based flush "+
+				"can never fire for 1:1 collections, so flushes will wait the full "+
+				"BULK_FLUSH_INTERVAL and batches stay undersized. Raise MAX_ACK_PENDING to "+
+				"%[3]d, or lower BULK_BATCH_SIZE to %[5]d, or drop PIPELINE_DEPTH.",
+			bulkBatchSize, pipelineDepth, needed, maxAckPending, maxAckPending/(pipelineDepth+1),
 		)
 	}
 	return ""
--- a/search-sync-worker/spotlight_org.go
+++ b/search-sync-worker/spotlight_org.go
@@ -12,6 +12,37 @@ import (
 	"github.com/hmchangw/chat/pkg/subject"
 )
 
+// spotlightOrgUpsertScriptID is the stored-script id for the last-write-wins guard.
+// Bump the suffix if the source ever changes incompatibly, so a rolling deploy doesn't
+// leave old and new pods sharing one mutated definition.
+const spotlightOrgUpsertScriptID = "search-sync-spotlight-org-upsert-v1"
+
+// spotlightOrgUpsertScript applies one HR org row under a last-write-wins guard keyed on
+// params.seq — the source message's JetStream stream sequence, a strict total order
+// assigned server-side at publish time.
+//
+// The HR employees.upsert payload is a bare array carrying no timestamp of its own, so
+// the stream sequence is the only ordering token available. Without it the write is a
+// plain doc_as_upsert with no ordering semantics: a redelivered batch, or a batch from a
+// second pod sharing the durable consumer, can land after a newer one and leave the older
+// row stored.
+//
+// Merging params.doc key-by-key reproduces doc-merge semantics — the row is marshalled
+// with omitempty, so fields absent from a partial publish keep their stored values.
+//
+// Runs under scripted_upsert so a brand-new sectId also records its sequence;
+// ctx._source starts as the empty upsert doc and stored reads 0.
+//
+// Operational note: the guard compares sequences from ONE stream. If the HR stream is
+// ever recreated, sequences restart at 1 and will lose to the stored values, silently
+// dropping every subsequent update — clear orgSeq (or reindex) as part of any rebuild.
+const spotlightOrgUpsertScript = `long stored = ctx._source.orgSeq == null ? 0L : ((Number)ctx._source.orgSeq).longValue(); ` +
+	`long incoming = ((Number)params.seq).longValue(); ` +
+	`if (incoming > stored) { ` +
+	`for (entry in params.doc.entrySet()) { ctx._source[entry.getKey()] = entry.getValue(); } ` +
+	`ctx._source.orgSeq = incoming; ` +
+	`} else { ctx.op = 'none'; }`
+
 // spotlightOrgCollection maintains the spotlight-org ES index, one
 // document per sectId. Many employees collapse to one document via
 // dedup in BuildAction.
@@ -58,8 +89,12 @@ func (c *spotlightOrgCollection) TemplateBody() json.RawMessage {
 	return spotlightOrgTemplateBody(c.indexName, c.devMode)
 }
 
+// StoredScripts registers the last-write-wins guard; BuildAction references it by id
+// so the source isn't repeated in every fan-out action.
 func (c *spotlightOrgCollection) StoredScripts() map[string]json.RawMessage {
-	return nil
+	return map[string]json.RawMessage{
+		spotlightOrgUpsertScriptID: searchindex.StoredScriptBody(spotlightOrgUpsertScript),
+	}
 }
 
 // MappingUpdate returns no update — spotlight-org writes to one fixed index.
@@ -74,7 +109,18 @@ func (c *spotlightOrgCollection) MappingUpdate() (string, json.RawMessage) {
 // (last-wins) and emits one ES _update per unique sectId with doc_as_upsert:true.
 // Doc-merge + omitempty means partial-field publishes preserve stored values for
 // unset fields.
-func (c *spotlightOrgCollection) BuildAction(data []byte) ([]searchengine.BulkAction, error) {
+// BuildAction satisfies Collection but cannot be used: without a stream sequence the
+// guard rejects every write — including the insert of a brand-new sectId, since the
+// upsert seeds stored=0 and the guard needs incoming > stored — and ES reports each one
+// as a successful no-op. Failing loudly beats a fleet of silently discarded writes;
+// Handler always routes through BuildActionSeq.
+func (c *spotlightOrgCollection) BuildAction([]byte) ([]searchengine.BulkAction, error) {
+	return nil, fmt.Errorf("spotlight-org requires the source stream sequence: use BuildActionSeq")
+}
+
+// BuildActionSeq is BuildAction plus the source message's JetStream stream sequence,
+// which the last-write-wins guard compares against the stored one.
+func (c *spotlightOrgCollection) BuildActionSeq(data []byte, streamSeq uint64) ([]searchengine.BulkAction, error) {
 	var rows []SpotlightOrgIndex
 	if err := json.Unmarshal(data, &rows); err != nil {
 		return nil, fmt.Errorf("unmarshal hr employees: %w", err)
@@ -98,7 +144,7 @@ func (c *spotlightOrgCollection) BuildAction(data []byte) ([]searchengine.BulkAc
 
 	actions := make([]searchengine.BulkAction, 0, len(deduped))
 	for sectID, row := range deduped {
-		body, err := buildSpotlightOrgUpdateBody(row)
+		body, err := buildSpotlightOrgUpdateBody(row, streamSeq)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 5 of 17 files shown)


julianshen:
**7 findings · 6 unique issues · 6 inline comments posted** — 🔴 5 blocking · 🟠 2 warning

### Needs clarification

Will deployment stop and drain every old worker before starting new workers?

`answer clar_bnodeei6dnbsm4sgmy7xuqeodx: <your answer>`

[Open the question](https://github.com/hmchangw/newchat/pull/311#issuecomment-5380770947)

### Related work

- [Issue #39](https://github.com/hmchangw/newchat/issues/39) — \[Fix\] Room-key cache TTL-safety validated against drift-prone per-service grace config (open)
- [Issue #34](https://github.com/hmchangw/newchat/issues/34) — \[Fix\] messageRead should publish a subscription.update \(action "read"\) to the reader (closed)
- [PR #313](https://github.com/hmchangw/newchat/pull/313) — search-sync-worker: point-read the thread parent instead of searching every monthly index (open)

### Local review memory

- `search-sync-worker/spotlight_org.go:149` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟡 Minor_ _⚡ Quick win_ **Wrap the update-body error with section context.** Line 149 returns err unchanged. Include the affected sectID and the current operation so a bulk-build failure can ide… ([thread](https://github.com/hmchangw/newchat/pull/311#discussion_r3834884527))
- `search-sync-worker/collection.go:69` — @coderabbitai: _🎯 Functional Correctness_ _🔴 Critical_ _⚡ Quick win_ **Remove the duplicate sequencedCollection declaration.** search-sync-worker/handler.go already declares sequencedCollection at lines 66-69. This added declaration causes s… ([thread](https://github.com/hmchangw/newchat/pull/311#discussion_r3834884515))
- `search-sync-worker/config_test.go:52` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟡 Minor_ _⚡ Quick win_ &lt;details> &lt;summary>🧩 Analysis chain&lt;/summary> 🏁 Script executed: shell #!/bin/bash set -eu printf '%s\n' '--- config_test.go ---' cat -n search-sync-worker/config_test.g… ([thread](https://github.com/hmchangw/newchat/pull/311#discussion_r3834884518))
- `search-sync-worker/consumer_pipeline_test.go:80` — @coderabbitai: _🩺 Stability & Availability_ _🟡 Minor_ _⚡ Quick win_ **A full entered channel can deadlock the test instead of failing it.** entered has a fixed capacity of 64 and Bulk uses a blocking send. After the test closes release , the… ([thread](https://github.com/hmchangw/newchat/pull/311#discussion_r3834884522))
- `search-sync-worker/integration_test.go:767` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟠 Major_ _⚡ Quick win_ **Make each ordering subtest establish its own document state.** These subtests share S1 and depend on writes from the parent flow. The later test also mutates that shared… ([thread](https://github.com/hmchangw/newchat/pull/311#discussion_r3834884525))

<details>
<summary>📒 Files reviewed (17)</summary>

* `.gitignore`
* `pkg/jsretry/jsretry.go`
* `pkg/jsretry/jsretry_test.go`
* `pkg/searchengine/classify.go`
* `pkg/searchengine/classify_test.go`
* `search-sync-worker/collection.go`
* `search-sync-worker/config_test.go`
* `search-sync-worker/consumer_config_test.go`
* `search-sync-worker/consumer_pipeline_test.go`
* `search-sync-worker/deploy/docker-compose.yml`
* `search-sync-worker/handler_test.go`
* `search-sync-worker/integration_test.go`
* `search-sync-worker/metrics_test.go`
* `search-sync-worker/spotlight_org.go`
* `search-sync-worker/spotlight_org_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

pkg/searchindex/userroomdoc.go
```diff
@@ -136,3 +136,37 @@ func BuildRemoveRoomUpdateBody(rid string, ts int64) (json.RawMessage, error) {
...
+// Runs under scripted_upsert so a brand-new sectId also records its sequence;
+// ctx._source starts as the empty upsert doc and stored reads 0.
+//
+// Operational note: the guard compares sequences from ONE stream. If the HR
+// stream is ever recreated, sequences restart at 1 and will lose to the stored
+// values, silently dropping every subsequent update — clear orgSeq (or reindex)
+// as part of any such rebuild.
+const SpotlightOrgUpsertScript = `long stored = ctx._source.orgSeq == null ? 0L : ((Number)ctx._source.orgSeq).longValue(); ` +
+	`long incoming = ((Number)params.seq).longValue(); ` +
+	`if (incoming > stored) { ` +
```
julianshen: _🎯 concurrency_ | _🔴 Blocking_ | _`distributed-system`, `performance-and-concurrency` (2 rules)_

**Document-level sequencing loses fields from reordered partial updates**

The new pipeline can execute consecutive spotlight-org bulks out of order, but this guard rejects an older row wholesale. For example, seq 10 may contain `deptId=D1` while seq 11 is a partial row containing only `sectName`. If seq 11 reaches Elasticsearch first, it records `orgSeq=11`; seq 10 then becomes a no-op, so `deptId` is never stored (or remains stale). In stream order, seq 10 would apply and seq 11 would preserve that field, as the documented merge semantics require. This is a reachable lost update because rows use `omitempty` and flushes now run concurrently. Track sequence per field so an older row can still populate fields not superseded by the newer row, or require complete snapshots. As an immediate mitigation, force spotlight-org to pipeline depth 1, though redelivery still requires the per-field or full-snapshot fix.

<details>

- **`distributed-system`**:167: Consider seq 100 updating only deptName and seq 101 updating only sectName for the same sectId. Concurrent bulks can apply seq 101 first, recording orgSeq=101; when seq 100 arrives, the script skips the entire document. The final row never receives deptName, although applying the partial updates in stream order would preserve both changes. This is reachable through the new pipeline, delayed retries, or multiple pods. Track sequence per field and apply each field whose incoming sequence is newer than that field's sequence, or require and enforce full-row snapshots.

</details>

<details>

```
In `pkg/searchindex/userroomdoc.go` around line 169: The new pipeline can execute consecutive spotlight-org bulks out of order, but this guard rejects an older row wholesale. For example, seq 10 may contain `deptId=D1` while seq 11 is a partial row containing only `sectName`. If seq 11 reaches Elasticsearch first, it records `orgSeq=11`; seq 10 then becomes a no-op, so `deptId` is never stored (or remains stale). In stream order, seq 10 would apply and seq 11 would preserve that field, as the documented merge semantics require. This is a reachable lost update because rows use `omitempty` and flushes now run concurrently. Track sequence per field so an older row can still populate fields not superseded by the newer row, or require complete snapshots. As an immediate mitigation, force spotlight-org to pipeline depth 1, though redelivery still requires the per-field or full-snapshot fix.

```

</details>

search-sync-worker/handler.go
```diff
@@ -148,30 +234,37 @@ func (h *Handler) Flush(ctx context.Context) {
...
 		// No duplicate processing, no lost events.
 		slog.Error("bulk result count mismatch", "expected", len(actions), "actual", len(results))
-		nakAll(pending, "bulk result count mismatch")
+		settleAll(pending, fmt.Errorf("bulk result count mismatch: sent %d actions, got %d results", len(actions), len(results)))
 		return
 	}
 
 	for _, p := range pending {
-		allOK := true
+		var itemErr error
```
julianshen, line 272: _🏷️ partial-failure_ | _🔴 Blocking_ | _`distributed-system`, `elasticsearch`, `nats`, `silent-failure` (4 rules)_

**The first failed bulk item can hide retryable failures.**

A source message may fan out to multiple independently executed bulk items. If its first failed result is a permanent 400 while a later item is rejected with 429 or fails with 500, this loop stops at the 400 and settles the whole source message as permanent. JetStream is acknowledged, so the later item is never retried and its update is lost. Inspect all results belonging to the source message, retry whenever any failure is retryable, and classify the message as permanent only when every failed item is permanent. Backpressure should likewise be accumulated across retryable failures.

<details>

- **`elasticsearch`**:265: The loop stops at the first failed action. For a fan-out spotlight-org message, if that item is a 400 but a later item returned 429/500, `settle` receives only the permanent error and acknowledges the source message. The later retryable item is therefore never retried and its organization row is lost. Inspect every result belonging to the source message, retry when any failure is retryable, and use the backpressure schedule when any retryable failure is backpressure. Add a test with a 400 followed by a 429 in one source message.
- **`silent-failure`**:258: A source message may fan out to multiple actions, but this loop stops at its first failed result. If that result is a permanent 400 while a later action failed transiently (for example, 429 or 500), itemErr remains permanent and settle acknowledges the entire source message. The later action is then permanently lost. Scan every action result, retry the source message whenever any failure is retryable, and acknowledge it as permanent only when all failed items are permanent; aggregate backpressure with logical OR while scanning.
- **`nats`**:267: The loop stops at the first failed action and settles the entire source message from only that result. For example, one chat.hr.<site>.employees.upsert message can produce actions for S1 and S2; if S1 receives a permanent 400 while S2 receives a retryable 429 or 500, the break leaves S2 unexamined and settle ACK-drops the source message. Scan every action belonging to the source message and retry whenever any failure is retryable; only ACK-drop when all failed items are genuinely permanent.

</details>

<details>

```
In `search-sync-worker/handler.go` around line 242: A source message may fan out to multiple independently executed bulk items. If its first failed result is a permanent 400 while a later item is rejected with 429 or fails with 500, this loop stops at the 400 and settles the whole source message as permanent. JetStream is acknowledged, so the later item is never retried and its update is lost. Inspect all results belonging to the source message, retry whenever any failure is retryable, and classify the message as permanent only when every failed item is permanent. Backpressure should likewise be accumulated across retryable failures.

```

</details>

julianshen (commented):
tGD inline review

pkg/searchengine/classify.go
```diff
@@ -47,3 +47,44 @@ func IsBulkItemSuccess(action ActionType, result BulkResult) bool {
...
+// IsBulkItemPermanent reports whether a failed bulk item can never succeed on
+// retry. A 400 means the backend rejected the document itself — a mapping
+// conflict, an unparseable field, an illegal argument — so redelivering
+// identical bytes can only fail identically. Callers drop it rather than
+// burning the consumer's redelivery budget on a doomed retry.
+//
+// Deliberately narrow: 429/503 are backpressure (IsBulkItemBackpressure), a 409
+// on an update is a live conflict a retry can win, and a 404
+// index_not_found_exception clears once the index or template exists. None of
+// those are permanent.
```
julianshen, line 87: _🏷️ delivery-guarantees_ | _🔴 Blocking_ | _`distributed-system`_

**Not every Elasticsearch 400 response is permanent.**

Elasticsearch can return status 400 for operational conditions such as index_closed_exception. During a temporary index closure, this classifier marks every affected item permanent; the handler then acknowledges and drops each JetStream message, so reopening the index cannot recover those updates. Classify permanence using a narrow allowlist of document-level error types and treat unknown, index-level, and operational errors as retryable by default.

<details>

```
In `pkg/searchengine/classify.go` around line 87: Elasticsearch can return status 400 for operational conditions such as index_closed_exception. During a temporary index closure, this classifier marks every affected item permanent; the handler then acknowledges and drops each JetStream message, so reopening the index cannot recover those updates. Classify permanence using a narrow allowlist of document-level error types and treat unknown, index-level, and operational errors as retryable by default.

```

</details>

search-sync-worker/spotlight_org.go
```diff
@@ -74,7 +78,16 @@ func (c *spotlightOrgCollection) MappingUpdate() (string, json.RawMessage) {
 // (last-wins) and emits one ES _update per unique sectId with doc_as_upsert:true.
 // Doc-merge + omitempty means partial-field publishes preserve stored values for
 // unset fields.
+// BuildAction satisfies Collection. Sequence 0 can only lose to a stored sequence, so a
+// caller with no stream metadata degrades to "skip" rather than clobbering a newer row;
+// the real path is BuildActionSeq, which Handler uses.
 func (c *spotlightOrgCollection) BuildAction(data []byte) ([]searchengine.BulkAction, error) {
+	return c.BuildActionSeq(data, 0)
```
julianshen: _🏷️ rolling-deploy_ | _🔴 Blocking_ | _`distributed-system`_

**Old pods can bypass the new ordering guard during rollout.**

The previous worker writes spotlight-org rows with unguarded doc_as_upsert. During a rolling deployment, an old pod can hold seq 100 while a new pod applies seq 101 and records orgSeq=101; the old pod can then overwrite fields with seq 100 while leaving orgSeq at 101. Equal-sequence retries are subsequently skipped, so the stale state may persist. Deployment must stop and drain all old consumers before new consumers process HR events, or introduce an isolated migration/index and cutover that prevents guarded and unguarded writers from targeting the same documents.

<details>

```
In `search-sync-worker/spotlight_org.go` around line 85: The previous worker writes spotlight-org rows with unguarded doc_as_upsert. During a rolling deployment, an old pod can hold seq 100 while a new pod applies seq 101 and records orgSeq=101; the old pod can then overwrite fields with seq 100 while leaving orgSeq at 101. Equal-sequence retries are subsequently skipped, so the stale state may persist. Deployment must stop and drain all old consumers before new consumers process HR events, or introduce an isolated migration/index and cutover that prevents guarded and unguarded writers from targeting the same documents.

```

</details>

search-sync-worker/spotlight_org.go
```diff
@@ -112,10 +125,16 @@ func (c *spotlightOrgCollection) BuildAction(data []byte) ([]searchengine.BulkAc
...
 }
 
-func buildSpotlightOrgUpdateBody(row *SpotlightOrgIndex) (json.RawMessage, error) {
+func buildSpotlightOrgUpdateBody(row *SpotlightOrgIndex, streamSeq uint64) (json.RawMessage, error) {
+	// scripted_upsert + an empty seed: the script owns both the insert and the merge, so a
+	// new sectId records its sequence instead of landing without one.
 	body := map[string]any{
-		"doc":           row,
-		"doc_as_upsert": true,
+		"scripted_upsert": true,
```
julianshen: _🏷️ rolling-deployment_ | _🔴 Blocking_ | _`elasticsearch`_

**Old pods can bypass the sequence guard during a rolling deployment.**

New pods use the scripted sequence guard, but old pods still issue unguarded `doc_as_upsert` updates to the same fixed spotlight-org index. While both versions share the durable consumer, an old pod can finish a stale update after a new pod stored sequence N, overwrite the newer fields, and leave `orgSeq` at N. A redelivery of sequence N then no-ops, preserving the stale values. Deploy this transition without overlapping writers—drain and stop all old workers before starting new ones—or use a staged/versioned-index migration that prevents old code from writing the guarded index.

<details>

```
In `search-sync-worker/spotlight_org.go` around line 132: New pods use the scripted sequence guard, but old pods still issue unguarded `doc_as_upsert` updates to the same fixed spotlight-org index. While both versions share the durable consumer, an old pod can finish a stale update after a new pod stored sequence N, overwrite the newer fields, and leave `orgSeq` at N. A redelivery of sequence N then no-ops, preserving the stale values. Deploy this transition without overlapping writers—drain and stop all old workers before starting new ones—or use a staged/versioned-index migration that prevents old code from writing the guarded index.

```

</details>

search-sync-worker/handler.go
```diff
@@ -112,26 +116,108 @@ func (h *Handler) AddWithContext(ctx context.Context, msg jetstream.Msg) {
...
+func (h *Handler) buildActions(msg jetstream.Msg, data []byte) ([]searchengine.BulkAction, error) {
+	sc, ok := h.collection.(sequencedCollection)
+	if !ok {
+		return h.collection.BuildAction(data)
+	}
+	return sc.BuildActionSeq(data, streamSequence(msg))
+}
+
+// streamSequence returns the message's position in its stream, or 0 when the metadata
+// can't be read. Zero is the fail-closed value: it loses to any stored sequence, so a
```
julianshen, line 186: _🏷️ data-loss_ | _🔴 Blocking_ | _`elasticsearch`, `nats`, `silent-failure` (3 rules)_

**Metadata failures are converted into acknowledged no-op writes**

When Msg.Metadata() fails or returns nil, streamSequence returns 0. SpotlightOrgUpsertScript then sees incoming == stored for a new or legacy document (or incoming < stored for an existing document), sets ctx.op to none, and Elasticsearch returns a successful bulk item. FlushBatch consequently acknowledges the source message even though its update was never applied. The added test explicitly confirms this path. Return an error for missing/zero sequence and settle it as retryable before buffering; do not pass it through the existing BuildAction error branch, which acknowledges validation errors.

<details>

- **`elasticsearch`**:169: When `Metadata()` fails, `streamSequence` returns zero. The spotlight-org script compares `incoming > stored`, so sequence zero no-ops even for a brand-new document; Elasticsearch reports that no-op as successful and the handler acknowledges the message. This silently loses the update instead of surfacing the ordering failure. For sequenced collections, detect missing/zero sequence before building the bulk action, log it, and settle the message as a delayed retry rather than routing it through the ordinary validation-error ACK path.
- **`nats`**:168: When Metadata fails or is nil, streamSequence returns zero and spotlight-org still sends a scripted update. Zero loses even against an empty stored sequence, Elasticsearch reports the no-op as successful, and FlushBatch ACKs the source message. Thus malformed or unavailable JetStream metadata is reported as success while the HR update is silently omitted. Return the metadata error from the sequenced build path and settle or surface it explicitly rather than constructing an action with sequence zero.

</details>

<details>

```
In `search-sync-worker/handler.go` around line 165: When Msg.Metadata() fails or returns nil, streamSequence returns 0. SpotlightOrgUpsertScript then sees incoming == stored for a new or legacy document (or incoming < stored for an existing document), sets ctx.op to none, and Elasticsearch returns a successful bulk item. FlushBatch consequently acknowledges the source message even though its update was never applied. The added test explicitly confirms this path. Return an error for missing/zero sequence and settle it as retryable before buffering; do not pass it through the existing BuildAction error branch, which acknowledges validation errors.

```

</details>

pkg/searchengine/classify.go
```diff
@@ -47,3 +47,44 @@ func IsBulkItemSuccess(action ActionType, result BulkResult) bool {
...
+func IsBulkItemPermanent(result BulkResult) bool {
+	return result.Status == 400
```
julianshen, line 89: _🏷️ message-settlement_ | _🔴 Blocking_ | _`nats`_

**Blanket 400 handling acknowledges retryable operational failures.**

IsBulkItemPermanent classifies every Elasticsearch item status 400 as permanent. FlushBatch wraps that result with errcode.Permanent, and jsretry then ACKs the JetStream message. Elasticsearch can return 400 for recoverable operational states such as index_closed_exception; reopening the index would allow the identical action to succeed, but the source event has already been discarded. Restrict permanent handling to explicitly irrecoverable document error types and retry unknown or operational 400 responses.

<details>

```
In `pkg/searchengine/classify.go` around line 89: IsBulkItemPermanent classifies every Elasticsearch item status 400 as permanent. FlushBatch wraps that result with errcode.Permanent, and jsretry then ACKs the JetStream message. Elasticsearch can return 400 for recoverable operational states such as index_closed_exception; reopening the index would allow the identical action to succeed, but the source event has already been discarded. Restrict permanent handling to explicitly irrecoverable document error types and retry unknown or operational 400 responses.

```

</details>

julianshen, line 89: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**HTTP 400 does not always mean a permanently invalid document.**

Elasticsearch can return per-item status 400 for recoverable operational failures such as `index_closed_exception`, not only malformed documents. `bulkItemError` wraps every result classified here as permanent, causing `jsretry` to Ack and permanently discard the source message even though the identical write could succeed after the index reopens. Classify permanent failures using a narrow allowlist of document-local error types rather than the status alone.

<details>

```
In `pkg/searchengine/classify.go` around line 89: Elasticsearch can return per-item status 400 for recoverable operational failures such as `index_closed_exception`, not only malformed documents. `bulkItemError` wraps every result classified here as permanent, causing `jsretry` to Ack and permanently discard the source message even though the identical write could succeed after the index reopens. Classify permanent failures using a narrow allowlist of document-local error types rather than the status alone.

```

</details>

search-sync-worker/handler.go
```diff
@@ -148,30 +234,37 @@ func (h *Handler) Flush(ctx context.Context) {
...
+			// leaves the Ack-drop of a permanent failure otherwise unlogged.
+			disposition := "retry"
+			if _, permanent := errcode.IsPermanent(itemErr); permanent {
+				disposition = "drop"
+			}
+			slog.ErrorContext(p.ctx, "bulk item failed",
 				"status", results[i].Status,
 				"error", results[i].Error,
+				"errorType", results[i].ErrorType,
+				"backpressure", backpressure,
```
julianshen: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**The first failed action can cause retryable sibling writes to be dropped.**

A source message can fan out to multiple actions; spotlight-org emits one action per unique `sectId`. This loop stops at the first failure, so if that result is classified permanent while a later action failed transiently with 429/500, `settle` Acks the entire source message and the later row is lost. Spotlight-org action order is derived from map iteration, making the outcome nondeterministic. Inspect every failed action, retry whenever any failure is retryable, and only Ack-drop when all failures are permanent; aggregate backpressure across retryable failures.

<details>

```
In `search-sync-worker/handler.go` around line 260: A source message can fan out to multiple actions; spotlight-org emits one action per unique `sectId`. This loop stops at the first failure, so if that result is classified permanent while a later action failed transiently with 429/500, `settle` Acks the entire source message and the later row is lost. Spotlight-org action order is derived from map iteration, making the outcome nondeterministic. Inspect every failed action, retry whenever any failure is retryable, and only Ack-drop when all failures are permanent; aggregate backpressure across retryable failures.

```

</details>

pkg/searchindex/userroomdoc.go
```diff
@@ -136,3 +136,37 @@ func BuildRemoveRoomUpdateBody(rid string, ts int64) (json.RawMessage, error) {
...
+// from params.doc and keep their stored values.
+//
+// Runs under scripted_upsert so a brand-new sectId also records its sequence;
+// ctx._source starts as the empty upsert doc and stored reads 0.
+//
+// Operational note: the guard compares sequences from ONE stream. If the HR
+// stream is ever recreated, sequences restart at 1 and will lose to the stored
+// values, silently dropping every subsequent update — clear orgSeq (or reindex)
+// as part of any such rebuild.
+const SpotlightOrgUpsertScript = `long stored = ctx._source.orgSeq == null ? 0L : ((Number)ctx._source.orgSeq).longValue(); ` +
```
julianshen: _🏷️ data-loss_ | _🟠 Warning_ | _`distributed-system`, `nats`, `silent-failure` (3 rules)_

**A recreated stream makes valid updates look successfully stale**

The ordering token contains only the stream sequence, which restarts when the HR stream is recreated. Existing documents retain larger orgSeq values, so valid updates from the new stream execute ctx.op = 'none'; Elasticsearch reports those no-ops as successful and the worker acknowledges the messages. Existing rows therefore remain stale until the recreated stream exceeds each stored sequence, potentially masking a large or effectively permanent outage. Include a stream incarnation in the ordering token, or enforce and automate clearing/reindexing orgSeq as part of stream recreation rather than relying only on this code comment.

<details>

- **`nats`**:163: The new last-write-wins token contains only the JetStream stream sequence, and the comment confirms that sequence resets after stream recreation. Existing spotlight-org documents then retain larger orgSeq values, so every new chat.hr.<site>.employees.upsert action becomes an Elasticsearch no-op and is ACKed as successful indefinitely. Include a stream epoch or durable source version in the ordering token, or enforce an automated startup/migration gate that clears or rebuilds orgSeq before the recreated stream can be consumed.
- **`distributed-system`**:167: If the HR stream is recreated while the search index survives, its sequence restarts at 1 while documents retain larger orgSeq values. Every new update then takes the script's no-op branch and is acknowledged as successful, silently freezing the index. A code comment does not enforce recovery. Include a stream-generation identifier in the ordering token and compare generation before sequence, or detect generation changes at startup and require/reset or reindex the destination before consuming.

</details>

<details>

```
In `pkg/searchindex/userroomdoc.go` around line 167: The ordering token contains only the stream sequence, which restarts when the HR stream is recreated. Existing documents retain larger orgSeq values, so valid updates from the new stream execute ctx.op = 'none'; Elasticsearch reports those no-ops as successful and the worker acknowledges the messages. Existing rows therefore remain stale until the recreated stream exceeds each stored sequence, potentially masking a large or effectively permanent outage. Include a stream incarnation in the ordering token, or enforce and automate clearing/reindexing orgSeq as part of stream recreation rather than relying only on this code comment.

```

</details>

search-sync-worker/handler.go
```diff
@@ -112,26 +116,108 @@ func (h *Handler) AddWithContext(ctx context.Context, msg jetstream.Msg) {
...
+	return sc.BuildActionSeq(data, streamSequence(msg))
+}
+
+// streamSequence returns the message's position in its stream, or 0 when the metadata
+// can't be read. Zero is the fail-closed value: it loses to any stored sequence, so a
+// message we can't order is skipped rather than allowed to overwrite a newer row.
+func streamSequence(msg jetstream.Msg) uint64 {
+	md, err := msg.Metadata()
+	if err != nil || md == nil {
+		return 0
```
julianshen, line 191: _🏷️ failure-handling_ | _🟠 Warning_ | _`distributed-system`_

**Metadata errors silently acknowledge an unapplied HR event.**

When Msg.Metadata() fails, streamSequence returns zero. The scripted upsert then compares incoming 0 with stored 0 or greater, performs a no-op, and the successful bulk result causes the source message to be acknowledged. Thus a metadata failure permanently loses even a brand-new row. Treat missing sequence metadata as a retryable processing error and settle it through jsretry rather than constructing a successful no-op action; if retries exhaust, route it through the consumer's terminal/DLQ handling.

<details>

```
In `search-sync-worker/handler.go` around line 170: When Msg.Metadata() fails, streamSequence returns zero. The scripted upsert then compares incoming 0 with stored 0 or greater, performs a no-op, and the successful bulk result causes the source message to be acknowledged. Thus a metadata failure permanently loses even a brand-new row. Treat missing sequence metadata as a retryable processing error and settle it through jsretry rather than constructing a successful no-op action; if retries exhaust, route it through the consumer's terminal/DLQ handling.

```

</details>

pkg/searchindex/userroomdoc.go
```diff
@@ -136,3 +136,37 @@ func BuildRemoveRoomUpdateBody(rid string, ts int64) (json.RawMessage, error) {
...
+// after a newer one and leave the older row stored.
+//
+// Merging params.doc key-by-key reproduces doc-merge semantics — the row is
+// marshalled with omitempty, so fields absent from a partial publish are absent
+// from params.doc and keep their stored values.
+//
+// Runs under scripted_upsert so a brand-new sectId also records its sequence;
+// ctx._source starts as the empty upsert doc and stored reads 0.
+//
+// Operational note: the guard compares sequences from ONE stream. If the HR
```
julianshen: _🏷️ index-lifecycle_ | _🟠 Warning_ | _`elasticsearch`_

**Recreating the HR stream silently blocks later organization updates.**

The stored `orgSeq` is compared without a stream-incarnation token. After the HR stream is recreated or the worker is pointed at a replacement stream/domain, sequences restart at one and valid updates no-op until they exceed historical values; those no-ops are acknowledged as successes. The comment identifies this failure mode but does not enforce recovery. Automate removal/reindexing of `orgSeq` before consuming from a recreated stream, or include a configured stream epoch in the stored ordering token so a new incarnation supersedes the old one.

<details>

```
In `pkg/searchindex/userroomdoc.go` around line 163: The stored `orgSeq` is compared without a stream-incarnation token. After the HR stream is recreated or the worker is pointed at a replacement stream/domain, sequences restart at one and valid updates no-op until they exceed historical values; those no-ops are acknowledged as successes. The comment identifies this failure mode but does not enforce recovery. Automate removal/reindexing of `orgSeq` before consuming from a recreated stream, or include a configured stream epoch in the stored ordering token so a new incarnation supersedes the old one.

```

</details>

search-sync-worker/main.go
```diff
@@ -386,31 +401,82 @@ func pushMappings(ctx context.Context, engine searchengine.SearchEngine, collect
...
+}
+
+// run detaches whatever the handler has buffered and flushes it in the background,
+// blocking for a slot once `depth` flushes are already in flight. Taking the batch first
+// frees the buffer immediately, so ActionCount() reads 0 on return.
+func (p *flushPipeline) run(ctx context.Context, h *Handler) {
+	batch := h.Take()
+	if batch == nil {
+		return
+	}
```
julianshen, line 450: _🏷️ consumer-lifecycle_ | _🟠 Warning_ | _`nats`_

**Queued pipeline batches can expire AckWait before flushing.**

run detaches the batch before blocking on the pipeline slot. When all slots contain slow or hung Elasticsearch requests, this newly fetched JetStream batch has no flusher and receives no InProgress acknowledgements; it can remain there beyond AckWait, causing duplicate redelivery and consuming MaxDeliver before its first bulk attempt begins. The same blocking send prevents runConsumer from observing stopCh. Bound each bulk request with a timeout and either heartbeat all queued/in-flight messages or configure and validate AckWait against the maximum queue-plus-request duration; slot acquisition should also be cancellation-aware.

<details>

```
In `search-sync-worker/main.go` around line 435: run detaches the batch before blocking on the pipeline slot. When all slots contain slow or hung Elasticsearch requests, this newly fetched JetStream batch has no flusher and receives no InProgress acknowledgements; it can remain there beyond AckWait, causing duplicate redelivery and consuming MaxDeliver before its first bulk attempt begins. The same blocking send prevents runConsumer from observing stopCh. Bound each bulk request with a timeout and either heartbeat all queued/in-flight messages or configure and validate AckWait against the maximum queue-plus-request duration; slot acquisition should also be cancellation-aware.

```

</details>

search-sync-worker/main.go
```diff
@@ -465,16 +531,21 @@ func runConsumer(
...
 
-// checkBatchAckCoupling warns when bulkBatchSize exceeds maxAckPending: a 1:1 collection then
-// stalls at maxAckPending un-acked messages before the size-based flush can fire, undersizing every batch. Fan-out collections are unaffected. Empty string = no issue.
-func checkBatchAckCoupling(bulkBatchSize, maxAckPending int) string {
-	if bulkBatchSize > maxAckPending {
+// checkBatchAckCoupling warns when the pipelined batches can't fit under maxAckPending.
+// depth batches can be in flight while one more fills, so a 1:1 collection needs headroom
+// for (depth+1)*bulkBatchSize un-acked messages; below that it stalls before the size-based
+// flush can fire, undersizing every batch. Fan-out collections are unaffected. "" = no issue.
+func checkBatchAckCoupling(bulkBatchSize, maxAckPending, pipelineDepth int) string {
```
julianshen, line 556: _🏷️ backpressure_ | _🟠 Warning_ | _`nats`_

**Ack-pending sizing ignores replicas sharing the durable consumer.**

MaxAckPending belongs to the JetStream consumer, not to each pod bound to that durable. The code explicitly expects multiple pods to share the durable, but needed only reserves (depth+1)*bulkBatchSize once. With two replicas, each can hold an in-flight batch and partial next batch until the shared 1500-message ceiling is reached, preventing either pod from reaching its configured per-pod pipeline depth. Size and validate the consumer-wide limit using the number of competing replicas, or reduce/document pipeline depth as an aggregate consumer budget.

<details>

```
In `search-sync-worker/main.go` around line 538: MaxAckPending belongs to the JetStream consumer, not to each pod bound to that durable. The code explicitly expects multiple pods to share the durable, but needed only reserves (depth+1)*bulkBatchSize once. With two replicas, each can hold an in-flight batch and partial next batch until the shared 1500-message ceiling is reached, preventing either pod from reaching its configured per-pod pipeline depth. Size and validate the consumer-wide limit using the number of competing replicas, or reduce/document pipeline depth as an aggregate consumer budget.

```

</details>

search-sync-worker/main.go
```diff
@@ -113,6 +121,12 @@ func main() {
 		slog.Error("invalid config", "name", "BULK_FLUSH_INTERVAL", "value", cfg.BulkFlushInterval, "reason", "must be > 0")
 		os.Exit(1)
 	}
+	// A non-positive depth would make the pipeline's slot channel unbuffered and park the
+	// consumer on its first flush forever.
+	if cfg.PipelineDepth <= 0 {
+		slog.Error("invalid config", "name", "PIPELINE_DEPTH", "value", cfg.PipelineDepth, "reason", "must be > 0")
```
julianshen, line 129: _⚡ memory-exhaustion_ | _🟠 Warning_ | _`oom`_

**PIPELINE_DEPTH has no upper bound on retained batches**

Validation rejects only non-positive values. With a large PIPELINE_DEPTH, a backlog combined with slow or stalled Elasticsearch requests lets flushPipeline start that many goroutines, each retaining its flushBatch, source messages, action documents, and request-related memory until Bulk returns; this is multiplied across collections and replicas. CONSUMER_MAX_ACK_PENDING may limit this only when configured conservatively, but the coupling check merely warns, can overflow for extreme values, and fan-out messages can produce oversized batches. A high but parseable deployment value can therefore cause GC thrashing, container eviction, or process OOM. Enforce a conservative maximum at startup, preferably also constrained by the ack-pending and per-pod memory budgets, and use checked arithmetic in the coupling calculation.

<details>

```
In `search-sync-worker/main.go` around line 127: Validation rejects only non-positive values. With a large PIPELINE_DEPTH, a backlog combined with slow or stalled Elasticsearch requests lets flushPipeline start that many goroutines, each retaining its flushBatch, source messages, action documents, and request-related memory until Bulk returns; this is multiplied across collections and replicas. CONSUMER_MAX_ACK_PENDING may limit this only when configured conservatively, but the coupling check merely warns, can overflow for extreme values, and fan-out messages can produce oversized batches. A high but parseable deployment value can therefore cause GC thrashing, container eviction, or process OOM. Enforce a conservative maximum at startup, preferably also constrained by the ack-pending and per-pod memory budgets, and use checked arithmetic in the coupling calculation.

```

</details>

search-sync-worker/main.go
```diff
@@ -465,16 +531,21 @@ func runConsumer(
...
-// checkBatchAckCoupling warns when bulkBatchSize exceeds maxAckPending: a 1:1 collection then
-// stalls at maxAckPending un-acked messages before the size-based flush can fire, undersizing every batch. Fan-out collections are unaffected. Empty string = no issue.
-func checkBatchAckCoupling(bulkBatchSize, maxAckPending int) string {
-	if bulkBatchSize > maxAckPending {
+// checkBatchAckCoupling warns when the pipelined batches can't fit under maxAckPending.
+// depth batches can be in flight while one more fills, so a 1:1 collection needs headroom
+// for (depth+1)*bulkBatchSize un-acked messages; below that it stalls before the size-based
+// flush can fire, undersizing every batch. Fan-out collections are unaffected. "" = no issue.
+func checkBatchAckCoupling(bulkBatchSize, maxAckPending, pipelineDepth int) string {
+	needed := (pipelineDepth + 1) * bulkBatchSize
```
julianshen, line 557: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**The coupling warning incorrectly says size-based flushes cannot fire.**

The `(depth+1)*bulkBatchSize` value is headroom for prebuilding another batch while every pipeline slot is occupied, not the minimum required for the size trigger. For example, with depth 2, bulk size 500, and MaxAckPending 1000, the first and second 500-message batches both reach the size trigger; later batches can do so after either in-flight batch is acknowledged. Only `bulkBatchSize > maxAckPending` makes the trigger unreachable. Keep the larger calculation as a pipeline-utilization advisory if desired, but change the warning and compose documentation so they do not claim batches must wait for the timer or remain undersized.

<details>

```
In `search-sync-worker/main.go` around line 539: The `(depth+1)*bulkBatchSize` value is headroom for prebuilding another batch while every pipeline slot is occupied, not the minimum required for the size trigger. For example, with depth 2, bulk size 500, and MaxAckPending 1000, the first and second 500-message batches both reach the size trigger; later batches can do so after either in-flight batch is acknowledged. Only `bulkBatchSize > maxAckPending` makes the trigger unreachable. Keep the larger calculation as a pipeline-utilization advisory if desired, but change the warning and compose documentation so they do not claim batches must wait for the timer or remain undersized.

```

</details>

search-sync-worker/spotlight_org.go
```diff
@@ -12,6 +12,37 @@ import (
...
+// row stored.
+//
+// Merging params.doc key-by-key reproduces doc-merge semantics — the row is marshalled
+// with omitempty, so fields absent from a partial publish keep their stored values.
+//
+// Runs under scripted_upsert so a brand-new sectId also records its sequence;
+// ctx._source starts as the empty upsert doc and stored reads 0.
+//
+// Operational note: the guard compares sequences from ONE stream. If the HR stream is
+// ever recreated, sequences restart at 1 and will lose to the stored values, silently
```
julianshen, line 37: _🏷️ ordering_ | _🔴 Blocking_ | _`distributed-system`, `elasticsearch`, `nats`, `silent-failure` (4 rules)_

**Stream recreation causes future updates to be silently discarded.**

The persisted ordering token contains only the stream sequence. If the HR stream is recreated after a document stored orgSeq 500000, the replacement stream starts at sequence 1; every new event compares lower, becomes an Elasticsearch no-op, and is ACKed as successfully processed until the new stream surpasses 500000. A comment-only runbook does not prevent this failure during recovery or migration. Include a stream generation/epoch in the ordering token and compare generation before sequence, or enforce a startup migration that clears/reindexes orgSeq before consumption and refuses readiness when the stored generation differs.

<details>

- **`nats`**:35: The ordering key contains only the JetStream stream sequence, which restarts at 1 when the HR stream is recreated. If Elasticsearch retains `orgSeq=500`, a post-recreation update at sequence 1 becomes a successful no-op and is ACKed, losing that update until incoming sequences exceed the stored value. A code comment does not enforce the required reset. Include a stream-generation identity in the ordering guard, or automate and verify clearing/reindexing `orgSeq` before consuming from a recreated stream.
- **`silent-failure`**:35: The ordering guard stores only the stream sequence. After the HR stream is recreated, its sequence restarts at 1 and new updates compare lower than existing orgSeq values. Elasticsearch returns each rejected update as a successful no-op, so the worker acknowledges messages while leaving stale documents indefinitely. Make stream-generation changes detectable and fail closed, or automatically clear/reindex orgSeq before consuming from a recreated stream rather than relying on a manual operational note.
- **`elasticsearch`**:35: The persisted `orgSeq` is compared without a stream generation or epoch. After the HR stream is recreated, sequences restart at 1 and every update loses to the old stored values; Elasticsearch returns successful no-ops and the worker acknowledges all events. A source-code comment does not enforce the required migration. Include a configured stream epoch in the guard, or make stream recreation atomically clear/rebuild `orgSeq` before consumption resumes.

</details>

<details>

```
In `search-sync-worker/spotlight_org.go` around line 37: The persisted ordering token contains only the stream sequence. If the HR stream is recreated after a document stored orgSeq 500000, the replacement stream starts at sequence 1; every new event compares lower, becomes an Elasticsearch no-op, and is ACKed as successfully processed until the new stream surpasses 500000. A comment-only runbook does not prevent this failure during recovery or migration. Include a stream generation/epoch in the ordering token and compare generation before sequence, or enforce a startup migration that clears/reindexes orgSeq before consumption and refuses readiness when the stored generation differs.

```

</details>

search-sync-worker/spotlight_org.go
```diff
@@ -12,6 +12,37 @@ import (
 	"github.com/hmchangw/chat/pkg/subject"
 )
 
+// spotlightOrgUpsertScriptID is the stored-script id for the last-write-wins guard.
+// Bump the suffix if the source ever changes incompatibly, so a rolling deploy doesn't
+// leave old and new pods sharing one mutated definition.
+const spotlightOrgUpsertScriptID = "search-sync-spotlight-org-upsert-v1"
```
julianshen, line 18: _🏷️ rolling-deploy_ | _🔴 Blocking_ | _`distributed-system`, `elasticsearch` (2 rules)_

**Old workers can bypass the new ordering guard during rollout.**

A normal multi-replica rolling deploy mixes this worker with the previous implementation, whose spotlight-org writes use unguarded doc_as_upsert. A new pod can apply sequence 200 and store orgSeq=200, after which an old pod completing sequence 199 can overwrite fields while leaving orgSeq at 200. Subsequent guarded writes treat sequence 200 or lower as stale, so the index can retain stale data. The versioned stored-script ID does not protect against requests emitted by old binaries. Use a staged stop-and-drain deployment, or write to a new index/durable and atomically cut over only after old workers are gone; document and test this compatibility boundary.

<details>

- **`elasticsearch`**:39: Existing spotlight-org documents have no `orgSeq`, and old worker pods continue issuing unguarded `doc_as_upsert` updates. During a rolling deployment, a new pod can apply sequence N+1 and store `orgSeq=N+1`, after which an old pod can apply delayed sequence N, overwrite the business fields, and leave `orgSeq=N+1` intact. The guarded retry is then unable to repair the stale document. Roll this out in stages: keep pipeline depth at 1 until all old writers are drained, initialize sequence state through a coordinated replay/reindex or versioned-index migration, and only then enable concurrent flushing.

</details>

<details>

```
In `search-sync-worker/spotlight_org.go` around line 18: A normal multi-replica rolling deploy mixes this worker with the previous implementation, whose spotlight-org writes use unguarded doc_as_upsert. A new pod can apply sequence 200 and store orgSeq=200, after which an old pod completing sequence 199 can overwrite fields while leaving orgSeq at 200. Subsequent guarded writes treat sequence 200 or lower as stale, so the index can retain stale data. The versioned stored-script ID does not protect against requests emitted by old binaries. Use a staged stop-and-drain deployment, or write to a new index/durable and atomically cut over only after old workers are gone; document and test this compatibility boundary.

```

</details>

search-sync-worker/spotlight_org.go
```diff
@@ -12,6 +12,37 @@ import (
...
+// Merging params.doc key-by-key reproduces doc-merge semantics — the row is marshalled
+// with omitempty, so fields absent from a partial publish keep their stored values.
+//
+// Runs under scripted_upsert so a brand-new sectId also records its sequence;
+// ctx._source starts as the empty upsert doc and stored reads 0.
+//
+// Operational note: the guard compares sequences from ONE stream. If the HR stream is
+// ever recreated, sequences restart at 1 and will lose to the stored values, silently
+// dropping every subsequent update — clear orgSeq (or reindex) as part of any rebuild.
+const spotlightOrgUpsertScript = `long stored = ctx._source.orgSeq == null ? 0L : ((Number)ctx._source.orgSeq).longValue(); ` +
```
julianshen, line 39: _🎯 concurrency_ | _🔴 Blocking_ | _`performance-and-concurrency`, `tgd-review` (2 rules)_

**A document-level sequence guard loses out-of-order partial updates.**

Concurrent flushes can reverse two partial updates for the same sectId. For example, seq 100 changes deptId and seq 101 changes only sectName. If seq 101 reaches Elasticsearch first, it preserves the old deptId and records orgSeq=101; seq 100 then becomes a no-op, permanently losing its deptId change. This contradicts the documented partial-merge semantics and is reachable with the default pipeline depth or multiple pods. Track sequence per field, require each event to contain authoritative full state, or otherwise preserve per-document stream order. Add a test that applies a newer partial update before an older update with disjoint fields.

<details>

- **`tgd-review`**:38: The collection explicitly supports partial publishes whose omitted fields preserve existing values, but the script rejects an older event wholesale once a newer sequence has landed. If sequence 10 updates only deptName and sequence 11 updates only sectName, concurrent flushes can apply 11 first and then skip 10, permanently losing the deptName update. Whole-document last-write-wins is only safe if every event contains a complete snapshot. Either guarantee and validate full rows, preserve processing order, or track ordering per field; add an out-of-order partial-update test.

</details>

<details>

```
In `search-sync-worker/spotlight_org.go` around line 39: Concurrent flushes can reverse two partial updates for the same sectId. For example, seq 100 changes deptId and seq 101 changes only sectName. If seq 101 reaches Elasticsearch first, it preserves the old deptId and records orgSeq=101; seq 100 then becomes a no-op, permanently losing its deptId change. This contradicts the documented partial-merge semantics and is reachable with the default pipeline depth or multiple pods. Track sequence per field, require each event to contain authoritative full state, or otherwise preserve per-document stream order. Add a test that applies a newer partial update before an older update with disjoint fields.

```

</details>

julianshen (commented):
tGD inline review

search-sync-worker/handler.go
```diff
@@ -140,7 +214,7 @@ func (h *Handler) Flush(ctx context.Context) {
 		slog.Error("bulk request failed", "error", err, "actions", len(actions))
```
julianshen, line 242: _🏷️ backpressure_ | _🟠 Warning_ | _`elasticsearch`_

**Request-level Elasticsearch overload uses the fast retry schedule.**

Backpressure classification is applied only to per-item results. If Elasticsearch rejects the entire _bulk request with HTTP 429 or a circuit-breaker error, Bulk returns an error here and settleAll uses DefaultBackoff, retrying after one second instead of the new 30-second backpressure delay. Preserve the request status/error type and select BackpressureBackoff for request-level overload failures.

<details>

```
In `search-sync-worker/handler.go` around line 214: Backpressure classification is applied only to per-item results. If Elasticsearch rejects the entire _bulk request with HTTP 429 or a circuit-breaker error, Bulk returns an error here and settleAll uses DefaultBackoff, retrying after one second instead of the new 30-second backpressure delay. Preserve the request status/error type and select BackpressureBackoff for request-level overload failures.

```

</details>

julianshen (commented):
tGD inline review

julianshen:
## Needs clarification

Is deployment guaranteed to stop and drain every old worker before starting this version?
Reply in this thread, or post: `answer clar_bq47kxijm6qhm6mwlvjayqco2a: <your answer>`

julianshen:
## Needs clarification

Will deployment stop and drain every old worker before starting new workers?
Reply in this thread, or post: `answer clar_bnodeei6dnbsm4sgmy7xuqeodx: <your answer>`

search-sync-worker/handler.go
```diff
@@ -194,35 +267,57 @@ func (h *Handler) Flush(ctx context.Context) {
...
-			if allOK {
-				// Log only the first failed item per message; the counter above still
-				// sees every failed action.
-				slog.Error("bulk item failed",
-					"status", results[i].Status,
-					"error", results[i].Error,
-					"docID", actions[i].DocID,
-					"index", actions[i].Index,
-				)
+			if failed {
```
julianshen, line 280: _🏷️ partial-failure_ | _🔴 Blocking_ | _`silent-failure`_

**The first failed item can hide later retryable failures.**

A source message can produce multiple bulk actions, but after the first failure this loop skips classification of all later failures. If the first failure is considered permanent and a later item returns 429 or 500, the whole source message is ACK-dropped and the later update is never retried. Inspect every result for the message, retry if any failure is retryable, and classify the message as permanent only when every failed item is permanent; select backpressure pacing if any retryable failure indicates backpressure.

<details>

```
In `search-sync-worker/handler.go` around line 280: A source message can produce multiple bulk actions, but after the first failure this loop skips classification of all later failures. If the first failure is considered permanent and a later item returns 429 or 500, the whole source message is ACK-dropped and the later update is never retried. Inspect every result for the message, retry if any failure is retryable, and classify the message as permanent only when every failed item is permanent; select backpressure pacing if any retryable failure indicates backpressure.

```

</details>

search-sync-worker/handler.go
```diff
@@ -146,18 +148,89 @@ func (h *Handler) AddWithContext(ctx context.Context, msg jetstream.Msg) {
...
+// Only the status and error type go into the message; the document body never belongs
+// in an error that reaches the server log.
+func bulkItemError(result searchengine.BulkResult) error {
+	if searchengine.IsBulkItemPermanent(result) {
+		return errcode.Permanent(errcode.BadRequest(
+			fmt.Sprintf("search backend rejected the document: status %d %q", result.Status, result.ErrorType)))
+	}
+	return fmt.Errorf("bulk item failed: status %d %q", result.Status, result.ErrorType)
+}
+
```
julianshen, line 165: _🏷️ backpressure_ | _🟠 Warning_ | _`elasticsearch`_

**Request-level 429s still use the aggressive default retry schedule**

Backpressure classification only accepts BulkResult items. If Elasticsearch rejects the entire _bulk request with a top-level 429 or circuit_breaking_exception, Store.Bulk returns an error and the whole-batch failure path uses DefaultBackoff, retrying as soon as one second rather than the new backpressure schedule. Preserve HTTP status and Elasticsearch error type in a typed Bulk error, classify request-level overload, and use BackpressureBackoff when settling every message in that batch.

<details>

```
In `search-sync-worker/handler.go` around line 165: Backpressure classification only accepts BulkResult items. If Elasticsearch rejects the entire _bulk request with a top-level 429 or circuit_breaking_exception, Store.Bulk returns an error and the whole-batch failure path uses DefaultBackoff, retrying as soon as one second rather than the new backpressure schedule. Preserve HTTP status and Elasticsearch error type in a typed Bulk error, classify request-level overload, and use BackpressureBackoff when settling every message in that batch.

```

</details>

julianshen (commented):
tGD inline review

search-sync-worker/handler.go
```diff
@@ -146,18 +148,89 @@ func (h *Handler) AddWithContext(ctx context.Context, msg jetstream.Msg) {
...
+	if !ok {
+		return h.collection.BuildAction(data)
+	}
+	return sc.BuildActionSeq(data, streamSequence(msg))
+}
+
+// streamSequence returns the message's position in its stream, or 0 when the metadata
+// can't be read. Zero is the fail-closed value: it loses to any stored sequence, so a
+// message we can't order is skipped rather than allowed to overwrite a newer row.
+func streamSequence(msg jetstream.Msg) uint64 {
```
julianshen, line 188: _🏷️ delivery-guarantees_ | _🔴 Blocking_ | _`distributed-system`_

**Metadata failures acknowledge unapplied HR events.**

When Msg.Metadata() fails or is nil, streamSequence returns 0. The spotlight-org script compares incoming 0 against stored 0 or greater, performs a no-op, and Elasticsearch reports success; FlushBatch then ACKs the source message although no update was applied. Treat a missing or zero sequence as a retryable processing failure and settle it separately from BuildAction validation errors, which are intentionally ACK-dropped.

<details>

```
In `search-sync-worker/handler.go` around line 188: When Msg.Metadata() fails or is nil, streamSequence returns 0. The spotlight-org script compares incoming 0 against stored 0 or greater, performs a no-op, and Elasticsearch reports success; FlushBatch then ACKs the source message although no update was applied. Treat a missing or zero sequence as a retryable processing failure and settle it separately from BuildAction validation errors, which are intentionally ACK-dropped.

```

</details>

search-sync-worker/handler.go
```diff
@@ -194,35 +267,57 @@ func (h *Handler) Flush(ctx context.Context) {
...
+	var acked, poisoned, nakked int
 	for _, p := range pending {
-		allOK := true
+		var itemErr error
+		backoff := jsretry.DefaultBackoff
+		failed, permanent := false, false
 		for i := p.actionStart; i < p.actionStart+p.actionCount; i++ {
 			if searchengine.IsBulkItemSuccess(actions[i].Action, results[i]) {
 				continue
 			}
```
julianshen, line 278: _🏷️ partial-failure_ | _🔴 Blocking_ | _`distributed-system`, `nats` (2 rules)_

**The first failed bulk item can hide retryable failures.**

One source message can fan out to multiple bulk actions. After the first failure, this continue prevents classification of later failures. If the first item is permanent while a later item returns 429 or 500, the source message is ACK-dropped and the later update is lost. Inspect every result for the message, retry if any failure is retryable, classify it as permanent only when all failures are permanent, and select backpressure pacing if any retryable failure indicates overload.

<details>

- **`nats`**:279: One HR message can fan out to several bulk actions. After the first failed item, this branch skips classification of every later failure. If the first failure is permanent but a later item returns 429 or 500, the entire source message is ACK-dropped and that later update is lost. Inspect all failures, retry if any is retryable, and select backpressure pacing if any retryable failure indicates overload.

</details>

<details>

```
In `search-sync-worker/handler.go` around line 278: One source message can fan out to multiple bulk actions. After the first failure, this continue prevents classification of later failures. If the first item is permanent while a later item returns 429 or 500, the source message is ACK-dropped and the later update is lost. Inspect every result for the message, retry if any failure is retryable, classify it as permanent only when all failures are permanent, and select backpressure pacing if any retryable failure indicates overload.

```

</details>

search-sync-worker/spotlight_org.go
```diff
@@ -12,6 +12,37 @@ import (
...
+// plain doc_as_upsert with no ordering semantics: a redelivered batch, or a batch from a
+// second pod sharing the durable consumer, can land after a newer one and leave the older
+// row stored.
+//
+// Merging params.doc key-by-key reproduces doc-merge semantics — the row is marshalled
+// with omitempty, so fields absent from a partial publish keep their stored values.
+//
+// Runs under scripted_upsert so a brand-new sectId also records its sequence;
+// ctx._source starts as the empty upsert doc and stored reads 0.
+//
```
julianshen, line 35: _🏷️ jetstream-ordering_ | _🔴 Blocking_ | _`nats`_

**Stream recreation makes new updates look permanently stale**

The persisted ordering token contains only the stream sequence. Recreating the HR stream restarts sequences at 1, so documents retaining a larger orgSeq reject every new update as a successful Elasticsearch no-op, after which the worker ACKs each message. Include a stream generation in the ordering token, or enforce and verify an orgSeq reset/reindex before consumption resumes.

<details>

```
In `search-sync-worker/spotlight_org.go` around line 35: The persisted ordering token contains only the stream sequence. Recreating the HR stream restarts sequences at 1, so documents retaining a larger orgSeq reject every new update as a successful Elasticsearch no-op, after which the worker ACKs each message. Include a stream generation in the ordering token, or enforce and verify an orgSeq reset/reindex before consumption resumes.

```

</details>

pkg/jsretry/jsretry_test.go
```diff
@@ -181,6 +181,29 @@ func TestDefaultBackoff_Shape(t *testing.T) {
...
+		"the first backpressure retry must skip the fast rungs that feed the overload")
+
+	sum := func(s []time.Duration) time.Duration {
+		var total time.Duration
+		for _, d := range s {
+			total += d
+		}
+		return total
+	}
+	assert.GreaterOrEqual(t, sum(BackpressureBackoff), sum(DefaultBackoff),
```
julianshen, line 201: _🧪 tests_ | _🔴 Blocking_ | _`nats`_

**The backpressure retry-budget assertion cannot pass**

BackpressureBackoff skips DefaultBackoff's positive fast rungs while retaining its later schedule, so its slice sum is necessarily smaller, not greater or equal. The new test therefore fails even though delayed retries clamp to the final rung. Test pacing over the configured MaxDeliver attempts, including clamping, or remove the incorrect cumulative-slice assertion.

<details>

```
In `pkg/jsretry/jsretry_test.go` around line 201: BackpressureBackoff skips DefaultBackoff's positive fast rungs while retaining its later schedule, so its slice sum is necessarily smaller, not greater or equal. The new test therefore fails even though delayed retries clamp to the final rung. Test pacing over the configured MaxDeliver attempts, including clamping, or remove the incorrect cumulative-slice assertion.

```

</details>

search-sync-worker/main.go
```diff
@@ -115,6 +123,12 @@ func main() {
 		slog.Error("invalid config", "name", "BULK_FLUSH_INTERVAL", "value", cfg.BulkFlushInterval, "reason", "must be > 0")
 		os.Exit(1)
 	}
+	// A non-positive depth would make the pipeline's slot channel unbuffered and park the
```
julianshen, line 126: _⚡ memory-exhaustion_ | _🟠 Warning_ | _`oom`_

**PIPELINE_DEPTH can allocate and retain an arbitrary number of batches**

The validation still accepts any positive PIPELINE_DEPTH. Each collection immediately allocates `make(chan struct{}, depth)`, so an excessively large but parseable deployment value can OOM during startup; under backlog with slow or hung Elasticsearch, it also permits that many goroutines to retain flush batches, source messages, action documents, and request memory, multiplied across collections and replicas. The advisory coupling check does not enforce a bound and `(pipelineDepth + 1) * bulkBatchSize` can overflow. Enforce a conservative maximum at startup, constrain depth against an enforced memory/ack-pending budget, and calculate the coupling with checked arithmetic.

<details>

```
In `search-sync-worker/main.go` around line 126: The validation still accepts any positive PIPELINE_DEPTH. Each collection immediately allocates `make(chan struct{}, depth)`, so an excessively large but parseable deployment value can OOM during startup; under backlog with slow or hung Elasticsearch, it also permits that many goroutines to retain flush batches, source messages, action documents, and request memory, multiplied across collections and replicas. The advisory coupling check does not enforce a bound and `(pipelineDepth + 1) * bulkBatchSize` can overflow. Enforce a conservative maximum at startup, constrain depth against an enforced memory/ack-pending budget, and calculate the coupling with checked arithmetic.

```

</details>

julianshen (commented):
tGD inline review

search-sync-worker/handler.go
```diff
@@ -146,18 +148,89 @@ func (h *Handler) AddWithContext(ctx context.Context, msg jetstream.Msg) {
...
+//
+// Only the status and error type go into the message; the document body never belongs
+// in an error that reaches the server log.
+func bulkItemError(result searchengine.BulkResult) error {
+	if searchengine.IsBulkItemPermanent(result) {
+		return errcode.Permanent(errcode.BadRequest(
+			fmt.Sprintf("search backend rejected the document: status %d %q", result.Status, result.ErrorType)))
+	}
+	return fmt.Errorf("bulk item failed: status %d %q", result.Status, result.ErrorType)
+}
```
julianshen, line 164: _🏷️ backpressure_ | _🟠 Warning_ | _`tgd-review`_

**Request-level overload still uses the fast retry schedule.**

itemBackoff only classifies per-item BulkResult values. A top-level bulk rejection such as HTTP 429 or circuit_breaking_exception follows the whole-request error path and uses DefaultBackoff, retrying after roughly one second despite the new overload policy. Preserve request status/error type in a typed Bulk error and apply BackpressureBackoff to every message in an overloaded request.

<details>

```
In `search-sync-worker/handler.go` around line 164: itemBackoff only classifies per-item BulkResult values. A top-level bulk rejection such as HTTP 429 or circuit_breaking_exception follows the whole-request error path and uses DefaultBackoff, retrying after roughly one second despite the new overload policy. Preserve request status/error type in a typed Bulk error and apply BackpressureBackoff to every message in an overloaded request.

```

</details>

mliu33 (approved):
Outstanding! Thanks

PR #312, opened 2026-08-18
fix(loadgen): restore created rooms after restart

flywindy:
## Summary

- restore create-lane and member-lane subscriptions when a soak process reloads its topology
- align loadgen membership checks and seeded subscription documents with the production model
- delete all subscriptions scoped to verified run-owned rooms during teardown
- add regression coverage for seed -> create -> restart, DM member removal, and teardown cleanup

## Root cause

Two independent conditions made a run fail after its first create-lane room:

1. `AppendOwnedRooms` marks the room with `soakRunId`, but room-service writes the room's subscriptions without that loadgen-private marker. Reload queried rooms by ownership and subscriptions by ownership, so it loaded the room without any members. Teardown used the same asymmetric filter and left orphan subscriptions.
2. Loadgen treated `IsSubscribed` as channel/DM membership. Production channel and DM membership is represented by subscription-row existence because leaving deletes the row; only botDM retains the row and uses `IsSubscribed` as a soft toggle.

At the default `SOAK_ROOM_CREATE_RATE=0.05`, one created room was enough to make every later pod restart abort with `soak room %q has no subscribed member`.

## Changes

- load subscriptions only after resolving the ownership chunk to rooms that still match the run and site, then scope the subscription query by those verified room IDs
- delete subscriptions by verified owned room IDs so server-written rows are cleaned up
- centralize channel/DM versus botDM membership semantics in `isSoakRoomMember`
- apply the shared membership contract to sender selection, read/mutation targets, room state, recipient expectations, and recipient observation
- leave `IsSubscribed` unset on seeded channel/DM rows so seed data matches production-created membership documents
- cover the DM removal model explicitly: a DM with only one remaining subscription row produces no requester/peer pair
- use named table-driven cases for the shared membership predicate

## Behavioral impact

Reloaded create-lane channel rooms now enter `newSoakRoomStatePool`, so member and room mutation lanes may operate on them after a process restart. This is intentional and safe:

- `UserCount` is not consumed by loadgen, so its absence on a created room does not create a state mismatch
- renames append `-r{n}` and preserve the `soak-{run}-created-` prefix used by `CountCreatedRooms`
- mute targets remain seeded members while membership candidates come from non-members, preserving the disjoint-set invariant

The pool is built once at process startup and is not updated dynamically. A room created by the current process therefore joins the pool only after the next restart; first-process and post-restart mutation coverage are intentionally different for now.

### Red

- `TestLoadTopology_RestoresRoomsCreatedDuringTheRun` reproduced the restart failure with a room-service-shaped subscription
- teardown coverage reproduced the orphaned subscription
- unit fixtures without `isSubscribed` failed across selector, room state, DM pairing, recipient expectation, and observer setup
- `TestBuildSoakSubscriptions_LeavesBotDMOnlyFlagUnset` failed while seed still wrote `IsSubscribed: true` on channel/DM rows

### Green / Refactor

- topology reload and teardown now use verified owned-room scope
- all membership consumers share the production-aligned helper, including botDM soft-toggle behavior
- seed and production-created channel/DM rows now use the same membership representation
- DM single-row coverage pins the existing incomplete-pair guard

## Validation

- `make test SERVICE=tools/loadgen`
- `make coverage-loadgen-soak`
  - soak aggregate: 83.23%
  - soak core excluding `soak_main.go` and `soak_store.go`: 90.74%
- `make test-integration SERVICE=tools/loadgen`
- `make lint` — 0 issues
- `make coverage-loadgen-failure`
  - failure core: 81.13%
  - recipient observer: 97.30%
  - failure metrics: 100%
- gosec — pass
- Semgrep — pass
- `make sast-vuln GOVULNCHECK=/c/Users/SheepRocket/go/bin/govulncheck.exe` — 0 called vulnerabilities

One intermediate integration run observed the existing end-to-end drain timing failure `message-worker still has pending: expected 0, got 1`; the exact prior head subsequently passed the complete integration/race suite. The review follow-up commit passed the full loadgen unit race suite and the soak integration/coverage target.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved load-test room membership detection by recognizing existing room membership records, including those created by room services.
  - Preserved filtering for inactive users, bots, empty accounts, and invalid memberships.
  - Ensured active accounts are selected for room reads and direct-message requests.
  - Improved topology restoration and teardown for rooms created during a test run.
- **Tests**
  - Added coverage for membership filtering, account selection, topology restoration, and cleanup behavior.

### DIFF
--- a/tools/loadgen/soak_topology.go
+++ b/tools/loadgen/soak_topology.go
@@ -343,14 +343,13 @@ func buildSoakSubscriptions(
 				ID:      members[i].ID,
 				Account: members[i].Account,
 			},
-			RoomID:       room.ID,
-			SiteID:       room.SiteID,
-			Roles:        roles,
-			Name:         name,
-			RoomType:     room.Type,
-			IsSubscribed: true,
-			Open:         true,
-			JoinedAt:     joinedAt,
+			RoomID:   room.ID,
+			SiteID:   room.SiteID,
+			Roles:    roles,
+			Name:     name,
+			RoomType: room.Type,
+			Open:     true,
+			JoinedAt: joinedAt,
 		}
 	}
 	return subscriptions
--- a/tools/loadgen/soak_topology.go
+++ b/tools/loadgen/soak_topology.go
@@ -286,7 +286,7 @@ func isActiveSoakSubscription(
 	subscription *model.Subscription,
 	active map[string]struct{},
 ) bool {
-	if subscription == nil || !subscription.IsSubscribed {
+	if !isSoakRoomMember(subscription) {
 		return false
 	}
 	if len(active) == 0 {
@@ -294,6 +294,16 @@ func isActiveSoakSubscription(
 	}
 	_, ok := active[subscription.User.ID]
 	return ok
+}
+
+// isSoakRoomMember follows the production subscription model: channel and DM
+// membership is represented by row existence because leave deletes the row.
+// Only botDM keeps the row and uses IsSubscribed as a soft toggle.
+func isSoakRoomMember(subscription *model.Subscription) bool {
+	if subscription == nil {
+		return false
+	}
+	return subscription.RoomType != model.RoomTypeBotDM || subscription.IsSubscribed
 }
 
 func reserveSoakPair(a, b *model.User, used map[string]struct{}) bool {
--- a/tools/loadgen/failure_platform_test.go
+++ b/tools/loadgen/failure_platform_test.go
@@ -495,8 +495,8 @@ func TestFailureRecipientObserver_SubscriptionsAreAttributedAndFlushedBeforeHeal
 	topology := &soakTopology{
 		Rooms: []model.Room{{ID: "channel-1", Type: model.RoomTypeChannel}},
 		Subscriptions: []model.Subscription{
-			{RoomID: "channel-1", RoomType: model.RoomTypeChannel, IsSubscribed: true, User: model.SubscriptionUser{Account: "alice"}},
-			{RoomID: "channel-1", RoomType: model.RoomTypeChannel, IsSubscribed: true, User: model.SubscriptionUser{Account: "bob"}},
+			{RoomID: "channel-1", RoomType: model.RoomTypeChannel, User: model.SubscriptionUser{Account: "alice"}},
+			{RoomID: "channel-1", RoomType: model.RoomTypeChannel, User: model.SubscriptionUser{Account: "bob"}},
 		},
 	}
 	subscriptions, err := startFailureRecipientSubscriptions(source, topology, observer)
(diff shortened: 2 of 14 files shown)


julianshen:
## Needs clarification

Can any existing channel or DM subscription retain isSubscribed:false after leave?
Reply in this thread, or post: `answer clar_zmx4dkp57p2efwpx5i4zi7j7tq: <your answer>`

julianshen:

### Needs clarification

Can any existing channel or DM subscription retain isSubscribed:false after leave?

`answer clar_zmx4dkp57p2efwpx5i4zi7j7tq: <your answer>`

[Open the question](https://github.com/hmchangw/newchat/pull/312#issuecomment-5333552912)

<details>

* `tools/loadgen/failure_platform_test.go`
* `tools/loadgen/soak_read_test.go`
* `tools/loadgen/soak_roomstate.go`
* `tools/loadgen/soak_roomstate_test.go`
* `tools/loadgen/soak_teardown_integration_test.go`
* `tools/loadgen/soak_topology.go`
* `tools/loadgen/soak_topology_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

julianshen (commented):
tGD inline review

tools/loadgen/soak_topology.go
```diff
@@ -333,14 +343,13 @@ func buildSoakSubscriptions(
...
 			},
-			RoomID:       room.ID,
-			SiteID:       room.SiteID,
-			Roles:        roles,
-			Name:         name,
-			RoomType:     room.Type,
-			IsSubscribed: true,
-			Open:         true,
-			JoinedAt:     joinedAt,
+			RoomID:   room.ID,
```
julianshen, line 346: _🏷️ compatibility_ | _🟠 Warning_ | _`distributed-system`_

**Omitting IsSubscribed breaks restart with the previous loadgen version.**

A new version now seeds channel and DM rows with IsSubscribed=false/omitted, while the previous version's selector, recipient observer, room-state pool, and user reader all reject such rows, as shown by the removed `!subscription.IsSubscribed` checks in this diff. If a run is resumed by an older binary after rollback or during a mixed-version deployment, that binary observes zero members and may fail initialization or silently omit recipient/read coverage. Keep `IsSubscribed: true` on loadgen-authored seed rows while retaining the new row-existence logic for production-authored rows that omit the field; update the new assertion accordingly.

<details>

```
In `tools/loadgen/soak_topology.go` around line 346: A new version now seeds channel and DM rows with IsSubscribed=false/omitted, while the previous version's selector, recipient observer, room-state pool, and user reader all reject such rows, as shown by the removed `!subscription.IsSubscribed` checks in this diff. If a run is resumed by an older binary after rollback or during a mixed-version deployment, that binary observes zero members and may fail initialization or silently omit recipient/read coverage. Keep `IsSubscribed: true` on loadgen-authored seed rows while retaining the new row-existence logic for production-authored rows that omit the field; update the new assertion accordingly.

```

</details>

mliu33 (approved):

PR #313, opened 2026-08-18
search-sync-worker: point-read the thread parent instead of searching every monthly index

hmchangw:
The thread-parent `createdAt` fallback searched `messages-<site>-v1-*` — every monthly index — for one timestamp on one document, synchronously on the consumer goroutine. Fixing it turned out to be a correctness fix, not just a cost one.

Independent of #311: that PR touches `main.go`, `handler.go`, `spotlight_org.go` and three `pkg/` files; this one touches `messages.go` and `thread_parent_resolver.go`. Disjoint — they merge in either order.

## The correctness bug

The message template sets **`refresh_interval: 30s`** (`pkg/searchindex/template.go:74`), and an ES search only sees refreshed segments. So **a parent indexed less than thirty seconds ago is not searchable.** The lookup paid its full cost to return not-found for exactly the case that dominates live chat — a quick reply to a recent message — and the field then stays unset forever, because nothing backfills it.

That field is not decorative. `search-service/query_messages.go:209-226`, restricted-room clause B, matches a thread reply only when `tshow=true` **or** `threadParentMessageCreatedAt >= historySharedSince`. An ES `range` query cannot match a document missing the field, so:

> An unset value hides a thread reply the user is entitled to see — silently, and permanently until a reindex.

Fail-closed, so not a leak. But a false negative, and one that gets *more* likely during an incident: the resolver only runs when the gatekeeper left the field nil, and `message-gatekeeper/handler.go:507-514` nils it on **any** parent-fetch error. A Cassandra or history-service wobble therefore stops the field being filled for every thread reply, and search-sync-worker answers by issuing an ES search per reply.

## The fix

A **get is realtime** — served from the translog, not bound by `refresh_interval` — and routes to a single shard by id. The obstacle is that it needs a concrete index, and the monthly index name comes from the parent's `createdAt`, which is the thing being looked up.

The reply's own `createdAt` breaks the cycle: a parent always predates its reply, and threads are overwhelmingly same-month. So the resolver probes the reply's month, then the month before, and only then falls back to the existing cross-month search for genuinely old parents.

| Case | Before | After |
|---|---|---|
| Parent same month | ~4 shards × N monthly indices, **misses if <30s old** | 1 shard read, always finds it |
| Parent previous month | same fan-out | 2 shard reads |
| Parent older | same fan-out | 2 reads + the same search |
| Parent nowhere | same fan-out | 2 reads + search — marginally dearer |

The worst case costs slightly more. Good trade: realtime gets make "not indexed yet" much rarer, since today that also covers "indexed but not yet refreshed".

## One trap worth reviewing closely

`prevMonth` deliberately avoids `AddDate(0, -1, 0)`. That normalises **31 March** into "31 February" and lands back on **3 March** — repeating the month just probed and silently losing the boundary case. Going through `time.Date` on the month number normalises correctly, January → previous December included. There's a table test for it.

## Testing

Written test-first; the resolver tests assert the *probe sequence*, so an implementation that searches too eagerly or skips the previous month fails.

- `prevMonth`: mid-month, the 31st, the 1st, January→December, leap February
- Resolver: same-month hit costs one read and **no** search; previous-month hit; fallback to search for older; not-found
- A new test pins that the reply's `createdAt` actually reaches the resolver — without it every probe would target the zero-time index and nothing else would catch that

`prevMonth`, `ResolveParentCreatedAt` and `resolveThreadParentCreatedAt` are at 100%. Package coverage moves 55.7% → 54.7%: `newESGet` adds wire-adapter statements reachable only from the integration test, exactly like the `newESRead` it replaces, which was already 0% on `main`.

**The integration test drops its `refreshIndex` call.** That call was the test compensating for the production bug — without it the test fails on the old search path and passes on the new one, which is the property worth pinning. Docker was unavailable here, so it is compile-checked (`go vet -tags integration`) but **not executed**; CI is its first real run, and it is the only thing that verifies the realtime behaviour.

`make lint` 0 issues, `gosec` clean, full package green under `-race`.

## What this does not fix

If parent and reply land in the **same flush batch**, no lookup can find the parent — it hasn't been written yet. The real fix there is filling the field upstream: `message-gatekeeper` already does so best-effort, and `bot-message-handler` passes through whatever the client sent. This change makes the fallback fast and correct where a fallback can work; it doesn't remove the need for one.

https://claude.ai/code/session_01NnEWtiYrtp72HsX78yjDLy

---
_Generated by [Claude Code](https://claude.ai/code/session_01NnEWtiYrtp72HsX78yjDLy)_

### DIFF
--- a/search-sync-worker/thread_parent_resolver.go
+++ b/search-sync-worker/thread_parent_resolver.go
@@ -5,44 +5,120 @@ import (
 	"encoding/json"
 	"log/slog"
 	"time"
+
+	"github.com/hmchangw/chat/pkg/searchindex"
 )
 
-// parentResolveTimeout bounds a single thread-parent createdAt lookup so a slow
-// search backend can't stall the indexing path.
+// parentResolveTimeout bounds one thread-parent createdAt lookup — including any
+// fallback search — so a slow search backend can't stall the indexing path.
 const parentResolveTimeout = 2 * time.Second
 
-// esReadFn reads a message's (server-stamped, authoritative) createdAt from the
-// search index. ok=false means not found — the caller leaves the field unset.
-type esReadFn func(ctx context.Context, messageID string) (time.Time, bool)
+// esGetFn reads a message's createdAt from one concrete index by document id.
+// ok=false means the index or the document is absent.
+type esGetFn func(ctx context.Context, index, messageID string) (time.Time, bool)
 
-// esParentResolver resolves a thread parent's createdAt via an ES self-lookup. It
+// esSearchFn reads a message's createdAt by searching every monthly index.
+// ok=false means not found — the caller leaves the field unset.
+type esSearchFn func(ctx context.Context, messageID string) (time.Time, bool)
+
+// esParentResolver resolves a thread parent's createdAt from the search index. It
 // never errors: a miss or backend failure degrades the field rather than stalling.
 type esParentResolver struct {
-	esRead  esReadFn
-	timeout time.Duration
-	metrics *collectionMetrics // optional, set by main; nil-safe recording
+	esGet       esGetFn
+	esSearch    esSearchFn
+	indexPrefix string
+	timeout     time.Duration
+	metrics     *collectionMetrics // optional, set by main; nil-safe recording
+}
+
+// prevMonth returns any instant inside the calendar month before at.
+//
+// Not at.AddDate(0, -1, 0): that normalises overflow, so 31 March becomes
+// "31 February" and lands back on 3 March — repeating the month the caller just
+// probed. Going via time.Date on the month number normalises correctly, including
+// January rolling back into the previous December.
+func prevMonth(at time.Time) time.Time {
+	at = at.UTC()
+	return time.Date(at.Year(), at.Month()-1, 1, 0, 0, 0, 0, time.UTC)
 }
 
 // ResolveParentCreatedAt returns the parent's createdAt and ok=true when resolved.
 // ok=false means the parent isn't indexed yet — the caller leaves the field unset.
-func (r *esParentResolver) ResolveParentCreatedAt(ctx context.Context, messageID string) (time.Time, bool) {
+//
+// A parent always predates its reply and threads are overwhelmingly same-month, so
+// this point-reads the reply's month and then the one before it before falling back
+// to the cross-month search. Two reasons that ordering matters:
+//
+//   - A get is realtime — served from the translog, not bound by the index's
+//     refresh_interval — so it sees a parent indexed seconds ago. The search cannot,
+//     which made it miss exactly the recent-parent case that dominates live chat.
+//   - A get routes to a single shard by id; the search fans out across every monthly
+//     index times its shard count.
+//
+// The search stays as the fallback for parents older than one month, where the month
+// probes can't guess the index.
+func (r *esParentResolver) ResolveParentCreatedAt(ctx context.Context, messageID string, replyCreatedAt time.Time) (time.Time, bool) {
 	ctx, cancel := context.WithTimeout(ctx, r.timeout)
 	defer cancel()
+
+	// Times the whole resolution — both month probes and any fallback search — because
+	// that is the latency the indexing path actually pays for one parent.
 	start := time.Now()
-	createdAt, ok := r.esRead(ctx, messageID)
+	createdAt, ok := r.resolve(ctx, messageID, replyCreatedAt)
 	r.metrics.recordParentResolve(ctx, time.Since(start).Seconds(), ok)
 	return createdAt, ok
 }
 
-// esSearcher is the narrow search surface the ES self-lookup needs.
+// resolve point-reads the reply's month and then the one before it, falling back to the
+// cross-month search only for a parent older than either.
+func (r *esParentResolver) resolve(ctx context.Context, messageID string, replyCreatedAt time.Time) (time.Time, bool) {
+	for _, at := range [...]time.Time{replyCreatedAt, prevMonth(replyCreatedAt)} {
+		if createdAt, ok := r.esGet(ctx, searchindex.MessageIndexName(r.indexPrefix, at), messageID); ok {
+			return createdAt, true
+		}
+	}
+	return r.esSearch(ctx, messageID)
+}
+
+// esReader is the narrow surface the parent lookup needs.
 // searchengine.SearchEngine satisfies it.
-type esSearcher interface {
+type esReader interface {
+	GetDoc(ctx context.Context, index, docID string) (json.RawMessage, bool, error)
 	Search(ctx context.Context, indices []string, body json.RawMessage) (json.RawMessage, error)
 }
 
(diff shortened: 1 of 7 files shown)


julianshen:

> [!NOTE]

### Related work

- [PR #311](https://github.com/hmchangw/newchat/pull/311) — search-sync-worker: pipeline ES flushes, pace failed-write retries, close ordering gaps (merged)

<details>

* `.gitignore`
* `search-sync-worker/messages_reresolve_test.go`
* `search-sync-worker/metrics_test.go`
* `search-sync-worker/parent_resolver_integration_test.go`
* `search-sync-worker/thread_parent_resolver.go`
* `search-sync-worker/thread_parent_resolver_test.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

search-sync-worker/thread_parent_resolver.go
```diff
@@ -5,40 +5,108 @@ import (
...
 	ctx, cancel := context.WithTimeout(ctx, r.timeout)
 	defer cancel()
-	return r.esRead(ctx, messageID)
+
+	for _, at := range [...]time.Time{replyCreatedAt, prevMonth(replyCreatedAt)} {
+		if createdAt, ok := r.esGet(ctx, searchindex.MessageIndexName(r.indexPrefix, at), messageID); ok {
+			return createdAt, true
+		}
+	}
+	return r.esSearch(ctx, messageID)
```
julianshen, line 80: _🏷️ consistency_ | _🟠 Warning_ | _`distributed-system`_

**Older parents can be permanently missed before an index refresh**

A concrete read-after-write gap remains for parents outside the two probed months: a worker can successfully bulk-index an older parent, then process its reply before Elasticsearch refreshes that older monthly index. Both point reads target the reply month and previous month, while the fallback search is near-real-time and therefore returns no hit. BuildAction then persists the reply without ThreadParentMessageCreatedAt; a later refresh exposes the parent but nothing shown retries or repairs the already-indexed reply. The unit test for older parents uses an immediately visible search stub, and the integration test only covers months eligible for point reads. Retry/requeue fallback misses after refresh, add a reconciliation path, or provide a way to point-read the actual parent index.

<details>

```
In `search-sync-worker/thread_parent_resolver.go` around line 68: A concrete read-after-write gap remains for parents outside the two probed months: a worker can successfully bulk-index an older parent, then process its reply before Elasticsearch refreshes that older monthly index. Both point reads target the reply month and previous month, while the fallback search is near-real-time and therefore returns no hit. BuildAction then persists the reply without ThreadParentMessageCreatedAt; a later refresh exposes the parent but nothing shown retries or repairs the already-indexed reply. The unit test for older parents uses an immediately visible search stub, and the integration test only covers months eligible for point reads. Retry/requeue fallback misses after refresh, add a reconciliation path, or provide a way to point-read the actual parent index.

```

</details>

search-sync-worker/thread_parent_resolver.go
```diff
@@ -5,40 +5,108 @@ import (
...
-func newESRead(searcher esSearcher, indexPrefix string) esReadFn {
+// newESGet returns an esGetFn that point-reads one index. A missing index is a
+// miss, not an error: GetDoc reports both as 404, which is what lets the caller
+// probe a month that may never have been created.
+func newESGet(reader esReader) esGetFn {
+	return func(ctx context.Context, index, messageID string) (time.Time, bool) {
+		raw, found, err := reader.GetDoc(ctx, index, messageID)
+		if err != nil {
+			slog.WarnContext(ctx, "ES parent createdAt get failed", "error", err, "index", index, "messageId", messageID)
+			return time.Time{}, false
```
julianshen, line 98: _🏷️ operational-risk_ | _🟠 Warning_ | _`elasticsearch`_

**GET failures trigger additional Elasticsearch requests**

newESGet collapses hard failures such as 429s, timeouts, and connection errors into the same false result as a 404. ResolveParentCreatedAt therefore proceeds with another GET and then a cross-month search, so each reply can turn one backend failure into three requests, with the final request fanning out across every monthly index. This can amplify an Elasticsearch overload. Preserve the distinction between not-found and failure (for example, return an error or tri-state from esGetFn); continue probing only for not-found, and stop resolution after a hard failure.

<details>

```
In `search-sync-worker/thread_parent_resolver.go` around line 86: newESGet collapses hard failures such as 429s, timeouts, and connection errors into the same false result as a 404. ResolveParentCreatedAt therefore proceeds with another GET and then a cross-month search, so each reply can turn one backend failure into three requests, with the final request fanning out across every monthly index. This can amplify an Elasticsearch overload. Preserve the distinction between not-found and failure (for example, return an error or tri-state from esGetFn); continue probing only for not-found, and stop resolution after a hard failure.

```

</details>

julianshen (commented):
tGD inline review

search-sync-worker/thread_parent_resolver.go
```diff
@@ -5,44 +5,120 @@ import (
...
+// resolve point-reads the reply's month and then the one before it, falling back to the
+// cross-month search only for a parent older than either.
+func (r *esParentResolver) resolve(ctx context.Context, messageID string, replyCreatedAt time.Time) (time.Time, bool) {
+	for _, at := range [...]time.Time{replyCreatedAt, prevMonth(replyCreatedAt)} {
+		if createdAt, ok := r.esGet(ctx, searchindex.MessageIndexName(r.indexPrefix, at), messageID); ok {
+			return createdAt, true
+		}
+	}
+	return r.esSearch(ctx, messageID)
+}
```
julianshen, line 81: _🎯 correctness_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`silent-failure`_

**Unrefreshed older parents are still permanently missed**

For a freshly indexed parent older than the two probed months, both GETs target unrelated indices and the fallback search cannot see the parent until refresh. The resolver therefore returns false and BuildAction persists the reply without the parent timestamp; no retry or repair is shown, so the described restricted-room query can continue hiding that reply after the parent becomes searchable. Make this fallback miss retriable after refresh, reconcile unresolved replies, or determine the parent's concrete index for a realtime GET.

<details>

```
In `search-sync-worker/thread_parent_resolver.go` around line 81: For a freshly indexed parent older than the two probed months, both GETs target unrelated indices and the fallback search cannot see the parent until refresh. The resolver therefore returns false and BuildAction persists the reply without the parent timestamp; no retry or repair is shown, so the described restricted-room query can continue hiding that reply after the parent becomes searchable. Make this fallback miss retriable after refresh, reconcile unresolved replies, or determine the parent's concrete index for a realtime GET.

```

</details>

---

julianshen (commented):
### tGDBot review of `f505128` — 1 finding in 7 files

1 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/313#issuecomment-5352956915), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

PR #314, opened 2026-08-19
fix(room-worker): self-correct migrated subscription joinedAt on replay

chenjr0719:
## Summary
Migrated subscriptions self-correct their `joinedAt` on a migration replay. Previously `BulkCreateSubscriptions` wrote the whole sub under `$setOnInsert`, so a sub first-inserted before the `createdDateTime` fix kept its stale migration-time `joinedAt` even when the batch was re-run. Now `joinedAt` is written via `$set`, and the Teams reconcile refreshes an already-present member whose `joinedAt` is stale — without re-adding, re-keying, or re-fanning-out.

## Changes
- `room-worker/store_mongo.go` — `BulkCreateSubscriptions` moves `joinedAt` from `$setOnInsert` into `$set` (via a small `bsonMExcept` marshal-and-remove, since a field can't be in both). On an existing `(roomId, u.account)` doc the `$setOnInsert` is a no-op, so `_id`, read state (`lastSeenAt`), roles and section are preserved and only `joinedAt` converges. `joinedAt` is deterministic from the event/chat and is not read-state, so overwriting it on every write is safe.
- `room-worker/teamsroomcreate.go` — `reconcileTeamsRoom` refreshes an already-present member whose `joinedAt != chat.createdDateTime` through the same upsert; unchanged members and current `joinedAt` write nothing.

## Verification
- Unit: `RefreshesStaleJoinedAt`, `NoRefreshWhenJoinedAtCurrent` (reconcile); `BulkCreateSubscriptions_UpsertRefreshesJoinedAt` (asserts `_id` + `lastSeenAt` + roles preserved, `joinedAt` refreshed), `BulkCreateSubscriptions_InsertSetsJoinedAt` (integration).
- Verified on our dev stack: publish a migrated room, force a stale `joinedAt` on a sub, replay the batch → `joinedAt` corrected to the chat's `createdDateTime` and the sub's read state preserved.

🤖 Generated with Claude Code

## Summary by CodeRabbit

* **New Features**
  * Corrected member join timestamps during Teams room reconciliation.
  * Propagated timestamp corrections across sites for existing room subscriptions.
  * Added support for processing joined-time refresh events.
  * Preserved subscription details and ignored accounts without existing subscriptions.

* **Bug Fixes**
  * Prevented stale `joinedAt` values from persisting after room reconciliation.

### DIFF
--- a/inbox-worker/handler.go
+++ b/inbox-worker/handler.go
@@ -20,6 +20,9 @@ import (
 type InboxStore interface {
 	CreateSubscription(ctx context.Context, sub *model.Subscription) error
 	BulkCreateSubscriptions(ctx context.Context, subs []*model.Subscription) error
+	// BulkRefreshJoinedAt sets joinedAt on existing (roomId, account) replicas —
+	// the Teams migration's cross-site joinedAt correction; a missing sub is a no-op.
+	BulkRefreshJoinedAt(ctx context.Context, roomID string, joinedAtByAccount map[string]time.Time) error
 	// UpsertRoom replicates room metadata, guarded by the incoming room's
 	// UpdatedAt: an event carrying an older (or equal) UpdatedAt than the
 	// stored one is a silent no-op, so out-of-order federated delivery cannot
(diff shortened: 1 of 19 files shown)


mliu33 (approved):
Thanks!

PR #316, opened 2026-08-19
fix(search-sync): re-index spotlight room name on rename

### LINKED ISSUE #315: [Fix] room rename does not re-index the spotlight room name (stale search-by-name)
> **Verified against:** `main`@dab21d38 — code read after a fresh `git pull`.

## Summary
Renaming a room updates the Mongo subscription names but does **not** re-index the room name in the **spotlight** search index. After a rename, searching by the *new* name fails to find the room, and the *old* name still matches. Affects any renamed room (not migration-specific).

## Reproduce steps
- **Preconditions:** a room with ≥1 member; search-sync-worker running (default mode); spotlight index populated.
- **Steps:**
  1. Create/have a room named "Alpha" with member alice; confirm `search.rooms` (or the spotlight index) returns it for "Alpha".
  2. Rename the room to "Beta" (the room-rename RPC → `room_renamed` inbox event).
  3. Search rooms for "Beta", then for "Alpha".
- **Expected vs actual:** expected — "Beta" finds it, "Alpha" does not. Actual — "Alpha" still finds it, "Beta" does not (spotlight `roomName` is stale).

## Root cause  (file:line)
- The **spotlight** ES doc indexes `roomName` (`pkg/searchindex/spotlightdoc.go`, `search_as_you_type`), sourced **only** from member events: `search-sync-worker/spotlight.go` `BuildAction` handles `member_added`/`member_removed` (+ `member_joinedat_refreshed`), and its consumer filter is `subject.InboxMemberEventSubjects` — **member events only**.
- A rename is `room_renamed` → `inbox-worker/handler.go handleRoomRenamed` → `UpdateSubscriptionNamesForRoom` (updates **Mongo** sub names). `room_renamed` is **not consumed by search-sync-worker**, so the spotlight `roomName` is never re-indexed.

## Message-path trace
1. Client request struct + docs/client-api.md — N/A (rename RPC unchanged)
2. Gatekeeper mapping — N/A
3. Canonical model / event structs (pkg/model) — no change (reuse `RoomRenamedInboxPayload` / room_renamed)
4. Workers — **change: search-sync-worker must consume `room_renamed`** and re-index spotlight roomName; room-worker already publishes room_renamed (internal + outbox)
5. Storage tables — Mongo already updated by inbox-worker; **ES spotlight index roomName** is the gap
6. Read / history path — N/A
7. Client-facing events — N/A (search results only)

## Fix  (Primary recommended + alternative)
- **Primary:** add `room_renamed` to the spotlight collection's consumed subjects (internal + external) and handle it in `spotlight.BuildAction` — a bulk/partial update of `roomName` for all spotlight docs of that roomId, guarded by the rename timestamp (external-version) so a stale rename can't overwrite a newer one. user-room ignores it (roomName not indexed there).
- **Alternative:** a painless-script partial update keyed on roomId prefix; heavier, only if the full re-index is undesirable.

## Contracts & docs
- docs/client-api.md — N/A (no wire change)
- Schema change? No Cassandra; no Mongo index; ES spotlight mapping already has roomName (no new field).

## Tests
- Unit (table-driven): spotlight `BuildAction` on `room_renamed` → produces the roomName update action with the right version guard; ignores when roomId missing.
- **e2e (dev stack):** rename a room → assert the spotlight ES doc's roomName updates + `search.rooms` returns the new name and not the old.

## Acceptance criteria
- After a rename, spotlight `roomName` = new name for every member's doc; search by new name finds the room; search by old name does not.
- Rename ordering guarded (a stale/duplicate rename event does not regress a newer name).

- The teams migration joinedAt refresh (#314) — separate.
- Whether `search.rooms` also needs the room-service enrichment path updated (confirm during implementation).

## PR / merge
- pr-self-audit is an internal pre-review gate (result NOT in PR body).
- Source repo `hmchangw/newchat`: draft-first → Jacob review → ready + `ready` label → maintainer merges.

🤖 Generated with Claude Code

chenjr0719:
## Summary

Room renames left the spotlight (room-name typeahead) index stale, so `search.rooms` returned the old name after a rename — issue #315.

Two independent gaps caused it:

1. **Producer gap.** `room-worker`'s rename handler only *federated* `room_renamed` to remote sites (via OUTBOX). A same-site rename — the common case, and the only case in a single-site deployment — published **no** INBOX event at all, so the search feed never saw it. Member events, by contrast, dual-emit: an internal-lane publish for the local search feed *and* the cross-site federate.
2. **Consumer gap.** `search-sync-worker`'s spotlight collection sourced `roomName` only from member events; it never consumed `room_renamed`.

Fix closes both. `roomName` lives on every member's spotlight doc (`{account}_{roomId}`) and the rename payload carries no account list, so the re-index is a single `_update_by_query` keyed on `roomId`.

Closes #315

## Changes

- **`room-worker/handler.go`** — after a successful rename, also publish `room_renamed` on the same-site internal INBOX lane (`chat.inbox.{siteID}.internal.room_renamed`), mirroring the existing `member_added`/`member_removed` dual-emit. Cross-site members remain covered by the existing OUTBOX federate.
- **`pkg/searchengine`** — new `UpdateByQuery(ctx, index, body)` primitive on the engine interface + HTTP adapter (`POST /{index}/_update_by_query?conflicts=proceed`), for field-keyed multi-doc mutations the DocID-keyed bulk API can't express.
- **`search-sync-worker`** — spotlight implements an optional `ByQueryCollection` seam: `room_renamed` is applied as one `_update_by_query` that sets `roomName` on every doc matching `roomId` (painless `ctx._source.roomName = params.name`), dispatched by the handler outside the bulk buffer and acked/naked on its own. `user-room`'s `BuildAction` ignores `room_renamed` (no actions = ack). The shared inbox subject filter (`subject.InboxMemberEventSubjects`) is extended with `room_renamed` on both the internal and external lanes.

**Ordering caveat (v1, by design):** the update-by-query has no per-doc version guard, so two renames delivered out of order could momentarily leave the older name — same-score typeahead reordering only, never data loss — self-correcting on the next rename or membership event. Documented inline with the upgrade path (add `roomNameUpdatedAt` to the doc + mapping and guard the script). No new ES mapping field added.

### NATS subjects

- `chat.inbox.{siteID}.internal.room_renamed` — now published by `room-worker` (same-site search feed) and consumed by `search-sync-worker`.
- `chat.inbox.{siteID}.external.room_renamed` — already federated by the OUTBOX path; now also consumed by `search-sync-worker`.

## Verification

- `go build -o /dev/null ./search-sync-worker/ ./pkg/...` — clean
- `go test ./search-sync-worker/... ./pkg/subject/... ./pkg/searchengine/... ./pkg/natsmetrics/... ./room-worker/...` — pass (new: spotlight by-query happy/fallthrough/error, user-room skip, handler dispatch ack/nak, room-worker internal-feed publish; updated every exact subject-list assertion)
- `go vet -tags integration ./search-sync-worker/... ./pkg/searchengine/... ./room-worker/...` — clean (integration compiles)
- `golangci-lint` on all touched packages — `0 issues.`

**Acceptance e2e (must run before ready):** rename a room on the dev stack → confirm the spotlight `roomName` updates and `search.rooms` returns the new name, not the old one, for a member of that room (same-site and cross-site).

## Summary by CodeRabbit

* **New Features**
  * Room renames now update searchable room names across all relevant member records.
  * Room rename events trigger same-site search reindexing before remote federation.
  * Updates safely ignore stale rename events and support retrying failed operations.

* **Bug Fixes**
  * Prevented unnecessary updates to indexes that do not store room names.
  * Added validation and clearer handling for malformed or incomplete rename events.

* **Tests**
  * Added coverage for successful updates, failures, retries, conflicts, and event validation.

### DIFF
--- a/pkg/searchengine/adapter.go
+++ b/pkg/searchengine/adapter.go
@@ -262,24 +262,33 @@ func (a *httpAdapter) UpdateByQuery(ctx context.Context, index string, body json
 	}
 	defer resp.Body.Close()
 
-	respBody, readErr := io.ReadAll(resp.Body)
+	// Bound the response read: a large per-doc failures array or a rogue proxy
+	// body shouldn't let io.ReadAll allocate without limit.
+	const maxRespBytes = 1 << 20 // 1 MiB
+	respBody, readErr := io.ReadAll(io.LimitReader(resp.Body, maxRespBytes))
 	if readErr != nil {
 		return fmt.Errorf("update by query: status %d, read body: %w", resp.StatusCode, readErr)
 	}
 	if resp.StatusCode != http.StatusOK {
 		return fmt.Errorf("update by query: status %d, body: %s", resp.StatusCode, respBody)
 	}
-	// A 200 can still carry per-doc failures or a server-side timeout; surface
-	// them so the caller Naks and retries (idempotent), matching the bulk path.
+	// A 200 can still carry per-doc failures, a server-side timeout, or (with
+	// conflicts=proceed) documents skipped on a concurrent version bump — surface
+	// all three so the caller Naks and retries. The retry is safe because the
+	// script guards on roomNameUpdatedAt (idempotent + order-independent).
 	var r struct {
-		TimedOut bool              `json:"timed_out"`
-		Failures []json.RawMessage `json:"failures"`
+		TimedOut         bool              `json:"timed_out"`
+		VersionConflicts int64             `json:"version_conflicts"`
+		Failures         []json.RawMessage `json:"failures"`
 	}
 	if err := json.Unmarshal(respBody, &r); err != nil {
 		return fmt.Errorf("update by query: decode response: %w", err)
 	}
 	if r.TimedOut {
 		return fmt.Errorf("update by query: timed out")
+	}
+	if r.VersionConflicts > 0 {
+		return fmt.Errorf("update by query: %d version conflict(s)", r.VersionConflicts)
 	}
 	if len(r.Failures) > 0 {
 		return fmt.Errorf("update by query: %d failure(s): %s", len(r.Failures), respBody)
--- a/pkg/searchengine/adapter.go
+++ b/pkg/searchengine/adapter.go
@@ -242,6 +242,51 @@ func (a *httpAdapter) UpdateMapping(ctx context.Context, indexPattern string, bo
 	return nil
 }
 
+// UpdateByQuery POSTs a painless `_update_by_query` to a single index with
+// conflicts=proceed (a concurrent version bump skips that doc, not the batch).
+func (a *httpAdapter) UpdateByQuery(ctx context.Context, index string, body json.RawMessage) error {
+	if index == "" {
+		return fmt.Errorf("update by query: index required")
+	}
+	path := fmt.Sprintf("/%s/_update_by_query?conflicts=proceed", url.PathEscape(index))
+	resp, err := a.do(ctx, "update_by_query", index, func(ctx context.Context) (*http.Request, error) {
+		req, err := http.NewRequestWithContext(ctx, http.MethodPost, path, bytes.NewReader(body))
+		if err != nil {
+			return nil, err
+		}
+		req.Header.Set("Content-Type", "application/json")
+		return req, nil
+	})
+	if err != nil {
+		return fmt.Errorf("update by query: %w", err)
+	}
+	defer resp.Body.Close()
+
+	respBody, readErr := io.ReadAll(resp.Body)
+	if readErr != nil {
+		return fmt.Errorf("update by query: status %d, read body: %w", resp.StatusCode, readErr)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("update by query: status %d, body: %s", resp.StatusCode, respBody)
+	}
+	// A 200 can still carry per-doc failures or a server-side timeout; surface
+	// them so the caller Naks and retries (idempotent), matching the bulk path.
+	var r struct {
+		TimedOut bool              `json:"timed_out"`
+		Failures []json.RawMessage `json:"failures"`
+	}
+	if err := json.Unmarshal(respBody, &r); err != nil {
+		return fmt.Errorf("update by query: decode response: %w", err)
+	}
+	if r.TimedOut {
+		return fmt.Errorf("update by query: timed out")
+	}
+	if len(r.Failures) > 0 {
+		return fmt.Errorf("update by query: %d failure(s): %s", len(r.Failures), respBody)
+	}
+	return nil
+}
+
 func (a *httpAdapter) PutScript(ctx context.Context, id string, body json.RawMessage) error {
 	resp, err := a.do(ctx, "put_script", "", func(ctx context.Context) (*http.Request, error) {
 		req, err := http.NewRequestWithContext(ctx, http.MethodPut, fmt.Sprintf("/_scripts/%s", id), bytes.NewReader(body))
--- a/pkg/searchengine/searchengine.go
+++ b/pkg/searchengine/searchengine.go
@@ -69,6 +69,14 @@ type SearchEngine interface {
 	// matching pattern; templates apply only at creation, existing indices need this.
 	UpdateMapping(ctx context.Context, indexPattern string, body json.RawMessage) error
 
+	// UpdateByQuery runs a painless `_update_by_query` against index with the
+	// given request body (query + script). It mutates every matching document,
+	// so callers use it for field-keyed multi-doc updates the DocID-keyed Bulk
+	// API can't express — e.g. re-indexing roomName on every member's spotlight
+	// doc after a room rename. Runs with conflicts=proceed so a concurrent
+	// version bump skips that one doc instead of aborting the whole batch.
+	UpdateByQuery(ctx context.Context, index string, body json.RawMessage) error
+
 	GetIndexMapping(ctx context.Context, index string) (json.RawMessage, error)
 
 	// Search executes a `_search` against the comma-joined list of indices
--- a/pkg/searchindex/spotlightdoc.go
+++ b/pkg/searchindex/spotlightdoc.go
@@ -12,6 +12,10 @@ type SpotlightDoc struct {
 	RoomType    string    `json:"roomType"    es:"keyword"`
 	SiteID      string    `json:"siteId"      es:"keyword"`
 	JoinedAt    time.Time `json:"joinedAt"    es:"date"`
+	// RoomNameUpdatedAt is the event-millis under which RoomName is valid — the
+	// LWW clock a room_renamed update-by-query guards on so an out-of-order rename
+	// can't overwrite a newer name. Stamped = the member/rename event timestamp.
+	RoomNameUpdatedAt int64 `json:"roomNameUpdatedAt" es:"long"`
 	// Origin marks provenance (e.g. model.OriginTeams for Teams-migrated
 	// rooms); empty for natively-created rooms. Filtered by search-service
 	// when SHOW_TEAMS_ROOM is false.
@@ -20,13 +24,14 @@ type SpotlightDoc struct {
 
 // SpotlightFields is the minimal, source-agnostic input to NewSpotlightDoc.
 type SpotlightFields struct {
-	UserAccount string
-	RoomID      string
-	RoomName    string
-	RoomType    string
-	SiteID      string
-	JoinedAt    time.Time
-	Origin      string
+	UserAccount       string
+	RoomID            string
+	RoomName          string
+	RoomType          string
+	SiteID            string
+	JoinedAt          time.Time
+	RoomNameUpdatedAt int64
+	Origin            string
 }
 
 // NewSpotlightDoc builds the ES document for the spotlight index from f.
--- a/pkg/subject/subject.go
+++ b/pkg/subject/subject.go
@@ -309,18 +309,22 @@ func InboxExternalAll(siteID string) string {
 	return fmt.Sprintf("chat.inbox.%s.external.>", siteID)
 }
 
-// InboxMemberEventSubjects returns the subject filters a consumer should use to
-// receive member_added/member_removed events on both the internal (same-site)
-// and external (cross-site) lanes for the given site. Use with
+// InboxMemberEventSubjects returns the subject filters a search-sync consumer
+// should use to receive the room-index-affecting events on both the internal
+// (same-site) and external (cross-site) lanes for the given site:
+// member_added/member_removed (per-member docs) and room_renamed (re-index the
+// room name across every member's doc). Use with
 // jetstream.ConsumerConfig.FilterSubjects (NATS 2.10+).
 func InboxMemberEventSubjects(siteID string) []string {
 	return []string{
 		InboxInternal(siteID, "member_added"),
 		InboxInternal(siteID, "member_removed"),
 		InboxInternal(siteID, "member_joinedat_refreshed"),
+		InboxInternal(siteID, "room_renamed"),
 		InboxExternal(siteID, "member_added"),
 		InboxExternal(siteID, "member_removed"),
 		InboxExternal(siteID, "member_joinedat_refreshed"),
+		InboxExternal(siteID, "room_renamed"),
 	}
 }
 
--- a/room-worker/handler.go
+++ b/room-worker/handler.go
@@ -2358,13 +2358,14 @@ func (h *Handler) processRoomRename(ctx context.Context, data []byte) (err error
 	if err != nil {
 		return fmt.Errorf("marshal internal rename event: %w", err)
 	}
-	// Return on failure so the ROOMS message redelivers rather than silently
-	// losing the spotlight re-index (for a single-site room there is no OUTBOX
-	// federate to cover it). Safe to retry: the publish carries a requestID-keyed
-	// dedup id, the sys message a deterministic id, and UpdateRoomName is an
-	// idempotent set — a redelivery re-runs them all without duplicating.
+	// Best-effort: log, don't Nak. Returning here would redeliver the whole
+	// handler, and UpdateRoomName / UpdateSubscriptionNamesForRoom (already
+	// committed above) are unconditional writes — a redelivered stale rename
+	// would revert newer canonical state. The spotlight re-index is a derived
+	// cache; a rare missed publish self-corrects on the next rename/membership
+	// event and is far cheaper than risking source-of-truth reversion.
 	if err := h.publish(ctx, subject.InboxInternal(h.siteID, model.InboxRoomRenamed), internalRenameData, natsutil.InboxDedupID(ctx, h.siteID, requestID)); err != nil {
-		return fmt.Errorf("publish local inbox room_renamed: %w", err)
+		slog.ErrorContext(ctx, "local inbox room_renamed publish failed (best-effort)", "error", err, "room_id", req.RoomID)
 	}
 	// Relay room_renamed through the OUTBOX ordered lane (not a direct INBOX
 	// publish) so it shares the per-destination FIFO consumer with member_added
--- a/room-worker/handler.go
+++ b/room-worker/handler.go
@@ -2354,9 +2354,17 @@ func (h *Handler) processRoomRename(ctx context.Context, data []byte) (err error
 		Payload:    renamedPayload,
 		Timestamp:  now,
 	}
-	internalRenameData, _ := json.Marshal(internalRename)
+	internalRenameData, err := json.Marshal(internalRename)
+	if err != nil {
+		return fmt.Errorf("marshal internal rename event: %w", err)
+	}
+	// Return on failure so the ROOMS message redelivers rather than silently
+	// losing the spotlight re-index (for a single-site room there is no OUTBOX
+	// federate to cover it). Safe to retry: the publish carries a requestID-keyed
+	// dedup id, the sys message a deterministic id, and UpdateRoomName is an
+	// idempotent set — a redelivery re-runs them all without duplicating.
 	if err := h.publish(ctx, subject.InboxInternal(h.siteID, model.InboxRoomRenamed), internalRenameData, natsutil.InboxDedupID(ctx, h.siteID, requestID)); err != nil {
-		slog.ErrorContext(ctx, "local inbox room_renamed publish failed", "error", err, "room_id", req.RoomID)
+		return fmt.Errorf("publish local inbox room_renamed: %w", err)
 	}
 	// Relay room_renamed through the OUTBOX ordered lane (not a direct INBOX
 	// publish) so it shares the per-destination FIFO consumer with member_added
--- a/room-worker/handler.go
+++ b/room-worker/handler.go
@@ -2342,6 +2342,22 @@ func (h *Handler) processRoomRename(ctx context.Context, data []byte) (err error
 	if err != nil {
 		return fmt.Errorf("marshal rename inbox payload: %w", err)
 	}
+	now := time.Now().UTC().UnixMilli()
+	// Same-site search feed: search-sync re-indexes the room name from this
+	// internal-lane event (mirrors the member_added/removed internal publish).
+	// Cross-site members are covered by the OUTBOX federate below; a single-site
+	// rename has no remote sites, so without this the spotlight index goes stale.
+	internalRename := model.InboxEvent{
+		Type:       model.InboxRoomRenamed,
+		SiteID:     h.siteID,
+		DestSiteID: h.siteID,
+		Payload:    renamedPayload,
+		Timestamp:  now,
+	}
+	internalRenameData, _ := json.Marshal(internalRename)
+	if err := h.publish(ctx, subject.InboxInternal(h.siteID, model.InboxRoomRenamed), internalRenameData, natsutil.InboxDedupID(ctx, h.siteID, requestID)); err != nil {
+		slog.ErrorContext(ctx, "local inbox room_renamed publish failed", "error", err, "room_id", req.RoomID)
+	}
 	// Relay room_renamed through the OUTBOX ordered lane (not a direct INBOX
 	// publish) so it shares the per-destination FIFO consumer with member_added
 	// and cannot overtake the add that creates a new member's subscription — a
@@ -2350,7 +2366,6 @@ func (h *Handler) processRoomRename(ctx context.Context, data []byte) (err error
 	// per-room ordering, so member.add and room.rename (separate ROOMS messages,
 	// concurrent workers) can still enqueue out of causal order. Full ordering
 	// awaits producer per-room lanes / reconciliation (design doc §5, §7).
-	now := time.Now().UTC().UnixMilli()
 	for _, remoteSiteID := range remoteSites {
 		if err := h.federate(ctx, req.RoomID, remoteSiteID, model.InboxRoomRenamed,
 			renamedPayload, natsutil.InboxDedupID(ctx, remoteSiteID, requestID), now); err != nil {
--- a/search-sync-worker/handler.go
+++ b/search-sync-worker/handler.go
@@ -102,15 +102,18 @@ func (h *Handler) AddWithContext(ctx context.Context, msg jetstream.Msg) {
 			// Parse/validation poison — Ack drops it (same as BuildAction).
 			slog.ErrorContext(ctx, "build by-query", "error", bqErr, "subject", msg.Subject(), "consumer", h.collection.ConsumerName())
 			natsutil.Ack(msg, "build by-query failed")
+			h.metrics.recordMessages(ctx, dispAckedPoison, 1)
 			return
 		}
 		if ok {
 			if err := h.store.UpdateByQuery(ctx, index, body); err != nil {
 				slog.ErrorContext(ctx, "update-by-query failed", "error", err, "index", index, "consumer", h.collection.ConsumerName())
 				natsutil.Nak(msg, "update-by-query failed")
+				h.metrics.recordMessages(ctx, dispNakkedRequestFailed, 1)
 				return
 			}
 			natsutil.Ack(msg, "update-by-query succeeded")
+			h.metrics.recordMessages(ctx, dispAckedSuccess, 1)
 			return
 		}
 	}
--- a/search-sync-worker/handler.go
+++ b/search-sync-worker/handler.go
@@ -90,6 +90,31 @@ func (h *Handler) AddWithContext(ctx context.Context, msg jetstream.Msg) {
 		h.metrics.recordMessages(ctx, dispAckedPoison, 1)
 		return
 	}
+
+	// By-query events (e.g. room rename) mutate many docs by field, not by
+	// DocID, so they can't ride the bulk buffer — apply each as a standalone
+	// _update_by_query and ack/nak it on its own. A collection that doesn't
+	// implement ByQueryCollection, or a message it doesn't claim (ok=false),
+	// falls through to the normal bulk path.
+	if bq, isBQ := h.collection.(ByQueryCollection); isBQ {
+		index, body, ok, bqErr := bq.BuildByQuery(data)
+		if bqErr != nil {
+			// Parse/validation poison — Ack drops it (same as BuildAction).
+			slog.ErrorContext(ctx, "build by-query", "error", bqErr, "subject", msg.Subject(), "consumer", h.collection.ConsumerName())
+			natsutil.Ack(msg, "build by-query failed")
+			return
+		}
+		if ok {
+			if err := h.store.UpdateByQuery(ctx, index, body); err != nil {
+				slog.ErrorContext(ctx, "update-by-query failed", "error", err, "index", index, "consumer", h.collection.ConsumerName())
+				natsutil.Nak(msg, "update-by-query failed")
+				return
+			}
+			natsutil.Ack(msg, "update-by-query succeeded")
+			return
+		}
+	}
+
 	actions, err := h.collection.BuildAction(data)
 	if err != nil {
 		// Every BuildAction error is parse/validation poison — Ack drops it for
--- a/search-sync-worker/spotlight.go
+++ b/search-sync-worker/spotlight.go
@@ -135,7 +135,10 @@ func (c *spotlightCollection) BuildByQuery(data []byte) (string, json.RawMessage
 	if p.NewName == "" {
 		return "", nil, false, fmt.Errorf("build spotlight rename: missing newName")
 	}
-	body, err := buildSpotlightRenameByQuery(p.RoomID, p.NewName)
+	if p.Timestamp <= 0 {
+		return "", nil, false, fmt.Errorf("build spotlight rename: missing timestamp")
+	}
+	body, err := buildSpotlightRenameByQuery(p.RoomID, p.NewName, p.Timestamp)
 	if err != nil {
 		return "", nil, false, fmt.Errorf("build spotlight rename: %w", err)
 	}
@@ -143,20 +146,25 @@ func (c *spotlightCollection) BuildByQuery(data []byte) (string, json.RawMessage
 }
 
 // buildSpotlightRenameByQuery is the _update_by_query that sets roomName on every
-// doc of the renamed room. Ordering caveat (v1, no guard): with out-of-order
-// renames the older name can win and persist until the next rename (a later
-// member event stamps whatever name it was born with, so it doesn't reliably
-// correct forward). Spotlight is a derived cache — the room doc is source of
-// truth — so this is stale typeahead text, never data loss. Upgrade path: add
-// roomNameUpdatedAt to SpotlightDoc + its mapping and guard the script with
-// `if (params.ts > ctx._source.roomNameUpdatedAt)`.
-func buildSpotlightRenameByQuery(roomID, newName string) (json.RawMessage, error) {
+// doc of the renamed room, guarded by roomNameUpdatedAt so an out-of-order rename
+// can't overwrite a newer name: only a strictly-newer ts wins, and it stamps the
+// clock it just advanced. A stale delivery is a no-op (updated:0), converging on
+// last-write-wins regardless of arrival order.
+//
+// Residual (bounded, documented): a member added *concurrently* with a rename can
+// have its doc created (from the member event's name snapshot) only after the
+// rename's update-by-query already ran on the then-absent doc, so that one new
+// member can see the pre-rename name until the next event touches their doc.
+// Closing it fully needs query-time room-name resolution (a room doc + join),
+// out of scope for this typeahead cache — spotlight is derived, not source of truth.
+func buildSpotlightRenameByQuery(roomID, newName string, ts int64) (json.RawMessage, error) {
 	body, err := json.Marshal(map[string]any{
 		"query": map[string]any{"term": map[string]any{"roomId": roomID}},
 		"script": map[string]any{
-			"lang":   "painless",
-			"source": "ctx._source.roomName = params.name",
-			"params": map[string]any{"name": newName},
+			"lang": "painless",
+			"source": "long stored = ctx._source.roomNameUpdatedAt == null ? 0L : ((Number)ctx._source.roomNameUpdatedAt).longValue(); " +
+				"if (params.ts > stored) { ctx._source.roomName = params.name; ctx._source.roomNameUpdatedAt = params.ts; } else { ctx.op = 'noop'; }",
+			"params": map[string]any{"name": newName, "ts": ts},
 		},
 	})
 	if err != nil {
@@ -173,7 +181,10 @@ func newSpotlightSearchIndex(account string, evt *model.InboxMemberEvent) search
 		RoomType:    string(evt.RoomType),
 		SiteID:      evt.SiteID,
 		JoinedAt:    convertJoinedAt(evt.JoinedAt),
-		Origin:      evt.Origin,
+		// Stamp the name's LWW clock so a later rename (higher ts) wins and this
+		// doc's own name can't be reverted by an older rename delivered late.
+		RoomNameUpdatedAt: evt.Timestamp,
+		Origin:            evt.Origin,
 	})
 }
 
--- a/search-sync-worker/spotlight.go
+++ b/search-sync-worker/spotlight.go
@@ -162,8 +162,11 @@ func buildSpotlightRenameByQuery(roomID, newName string, ts int64) (json.RawMess
 		"query": map[string]any{"term": map[string]any{"roomId": roomID}},
 		"script": map[string]any{
 			"lang": "painless",
+			// >= (not >) so two renames sharing a millisecond both apply (last
+			// processed wins the tie) instead of dropping the second, and a
+			// redelivery of the same rename re-applies its own name idempotently.
 			"source": "long stored = ctx._source.roomNameUpdatedAt == null ? 0L : ((Number)ctx._source.roomNameUpdatedAt).longValue(); " +
-				"if (params.ts > stored) { ctx._source.roomName = params.name; ctx._source.roomNameUpdatedAt = params.ts; } else { ctx.op = 'noop'; }",
+				"if (params.ts >= stored) { ctx._source.roomName = params.name; ctx._source.roomNameUpdatedAt = params.ts; } else { ctx.op = 'noop'; }",
 			"params": map[string]any{"name": newName, "ts": ts},
 		},
 	})
--- a/search-sync-worker/spotlight.go
+++ b/search-sync-worker/spotlight.go
@@ -113,6 +113,58 @@ func (c *spotlightCollection) BuildAction(data []byte) ([]searchengine.BulkActio
 	return actions, nil
 }
 
+// BuildByQuery handles room_renamed: roomName lives on every member's spotlight
+// doc (`{account}_{roomId}`), and the rename payload carries no account list, so
+// a single _update_by_query keyed on roomId re-indexes them all. Returns ok=false
+// for member events, which fall through to BuildAction's bulk path.
+func (c *spotlightCollection) BuildByQuery(data []byte) (string, json.RawMessage, bool, error) {
+	var evt model.InboxEvent
+	if err := json.Unmarshal(data, &evt); err != nil {
+		return "", nil, false, fmt.Errorf("unmarshal inbox event: %w", err)
+	}
+	if evt.Type != model.InboxRoomRenamed {
+		return "", nil, false, nil
+	}
+	var p model.RoomRenamedInboxPayload
+	if err := json.Unmarshal(evt.Payload, &p); err != nil {
+		return "", nil, false, fmt.Errorf("unmarshal room_renamed payload: %w", err)
+	}
+	if p.RoomID == "" {
+		return "", nil, false, fmt.Errorf("build spotlight rename: missing roomId")
+	}
+	if p.NewName == "" {
+		return "", nil, false, fmt.Errorf("build spotlight rename: missing newName")
+	}
+	body, err := buildSpotlightRenameByQuery(p.RoomID, p.NewName)
+	if err != nil {
+		return "", nil, false, fmt.Errorf("build spotlight rename: %w", err)
+	}
+	return c.indexName, body, true, nil
+}
+
+// buildSpotlightRenameByQuery is the _update_by_query that sets roomName on every
+// doc of the renamed room. Ordering caveat (v1, no guard): with out-of-order
+// renames the older name can win and persist until the next rename (a later
+// member event stamps whatever name it was born with, so it doesn't reliably
+// correct forward). Spotlight is a derived cache — the room doc is source of
+// truth — so this is stale typeahead text, never data loss. Upgrade path: add
+// roomNameUpdatedAt to SpotlightDoc + its mapping and guard the script with
+// `if (params.ts > ctx._source.roomNameUpdatedAt)`.
+func buildSpotlightRenameByQuery(roomID, newName string) (json.RawMessage, error) {
+	body, err := json.Marshal(map[string]any{
+		"query": map[string]any{"term": map[string]any{"roomId": roomID}},
+		"script": map[string]any{
+			"lang":   "painless",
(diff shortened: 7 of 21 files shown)


julianshen:
**1 finding · 1 inline comment posted** — 🔴 1 blocking

### Disputed

- `search-sync-worker/spotlight.go:162` (`distributed-system`) — The room-worker comments acknowledge that separate room messages can be enqueued out of causal order. If room_renamed reaches this term query before a delayed member_added event, the query matches nothing and the rename is permanently acknowledged. The later member event can then create a document containing its stale name snapshot, and no subsequent event is guaranteed to repair it. A local bulk barrier does not address this producer-side ordering. Establish per-room ordering, resolve the canonical room name/version when creating spotlight documents, or add durable reconciliation for documents created after a rename.
- `room-worker/handler.go:2367` (`distributed-system`) — For a single-site room, UpdateRoomName and UpdateSubscriptionNamesForRoom can commit, then this publish can fail before JetStream accepts the event. The failure is only logged, so the source message is acknowledged and every existing spotlight document can retain the old name indefinitely. A later membership event only updates that member's document and is not room-wide reconciliation. Persist the local event in an outbox atomically with the canonical rename, or add a monotonic/idempotency fence that makes retrying publication safe without replaying stale database writes.
- `search-sync-worker/spotlight.go:162` (`elasticsearch`) — This term query only updates documents present in the index snapshot. The room-worker comments acknowledge that member_added and room_renamed can be produced out of causal order; if the rename arrives first, Elasticsearch returns success with zero matches and the worker acknowledges it, after which the delayed member event creates a document with the old name. This remains possible even after adding a local bulk barrier. Establish producer ordering, resolve an authoritative room-name revision during member writes, or add durable reconciliation.
- `room-worker/handler.go:2367` (`elasticsearch`) — If this publish fails before broker acceptance, the handler logs the error and returns success for a single-site room, so no `_update_by_query` event remains to repair existing spotlight documents. A later membership event only rewrites its own member document and does not reconcile all existing members, so the stale names can persist indefinitely. Persist the local rename in an outbox, or add a monotonic/idempotency fence that permits safe retry after the canonical writes.
- `room-worker/handler.go:2367` (`mongodb`) — After the MongoDB room and subscription-name updates commit, a failure publishing the local room_renamed event is only logged. For a same-site room with no remote destinations, the handler then succeeds and the source message is acknowledged, while every existing spotlight document can retain the old name indefinitely. A later membership event repairs only that member's document, not the room-wide duplication. Persist a retryable outbox record atomically with the canonical rename, or add a monotonic MongoDB revision/idempotency fence so the handler can safely retry publication without allowing an older redelivery to overwrite newer room state.
- `search-sync-worker/spotlight.go:163` (`nats`) — The room-worker comments acknowledge that concurrent room messages can enter NATS out of causal order. If room_renamed is consumed before a delayed member_added carrying the old name, the term query succeeds with zero matches and is Acked; the later member event then creates the stale document with no remaining rename to repair it. A local bulk barrier cannot fix reversed stream order. Establish per-room producer ordering, resolve authoritative room name/revision when processing member events, or add durable reconciliation.
- `room-worker/handler.go:2367` (`nats`) — After the canonical database writes commit, failure to publish chat.inbox.<site>.internal.room_renamed is only logged. For a single-site room, processRoomRename then succeeds and the source message is Acked even though NATS never received the search event, leaving all existing spotlight documents stale indefinitely. A later membership event repairs only its own document. Persist the local event in an outbox with the rename, or add a monotonic/idempotency fence that makes retrying the handler safe without reverting newer canonical state.
- `room-worker/handler.go:2367` (`silent-failure`) — If the internal room_renamed publish fails, the handler only logs the error and can return success, causing the source message to be acknowledged. For a single-site room this permanently loses the only event that updates existing spotlight documents; a later membership event updates only its affected member document, so the other stale documents are not necessarily repaired. Make publication durable with an outbox committed alongside the rename, or add a monotonic/idempotency fence to the canonical writes so the handler can safely return the publish error and retry without reverting a newer rename.
- `search-sync-worker/spotlight.go:162` (`tgd-review`) — This term query only updates documents that already exist. Independently of the local bulk-buffer issue, the room-worker comments acknowledge that concurrent room messages can be enqueued out of causal order. If the rename runs before a delayed member_added, Elasticsearch reports success with zero matches and the rename is acknowledged; the later event can create a stale document with no guaranteed future repair. Establish per-room producer ordering, resolve canonical room state during document creation, or add durable reconciliation.
- `room-worker/handler.go:2367` (`tgd-review`) — After the canonical writes commit, a broker failure here is only logged, so the source message can be acknowledged without any local room_renamed event. For a single-site room, all existing spotlight documents may then remain stale indefinitely; a later membership event only touches its own document and is not full-room reconciliation. Persist publication through an outbox, or add a monotonic database fence that makes retrying the handler safe without allowing stale canonical writes to revert newer state.

### Related work

- [Issue #315](https://github.com/hmchangw/newchat/issues/315) — \[Fix\] room rename does not re-index the spotlight room name \(stale search-by-name\) (open)

### Local review memory

- `search-sync-worker/spotlight.go:163` — @chenjr0719: This is the concurrent-member-during-rename residual documented in the code comment on buildSpotlightRenameByQuery . Spotlight is a derived typeahead cache (the room doc is the source of truth), so a newly-added member briefly seeing the… ([thread](https://github.com/hmchangw/newchat/pull/316#discussion_r3819724508))
- `room-worker/handler.go:2370` — @chenjr0719: Good catch — reverting this to best-effort (log, don't Nak) in the next push. You're right that Naking redelivers the whole handler and the already-committed UpdateRoomName / UpdateSubscriptionNamesForRoom are unconditional writes, so a re… ([thread](https://github.com/hmchangw/newchat/pull/316#discussion_r3819724521))
- `pkg/searchindex/spotlightdoc.go:18` — @chenjr0719: NewSpotlightDoc uses SpotlightDoc(f) — a Go struct conversion between two structs with identical fields (conversion ignores tags), so it copies **every** field by position, including RoomNameUpdatedAt . A field mismatch would be a com… ([thread](https://github.com/hmchangw/newchat/pull/316#discussion_r3819924820))
- `room-worker/handler.go:2367` — @chenjr0719: This is in tension with the earlier finding on this same line, where Naking here was flagged for letting a redelivered stale rename revert canonical state via the unconditional UpdateRoomName / UpdateSubscriptionNamesForRoom . Nak risks s… ([thread](https://github.com/hmchangw/newchat/pull/316#discussion_r3819924817))

<details>

* `pkg/searchengine/adapter.go`
* `pkg/searchengine/adapter_test.go`
* `pkg/searchengine/searchengine.go`
* `pkg/searchindex/spotlightdoc.go`
* `pkg/searchindex/spotlightdoc_test.go`
* `pkg/subject/subject.go`
* `room-worker/handler.go`
* `search-sync-worker/collection.go`
* `search-sync-worker/handler_test.go`
* `search-sync-worker/inbox_stream.go`
* `search-sync-worker/main_test.go`
* `search-sync-worker/mock_store_test.go`
* `search-sync-worker/spotlight.go`
* `search-sync-worker/spotlight_test.go`
* `search-sync-worker/store.go`
* `search-sync-worker/user_room.go`
* `search-sync-worker/user_room_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

search-sync-worker/handler.go
```diff
@@ -90,6 +90,31 @@ func (h *Handler) AddWithContext(ctx context.Context, msg jetstream.Msg) {
...
+	if bq, isBQ := h.collection.(ByQueryCollection); isBQ {
+		index, body, ok, bqErr := bq.BuildByQuery(data)
+		if bqErr != nil {
+			// Parse/validation poison — Ack drops it (same as BuildAction).
+			slog.ErrorContext(ctx, "build by-query", "error", bqErr, "subject", msg.Subject(), "consumer", h.collection.ConsumerName())
+			natsutil.Ack(msg, "build by-query failed")
+			return
+		}
+		if ok {
+			if err := h.store.UpdateByQuery(ctx, index, body); err != nil {
```
julianshen, line 109: _🏷️ ordering_ | _🔴 Blocking_ | _`distributed-system`, `performance-and-concurrency` (2 rules)_

**Standalone renames overtake buffered member writes.**

A member_added event can be accepted first and remain in the bulk buffer. Before it is flushed, the next room_renamed event immediately executes UpdateByQuery; the new member document is absent, so Elasticsearch reports success and the rename is acknowledged. A later flush then creates that document with the pre-rename roomName, with no remaining event to correct it. Preserve the consumer's ordering by treating a rename as a barrier: successfully flush and acknowledge all prior bulk work before running it, while serializing against concurrent Add/Flush calls, or place both operations in one ordered pipeline. Add an add→rename→flush regression test.

<details>

- **`performance-and-concurrency`**:110: This standalone call is not ordered with the handler's buffered or in-flight bulk work. For example, a member_added event can buffer creation of a spotlight document with the old room name, then a following room_renamed executes immediately and finds no document; the later bulk flush creates the document with the stale name. An already-running flush can similarly complete after the rename and overwrite it. Before executing the by-query operation, establish a barrier that flushes and waits for all preceding bulk work and prevents subsequent work from overtaking it, or put both operation types through one ordered execution queue.

</details>

<details>

```
In `search-sync-worker/handler.go` around line 108: A member_added event can be accepted first and remain in the bulk buffer. Before it is flushed, the next room_renamed event immediately executes UpdateByQuery; the new member document is absent, so Elasticsearch reports success and the rename is acknowledged. A later flush then creates that document with the pre-rename roomName, with no remaining event to correct it. Preserve the consumer's ordering by treating a rename as a barrier: successfully flush and acknowledge all prior bulk work before running it, while serializing against concurrent Add/Flush calls, or place both operations in one ordered pipeline. Add an add→rename→flush regression test.

```

</details>

room-worker/handler.go
```diff
@@ -2342,6 +2342,22 @@ func (h *Handler) processRoomRename(ctx context.Context, data []byte) (err error
...
+	internalRename := model.InboxEvent{
+		Type:       model.InboxRoomRenamed,
+		SiteID:     h.siteID,
+		DestSiteID: h.siteID,
+		Payload:    renamedPayload,
+		Timestamp:  now,
+	}
+	internalRenameData, _ := json.Marshal(internalRename)
+	if err := h.publish(ctx, subject.InboxInternal(h.siteID, model.InboxRoomRenamed), internalRenameData, natsutil.InboxDedupID(ctx, h.siteID, requestID)); err != nil {
+		slog.ErrorContext(ctx, "local inbox room_renamed publish failed", "error", err, "room_id", req.RoomID)
```
julianshen: _🏷️ delivery_ | _🔴 Blocking_ | _`distributed-system`, `elasticsearch`, `nats`, `silent-failure`, `tgd-review` (5 rules)_

**A failed local rename publication is permanently ignored.**

The room and subscription names are updated before this publish. If the local JetStream publish then fails, the error is only logged; for a room with no remote sites, processRoomRename can still return success and the source message is acknowledged. The spotlight rename event is therefore permanently lost and local search remains stale. Durably enqueue this event transactionally with the rename via an outbox and relay it asynchronously. At minimum, propagate the publish error so the source message is retried, although that alone does not close the crash window between the database writes and publication.

<details>

- **`nats`**:2358: On the single-site path, if publishing `chat.inbox.<site>.internal.room_renamed` fails, the error is only logged. With no remote sites, `processRoomRename` then returns success, so the source message can be acknowledged even though the search-sync event was never persisted. A transient NATS disconnect therefore leaves the spotlight name stale with no retry. Return the publish error so the existing message retry path can republish it; the deduplication ID handles an ambiguous prior publish.
- **`tgd-review`**:2358: The local search-feed publish logs its error and lets processRoomRename continue successfully. Because the source room and subscriptions were already updated, the message can be acknowledged without emitting room_renamed, leaving same-site spotlight documents stale with no retry or reconciliation. Propagate the error so message redelivery and the deduplication ID can recover, and cover this failure path in a test.
- **`elasticsearch`**:2358: When the internal search-feed publish fails, the error is only logged. For a room with no remote sites, processing then returns success, so the originating rename is acknowledged even though search-sync never receives an event and every spotlight document keeps the old name. Propagate the publish error so the request is retried; the existing deterministic deduplication ID prevents duplicate feed publication.
- **`silent-failure`**:2358: If the internal INBOX publish fails, the handler only logs the error and continues. With no remote sites, processRoomRename then succeeds, allowing the source message to be acknowledged even though search-sync never receives the rename; every same-site spotlight document can remain stale permanently. Return the publish error so the source message is retried, and add a test covering this failure path.

</details>

<details>

```
In `room-worker/handler.go` around line 2359: The room and subscription names are updated before this publish. If the local JetStream publish then fails, the error is only logged; for a room with no remote sites, processRoomRename can still return success and the source message is acknowledged. The spotlight rename event is therefore permanently lost and local search remains stale. Durably enqueue this event transactionally with the rename via an outbox and relay it asynchronously. At minimum, propagate the publish error so the source message is retried, although that alone does not close the crash window between the database writes and publication.

```

</details>

pkg/searchengine/adapter.go
```diff
@@ -242,6 +242,51 @@ func (a *httpAdapter) UpdateMapping(ctx context.Context, indexPattern string, bo
...
+	respBody, readErr := io.ReadAll(resp.Body)
+	if readErr != nil {
+		return fmt.Errorf("update by query: status %d, read body: %w", resp.StatusCode, readErr)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("update by query: status %d, body: %s", resp.StatusCode, respBody)
+	}
+	// A 200 can still carry per-doc failures or a server-side timeout; surface
+	// them so the caller Naks and retries (idempotent), matching the bulk path.
+	var r struct {
```
julianshen, line 279: _🏷️ consistency_ | _🔴 Blocking_ | _`distributed-system`, `elasticsearch`, `nats`, `performance-and-concurrency`, `tgd-review` (5 rules)_

**Version-conflicted documents are acknowledged without being renamed.**

With conflicts=proceed, Elasticsearch reports skipped documents in the numeric version_conflicts field, normally leaving failures empty. If a concurrent bulk write increments a matching document's version, that document is skipped, this adapter returns nil, and the handler acknowledges the rename permanently. The added test incorrectly models a conflict as a failures entry. Parse version_conflicts and return an error so the idempotent rename is retried, and add a test using a real response shape such as {"version_conflicts":1,"failures":[]}.

<details>

- **`performance-and-concurrency`**:275: Elasticsearch reports documents skipped because of version conflicts in the top-level `version_conflicts` count, not in `failures`. If a concurrent bulk update bumps a matching document after the query snapshot, this request returns 200 with `version_conflicts > 0` and an empty failures array; the adapter returns nil and the handler ACKs the rename, leaving that document stale permanently. Treat a nonzero count as retryable and add a test using the actual response shape.

<details>

```text
var r struct {
		TimedOut        bool              `json:"timed_out"`
		VersionConflicts int              `json:"version_conflicts"`
		Failures        []json.RawMessage `json:"failures"`
	}
	if err := json.Unmarshal(respBody, &r); err != nil {
		return fmt.Errorf("update by query: decode response: %w", err)
	}
	if r.TimedOut {
		return fmt.Errorf("update by query: timed out")
	}
	if r.VersionConflicts > 0 {
		return fmt.Errorf("update by query: %d version conflict(s)", r.VersionConflicts)
	}
	if len(r.Failures) > 0 {
		return fmt.Errorf("update by query: %d failure(s): %s", len(r.Failures), respBody)
	}
```

</details>

- **`nats`**:274: With `conflicts=proceed`, Elasticsearch reports skipped concurrent updates in the top-level `version_conflicts` count rather than necessarily placing them in `failures`. The adapter currently returns nil for such a 200 response, causing the room-renamed JetStream message to be Acked while some member documents retain the old name. Treat a nonzero count as an error so the handler Naks and retries the idempotent update, and add a response test covering this field.
- **`elasticsearch`**:274: With `conflicts=proceed`, Elasticsearch reports skipped conflicting documents in the top-level `version_conflicts` count, not necessarily in `failures`. The current decoder therefore returns success, causing the message to be Acked while those documents retain the old room name. Decode `version_conflicts` and return an error so the idempotent assignment is retried. Add a test using `{"version_conflicts":1,"failures":[]}`.
- **`tgd-review`**:274: With conflicts=proceed, Elasticsearch reports skipped version conflicts in the top-level version_conflicts count while still returning HTTP 200; they are not normally included in failures. This response decoder therefore returns nil, causing the handler to acknowledge the rename even though some matching documents were not updated. Decode the count and return a retryable error when it is nonzero.

</details>

<details>

```
In `pkg/searchengine/adapter.go` around line 274: With conflicts=proceed, Elasticsearch reports skipped documents in the numeric version_conflicts field, normally leaving failures empty. If a concurrent bulk write increments a matching document's version, that document is skipped, this adapter returns nil, and the handler acknowledges the rename permanently. The added test incorrectly models a conflict as a failures entry. Parse version_conflicts and return an error so the idempotent rename is retried, and add a test using a real response shape such as {"version_conflicts":1,"failures":[]}.

```

</details>

search-sync-worker/spotlight.go
```diff
@@ -113,6 +113,58 @@ func (c *spotlightCollection) BuildAction(data []byte) ([]searchengine.BulkActio
...
+// doc of the renamed room. Ordering caveat (v1, no guard): with out-of-order
+// renames the older name can win and persist until the next rename (a later
+// member event stamps whatever name it was born with, so it doesn't reliably
+// correct forward). Spotlight is a derived cache — the room doc is source of
+// truth — so this is stale typeahead text, never data loss. Upgrade path: add
+// roomNameUpdatedAt to SpotlightDoc + its mapping and guard the script with
+// `if (params.ts > ctx._source.roomNameUpdatedAt)`.
+func buildSpotlightRenameByQuery(roomID, newName string) (json.RawMessage, error) {
+	body, err := json.Marshal(map[string]any{
+		"query": map[string]any{"term": map[string]any{"roomId": roomID}},
```
julianshen, line 162: _🏷️ event-ordering_ | _🔴 Blocking_ | _`elasticsearch`_

**A rename can be acknowledged before the member document exists.**

This term-scoped update can only mutate documents already present. The changed room-worker code explicitly notes that member-added and room-renamed events can be enqueued out of causal order. If the rename runs first, Elasticsearch returns success with `updated: 0`; the worker Acks it, and the delayed member event can then create a document containing the old name with no later correction. A timestamp guard cannot fix a document absent during the rename. Establish per-room ordering, consult canonical/current rename state when creating spotlight documents, or add reconciliation that repairs documents created after the rename.

<details>

```
In `search-sync-worker/spotlight.go` around line 155: This term-scoped update can only mutate documents already present. The changed room-worker code explicitly notes that member-added and room-renamed events can be enqueued out of causal order. If the rename runs first, Elasticsearch returns success with `updated: 0`; the worker Acks it, and the delayed member event can then create a document containing the old name with no later correction. A timestamp guard cannot fix a document absent during the rename. Establish per-room ordering, consult canonical/current rename state when creating spotlight documents, or add reconciliation that repairs documents created after the rename.

```

</details>

julianshen (commented):
tGD inline review

search-sync-worker/spotlight.go
```diff
@@ -113,6 +113,58 @@ func (c *spotlightCollection) BuildAction(data []byte) ([]searchengine.BulkActio
...
+		"script": map[string]any{
+			"lang":   "painless",
+			"source": "ctx._source.roomName = params.name",
```
julianshen: _🎯 concurrency_ | _🔴 Blocking_ | _`performance-and-concurrency`_

**Out-of-order renames can permanently restore an older name**

The script applies every rename unconditionally even though the room-worker comments establish that concurrent room messages can be published out of causal order. If rename B is processed before older rename A, A overwrites B and no reconciliation corrects the spotlight documents. Persist a mapped room-name update timestamp/version, pass `p.Timestamp` into this request, and make the script update only when the incoming version is newer. Member-document writes must honor the same version so they cannot overwrite a newer rename.

<details>

```
In `search-sync-worker/spotlight.go` around line 158: The script applies every rename unconditionally even though the room-worker comments establish that concurrent room messages can be published out of causal order. If rename B is processed before older rename A, A overwrites B and no reconciliation corrects the spotlight documents. Persist a mapped room-name update timestamp/version, pass `p.Timestamp` into this request, and make the script update only when the incoming version is newer. Member-document writes must honor the same version so they cannot overwrite a newer rename.

```

</details>

search-sync-worker/spotlight.go
```diff
@@ -113,6 +113,58 @@ func (c *spotlightCollection) BuildAction(data []byte) ([]searchengine.BulkActio
...
+	if err := json.Unmarshal(evt.Payload, &p); err != nil {
+		return "", nil, false, fmt.Errorf("unmarshal room_renamed payload: %w", err)
+	}
+	if p.RoomID == "" {
+		return "", nil, false, fmt.Errorf("build spotlight rename: missing roomId")
+	}
+	if p.NewName == "" {
+		return "", nil, false, fmt.Errorf("build spotlight rename: missing newName")
+	}
+	body, err := buildSpotlightRenameByQuery(p.RoomID, p.NewName)
```
julianshen: _🏷️ ordering_ | _🟠 Warning_ | _`distributed-system`, `elasticsearch`, `nats`, `tgd-review` (4 rules)_

**An older rename delivery can overwrite a newer room name.**

The payload timestamp is discarded and the Painless script unconditionally assigns `roomName`. Two rename commands can be published out of causal order by concurrent room workers, and a rename Naked after an Elasticsearch failure can also be redelivered after a newer rename was Acked. In either path, the older event executes last and permanently restores the old name. Persist an authoritative monotonic rename revision on spotlight documents and update only when the incoming revision is newer; if `Timestamp` is guaranteed monotonic for a room, pass it into the query, add a mapped `roomNameUpdatedAt` field, and guard both assignments in the script.

<details>

- **`distributed-system`**:138: RoomRenamedInboxPayload carries a timestamp, but BuildByQuery discards it and the script assigns roomName unconditionally. A concrete failure occurs when rename R1 is NAKed after a timeout or partial execution, newer R2 succeeds, and R1 is then redelivered: R1 runs last, is acknowledged, and restores the old name indefinitely. Store an authoritative monotonic room revision on spotlight documents and update only when the incoming revision is newer; member-document writes must preserve the same revision contract. A wall-clock timestamp is safe only if the producer guarantees monotonicity.
- **`elasticsearch`**:158: The script assigns `roomName` unconditionally and `BuildByQuery` discards the payload timestamp. If a newer rename is applied before an older delayed or redelivered rename, the older event succeeds and permanently restores the stale name. Add an explicitly mapped monotonic rename version/timestamp to spotlight documents, pass it as a script parameter, and update both fields only when the incoming version is newer, handling existing documents where the version field is absent.
- **`tgd-review`**:138: The payload timestamp is decoded but discarded, and the generated script assigns roomName unconditionally. If rename events are delivered or retried out of order, an older event can overwrite the newer name permanently; the surrounding room-worker comments explicitly note that per-room ordering is not guaranteed. Add a mapped room-name version/timestamp to spotlight documents and make both member writes and this script update only when the incoming version is newer.

</details>

<details>

```
In `search-sync-worker/spotlight.go` around line 138: The payload timestamp is discarded and the Painless script unconditionally assigns `roomName`. Two rename commands can be published out of causal order by concurrent room workers, and a rename Naked after an Elasticsearch failure can also be redelivered after a newer rename was Acked. In either path, the older event executes last and permanently restores the old name. Persist an authoritative monotonic rename revision on spotlight documents and update only when the incoming revision is newer; if `Timestamp` is guaranteed monotonic for a room, pass it into the query, add a mapped `roomNameUpdatedAt` field, and guard both assignments in the script.

```

</details>

pkg/searchengine/adapter.go
```diff
@@ -242,6 +242,51 @@ func (a *httpAdapter) UpdateMapping(ctx context.Context, indexPattern string, bo
...
+		}
+		req.Header.Set("Content-Type", "application/json")
+		return req, nil
+	})
+	if err != nil {
+		return fmt.Errorf("update by query: %w", err)
+	}
+	defer resp.Body.Close()
+
+	respBody, readErr := io.ReadAll(resp.Body)
```
julianshen: _⚡ memory-exhaustion_ | _🟠 Warning_ | _`oom`_

**The update response is read and copied without a size limit.**

A room rename whose update produces a large Elasticsearch per-document `failures` response, or a large proxy/error response, makes `io.ReadAll` allocate until EOF. Successful JSON decoding then copies each failure into `json.RawMessage`, and the failure error formats the entire body again, multiplying peak memory and risking GC thrashing or process OOM. Bound the response before decoding; normal successful `_update_by_query` responses are small, while an oversized response can safely be treated as a retryable error.

<details>

```
In `pkg/searchengine/adapter.go` around line 265: A room rename whose update produces a large Elasticsearch per-document `failures` response, or a large proxy/error response, makes `io.ReadAll` allocate until EOF. Successful JSON decoding then copies each failure into `json.RawMessage`, and the failure error formats the entire body again, multiplying peak memory and risking GC thrashing or process OOM. Bound the response before decoding; normal successful `_update_by_query` responses are small, while an oversized response can safely be treated as a retryable error.

```

</details>

pkg/searchengine/searchengine.go
```diff
@@ -69,6 +69,14 @@ type SearchEngine interface {
 	// matching pattern; templates apply only at creation, existing indices need this.
 	UpdateMapping(ctx context.Context, indexPattern string, body json.RawMessage) error
 
+	// UpdateByQuery runs a painless `_update_by_query` against index with the
+	// given request body (query + script). It mutates every matching document,
+	// so callers use it for field-keyed multi-doc updates the DocID-keyed Bulk
+	// API can't express — e.g. re-indexing roomName on every member's spotlight
+	// doc after a room rename. Runs with conflicts=proceed so a concurrent
+	// version bump skips that one doc instead of aborting the whole batch.
+	UpdateByQuery(ctx context.Context, index string, body json.RawMessage) error
```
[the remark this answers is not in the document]
chenjr0719: Keeping the raw-HTTP form here on purpose. This adapter (`httpAdapter`) is a thin raw-HTTP wrapper — `UpdateMapping`, `PutScript`, `Bulk`, and `Search` all issue their request through the same `Transporter.Perform` path and parse the response themselves. `UpdateByQuery` follows that established shape, and the response fields we act on (`version_conflicts`, `timed_out`, `failures`) are already parsed. Switching one method to the typed client would pull in a second client instance and diverge from the rest of the adapter for no behavioral gain, so we're keeping it consistent.

search-sync-worker/spotlight.go
```diff
@@ -113,6 +113,66 @@ func (c *spotlightCollection) BuildAction(data []byte) ([]searchengine.BulkActio
...
+		return "", nil, false, fmt.Errorf("build spotlight rename: missing newName")
+	}
+	if p.Timestamp <= 0 {
+		return "", nil, false, fmt.Errorf("build spotlight rename: missing timestamp")
+	}
+	body, err := buildSpotlightRenameByQuery(p.RoomID, p.NewName, p.Timestamp)
+	if err != nil {
+		return "", nil, false, fmt.Errorf("build spotlight rename: %w", err)
+	}
+	return c.indexName, body, true, nil
```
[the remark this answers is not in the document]
chenjr0719: We chose an acceptance-time-millis LWW guard over a per-room monotonic revision deliberately. Spotlight is a derived typeahead cache (the room doc is source of truth), so the goal is convergence, not a strict total order: the `roomNameUpdatedAt` guard already makes an out-of-order rename a no-op and lets the newest win regardless of arrival order. The residual — a same-millisecond collision or cross-replica clock skew — is rare and non-data-loss (stale typeahead text only, self-corrected by the next rename/membership event), and it's documented in the code. A monotonic per-room rename revision would require the room record + producer to carry and increment a counter, a larger cross-service change that isn't proportionate for this cache. The new field is additive (backward-compatible via the index template), not a breaking schema change.

search-sync-worker/spotlight.go
```diff
@@ -113,6 +113,66 @@ func (c *spotlightCollection) BuildAction(data []byte) ([]searchengine.BulkActio
...
+// Residual (bounded, documented): a member added *concurrently* with a rename can
+// have its doc created (from the member event's name snapshot) only after the
+// rename's update-by-query already ran on the then-absent doc, so that one new
+// member can see the pre-rename name until the next event touches their doc.
+// Closing it fully needs query-time room-name resolution (a room doc + join),
+// out of scope for this typeahead cache — spotlight is derived, not source of truth.
+func buildSpotlightRenameByQuery(roomID, newName string, ts int64) (json.RawMessage, error) {
+	body, err := json.Marshal(map[string]any{
+		"query": map[string]any{"term": map[string]any{"roomId": roomID}},
+		"script": map[string]any{
```
julianshen, line 163: _🏷️ event-ordering_ | _🔴 Blocking_ | _`distributed-system`_

**A rename is lost when it arrives before the member document.**

Even with a local bulk barrier, the room-worker comments acknowledge that concurrent room messages can be enqueued out of causal order. If room_renamed reaches this term query before a delayed member_added event, Elasticsearch returns success with zero matches and the handler acknowledges the rename. The later member event then creates a document containing its stale name snapshot, with no event left to repair it. Establish per-room ordering at production, resolve the current canonical room name/version when creating member documents, or add durable reconciliation for documents created after a rename.

<details>

```
In `search-sync-worker/spotlight.go` around line 163: Even with a local bulk barrier, the room-worker comments acknowledge that concurrent room messages can be enqueued out of causal order. If room_renamed reaches this term query before a delayed member_added event, Elasticsearch returns success with zero matches and the handler acknowledges the rename. The later member event then creates a document containing its stale name snapshot, with no event left to repair it. Establish per-room ordering at production, resolve the current canonical room name/version when creating member documents, or add durable reconciliation for documents created after a rename.

```

</details>

search-sync-worker/spotlight.go
```diff
@@ -113,6 +113,66 @@ func (c *spotlightCollection) BuildAction(data []byte) ([]searchengine.BulkActio
...
+			"lang": "painless",
+			"source": "long stored = ctx._source.roomNameUpdatedAt == null ? 0L : ((Number)ctx._source.roomNameUpdatedAt).longValue(); " +
+				"if (params.ts > stored) { ctx._source.roomName = params.name; ctx._source.roomNameUpdatedAt = params.ts; } else { ctx.op = 'noop'; }",
```
julianshen: _🏷️ time-ordering_ | _🔴 Blocking_ | _`distributed-system`_

**Millisecond timestamps do not define a total rename order.**

The script accepts only params.ts > stored, but RoomNameUpdatedAt is explicitly an event-millisecond value. Two distinct renames can therefore share a timestamp: the first stores its name and timestamp, the second correctly processed rename compares equal, becomes a noop, and is acknowledged while the canonical room has the second name. Changing the comparison to >= would make replay order determine the result. Use a monotonic per-room revision, or persist and compare a deterministic compound order such as timestamp plus a unique event identifier.

<details>

```
In `search-sync-worker/spotlight.go` around line 166: The script accepts only params.ts > stored, but RoomNameUpdatedAt is explicitly an event-millisecond value. Two distinct renames can therefore share a timestamp: the first stores its name and timestamp, the second correctly processed rename compares equal, becomes a noop, and is acknowledged while the canonical room has the second name. Changing the comparison to >= would make replay order determine the result. Use a monotonic per-room revision, or persist and compare a deterministic compound order such as timestamp plus a unique event identifier.

```

</details>

room-worker/handler.go
```diff
@@ -2342,6 +2342,30 @@ func (h *Handler) processRoomRename(ctx context.Context, data []byte) (err error
...
+	}
+	// Return on failure so the ROOMS message redelivers rather than silently
+	// losing the spotlight re-index (for a single-site room there is no OUTBOX
+	// federate to cover it). Safe to retry: the publish carries a requestID-keyed
+	// dedup id, the sys message a deterministic id, and UpdateRoomName is an
+	// idempotent set — a redelivery re-runs them all without duplicating.
+	if err := h.publish(ctx, subject.InboxInternal(h.siteID, model.InboxRoomRenamed), internalRenameData, natsutil.InboxDedupID(ctx, h.siteID, requestID)); err != nil {
+		return fmt.Errorf("publish local inbox room_renamed: %w", err)
+	}
 	// Relay room_renamed through the OUTBOX ordered lane (not a direct INBOX
```
julianshen, line 2370: _🏷️ partial-failure_ | _🔴 Blocking_ | _`distributed-system`_

**A failed publish can make a stale rename revert newer state.**

This new error return occurs after UpdateRoomName and UpdateSubscriptionNamesForRoom have succeeded. Rename R1 can set name A and then fail this publish; rename R2 can subsequently set name B; when R1 is redelivered, its unconditional setters run again and revert the canonical room and subscriptions to A. Broker request-ID deduplication does not fence these database writes, and setting a value is not order-safe under this interleaving. Persist the rename and an outbox record atomically, or guard database updates with a monotonic per-room revision/idempotency record so a retry resumes missing publication without reapplying stale state.

<details>

```
In `room-worker/handler.go` around line 2369: This new error return occurs after UpdateRoomName and UpdateSubscriptionNamesForRoom have succeeded. Rename R1 can set name A and then fail this publish; rename R2 can subsequently set name B; when R1 is redelivered, its unconditional setters run again and revert the canonical room and subscriptions to A. Broker request-ID deduplication does not fence these database writes, and setting a value is not order-safe under this interleaving. Persist the rename and an outbox record atomically, or guard database updates with a monotonic per-room revision/idempotency record so a retry resumes missing publication without reapplying stale state.

```

</details>

search-sync-worker/spotlight.go
```diff
@@ -113,6 +113,66 @@ func (c *spotlightCollection) BuildAction(data []byte) ([]searchengine.BulkActio
...
+// rename's update-by-query already ran on the then-absent doc, so that one new
+// member can see the pre-rename name until the next event touches their doc.
+// Closing it fully needs query-time room-name resolution (a room doc + join),
+// out of scope for this typeahead cache — spotlight is derived, not source of truth.
+func buildSpotlightRenameByQuery(roomID, newName string, ts int64) (json.RawMessage, error) {
+	body, err := json.Marshal(map[string]any{
+		"query": map[string]any{"term": map[string]any{"roomId": roomID}},
+		"script": map[string]any{
+			"lang": "painless",
+			"source": "long stored = ctx._source.roomNameUpdatedAt == null ? 0L : ((Number)ctx._source.roomNameUpdatedAt).longValue(); " +
```
julianshen, line 168: _🎯 correctness_ | _🔴 Blocking_ | _`elasticsearch`, `tgd-review` (2 rules)_

**Equal-millisecond renames cause the later name to be discarded.**

The guard accepts only params.ts > stored, but rename timestamps are millisecond integers and validation does not require uniqueness. If two successful renames for one room carry the same timestamp, the first update stores that timestamp and the second always no-ops, even when the canonical room state contains the second name. Use a strictly monotonic per-room revision or persist and compare an additional authoritative tie-breaker, and test two distinct renames with equal timestamps.

<details>

- **`tgd-review`**:166: The strict `params.ts > stored` comparison uses an event-millisecond clock without a tie-breaker. Two distinct renames in the same millisecond retain whichever name is delivered first, so different delivery orders can produce different final names despite the order-independent claim. Store and compare a deterministic tie-breaker such as a per-room sequence or request ID alongside the timestamp.

</details>

<details>

```
In `search-sync-worker/spotlight.go` around line 165: The guard accepts only params.ts > stored, but rename timestamps are millisecond integers and validation does not require uniqueness. If two successful renames for one room carry the same timestamp, the first update stores that timestamp and the second always no-ops, even when the canonical room state contains the second name. Use a strictly monotonic per-room revision or persist and compare an additional authoritative tie-breaker, and test two distinct renames with equal timestamps.

```

</details>

search-sync-worker/handler.go
```diff
@@ -90,6 +90,31 @@ func (h *Handler) AddWithContext(ctx context.Context, msg jetstream.Msg) {
...
 	}
+
+	// By-query events (e.g. room rename) mutate many docs by field, not by
+	// DocID, so they can't ride the bulk buffer — apply each as a standalone
+	// _update_by_query and ack/nak it on its own. A collection that doesn't
+	// implement ByQueryCollection, or a message it doesn't claim (ok=false),
+	// falls through to the normal bulk path.
+	if bq, isBQ := h.collection.(ByQueryCollection); isBQ {
+		index, body, ok, bqErr := bq.BuildByQuery(data)
+		if bqErr != nil {
```
julianshen, line 101: _🏷️ observability_ | _🟠 Warning_ | _`tgd-review`_

**By-query outcomes bypass disposition metrics.**

Every return in the new by-query branch acknowledges or negatively acknowledges the message without calling `h.metrics.recordMessages`. The surrounding poison and bulk paths record their dispositions, so room-renamed successes, validation failures, and retryable failures will be absent from the worker's message metrics. Record the corresponding disposition before each return.

<details>

```
In `search-sync-worker/handler.go` around line 101: Every return in the new by-query branch acknowledges or negatively acknowledges the message without calling `h.metrics.recordMessages`. The surrounding poison and bulk paths record their dispositions, so room-renamed successes, validation failures, and retryable failures will be absent from the worker's message metrics. Record the corresponding disposition before each return.

```

</details>

julianshen (commented):
tGD inline review

search-sync-worker/spotlight.go
chenjr0719: This is the concurrent-member-during-rename residual documented in the code comment on `buildSpotlightRenameByQuery`. Spotlight is a derived typeahead cache (the room doc is the source of truth), so a newly-added member briefly seeing the pre-rename name until the next event touches their doc is bounded stale text, not data loss. Fully closing it needs query-time room-name resolution (a room-name doc + a join at search time) — a larger change out of scope for this cache re-index. Kept as a documented limitation.

room-worker/handler.go
chenjr0719: Good catch — reverting this to best-effort (log, don't Nak) in the next push. You're right that Naking redelivers the whole handler and the already-committed UpdateRoomName / UpdateSubscriptionNamesForRoom are unconditional writes, so a redelivered stale rename could revert newer canonical state. The spotlight re-index is a derived cache, so a rare missed publish self-corrects on the next rename/membership event — far cheaper than risking source-of-truth reversion.

julianshen:
## Needs clarification

Is a fully drained consumer-first rollout guaranteed before room\_renamed is produced?
Reply in this thread, or post: `answer clar_yggqvz6voj3ef7kvgemoplng2p: <your answer>`

room-worker/handler.go
```diff
@@ -2342,6 +2342,31 @@ func (h *Handler) processRoomRename(ctx context.Context, data []byte) (err error
...
+	if err != nil {
+		return fmt.Errorf("marshal internal rename event: %w", err)
+	}
+	// Best-effort: log, don't Nak. Returning here would redeliver the whole
+	// handler, and UpdateRoomName / UpdateSubscriptionNamesForRoom (already
+	// committed above) are unconditional writes — a redelivered stale rename
+	// would revert newer canonical state. The spotlight re-index is a derived
+	// cache; a rare missed publish self-corrects on the next rename/membership
+	// event and is far cheaper than risking source-of-truth reversion.
+	if err := h.publish(ctx, subject.InboxInternal(h.siteID, model.InboxRoomRenamed), internalRenameData, natsutil.InboxDedupID(ctx, h.siteID, requestID)); err != nil {
```
julianshen, line 2367: _🏷️ partial-failure_ | _🔴 Blocking_ | _`distributed-system`, `tgd-review` (2 rules)_

**A failed local publish permanently loses the rename event.**

For a single-site room, the database updates can commit, this publish can fail before the broker accepts it, and the handler then logs the error and returns success because there are no remote sites to federate. The source message is consequently acknowledged while every existing spotlight document retains the old name indefinitely; a later membership event only touches its own member document and is not reconciliation for the room. Persist the local event in an outbox atomically with the rename, or add a monotonic/idempotency fence that makes retrying the whole handler safe.

<details>

- **`tgd-review`**:2367: Logging and continuing avoids replaying the already-committed canonical writes, but it also acknowledges the source event without ensuring the new search event exists. For a same-site room with no later rename, every existing spotlight document can remain stale indefinitely; a membership event only touches its own document. Persist a retryable outbox record independently of the canonical update, or otherwise make publication durable without replaying stale database writes.

</details>

<details>

```
In `room-worker/handler.go` around line 2367: For a single-site room, the database updates can commit, this publish can fail before the broker accepts it, and the handler then logs the error and returns success because there are no remote sites to federate. The source message is consequently acknowledged while every existing spotlight document retains the old name indefinitely; a later membership event only touches its own member document and is not reconciliation for the room. Persist the local event in an outbox atomically with the rename, or add a monotonic/idempotency fence that makes retrying the whole handler safe.

```

</details>

pkg/searchindex/spotlightdoc.go
```diff
@@ -12,6 +12,10 @@ type SpotlightDoc struct {
 	RoomType    string    `json:"roomType"    es:"keyword"`
 	SiteID      string    `json:"siteId"      es:"keyword"`
 	JoinedAt    time.Time `json:"joinedAt"    es:"date"`
+	// RoomNameUpdatedAt is the event-millis under which RoomName is valid — the
+	// LWW clock a room_renamed update-by-query guards on so an out-of-order rename
+	// can't overwrite a newer name. Stamped = the member/rename event timestamp.
+	RoomNameUpdatedAt int64 `json:"roomNameUpdatedAt" es:"long"`
```
julianshen, line 18: _🎯 elasticsearch-correctness_ | _🔴 Blocking_ | _`elasticsearch`_

**The constructor drops the supplied room-name clock.**

The caller now sets SpotlightFields.RoomNameUpdatedAt, but the unchanged NewSpotlightDoc initializer does not copy this newly introduced field into SpotlightDoc. Newly indexed documents therefore store the zero value, allowing any delayed rename with a positive timestamp to overwrite a newer room-name snapshot. Add RoomNameUpdatedAt: f.RoomNameUpdatedAt to the SpotlightDoc initializer and test the serialized document value.

<details>

```
In `pkg/searchindex/spotlightdoc.go` around line 18: The caller now sets SpotlightFields.RoomNameUpdatedAt, but the unchanged NewSpotlightDoc initializer does not copy this newly introduced field into SpotlightDoc. Newly indexed documents therefore store the zero value, allowing any delayed rename with a positive timestamp to overwrite a newer room-name snapshot. Add RoomNameUpdatedAt: f.RoomNameUpdatedAt to the SpotlightDoc initializer and test the serialized document value.

```

</details>

julianshen (commented):
tGD inline review

pkg/searchindex/spotlightdoc.go
chenjr0719: `NewSpotlightDoc` uses `SpotlightDoc(f)` — a Go struct conversion between two structs with identical fields (conversion ignores tags), so it copies **every** field by position, including `RoomNameUpdatedAt`. A field mismatch would be a compile error, and the build passes. I added a test asserting `NewSpotlightDoc` carries `RoomNameUpdatedAt` (and `Origin`) to pin it against a future field-order drift. New documents therefore store the supplied clock, not zero.

room-worker/handler.go
chenjr0719: This is in tension with the earlier finding on this same line, where Naking here was flagged for letting a redelivered stale rename revert canonical state via the unconditional `UpdateRoomName`/`UpdateSubscriptionNamesForRoom`. Nak risks source-of-truth reversion; log-and-continue risks a missed re-index. I chose best-effort deliberately: the spotlight re-index is a **derived typeahead cache**, so a rare missed publish is bounded staleness (the room shows the pre-rename name until the next rename/membership event re-indexes it — no data loss), whereas Naking risks corrupting the canonical room/subscription state. Closing the miss without the reversion risk would require guarding the canonical writes with a rename version — a larger cross-service change out of scope for this cache re-index.

julianshen (commented):
tGD inline review

pkg/subject/subject.go
```diff
@@ -309,18 +309,22 @@ func InboxExternalAll(siteID string) string {
...
+// room name across every member's doc). Use with
 // jetstream.ConsumerConfig.FilterSubjects (NATS 2.10+).
 func InboxMemberEventSubjects(siteID string) []string {
 	return []string{
 		InboxInternal(siteID, "member_added"),
 		InboxInternal(siteID, "member_removed"),
 		InboxInternal(siteID, "member_joinedat_refreshed"),
+		InboxInternal(siteID, "room_renamed"),
 		InboxExternal(siteID, "member_added"),
 		InboxExternal(siteID, "member_removed"),
```
julianshen, line 325: _🏷️ rolling-deployment_ | _🔴 Blocking_ | _`elasticsearch`_

**Old workers can acknowledge rename events as poison.**

This adds room_renamed to the existing durable consumer filters without versioning the consumer or staging compatibility. During a rolling deployment, an old spotlight worker attached to that durable can receive the newly enabled event; it lacks ByQueryCollection handling, falls through to the member-event parser, and acknowledges the resulting validation error as poison. That permanently loses the rename. Use a staged rollout where all workers first understand the event before enabling the filter, or introduce a versioned durable consumer for this event.

<details>

```
In `pkg/subject/subject.go` around line 325: This adds room_renamed to the existing durable consumer filters without versioning the consumer or staging compatibility. During a rolling deployment, an old spotlight worker attached to that durable can receive the newly enabled event; it lacks ByQueryCollection handling, falls through to the member-event parser, and acknowledges the resulting validation error as poison. That permanently loses the rename. Use a staged rollout where all workers first understand the event before enabling the filter, or introduce a versioned durable consumer for this event.

```

</details>

mliu33 (approved):
Thanks!

pkg/subject/subject.go
chenjr0719: Resolved as an accepted, bounded residual (merged as-is by the maintainer). Spotlight is a derived typeahead cache, so a room_renamed dropped by an old worker during the rolling-deploy window just leaves those docs on the pre-rename name until the next rename/membership event re-indexes them — no source-of-truth loss. It's handled operationally with a **consumer-first / drained rollout** (new by-query-capable workers up before room_renamed is produced), which also answers the earlier consumer-first-rollout clarification. If we later want it airtight regardless of rollout order, the follow-up is to isolate room_renamed on its own durable so old workers on the member consumer never receive it.

PR #319, opened 2026-08-19
fix(natsmetrics): drop inline service_name label that 500s /metrics

flywindy:
## Problem

Every `chat_nats_*` metric family disappeared: the Prometheus `/metrics` endpoint returned HTTP 500 with a duplicate-label error.

The o11y SDK's Prometheus exporter renders the resource's `service.name` (from `OTEL_SERVICE_NAME`) as a constant `service_name` label on every series (`WithResourceAsConstantLabels`). `pkg/natsmetrics` also attached its own inline `service_name` attribute to every consumer/publisher instrument. `client_golang` rejects a metric whose constant and variable label names collide, so `Gather()` failed with `duplicate label names in constant and variable labels` on all 8 recorded families and promhttp turned the entire scrape into a 500.

`site` was verified NOT to collide: it is not in the exporter's constant-label allow list, and deployments put `site.id` (→ `site_id`) into the resource, not `site`.

## Fix

- Remove the inline `service_name` attribute and its API surface: `ConsumerConfig.ServiceName` and `Publisher`'s `serviceName` parameter (`pkg/natsmetrics`). The `site` label stays.
- Update all seven adopter services (broadcast-worker, message-worker, notification-worker, message-gatekeeper, room-service, room-worker, history-service) and remove their now-dead `OTEL_SERVICE_NAME` config fields plus the config tests that only covered them. `pkg/obs` still reads `OTEL_SERVICE_NAME` for the resource, so deployments are unchanged.
- Update `docs/specs/o11y/o11y-metrics-inventory.md` §2.1: the shared label base is now `site` only; service identity is the resource-derived constant label.
- `go.mod`: `go.opentelemetry.io/otel/exporters/prometheus` moves from indirect to direct (test-only import; already in the module graph).

Queries are unaffected: the constant label supplies the same `service_name` name and value (both sides read `OTEL_SERVICE_NAME`), so existing `{service_name="…"}` selectors and Grafana trace↔metric correlation keep working. Bot/Teams deployments keep distinct identities the same way (`room-worker`'s compose-distinctness test is retained).

## Testing

- New `pkg/natsmetrics/prometheus_export_test.go` mirrors the SDK's exporter setup (`WithResourceAsConstantLabels` + populated resource). Written TDD-first: it reproduced the 500 (all 8 families failing gather) before the fix, and now pins that gather succeeds, every label appears exactly once per series, and `service_name` arrives from the resource.
- `make lint`: 0 issues. `make test`: all 112 packages pass with `-race`.
- `gosec` passes; govulncheck/semgrep could not run in this sandbox (network/tooling) and are covered by the CI `sast` gate.

https://claude.ai/code/session_01NHmKan8izZjruS9x7U15n1

---
_Generated by [Claude Code](https://claude.ai/code/session_01NHmKan8izZjruS9x7U15n1)_

## Summary by CodeRabbit

* **Metrics**
  * Standardized metrics identity around site information.
  * Service names are now exported consistently as resource-derived labels rather than inline metric attributes.
  * Improved Prometheus label handling to prevent duplicate labels.

* **Refactor**
  * Removed service-name configuration from supported services and metric setup.
  * Updated metrics interfaces and integrations to use site-only identification.

* **Documentation**
  * Clarified metric labels and how site and service identity are represented.


flywindy:
@coderabbitai review

flywindy:
@coderabbitai review

hmchangw (approved):

PR #320, opened 2026-08-19
feat(loadgen): make the subscription-list request shape configurable

flywindy:
`subscription_list` has p95 pinned at the 5s request timeout in staging, with
3–4 timeouts/s and the same again in retries. #317 confirmed it is not the
loadgen's memory: it persists with the pod stable at 8GB with `GOMEMLIMIT`
set. This does not fix it — it makes the next step a measurement instead of an
argument.

## What the lane was actually sending

```go
soakSubscriptionListRequest{Type: soakSubscriptionListType}
```

That is the whole body. Two things the lane never chose were therefore deciding
what it measured:

- **No `limit`.** user-service applies `SUBSCRIPTION_DEFAULT_LIMIT`, currently
  `40`. The lane was reporting on whatever page size the service happens to
  default to, not on a workload anybody picked.
- **No `includeLastMessage`.** user-service reads a missing value as `true`, so
  every request runs `enrichLastMessage`, which fans each listed room out to
  history-service `rooms.get` — for **all** sites including the local one.

`rooms.get` resolves each room's preview, and a room with no cached preview
walks Cassandra buckets to find one. The preview cache is positives-only, so an
empty room is never cached and re-walks on every call. The walk's floor is
`createdAt` clamped to `now - MESSAGE_HISTORY_FLOOR_DAYS` (730), and the history
seeder writes room `createdAt` as `time.Unix(0, 0)` — 1970 — so the clamp always
lands at the widest bound, up to `MESSAGE_READ_MAX_BUCKETS` (122) per room.

Forty rooms behind one 5s deadline, each capable of a 122-bucket walk, with
`maxRoomsGetConcurrency = 16`. That is the shape of the problem; what this PR
adds is the ability to test it.

## What this changes

| | |
|---|---|
| `SOAK_SUBSCRIPTION_LIST_LIMIT` | Page size for the read. Unset leaves the service default in force. |
| `SOAK_SUBSCRIPTION_LIST_INCLUDE_LAST_MESSAGE` | Enrichment control. Unset sends nothing; `false` drops the per-room fan-out. |

Setting the second to `false` for one run separates *the page is expensive*
from *the enrichment is expensive* — without re-seeding anything, which is what
the `createdAt` fix would require.

## Both default to unset, deliberately

Every percentile recorded so far was measured against the old request. If an
unconfigured run sent anything different, none of that history would be
comparable. So both wire fields carry `omitempty` and both config values default
to zero/nil, and a test asserts the marshalled body is still exactly
`{"type":"rooms"}` when neither is set.

`omitempty` on `IncludeLastMessage` is load-bearing rather than cosmetic:
sending `false` is a different request from omitting the field, and only the
omission means "let the service decide". That is why it is `*bool` and not
`bool`.

## Two things worth reviewing

**The mapping is a named function.** `soakRoomReadConfigFrom` replaces a struct
literal at the call site so a test can assert the values actually arrive. A knob
that parses but never reaches the lane leaves a run looking configured while it
measures the old thing — a failure mode this branch has already shipped once.

**The enrichment guard is not vacuous.** Turning enrichment off must not be able
to manufacture a finding. Today it cannot: the lane samples the row count, and
the room-state observer reads a different RPC entirely
(`subscription.getByRoomID`). `TestSoakRoomReader_SubscriptionListSamplesTheSameWithoutEnrichment`
pins that. I verified it can fail by making the lane count only rows carrying a
last message — it fails on both assertions with that mutation and passes once
reverted.

## Docs

`docs/load-testing/loadgen/observation.md`'s config table gains the two new
knobs and the six #317 shipped without listing: the journal's
compact/size/expire-batch bounds and the profiling controls. That omission has a
cost — `SOAK_LEDGER_MAX_BYTES` sat at its 512MB default through an OOM
investigation whose journal peaked around 400MB and so never reached the
trigger.

`docs/client-api.md` needs no change: it already documents
`includeLastMessage: false` on the `previewMessage` field, and nothing here
touches a client-facing handler or a `pkg/model` struct.

## What this does not do

- It does not fix `subscription_list`. It makes the next run answer whether the
  enrichment path is the cost.
- It does not touch the seeder `createdAt`. That needs a re-seed and should be
  gated on what the probe returns.
- It does not address the hourly OOM still seen after #317, which is being
  investigated from heap profiles separately.

## Verification

`make lint` clean (0 issues), full `tools/loadgen` unit suite green under
`-race` (268s), `go vet -tags integration` compiles. `helm` is not available in
this environment, so the chart changes are unvalidated beyond the schema being
well-formed JSON.

---

## Summary by CodeRabbit

* **New Features**
  * Added optional subscription-list page limits for soak tests.
  * Added controls to include or exclude the last message in subscription-list results.
  * Added validation for subscription-list settings and preserved existing defaults when unset.
  * Added deployment configuration options and documentation for these diagnostics.

* **Tests**
  * Added coverage for configuration parsing, validation, request behavior, and enriched results.


flywindy:
Not acting on the docstring-coverage warning (71.43% vs an 80% threshold). It is measuring the touched files, not this diff, and the two are very different numbers.

Every symbol this PR **adds** carries a doc comment: `soakRoomReadConfigFrom`, both new `soakConfig` fields, both new `soakSubscriptionListRequest` fields, both new `soakRoomReadConfig` fields, and the test fixture. The only two additions without one are `TestSoakConfig_SubscriptionListProbeReadsItsEnvironment` and `TestSoakRoomReadConfigFrom_LeavesAnUnsetProbeUnset`, whose names state the assertion and which sit immediately under a commented sibling that supplies the shared rationale.

What drags the percentage down is what was already there. `soak_wire.go` alone contributes ~55 undocumented declarations — the request/response carriers, which are covered by a group comment explaining that they mirror history-service's internal models and are protected by contract tests, rather than by 55 restatements of their own field names. `soak_roomread.go` and `soak_main.go` are the same shape.

Writing docstrings to move that number would work against the house style rather than toward it: this repo carries a `remove_comments` skill whose whole purpose is deleting comments that restate *what* the code does, and CLAUDE.md asks for comments that explain *why*. `// soakRoomInfo is a room info.` satisfies a linter and costs the next reader time.

The repo's blocking gates are `lint`, `test`, `test-integration` and `sast`, all of which this branch passes.

---

julianshen:
**No actionable comments.** ✅

### Related work

- [PR #317](https://github.com/hmchangw/newchat/pull/317) — fix\(loadgen\): stop the soak OOM loop and make its failures readable (merged)

<details>
<summary>📒 Files reviewed (9)</summary>

* `tools/loadgen/soak_subscription_probe_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

yenta (approved):

PR #324, opened 2026-08-20
feat: badge cache precise read invalidation with bounded-staleness marker

hmchangw:
## Summary

Eliminate the MongoDB `$lookup` aggregation on every room read by making the badge cache invalidation precise and bounding staleness via an independent marker TTL. Room reads now remove exactly the room (`ClearRoom`) instead of dropping the whole set (`ClearAll`), and the freshness marker gets its own shorter countdown that is stamped only by Mongo-verifying paths (Seed/Reseed), never refreshed by bumps.

### `pkg/badgecache` — Marker TTL and Script Semantics
- Add `Option` functional type and `WithMarkerTTL(d time.Duration)` to give the freshness marker its own lifetime, shorter than the set's TTL
- Default behavior unchanged: absent option means `markerTTL == ttl` (backward compatible with existing call sites in room-service and inbox-worker)
- Rewrite `bumpScript`: marker existence is now the hit condition (not set existence); bumps no longer re-stamp the marker, so its countdown continues from the last Mongo verification
- Rewrite `seedScript`: replace the set instead of unioning into it, so an unverified leftover cannot contaminate a freshly computed answer
- Add jitter to marker TTL to spread expiry across accounts and avoid thundering herd on reseed
- Update `Reseed` and `Seed` to pass both `setTTL` and `markerTTL` to scripts
- Add integration tests: marker TTL independence, bump-miss-on-expired-marker, seed replace semantics, marker-only state handling

### `room-service` — Precise `ClearRoom` on Room Read
- Change `RoomStore.UpdateSubscriptionRead` signature to return `(threadUnread int, err error)` — the post-update count of unread followed threads
- Update `store_mongo.go`: switch from `UpdateOne` to `FindOneAndUpdate` with `SetProjection({threadUnread: 1})` and `ReturnDocument(After)` to get the post-write value in a single round trip
- Update `handler.go:messageRead`: replace `ClearAll` with conditional `ClearRoom` — only clear the room if `threadUnread == 0` (fully read), otherwise no cache write (room was and remains unread)
- Add unit tests: no-thread-unread clears room only, thread-unread remains means no badge write, post-update value is used (not pre-write snapshot)

### `inbox-worker` — Matching Shape for Consistency
- Update `InboxStore.UpdateSubscriptionRead` signature to return `(applied bool, threadUnread int, err error)` for consistency with room-service
- Update `main.go` implementation to use `FindOneAndUpdate` and return the post-update `threadUnread` count
- Update `handleSubscriptionRead` to gate badge operations on the `applied` flag (skip if the write was out-of-order)
- Update integration tests to assert the new return values

### `user-service` — Config, Wiring, and Degraded Handling
- Add `BadgeMarkerTTL` config field with default `10m` and validation that it is positive and ≤ `BadgeCacheTTL`
- Pass `WithMarkerTTL(cfg.BadgeMarkerTTL)` when constructing the badge cache in `main.go`
- Update `unreadRooms` to return a `degraded` flag when a cross-site RPC fails (partial set is unreliable)
- Update `CountSubscriptions` and `BadgeCountBatch` to skip `Reseed` when degraded, so a partial set does not get stamped as fresh
- Add config tests for marker TTL defaulting and validation

### Documentation
- Add spec: `docs/superpowers/specs/2026-08-19-badge-precise-read-invalidation-design.md` — full design rationale, invariants, and accuracy model
- Add implementation plan: `docs/superpowers/plans/2026-08-19-badge-precise-read-invalidation.md` — task-by-task checklist with TDD guidance
- Update `docs/notification-worker-downstream-contracts.md`

https://claude.ai/code/session_019unBWDkdqrQWtWxUvmJF4v

## Summary by CodeRabbit

* **Improvements**
  * Room reads now clear only rooms with no remaining unread followed threads, preserving accurate badges.
  * Badge freshness is bounded by a configurable marker lifetime, preventing stale counts from persisting indefinitely.
  * Incomplete cross-site results still return best-effort counts but are not saved as verified cache data.
  * Badge counts now align more closely with counts delivered through notifications.

* **Documentation**
  * Updated API and downstream contract documentation to describe badge accuracy, invalidation, degradation, and freshness behavior.


user-service/service/subscriptions.go
```diff
@@ -562,28 +562,35 @@ func (s *UserService) CountSubscriptions(c *natsrouter.Context, req models.Count
 			return &models.CountResponse{Count: n}, nil
 		}
 	}
-	ids, err := s.unreadRooms(c, account)
+	ids, degraded, err := s.unreadRooms(c, account)
 	if err != nil {
 		return nil, err
 	}
```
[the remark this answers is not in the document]
hmchangw: Not taking this one, and the related `inbox-worker/main.go:450` nitpick either — both would stutter, because the callee already wraps.

`unreadRooms` returns `fmt.Errorf("unread rooms: %w", err)` (`user-service/service/subscriptions.go:589`). Adding `fmt.Errorf("compute unread rooms: %w", err)` at this call site yields:

```
compute unread rooms: unread rooms: get active subscriptions: <cause>
```

Same for the nitpick: `naksIfSubscriptionMissing` (`inbox-worker/main.go:128`) wraps both of its paths — `check subscription exists for %q in room %q: %w` and `subscription not found for %q in room %q` — so wrapping its return again adds a second layer describing the same step.

CLAUDE.md's rule is to "describe what the current function was doing, not what failed underneath." The callee has already done exactly that, and the guidance against a bare `err` is aimed at errors that reach a boundary with no context at all. Neither of these does.

Worth noting both lines are also pre-existing and untouched by this PR — the diff only changed the arity of the call above them.

The other four findings are fixed in `67ee3f1`. The marker-TTL one was real and my own doc comment contradicted the code: it claimed "Never returns 0" while returning exactly that for a sub-second TTL. Reproduced it as a failing test first, then floored it in the package rather than relying on user-service's config validation, since `pkg/badgecache` is shared and can't assume every caller validates the same way.

I skipped the third site in that finding — the plan document under `docs/superpowers/plans/`. It's a point-in-time planning artifact, and the validation it describes already exists in `config.go`; the spec and the code are the sources of truth.

---

julianshen:

<details>

* `docs/client-api.md`
* `docs/superpowers/plans/2026-08-19-badge-precise-read-invalidation.md`
* `docs/superpowers/specs/2026-08-10-badge-cache-invalidation-design.md`
* `docs/superpowers/specs/2026-08-19-badge-precise-read-invalidation-design.md`
* `inbox-worker/handler.go`
* `inbox-worker/main.go`
* `pkg/badgecache/badgecache_integration_test.go`
* `pkg/badgecache/badgecache_test.go`
* `room-service/handler.go`
* `room-service/mock_store_test.go`
* `room-service/store.go`
* `user-service/main.go`
* `user-service/service/badge.go`
* `user-service/service/badge_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

room-service/handler.go
```diff
@@ -1384,11 +1387,12 @@ func (h *Handler) messageRead(c *natsrouter.Context) (*model.StatusReply, error)
...
-	// A read shrinks the unread set — drop it and recompute on next count.
-	// Home-local only; inbox-worker clears cross-site replicas when the
-	// federated subscription_read lands.
-	if userSiteID == h.siteID && h.badge != nil {
-		h.badge.ClearAll(ctx, account)
+	// A read settles message-unread (lastSeenAt = now), so the room stays unread
+	// only if a followed thread is unread. None ⇒ exact removal, freshness marker
+	// survives. Some ⇒ the room was and remains unread, so the set needs no edit.
+	// Home-local only; inbox-worker handles cross-site replicas.
+	if userSiteID == h.siteID && h.badge != nil && threadUnread == 0 {
```
julianshen, line 1394: _🏷️ rolling-deploy_ | _🔴 Blocking_ | _`distributed-system`_

**Precise clears are unsafe while any old user-service instance remains.**

A reverse-order rollout or user-service-only rollback creates an unbounded stale state. A reply bump can land before this new ClearRoom, then the clear removes the room; an old user-service subsequently refreshes the freshness marker on every bump, so marker expiry never forces reconciliation. The user can therefore miss the unread badge indefinitely until an unrelated ClearAll. The documentation's manual deployment order does not protect independent rollback or accidental mixed versions. Keep ClearAll as the default behind a capability/feature flag, enable precise clears only after all user-service instances are upgraded and legacy markers are purged, and disable it before rolling user-service back.

<details>

```
In `room-service/handler.go` around line 1394: A reverse-order rollout or user-service-only rollback creates an unbounded stale state. A reply bump can land before this new ClearRoom, then the clear removes the room; an old user-service subsequently refreshes the freshness marker on every bump, so marker expiry never forces reconciliation. The user can therefore miss the unread badge indefinitely until an unrelated ClearAll. The documentation's manual deployment order does not protect independent rollback or accidental mixed versions. Keep ClearAll as the default behind a capability/feature flag, enable precise clears only after all user-service instances are upgraded and legacy markers are purged, and disable it before rolling user-service back.

```

</details>

user-service/service/badge.go
```diff
@@ -26,11 +26,17 @@ func (s *UserService) BadgeCountBatch(c *natsrouter.Context, req model.BadgeCoun
...
 			continue
 		}
-		ids, err := s.unreadRooms(c, account)
+		ids, degraded, err := s.unreadRooms(c, account)
 		if err != nil {
 			slog.WarnContext(c, "badge seed degraded", "account", account, "room_id", req.RoomID, "request_id", natsutil.RequestIDFromContext(c), "error", err)
 			continue
 		}
+		// A partial result must not be cached (it would stamp the freshness
+		// marker); answer from it directly instead.
```
julianshen, line 35: _🏷️ backpressure_ | _🟠 Warning_ | _`distributed-system`_

**A remote-site outage causes unbounded badge recompute retries.**

When a marker expires, every message causes BumpBatch to miss and this path calls unreadRooms. If one remote site is unavailable, degraded remains true and no marker or other retry state is written. The next message therefore repeats the Mongo aggregation and failed cross-site RPC for every affected account. BadgeCountBatch processes accounts sequentially, so one sustained site outage can turn each full-audience message into many timeout-length RPCs, amplify load against the failed site, and cause notification requests to time out repeatedly. Add bounded per-site circuit breaking/backoff plus per-account singleflight or a short-lived retry-cooldown key distinct from the freshness marker, while continuing to label the result as degraded.

<details>

```
In `user-service/service/badge.go` around line 35: When a marker expires, every message causes BumpBatch to miss and this path calls unreadRooms. If one remote site is unavailable, degraded remains true and no marker or other retry state is written. The next message therefore repeats the Mongo aggregation and failed cross-site RPC for every affected account. BadgeCountBatch processes accounts sequentially, so one sustained site outage can turn each full-audience message into many timeout-length RPCs, amplify load against the failed site, and cause notification requests to time out repeatedly. Add bounded per-site circuit breaking/backoff plus per-account singleflight or a short-lived retry-cooldown key distinct from the freshness marker, while continuing to label the result as degraded.

```

</details>

julianshen (commented):
tGD inline review

user-service/config/config.go
```diff
@@ -106,8 +116,14 @@ func Load() (Config, error) {
...
 	}
-	if cfg.BadgeCacheTTL <= 0 {
-		return Config{}, fmt.Errorf("BADGE_CACHE_TTL must be > 0, got %s", cfg.BadgeCacheTTL)
+	if cfg.BadgeCacheTTL < time.Second {
+		return Config{}, fmt.Errorf("BADGE_CACHE_TTL must be >= 1s, got %s", cfg.BadgeCacheTTL)
+	}
+	if cfg.BadgeMarkerTTL < time.Second {
+		return Config{}, fmt.Errorf("BADGE_MARKER_TTL must be >= 1s, got %s", cfg.BadgeMarkerTTL)
+	}
+	if cfg.BadgeMarkerTTL > cfg.BadgeCacheTTL {
```
julianshen, line 125: _🏷️ rolling-deploy_ | _🟠 Warning_ | _`distributed-system`_

**The new marker default rejects previously valid cache TTLs.**

An existing deployment with BADGE_CACHE_TTL=5m and no new BADGE_MARKER_TTL was valid before this change. During rollout, the new binary applies the 10m marker default, fails this validation, and crash-loops, potentially stalling the rollout or reducing service capacity. Preserve compatibility by deriving an unset marker TTL as min(10m, BadgeCacheTTL), while still rejecting an explicitly configured marker TTL that exceeds the cache TTL.

<details>

```
In `user-service/config/config.go` around line 125: An existing deployment with BADGE_CACHE_TTL=5m and no new BADGE_MARKER_TTL was valid before this change. During rollout, the new binary applies the 10m marker default, fails this validation, and crash-loops, potentially stalling the rollout or reducing service capacity. Preserve compatibility by deriving an unset marker TTL as min(10m, BadgeCacheTTL), while still rejecting an explicitly configured marker TTL that exceeds the cache TTL.

```

</details>

pkg/badgecache/badgecache.go
```diff
@@ -16,59 +17,97 @@ import (
...
-// marker. KEYS=[set, marker], ARGV=[ttlSec, roomIDs...]; returns SCARD.
+// seedScript replaces the set from Mongo-derived IDs and stamps the marker.
+// Replace (not union): seed runs on a miss, where any surviving set is
+// unverified and must not contaminate a freshly computed answer.
+// KEYS=[set, marker], ARGV=[setTtlSec, markerTtlSec, roomIDs...]; returns SCARD.
 var seedScript = redis.NewScript(`
-if #ARGV > 1 then
-  redis.call('SADD', KEYS[1], unpack(ARGV, 2))
+redis.call('DEL', KEYS[1])
+if #ARGV > 2 then
```
julianshen, line 46: _🎯 concurrency_ | _🟠 Warning_ | _`distributed-system`_

**Late seed replacements can erase successful concurrent bumps.**

A caller can miss the marker and begin a slow Mongo/RPC recompute; another caller then seeds the account, and a later message successfully bumps its room into the now-fresh set. When the first caller finally executes Seed, this unconditional DEL replaces the set from its older snapshot, erases the acknowledged bump, and stamps a fresh marker, hiding the loss until the next marker expiry. Reseed has the same stale-writer race. Fence miss-driven writes: atomically check that the marker generation observed before recomputation is still absent/unchanged; if another caller has already established freshness, preserve that set and merge only the triggering room without extending the marker. Apply equivalent fencing to Reseed or serialize and recheck per account.

<details>

```
In `pkg/badgecache/badgecache.go` around line 46: A caller can miss the marker and begin a slow Mongo/RPC recompute; another caller then seeds the account, and a later message successfully bumps its room into the now-fresh set. When the first caller finally executes Seed, this unconditional DEL replaces the set from its older snapshot, erases the acknowledged bump, and stamps a fresh marker, hiding the loss until the next marker expiry. Reseed has the same stale-writer race. Fence miss-driven writes: atomically check that the marker generation observed before recomputation is still absent/unchanged; if another caller has already established freshness, preserve that set and merge only the triggering room without extending the marker. Apply equivalent fencing to Reseed or serialize and recheck per account.

```

</details>

pkg/badgecache/badgecache.go
```diff
@@ -16,59 +17,97 @@ import (
...
+  redis.call('SADD', KEYS[1], unpack(ARGV, 3))
   redis.call('EXPIRE', KEYS[1], ARGV[1])
 end
-redis.call('SET', KEYS[2], '1', 'EX', ARGV[1])
+redis.call('SET', KEYS[2], '1', 'EX', ARGV[2])
 return redis.call('SCARD', KEYS[1])
```
julianshen, line 51: _🎯 NATS concurrency_ | _🟠 Warning_ | _`mongodb`, `nats`, `performance-and-concurrency` (3 rules)_

**Concurrent marker-miss seeds overwrite each other's trigger rooms.**

Two concurrent badge.count.batch requests for different rooms can both miss an expired marker and compute before Mongo's room metadata reflects either message. Each Seed then executes DEL followed by replacement; the last request leaves only its own trigger room, losing the other request's trigger even though both NATS requests report success. The new marker prevents another recomputation until expiry. Make Seed distinguish initialization from a concurrent post-initialization call atomically: only the first marker-miss winner should replace the stale set and stamp the marker; if the marker already exists when a later Seed reaches Lua, it should SADD only that request's trigger room without refreshing the marker. A distributed per-account recomputation lock is another valid fix; process-local singleflight alone would not cover multiple user-service responders.

<details>

- **`mongodb`**:50: The new `seedScript` unconditionally deletes and replaces the set after a Mongo recomputation. When the marker expires, two messages for different rooms can both miss, query Mongo before `lastMsgAt` has caught up, and then seed `ids ∪ triggerRoomID`; whichever script runs last deletes the other trigger room while stamping a fresh marker. The badge then undercounts until marker expiry. Make replacement conditional on the marker still being absent; if another computation has already stamped it, atomically add only this request's trigger room. The trigger must therefore be passed separately from the Mongo-derived IDs. This CAS must happen in Valkey, not process-local singleflight, because concurrent requests can reach different user-service replicas.
- **`performance-and-concurrency`**:48: When a marker expires, concurrent messages for the same account all miss and independently run the expensive Mongo/RPC recomputation. If rooms A and B trigger two Seeds, B can install ids plus B, then A's unconditional DEL replaces it with ids plus A. Because room metadata is coalesced, neither Mongo snapshot is guaranteed to contain the other trigger; B is therefore lost while A stamps a fresh marker, producing an undercount for the full marker window. Add distributed per-account refresh coalescing, or make Seed atomically elect one replacement winner and have later callers only SADD their trigger room when the marker has already been restored.

</details>

<details>

```
In `pkg/badgecache/badgecache.go` around line 51: Two concurrent badge.count.batch requests for different rooms can both miss an expired marker and compute before Mongo's room metadata reflects either message. Each Seed then executes DEL followed by replacement; the last request leaves only its own trigger room, losing the other request's trigger even though both NATS requests report success. The new marker prevents another recomputation until expiry. Make Seed distinguish initialization from a concurrent post-initialization call atomically: only the first marker-miss winner should replace the stale set and stamp the marker; if the marker already exists when a later Seed reaches Lua, it should SADD only that request's trigger room without refreshing the marker. A distributed per-account recomputation lock is another valid fix; process-local singleflight alone would not cover multiple user-service responders.

```

</details>

user-service/service/subscriptions.go
```diff
@@ -562,28 +562,35 @@ func (s *UserService) CountSubscriptions(c *natsrouter.Context, req models.Count
...
-	// Best-effort reconciliation from the Mongo source of truth (fail-open).
-	s.badge.Reseed(c, account, ids)
+	// Best-effort reconciliation from the Mongo source of truth (fail-open) —
+	// skipped when degraded, since caching a partial set would stamp the
+	// freshness marker on data we already know is incomplete.
+	if !degraded {
+		s.badge.Reseed(c, account, ids)
```
julianshen, line 573: _🏷️ consistency_ | _🟠 Warning_ | _`mongodb`_

**A secondary read can stamp stale Mongo data as freshly verified.**

The shipped configuration uses `MONGO_READ_PREFERENCE=secondaryPreferred`, but this path now calls `Reseed` and stamps the freshness marker whenever cross-site RPCs did not fail. Replica lag is not represented by `degraded`: for example, after membership invalidation, an immediate count can read a lagging secondary that does not yet contain the new subscription, reseed an empty set, and report a fresh zero for the marker window. Replica lag can also exceed the marker TTL, so the documented TTL is not a strict staleness bound. Run this reconciliation aggregation against the primary, preferably with majority read concern if rollback durability is part of the contract, while retaining secondary reads for paths that permit stale results.

<details>

```
In `user-service/service/subscriptions.go` around line 573: The shipped configuration uses `MONGO_READ_PREFERENCE=secondaryPreferred`, but this path now calls `Reseed` and stamps the freshness marker whenever cross-site RPCs did not fail. Replica lag is not represented by `degraded`: for example, after membership invalidation, an immediate count can read a lagging secondary that does not yet contain the new subscription, reseed an empty set, and report a fresh zero for the marker window. Replica lag can also exceed the marker TTL, so the documented TTL is not a strict staleness bound. Run this reconciliation aggregation against the primary, preferably with majority read concern if rollback durability is part of the contract, while retaining secondary reads for paths that permit stale results.

```

</details>

inbox-worker/main.go
```diff
@@ -437,14 +437,22 @@ func (s *mongoInboxStore) UpdateSubscriptionRead(ctx context.Context, roomID, ac
...
-	res, err := s.subCol.UpdateOne(ctx, filter, update)
-	if err != nil {
-		return fmt.Errorf("update subscription read for %q in room %q: %w", account, roomID, err)
+	var out struct {
+		ThreadUnread []string `bson:"threadUnread"`
 	}
-	if res.MatchedCount == 0 {
-		return s.naksIfSubscriptionMissing(ctx, account, roomID)
+	opts := options.FindOneAndUpdate().
+		SetProjection(bson.D{{Key: "threadUnread", Value: 1}}).
```
julianshen, line 444: _🏷️ data-modeling_ | _🟠 Warning_ | _`mongodb`, `oom` (2 rules)_

**Room reads return the entire unbounded threadUnread array.**

Both new `FindOneAndUpdate` queries project all of `threadUnread`, although their callers only test whether its length is zero. An account can accumulate many unread followed-thread IDs, so every federated read now transfers and decodes that array, potentially approaching the subscription document's 16 MiB BSON limit; `room-service/store_mongo.go` repeats the same projection. Change the return contract to `hasThreadUnread bool` and project only one element with `$slice: 1`. This bounds the response while preserving the required post-update atomic snapshot.

<details>

- **`oom`**:444: Each applied `subscription_read` now fetches and decodes the complete `threadUnread` array. An account can accumulate many unread followed-thread parent IDs, weakly bounded only by MongoDB's roughly 16 MB document limit. A federated-read backlog, redeliveries, or concurrent workers can multiply that allocation per event and cause worker OOM or sustained GC pressure. Since the handler only checks `threadUnread == 0`, project at most one array element and return `hasThreadUnread`, or project a server-computed count instead of transferring the array.

</details>

<details>

```
In `inbox-worker/main.go` around line 444: Both new `FindOneAndUpdate` queries project all of `threadUnread`, although their callers only test whether its length is zero. An account can accumulate many unread followed-thread IDs, so every federated read now transfers and decodes that array, potentially approaching the subscription document's 16 MiB BSON limit; `room-service/store_mongo.go` repeats the same projection. Change the return contract to `hasThreadUnread bool` and project only one element with `$slice: 1`. This bounds the response while preserving the required post-update atomic snapshot.

```

</details>

room-service/handler.go
julianshen, line 1394: _🏷️ NATS ordering_ | _🟠 Warning_ | _`nats`_

**Independent consumers can make a read delete a still-unread room.**

A threaded reply is processed independently by notification-worker and message-worker from MESSAGES-CANONICAL. If notification-worker first bumps room R, the read then observes an empty threadUnread and executes ClearRoom, and message-worker subsequently adds the unread thread, no later event restores R because that reply's bump was already handled. The same interleaving exists for the federated inbox-worker path. The event is successfully acknowledged while the freshness marker remains valid, so cache-first counts under-report until marker expiry. The robust fix is to order the threadUnread mutation and badge mutation under one consumer/owner or carry a stream-sequence barrier into the read path; until that is available, precise SREM must not preserve a fresh marker when the thread state may still lag its corresponding message delivery.

<details>

```
In `room-service/handler.go` around line 1394: A threaded reply is processed independently by notification-worker and message-worker from MESSAGES-CANONICAL. If notification-worker first bumps room R, the read then observes an empty threadUnread and executes ClearRoom, and message-worker subsequently adds the unread thread, no later event restores R because that reply's bump was already handled. The same interleaving exists for the federated inbox-worker path. The event is successfully acknowledged while the freshness marker remains valid, so cache-first counts under-report until marker expiry. The robust fix is to order the threadUnread mutation and badge mutation under one consumer/owner or carry a stream-sequence barrier into the read path; until that is available, precise SREM must not preserve a fresh marker when the thread state may still lag its corresponding message delivery.

```

</details>

room-service/store_mongo.go
```diff
@@ -1093,20 +1093,29 @@ func (s *MongoStore) FindExistingAccounts(ctx context.Context, accounts []string
...
-func (s *MongoStore) UpdateSubscriptionRead(ctx context.Context, roomID, account string, lastSeenAt time.Time) error {
-	res, err := s.subscriptions.UpdateOne(ctx,
+// (roomID, account), clearing alert and hasMention, and returns the post-update
+// unread-thread count. FindOneAndUpdate rather than UpdateOne: the same single
+// round trip also yields the post-state the badge hook needs.
+// Returns model.ErrSubscriptionNotFound when no subscription matches.
+func (s *MongoStore) UpdateSubscriptionRead(ctx context.Context, roomID, account string, lastSeenAt time.Time) (int, error) {
+	var out struct {
+		ThreadUnread []string `bson:"threadUnread"`
+	}
```
julianshen, line 1103: _⚡ memory-exhaustion_ | _🟠 Warning_ | _`oom`_

**Room reads materialize the entire unread-thread array.**

The new FindOneAndUpdate projection returns and decodes every `threadUnread` ID merely to test whether the array is empty. This array grows through unread followed threads and has no application-level bound shown here; MongoDB's document limit still permits nearly 16 MB per subscription. Concurrent `messageRead` requests for such subscriptions can therefore allocate hundreds of megabytes or more, causing GC thrashing or process OOM. Project only the first array element and return a boolean, or use a server-side computed count, so memory remains constant regardless of thread cardinality.

<details>

```
In `room-service/store_mongo.go` around line 1103: The new FindOneAndUpdate projection returns and decodes every `threadUnread` ID merely to test whether the array is empty. This array grows through unread followed threads and has no application-level bound shown here; MongoDB's document limit still permits nearly 16 MB per subscription. Concurrent `messageRead` requests for such subscriptions can therefore allocate hundreds of megabytes or more, causing GC thrashing or process OOM. Project only the first array element and return a boolean, or use a server-side computed count, so memory remains constant regardless of thread cardinality.

```

</details>

room-service/handler.go
```diff
@@ -1384,11 +1387,12 @@ func (h *Handler) messageRead(c *natsrouter.Context) (*model.StatusReply, error)
...
+		h.badge.ClearRoom(ctx, account, roomID)
```
julianshen, line 1395: _🎯 concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**ClearRoom can erase a concurrent threaded-reply bump**

A threaded reply is handled by independent notification-worker and message-worker consumers. If notification-worker first bumps room R, a concurrent room read can then observe the still-empty post-update threadUnread array and SREM R, after which message-worker adds the thread parent to threadUnread without another badge bump. The surviving freshness marker makes subsequent counts omit R until marker expiry (10 minutes by default). inbox-worker/handler.go has the same interleaving. Keep ClearAll for room reads until these writes are ordered, or move the threadUnread and badge mutations behind one ordered owner so the removal decision is atomic with the thread update.

<details>

```
In `room-service/handler.go` around line 1395: A threaded reply is handled by independent notification-worker and message-worker consumers. If notification-worker first bumps room R, a concurrent room read can then observe the still-empty post-update threadUnread array and SREM R, after which message-worker adds the thread parent to threadUnread without another badge bump. The surviving freshness marker makes subsequent counts omit R until marker expiry (10 minutes by default). inbox-worker/handler.go has the same interleaving. Keep ClearAll for room reads until these writes are ordered, or move the threadUnread and badge mutations behind one ordered owner so the removal decision is atomic with the thread update.

```

</details>

user-service/service/badge.go
```diff
@@ -26,11 +26,17 @@ func (s *UserService) BadgeCountBatch(c *natsrouter.Context, req model.BadgeCoun
 			resp.Counts[account] = n
 			continue
 		}
-		ids, err := s.unreadRooms(c, account)
+		ids, degraded, err := s.unreadRooms(c, account)
```
julianshen, line 29: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Expired audiences trigger serial per-account recomputations**

BadgeCountBatch handles misses sequentially, and each unreadRooms call performs a subscription aggregation plus cross-site RPCs. After a room is idle longer than the marker TTL, every recipient's marker has expired despite jitter; the next message therefore performs one synchronous recomputation per audience member, with request latency equal to the sum of those operations and database/RPC load proportional to the full audience. Process misses through a bounded concurrent/bulk path with cancellation and backpressure, or serve the surviving set while scheduling deduplicated per-account refreshes instead of synchronously recomputing the whole expired audience.

<details>

```
In `user-service/service/badge.go` around line 29: BadgeCountBatch handles misses sequentially, and each unreadRooms call performs a subscription aggregation plus cross-site RPCs. After a room is idle longer than the marker TTL, every recipient's marker has expired despite jitter; the next message therefore performs one synchronous recomputation per audience member, with request latency equal to the sum of those operations and database/RPC load proportional to the full audience. Process misses through a bounded concurrent/bulk path with cancellation and backpressure, or serve the surviving set while scheduling deduplicated per-account refreshes instead of synchronously recomputing the whole expired audience.

```

</details>

inbox-worker/main.go
```diff
@@ -437,14 +437,22 @@ func (s *mongoInboxStore) UpdateSubscriptionRead(ctx context.Context, roomID, ac
...
+		SetReturnDocument(options.After)
+	err := s.subCol.FindOneAndUpdate(ctx, filter, update, opts).Decode(&out)
+	if errors.Is(err, mongo.ErrNoDocuments) {
+		// Either the subscription is missing (NAK to retry) or the order guard
+		// rejected a stale event (a no-op, not a failure).
+		return false, 0, s.naksIfSubscriptionMissing(ctx, account, roomID)
 	}
```
julianshen, line 451: _🏷️ reliability_ | _🟠 Warning_ | _`silent-failure`_

**A committed read can be acknowledged on retry without badge cleanup**

`FindOneAndUpdate` can commit successfully but return a timeout or transport error before the client observes the result. The event is then redelivered with the same timestamp; the strict `$lt` filter no longer matches, `naksIfSubscriptionMissing` sees that the subscription exists and returns nil, and this branch reports `applied=false`. `handleSubscriptionRead` consequently acknowledges the retry without calling `ClearRoom`, leaving the room in a marker-fresh badge set until marker expiry. Invalidate the account's badge cache when the update returns an ambiguous error, or distinguish an equal-timestamp duplicate from an older stale event and use `ClearAll` for the duplicate case. Add a test where the store applies the update but returns an error, followed by a redelivery.

<details>

```
In `inbox-worker/main.go` around line 451: `FindOneAndUpdate` can commit successfully but return a timeout or transport error before the client observes the result. The event is then redelivered with the same timestamp; the strict `$lt` filter no longer matches, `naksIfSubscriptionMissing` sees that the subscription exists and returns nil, and this branch reports `applied=false`. `handleSubscriptionRead` consequently acknowledges the retry without calling `ClearRoom`, leaving the room in a marker-fresh badge set until marker expiry. Invalidate the account's badge cache when the update returns an ambiguous error, or distinguish an equal-timestamp duplicate from an older stale event and use `ClearAll` for the duplicate case. Add a test where the store applies the update but returns an error, followed by a redelivery.

```

</details>

pkg/badgecache/badgecache.go
```diff
@@ -146,11 +186,11 @@ func (c *Cache) BumpBatch(ctx context.Context, accounts []string, roomID string)
 	return counts
 }
 
-// Seed creates or extends account's set from roomIDs plus triggerRoomID
-// (deduplicated, empty skipped) and returns the capped size. ok=false on any
-// Valkey error (fail-open).
+// Seed replaces account's set with roomIDs plus triggerRoomID (deduplicated,
+// empty skipped) and returns the capped size. ok=false on any Valkey error
```
julianshen, line 190: _🏷️ intent-match_ | _🟠 Warning_ | _`tgd-review`_

**Default construction no longer preserves the previous marker TTL**

The design promises that omitting WithMarkerTTL preserves today's behavior, but Seed and Reseed always call markerTTLSeconds, which shortens every marker by up to 20% even when markerTTL defaults to the set TTL. The unit test only checks the stored field, not the observable Redis expiry. Either remove/gate jitter behind an explicit option, or update the compatibility contract and add an integration test for the intended default behavior.

<details>

```
In `pkg/badgecache/badgecache.go` around line 190: The design promises that omitting WithMarkerTTL preserves today's behavior, but Seed and Reseed always call markerTTLSeconds, which shortens every marker by up to 20% even when markerTTL defaults to the set TTL. The unit test only checks the stored field, not the observable Redis expiry. Either remove/gate jitter behind an explicit option, or update the compatibility contract and add an integration test for the intended default behavior.

```

</details>

docs/notification-worker-downstream-contracts.md
```diff
@@ -147,16 +147,24 @@ Request / reply (`pkg/model/subscription.go`):
...
+(`BADGE_MARKER_TTL`, user-service only) rather than the set's TTL
+(`BADGE_CACHE_TTL`): the marker is stamped only by the Mongo-verifying
+seed/reseed paths and is never refreshed by a bump, so its expiry is the
+upper bound on how long the cached set can go unverified before the next
+count or push recomputes from Mongo and re-stamps it. A degraded computation
+— a cross-site `GetRoomsMeta` RPC failed and that site's rooms
+were dropped — is never cached: the best-effort count is still returned, but
+the marker is not stamped, so a knowingly-incomplete set is never blessed as
+verified. Failure direction is undercount-until-recompute, never a stuck
+overcount. `subscription.count` (unread=true) may itself be served from this
```
julianshen, line 165: _📝 documentation_ | _🟠 Warning_ | _`tgd-review`_

**The accuracy model incorrectly rules out overcount drift**

A failed ClearRoom can leave the room in the set while leaving its freshness marker intact, producing an overcount until marker expiry. A Seed computed before a concurrent read can likewise land after that read's SREM and restore the now-read room. The marker TTL bounds both cases, but the statement that failure direction is exclusively undercount is false. Update this contract and the new design's accuracy section to describe bounded undercount or overcount.

<details>

```
In `docs/notification-worker-downstream-contracts.md` around line 165: A failed ClearRoom can leave the room in the set while leaving its freshness marker intact, producing an overcount until marker expiry. A Seed computed before a concurrent read can likewise land after that read's SREM and restore the now-read room. The marker TTL bounds both cases, but the statement that failure direction is exclusively undercount is false. Update this contract and the new design's accuracy section to describe bounded undercount or overcount.

```

</details>

inbox-worker/handler.go
```diff
@@ -358,12 +358,15 @@ func (h *Handler) handleSubscriptionRead(ctx context.Context, evt *model.InboxEv
...
 		return fmt.Errorf("update subscription read for %q in room %q: %w", e.Account, e.RoomID, err)
 	}
-	// A read shrinks the unread set — drop it and recompute on next count.
-	if h.badge != nil {
-		h.badge.ClearAll(ctx, e.Account)
+	// The read settles message-unread, so the room stays unread only via an
+	// unread followed thread. Skip entirely when the order guard rejected the
+	// event — nothing was written, so nothing should be invalidated.
+	if applied && h.badge != nil && threadUnread == 0 {
+		h.badge.ClearRoom(ctx, e.Account, e.RoomID)
```
julianshen, line 369: _🎯 concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Delayed read events can delete newer-message badge bumps.**

A federated read carries a fixed `LastSeenAt` timestamp but is processed independently from message badge bumps. If a read at T is delayed, a newer message at T+1 can first `SADD` the room; the read event then applies because the stored read position is older and `ClearRoom` removes that newer message's bump. Plain messages do not update `threadUnread`, and the bump was already acknowledged, so nothing restores the room while the freshness marker remains valid, causing an undercount until marker expiry. The `$lt` guard only orders read events and does not prevent this interleaving. Keep `ClearAll` for federated reads until read events and message bumps share an ordering barrier, or carry enough message high-water state to prove no post-read message was already bumped before performing `ClearRoom`.

<details>

```
In `inbox-worker/handler.go` around line 369: A federated read carries a fixed `LastSeenAt` timestamp but is processed independently from message badge bumps. If a read at T is delayed, a newer message at T+1 can first `SADD` the room; the read event then applies because the stored read position is older and `ClearRoom` removes that newer message's bump. Plain messages do not update `threadUnread`, and the bump was already acknowledged, so nothing restores the room while the freshness marker remains valid, causing an undercount until marker expiry. The `$lt` guard only orders read events and does not prevent this interleaving. Keep `ClearAll` for federated reads until read events and message bumps share an ordering barrier, or carry enough message high-water state to prove no post-read message was already bumped before performing `ClearRoom`.

```

</details>

julianshen (commented):
tGD inline review

user-service/service/subscriptions.go
mliu33: This will be addressed in separate PR

mliu33 (approved):
Awesome work, thanks a lot!

PR #326, opened 2026-08-20
fix(broadcast-worker): attach resolved mentions to the message_edited event

chenjr0719:
## Summary

When a message is edited to **add** an `@mention`, the mentioned user's subscription is badged (handled earlier), but the `message_edited` event carried **no** structured `mentions[]` — unlike a freshly-sent message, whose event carries the resolved participants. A client that renders a mention chip from that structured array therefore couldn't render it on an edited message; it would only have the raw `@name` text.

This attaches the resolved participants to the edit event, mirroring the create path.

## Changes

- `pkg/model`: `EditRoomEvent` gains `mentions []Participant` (`omitempty`).
- `broadcast-worker`: new `resolveMentionParticipants` (parse the edited content → resolve to participants, same as the create path) and set `edit.Mentions` in **both** edit paths — main-room (`handleUpdated`) and thread-reply (`handleThreadUpdated`). The existing badge behavior is unchanged.
- `docs/client-api.md` + the events view: document the new `mentions` field on `message_edited`.

## Verification

- Unit: a new test asserts the `message_edited` event carries the resolved mention; existing edit tests updated for the added participant resolution. `broadcast-worker` + `pkg/model` suites green.
- Verified on our dev stack at the wire layer: alice sends a plain message then edits it to add `@bob`; the `message_edited` event now carries `mentions: [{account:"bob", ...}]` (previously absent), for both main-room and thread-reply edits.

## Summary by CodeRabbit

* **New Features**
  * Edited messages now include resolved participants for mentions in the updated content.
  * Mentions in edited thread replies are recognized and routed correctly.
  * Mention data is reused for cross-site message federation.
  * Reaction keys now support visible text and Unicode values with normalization and validation.

* **Documentation**
  * Updated client API documentation for edited-message mentions, reaction keys, and subscription-list ordering behavior.

### DIFF
--- a/broadcast-worker/handler.go
+++ b/broadcast-worker/handler.go
@@ -383,7 +383,12 @@ func (h *Handler) handleUpdated(ctx context.Context, evt *model.MessageEvent) er
 		}
 	}
 
+	// Resolve mentionees once: the same participants render on the edit event
+	// and route the cross-site badge, so we avoid a second FindUsersByAccounts.
+	participants := h.resolveEditMentions(ctx, parsed)
+
 	edit := buildEditRoomEvent(room, evt)
+	edit.Mentions = participants
 	if room.Type == model.RoomTypeChannel && h.encrypt {
 		if err := h.encryptEditedContent(ctx, room.ID, &edit); err != nil {
 			return fmt.Errorf("encrypt edit content for room %s: %w", room.ID, err)
@@ -392,26 +397,10 @@ func (h *Handler) handleUpdated(ctx context.Context, evt *model.MessageEvent) er
 	if err := h.publishMutation(ctx, room, model.RoomEventMessageEdited, msg.ID, &edit); err != nil {
 		return err
 	}
-	h.federateEditMentions(ctx, room.ID, msg.ID, parsed, *msg.EditedAt)
+	// Routing runs after the client broadcast; an unresolved mentionee has no
+	// site, so federateMentions simply relays nothing rather than failing the edit.
+	h.federateMentions(ctx, room.ID, msg.ID, participants, *msg.EditedAt)
 	return nil
-}
-
-// federateEditMentions resolves an edit's mentionees to their home sites and
-// relays the badge. The lookup is routing-only, so it runs after the client
-// broadcast, is skipped when federation is off, and a failure costs the relay
-// rather than the local badge.
-func (h *Handler) federateEditMentions(ctx context.Context, roomID, msgID string, parsed mention.ParseResult, at time.Time) {
-	if h.publish == nil || len(parsed.Accounts) == 0 {
-		return
-	}
-	users, err := h.userStore.FindUsersByAccounts(ctx, parsed.Accounts)
-	if err != nil {
-		slog.WarnContext(ctx, "user lookup failed for edited mentions, skipping federation",
-			"error", err, "room_id", roomID,
-			"request_id", natsutil.RequestIDFromContext(ctx))
-		return
-	}
-	h.federateMentions(ctx, roomID, msgID, mention.ResolveFromParsed(parsed, usersByAccount(users)).Participants, at)
 }
 
 // federateMentions relays the badge to each mentionee's home site, one event per
@@ -494,6 +483,26 @@ func (h *Handler) federateMentions(ctx context.Context, roomID, msgID string, pa
 	}
 }
 
+// resolveEditMentions resolves parsed @-mentions to participants (account +
+// display info) for the edit event, mirroring the create path so an edit-added
+// mention renders like a fresh one. The event's mentions[] is best-effort
+// enrichment, NOT the durable signal: the unread badge is set separately by
+// SetSubscriptionMentions and newContent still carries the raw @account, so on a
+// user-lookup error we drop the mentions[] enrichment entirely (return nil)
+// rather than emitting a partial set or failing/retrying the edit. nil when none.
+func (h *Handler) resolveEditMentions(ctx context.Context, parsed mention.ParseResult) []model.Participant {
+	if len(parsed.Accounts) == 0 {
+		return nil
+	}
+	users, err := h.userStore.FindUsersByAccounts(ctx, parsed.Accounts)
+	if err != nil {
+		slog.WarnContext(ctx, "user lookup failed resolving edit mentions, dropping edit mentions",
+			"error", err, "request_id", natsutil.RequestIDFromContext(ctx))
+		return nil
+	}
+	return mention.ResolveFromParsed(parsed, usersByAccount(users)).Participants
+}
+
 func (h *Handler) handleThreadUpdated(ctx context.Context, evt *model.MessageEvent) error {
 	msg := evt.Message
 	if msg.EditedAt == nil || msg.UpdatedAt == nil {
@@ -509,11 +518,12 @@ func (h *Handler) handleThreadUpdated(ctx context.Context, evt *model.MessageEve
 		return fmt.Errorf("get room %s: %w", msg.RoomID, err)
 	}
 
+	parsed := mention.Parse(msg.Content)
 	edit := buildEditRoomEvent(room, evt)
+	edit.Mentions = h.resolveEditMentions(ctx, parsed)
 
 	switch room.Type {
 	case model.RoomTypeChannel:
-		parsed := mention.Parse(msg.Content)
 		fanOut, err := h.channelThreadFanOut(ctx, room.ID, room.SiteID, parentMsgID, msg.UserAccount, parsed.Accounts, msg.ThreadParentMessageCreatedAt, evt.ThreadParentSenderAccount)
 		if err != nil {
 			return fmt.Errorf("channel thread fan-out for thread update of parent %s: %w", parentMsgID, err)
--- a/broadcast-worker/handler_test.go
+++ b/broadcast-worker/handler_test.go
@@ -840,6 +840,41 @@ func TestHandleUpdated_ChannelRoomScopedPublish(t *testing.T) {
 	assert.True(t, roomEvt.UpdatedAt.Equal(edited))
 }
 
+func TestHandleUpdated_AttachesMentionsToEditEvent(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	store := NewMockStore(ctrl)
+	us := NewMockUserStore(ctrl)
+	pub := &mockPublisher{}
+	keyStore := NewMockRoomKeyProvider(ctrl)
+
+	roomID := "r1"
+	store.EXPECT().GetRoom(gomock.Any(), roomID).Return(&model.Room{ID: roomID, Type: model.RoomTypeChannel, SiteID: "site-a"}, nil)
+	store.EXPECT().SetSubscriptionMentions(gomock.Any(), roomID, gomock.InAnyOrder([]string{"bob"}), gomock.Any()).Return(nil)
+	us.EXPECT().FindUsersByAccounts(gomock.Any(), []string{"bob"}).
+		Return([]model.User{{ID: "u-bob", Account: "bob", EngName: "Bob"}}, nil)
+
+	edited := time.Date(2026, 5, 14, 12, 5, 0, 0, time.UTC)
+	evt := model.MessageEvent{
+		Event: model.EventUpdated, SiteID: "site-a", Timestamp: edited.UnixMilli(),
+		Message: model.Message{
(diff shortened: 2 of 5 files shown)


julianshen:
**No actionable comments.** ✅

### Disputed

- `broadcast-worker/handler.go:388` (`mongodb`) — `FindUsersByAccounts` now runs synchronously with the parent handler context before `publishMutation`; the thread-edit path likewise performs it before fan-out and publication. During MongoDB server selection, pool contention, or replica-set failover, the query can block until that context expires. `resolveEditMentions` then swallows the error, but downstream operations still receive the canceled context, so the supposedly optional enrichment can prevent the authoritative edit from being published. For top-level edits, `SetSubscriptionMentions` may already have succeeded, leaving partial processing before broker redelivery. Use a bounded child context that expires early enough to preserve the parent publication budget, and test a lookup that blocks until the child deadline while verifying publication still occurs.
- `broadcast-worker/handler.go:388` (`nats`) — `FindUsersByAccounts` now runs with the handler context before `publishMutation`; the thread-edit path likewise runs it before fan-out and publication. If the store stalls until the handler deadline, `resolveEditMentions` swallows the error but returns with `ctx` canceled, so the subsequent NATS operation fails after earlier work such as `SetSubscriptionMentions` may have committed. A broker retry can then repeat that work, while an unretried failure leaves the edit unbroadcast. The enrichment being best-effort does not address the exhausted shared context. Use a bounded child deadline that reserves time for publication, falling back when it expires, and apply the same protection to both edit paths.
- `broadcast-worker/handler.go:388` (`performance-and-concurrency`) — This lookup now runs synchronously before the authoritative client publish, using the same context. Under user-store latency, FindUsersByAccounts can block until the handler deadline; resolveEditMentions then swallows the timeout, but publishMutation receives an already-canceled context and fails. Earlier badge writes may already have succeeded, so redelivery can repeat work, and this regression also occurs when federation is disabled. The same ordering affects thread edits before fan-out/publication. The best-effort nature of the payload does not prevent this failure mode. Bound enrichment with a short child timeout that preserves publication budget, or publish first and perform optional enrichment/routing without risking the main context.
- `broadcast-worker/handler.go:388` (`silent-failure`) — The optional `FindUsersByAccounts` lookup now runs on the handler context before the authoritative publish (and before thread fan-out). If it blocks until that context expires, `resolveEditMentions` logs and suppresses the lookup error, but the canceled context then makes `publishMutation` or fan-out fail. For top-level edits, earlier badge writes may already have succeeded, so redelivery can repeat partial work while clients never receive the edit. Calling this enrichment best-effort does not address the regression: previously the top-level routing lookup ran after publication. Bound the lookup with a child deadline that preserves enough parent-context budget for downstream work, then use the documented fallback when that child deadline expires.
- `broadcast-worker/handler.go:388` (`tgd-review`) — `resolveEditMentions` now performs a user-store call on the same context before `publishMutation`. If that call blocks until the handler deadline, it may return an error that is swallowed while leaving `ctx` canceled, causing the authoritative edit publish to fail rather than merely omitting enrichment. The same ordering affects thread edits before fan-out and publication. Use a bounded child context that expires before the handler's publication budget, then fall back to unresolved/empty participants while the parent context remains usable.

### Local review memory

- `broadcast-worker/handler.go:388` — @chenjr0719: The event's mentions[] is best-effort enrichment, not the durable signal: the unread mention badge is set on an independent path (SetSubscriptionMentions) and newContent still carries the raw @account, so a slow or failed lookup degrades e… ([thread](https://github.com/hmchangw/newchat/pull/326#discussion_r3840722928))
- `broadcast-worker/handler.go:498` — @chenjr0719: Same best-effort posture: a transient lookup error omits the display-name enrichment, but the unread badge (independent write) and the raw @account in newContent both survive, so no correctness signal is lost and a retry isn't needed. Docu… ([thread](https://github.com/hmchangw/newchat/pull/326#discussion_r3840722931))
- `broadcast-worker/handler_test.go:2970` — @chenjr0719: Fixed — added TestHandleThreadUpdated_AttachesMentionsToEditEvent, which decodes the emitted thread-edit event and asserts its Mentions, so dropping edit.Mentions from the thread path now fails the suite. (96b71ea3) ([thread](https://github.com/hmchangw/newchat/pull/326#discussion_r3840722934))

<details>

* `docs/client-api.md`
* `docs/client-api/events.md`
* `pkg/model/event.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

broadcast-worker/handler.go
```diff
@@ -383,7 +383,12 @@ func (h *Handler) handleUpdated(ctx context.Context, evt *model.MessageEvent) er
 		}
 	}
 
+	// Resolve mentionees once: the same participants render on the edit event
+	// and route the cross-site badge, so we avoid a second FindUsersByAccounts.
+	participants := h.resolveEditMentions(ctx, parsed)
```
julianshen, line 388: _🏷️ failure handling_ | _🟠 Warning_ | _`distributed-system`, `nats` (2 rules)_

**The lookup can consume the context before the edit is published.**

A slow user store now sits on the critical path before `publishMutation`, including when federation is disabled. Concrete timeline: an edit adds a mention; `FindUsersByAccounts` blocks until the handler context deadline; `resolveEditMentions` swallows the resulting error, but returns with the parent context already canceled; `publishMutation` then cannot broadcast the edit, so the caller observes failure despite the lookup being documented as non-fatal. Any earlier badge/database work may already have succeeded, and broker redelivery can repeat it. Give this optional lookup a bounded child timeout shorter than the handler's publication budget (then use account-only fallback), or otherwise ensure participant enrichment cannot exhaust the context needed for the authoritative client publish.

<details>

- **`nats`**:388: The optional `FindUsersByAccounts` call now runs before `publishMutation` using the same context. If that lookup stalls until the handler deadline, `resolveEditMentions` swallows the lookup error but returns with `ctx` already canceled, so the subsequent NATS publish fails and the edit may be retried or never broadcast. The same new ordering precedes thread fan-out and publication in `handleThreadUpdated`. This regresses the previous top-level ordering, whose removed comment explicitly kept the routing-only lookup after the client broadcast. Use a short child timeout for mention enrichment that expires well before the handler deadline, then fall back to empty/unresolved participants while publishing with the still-live parent context.

</details>

<details>

```
In `broadcast-worker/handler.go` around line 388: A slow user store now sits on the critical path before `publishMutation`, including when federation is disabled. Concrete timeline: an edit adds a mention; `FindUsersByAccounts` blocks until the handler context deadline; `resolveEditMentions` swallows the resulting error, but returns with the parent context already canceled; `publishMutation` then cannot broadcast the edit, so the caller observes failure despite the lookup being documented as non-fatal. Any earlier badge/database work may already have succeeded, and broker redelivery can repeat it. Give this optional lookup a bounded child timeout shorter than the handler's publication budget (then use account-only fallback), or otherwise ensure participant enrichment cannot exhaust the context needed for the authoritative client publish.

```

</details>

broadcast-worker/handler.go
```diff
@@ -494,6 +483,23 @@ func (h *Handler) federateMentions(ctx context.Context, roomID, msgID string, pa
...
+// display info) for the edit event, mirroring the create path so an edit-added
+// mention renders like a fresh one. On a user-lookup error it logs and returns
+// whatever resolved — unresolved accounts are omitted (same as create), and the
+// local mention badge still covers every account. nil when there are none.
+func (h *Handler) resolveEditMentions(ctx context.Context, parsed mention.ParseResult) []model.Participant {
+	if len(parsed.Accounts) == 0 {
+		return nil
+	}
+	users, err := h.userStore.FindUsersByAccounts(ctx, parsed.Accounts)
+	if err != nil {
```
julianshen, line 498: _🏷️ consistency_ | _🟠 Warning_ | _`mongodb`, `silent-failure` (2 rules)_

**Transient user lookup errors permanently omit edit mentions**

`FindUsersByAccounts` is a MongoDB-backed lookup, but its error is only logged before publishing the edit successfully. During a timeout or replica-set failover, resolved participants can therefore be missing from both top-level and thread edit events even though the content contains mentions; because the handler returns success, the message is not retried and clients permanently receive an incomplete event. This also conflicts with the documented behavior that `mentions` is omitted when there are none. Return the lookup error from `resolveEditMentions` and propagate it from both callers before publishing, so transient database failures trigger the worker's retry path.

<details>

- **`silent-failure`**:496: If FindUsersByAccounts fails, the handler continues with incomplete participants, publishes the edit, and returns nil. The edit is therefore acknowledged without retry even though its documented mentions payload may omit mentionees and federateMentions cannot route badges for participants lacking a site. Return the lookup error from resolveEditMentions and propagate it from both edit handlers before publishing; federation failures after publishing can remain best-effort to avoid duplicate broadcasts.

</details>

<details>

```
In `broadcast-worker/handler.go` around line 496: `FindUsersByAccounts` is a MongoDB-backed lookup, but its error is only logged before publishing the edit successfully. During a timeout or replica-set failover, resolved participants can therefore be missing from both top-level and thread edit events even though the content contains mentions; because the handler returns success, the message is not retried and clients permanently receive an incomplete event. This also conflicts with the documented behavior that `mentions` is omitted when there are none. Return the lookup error from `resolveEditMentions` and propagate it from both callers before publishing, so transient database failures trigger the worker's retry path.

```

</details>

broadcast-worker/handler_test.go
```diff
@@ -2886,6 +2923,8 @@ func TestHandleThreadUpdated_ChannelExcludesRestrictedAndNonMemberMentions(t *te
 
 	room := &model.Room{ID: "r1", Type: model.RoomTypeChannel, SiteID: "site-a"}
 	store.EXPECT().GetRoom(gomock.Any(), "r1").Return(room, nil)
+	// The edit event resolves participants for its mentions[] (mirrors create).
+	us.EXPECT().FindUsersByAccounts(gomock.Any(), gomock.Any()).Return(nil, nil)
```
julianshen, line 2970: _🧪 tests_ | _🟠 Warning_ | _`tgd-review`_

**Thread-edit mention payload has no regression assertion.**

The thread-update test only adds the new user-store expectation; it never verifies that the published EditRoomEvent contains the resolved participant. The top-level edit test does not exercise handleThreadUpdated, so removing `edit.Mentions = h.resolveEditMentions(...)` from that path would leave the suite passing. Decode the emitted thread-edit event and assert its Mentions account and resolved display fields.

<details>

```
In `broadcast-worker/handler_test.go` around line 2927: The thread-update test only adds the new user-store expectation; it never verifies that the published EditRoomEvent contains the resolved participant. The top-level edit test does not exercise handleThreadUpdated, so removing `edit.Mentions = h.resolveEditMentions(...)` from that path would leave the suite passing. Decode the emitted thread-edit event and assert its Mentions account and resolved display fields.

```

</details>

julianshen (commented):
tGD inline review

broadcast-worker/handler.go
chenjr0719: The event's mentions[] is best-effort enrichment, not the durable signal: the unread mention badge is set on an independent path (SetSubscriptionMentions) and newContent still carries the raw @account, so a slow or failed lookup degrades enrichment only, not correctness — documented on resolveEditMentions. Bounding the lookup with its own context to protect the publish budget applies equally to the create path, so tracking that as a separate follow-up rather than diverging one path here. (96b71ea3)

broadcast-worker/handler.go
chenjr0719: Same best-effort posture: a transient lookup error omits the display-name enrichment, but the unread badge (independent write) and the raw @account in newContent both survive, so no correctness signal is lost and a retry isn't needed. Documented on resolveEditMentions. (96b71ea3)

broadcast-worker/handler_test.go
chenjr0719: Fixed — added TestHandleThreadUpdated_AttachesMentionsToEditEvent, which decodes the emitted thread-edit event and asserts its Mentions, so dropping edit.Mentions from the thread path now fails the suite. (96b71ea3)

broadcast-worker/handler.go
mliu33: Should we return nil when user lookup fails here ? Please update @chenjr0719

chenjr0719: Done in 9333491a — `resolveEditMentions` now returns nil on the `FindUsersByAccounts` error path (no partial mentions[] on the event), and I updated the comment to match. broadcast-worker tests + lint green. Thanks!

mliu33 (approved):
Great work, thanks

PR #330, opened 2026-08-20
fix(loadgen): bound the recipient expectation map by the operations the ledger retires

flywindy:
The soak pod is OOM-killed roughly hourly at 8GB with `GOMEMLIMIT` set. #317
removed the journal's growth and it kept dying, so the remaining ~5GB was
unaccounted for. A heap profile names it.

## The evidence

```
      flat  flat%   sum%
 2472.41MB 60.79% 60.79%  main.incrementRecipientRoute (inline)
  413.86MB 10.18% 70.97%  main.(*recipientEvidence).ExpectDelivery
```

Both serve exactly one structure: `recipientEvidence.operations`. 71% of the
live heap is one map.

## Why it could only grow

| | |
|---|---|
| **In** | `ExpectDelivery` — one entry per send, no cap |
| **Out** | `Finalize` only, and its single caller is inside `soakFailureReconciler.Try` |

`failureLedger.Expire` finalizes an operation at its deadline, marks the
unobserved observers `unverified`, drops it from `l.active` — and never told
the evidence. So any operation the reconciler did not claim left its
expectation behind permanently.

## The arithmetic says the same thing

`loadgen_failure_inflight` sat at **33K**, which is exactly
`SEND_RATE 55/s × RECONCILE_DEADLINE 10m`. The ledger is bounded and behaving
precisely as designed.

The evidence map held enough to account for **2886MB** — an order of magnitude
more entries than the ledger. It was holding operations the ledger had already
finalized and forgotten.

`incrementRecipientRoute` beating `ExpectDelivery` six to one is the same story
from the allocation side: it allocates **one `map[recipientDeliveryRoute]int`
per (operation, recipient) pair**, and a channel room has 100 recipients. At 55
sends/s that is roughly 20M small maps an hour, each with its own header and
bucket.

## The fix

One rule, applied everywhere: **an expectation is released exactly when the
ledger stops holding its operation.** Identity, not time — the evidence and the
ledger cannot disagree about which operations exist, because only one of them
decides.

`failureLedger.Expire` returns the IDs it retired and the sweep forgets those.
That the ledger decides matters, because "everything overdue" is a different
set: `Expire` stops at `LEDGER_EXPIRE_BATCH`, and it skips **claimed**
operations the reconciler is mid-verification on — the ordinary path, not a
configuration edge. Both return `nil`, so no flag derived from "did the sweep
succeed" can tell them apart. Only the list of IDs can.

The same rule covers the paths that retire an operation without going through
expiry, each of which review surfaced in turn:

- `Expire`'s own error path. `finalizeLocked` deletes from `l.active` and
  compacts afterwards, so a compaction failure returns an error on an operation
  it has already retired. `Expire` now reports by what left `l.active` rather
  than by whether the call succeeded — an ID it does not report can never be
  reported by a later sweep.
- `AbandonUnsent`, which finalizes the operation out of the ledger entirely.
  The expectation is registered before the publish is attempted, so every
  locally-failed publish used to leave one behind for the life of the run. It
  asks `failureLedger.Retired` and releases on that answer alone, so the
  compaction case is covered there too rather than being re-derived per branch.

It is still not driven from *inside* `finalizeLocked`. Calling
`evidence.Forget` there would put the evidence lock underneath the ledger lock,
while the observer already takes them the other way round (`r.mu` released,
then `l.mu` via `Invalidate`). Every release happens after the ledger lock is
gone, so the evidence lock is still taken alone and no ordering between the two
exists at all.

The map also gains the ceiling it never had, because forgetting only what the
ledger retires means a ledger that stops retiring stops bounding the map. It is
set to twice `LEDGER_CAPACITY`: one ledger generation of headroom, since a
sweep can retire the whole ledger before cleanup runs, and sizing the two
equally would refuse a send inside that window.

## Three attempts at the same bound

Worth recording, because the first two both looked right.

1. **Registration time + a retention window.** Rejected during review: a
   recovered operation re-registers at replay time, so a shortened
   `RECONCILE_DEADLINE` deletes evidence the ledger correctly keeps alive.
2. **The operation's own persisted deadline.** Rejected during review: it still
   deletes evidence for operations `Expire` deferred by its batch limit or
   skipped because they were claimed.
3. **The IDs the ledger reports.** This one.

The first two were clocks approximating what the ledger did. `Expire` has two
exceptions no clock can model, so the approximation could only ever be close.
The net change removes `registeredAt`, the injected clock, the per-entry
deadline, the optional `Deadline` config field and the retention parameter,
along with the guards that existed to make them safe.

## Observability

Adds `loadgen_failure_recipient_expectations`. Without it, the only way to see
this map growing again is a heap profile — which is exactly how it stayed
hidden through an entire OOM investigation. A healthy run tracks
`loadgen_failure_inflight`; climbing past it is this regressing.

Documented in `docs/specs/o11y/storage-dependency-metrics.md`.

## Also in this PR

`TestGenerator_PacedRunBeatsSingleTickerCeiling` was failing on CI on code this
branch does not touch. It compared the two dispatch paths over one 200ms window
and required `paced > serial × 1.3`, but the pacer releases `rate × elapsed` and
no more, so the paced count clipped at exactly the target while a fast runner
drove the serial ticker close to it — leaving no result that could pass.

Two attempts to rescue the comparison failed for the same underlying reason. A
higher fixed target is reachable on a fast host and unreachable on a slow one;
a target scaled from the measured serial plateau fails too, because under load
the serial ticker collapses far harder than the worker pool does (a run
measured plateau 400 against paced 8000 — a ratio of 20 where an idle host
shows 2.2). Both counts are properties of the host, and their ratio has no
upper bound to write a threshold against.

So the blocking assertion moved to the seam with no timer in it.
`pacedDispatchRateWithTicks` accepts the tick channel, so
`TestPacedDispatch_ReleasesTheWholeBatchOnOneTick` drives one synthetic tick
through the real dispatch loop and asserts it releases the whole batch the rate
owes — 200 events at 100k rps against the 2ms floor — where a one-per-tick
regression releases exactly one. Verified by making that regression: it reports
200 against 1.

What remains on real timers is a measurement. `TestGenerator_MeasuresBothDispatchPaths`
logs both counts, which is what makes it useful when tuning the pacer, and
asserts only that each path dispatched at all.

## What this does not fix

- **Reconciliation appears not to be running.** `inflight` sitting at exactly
  `rate × deadline` means essentially every operation reaches its deadline
  rather than being reconciled — even though capacity looks sufficient
  (`READ_RATE 110 × RECONCILE_READ_SHARE 0.75 = 82.5/s` against 55/s of sends).
  This PR makes that survivable rather than fatal; the reason reconcile is idle
  is a separate investigation.
- `subscription_list` p95 pinned at 5s — unrelated, and #320 added the probe
  for it.
- The hourly `thread_subscriptions` scan wave in MongoDB, which shares the OOM
  restart period. If it disappears once the pod stops restarting, it was a
  consequence; if it survives, it is its own problem.

## Verification

`make lint` clean (0 issues), full `tools/loadgen` suite green under `-race`
(316s), `go vet -tags integration` compiles.

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Recipient expectation metrics now reflect current values at scrape time, even when cleanup is delayed.
  - Failure evidence is released reliably when operations complete, are abandoned, or are removed.
  - Evidence is preserved when cleanup is deferred or encounters errors, preventing premature data loss.
  - Improved pacing test reliability across different environments.

- **Documentation**
  - Simplified storage dependency metrics guidance and clarified telemetry, ledger coverage, and MongoDB support.
  - Documented recipient expectation metric collection behavior.

### DIFF
--- a/tools/loadgen/failure_ledger.go
+++ b/tools/loadgen/failure_ledger.go
@@ -692,6 +692,18 @@ func (l *failureLedger) AbandonWithReason(
 	return nil
 }
 
+// Retired reports that the ledger no longer holds the operation. An error from
+// a finalizing call does not answer that question: finalizeLocked commits the
+// removal and compacts afterwards, so a compaction failure returns an error on
+// an operation that is already gone — and one nothing reports can never be
+// reported again.
+func (l *failureLedger) Retired(operationID string) bool {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	_, active := l.active[operationID]
+	return !active
+}
+
 func (l *failureLedger) Observe(
 	operationID string,
 	observer failureObserver,
--- a/tools/loadgen/failure_ledger.go
+++ b/tools/loadgen/failure_ledger.go
@@ -887,12 +887,19 @@ func (l *failureLedger) Expire(now time.Time) ([]string, error) {
 		}
 		result := failureOperationResult(operation)
 		operationID := operation.ID
-		if err := l.finalizeLocked(
+		err := l.finalizeLocked(
 			operation, result, failureOperationFinalReason(operation, result), now,
-		); err != nil {
+		)
+		// Report by what left the ledger, not by whether the call succeeded.
+		// finalizeLocked drops the operation from l.active and compacts
+		// afterwards, so it can fail on one it has already retired — and an ID
+		// this loop does not report can never be reported by a later sweep.
+		if _, stillActive := l.active[operationID]; !stillActive {
+			finalized = append(finalized, operationID)
+		}
+		if err != nil {
 			return finalized, err
 		}
-		finalized = append(finalized, operationID)
 	}
 	return finalized, nil
 }
--- a/tools/loadgen/failure_ledger.go
+++ b/tools/loadgen/failure_ledger.go
@@ -832,18 +832,23 @@ func defaultFailureReason(
 	}
 }
 
-func (l *failureLedger) Expire(now time.Time) (int, error) {
+// Expire finalizes every operation past its deadline and returns the IDs it
+// actually finalized. The IDs matter to callers holding per-operation state:
+// a successful call is not a complete one — it stops at expireBatch and skips
+// claimed operations mid-verification — so "the sweep succeeded" is not a
+// licence to discard anything else that looks overdue.
+func (l *failureLedger) Expire(now time.Time) ([]string, error) {
 	if now.IsZero() {
 		now = l.now()
 	}
 	l.mu.Lock()
 	defer l.mu.Unlock()
 	if err := l.ensureOpen(); err != nil {
-		return 0, err
+		return nil, err
 	}
-	finalized := 0
+	var finalized []string
 	for _, operation := range l.active {
-		if l.expireBatch > 0 && finalized >= l.expireBatch {
+		if l.expireBatch > 0 && len(finalized) >= l.expireBatch {
 			break
 		}
 		if now.Before(operation.Deadline) {
@@ -881,12 +886,13 @@ func (l *failureLedger) Expire(now time.Time) (int, error) {
 			}
 		}
 		result := failureOperationResult(operation)
+		operationID := operation.ID
 		if err := l.finalizeLocked(
 			operation, result, failureOperationFinalReason(operation, result), now,
 		); err != nil {
 			return finalized, err
 		}
-		finalized++
+		finalized = append(finalized, operationID)
 	}
 	return finalized, nil
 }
--- a/tools/loadgen/failure_recipient.go
+++ b/tools/loadgen/failure_recipient.go
@@ -822,14 +822,26 @@ func withFailureRecipientEvidenceDir(directory string) failureRecipientObserverO
 	return func(observer *failureRecipientObserver) { observer.evidenceDir = directory }
 }
 
+// recipientEvidenceCapacityFactor is the headroom the expectation map keeps over
+// the ledger's own capacity.
+//
+// The sweep frees ledger slots inside Expire and frees the matching evidence
+// afterwards, in ForgetAll, so between the two the ledger has room the evidence
+// does not. Sized exactly like the ledger, a send arriving in that window is
+// refused registration even though the ledger would have admitted its
+// operation. One extra generation closes the window by construction: an
+// unbounded sweep can retire at most the whole ledger before cleanup runs, so
+// the lag can never exceed one capacity.
+const recipientEvidenceCapacityFactor = 2
+
 // withFailureRecipientEvidenceCapacity bounds the expectation map. Callers pass
 // the ledger's own capacity: an expectation exists per ledger operation, so the
 // ledger's admission limit binds first in a healthy run and this only engages
-// when expiry has stopped working.
+// once expiry has stopped working.
 func withFailureRecipientEvidenceCapacity(capacity int) failureRecipientObserverOption {
 	return func(o *failureRecipientObserver) {
 		if capacity > 0 {
-			o.evidence.capacity = capacity
+			o.evidence.capacity = capacity * recipientEvidenceCapacityFactor
 		}
 	}
 }
--- a/tools/loadgen/failure_recipient.go
+++ b/tools/loadgen/failure_recipient.go
@@ -613,13 +613,20 @@ func (r *recipientEvidence) Forget(operationID string) {
 }
 
 // ExpireBefore releases every expectation registered at or before cutoff and
-// returns how many went. The caller passes now minus the operation deadline, so
-// an expectation cannot outlive the operation it describes. It exists because the ledger's own expiry finalizes an
-// operation and forgets it without telling the evidence: Finalize is the only
-// other way out of this map and it runs inside the reconciler, so anything the
-// reconciler does not reach was retained for the life of the process. The
-// per-recipient route maps incrementRecipientRoute allocates make that the
-// largest object graph in the run.
+// returns how many went. Callers pass now minus the operation deadline, so an
+// expectation cannot outlive the operation it describes.
+//
+// It exists because the ledger's own expiry finalizes an operation and forgets
+// it without telling the evidence, while Finalize — the only other way out of
+// this map — runs inside the reconciler. Anything the reconciler never reaches
+// was therefore retained for the life of the process, and each entry holds one
+// route map per recipient, which made this the largest object graph in the run.
+//
+// A recovered expectation is registered when the restart replays it rather than
+// when its operation started, so it can outlive that operation by up to one
+// deadline. That over-retention is bounded and costs one deadline of entries
+// once per restart; treating a replayed entry as already aged would expire
+// evidence the run can still resolve.
 //
 // It is deliberately not driven from the ledger's finalize path. That would put
 // the evidence lock underneath the ledger lock, while the observer already
--- a/tools/loadgen/failure_recipient.go
+++ b/tools/loadgen/failure_recipient.go
@@ -351,19 +351,9 @@ type recipientExpectationConfig struct {
 	Route       recipientExpectedRoute
 	Source      recipientSetSource
 	Complete    bool
-	// Deadline is the operation's own deadline when the caller knows it. The
-	// ledger honours each operation's persisted deadline, so an entry expired
-	// on the sweep's current retention alone can be released while its
-	// operation is still live — a recovered operation accepted under a longer
-	// deadline is the reachable case. Optional: without it the sweep falls back
-	// to registration time, which matches for operations started in this
-	// process under this configuration.
-	Deadline time.Time
 }
 
 type recipientExpectation struct {
-	registeredAt      time.Time
-	deadline          time.Time
 	expected          map[string]struct{}
 	observed          map[string]map[recipientDeliveryRoute]int
 	mismatches        map[string]struct{}
@@ -386,15 +376,6 @@ type recipientEvidence struct {
 	// failing to expire would otherwise reinstate the growth this file exists
 	// to stop. Zero means unbounded.
 	capacity int
-	// now is injectable so expiry can be asserted without sleeping.
-	now func() time.Time
-}
-
-func (r *recipientEvidence) clock() time.Time {
-	if r.now == nil {
-		return time.Now().UTC()
-	}
-	return r.now().UTC()
 }
 
 type recipientEvidenceDisposition string
@@ -411,7 +392,6 @@ func newRecipientEvidence(allowDuplicates bool) *recipientEvidence {
 	return &recipientEvidence{
 		allowDuplicates: allowDuplicates,
 		operations:      make(map[string]*recipientExpectation),
-		now:             func() time.Time { return time.Now().UTC() },
 	}
 }
 
@@ -474,9 +454,7 @@ func (r *recipientEvidence) ExpectDelivery(config *recipientExpectationConfig) e
 			r.capacity)
 	}
 	r.operations[config.OperationID] = &recipientExpectation{
-		registeredAt: r.clock(),
-		deadline:     config.Deadline,
-		expected:     expected, observed: make(map[string]map[recipientDeliveryRoute]int), mismatches: make(map[string]struct{}),
+		expected: expected, observed: make(map[string]map[recipientDeliveryRoute]int), mismatches: make(map[string]struct{}),
 		roomID: config.RoomID, eventType: config.EventType,
 		route: config.Route, source: config.Source, complete: config.Complete,
 		durableMissing: make(map[string]struct{}), durableUnexpected: make(map[string]struct{}),
@@ -631,46 +609,30 @@ func (r *recipientEvidence) Forget(operationID string) {
 	delete(r.operations, operationID)
 }
 
-// ExpireBefore releases every expectation registered at or before cutoff and
-// returns how many went. Callers pass now minus the operation deadline, so an
-// expectation cannot outlive the operation it describes.
+// ForgetAll releases the expectations for the operations the ledger reports it
+// finalized, and returns how many went. Taking the IDs from the ledger is what
+// keeps the two aligned: a time-based bound cannot be, because Expire stops at
+// its batch limit and skips claimed operations mid-verification, so "past its
+// deadline" and "the ledger is done with it" are different sets.
 //
-// It exists because the ledger's own expiry finalizes an operation and forgets
-// it without telling the evidence, while Finalize — the only other way out of
-// this map — runs inside the reconciler. Anything the reconciler never reaches
-// was therefore retained for the life of the process, and each entry holds one
-// route map per recipient, which made this the largest object graph in the run.
-//
-// A recovered expectation is registered when the restart replays it rather than
-// when its operation started, so it can outlive that operation by up to one
-// deadline. That over-retention is bounded and costs one deadline of entries
-// once per restart; treating a replayed entry as already aged would expire
-// evidence the run can still resolve.
-//
-// It is deliberately not driven from the ledger's finalize path. That would put
-// the evidence lock underneath the ledger lock, while the observer already
-// takes them the other way round; a sweep that holds neither cannot deadlock
-// however those paths change.
-func (r *recipientEvidence) ExpireBefore(cutoff time.Time) int {
-	if r == nil {
+// The caller passes IDs the ledger has already returned, so this takes only the
+// evidence lock — the ledger's is released by then, and no ordering between the
+// two is created.
+func (r *recipientEvidence) ForgetAll(operationIDs []string) int {
+	if r == nil || len(operationIDs) == 0 {
 		return 0
 	}
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	expired := 0
-	for operationID, expectation := range r.operations {
-		if !expectation.deadline.IsZero() {
-			// It knows when its operation ends, so nothing else may release it.
-			if !expectation.deadline.Before(r.clock()) {
-				continue
-			}
-		} else if expectation.registeredAt.IsZero() || expectation.registeredAt.After(cutoff) {
+	forgotten := 0
+	for _, operationID := range operationIDs {
+		if _, tracked := r.operations[operationID]; !tracked {
 			continue
 		}
 		delete(r.operations, operationID)
-		expired++
+		forgotten++
 	}
-	return expired
+	return forgotten
 }
 
 // Len reports how many expectations are retained. The run needs this as a gauge:
@@ -973,17 +935,12 @@ func (o *failureRecipientObserver) Recover(operations []failureOperation) error 
 			err = o.ExpectDelivery(&recipientExpectationConfig{
 				OperationID: operation.ID, Recipients: recipients,
 				Route: recipientExpectedRouteAny, Source: recipientSetSourceLegacy, Complete: false,
-				Deadline: operation.Deadline,
 			})
 		} else {
 			err = o.ExpectDelivery(&recipientExpectationConfig{
 				OperationID: operation.ID, Recipients: recipients,
 				RoomID: operation.Targets["roomId"], EventType: eventType,
 				Route: route, Source: source, Complete: complete,
-				// The persisted deadline, not this process's retention: the
-				// ledger keeps the operation until this instant regardless of
-				// how SOAK_RECONCILE_DEADLINE was changed between runs.
-				Deadline: operation.Deadline,
 			})
 		}
 		if err != nil {
--- a/tools/loadgen/failure_recipient.go
+++ b/tools/loadgen/failure_recipient.go
@@ -351,10 +351,19 @@ type recipientExpectationConfig struct {
 	Route       recipientExpectedRoute
 	Source      recipientSetSource
 	Complete    bool
+	// Deadline is the operation's own deadline when the caller knows it. The
+	// ledger honours each operation's persisted deadline, so an entry expired
+	// on the sweep's current retention alone can be released while its
+	// operation is still live — a recovered operation accepted under a longer
+	// deadline is the reachable case. Optional: without it the sweep falls back
+	// to registration time, which matches for operations started in this
+	// process under this configuration.
+	Deadline time.Time
 }
 
 type recipientExpectation struct {
 	registeredAt      time.Time
+	deadline          time.Time
 	expected          map[string]struct{}
 	observed          map[string]map[recipientDeliveryRoute]int
 	mismatches        map[string]struct{}
@@ -373,6 +382,10 @@ type recipientEvidence struct {
 	mu              sync.Mutex
 	allowDuplicates bool
 	operations      map[string]*recipientExpectation
+	// capacity bounds the map when expiry cannot run — a ledger that keeps
+	// failing to expire would otherwise reinstate the growth this file exists
+	// to stop. Zero means unbounded.
+	capacity int
 	// now is injectable so expiry can be asserted without sleeping.
 	now func() time.Time
 }
@@ -455,8 +468,14 @@ func (r *recipientEvidence) ExpectDelivery(config *recipientExpectationConfig) e
 	if _, exists := r.operations[config.OperationID]; exists {
 		return fmt.Errorf("recipient expectation %q already exists", config.OperationID)
 	}
+	if r.capacity > 0 && len(r.operations) >= r.capacity {
+		return fmt.Errorf(
+			"recipient expectation capacity %d exceeded; evidence is not being expired",
+			r.capacity)
+	}
 	r.operations[config.OperationID] = &recipientExpectation{
 		registeredAt: r.clock(),
+		deadline:     config.Deadline,
 		expected:     expected, observed: make(map[string]map[recipientDeliveryRoute]int), mismatches: make(map[string]struct{}),
 		roomID: config.RoomID, eventType: config.EventType,
 		route: config.Route, source: config.Source, complete: config.Complete,
@@ -640,7 +659,12 @@ func (r *recipientEvidence) ExpireBefore(cutoff time.Time) int {
 	defer r.mu.Unlock()
 	expired := 0
 	for operationID, expectation := range r.operations {
-		if expectation.registeredAt.IsZero() || expectation.registeredAt.After(cutoff) {
+		if !expectation.deadline.IsZero() {
+			// It knows when its operation ends, so nothing else may release it.
+			if !expectation.deadline.Before(r.clock()) {
+				continue
+			}
+		} else if expectation.registeredAt.IsZero() || expectation.registeredAt.After(cutoff) {
 			continue
 		}
 		delete(r.operations, operationID)
@@ -836,6 +860,18 @@ func withFailureRecipientEvidenceDir(directory string) failureRecipientObserverO
 	return func(observer *failureRecipientObserver) { observer.evidenceDir = directory }
 }
 
+// withFailureRecipientEvidenceCapacity bounds the expectation map. Callers pass
+// the ledger's own capacity: an expectation exists per ledger operation, so the
+// ledger's admission limit binds first in a healthy run and this only engages
+// when expiry has stopped working.
+func withFailureRecipientEvidenceCapacity(capacity int) failureRecipientObserverOption {
+	return func(o *failureRecipientObserver) {
+		if capacity > 0 {
+			o.evidence.capacity = capacity
+		}
+	}
+}
+
 func withFailureRecipientDuplicatePolicy(allow bool) failureRecipientObserverOption {
 	return func(observer *failureRecipientObserver) {
 		observer.evidence.allowDuplicates = allow
@@ -937,12 +973,17 @@ func (o *failureRecipientObserver) Recover(operations []failureOperation) error 
 			err = o.ExpectDelivery(&recipientExpectationConfig{
 				OperationID: operation.ID, Recipients: recipients,
 				Route: recipientExpectedRouteAny, Source: recipientSetSourceLegacy, Complete: false,
+				Deadline: operation.Deadline,
 			})
 		} else {
 			err = o.ExpectDelivery(&recipientExpectationConfig{
 				OperationID: operation.ID, Recipients: recipients,
 				RoomID: operation.Targets["roomId"], EventType: eventType,
 				Route: route, Source: source, Complete: complete,
+				// The persisted deadline, not this process's retention: the
+				// ledger keeps the operation until this instant regardless of
+				// how SOAK_RECONCILE_DEADLINE was changed between runs.
+				Deadline: operation.Deadline,
 			})
 		}
 		if err != nil {
--- a/tools/loadgen/failure_recipient.go
+++ b/tools/loadgen/failure_recipient.go
@@ -354,6 +354,7 @@ type recipientExpectationConfig struct {
 }
 
 type recipientExpectation struct {
+	registeredAt      time.Time
 	expected          map[string]struct{}
 	observed          map[string]map[recipientDeliveryRoute]int
 	mismatches        map[string]struct{}
@@ -372,6 +373,15 @@ type recipientEvidence struct {
 	mu              sync.Mutex
 	allowDuplicates bool
 	operations      map[string]*recipientExpectation
+	// now is injectable so expiry can be asserted without sleeping.
+	now func() time.Time
+}
+
+func (r *recipientEvidence) clock() time.Time {
+	if r.now == nil {
+		return time.Now().UTC()
+	}
+	return r.now().UTC()
 }
 
 type recipientEvidenceDisposition string
@@ -385,7 +395,11 @@ const (
 )
 
 func newRecipientEvidence(allowDuplicates bool) *recipientEvidence {
-	return &recipientEvidence{allowDuplicates: allowDuplicates, operations: make(map[string]*recipientExpectation)}
+	return &recipientEvidence{
+		allowDuplicates: allowDuplicates,
+		operations:      make(map[string]*recipientExpectation),
+		now:             func() time.Time { return time.Now().UTC() },
+	}
 }
 
 func (r *recipientEvidence) Expect(operationID string, recipients []string) error {
@@ -442,7 +456,8 @@ func (r *recipientEvidence) ExpectDelivery(config *recipientExpectationConfig) e
 		return fmt.Errorf("recipient expectation %q already exists", config.OperationID)
 	}
 	r.operations[config.OperationID] = &recipientExpectation{
-		expected: expected, observed: make(map[string]map[recipientDeliveryRoute]int), mismatches: make(map[string]struct{}),
+		registeredAt: r.clock(),
+		expected:     expected, observed: make(map[string]map[recipientDeliveryRoute]int), mismatches: make(map[string]struct{}),
 		roomID: config.RoomID, eventType: config.EventType,
 		route: config.Route, source: config.Source, complete: config.Complete,
 		durableMissing: make(map[string]struct{}), durableUnexpected: make(map[string]struct{}),
@@ -595,6 +610,48 @@ func (r *recipientEvidence) Forget(operationID string) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 	delete(r.operations, operationID)
+}
+
+// ExpireBefore releases every expectation registered at or before cutoff and
+// returns how many went. The caller passes now minus the operation deadline, so
+// an expectation cannot outlive the operation it describes. It exists because the ledger's own expiry finalizes an
+// operation and forgets it without telling the evidence: Finalize is the only
+// other way out of this map and it runs inside the reconciler, so anything the
+// reconciler does not reach was retained for the life of the process. The
+// per-recipient route maps incrementRecipientRoute allocates make that the
+// largest object graph in the run.
+//
+// It is deliberately not driven from the ledger's finalize path. That would put
+// the evidence lock underneath the ledger lock, while the observer already
+// takes them the other way round; a sweep that holds neither cannot deadlock
+// however those paths change.
+func (r *recipientEvidence) ExpireBefore(cutoff time.Time) int {
+	if r == nil {
+		return 0
+	}
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	expired := 0
+	for operationID, expectation := range r.operations {
+		if expectation.registeredAt.IsZero() || expectation.registeredAt.After(cutoff) {
+			continue
+		}
+		delete(r.operations, operationID)
+		expired++
+	}
+	return expired
+}
+
+// Len reports how many expectations are retained. The run needs this as a gauge:
+// a count that tracks the ledger's own in-flight number is healthy, one that
+// climbs past it is this map leaking again.
+func (r *recipientEvidence) Len() int {
+	if r == nil {
+		return 0
+	}
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	return len(r.operations)
 }
 
 func (r *recipientEvidence) Complete(operationID string) bool {
--- a/tools/loadgen/generator_test.go
+++ b/tools/loadgen/generator_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -368,78 +369,66 @@ func dispatchPeak(t *testing.T, maxInFlight, rate int, window time.Duration, tri
 	return best
 }
 
-// pacedCeilingProbeRate is only used to measure what the serial ticker manages
-// on this host; the rate the two paths are then compared at is derived from
-// that measurement. pacedCeilingHeadroom is how far past the measured plateau
-// that derived rate sits.
 const (
-	pacedCeilingProbeRate = 100_000
-	pacedCeilingHeadroom  = 20
-	pacedCeilingWindow    = 200 * time.Millisecond
-	pacedCeilingTrials    = 3
+	pacedMeasureRate   = 100_000
+	pacedMeasureWindow = 200 * time.Millisecond
+	pacedMeasureTrials = 3
 )
 
-// pacedCeilingRateFrom turns a dispatch count measured over window into a rate
-// headroom times higher, falling back to the probe rate when the measurement
-// was empty.
-func pacedCeilingRateFrom(measured int, window time.Duration, headroom int) int {
-	if measured <= 0 {
-		return pacedCeilingProbeRate
-	}
-	return int(float64(measured)/window.Seconds()) * headroom
-}
-
-// Regression for the single-ticker RPS ceiling: the legacy serial path
-// (MaxInFlight=0) releases one event per delivered tick, and the runtime cannot
-// deliver a tick per microsecond, so it plateaus far below target. The batched
-// pacer releases rate*interval events per coarse tick and must out-dispatch it
-// by a wide margin. A relative comparison (same process, back to back) cancels
-// host-speed and race-detector overhead.
+// The regression that produced the batched pacer: the serial path releases one
+// event per delivered tick, and the runtime cannot deliver a tick per
+// microsecond, so it plateaus far below any high target. This is that property
+// with the timer taken out of it. pacedDispatchRateWithTicks accepts the tick
+// channel, so one tick can be shown to release the whole batch the rate is owed
+// — 200 events here — where a one-per-tick regression releases exactly one.
 //
-// The rate they are compared at is measured, not chosen. Both counts are only
-// ceilings while the target stays out of reach — a target the paced path can
-// reach reports the target instead, and the comparison stops being about the
-// pacer (see TestGenerator_PacedDispatchReportsAReachableTarget). A constant
-// cannot hold that open: the fastest host observed drove the serial ticker to
-// 78k/s, which is above the rate the slowest one could dispatch at all, so no
-// single number is simultaneously out of reach on one and in reach on the
-// other. Deriving it from this host's own plateau is.
-func TestGenerator_PacedRunBeatsSingleTickerCeiling(t *testing.T) {
-	plateau := dispatchPeak(t, 0, pacedCeilingProbeRate, pacedCeilingWindow, pacedCeilingTrials)
-	require.Positive(t, plateau, "the serial ticker dispatched nothing to scale from")
-	rate := pacedCeilingRateFrom(plateau, pacedCeilingWindow, pacedCeilingHeadroom)
+// This is the blocking assertion. The comparison against the serial path can
+// only be measured on real timers, and no measurement of it is safe to assert
+// on: the two paths degrade at completely different rates under load, so their
+// ratio has no upper bound to write a threshold against.
+func TestPacedDispatch_ReleasesTheWholeBatchOnOneTick(t *testing.T) {
+	start := time.Unix(0, 0).UTC()
+	pacer := newPacer(pacedMeasureRate, start)
+	require.Equal(t, minEmitInterval, pacer.interval,
+		"a target this high must clamp to the tick floor for a batch to exist")
+	require.Greater(t, pacer.perTick, 1.0)
 
-	serial := dispatchPeak(t, 0, rate, pacedCeilingWindow, pacedCeilingTrials)
-	paced := dispatchPeak(t, 5000, rate, pacedCeilingWindow, pacedCeilingTrials)
+	var dispatched atomic.Int64
+	ticks := make(chan time.Time)
+	done := make(chan struct{})
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		defer close(done)
+		pacedDispatchRateWithTicks(ctx, pacer, ticks, 4096,
+			func(int) {}, func() {},
+			func(context.Context) { dispatched.Add(1) })
+	}()
 
-	require.Positive(t, serial)
-	t.Logf("plateau=%d derived rate=%d serial=%d paced=%d", plateau, rate, serial, paced)
-	require.Less(t, paced, int(float64(rate)*pacedCeilingWindow.Seconds()),
-		"this host dispatches more than %dx what its own serial ticker manages, so the margin below measures the target rather than the pacer: raise pacedCeilingHeadroom",
-		pacedCeilingHeadroom)
-	assert.Greater(t, float64(paced), float64(serial)*1.3,
-		"batched pacer (%d) should out-dispatch the serial ticker (%d) at %d rps",
-		paced, serial, rate)
+	// Unbuffered: the send returns only once the dispatch loop has taken the
+	// tick, so cancelling afterwards cannot race ahead of the batch it owes.
+	ticks <- start.Add(pacer.interval)
+	cancel()
+	<-done
+
+	assert.Equal(t, int64(pacer.perTick), dispatched.Load(),
+		"one tick must release the events the rate owes for that interval")
 }
 
-// The pacer releases rate*elapsed events and no more, so a target the host can
-// reach is what the count reports — the path's own ceiling stays invisible, and
-// the two paths converge on the same number. That is how the ceiling assertion
-// above failed on CI at 100k rps: the paced count clipped at exactly 20000
-// while the runner drove the serial ticker to 15586, which no 1.3x margin can
-// sit above. The failure mode belongs to a *faster* host, so it needs a test
-// that does not depend on host speed to surface it.
-func TestGenerator_PacedDispatchReportsAReachableTarget(t *testing.T) {
-	const rate = 5000
-	window := 200 * time.Millisecond
-	target := int(float64(rate) * window.Seconds())
+// A measurement, not an assertion. Both counts are host-dependent and the
+// review history of this file is three attempts to find a threshold that holds
+// across machines: a fixed target is reachable on a fast host and unreachable
+// on a slow one, and a target scaled from the serial plateau fails too, because
+// under load the serial ticker collapses far harder than the pool does
+// (observed: plateau 400, paced 8000 — a ratio of 20 where an idle host shows
+// 2.2). The numbers are logged because they are useful when tuning the pacer;
+// the only thing asserted is that both paths dispatched at all.
+func TestGenerator_MeasuresBothDispatchPaths(t *testing.T) {
+	serial := dispatchPeak(t, 0, pacedMeasureRate, pacedMeasureWindow, pacedMeasureTrials)
+	paced := dispatchPeak(t, 5000, pacedMeasureRate, pacedMeasureWindow, pacedMeasureTrials)
 
-	paced := dispatchPeak(t, 5000, rate, window, 2)
-
-	assert.LessOrEqual(t, paced, target,
-		"the pacer must never release more than the target rate allows")
-	assert.InDelta(t, target, paced, float64(target)*0.1,
-		"a reachable target caps the count, leaving nothing of the pacer's own ceiling to compare")
+	t.Logf("at %d rps over %s: serial=%d paced=%d", pacedMeasureRate, pacedMeasureWindow, serial, paced)
+	require.Positive(t, serial, "the serial ticker dispatched nothing")
+	require.Positive(t, paced, "the batched pacer dispatched nothing")
 }
 
 func TestParseInjectMode(t *testing.T) {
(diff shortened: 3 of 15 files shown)


flywindy:
Not acting on the docstring-coverage warning, and this time the number itself is wrong rather than merely misdirected.

It reports **0.00%**. I counted every top-level declaration this PR adds and checked whether the preceding line is a comment:

| | |
|---|---|
| Added declarations | 16 |
| Carrying a doc comment | **13 (81%)** |

That is above the 80% threshold the check is enforcing, so 0.00% is not a strict reading of this diff — it is not measuring this diff at all. Worth noting the same check reported 71.43% on #320, where the touched files were genuinely mostly undocumented; the two results are not consistent with each other.

The three without their own comment:

- `soakFailureExpirySettings` and `withSoakFailureExpiryMetrics` sit directly beneath the `soakFailureExpiryOption` doc comment, which states why the trio exists ("Metrics are optional so the tests that only assert expiry do not have to build a registry"). Repeating that on each of the three is the restatement CLAUDE.md asks us not to write.
- `clock()` is a two-line accessor that returns the injected clock or `time.Now().UTC()`.

Everything a reviewer would actually stop at is documented, including the two decisions in this change that are not obvious from the code: why expiry runs on the sweep rather than the ledger's finalize path (lock ordering), and why a recovered expectation is registered at replay time rather than aged to its original start.

CI is 5/5 green on `57f5c47`.

---

julianshen:
**2 findings · 1 inline comment posted** — 🟠 2 warning

### 💬 Additional comments (1)


**`tools/loadgen/generator_test.go:439`**

_🧪 tests_ | _🟠 Warning_ | _`tgd-review`_

**The generator regression test no longer verifies paced-path selection**

The replacement integration test only requires both configurations to publish at least once. It therefore passes if `Generator.Run` ignores `MaxInFlight` and always selects the serial path, while `TestPacedDispatch_ReleasesTheWholeBatchOnOneTick` still passes because it calls the paced helper directly. Preserve deterministic coverage of the wiring by injecting a tick source/clock into `Generator` and asserting that a generator configured with `MaxInFlight > 0` releases a full batch for one tick.

### Related work

- [PR #317](https://github.com/hmchangw/newchat/pull/317) — fix\(loadgen\): stop the soak OOM loop and make its failures readable (merged)
- [PR #320](https://github.com/hmchangw/newchat/pull/320) — feat\(loadgen\): make the subscription-list request shape configurable (merged)

<details>
<summary>📒 Files reviewed (15)</summary>

* `tools/loadgen/failure_ledger_expiry_test.go`
* `tools/loadgen/failure_ledger_review_test.go`
* `tools/loadgen/failure_ledger_test.go`
* `tools/loadgen/failure_observation_runtime.go`
* `tools/loadgen/failure_observation_runtime_test.go`
* `tools/loadgen/failure_recipient_expiry_test.go`
* `tools/loadgen/generator_test.go`
* `tools/loadgen/metrics.go`
* `tools/loadgen/soak_failure_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

julianshen (commented):
tGD inline review

tools/loadgen/soak_failure.go
```diff
@@ -81,6 +100,23 @@ func runSoakFailureExpiry(
...
+			// retained for the life of the process — and each entry holds one
+			// route map per recipient, which is what made this the largest
+			// object graph in the run. Sweeping here bounds the evidence by the
+			// same deadline that bounds the ledger.
+			// A zero window would put the cutoff at now and release every
+			// expectation, including operations still inside their deadline.
+			// Config validation forbids a zero SOAK_RECONCILE_DEADLINE, so
+			// this guards a caller mistake rather than a reachable setting.
+			if retention > 0 {
+				evidence.ExpireBefore(at.Add(-retention))
```
julianshen: _🏷️ failure-recovery_ | _🟠 Warning_ | _`distributed-system`_

**Current retention can expire evidence for a still-live recovered operation.**

Evidence expiry uses replay-time `registeredAt` plus the current process's `ReconcileDeadline`, independently of the deadline stored by the ledger. Concrete rollout path: an operation accepted under a 60-minute deadline is recovered after restart with the deadline changed to 10 minutes; replay registers its expectation at startup, and after 10 minutes this call deletes it while `ledger.Expire(at)` correctly leaves the operation live until its original deadline. A later delivery or reconciliation before that original deadline is then treated as untracked/unverified. Expire evidence using each operation's persisted deadline, or remove expectations only for operation IDs the ledger reports as finalized/expired, rather than applying the current global retention to replay-time registration.

<details>

```
In `tools/loadgen/soak_failure.go` around line 115: Evidence expiry uses replay-time `registeredAt` plus the current process's `ReconcileDeadline`, independently of the deadline stored by the ledger. Concrete rollout path: an operation accepted under a 60-minute deadline is recovered after restart with the deadline changed to 10 minutes; replay registers its expectation at startup, and after 10 minutes this call deletes it while `ledger.Expire(at)` correctly leaves the operation live until its original deadline. A later delivery or reconciliation before that original deadline is then treated as untracked/unverified. Expire evidence using each operation's persisted deadline, or remove expectations only for operation IDs the ledger reports as finalized/expired, rather than applying the current global retention to replay-time registration.

```

</details>

tools/loadgen/soak_failure.go
```diff
@@ -81,6 +100,23 @@ func runSoakFailureExpiry(
...
+			// reconciler. Anything the reconciler never reaches was therefore
+			// retained for the life of the process — and each entry holds one
+			// route map per recipient, which is what made this the largest
+			// object graph in the run. Sweeping here bounds the evidence by the
+			// same deadline that bounds the ledger.
+			// A zero window would put the cutoff at now and release every
+			// expectation, including operations still inside their deadline.
+			// Config validation forbids a zero SOAK_RECONCILE_DEADLINE, so
+			// this guards a caller mistake rather than a reachable setting.
+			if retention > 0 {
```
julianshen: _🏷️ reliability_ | _🟠 Warning_ | _`silent-failure`_

**Evidence is discarded even when durable ledger expiry fails**

When `ledger.Expire(at)` fails, the callback in `soak_main.go` only logs the error, but this code still deletes every aged recipient expectation. A WAL or other persistence failure can therefore leave the ledger operation unresolved while irrevocably removing the observations needed to reconcile it, causing a later retry or reconciliation to report missing/unverified evidence instead of the actual recipient result. Retain the evidence when `ledger.Expire` returns an error and retry both operations on the next tick; only call `ExpireBefore` after ledger expiry succeeds.

<details>

```
In `tools/loadgen/soak_failure.go` around line 114: When `ledger.Expire(at)` fails, the callback in `soak_main.go` only logs the error, but this code still deletes every aged recipient expectation. A WAL or other persistence failure can therefore leave the ledger operation unresolved while irrevocably removing the observations needed to reconcile it, causing a later retry or reconciliation to report missing/unverified evidence instead of the actual recipient result. Retain the evidence when `ledger.Expire` returns an error and retry both operations on the next tick; only call `ExpireBefore` after ledger expiry succeeds.

```

</details>

tools/loadgen/soak_failure.go
flywindy: Confirmed and fixed in `4d81aaf`. The rollout path you describe is exactly reachable, and I had reasoned about the wrong direction.

I checked `failureLedger.Expire` — it gates on `now.Before(operation.Deadline)` using the replayed operation's own `Deadline`, so the ledger is authoritative and my sweep was using a different clock. My doc comment even discussed recovered expectations, but only the case where replay-time registration makes evidence live *longer* than its operation. Your case is the opposite one, and it loses data rather than wasting memory.

The fix takes your first option. `recipientExpectationConfig` gains an optional `Deadline`, set at the two production sites that know it:

- the tracker passes `startedAt.Add(t.deadline)` — literally the same value it hands `ledger.Start` two statements later
- `Recover` passes `operation.Deadline`, the persisted one

`ExpireBefore` holds an expectation that knows its deadline to that, and falls back to registration time only when it does not. Optional matters here: making it required was my first attempt and it broke 25 tests, which is why the original commit derived the bound instead.

Not taking the second option — releasing only for IDs the ledger reports finalized. That is the coupling this PR deliberately avoids: it puts the evidence lock under the ledger lock while the observer already takes them the other way round (`r.mu` released, then `l.mu` via `Invalidate`). Safe today, but on an unwritten invariant. Carrying the deadline gets the same correctness with no lock relationship at all.

Covered by `TestRecipientEvidence_HoldsAnExpectationToItsOwnDeadline`, which fails on the old code at the ten-minute mark.

---

tools/loadgen/soak_failure.go
flywindy: Confirmed and fixed in `4d81aaf`.

I checked whether the stated consequence actually holds, because it depends on what a failed `Expire` leaves behind. It does: `invalidateLocked` only sets `invalidReason` and notifies the recorder — it does not close the ledger or stop it accepting observations. So after a WAL failure the operations are still in `l.active` and still reconcilable, and releasing their evidence on the same tick destroyed the only record of what was delivered while the operation itself survived. `ExpireBefore` is now gated on the ledger's expiry succeeding, and both retry on the next tick.

One consequence worth naming, because it is the reason I did not stop at the gate. If `Expire` keeps failing, gating means the evidence map grows unbounded again — which is precisely the failure this PR exists to fix, so the fix for your finding would have handed back the OOM.

So the map also gains the ceiling it never had, set to the ledger's own `LedgerCapacity`. An expectation exists per ledger operation, so the ledger's admission limit binds first in any healthy run and the ceiling only engages once expiry has stopped working — at which point refusing to register a new expectation, loudly, is the honest answer rather than silently ballooning until the pod dies.

Two tests: `TestRunSoakFailureExpiry_KeepsEvidenceWhenLedgerExpiryFails` (closed ledger, evidence survives the tick, error reported) and `TestRecipientEvidence_RefusesToGrowPastItsCapacity`.

---

tools/loadgen/soak_failure.go
```diff
@@ -81,6 +102,28 @@ func runSoakFailureExpiry(
...
+			// leaves those operations live and still reconcilable, so
+			// releasing their evidence on the same tick would destroy the only
+			// record of what was delivered. The next tick retries both.
+			//
+			// A zero window would put the cutoff at now and release every
+			// expectation, including operations still inside their deadline.
+			// Config validation forbids a zero SOAK_RECONCILE_DEADLINE, so
+			// this guards a caller mistake rather than a reachable setting.
+			if ledgerExpired && retention > 0 {
+				evidence.ExpireBefore(at.Add(-retention))
```
flywindy: Still not fully fixed in `4d81aaf`. The persisted-deadline and `Expire`-error paths are fixed, but a successful `ledger.Expire` can still leave overdue operations active: `ExpireBatch` may process only part of the backlog, and claimed operations are skipped. In both cases this call deletes their evidence; I reproduced `ledger active=1, evidence=0` for each on the current head under `-race`. Please forget only the exact IDs actually removed by the ledger; returning those IDs and deleting evidence after `Expire` returns does not nest the locks.

tools/loadgen/failure_recipient.go
```diff
@@ -597,6 +631,60 @@ func (r *recipientEvidence) Forget(operationID string) {
...
+	for operationID, expectation := range r.operations {
+		if !expectation.deadline.IsZero() {
+			// It knows when its operation ends, so nothing else may release it.
+			if !expectation.deadline.Before(r.clock()) {
+				continue
+			}
+		} else if expectation.registeredAt.IsZero() || expectation.registeredAt.After(cutoff) {
+			continue
+		}
+		delete(r.operations, operationID)
```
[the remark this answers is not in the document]
flywindy: Confirmed and fixed in `3f45afe`. Took your first instruction; declined the second, and the reason is the more interesting half.

**Returning the finalized IDs** — done. `failureLedger.Expire` now returns `[]string` of the operations it actually retired, and the sweep forgets exactly those. As you note, this costs no lock ordering: the IDs come back after `Expire` has released the ledger lock, so the evidence lock is still taken alone. That was my whole reason for avoiding a ledger-driven forget in the first two attempts, and it turns out not to have been a reason at all.

**Not using the sweep timestamp for the deadline comparison** — because there is no longer a deadline comparison to make. The per-entry deadline is gone, along with `registeredAt`, the injected clock, the optional `Deadline` config field and the retention window. Once the ledger reports what it retired, every time-based approximation of that set is redundant.

Worth recording why the cheaper fix I reached for first does not work. I tried gating the evidence sweep on whether `Expire` had been truncated by `expireBatch` — but `Expire` also skips **claimed** operations, and that is not truncation: the loop completes normally and returns `nil`. No flag derived from "did the sweep run to completion" can distinguish it, and that path is the ordinary one, not a configuration edge. The reconciler is mid-verification and about to call `Finalize`.

So this is the third bound on the same map, and the first two were both clocks pretending to know what the ledger did. The net change removes more than it adds.

The capacity ceiling stays: forgetting only what the ledger retired means a ledger that stops retiring stops bounding the map.

Two tests, both failing on `4d81aaf`:
`TestRunSoakFailureExpiry_KeepsEvidenceTheLedgerBatchDeferred` and
`TestRunSoakFailureExpiry_KeepsEvidenceForAClaimedOperation`.

`make lint` clean, full `tools/loadgen` suite green under `-race` (322s).

---

tools/loadgen/soak_failure.go
flywindy: You were right, and the claimed-operation half is the part I would not have found on my own. Fixed in `3f45afe`.

I had reached for a cheaper version first — gate the evidence sweep on whether `Expire` was truncated by `expireBatch`. That covers your batch case and completely misses the claimed one, because skipping a claimed operation is not truncation: the loop completes normally and returns `nil`. So no flag derived from "did the sweep run to completion" can tell the two apart, and the claimed path is the ordinary one — the reconciler is mid-verification and about to call `Finalize` against exactly the evidence I was deleting.

`failureLedger.Expire` now returns the IDs it actually retired and the sweep forgets exactly those. As you said, that does not nest the locks: the IDs come back after `Expire` has released the ledger lock.

That also lets the change shrink. `registeredAt`, the injected clock, the per-entry deadline, the optional `Deadline` config field and the retention parameter are all gone, with the guards that existed to make them safe — three attempts at this bound, and the first two were clocks pretending to know what the ledger did.

The capacity ceiling stays, since forgetting only what the ledger retired means a ledger that stops retiring stops bounding the map.

Both shapes now have a test that fails on `4d81aaf`:
`TestRunSoakFailureExpiry_KeepsEvidenceTheLedgerBatchDeferred` and
`TestRunSoakFailureExpiry_KeepsEvidenceForAClaimedOperation`.

---

julianshen (commented):
tGD inline review

tools/loadgen/failure_recipient.go
```diff
@@ -441,6 +448,11 @@ func (r *recipientEvidence) ExpectDelivery(config *recipientExpectationConfig) e
 	if _, exists := r.operations[config.OperationID]; exists {
 		return fmt.Errorf("recipient expectation %q already exists", config.OperationID)
 	}
+	if r.capacity > 0 && len(r.operations) >= r.capacity {
```
julianshen, line 451: _🎯 concurrency_ | _🟠 Warning_ | _`distributed-system`, `performance-and-concurrency` (2 rules)_

**Expiry can reject a valid expectation at ledger capacity.**

The evidence limit equals the ledger limit, but expiry releases ledger capacity before `ForgetAll` releases the corresponding evidence. Concrete interleaving: both structures contain N operations; `Expire` finalizes one or more operations and unlocks the ledger; a workload goroutine occupies a newly freed ledger slot before the sweeper calls `ForgetAll`; `ExpectDelivery` still sees N entries and rejects the valid expectation. The operation then lacks recipient evidence or must be abandoned, producing false/unverified failure results during ordinary deadline pressure. Coordinate admission with expiry cleanup, or retain bounded headroom of at least one ledger capacity (2N total, since an unbounded sweep can finalize all N before cleanup), and add a synchronized regression test for this interleaving.

<details>

- **`performance-and-concurrency`**:451: At capacity, Expire removes a batch from the ledger and unlocks it before ForgetAll removes the corresponding evidence. A waiting workload goroutine can therefore acquire the ledger lock, successfully start a replacement operation, and then hit this still-full evidence map, causing ExpectDelivery to fail despite available ledger capacity. Mutex handoff makes this reachable under sustained contention. Coordinate ledger admission with evidence cleanup, or provide bounded handoff headroom sufficient for an expiry batch so newly admitted operations are not rejected during this window.

</details>

<details>

```
In `tools/loadgen/failure_recipient.go` around line 451: The evidence limit equals the ledger limit, but expiry releases ledger capacity before `ForgetAll` releases the corresponding evidence. Concrete interleaving: both structures contain N operations; `Expire` finalizes one or more operations and unlocks the ledger; a workload goroutine occupies a newly freed ledger slot before the sweeper calls `ForgetAll`; `ExpectDelivery` still sees N entries and rejects the valid expectation. The operation then lacks recipient evidence or must be abandoned, producing false/unverified failure results during ordinary deadline pressure. Coordinate admission with expiry cleanup, or retain bounded headroom of at least one ledger capacity (2N total, since an unbounded sweep can finalize all N before cleanup), and add a synchronized regression test for this interleaving.

```

</details>

julianshen, line 451: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Equal capacities can reject operations the ledger already admitted**

The evidence limit equals the ledger limit, but ledger expiry releases its lock and capacity before `runSoakFailureExpiry` calls `ForgetAll`. At full capacity, a concurrent start can therefore occupy the newly freed ledger slot while the expired expectation still occupies this map; this check then rejects the new expectation even though the ledger admitted its operation. That leaves admission and recipient tracking inconsistent (or causes a spurious operation failure, depending on the caller's error handling). Coordinate cleanup/admission atomically or give the evidence map bounded headroom (for example, enough for both one stale and one active ledger generation), and add a concurrent boundary test.

<details>

```
In `tools/loadgen/failure_recipient.go` around line 451: The evidence limit equals the ledger limit, but ledger expiry releases its lock and capacity before `runSoakFailureExpiry` calls `ForgetAll`. At full capacity, a concurrent start can therefore occupy the newly freed ledger slot while the expired expectation still occupies this map; this check then rejects the new expectation even though the ledger admitted its operation. That leaves admission and recipient tracking inconsistent (or causes a spurious operation failure, depending on the caller's error handling). Coordinate cleanup/admission atomically or give the evidence map bounded headroom (for example, enough for both one stale and one active ledger generation), and add a concurrent boundary test.

```

</details>

flywindy: Taking the headroom fix in `d4b9032`, but correcting the premise: the causal chain runs the other way round from the code.

`soak_failure.go` calls `t.recipient.ExpectDelivery` at line 350 and `t.ledger.Start` at line 367. Evidence is the **first** gate. So a refusal happens before the ledger is ever asked, and the two cannot end up disagreeing about an operation the ledger already admitted — `Start` never runs. "Leaves admission and recipient tracking inconsistent" does not describe this code.

What is real is the window itself, just with a different consequence. `Expire` frees ledger slots and `ForgetAll` frees the matching evidence afterwards, so in between the ledger has room the evidence does not, and a send arriving there is refused registration — an untracked send rather than an inconsistency.

I am also not accepting "during ordinary deadline pressure". The window only opens with the map at its ceiling. That ceiling is `LEDGER_CAPACITY` — 200,000 by default against a steady state of `SEND_RATE × RECONCILE_DEADLINE` ≈ 33,000, six times below it. Reaching it means expiry has already stopped working entirely, which is the emergency the ceiling exists for.

Fixed anyway, because it is cheaper to remove than to keep reasoning about, and adding a spurious untracked send to an already-degraded run is a poor trade. `recipientEvidenceCapacityFactor = 2`, exactly your suggested bound: an unbounded sweep retires at most the whole ledger before cleanup, so the lag can never exceed one capacity and the window is closed by construction.

Doubling does not weaken the ceiling. Its purpose is that a ledger which stops retiring cannot grow this map without limit, and 2N bounds that as well as N.

On the synchronized regression test: I did not add one. A test that has to interleave two goroutines between `Expire` returning and `ForgetAll` running would be asserting on a scheduling window rather than on behaviour, and would be the flaky kind. `TestFailureRecipientObserver_EvidenceCarriesHeadroomOverTheLedger` asserts the invariant that makes the window unreachable instead, and fails on `3f45afe`.

`make lint` clean, full suite green under `-race` (333s).

---

flywindy: Same finding as the `concurrency` rule's; fixed in `d4b9032`, reasoning on that thread.

Short version: taking the headroom (`recipientEvidenceCapacityFactor = 2`, exactly the bound you suggest), but the stated chain is inverted — `soak_failure.go:350` calls `ExpectDelivery` before `soak_failure.go:367` calls `ledger.Start`, so a refusal happens before the ledger is asked and "the ledger admitted its operation" never occurs. The real consequence is an untracked send, not inconsistent state, and it only arises with the map at its ceiling — 200,000 against a ~33,000 steady state, i.e. after expiry has already stopped working.

---

tools/loadgen/soak_failure.go
```diff
@@ -70,17 +88,27 @@ func runSoakFailureExpiry(
...
+			// claimed operations mid-verification. Forgetting exactly these
+			// keeps the evidence and the ledger in step, and taking them after
+			// Expire returns creates no ordering between the two locks.
+			expired, err := ledger.Expire(at)
+			if err != nil {
 				reportSoakFailureSweepError(
 					onError, fmt.Errorf("expire failure operations: %w", err))
 			}
+			evidence.ForgetAll(expired)
 			// Reclaiming on the same tick means a run that finalizes nothing
```
julianshen, line 85: _🎯 concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Expiry can miss a concurrently registered expectation.**

Ledger finalization and evidence registration are not atomic. If an operation is started in the ledger, its recipient registration is delayed past the deadline, and this sweep finalizes it before ExpectDelivery inserts the expectation, ForgetAll finds nothing; the subsequent insertion can never appear in another Expire result. Under repeated delayed sends, these orphaned entries accumulate to the new capacity, after which all recipient registrations are rejected. Add a post-insertion ledger-active check with rollback, or otherwise coordinate registration with ledger finalization so insertion-before-expiry is cleaned by ForgetAll and insertion-after-expiry is immediately discarded.

<details>

```
In `tools/loadgen/soak_failure.go` around line 102: Ledger finalization and evidence registration are not atomic. If an operation is started in the ledger, its recipient registration is delayed past the deadline, and this sweep finalizes it before ExpectDelivery inserts the expectation, ForgetAll finds nothing; the subsequent insertion can never appear in another Expire result. Under repeated delayed sends, these orphaned entries accumulate to the new capacity, after which all recipient registrations are rejected. Add a post-insertion ledger-active check with rollback, or otherwise coordinate registration with ledger finalization so insertion-before-expiry is cleaned by ForgetAll and insertion-after-expiry is immediately discarded.

```

</details>

julianshen (commented):
tGD inline review

flywindy:
CI `test` was red on `d4b9032` in `TestGenerator_PacedRunBeatsSingleTickerCeiling`, a test this branch does not touch (last change: `c1b7854`). It is not flaky — the assertion was unsatisfiable on a fast runner, and `cb35c3d` fixes the measurement.

The test took two dispatch counts over the same 200ms window at a 100k rps target and required `paced > serial * 1.3`. But the pacer releases `rate * elapsed` events and no more (`pacer.tick`, pinned by `TestPacer_SteadyStateTracksTargetRate`), so the paced count was clipped at exactly `100000 * 0.2 = 20000` — which is precisely what the runner reported. The serial ticker reached 15586, putting the required margin at 20261.8. Above the clip, so no result could pass:

```
Error:    "20000" is not greater than "20261.8"
Messages: batched pacer (20000) should out-dispatch the serial ticker (15586) at 100k rps
```

The margin shrinks as the host gets faster, which is why it passes locally (serial 7287 / paced 16007, 2.20x) and fails on CI. The earlier best-of-3 peak sampling made it worse, not better: the paced side was already pinned at its clip, so taking the peak only lifted the serial side.

`cb35c3d`:
- both paths are measured at 1M rps, out of reach of any host, so each count is that path's own ceiling rather than the shared target. Local, race detector: serial 3020 / paced 15917 (5.27x).
- a `require.Less(paced, target)` guard states the precondition, so a host that ever does reach the target fails saying so instead of failing on a bare ratio.
- `TestGenerator_PacedDispatchReportsAReachableTarget` pins the clipping itself at a rate every host reaches, so the failure mode is covered without depending on host speed. Verified red first: with the ceiling rate temporarily lowered to 5000, the guard fires with `"1000" is not less than "1000"`.

`make lint` clean, `go test -race ./tools/loadgen/` green (324s).

---

tools/loadgen/soak_failure.go
flywindy: Not reachable on this code — the two calls run in the opposite order from the one the finding assumes, and the gap it leaves is already closed by an explicit rollback.

`soakFailureTracker.Start` registers the evidence **first** and starts the ledger operation **second**:

- `soak_failure.go:350` — `t.recipient.ExpectDelivery(...)`
- `soak_failure.go:367` — `t.ledger.Start(...)`

So "an operation is started in the ledger, its recipient registration is delayed past the deadline, and this sweep finalizes it before `ExpectDelivery` inserts the expectation" has no window to occur in. At the only moment where the expectation is absent, the ledger does not hold the operation either — `Expire` iterates `l.active`, which the operation has not entered yet, so there is nothing for the sweep to finalize and nothing for `ForgetAll` to miss.

The reverse gap is the one that exists, and it is handled: evidence inserted, then `ledger.Start` refuses (capacity, WAL, validation). `soak_failure.go:384` rolls the expectation back on that error path — `t.recipient.evidence.Forget(pending.MessageID)`. That is the "post-insertion ledger-active check with rollback" being asked for, placed at the point where the ledger's answer is known rather than re-derived afterwards. Pinned by `TestSoakFailureTracker_ForgetsRecipientExpectationWhenLedgerRejectsStart`, which fills a `Capacity: 1` ledger, drives `tracker.Start` into the rejection, and asserts the expectation is gone.

Worth naming, since it is the third finding on this PR built on the same inverted premise (`#discussion_r3821511112`, `#discussion_r3821511114`): in this tracker, recipient evidence is always the first gate and the ledger the second. Any reasoning that starts "the ledger admitted the operation, then the evidence…" is describing code that is not there. The accumulation this predicts would also be visible: `loadgen_failure_recipient_expectations` climbing away from `loadgen_failure_inflight` is exactly what this PR adds the gauge to show.

---

flywindy:
Two items from the latest review, plus the title.

**Title** — fair, and fixed. It still described the deadline-based bound that `3f45afe` removed. Now `fix(loadgen): bound the recipient expectation map by the operations the ledger retires`.

**Nitpick: bare `err` returns in `Expire`** — declining, with the counts.

Both returns delegate to callees that already describe themselves, and both follow the convention every sibling site in the file uses:

- `ensureOpen()` is guarded at six sites — `failure_ledger.go` 561, 616, 679, 734, 846, 1234 — and all six return the error unchanged. It returns `fmt.Errorf("failure ledger is closed")`: a self-describing sentinel with nothing underneath it to lose.
- `finalizeLocked(...)` is called at three sites — 689, 809, 890 — and all three return it unchanged. It wraps both of its own failure modes already: `"persist finalized failure operation %q: %w"` and `"compact failure ledger: %w"`.

Line 874 is not the same kind of site, so it does not show the function to be internally inconsistent. It wraps `appendLocked`, which returns a raw file-write error carrying no context of its own — which is exactly when the guideline's `fmt.Errorf("short description: %w", err)` earns its keep.

The suggested text would also stutter: `runSoakFailureExpiry` already wraps this call with `"expire failure operations: %w"`, so the log line would read `expire failure operations: expire failure operations: failure ledger is closed`.

Changing two of the nine sites would make the file less consistent, not more. If the guideline should be applied literally here, it is a nine-site sweep of `failure_ledger.go` on its own — happy to do that as a separate change.

**Merge Risk note** — stale, and it does not hold on this head. It reads: "the current expiry behavior can still remove recipient evidence while operations remain active when expiry is batch-limited". That describes `4d81aaf`. On `cb35c`/`6be1ac3`, `Expire` appends to `finalized` only *after* `finalizeLocked` returns, and `finalizeLocked` ends with `delete(l.active, operation.ID)`. The batch limit `break`s and a claimed operation `continue`s — neither reaches the append. So every ID handed to `ForgetAll` names an operation already out of `l.active`; a batch-deferred one cannot be among them. `TestRunSoakFailureExpiry_KeepsEvidenceTheLedgerBatchDeferred` and `TestRunSoakFailureExpiry_KeepsEvidenceForAClaimedOperation` pin both shapes, and this is the same conclusion the review already reached and recorded as a learning on `#discussion_r3821456428`.

---

tools/loadgen/generator_test.go
```diff
@@ -337,51 +337,87 @@ func TestGenerator_PoolSaturationCountedAsError(t *testing.T) {
...
+	// The batched pacer releases rate*interval events per coarse tick and must
+	// out-dispatch it by a wide margin. A relative comparison (same process,
+	// back to back) cancels host-speed and race-detector overhead.
+	serial := dispatchPeak(t, 0, pacedCeilingRate, pacedCeilingWindow, pacedCeilingTrials)
+	paced := dispatchPeak(t, 5000, pacedCeilingRate, pacedCeilingWindow, pacedCeilingTrials)
 
-	serial := peakOf(0, 3)   // legacy one-per-tick ceiling
-	paced := peakOf(5000, 3) // batched pacer
 	require.Positive(t, serial)
+	require.Less(t, paced, pacedCeilingTarget(),
```
julianshen: _🧪 tests_ | _🟠 Warning_ | _`tgd-review`_

**The ceiling test fails on sufficiently fast hosts.**

This assertion requires the paced implementation not to reach its configured target. On a host capable of dispatching 1,000,000 events/s for this window, correct behavior can reach pacedCeilingTarget and fail the test; the failure message explicitly asks maintainers to raise the constant. That preserves the hardware-dependent flakiness the change is intended to remove. Use an adaptive rate or a deterministic clock/pacer test rather than assuming a fixed target is unreachable.

<details>

```
In `tools/loadgen/generator_test.go` around line 392: This assertion requires the paced implementation not to reach its configured target. On a host capable of dispatching 1,000,000 events/s for this window, correct behavior can reach pacedCeilingTarget and fail the test; the failure message explicitly asks maintainers to raise the constant. That preserves the hardware-dependent flakiness the change is intended to remove. Use an adaptive rate or a deterministic clock/pacer test rather than assuming a fixed target is unreachable.

```

</details>

julianshen (commented):
tGD inline review

tools/loadgen/generator_test.go
flywindy: You are right, and the part I want to concede plainly is the one I had talked myself out of: the guard turns a wrong measurement into a loud failure, but it does not make the test host-independent. Trading a fixed 100k for a fixed 1M widened the margin and left the shape of the problem exactly as it was. Fixed in `bcb8244`.

On why swapping constants could never close it. The two quantities being compared are both properties of the host, and they move in opposite directions:

| | serial plateau | paced ceiling |
|---|---|---|
| this box | ~36k/s | ~80k/s |
| CI runner | ~78k/s | ≥100k/s (it clipped, so this is a floor) |

A fixed rate has to be above the fastest host's serial plateau and below the slowest host's paced ceiling. 78k/s is already above 80k/s far enough that there is no number left to pick. So no constant works, and picking a bigger one only defers the same failure to a bigger machine — which is the objection.

Taking the adaptive option. The comparison now measures the serial ticker's own plateau on the host running the test and derives the rate from it, `pacedCeilingHeadroom = 20` times higher. The precondition that remains is no longer "1M is unreachable" but "this host does not out-dispatch its own serial ticker twentyfold" — a ratio between two numbers measured back to back in the same process, observed at 2.2–2.6x. It scales with the machine instead of aging with it.

`pacedCeilingRateFrom` is a pure function so the scaling is pinned by a deterministic table test rather than inferred from a timing run, and the ceiling test now logs `plateau`, the derived rate and both counts, so a future failure arrives carrying its own evidence instead of a bare ratio.

On the other half of your suggestion — a deterministic clock/pacer test — that layer already exists and carries the actual regression, which is worth saying because it bounds how much this timing test needs to prove:

- `pacer_test.go` drives `pacer.tick` on a synthetic clock, and `pacedDispatchRateWithTicks` takes an injectable tick channel, so the release arithmetic is tested with no real timer at all.
- `TestGenerator_PacedDispatchReportsAReachableTarget` runs at 5000 rps, where `minEmitInterval` forces a 2ms tick and therefore ten events per tick. A batcher that regressed to one-per-tick yields 100 instead of 1000 and fails on every host, with no timing sensitivity.

What only a real-timer comparison can show is that the runtime cannot deliver a tick per microsecond — an empirical property of the host, which is why that one test is measured rather than asserted.

Local, race detector: plateau 7437, derived rate 743700, serial 7517, paced 19695 — 2.6x, and 7.5x below the derived target. `make lint` clean, full `tools/loadgen` suite green under `-race` (316s).

---

tools/loadgen/failure_ledger.go
```diff
@@ -881,12 +886,13 @@ func (l *failureLedger) Expire(now time.Time) (int, error) {
 			}
 		}
 		result := failureOperationResult(operation)
+		operationID := operation.ID
 		if err := l.finalizeLocked(
 			operation, result, failureOperationFinalReason(operation, result), now,
 		); err != nil {
 			return finalized, err
```
flywindy, line 913: The exact-ID approach fixes the batch and claimed-operation cases, but this error path can still omit an operation that was already retired. `finalizeLocked` deletes it from `l.active` before running compaction; if compaction fails, `Expire` returns here before appending `operationID`, so `ForgetAll` never removes its evidence and no later sweep can report that ID. I reproduced this on `bcb8244` under `-race` (`active=false`, `evidence=1`). Please return the ID when finalization committed despite the compaction error, or move compaction outside per-operation finalization, and add a regression test.

tools/loadgen/generator_test.go
```diff
@@ -337,51 +337,109 @@ func TestGenerator_PoolSaturationCountedAsError(t *testing.T) {
...
+	rate := pacedCeilingRateFrom(plateau, pacedCeilingWindow, pacedCeilingHeadroom)
+
+	serial := dispatchPeak(t, 0, rate, pacedCeilingWindow, pacedCeilingTrials)
+	paced := dispatchPeak(t, 5000, rate, pacedCeilingWindow, pacedCeilingTrials)
 
-	serial := peakOf(0, 3)   // legacy one-per-tick ceiling
-	paced := peakOf(5000, 3) // batched pacer
 	require.Positive(t, serial)
+	t.Logf("plateau=%d derived rate=%d serial=%d paced=%d", plateau, rate, serial, paced)
+	require.Less(t, paced, int(float64(rate)*pacedCeilingWindow.Seconds()),
```
flywindy: This is still host/load dependent on `bcb8244`. `make test SERVICE=tools/loadgen` failed under `-race` with `plateau=400 derived rate=40000 serial=398 paced=8000`: the paced count exactly reached the 8000 target, so this `require.Less` failed. Scaling from the serial plateau does not guarantee that the paced target is unreachable; a larger fixed headroom only moves the boundary. Please keep the blocking assertion on the deterministic injected-clock/tick coverage, or make this real-timer measurement non-failing.

tools/loadgen/failure_ledger.go
flywindy: Confirmed and fixed in `9b3b4f2`. Your trace matches the code exactly, and following it turned up a second instance of the same defect.

Reproduced before changing anything — a ledger whose journal fails `Compact`, with `CompactEvery: 1`, and one operation past its deadline: `Expire` returned an error with `finalized` empty while `ledger.active` no longer held the operation. `finalizeLocked` does `delete(l.active, operation.ID)` and only then compacts, so the compaction error arrives on an operation already retired, and nothing is left in `l.active` for a later sweep to find.

Taking your first option. The second changes when compaction runs for all three `finalizeLocked` callers, which is a larger change than this defect warrants. `Expire` now decides by what left the ledger rather than by whether the call returned an error — which also states the invariant instead of inferring it:

```go
err := l.finalizeLocked(...)
if _, stillActive := l.active[operationID]; !stillActive {
    finalized = append(finalized, operationID)
}
if err != nil {
    return finalized, err
}
```

**The second one.** Your finding generalises to "an operation can leave `l.active` without its ID reaching `ForgetAll`", so I went looking for other paths with that shape. `AbandonUnsent` has it and released no evidence at all: it calls `ledger.Abandon` → `finalizeLocked`, which retires the operation out of `l.active`, putting it equally beyond the sweep. The expectation is registered at `soak_failure.go:350`, before the publish is attempted, so every locally-failed publish left one behind for the life of the run. Same reproduction: evidence still at 1 after a successful `AbandonUnsent`.

It now releases the expectation on the abandoned branch and on the already-finalized (`errFailureOperationNotActive`) branch, through a helper shared with the `Start` rollback. Deliberately not on the remaining error branch — there the operation may still be active and reconcilable, and its evidence is the only record of what was delivered, which is the reasoning from `#discussion_r3821345606`.

The reconciler path is clean, for completeness: `soak_failure.go:661` calls `recipient.Finalize` before `ledger.ObserveWithReason`, so it releases evidence regardless of what the ledger does next.

Two regression tests, both failing on `bcb8244`: `TestFailureLedger_ExpireReportsAnOperationRetiredBeforeTheFailure` and `TestSoakFailureTracker_AbandonUnsentReleasesTheRecipientExpectation`. `make lint` clean, full `tools/loadgen` suite green under `-race` (317s).

---

tools/loadgen/generator_test.go
flywindy: You are right, and the data you posted is what settles it. Fixed in `a5f7d92`, taking both halves of your suggestion.

`plateau=400 derived rate=40000 serial=398 paced=8000` is a ratio of 20 where an idle host shows 2.2, and it shows my premise was backwards. I assumed the two paths degrade together, so a ratio measured on the host would stay bounded. They do not: under load the serial ticker collapses — 2k/s, from ~36k/s idle — while the worker pool still manages 40k/s, because one is bound by timer delivery and the other by throughput. So the ratio *grows* with contention and has no upper bound to write a threshold against. Scaling from the plateau did not fix the shape of the problem any more than the constant did; it just made the boundary move with the load instead of with the machine.

That is three attempts, and the pattern in all three is that I kept trying to assert on a measurement whose value is a property of the host.

**The blocking assertion moves to the seam with no timer in it.** `pacedDispatchRateWithTicks` takes the tick channel, so `TestPacedDispatch_ReleasesTheWholeBatchOnOneTick` drives one synthetic tick through the real dispatch loop and asserts it releases the whole batch the rate owes — 200 events at 100k rps against the 2ms floor — where a one-per-tick regression releases exactly one. No timer, no host dependence. The tick channel is unbuffered so the send returns only once the loop has taken it, which makes the subsequent cancel deterministic rather than a race.

Verified by making the regression it exists to catch: changing the emit loop to release one event per tick fails it with `expected: 200, actual: 1`.

**The real-timer test is now a measurement.** It logs both counts, which is the part that is genuinely useful when tuning the pacer, and asserts only that each path dispatched at all — which cannot depend on host speed. I also dropped `TestGenerator_PacedDispatchReportsAReachableTarget`: it asserted the paced path reaches a 5000 rps target within 10%, and your load numbers show a host under contention need not, so it carried the same latent failure. The new deterministic test covers what it was there for.

`make lint` clean, full `tools/loadgen` suite green under `-race` (317s).

---

tools/loadgen/soak_failure.go
```diff
@@ -385,15 +411,30 @@ func (t *soakFailureTracker) AbandonUnsent(pending *soakPendingSend) error {
 	}
 	err := t.ledger.Abandon(pending.MessageID, failureResultNotSent, t.now().UTC())
 	if errors.Is(err, errFailureOperationNotActive) {
+		// Already finalized, so no sweep will ever report this ID again.
+		t.forgetRecipientExpectation(pending.MessageID)
 		t.countUntracked(failureUntrackedReasonAbandon)
 		return nil
 	}
 	if err != nil {
```
flywindy, line 406: `Abandon` can return a compaction error after `finalizeLocked` has already deleted the operation from `l.active`. This branch then retains the recipient expectation, and no expiry sweep can ever report that ID. I reproduced this on `9b3b4f2` (`active=false`, `evidence=1`). Please forget the expectation when the operation was retired, while still propagating the compaction error.

flywindy: Confirmed and fixed in `414edfd`. This is my own inconsistency, and it is worth naming as such.

The previous commit fixed `Expire` by reporting what left `l.active` instead of inferring it from the error — and then, four lines away, I left `AbandonUnsent` inferring it from the error. I even wrote the reasoning out: *"the operation may still be active and reconcilable"*. That is true for a failure **before** the commit (`appendLocked`, `ensureOpen`, the `not_sent` validation) and false for one **after** it, which is exactly the case you found. Having just replaced that inference in one function I reproduced it in the next.

Reproduced before changing anything, same shape as your run: a journal whose `Compact` fails with `CompactEvery: 1` left the operation out of `l.active` with `evidence` still at 1, and `AbandonUnsent` returned the error.

The fix puts the question where it can actually be answered. `failureLedger.Retired(operationID)` reports whether the ledger still holds the operation, and `AbandonUnsent` makes **one** decision from it rather than three from the error's shape:

```go
err := t.ledger.Abandon(pending.MessageID, failureResultNotSent, t.now().UTC())
if t.ledger.Retired(pending.MessageID) {
    t.forgetRecipientExpectation(pending.MessageID)
}
if errors.Is(err, errFailureOperationNotActive) { ... }
if err != nil {
    return fmt.Errorf("abandon unsent soak operation: %w", err)
}
```

An operation the ledger still holds is reconcilable and keeps its evidence; one it has let go releases it. The compaction error still propagates unchanged, so the caller's handling and the ledger's `invalidateLocked("wal")` are untouched. `Retired` takes only the ledger lock and is called after `Abandon` has released it, so the evidence lock is still taken alone.

I also checked the two remaining paths for the same shape rather than waiting to be told:

- `soak_roommember.go:556` also calls `Abandon`, but room mutations carry no recipient expectation, so there is nothing to strand.
- `ObserveWithReason` can likewise return `false, err` on an operation it retired, but the reconciler at `soak_failure.go:661` calls `recipient.Finalize` *before* it, so the evidence is already released regardless.

`TestSoakFailureTracker_AbandonUnsentReleasesTheExpectationItRetired` fails on `9b3b4f2`. `make lint` clean, full `tools/loadgen` suite green under `-race` (415s).

---

flywindy:
Same nitpick as `#issuecomment-5356366217`, and the position is unchanged — but the new justification does not hold, so it is worth correcting rather than restating.

> This preserves `errors.Is` behavior

`errors.Is` is not at risk either way. Returning `err` unchanged preserves the chain trivially; wrapping with `%w` preserves it too. Neither form can break `errors.Is`, so this is not a reason to prefer one.

The measured position from last time still stands: `ensureOpen` is guarded at six sites in `failure_ledger.go` and all six return unchanged, and it returns `fmt.Errorf("failure ledger is closed")` — a self-describing sentinel with nothing underneath to lose. `finalizeLocked` is called at three sites and all three return unchanged, and it wraps both of its own failure modes already. Changing two of nine makes the file less consistent, not more.

One thing has genuinely changed since that reply, and it cuts the other way for the second site only. `finalizeLocked` can now return an error on an operation it has already retired, so naming the operation there would say something the sweep-level wrapper (`"expire failure operations: %w"`) does not. That is a real, if small, diagnostic gain — worth doing as a nine-site sweep of the file rather than two lines inside a PR that is green and ready, since it is explicitly a non-blocker in your own Merge Risk note.

The PR description has also been brought back in line with the code: the pacer section described the `require.Less` unreachable-target design, which two rounds of review replaced with `TestPacedDispatch_ReleasesTheWholeBatchOnOneTick` on an injected tick channel.

---

flywindy:
## On the open item: "reconciliation appears not to be running"

The memory bound here looks right and it does stop the OOM. But I think the capacity arithmetic in the PR body is one term short, and the missing term explains the symptom exactly.

`soakFailureReconciler.Try` advances **one observer per claim**. With the recipient observer enabled a message needs two claims to finalize: the first verifies Cassandra history, and the recipient finalize is only reached on a *later* claim of the same operation (`soak_failure.go:626-666`). `TestFailureLedger_HistoryObservationReleasesClaimForAdmissionExpiry` pins that — observing history returns `finalized == false` and releases the claim.

```
demand   = SEND_RATE 55/s x 2 claims       = 110/s
capacity = READ_RATE 110/s x SHARE 0.75    =  82.5/s
```

So the backlog can only grow, every operation ages out at `RECONCILE_DEADLINE`, and `inflight` parks at `rate x deadline` = 33K — which is what you measured. A history probe that lands before the persist grace costs a third claim, so real demand is above 110/s.

Nothing catches this at startup. `validateSoakSearchReconcileCapacity` already models multi-step reconciliation (`reconcileStepsPerMessage = 2`), but it is only reached when `SearchObserverEnabled` (`soak_config.go:556`), and the search observer is refused at startup as "not usable yet". With recipient enabled and search disabled there is no reconcile capacity check at all.

**Question:** is this run using `SOAK_RECIPIENT_OBSERVER_ENABLED=true`? The heap profile suggests yes, and if so the numbers above are this run's numbers.

If that is right, this PR removes an aggravating factor — the hourly OOM restart replayed 33K unresolved operations that were all immediately due — but the deficit survives it.

**Check after merge, to confirm or kill the diagnosis:** if `loadgen_failure_inflight` still sits at `SEND_RATE x RECONCILE_DEADLINE` and `unverified` still dominates `loadgen_failure_operations_total`, this is the cause. If inflight falls and `good` becomes the majority, it was the restart loop and this diagnosis is wrong. Either way it is visible within ~30 minutes of the merge, well before the MongoDB failure campaign.

Two fixes, both separable from this PR:

- **Config, usable immediately:** raise `SOAK_READ_RATE` / `SOAK_RECONCILE_READ_SHARE` so capacity is at least `steps x SEND_RATE` with headroom for retried probes.
- **Code:** derive the step count from the observers actually enabled, and fail startup when capacity is short — rather than only when the search observer is on.

Worth stating explicitly for the MongoDB campaign: while the deficit holds, nearly every operation terminates `unverified`, which is the same result the ledger reports when MongoDB is genuinely unreachable. The two are indistinguishable on the dashboard, so the campaign would produce no usable evidence.

---

flywindy:
Confirmed, all three steps of it, and the answer to your question is yes — but from stronger evidence than the config.

**Was the recipient observer on?** It has to have been. `ExpectDelivery` is only reached under `if t.recipient != nil` (`soak_failure.go:349`), and the heap profile that opened this PR shows `ExpectDelivery` at 413.86MB and `incrementRecipientRoute` at 2472.41MB. Those allocations cannot exist with the observer off. So the arithmetic below is this run's arithmetic, not a hypothetical.

**One observer per claim** — verified in `Try`. Every branch ends in `return true, ...` after recording exactly one observation, and `Observe` re-enqueues rather than finalizing while `len(Observations) != len(Expected)`. So history and recipient cannot be resolved in the same claim.

**No capacity check covers this shape** — verified. `validateSoakSearchConfig` returns `nil` immediately when `!cfg.SearchObserverEnabled` (`soak_config.go:556`), and `validateSoakSearchReconcileCapacity` sits behind that early return. With recipient on and search off, nothing validates reconcile capacity at all. The constant is even named for the wrong reason: `reconcileStepsPerMessage = 2` is documented as "one step for the history observer, one for the search observer", when the recipient observer adds exactly the same second step and is the one actually enabled.

So the deficit is real and it is worse than 110/s. The history branch releases the claim and retries whenever the message has not landed in Cassandra yet (`soak_failure.go:735`), so every probe that arrives inside the persist grace costs an extra claim on top of the two.

```
capacity = READ_RATE 110 x SHARE 0.75              =  82.5/s
demand   = SEND_RATE 55 x 2 steps                  = 110/s   (floor, before history retries)
sustainable send rate at the current read config   =  41.25/s
```

`inflight` parking at exactly `55 x 600s = 33,000` is that deficit at steady state, and it also explains the `unverified` majority in `loadgen_failure_operations_total` without needing any service to be unhealthy.

Your point about the MongoDB campaign is the one that matters most and I had not drawn it: while the deficit holds, an operation that expires because the reconciler never reached it is recorded `unverified` — the same verdict as one that expires because MongoDB was genuinely unreachable. The campaign cannot distinguish them, so it would produce no usable evidence. That is a reason to fix the deficit before running it, not after.

One correction to the proposed remedies, for whoever picks this up. Raising `SOAK_RECONCILE_READ_SHARE` alone cannot close it: at `READ_RATE 110`, even a share of 1.0 supplies 110/s, which only just meets the floor and leaves nothing for history retries — and it does so by taking the entire read lane away from the read traffic the soak exists to generate. `SOAK_READ_RATE` has to rise, or `SOAK_SEND_RATE` has to fall to ~41. Raising `SOAK_RECONCILE_DEADLINE` is the one lever that actively hurts: it does not add capacity, it just parks a larger backlog, since `inflight` is `rate x deadline`.

Not folding any of this into this PR — it is green, and this is a configuration deficit plus a validation gap, neither of which is what the PR fixes. The check after merge you describe is exactly the right way to confirm or kill it, and it is already in the PR body as the open item.

---

tools/loadgen/soak_failure.go
```diff
@@ -70,17 +88,27 @@ func runSoakFailureExpiry(
...
 					onError, fmt.Errorf("expire failure operations: %w", err))
 			}
+			evidence.ForgetAll(expired)
 			// Reclaiming on the same tick means a run that finalizes nothing
 			// still bounds the journal; the finalize counter alone would let it
 			// grow until the next restart made recovery more expensive.
 			if err := ledger.MaybeCompact(at); err != nil {
 				reportSoakFailureSweepError(
 					onError, fmt.Errorf("compact failure journal: %w", err))
 			}
```
julianshen, line 91: _🏷️ observability_ | _🟠 Warning_ | _`distributed-system`_

**The expectation gauge freezes when the expiry sweep stalls**

The gauge is updated only after both `ledger.Expire` and `ledger.Compact` return. If the PVC or journal blocks in either call, this goroutine cannot refresh the metric. Concurrent sends can still add expectations before blocking on the ledger lock, growing the map toward its new capacity while Prometheus continues reporting the previous healthy value. Operators therefore cannot observe the failure mode this metric is intended to expose, and the first visible symptom may be rejected load generation at capacity. Update the gauge on expectation-map mutations, or expose it through an independently collected `GaugeFunc`, so journal/expiry stalls cannot prevent observation.

<details>

```
In `tools/loadgen/soak_failure.go` around line 108: The gauge is updated only after both `ledger.Expire` and `ledger.Compact` return. If the PVC or journal blocks in either call, this goroutine cannot refresh the metric. Concurrent sends can still add expectations before blocking on the ledger lock, growing the map toward its new capacity while Prometheus continues reporting the previous healthy value. Operators therefore cannot observe the failure mode this metric is intended to expose, and the first visible symptom may be rejected load generation at capacity. Update the gauge on expectation-map mutations, or expose it through an independently collected `GaugeFunc`, so journal/expiry stalls cannot prevent observation.

```

</details>

julianshen (commented):
tGD inline review

flywindy:
On the open item — "reconciliation appears not to be running". I think the arithmetic in the description is one term short, and the missing term explains the symptom exactly.

`soakFailureReconciler.Try` advances **one observer per claim**. With the recipient observer enabled, a message needs two claims to finalize: the first verifies Cassandra history and returns; the recipient finalize is only reached on a *later* claim of the same operation (`soak_failure.go:626-666`).

```
demand   = SEND_RATE 55/s x 2 claims     = 110/s
capacity = READ_RATE 110/s x SHARE 0.75  =  82.5/s
```

The backlog can then only grow, every operation ages out at `RECONCILE_DEADLINE`, and `inflight` parks at `rate x deadline` = 33K — which is what you measured. A history probe that lands before the persist grace costs a third claim, so real demand is above 110/s.

Nothing catches this at startup. `validateSoakSearchReconcileCapacity` already models multi-step reconciliation (`reconcileStepsPerMessage = 2`), but it is only called when `SearchObserverEnabled` (`soak_config.go:556`), and the search observer is refused at startup as "not usable yet". With recipient enabled and search disabled there is no capacity check at all.

Question: is the run that produced the heap profile using `SOAK_RECIPIENT_OBSERVER_ENABLED=true`? The 2.4GB in `incrementRecipientRoute` suggests yes — if so the numbers above are that run's numbers.

If this is right, the PR still stands on its own: it bounds the map, and it removes an aggravating factor, since each hourly restart replayed ~33K unresolved operations that were all immediately due. But the deficit survives it. Falsifiable prediction for the next run: `loadgen_failure_inflight` stays at ~`SEND_RATE x RECONCILE_DEADLINE` and `unverified` still dominates `loadgen_failure_operations_total{result}`. If inflight drops and `good` becomes the majority, this diagnosis is wrong.

Two fixes, both outside this PR:

- config: raise `SOAK_READ_RATE` / `SOAK_RECONCILE_READ_SHARE` so capacity covers `steps x SEND_RATE` plus headroom for retries;
- code: derive the step count from the observers actually enabled and fail startup when capacity is short, instead of only checking it on a path that cannot run.

This matters beyond the OOM: for the MongoDB failure test, a run where nearly every operation expires `unverified` looks identical with and without the fault, so the ledger would carry no evidence either way.

---

tools/loadgen/soak_failure.go
flywindy: Confirmed on the mechanism and fixed in `37c3230`. One part of the consequence is overstated, and it is worth separating because the fix is the same either way.

**The stale gauge is real.** `Set` was the last statement in the tick, after both `ledger.Expire` and `ledger.MaybeCompact`. A journal blocked in `appendLocked` holds that goroutine, so the gauge stops updating and keeps reporting the last healthy value — during exactly the condition it exists to expose. A frozen gauge is worse than a missing one: it reads as normal.

**The unbounded growth is not.** `ExpectDelivery` takes only `r.mu`, as you say, so a send can add an expectation while the sweep is stuck. But `soakFailureTracker.Start` calls `ledger.Start` immediately afterwards (`soak_failure.go:350` then `:367`), and that blocks on the ledger lock the stalled `Expire` is holding. So each concurrent sender contributes exactly one expectation and then parks. Growth is bounded by send-lane concurrency, not by the map's capacity, and "rejected load generation at capacity" would need concurrency on the order of `LEDGER_CAPACITY`.

Taking your `GaugeFunc` option. It is now collected when Prometheus scrapes:

```go
m.FailureRecipientExpectations = prometheus.NewGaugeFunc(opts, m.recipientExpectationCount)
```

The source is installed once the evidence exists — well after the registry is built — so it lives in an `atomic.Pointer[func() int]` that the scrape goroutine reads. Before it is installed the gauge reports zero, which is the honest answer for a run with no recipient observer rather than a collection failure. `evidence.Len` takes only the evidence lock and holds it for a map length, so a stalled ledger cannot block a scrape either.

Not taking the other option you offered — updating on expectation-map mutations. It puts metric knowledge inside `recipientEvidence` at four call sites, and it still leaves the value only as fresh as the last mutation.

One thing fell out of it: that `Set` was the sweep's only use of `Metrics`, so `soakFailureExpirySettings`, `soakFailureExpiryOption` and the variadic parameter were dead once it went, and are gone.

`TestMetrics_RecipientExpectationsReportsWithoutTheSweep` replaces the old test that asserted the sweep published it — that test encoded the coupling you flagged. It reads the gauge with no tick at all, before and after mutations, and checks the zero-before-installed case. `make lint` clean, full `tools/loadgen` suite green under `-race` (271s).

---

flywindy:
@coderabbitai review

---

flywindy:
CI red on `sast` on `37c3230`, failing on the base branch too — not this PR.

gosec reports one finding, twice:

```
[pkg/mongoutil/indexes.go:138] - G115 (CWE-190): integer overflow conversion int64 -> int32
    137: 	case int64:
  > 138: 		opts = opts.SetExpireAfterSeconds(int32(v))
```

That file arrived on `main` in `ccd3d40` (#333, merged 00:43Z) — after this PR's last green run at 14:35Z. This branch touches no file under `pkg/` at all (`git diff origin/main...HEAD -- pkg/` is empty), and gosec runs with `-exclude-dir=tools`, so nothing in this diff is even scanned by it.

Verified rather than assumed: running the same gosec invocation against a pristine `origin/main` worktree reproduces both findings, `Nosec: 0`, `Issues: 2`.

`govulncheck` and `semgrep` both passed in the same job; `lint` and `prewarm` are green on this head, and `test` / `test-integration` are still running. I will re-run `sast` once `main` recovers. Not touching `pkg/mongoutil` from here — it belongs to #333's author, and a `// #nosec G115` there is their call to make with the range argument that justifies it.

---

flywindy:
Rebased onto `d2fdb33` to pick up the base-branch gosec fix. The branch is now `b3f0c9e`; the 12 commits replayed with no conflicts, and nothing in the diff changed — this branch only touches `tools/loadgen` and `docs/specs/o11y`, while the base moved under `pkg/mongoutil`, `room-service` and `docs/load-testing`.

The `sast` failure is resolved on the base, by #338 rather than #334 — `pkg/mongoutil/indexes.go` now routes the captured TTL through a `ttlSeconds` helper instead of narrowing `int64` inline. Verified with the same gosec invocation CI uses, not assumed:

- pristine `origin/main` @ `d2fdb33`, `pkg/...` — `Issues: 0`
- this branch after the rebase, full `./...` — 1520 files, 268955 lines, `Nosec: 76`, `Issues: 0`, exit 0

Previously that same command reported `Issues: 2` on `ccd3d40`.

Also green locally on the rebased head: `make lint` 0 issues, full `tools/loadgen` suite under `-race` (269s).

Every SHA changed, so CodeRabbit's incremental state no longer matches and automatic reviews are still paused on this branch — triggering one below.

---

flywindy:
@coderabbitai review

---

yenta (approved):

PR #13, opened 2026-07-14
message-worker: cut per-reply DB work on the thread-reply hot path

hmchangw:
## Context

Investigating high MongoDB CPU under the **thread `max-rps`** load test. This PR removes work that `message-worker` does on **every subsequent thread reply**. It&#39;s three focused reductions on the thread-reply hot path.

## What changed

**1. `thread_subscriptions` writes: 3 → 1 per reply**
- Drop the parent-author subscription re-upsert on subsequent replies — it&#39;s created once, on the first reply — along with its owner-site `FindUserByID` read.
- Fold the replier&#39;s own `lastSeenAt` `$max` into the replier subscription upsert via a new `UpsertThreadSubscriptionAdvancingLastSeen` (`$setOnInsert` + `$max` in one write, non-conflicting because `lastSeenAt` is owned solely by `$max`).
- The standalone `AdvanceThreadSubscriptionLastSeen` now runs only on paths that write no replier sub (migration, self-reply, system message).

**2. Thread-room resolution: 1 round trip (both first and subsequent replies)**
- Replace the create-first pattern (`CreateThreadRoom` → duplicate-key → `GetThreadRoomByParentMessageID`) with an upserting `EnsureThreadRoom` (`FindOneAndUpdate` with `$setOnInsert` + `ReturnDocument:After`).
- One round trip, no failed unique-index insert on the hot path; the caller distinguishes first-vs-subsequent by comparing the returned `_id` to the candidate&#39;s. A rare concurrent-first-reply dup-key is resolved with a single read.

**3. Stop re-stamping the parent&#39;s `thread_room_id` on every subsequent reply**
- It&#39;s immutable and already stamped once, on the first reply (`handleFirstThreadReply`). Removes one Cassandra write per subsequent reply.

## Trade-offs (intentional)
Narrow crash windows, not worth a per-reply write to self-heal: if a first reply creates the room but crashes before writing the parent subscription / parent stamp, those stay unset for that thread. Same class of trade-off across all three changes.

## Important finding — this does not move max RPS on its own
Profiling during the load test showed the thread `max-rps` ceiling (~500–600, with a growing consumer backlog and E1/E2 P95 creeping over SLO) is bounded by **worker concurrency, not DB CPU**: `MAX_WORKERS` and the Mongo connection pool both default to **100**, while every resource sits at **≤70%**. `drain rate ≈ MaxWorkers / per-msg-latency ≈ 100/200ms ≈ 500/s`, so the backlog grows at ~600 offered.

[...]

PR #331, opened 2026-08-20
message-worker: cut per-reply DB work on the thread-reply hot path

hmchangw:
Replaces #13, which could not be made mergeable on its own branch — see **Why this replaces #13** below.

## Context

Investigating high MongoDB CPU under the **thread `max-rps`** load test. This PR removes work that `message-worker` does on **every subsequent thread reply**. It's three focused reductions on the thread-reply hot path.

## What changed

**1. `thread_subscriptions` writes: 3 → 1 per reply**
- Drop the parent-author subscription re-upsert on subsequent replies — it's created once, on the first reply — along with its owner-site lookup.
- Fold the replier's own `lastSeenAt` `$max` into the replier subscription upsert via a new `UpsertThreadSubscriptionAdvancingLastSeen` (`$setOnInsert` + `$max` in one write, non-conflicting because `lastSeenAt` is owned solely by `$max`).
- The standalone `AdvanceThreadSubscriptionLastSeen` now runs only on paths that write no replier sub (migration, self-reply, system message).
- That upsert filters on `(threadRoomId, userAccount)`, which carries a unique index, so two concurrent replies by the same account in one thread can both miss and both attempt the insert. The loser's `E11000` is recovered by replaying the `$max` alone rather than NAKing the reply. The recovery is deliberately narrow: if the replayed `$max` matches nothing, the duplicate came from a different unique key (an `_id` collision, say) and the original error is returned rather than silently dropping the write.

**2. Thread-room resolution: 1 round trip (both first and subsequent replies)**
- Replace the create-first pattern (`CreateThreadRoom` → duplicate-key → `GetThreadRoomByParentMessageID`) with an upserting `EnsureThreadRoom` (`FindOneAndUpdate` with `$setOnInsert` + `ReturnDocument:After`).
- One round trip, no failed unique-index insert on the hot path; the caller distinguishes first-vs-subsequent by comparing the returned `_id` to the candidate's. A rare concurrent-first-reply dup-key is resolved with a single read.

**3. Stop re-stamping the parent's `thread_room_id` on every subsequent reply**
- It's immutable and already stamped once, on the first reply (`handleFirstThreadReply`). Removes one Cassandra write per subsequent reply.

## Trade-offs (intentional)
Narrow crash windows, not worth a per-reply write to self-heal: if a first reply creates the room but crashes before writing the parent subscription / parent stamp, those stay unset for that thread. Same class of trade-off across all three changes.

## Why this replaces #13

#13's head branch and `main` have **unrelated git histories** (different root commits), so `git merge` refuses outright and GitHub reports the PR as `dirty` with no resolvable conflict set. Since a PR's head branch cannot be changed after it is opened, the change was re-applied onto current `main` on a fresh branch and #13 closed in favour of this PR. The content is #13's, rebuilt on a branch that fast-forwards from `main`.

### Conflicts resolved during the re-apply

`main` had, since #13's base, added a thread-unread fanout (`AddThreadUnread` / `fanOutThreadUnread` and its `followers` audience) and switched the sender lookup from `FindUserByID` to `FindUserByAccount`. Both touch the same functions this PR reshapes:

- `handleThreadRoomAndSubscriptions` / `handleSubsequentThreadReply` now return `(threadRoomID, followers, replierLastSeenAdvanced, error)` — `main`'s unread-fanout audience and this PR's folded-`lastSeen` flag both survive.
- `handleSubsequentThreadReply` takes the room resolved by `EnsureThreadRoom` (no second fetch) while keeping `main`'s pre-reply `followers` snapshot and the parent-author append that repairs legacy `thread_rooms`.
- Sender lookups stay on `main`'s `FindUserByAccount`; the parent owner-site lookup on the subsequent-reply path is gone along with the parent re-upsert.
- `store.go` keeps both interface additions: `AddThreadUnread` and `UpsertThreadSubscriptionAdvancingLastSeen`.
- Two `main`-only tests that #13's diff never touched (`..._NoThreadUnreadFanout`, `..._LegacyThreadRoom_ParentAuthorGetsUnread`) were updated for the new store API and the dropped re-stamp.

### Follow-up commits on this branch

- `7a219a1` — the concurrency recovery described above, plus: extracted the duplicated replier-subscription write into `writeReplierThreadSub`; dropped `UpsertThreadSubscription` from `ThreadStore` (no `message-worker` caller left after the combined write replaced it — `inbox-worker`'s same-named method is a separate interface and is untouched); made the new integration subtests independent of each other.
- `e1474a5` — narrowed the duplicate-key recovery to the race it was written for. It had been swallowing *any* duplicate key, so an `_id` collision took the same path, matched nothing on the replayed `$max`, and dropped the write with no error.

## Important finding — this does not move max RPS on its own

Profiling during the load test showed the thread `max-rps` ceiling (~500–600, with a growing consumer backlog and E1/E2 P95 creeping over SLO) is bounded by **worker concurrency, not DB CPU**: `MAX_WORKERS` and the Mongo connection pool both default to **100**, while every resource sits at **≤70%**. `drain rate ≈ MaxWorkers / per-msg-latency ≈ 100/200ms ≈ 500/s`, so the backlog grows at ~600 offered.

These changes lower per-reply Mongo/Cassandra **load**, which raises the DB ceiling **once concurrency is unblocked** — but the lever that actually moves the number is raising `MAX_WORKERS` and `maxPoolSize` together (they must move together, or the pool re-caps at 100). That's a config change, tracked separately.

## Scope & testing

- Contained to `message-worker` (a `MESSAGES-CANONICAL` JetStream consumer — not a `chat.user.*` client-facing handler, so no `docs/client-api.md` change).
- `make test` (full suite, race), `make lint`, `make sast-gosec`: green locally.
- Integration coverage added for `EnsureThreadRoom`, the combined upsert, and an 8-way concurrent-upsert race asserting no call errors, exactly one subscription survives, and `lastSeenAt` lands on the newest time. These cannot run in the authoring environment (no Docker daemon), so CI is the check on them — and it is what caught the `e1474a5` bug.
- `make generate` produces no diff, so the checked-in mocks match the interface.

https://claude.ai/code/session_013HBsHZQ71jNjjtNcEkRQbq

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when creating and preserving thread rooms, including concurrent activity.
  - Ensured reply subscriptions are created or updated consistently.
  - Prevented last-seen timestamps from moving backward.
  - Improved handling of subsequent replies, including replies with missing or resolved parents.
  - Reduced unnecessary unread and subscription updates for migration replies.
  - Preserved remote inbox notifications and parent-author unread updates where applicable.

- **Tests**
  - Expanded coverage for thread creation, subscription updates, concurrency, reply handling, and message history behavior.

### DIFF
--- a/message-worker/handler.go
+++ b/message-worker/handler.go
@@ -428,6 +428,21 @@ func (h *Handler) handleFirstThreadReply(ctx context.Context, msg *model.Message
 	return []string{parentSender.Account}, nil
 }
 
+// writeReplierThreadSub performs the replier's combined subscription upsert +
+// lastSeenAt advance on the subsequent-reply path, then publishes the cross-site
+// inbox copy when the replier's home site is remote. Callers keep their own
+// eligibility guard and set replierLastSeenAdvanced on success.
+func (h *Handler) writeReplierThreadSub(ctx context.Context, msg *model.Message, eventSiteID, threadRoomID string, replier *model.User, now time.Time) error {
+	replierSub := h.buildThreadSubscription(msg, threadRoomID, msg.UserID, msg.UserAccount, eventSiteID, now)
+	if err := h.threadStore.UpsertThreadSubscriptionAdvancingLastSeen(ctx, replierSub, msg.CreatedAt); err != nil {
+		return fmt.Errorf("upsert replier thread subscription: %w", err)
+	}
+	if err := h.publishThreadSubInboxIfRemote(ctx, replierSub, replier.SiteID, msg.ID); err != nil {
+		return fmt.Errorf("publish replier thread subscription inbox: %w", err)
+	}
+	return nil
+}
+
 // handleSubsequentThreadReply runs when the thread room already existed (resolved by the
 // caller's EnsureThreadRoom and passed in as existingRoom, so no second fetch is needed).
 // It upserts only the replier's subscription — folding the replier's own lastSeenAt
@@ -459,14 +474,10 @@ func (h *Handler) handleSubsequentThreadReply(ctx context.Context, msg *model.Me
 	switch {
 	case err == nil:
 		if !isMigration && replier != nil && msg.UserID != parentSender.ID {
-			replierSub := h.buildThreadSubscription(msg, existingRoom.ID, msg.UserID, msg.UserAccount, eventSiteID, now)
-			if err := h.threadStore.UpsertThreadSubscriptionAdvancingLastSeen(ctx, replierSub, msg.CreatedAt); err != nil {
-				return "", nil, false, fmt.Errorf("upsert replier thread subscription: %w", err)
+			if err := h.writeReplierThreadSub(ctx, msg, eventSiteID, existingRoom.ID, replier, now); err != nil {
+				return "", nil, false, err
 			}
 			replierLastSeenAdvanced = true
-			if err := h.publishThreadSubInboxIfRemote(ctx, replierSub, replier.SiteID, msg.ID); err != nil {
-				return "", nil, false, fmt.Errorf("publish replier thread subscription inbox: %w", err)
-			}
 		}
 	case errors.Is(err, errMessageNotFound):
 		parentFound = false
@@ -475,14 +486,10 @@ func (h *Handler) handleSubsequentThreadReply(ctx context.Context, msg *model.Me
 			"replyID", msg.ID,
 			"request_id", natsutil.RequestIDFromContext(ctx))
 		if !isMigration && replier != nil {
-			replierSub := h.buildThreadSubscription(msg, existingRoom.ID, msg.UserID, msg.UserAccount, eventSiteID, now)
-			if err := h.threadStore.UpsertThreadSubscriptionAdvancingLastSeen(ctx, replierSub, msg.CreatedAt); err != nil {
-				return "", nil, false, fmt.Errorf("upsert replier thread subscription: %w", err)
+			if err := h.writeReplierThreadSub(ctx, msg, eventSiteID, existingRoom.ID, replier, now); err != nil {
+				return "", nil, false, err
 			}
 			replierLastSeenAdvanced = true
-			if err := h.publishThreadSubInboxIfRemote(ctx, replierSub, replier.SiteID, msg.ID); err != nil {
-				return "", nil, false, fmt.Errorf("publish replier thread subscription inbox: %w", err)
-			}
 		}
 	default:
 		return "", nil, false, fmt.Errorf("get parent message sender: %w", err)
--- a/message-worker/handler.go
+++ b/message-worker/handler.go
@@ -158,17 +158,21 @@ func (h *Handler) processMessage(ctx context.Context, data []byte, isMigration b
 
 		// Resolve (or create) the thread room first so we have the threadRoomID
 		// before persisting the message to Cassandra.
-		threadRoomID, followers, err := h.handleThreadRoomAndSubscriptions(ctx, &evt.Message, evt.SiteID, user, isMigration)
+		threadRoomID, followers, replierLastSeenAdvanced, err := h.handleThreadRoomAndSubscriptions(ctx, &evt.Message, evt.SiteID, user, isMigration)
 		if err != nil {
 			return fmt.Errorf("handle thread room and subscriptions: %w", err)
 		}
 		// Replying implies the replier read up to their own reply: advance their thread
-		// lastSeenAt so the read-floor doesn't count them (#396). Best-effort. Runs on
-		// migration too ($max only moves forward → lands on the replier's last reply).
-		if err := h.threadStore.AdvanceThreadSubscriptionLastSeen(ctx, threadRoomID, evt.Message.UserAccount, evt.Message.CreatedAt); err != nil {
-			slog.WarnContext(ctx, "advance replier thread lastSeenAt failed",
-				"error", err, "thread_room_id", threadRoomID, "account", evt.Message.UserAccount,
-				"request_id", natsutil.RequestIDFromContext(ctx))
+		// lastSeenAt so the read-floor doesn't count them (#396). The hot subsequent-reply
+		// path folds this into the replier's subscription upsert (one write instead of two),
+		// reporting replierLastSeenAdvanced=true; this standalone $max covers the paths that
+		// write no replier subscription (migration, self-reply, system message). Best-effort.
+		if !replierLastSeenAdvanced {
+			if err := h.threadStore.AdvanceThreadSubscriptionLastSeen(ctx, threadRoomID, evt.Message.UserAccount, evt.Message.CreatedAt); err != nil {
+				slog.WarnContext(ctx, "advance replier thread lastSeenAt failed",
+					"error", err, "thread_room_id", threadRoomID, "account", evt.Message.UserAccount,
+					"request_id", natsutil.RequestIDFromContext(ctx))
+			}
 		}
 		mentionedAccounts, err := h.markThreadMentions(ctx, &evt.Message, threadRoomID, evt.SiteID, isMigration)
 		if err != nil {
@@ -302,17 +306,20 @@ func debugFlowPersisted(ctx context.Context, messageID string, thread bool) {
 		"message_id", messageID, "thread", thread)
 }
 
-// handleThreadRoomAndSubscriptions creates the ThreadRoom on first reply and
-// inserts ThreadSubscriptions for the parent author and replier. On subsequent
-// replies it upserts both subscriptions and bumps the room's last-message pointer.
-// It returns the threadRoomID so the caller can pass it to SaveThreadMessage,
-// plus the thread's pre-existing followers (the accounts fanOutThreadUnread
-// should mark unread — the parent author alone on a first reply, or the room's
-// established ReplyAccounts on a subsequent one).
+// handleThreadRoomAndSubscriptions resolves the ThreadRoom via a single upserting
+// EnsureThreadRoom call (one round trip, no failed insert on the hot path). On the first
+// reply (EnsureThreadRoom created the room) it inserts ThreadSubscriptions for the parent
+// author and replier; on subsequent replies it upserts only the replier's subscription
+// (the parent author's was created on the first reply) and bumps the last-message pointer.
+// It returns the threadRoomID so the caller can pass it to SaveThreadMessage, the thread's
+// pre-existing followers (the accounts fanOutThreadUnread should mark unread — the parent
+// author alone on a first reply, or the room's established ReplyAccounts on a subsequent
+// one), and a bool reporting whether the replier's lastSeenAt was already advanced as part
+// of that subscription write (so the caller can skip the standalone $max).
 //
 // `replier` may be nil for system messages with no real user (rare in thread
 // paths); subscriptions for the replier are skipped in that case.
-func (h *Handler) handleThreadRoomAndSubscriptions(ctx context.Context, msg *model.Message, eventSiteID string, replier *model.User, isMigration bool) (string, []string, error) {
+func (h *Handler) handleThreadRoomAndSubscriptions(ctx context.Context, msg *model.Message, eventSiteID string, replier *model.User, isMigration bool) (string, []string, bool, error) {
 	now := msg.CreatedAt
 
 	var parentCreatedAt time.Time
@@ -332,16 +339,17 @@ func (h *Handler) handleThreadRoomAndSubscriptions(ctx context.Context, msg *mod
 		UpdatedAt:             now,
 	}
 
-	err := h.threadStore.CreateThreadRoom(ctx, &threadRoom)
-	switch {
-	case err == nil:
-		followers, ferr := h.handleFirstThreadReply(ctx, msg, eventSiteID, threadRoom.ID, replier, now, isMigration)
-		return threadRoom.ID, followers, ferr
-	case errors.Is(err, errThreadRoomExists):
-		return h.handleSubsequentThreadReply(ctx, msg, eventSiteID, replier, now, isMigration)
-	default:
-		return "", nil, fmt.Errorf("create thread room: %w", err)
+	existingRoom, created, err := h.threadStore.EnsureThreadRoom(ctx, &threadRoom)
+	if err != nil {
+		return "", nil, false, fmt.Errorf("ensure thread room: %w", err)
 	}
+	if created {
+		// First reply (once per thread): it advances the replier's lastSeenAt via the standalone
+		// $max in the caller, so it reports replierLastSeenAdvanced=false.
+		followers, ferr := h.handleFirstThreadReply(ctx, msg, eventSiteID, existingRoom.ID, replier, now, isMigration)
+		return existingRoom.ID, followers, false, ferr
+	}
+	return h.handleSubsequentThreadReply(ctx, msg, eventSiteID, existingRoom, replier, now, isMigration)
 }
 
 // handleFirstThreadReply runs after the thread room has just been created.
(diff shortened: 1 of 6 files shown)


julianshen:
**No actionable comments.** ✅

### Related work

- [PR #13](https://github.com/hmchangw/newchat/pull/13) — message-worker: cut per-reply DB work on the thread-reply hot path (closed)

<details>

* `message-worker/handler.go`
* `message-worker/mock_store_test.go`
* `message-worker/store.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

message-worker/handler.go
```diff
@@ -508,32 +510,13 @@ func (h *Handler) handleSubsequentThreadReply(ctx context.Context, msg *model.Me
 		followers = append(followers, parentSender.Account)
```
julianshen, line 517: _🎯 cassandra-correctness_ | _🔴 Blocking_ | _`cassandra`_

**Retries can permanently leave the Cassandra parent unstamped**

The affected mutation is `UpdateParentMessageThreadRoomID` on the Cassandra parent-message row. If `EnsureThreadRoom` creates the Mongo room and the worker crashes or returns an error before the first-reply stamp completes, redelivery observes `created=false` and enters this subsequent-reply path. Because this path now skips the idempotent stamp, the retry can save the reply successfully while the parent permanently lacks `thread_room_id`, breaking readers that discover the thread through the parent row. Restore the retry-time stamp when the parent and its clustering timestamp are available; the existing mutation is already designed for this recovery path.

<details>

```
In `message-worker/handler.go` around line 510: The affected mutation is `UpdateParentMessageThreadRoomID` on the Cassandra parent-message row. If `EnsureThreadRoom` creates the Mongo room and the worker crashes or returns an error before the first-reply stamp completes, redelivery observes `created=false` and enters this subsequent-reply path. Because this path now skips the idempotent stamp, the retry can save the reply successfully while the parent permanently lacks `thread_room_id`, breaking readers that discover the thread through the parent row. Restore the retry-time stamp when the parent and its clustering timestamp are available; the existing mutation is already designed for this recovery path.

```

</details>

message-worker/handler.go
```diff
@@ -420,51 +428,44 @@ func (h *Handler) handleFirstThreadReply(ctx context.Context, msg *model.Message
...
-func (h *Handler) handleSubsequentThreadReply(ctx context.Context, msg *model.Message, eventSiteID string, replier *model.User, now time.Time, isMigration bool) (string, []string, error) {
-	existingRoom, err := h.threadStore.GetThreadRoomByParentMessageID(ctx, msg.ThreadParentMessageID)
-	if err != nil {
-		return "", nil, fmt.Errorf("get existing thread room: %w", err)
-	}
+// handleSubsequentThreadReply runs when the thread room already existed (resolved by the
+// caller's EnsureThreadRoom and passed in as existingRoom, so no second fetch is needed).
+// It upserts only the replier's subscription — folding the replier's own lastSeenAt
+// advance into that single write (#396) — then bumps the room's last-message pointer.
+// The parent author's subscription is intentionally NOT re-upserted here: it was
```
julianshen, line 450: _🎯 distributed-correctness_ | _🔴 Blocking_ | _`distributed-system`_

**A retried first reply leaves the thread permanently half-initialized**

After EnsureThreadRoom inserts the Mongo room, the process can crash, lose the successful response, or fail while inserting/publishing the parent subscription or stamping Cassandra. On redelivery, EnsureThreadRoom returns created=false, so this path permanently skips the parent subscription, its cross-site inbox publication, and UpdateParentMessageThreadRoomID; the reply can then persist and be acknowledged. This is also reachable during rolling deploys or concurrent redelivery. Restore the idempotent repair writes on subsequent replies, or persist an initialization marker and retry all side effects through an idempotent outbox before marking the room initialized.

<details>

```
In `message-worker/handler.go` around line 435: After EnsureThreadRoom inserts the Mongo room, the process can crash, lose the successful response, or fail while inserting/publishing the parent subscription or stamping Cassandra. On redelivery, EnsureThreadRoom returns created=false, so this path permanently skips the parent subscription, its cross-site inbox publication, and UpdateParentMessageThreadRoomID; the reply can then persist and be acknowledged. This is also reachable during rolling deploys or concurrent redelivery. Restore the idempotent repair writes on subsequent replies, or persist an initialization marker and retry all side effects through an idempotent outbox before marking the room initialized.

```

</details>

message-worker/handler.go
```diff
@@ -420,51 +428,44 @@ func (h *Handler) handleFirstThreadReply(ctx context.Context, msg *model.Message
...
-// this reply's unread mark. That combined list is the audience
-// fanOutThreadUnread should mark unread; it dedups, and the sender is
-// filtered out there regardless of whether they were already a follower.
-func (h *Handler) handleSubsequentThreadReply(ctx context.Context, msg *model.Message, eventSiteID string, replier *model.User, now time.Time, isMigration bool) (string, []string, error) {
-	existingRoom, err := h.threadStore.GetThreadRoomByParentMessageID(ctx, msg.ThreadParentMessageID)
-	if err != nil {
-		return "", nil, fmt.Errorf("get existing thread room: %w", err)
-	}
+// handleSubsequentThreadReply runs when the thread room already existed (resolved by the
+// caller's EnsureThreadRoom and passed in as existingRoom, so no second fetch is needed).
```
julianshen, line 447: _🏷️ atomicity_ | _🔴 Blocking_ | _`mongodb`_

**Room creation can permanently skip the parent subscription**

`EnsureThreadRoom` commits the `thread_rooms` document before `handleFirstThreadReply` creates the parent author's `thread_subscriptions` document. If the worker exits or a transient lookup/insert error occurs between those operations, redelivery finds the room with `created=false`; this changed subsequent-reply path only upserts the replier, so the parent subscription is never recovered. `AddThreadUnread` only updates existing subscriptions, meaning the parent can permanently miss unread state and remote subscription delivery. Restore the idempotent parent upsert on the subsequent path, or atomically initialize the room and required subscriptions with a durable initialization state/transaction.

<details>

```
In `message-worker/handler.go` around line 432: `EnsureThreadRoom` commits the `thread_rooms` document before `handleFirstThreadReply` creates the parent author's `thread_subscriptions` document. If the worker exits or a transient lookup/insert error occurs between those operations, redelivery finds the room with `created=false`; this changed subsequent-reply path only upserts the replier, so the parent subscription is never recovered. `AddThreadUnread` only updates existing subscriptions, meaning the parent can permanently miss unread state and remote subscription delivery. Restore the idempotent parent upsert on the subsequent path, or atomically initialize the room and required subscriptions with a durable initialization state/transaction.

```

</details>

message-worker/handler.go
```diff
@@ -420,51 +428,44 @@ func (h *Handler) handleFirstThreadReply(ctx context.Context, msg *model.Message
...
+// (replier.SiteID) drives cross-site inbox routing without an extra lookup.
+func (h *Handler) handleSubsequentThreadReply(ctx context.Context, msg *model.Message, eventSiteID string, existingRoom *model.ThreadRoom, replier *model.User, now time.Time, isMigration bool) (string, []string, bool, error) {
 	followers := existingRoom.ReplyAccounts
 
 	// Migrated replies: resolve the parent for replyAccounts, but skip all thread_subscription writes (collections owns them).
 	parentFound := true
+	replierLastSeenAdvanced := false
 	parentSender, err := h.store.GetMessageSender(ctx, msg.ThreadParentMessageID)
 	switch {
 	case err == nil:
```
julianshen, line 475: _🏷️ nats-redelivery_ | _🔴 Blocking_ | _`nats`_

**Redelivery skips the failed parent inbox publish**

A first delivery can create the room and then fail while publishing the parent subscription to the cross-site inbox, causing the JetStream message to be NAKed. On redelivery, EnsureThreadRoom returns created=false and this branch only upserts and publishes the replier subscription, so processing can succeed and ACK without retrying the failed parent publication. A worker crash between room creation and that publication has the same outcome, permanently leaving the remote parent without the thread subscription. Restore the idempotent parent upsert/inbox publication on the existing-room path, or persist an initialization/outbox marker and do not treat the room as initialized until that publication is durably delivered.

<details>

```
In `message-worker/handler.go` around line 460: A first delivery can create the room and then fail while publishing the parent subscription to the cross-site inbox, causing the JetStream message to be NAKed. On redelivery, EnsureThreadRoom returns created=false and this branch only upserts and publishes the replier subscription, so processing can succeed and ACK without retrying the failed parent publication. A worker crash between room creation and that publication has the same outcome, permanently leaving the remote parent without the thread subscription. Restore the idempotent parent upsert/inbox publication on the existing-room path, or persist an initialization/outbox marker and do not treat the room as initialized until that publication is durably delivered.

```

</details>

message-worker/handler.go
```diff
@@ -420,51 +428,44 @@ func (h *Handler) handleFirstThreadReply(ctx context.Context, msg *model.Message
...
-			parentSub := h.buildThreadSubscription(msg, existingRoom.ID, parentSender.ID, parentSender.Account, eventSiteID, now)
-			if err := h.threadStore.UpsertThreadSubscription(ctx, parentSub); err != nil {
-				return "", nil, fmt.Errorf("upsert parent author thread subscription: %w", err)
-			}
-			if parentOwnerSite != "" {
-				if err := h.publishThreadSubInboxIfRemote(ctx, parentSub, parentOwnerSite, msg.ID); err != nil {
-					return "", nil, fmt.Errorf("publish parent thread subscription inbox: %w", err)
-				}
+		if !isMigration && replier != nil && msg.UserID != parentSender.ID {
+			replierSub := h.buildThreadSubscription(msg, existingRoom.ID, msg.UserID, msg.UserAccount, eventSiteID, now)
```
julianshen: _🎯 concurrency_ | _🔴 Blocking_ | _`performance-and-concurrency`_

**Concurrent first replies can miss parent initialization.**

EnsureThreadRoom makes the new room visible before handleFirstThreadReply creates the parent subscription. If reply A creates the room and pauses, concurrent reply B observes created=false, enters this branch, and now upserts only B's subscription. B can reach AddThreadUnread before A creates the parent subscription, so its parent unread update matches nothing and is permanently lost even when both replies succeed. The same gap breaks redelivery: any transient error or crash after room insertion but before the first-reply subscriptions or parent thread_room_id stamp causes the retry to take this subsequent path and permanently skip those initialization writes. Restore the idempotent parent subscription/stamp writes on the existing-room path, or persist an initialization marker and require callers observing an uninitialized room to complete those writes before fan-out.

<details>

```
In `message-worker/handler.go` around line 462: EnsureThreadRoom makes the new room visible before handleFirstThreadReply creates the parent subscription. If reply A creates the room and pauses, concurrent reply B observes created=false, enters this branch, and now upserts only B's subscription. B can reach AddThreadUnread before A creates the parent subscription, so its parent unread update matches nothing and is permanently lost even when both replies succeed. The same gap breaks redelivery: any transient error or crash after room insertion but before the first-reply subscriptions or parent thread_room_id stamp causes the retry to take this subsequent path and permanently skip those initialization writes. Restore the idempotent parent subscription/stamp writes on the existing-room path, or persist an initialization marker and require callers observing an uninitialized room to complete those writes before fan-out.

```

</details>

message-worker/handler.go
julianshen, line 450: _🏷️ error-handling_ | _🔴 Blocking_ | _`silent-failure`_

**Redelivery skips failed first-reply subscription initialization**

EnsureThreadRoom durably creates the room before handleFirstThreadReply inserts the parent subscription and publishes its remote inbox event. If either later operation fails, the handler returns an error, but redelivery sees created=false and this path deliberately performs neither operation. The retry can then persist and acknowledge the reply while leaving the parent permanently unsubscribed or the remote inbox event permanently missing. Concurrent first replies expose the same incomplete-initialization window. Make initialization retryable by idempotently upserting and publishing the parent subscription on this path, or persist a completion marker and only skip initialization after all required side effects succeed.

<details>

```
In `message-worker/handler.go` around line 435: EnsureThreadRoom durably creates the room before handleFirstThreadReply inserts the parent subscription and publishes its remote inbox event. If either later operation fails, the handler returns an error, but redelivery sees created=false and this path deliberately performs neither operation. The retry can then persist and acknowledge the reply while leaving the parent permanently unsubscribed or the remote inbox event permanently missing. Concurrent first replies expose the same incomplete-initialization window. Make initialization retryable by idempotently upserting and publishing the parent subscription on this path, or persist a completion marker and only skip initialization after all required side effects succeed.

```

</details>

message-worker/handler.go
```diff
@@ -508,32 +510,13 @@ func (h *Handler) handleSubsequentThreadReply(ctx context.Context, msg *model.Me
...
-	default: // msg.ThreadParentMessageCreatedAt == nil
-		slog.ErrorContext(ctx, "subsequent thread reply: ThreadParentMessageCreatedAt is nil, parent thread_room_id stamp skipped",
-			"request_id", natsutil.RequestIDFromContext(ctx),
-			"replyID", msg.ID,
-			"parentMessageID", msg.ThreadParentMessageID,
-			"threadRoomID", existingRoom.ID,
-			"room_id", msg.RoomID,
-		)
-	}
+	// The parent message's thread_room_id is stamped once, on the first reply
```
julianshen, line 520: _🏷️ error-handling_ | _🔴 Blocking_ | _`silent-failure`_

**Redelivery masks a failed parent thread-room stamp**

UpdateParentMessageThreadRoomID can fail after the room has already been created, causing the first attempt to return an error. On redelivery, EnsureThreadRoom reports an existing room and this path now skips the stamp, allowing the retry to succeed while the parent Cassandra row remains permanently unstamped. Restore the idempotent re-stamp and continue propagating its write failure.

<details>

```
In `message-worker/handler.go` around line 513: UpdateParentMessageThreadRoomID can fail after the room has already been created, causing the first attempt to return an error. On redelivery, EnsureThreadRoom reports an existing room and this path now skips the stamp, allowing the retry to succeed while the parent Cassandra row remains permanently unstamped. Restore the idempotent re-stamp and continue propagating its write failure.

```

</details>

message-worker/handler.go
julianshen, line 475: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**Retries can permanently skip first-reply initialization**

EnsureThreadRoom persists the room before handleFirstThreadReply completes. If sender lookup, parent subscription insertion, inbox publication, replier insertion, or parent stamping then fails, redelivery sees created=false and enters this subsequent-reply path. This path now writes only the replier subscription and no longer re-stamps the parent, so the parent subscription, remote inbox event, or Cassandra thread_room_id can remain missing permanently. This is not limited to a process crash; any ordinary transient error after room creation triggers it. Preserve idempotent self-healing on subsequent replies, or record initialization completion and retry the missing first-reply work until complete.

<details>

```
In `message-worker/handler.go` around line 460: EnsureThreadRoom persists the room before handleFirstThreadReply completes. If sender lookup, parent subscription insertion, inbox publication, replier insertion, or parent stamping then fails, redelivery sees created=false and enters this subsequent-reply path. This path now writes only the replier subscription and no longer re-stamps the parent, so the parent subscription, remote inbox event, or Cassandra thread_room_id can remain missing permanently. This is not limited to a process crash; any ordinary transient error after room creation triggers it. Preserve idempotent self-healing on subsequent replies, or record initialization completion and retry the missing first-reply work until complete.

```

</details>

message-worker/integration_test.go
```diff
@@ -2110,69 +2106,64 @@ func TestAdvanceThreadSubscriptionLastSeen_OnlyAdvances(t *testing.T) {
 	require.NoError(t, store.AdvanceThreadSubscriptionLastSeen(ctx, "no-room", "nobody", t2))
```
julianshen, line 2115: _🧪 tests_ | _🟠 Warning_ | _`tgd-review`_

**Unrelated integration coverage was removed**

Adding the combined-upsert test also deletes TestThreadStoreMongo_GetHistorySharedSince and TestCassandraStore_SaveThreadMessage_TShowWritesAllTables. Neither behavior is changed by this PR, and the latter verified persistence across three Cassandra tables. Restore these tests alongside the new test so this optimization does not silently reduce coverage of unrelated functionality.

<details>

```
In `message-worker/integration_test.go` around line 2106: Adding the combined-upsert test also deletes TestThreadStoreMongo_GetHistorySharedSince and TestCassandraStore_SaveThreadMessage_TShowWritesAllTables. Neither behavior is changed by this PR, and the latter verified persistence across three Cassandra tables. Restore these tests alongside the new test so this optimization does not silently reduce coverage of unrelated functionality.

```

</details>

julianshen (commented):
tGD inline review

message-worker/store_mongo.go
```diff
@@ -171,6 +169,35 @@ func (s *threadStoreMongo) AdvanceThreadSubscriptionLastSeen(ctx context.Context
...
+			"siteId":          sub.SiteID,
+			"hasMention":      sub.HasMention,
+			"createdAt":       sub.CreatedAt,
+			"updatedAt":       sub.UpdatedAt,
+		},
+		"$max": bson.M{"lastSeenAt": at},
+	}
+	if _, err := s.threadSubscriptions.UpdateOne(ctx, filter, update, options.UpdateOne().SetUpsert(true)); err != nil {
+		return fmt.Errorf("upsert thread subscription advancing lastSeen: %w", err)
+	}
```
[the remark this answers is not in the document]
hmchangw: Confirmed and fixed in 7a219a1.

The race is real: `EnsureIndexes` puts a unique index on `(threadRoomId, userAccount)`, which is also the upsert filter, so two concurrent replies by the same account in one thread can both miss and both attempt the insert. The loser took an E11000, which surfaced as an error and NAKed the reply — the write then only landed on JetStream redelivery.

On a duplicate key the subscription is known to exist, so the fix replays the `$max` alone with upsert disabled, mirroring what `EnsureThreadRoom` already does for the same race one collection over.

Also added a regression test (`..._ConcurrentInsertRace`) driving 8 racing upserts at one `(threadRoomId, userAccount)`: none may error, exactly one subscription survives, and `lastSeenAt` lands on the newest time.

---

message-worker/integration_test.go
hmchangw: Confirmed and restored in 8b6a889.

Both tests exist on `main` and cover behaviour this PR does not change — `GetHistorySharedSince` and `SaveThreadMessage` are untouched — so dropping them was accidental. Restored verbatim from `main`.

For the record, the removal came from the original PR's diff rather than the re-apply onto `main`: both are absent from #13's head too. Comparing test functions between `main` and this branch now leaves only `TestThreadStoreMongo_CreateThreadRoom` and `TestThreadStoreMongo_GetThreadRoomByParentMessageID` removed, which cover the two store methods this PR deletes.

---

hmchangw:
Superseded: this change has been re-applied onto current `main` and split into four independent PRs. Closing.

## Why it could not be merged as-is

The same problem that killed #13. This PR's head branch and `main` have **unrelated git histories** — different root commits (`93fe60e` vs `fcfb22c`) — because `main` was rewritten again after this branch was cut. `git merge-base` returns nothing, so GitHub reports `dirty` with no resolvable conflict set, and a PR's head branch cannot be changed after it is opened. The work had to be re-applied on a fresh branch regardless of whether it was split.

## The split

Each is a single commit based directly on `main`, and each builds, lints and passes `make test` (race) on its own:

| PR | Change | Per-reply saving on the hot path |
|----|--------|----------------------------------|
| #507 | Stop re-stamping the parent's `thread_room_id` | 1 Cassandra write |
| #508 | `EnsureThreadRoom` — resolve the room in one round trip | 1 rejected Mongo insert + 1 read |
| #509 | Stop re-upserting the parent author's subscription | 1 Mongo write + 1 Mongo read |
| #510 | Fold the replier's `lastSeenAt` `$max` into their subscription upsert | 1 Mongo write |

They can merge in any order. #508, #509 and #510 all touch `handleSubsequentThreadReply`, so each one after the first needs a mechanical rebase; #510's description notes the one non-mechanical follow-up (dropping `UpsertThreadSubscription` from `ThreadStore` once #509 has also landed).

## Differences from this PR's content

Not a pure replay — `main` has moved, and two things had to be reworked rather than carried over:

- **The index gates.** `main` added `mongoutil.IndexGate` to `CreateThreadRoom` since this branch was cut. This PR's `EnsureThreadRoom` predates them and would have dropped both `Ready` calls. #508 keeps them, and the `parentMessageId` gate is load-bearing in a new way there: the upsert *filters* on `parentMessageId`, so without the unique key two concurrent first replies both miss and both insert.
- **The duplicate-key narrowing** from `e1474a5` is folded into #510's implementation from the start rather than arriving as a follow-up fix, and has a dedicated integration test.

Integration tests still cannot run in the authoring environment (no Docker daemon), so CI remains the check on them.

The **"this does not move max RPS on its own"** finding from this PR's description still stands: the thread `max-rps` ceiling is bounded by `MAX_WORKERS` and `maxPoolSize` (both 100, every resource ≤70%), not DB CPU. These four PRs lower per-reply DB load, which raises the ceiling once concurrency is unblocked; raising the two concurrency knobs together is still tracked separately.

---

Issue #332, opened 2026-08-20
[Refactor] MongoDB index ops: single-owner creates + log-and-continue (no os.Exit crashloop)


chenjr0719:
> **Verified against:** `main`@b7e9047f — code read after a fresh `git pull`.

## Summary
Every MongoDB `EnsureIndexes`/`ensureIndexes` call site (~20 services) fails **fatally** — `slog.Error(...) → os.Exit(1)` — so any index error crashloops the service at startup. The dangerous trigger is a **cross-service spec conflict**: `CreateOne` is idempotent only when the spec matches, so when two services create the same index with a different option/name/key Mongo returns `IndexKeySpecsConflict`/`IndexOptionsConflict` and both crash. This is the #158/#159 failure mode, and the audit below shows it is systemic, not a one-off.

**Proposal:** (1) give each collection a **single owner** service for index create/drop; (2) make **all** `EnsureIndexes` **log-and-continue** (owner and dependents) instead of `os.Exit` — a missing performance index is a slowdown, not a correctness break; (3) index-dependents that don't own a collection only *verify + log*, never create (so a conflict can't be reintroduced).

## Audit — collections created by MORE than one service (the conflict class)
| collection | creators | duplicated index |
|---|---|---|
| `users` | user-service, room-service, search-service, admin-service, oplog-collections-transformer | `{account}` **unique** ×5 |
| `apps` | user-service, room-service, search-service | `{assistant.name}` `assistant_name_idx` ×3 |
| `subscriptions` | user-service, room-service, bot-room-service (+ search-service reads) | `{roomId,u.account}` **unique** ×3 |
| `thread_subscriptions` | room-service, message-worker, inbox-worker, history-service (+ user-service) | `{threadRoomId,userAccount}` **unique** ×4; legacy `threadRoomId_1_userId_1` dropped by 2 |
| `thread_rooms` | message-worker, broadcast-worker, history-service | (mostly distinct keys) |
| `sessions` (via `pkg/session`) | admin-service, botplatform-service | 3 indexes ×2 |

**Single-owner already (no conflict, just flip to log-and-continue):** `sso_tokens`, `fab_domain_mapping` (user-service), `hr_employee` (portal-service), `cards` (tcard-service), `teams_chat` (teams-chat-sync), `teams_meetings`, `bot_cmd_menu`, `room_members` (room-service), `admin_audit`, `permission_grants` (admin-service).

[...]

PR #333, opened 2026-08-20
fix(mongo-indexes): non-fatal index ensure + single-owner shared indexes

chenjr0719:
Refs #332

## Summary

Index creation at service startup is currently fatal: every `EnsureIndexes` failure
calls `os.Exit(1)` (or returns an error that aborts `run()`), so a single index problem —
most often a cross-service spec conflict on a shared collection — crashloops the service.

This makes index ensure **non-fatal** everywhere (log-and-continue), and **consolidates each
shared/unique index to a single owning service**. Services that only depend on a shared index
no longer create it; they verify-and-warn, so they can neither conflict with the owner's spec
nor die when the index is briefly absent.

## What's included

- **Non-fatal index ensure (all services).** Every `EnsureIndexes`/`ensureIndexes` call site
  now logs a warning and continues instead of exiting. Indexes are best-effort at startup; a
  build failure no longer takes the service down.
- **Single-owner shared indexes.** The indexes previously created by more than one service are
  now created by exactly one owner:
  - `users.account` (unique) and `apps.assistant_name_idx` → **user-service**
  - `subscriptions.{roomId,u.account}` (unique) → **room-service**
  - `thread_subscriptions.{threadRoomId,userAccount}` (unique), plus the legacy
    `threadRoomId_1_userId_1` drop → **room-service**
- **Dependents check-and-log.** A service that depends on an owned index no longer creates it;
  it calls the new `mongoutil.WarnMissingIndexes` helper, which warns if the index is absent
  (its owner builds it) and never errors.
- **Owners self-heal a wrong-spec index.** Owner uniques go through the new
  `mongoutil.EnsureIndexWithRepair`: if an index with the same keys but a conflicting spec
  already exists (`IndexOptionsConflict`/`IndexKeySpecsConflict` — e.g. a non-unique `account_1`,
  the #159 case), it drops the conflicting index and recreates it to the expected spec at
  startup, so an environment carrying the old wrong index converges without operator action.
  A duplicate-*data* failure (E11000) is not repairable by an index drop and still surfaces the
  one-time dedupe-preflight guidance. Applied to `users.account`, `apps.assistant_name_idx`,
  `subscriptions.{roomId,u.account}`, `thread_subscriptions.{threadRoomId,userAccount}`, `room_members`.
- Service-specific, non-shared indexes stay with their service. `thread_rooms` and `sessions`
  keep their existing creators — they have no divergent-spec duplicate, so consolidating them
  would add cross-service coupling with no conflict to prevent; the non-fatal change covers them.

## Changes

- `pkg/mongoutil/indexes.go` — new `WarnMissingIndexes(ctx, coll, names...)` (list-and-warn,
  never returns an error).
- ~20 `EnsureIndexes` call sites across `*/main.go` → warn + continue (no `os.Exit`, no aborting
  error return).
- Owner stores keep their creates; dependent stores (room-service, search-service, admin-service,
  bot-room-service, user-service, message-worker, inbox-worker, history-service) drop the shared
  create and warn instead. The `users.account` operator-guidance errors (E11000 dedupe preflight,
  non-unique conflict) now live only on the owner, user-service.
- Tests moved to the new ownership: the unique-constraint and operator-guidance assertions moved
  to the owning service's tests.

## Rollout note

Because a unique index now has a single creator, the guarantee it enforces depends on the owner
service having built it. On a brand-new environment, deploy the owner ahead of its dependents
(or accept a brief startup window where the index is not yet present). Existing environments
already carry these indexes, so this only affects a from-scratch bring-up.

## Verification

- `go build` green across all touched services.
- `go vet -tags integration` green (integration tests updated to the new ownership).
- golangci-lint clean.
- Index ensure is exercised by the per-service testcontainer integration tests (real Mongo),
  which is where this behavior is covered.

## Summary by CodeRabbit

* **Reliability**
  * Services and workers now continue starting when database index setup encounters errors, while logging warnings.
  * Index initialization proceeds independently, reducing disruption from individual setup failures.

* **Data Integrity**
  * Conflicting indexes can be repaired automatically while preserving safeguards against duplicate data.
  * Shared indexes are verified consistently, with warnings when required indexes are missing.

* **Bug Fixes**
  * Updated index ownership and migration behavior to prevent competing services from managing the same indexes.
  * Improved duplicate-account and subscription constraint validation.

### DIFF
--- a/admin-service/store_mongo.go
+++ b/admin-service/store_mongo.go
@@ -11,6 +11,7 @@ import (
 	"go.mongodb.org/mongo-driver/v2/mongo/options"
 
 	"github.com/hmchangw/chat/pkg/model"
+	"github.com/hmchangw/chat/pkg/mongoutil"
 	"github.com/hmchangw/chat/pkg/session"
 )
 
@@ -30,21 +31,15 @@ func newStoreMongo(db *mongo.Database) *storeMongo {
 
 // EnsureIndexes creates required indexes idempotently.
 func (s *storeMongo) EnsureIndexes(ctx context.Context) error {
-	// Matches botplatform's auto-named "account_1" unique index for idempotency on the shared collection.
-	_, err := s.users.Indexes().CreateOne(ctx, mongo.IndexModel{
-		Keys:    bson.D{{Key: "account", Value: 1}},
-		Options: options.Index().SetUnique(true),
-	})
-	if err != nil {
-		return fmt.Errorf("create users account index: %w", err)
-	}
+	// users.account (unique) is owned by user-service; verify + warn only, never create.
+	mongoutil.WarnMissingIndexes(ctx, s.users, "account_1")
 
 	// Backs SearchUsers, whose only non-regex predicate is siteId: no other service
 	// declares a siteId-prefixed index on the shared users collection, so without
 	// this both the count and the paged find scan every user document. account
 	// trails so the unfiltered count is answered from the index alone (a q-filtered
 	// one still fetches, since engName/chineseName aren't in the key).
-	_, err = s.users.Indexes().CreateOne(ctx, mongo.IndexModel{
+	_, err := s.users.Indexes().CreateOne(ctx, mongo.IndexModel{
 		Keys: bson.D{{Key: "siteId", Value: 1}, {Key: "account", Value: 1}},
 	})
 	if err != nil {
--- a/bot-room-service/main.go
+++ b/bot-room-service/main.go
@@ -77,7 +77,7 @@ func run() error {
 	ensureCtx, ensureCancel := context.WithTimeout(ctx, 30*time.Second)
 	defer ensureCancel()
 	if err := store.EnsureIndexes(ensureCtx); err != nil {
-		return fmt.Errorf("ensure store indexes: %w", err)
+		slog.Warn("ensure store indexes failed; continuing (indexes are best-effort)", "error", err)
 	}
 
 	if cfg.RoomKeyGracePeriod <= 0 {
--- a/bot-room-service/store_mongo.go
+++ b/bot-room-service/store_mongo.go
@@ -11,6 +11,7 @@ import (
 	"go.mongodb.org/mongo-driver/v2/mongo/options"
 
 	"github.com/hmchangw/chat/pkg/model"
+	"github.com/hmchangw/chat/pkg/mongoutil"
 )
 
 type storeMongo struct {
@@ -31,14 +32,8 @@ func newStoreMongo(db *mongo.Database) *storeMongo {
 // writes, and must be invoked once at startup. The rooms and users access here
 // (including the room-key store) is entirely _id-keyed, so neither needs one.
 func (s *storeMongo) EnsureIndexes(ctx context.Context) error {
-	// Mirrors room-service's declaration — same keys AND same unique option, or
-	// whichever service starts second hits IndexOptionsConflict on the shared collection.
-	if _, err := s.subs.Indexes().CreateOne(ctx, mongo.IndexModel{
-		Keys:    bson.D{{Key: "roomId", Value: 1}, {Key: "u.account", Value: 1}},
-		Options: options.Index().SetUnique(true),
-	}); err != nil {
-		return fmt.Errorf("ensure subscriptions (roomId,u.account) unique index: %w", err)
-	}
+	// subscriptions.{roomId,u.account} (unique) is owned by room-service; verify + warn only, never create.
+	mongoutil.WarnMissingIndexes(ctx, s.subs, "roomId_1_u.account_1")
 	// Upsert and delete identify a subscription by u._id — the remove RPC carries
 	// user ids, not accounts — which the unique index above cannot serve.
 	if _, err := s.subs.Indexes().CreateOne(ctx, mongo.IndexModel{
--- a/broadcast-worker/main.go
+++ b/broadcast-worker/main.go
@@ -137,8 +137,7 @@ func main() {
 	}
 	store := NewMongoStore(db.Collection("rooms"), db.Collection("subscriptions"), db.Collection("thread_rooms"), metaValkey, cfg.RoomMetaL2TTL)
 	if err := store.EnsureIndexes(ctx); err != nil {
-		slog.Error("ensure indexes failed", "error", err)
-		os.Exit(1)
+		slog.Warn("ensure indexes failed; continuing (indexes are best-effort)", "error", err)
 	}
 	cachedStore, err := newCachedMetaStore(store, cfg.RoomMetaCacheSize, cfg.RoomMetaCacheTTL)
 	if err != nil {
--- a/inbox-worker/integration_test.go
+++ b/inbox-worker/integration_test.go
@@ -15,7 +15,6 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.mongodb.org/mongo-driver/v2/bson"
 	"go.mongodb.org/mongo-driver/v2/mongo"
-	"go.mongodb.org/mongo-driver/v2/mongo/options"
 
 	"github.com/hmchangw/chat/pkg/model"
 	"github.com/hmchangw/chat/pkg/stream"
@@ -1216,43 +1215,6 @@ func TestIntegration_HandleRoomVisibilityChanged(t *testing.T) {
 	assert.Equal(t, []model.Role{model.RoleOwner}, rolesByAccount["bob"], "bob should be owner")
 	assert.Equal(t, []model.Role{model.RoleMember}, rolesByAccount["alice"], "alice should be member")
 	assert.Equal(t, []model.Role{model.RoleMember}, rolesByAccount["carol"], "carol should be member")
-}
-
-// ensureIndexes must standardize on (threadRoomId, userAccount) — the same
-// natural key room-service, message-worker, and history-service create — and
-// drop the legacy (threadRoomId, userId) index that message-worker explicitly
-// removes. Otherwise the two services thrash the index across restarts and the
-// collection ends up with two conflicting unique constraints.
-func TestInboxStore_EnsureIndexes_DropsLegacyAndCreatesUserAccount_Integration(t *testing.T) {
-	db := setupMongo(t)
-	ctx := context.Background()
-	threadSubs := db.Collection("thread_subscriptions")
-	store := &mongoInboxStore{threadSubCol: threadSubs}
-
-	// Simulate a DB where the legacy index already exists (older inbox-worker).
-	_, err := threadSubs.Indexes().CreateOne(ctx, mongo.IndexModel{
-		Keys:    bson.D{{Key: "threadRoomId", Value: 1}, {Key: "userId", Value: 1}},
-		Options: options.Index().SetUnique(true),
-	})
-	require.NoError(t, err)
-
-	require.NoError(t, store.ensureIndexes(ctx))
-
-	cur, err := threadSubs.Indexes().List(ctx)
-	require.NoError(t, err)
-	var idxs []bson.M
-	require.NoError(t, cur.All(ctx, &idxs))
-
-	names := make(map[string]bool, len(idxs))
-	for _, ix := range idxs {
-		if n, ok := ix["name"].(string); ok {
-			names[n] = true
-		}
-	}
-	assert.True(t, names["threadRoomId_1_userAccount_1"],
-		"ensureIndexes must create the canonical (threadRoomId, userAccount) unique index")
-	assert.False(t, names["threadRoomId_1_userId_1"],
-		"ensureIndexes must drop the legacy (threadRoomId, userId) index")
 }
 
 // Regression: a federated upsert for an existing (threadRoomId, userAccount)
--- a/inbox-worker/integration_test.go
+++ b/inbox-worker/integration_test.go
@@ -434,7 +434,7 @@ func TestInboxWorker_ThreadSubscriptionUpserted_Insert_Integration(t *testing.T)
 		userCol:      db.Collection("users"),
 		threadSubCol: db.Collection("thread_subscriptions"),
 	}
-	require.NoError(t, store.ensureIndexes(ctx))
+	store.ensureIndexes(ctx)
 
 	handler := NewHandler(store)
 
@@ -483,7 +483,7 @@ func TestInboxWorker_ThreadSubscription_DedupByUserAccount_Integration(t *testin
 		userCol:      db.Collection("users"),
 		threadSubCol: db.Collection("thread_subscriptions"),
 	}
-	require.NoError(t, store.ensureIndexes(ctx))
+	store.ensureIndexes(ctx)
 
 	now := time.Date(2026, 4, 1, 12, 0, 0, 0, time.UTC)
 	first := model.ThreadSubscription{
@@ -520,7 +520,7 @@ func TestInboxWorker_ThreadSubscriptionUpserted_MonotonicMention_Integration(t *
 		userCol:      db.Collection("users"),
 		threadSubCol: db.Collection("thread_subscriptions"),
 	}
-	require.NoError(t, store.ensureIndexes(ctx))
+	store.ensureIndexes(ctx)
 
 	handler := NewHandler(store)
 	now := time.Date(2026, 4, 1, 12, 0, 0, 0, time.UTC)
@@ -1229,7 +1229,7 @@ func TestInboxStore_UpsertThreadSubscription_DedupesByUserAccount_Integration(t 
 	ctx := context.Background()
 	threadSubs := db.Collection("thread_subscriptions")
 	store := &mongoInboxStore{threadSubCol: threadSubs}
-	require.NoError(t, store.ensureIndexes(ctx))
+	store.ensureIndexes(ctx)
 
 	now := time.Now().UTC().Truncate(time.Millisecond)
 
--- a/media-service/main.go
+++ b/media-service/main.go
@@ -62,7 +62,7 @@ func run() error {
 	defer mongoutil.Disconnect(ctx, mongoClient)
 
 	if err := store.EnsureEmojiIndexes(ctx); err != nil {
-		return fmt.Errorf("ensure emoji indexes: %w", err)
+		slog.Warn("ensure emoji indexes failed; continuing (indexes are best-effort)", "error", err)
 	}
 
 	minioClient, err := minioutil.Connect(ctx, cfg.MinioEndpoint, cfg.MinioUseSSL, cfg.MinioAccessKey, cfg.MinioSecretKey, minioutil.WithObservability(sdk))
--- a/message-worker/store_mongo.go
+++ b/message-worker/store_mongo.go
@@ -11,6 +11,7 @@ import (
 	"go.mongodb.org/mongo-driver/v2/mongo/options"
 
 	"github.com/hmchangw/chat/pkg/model"
+	"github.com/hmchangw/chat/pkg/mongoutil"
 )
 
 var (
@@ -44,17 +45,9 @@ func (s *threadStoreMongo) EnsureIndexes(ctx context.Context) error {
 		return fmt.Errorf("ensure thread_rooms parentMessageId index: %w", err)
 	}
 
-	// Best-effort: drop the legacy (threadRoomId, userId) unique index so the new
-	// (threadRoomId, userAccount) index can be created without a key conflict.
-	// The collection or index may not exist (fresh deploy / test container) — ignore all errors.
-	_ = s.threadSubscriptions.Indexes().DropOne(ctx, "threadRoomId_1_userId_1") //nolint:errcheck
-
-	if _, err := s.threadSubscriptions.Indexes().CreateOne(ctx, mongo.IndexModel{
-		Keys:    bson.D{{Key: "threadRoomId", Value: 1}, {Key: "userAccount", Value: 1}},
-		Options: options.Index().SetUnique(true),
-	}); err != nil {
-		return fmt.Errorf("ensure thread_subscriptions (threadRoomId,userAccount) index: %w", err)
-	}
+	// thread_subscriptions.{threadRoomId,userAccount} (unique) is owned by room-service
+	// (which also drops the legacy threadRoomId_1_userId_1 index); verify + warn only, never create.
+	mongoutil.WarnMissingIndexes(ctx, s.threadSubscriptions, "threadRoomId_1_userAccount_1")
 
 	return nil
 }
--- a/message-worker/store_mongo.go
+++ b/message-worker/store_mongo.go
@@ -38,7 +38,7 @@ func newThreadStoreMongo(db *mongo.Database) *threadStoreMongo {
 
 // EnsureIndexes creates the unique indexes required by the thread store.
 func (s *threadStoreMongo) EnsureIndexes(ctx context.Context) error {
-	if _, err := s.threadRooms.Indexes().CreateOne(ctx, mongo.IndexModel{
+	if err := mongoutil.EnsureIndexWithRepair(ctx, s.threadRooms, mongo.IndexModel{
 		Keys:    bson.D{{Key: "parentMessageId", Value: 1}},
 		Options: options.Index().SetUnique(true),
 	}); err != nil {
--- a/pkg/mongoutil/indexes.go
+++ b/pkg/mongoutil/indexes.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log/slog"
 	"strings"
+	"time"
 
 	"go.mongodb.org/mongo-driver/v2/bson"
 	"go.mongodb.org/mongo-driver/v2/mongo"
@@ -43,40 +44,54 @@ func WarnMissingIndexes(ctx context.Context, coll *mongo.Collection, names ...st
 	}
 }
 
+// indexRestoreTimeout bounds the best-effort restore of the old index after a
+// failed repair, on a context detached from the (possibly already-expired) caller.
+const indexRestoreTimeout = 10 * time.Second
+
 // EnsureIndexWithRepair creates model on coll and self-heals a pre-existing index
 // with the SAME keys but a conflicting spec (e.g. a non-unique index where model
 // is unique): it drops the conflicting index and recreates it from model.
 //
 // If the recreate fails — typically E11000 when the DATA holds duplicate values a
-// unique index can't be built over — the old index is restored so the collection
-// is never left without one, and the E11000 is returned so the caller can surface
-// the dedupe-preflight guidance.
+// unique index can't be built over — the old index is restored faithfully (its
+// full spec, under a fresh context since the caller's may be the one that just
+// expired) so the collection is never left without one, and the error is returned
+// so the caller can surface the dedupe-preflight guidance.
+//
+// ponytail: concurrent repair across replicas on the same dirty index is narrowed
+// (a retry before the destructive drop skips it once a peer has repaired, and a
+// drop of an already-absent index is tolerated) but not fully serialized — a
+// distributed lock is deliberately avoided for a one-time startup convergence; the
+// write-gating follow-up closes the residual window.
 func EnsureIndexWithRepair(ctx context.Context, coll *mongo.Collection, model mongo.IndexModel) error {
-	_, err := coll.Indexes().CreateOne(ctx, model)
-	if err == nil || !isIndexSpecConflict(err) {
+	if _, err := coll.Indexes().CreateOne(ctx, model); err == nil || !isIndexSpecConflict(err) {
 		return err
 	}
-	name := indexNameByKeys(ctx, coll, model.Keys)
-	if name == "" {
-		return err // can't identify the conflicting index; surface the original error
+	// A peer replica may have repaired it since the first attempt — retry before the
+	// destructive drop so we never drop a peer's freshly created correct index.
+	if _, err := coll.Indexes().CreateOne(ctx, model); err == nil || !isIndexSpecConflict(err) {
+		return err
 	}
-	// Best-effort: a peer replica may have already dropped it — the recreate below
-	// is the real check.
-	_ = coll.Indexes().DropOne(ctx, name) //nolint:errcheck
+	old := existingIndexByKeys(ctx, coll, model.Keys)
+	if old.name == "" {
+		return fmt.Errorf("repair index on %s: conflicting index not found", coll.Name())
+	}
+	if err := coll.Indexes().DropOne(ctx, old.name); err != nil && !IsIndexNotFound(err) {
+		return fmt.Errorf("repair index %q on %s: drop conflicting index: %w", old.name, coll.Name(), err)
+	}
 	if _, cerr := coll.Indexes().CreateOne(ctx, model); cerr != nil {
-		// Recreate failed (e.g. E11000: duplicate values block a unique index).
-		// Restore the old index so the collection is never left without one; the
-		// returned error still carries the cause so the caller surfaces its guidance.
-		if _, rerr := coll.Indexes().CreateOne(ctx, mongo.IndexModel{
-			Keys: model.Keys, Options: options.Index().SetName(name),
-		}); rerr != nil {
+		// Restore the old index — its full spec, on a fresh context — so the
+		// collection keeps an index; the returned error still carries the cause.
+		restoreCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), indexRestoreTimeout)
+		defer cancel()
+		if _, rerr := coll.Indexes().CreateOne(restoreCtx, old.model()); rerr != nil {
 			slog.WarnContext(ctx, "mongo: index repair failed and the old index could not be restored",
-				"collection", coll.Name(), "index", name, "error", rerr)
+				"collection", coll.Name(), "index", old.name, "error", rerr)
 		}
-		return fmt.Errorf("repair index %q on %s: %w", name, coll.Name(), cerr)
+		return fmt.Errorf("repair index %q on %s: %w", old.name, coll.Name(), cerr)
 	}
 	slog.WarnContext(ctx, "mongo: repaired a conflicting index (dropped and recreated to the expected spec)",
-		"collection", coll.Name(), "droppedIndex", name)
+		"collection", coll.Name(), "droppedIndex", old.name)
 	return nil
 }
 
@@ -87,16 +102,57 @@ func isIndexSpecConflict(err error) bool {
 	return errors.As(err, &se) && (se.HasErrorCode(85) || se.HasErrorCode(86))
 }
 
-// indexNameByKeys returns the name of the existing index on coll whose key spec
-// matches keys (order-sensitive), or "" if none/unreadable.
-func indexNameByKeys(ctx context.Context, coll *mongo.Collection, keys any) string {
+// IsIndexNotFound reports whether err is Mongo's IndexNotFound (27), so dropping an
+// index a peer already removed is not treated as a failure.
+func IsIndexNotFound(err error) bool {
+	var se mongo.ServerError
+	return errors.As(err, &se) && se.HasErrorCode(27)
+}
+
+// existingIndex is a pre-existing index captured for a faithful restore.
+type existingIndex struct {
+	name string
+	keys bson.D
+	spec bson.M
+}
+
+// model reconstructs an IndexModel reproducing the captured index, preserving the
+// correctness-relevant options (unique, sparse, hidden, TTL, partial filter) so a
+// restore never silently weakens the constraint. Collation is not reconstructed —
+// no repaired index in this repo uses it.
+func (e existingIndex) model() mongo.IndexModel {
+	opts := options.Index().SetName(e.name)
+	if v, _ := e.spec["unique"].(bool); v {
+		opts = opts.SetUnique(true)
+	}
+	if v, _ := e.spec["sparse"].(bool); v {
+		opts = opts.SetSparse(true)
+	}
+	if v, _ := e.spec["hidden"].(bool); v {
+		opts = opts.SetHidden(true)
+	}
+	switch v := e.spec["expireAfterSeconds"].(type) {
+	case int32:
+		opts = opts.SetExpireAfterSeconds(v)
+	case int64:
+		opts = opts.SetExpireAfterSeconds(int32(v))
+	}
+	if v, ok := e.spec["partialFilterExpression"]; ok {
+		opts = opts.SetPartialFilterExpression(v)
+	}
+	return mongo.IndexModel{Keys: e.keys, Options: opts}
+}
+
+// existingIndexByKeys returns the index on coll whose key spec matches keys
+// (order-sensitive), captured for restore, or a zero value if none/unreadable.
+func existingIndexByKeys(ctx context.Context, coll *mongo.Collection, keys any) existingIndex {
 	want := keySpec(keys)
 	if want == "" {
-		return ""
+		return existingIndex{}
 	}
 	cur, err := coll.Indexes().List(ctx)
 	if err != nil {
-		return ""
+		return existingIndex{}
 	}
 	defer func() { _ = cur.Close(ctx) }()
 	for cur.Next(ctx) {
@@ -104,11 +160,14 @@ func indexNameByKeys(ctx context.Context, coll *mongo.Collection, keys any) stri
 			Name string `bson:"name"`
 			Key  bson.D `bson:"key"`
 		}
-		if cur.Decode(&idx) == nil && keySpec(idx.Key) == want {
-			return idx.Name
+		if cur.Decode(&idx) != nil || keySpec(idx.Key) != want {
+			continue
 		}
+		var spec bson.M
+		_ = bson.Unmarshal(cur.Current, &spec)
+		return existingIndex{name: idx.Name, keys: idx.Key, spec: spec}
 	}
-	return ""
+	return existingIndex{}
 }
 
 // keySpec renders an index key document as "field:dir,..." preserving order, so
--- a/pkg/mongoutil/indexes.go
+++ b/pkg/mongoutil/indexes.go
@@ -1,0 +1,41 @@
+package mongoutil
+
+import (
+	"context"
+	"log/slog"
+
+	"go.mongodb.org/mongo-driver/v2/mongo"
+)
+
+// WarnMissingIndexes logs a warning for each named index absent from coll and
+// never returns an error. It is for services that DEPEND on an index another
+// service owns: a dependent must not create the shared index (a divergent spec
+// crashloops whichever service starts second) nor die when the owner has not
+// built it yet. names are the owner's resolved index names (e.g. "account_1",
+// "assistant_name_idx"); a name that drifts from the owner's spec only costs a
+// spurious warning, never a failure.
+func WarnMissingIndexes(ctx context.Context, coll *mongo.Collection, names ...string) {
+	cur, err := coll.Indexes().List(ctx)
+	if err != nil {
+		slog.WarnContext(ctx, "mongo: cannot list indexes to verify dependencies",
+			"collection", coll.Name(), "error", err)
+		return
+	}
+	defer func() { _ = cur.Close(ctx) }()
+
+	have := make(map[string]bool)
+	for cur.Next(ctx) {
+		var idx struct {
+			Name string `bson:"name"`
+		}
+		if err := cur.Decode(&idx); err == nil {
+			have[idx.Name] = true
+		}
+	}
+	for _, n := range names {
+		if !have[n] {
+			slog.WarnContext(ctx, "mongo: depended-on index missing; its owner service must create it",
+				"collection", coll.Name(), "index", n)
+		}
+	}
+}
--- a/pkg/mongoutil/indexes.go
+++ b/pkg/mongoutil/indexes.go
@@ -2,8 +2,12 @@ package mongoutil
 
 import (
 	"context"
+	"errors"
+	"fmt"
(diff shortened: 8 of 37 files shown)


pkg/mongoutil/indexes.go
```diff
@@ -0,0 +1,126 @@
...
+		// Recreate failed (e.g. E11000: duplicate values block a unique index).
+		// Restore the old index so the collection is never left without one; the
+		// returned error still carries the cause so the caller surfaces its guidance.
+		if _, rerr := coll.Indexes().CreateOne(ctx, mongo.IndexModel{
+			Keys: model.Keys, Options: options.Index().SetName(name),
+		}); rerr != nil {
+			slog.WarnContext(ctx, "mongo: index repair failed and the old index could not be restored",
+				"collection", coll.Name(), "index", name, "error", rerr)
+		}
+		return fmt.Errorf("repair index %q on %s: %w", name, coll.Name(), cerr)
```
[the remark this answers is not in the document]
chenjr0719: Thanks — but I think a full-spec snapshot is dead generality here: every index this helper repairs is a simple single- or compound-key index distinguished only by the `unique` flag. None use TTL, partial filters, collation, sparse, or hidden state. So the restore (same keys + same name, non-unique) faithfully reproduces the only pre-existing shape that can occur — the non-unique variant of the target key. Reconstructing an arbitrary IndexModel from a listed spec (mapping the raw index doc back through all option setters) would be untested code for a case that cannot arise in this repo. Leaving it as-is; if a repaired index ever gains TTL/partial/collation, the snapshot becomes worth adding.

inbox-worker/integration_test.go
```diff
@@ -484,7 +483,7 @@ func TestInboxWorker_ThreadSubscription_DedupByUserAccount_Integration(t *testin
 		userCol:      db.Collection("users"),
 		threadSubCol: db.Collection("thread_subscriptions"),
 	}
-	require.NoError(t, store.ensureIndexes(ctx))
+	store.ensureIndexes(ctx)
```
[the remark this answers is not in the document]
chenjr0719: These two tests assert the *upsert-filter* dedup — the subscription is keyed on `userAccount`, so a second upsert with a different `userId` must merge, not insert. That is the actual mechanism, and it is what the assertion verifies: a mis-keying regression (filtering on `userId`) makes the second upsert insert a second row → the `CountDocuments == 1` assertion fails, with or without the unique index. The unique index is a separate backstop, and it is exercised in the owner test (room-service `TestEnsureIndexes_ThreadSubsDropsLegacyAndCreatesUnique`, which now also asserts the index is unique). So seeding it here would not strengthen what these tests prove. Leaving them as-is.

bot-room-service/main.go
```diff
@@ -77,7 +77,7 @@ func run() error {
 	ensureCtx, ensureCancel := context.WithTimeout(ctx, 30*time.Second)
 	defer ensureCancel()
 	if err := store.EnsureIndexes(ensureCtx); err != nil {
-		return fmt.Errorf("ensure store indexes: %w", err)
+		slog.Warn("ensure store indexes failed; continuing (indexes are best-effort)", "error", err)
```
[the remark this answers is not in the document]
chenjr0719: Good point, and we are addressing it — as a follow-up PR rather than here. Instead of failing startup or blocking writes (which reintroduces the crashloop this PR intentionally removes), we will add a **readiness gate**: the service still starts and never crashes, but reports **not-ready** (its readiness probe fails, so it receives no traffic) until its owned unique indexes exist. That keeps a service from serving writes without the constraint while preserving the no-crash behavior. This PR keeps the log-and-continue + self-heal; the readiness gating lands separately.

room-service/main.go
```diff
@@ -206,9 +206,7 @@ func main() {
 	// Bounded timeout so a hung createIndexes surfaces at startup.
 	ensureCtx, ensureCancel := context.WithTimeout(ctx, 30*time.Second)
 	if err := store.EnsureIndexes(ensureCtx); err != nil {
-		ensureCancel()
-		slog.Error("ensure store indexes failed", "error", err)
-		os.Exit(1)
+		slog.Warn("ensure store indexes failed; continuing (indexes are best-effort)", "error", err)
 	}
```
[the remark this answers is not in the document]
chenjr0719: Good point, and we are addressing it — as a follow-up PR rather than here. Instead of failing startup or blocking writes (which reintroduces the crashloop this PR intentionally removes), we will add a **readiness gate**: the service still starts and never crashes, but reports **not-ready** (its readiness probe fails, so it receives no traffic) until its owned unique indexes exist. That keeps a service from serving writes without the constraint while preserving the no-crash behavior. This PR keeps the log-and-continue + self-heal; the readiness gating lands separately.

broadcast-worker/main.go
```diff
@@ -137,8 +137,7 @@ func main() {
 	}
 	store := NewMongoStore(db.Collection("rooms"), db.Collection("subscriptions"), db.Collection("thread_rooms"), metaValkey, cfg.RoomMetaL2TTL)
 	if err := store.EnsureIndexes(ctx); err != nil {
-		slog.Error("ensure indexes failed", "error", err)
-		os.Exit(1)
+		slog.Warn("ensure indexes failed; continuing (indexes are best-effort)", "error", err)
 	}
```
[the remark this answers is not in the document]
chenjr0719: Good point — addressing it in a follow-up PR (the readiness gate described on the other threads): the service starts and never crashes, but reports not-ready until its owned unique indexes exist, so it takes no traffic without the constraint. The bounded startup context you flagged goes in that same follow-up — the ensure calls still using context.Background() will get a bounded timeout (a few services already do). This PR keeps the log-and-continue + self-heal.

julianshen:
**8 findings · 4 unique issues · 4 inline comments posted** — 🔴 7 blocking · 🟠 1 warning

### Related work

- [Issue #332](https://github.com/hmchangw/newchat/issues/332) — \[Refactor\] MongoDB index ops: single-owner creates \+ log-and-continue \(no os.Exit crashloop\) (open)
- [PR #159](https://github.com/hmchangw/newchat/pull/159) — fix\(search-service\): make users\(account\) index unique to match canonical (merged)

### Local review memory

- `bot-room-service/main.go:80` — @chenjr0719: Good point, and we are addressing it — as a follow-up PR rather than here. Instead of failing startup or blocking writes (which reintroduces the crashloop this PR intentionally removes), we will add a **readiness gate**: the service still… ([thread](https://github.com/hmchangw/newchat/pull/333#discussion_r3824721699))
- `bot-room-service/main.go:80` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟠 Major_ _🏗️ Heavy lift_ &lt;details> &lt;summary>🧩 Analysis chain&lt;/summary> 🏁 Script executed: shell #!/bin/bash set -euo pipefail rg -n -C 20 'EnsureIndexes EnsureIndexWithRepair' bot-room-service… ([thread](https://github.com/hmchangw/newchat/pull/333#discussion_r3824721699))
- `bot-room-service/main.go:80` — @coderabbitai: @chenjr0719 , thanks for the clarification. A readiness gate addresses the concern without making index initialization fatal. I will keep this thread open for the follow-up change. Would you like me to create a GitHub issue for the readin… ([thread](https://github.com/hmchangw/newchat/pull/333#discussion_r3824721699))
- `room-service/main.go:210` — @chenjr0719: Good point, and we are addressing it — as a follow-up PR rather than here. Instead of failing startup or blocking writes (which reintroduces the crashloop this PR intentionally removes), we will add a **readiness gate**: the service still… ([thread](https://github.com/hmchangw/newchat/pull/333#discussion_r3824721735))
- `room-service/main.go:210` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟠 Major_ _🏗️ Heavy lift_ &lt;details> &lt;summary>🧩 Analysis chain&lt;/summary> 🏁 Script executed: shell #!/bin/bash set -eu printf '%s\n' '--- relevant files ---' git ls-files rg '(^ /)(room-service… ([thread](https://github.com/hmchangw/newchat/pull/333#discussion_r3824721735))
- `room-service/main.go:210` — @coderabbitai: @chenjr0719 , the readiness gate is an appropriate follow-up design. It preserves the non-fatal startup behavior in this PR. It also prevents traffic from reaching a service until its owned unique indexes are available. I will keep this t… ([thread](https://github.com/hmchangw/newchat/pull/333#discussion_r3824721735))
- `broadcast-worker/main.go:141` — @chenjr0719: Good point — addressing it in a follow-up PR (the readiness gate described on the other threads): the service starts and never crashes, but reports not-ready until its owned unique indexes exist, so it takes no traffic without the constrai… ([thread](https://github.com/hmchangw/newchat/pull/333#discussion_r3824721708))
- `broadcast-worker/main.go:141` — @coderabbitai: _🩺 Stability & Availability_ _🟠 Major_ _⚡ Quick win_ &lt;details> &lt;summary>🧩 Analysis chain&lt;/summary> 🏁 Script executed: shell #!/bin/bash set -eu printf '%s\n' '--- candidate files ---' git ls-files '*main.go' '*repository*.go' '*… ([thread](https://github.com/hmchangw/newchat/pull/333#discussion_r3824721708))
- `broadcast-worker/main.go:141` — @coderabbitai: @chenjr0719 , thanks for the clarification. The readiness gate addresses the safety concern without changing the availability goal. A bounded context for EnsureIndexes also prevents startup from waiting indefinitely. This finding remain… ([thread](https://github.com/hmchangw/newchat/pull/333#discussion_r3824721708))

<details>
<summary>📒 Files reviewed (37)</summary>

* `admin-service/main.go`
* `bot-room-service/integration_test.go`
* `bot-room-service/main.go`
* `bot-room-service/store_mongo.go`
* `broadcast-worker/main.go`
* `data-migration/oplog-collections-transformer/main.go`
* `data-migration/oplog-collections-transformer/targetstore.go`
* `history-service/internal/mongorepo/threadsubscription.go`
* `inbox-worker/main.go`
* `media-service/main.go`
* `media-service/store_mongo.go`
* `message-worker/main.go`
* `pkg/mongoutil/indexes.go`
* `pkg/mongoutil/indexes_integration_test.go`
* `portal-service/main.go`
* `portal-service/store_mongo.go`
* `room-service/main.go`
* `search-service/main.go`
* `search-service/store_mongo.go`
* `tcard-service/main.go`
* `tcard-service/store_mongo.go`
* `teams-chat-sync/main.go`
* `user-service/main.go`
* `user-service/mongorepo/apps.go`
* `user-service/mongorepo/indexes_test.go`
* `user-service/mongorepo/ssotokens.go`
* `user-service/mongorepo/users.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

room-service/main.go
```diff
@@ -206,9 +206,7 @@ func main() {
 	// Bounded timeout so a hung createIndexes surfaces at startup.
 	ensureCtx, ensureCancel := context.WithTimeout(ctx, 30*time.Second)
 	if err := store.EnsureIndexes(ensureCtx); err != nil {
-		ensureCancel()
-		slog.Error("ensure store indexes failed", "error", err)
-		os.Exit(1)
+		slog.Warn("ensure store indexes failed; continuing (indexes are best-effort)", "error", err)
```
julianshen, line 209: _🏷️ idempotency_ | _🔴 Blocking_ | _`distributed-system`, `mongodb`, `tgd-review` (3 rules)_

**Owner services serve writes after unique-index setup fails.**

A timeout, Mongo outage, or failed unique-index rebuild makes EnsureIndexes return an error, but startup continues. The service then accepts operations without the room_members, subscriptions, thread_subscriptions, or teams_meetings uniqueness constraints. For example, two concurrent or retried meeting requests can both insert idempotency records and publish duplicate teams_meet_started messages; similarly, redelivered member operations can insert duplicate rows under fresh IDs. Keep the process alive if desired, but do not make affected RPCs or consumers ready until every correctness-critical unique index has been verified, and retry initialization separately from performance indexes. The planned follow-up is not present in this diff, so the prior concern remains valid.

<details>

- **`mongodb`**:209: The service continues serving after EnsureIndexes fails, including failures for the room_members, subscriptions, thread_subscriptions, and teams_meetings unique constraints. Retries or concurrent requests can then insert duplicate logical records and, for meetings, publish duplicate start messages. The same pattern lets message-worker consume without thread_rooms.parentMessageId uniqueness and lets other collection owners continue without their declared unique constraints. The proposed follow-up does not protect this revision. Separate correctness-critical index status from performance-index warnings and keep readiness false until every owned unique index is verified.
- **`tgd-review`**:209: EnsureIndexes includes correctness constraints for room members, subscriptions, thread subscriptions, and Teams meeting idempotency, but this branch now logs every failure and proceeds to serve traffic. Without those constraints, retries or concurrent requests can create duplicate logical records. The discussed readiness gate is not present in this diff, so the unsafe interval remains. Keep the service unready, or disable the affected writes, until all owner-managed unique indexes are confirmed while allowing only performance-index failures to remain best-effort.

</details>

<details>

```
In `room-service/main.go` around line 209: A timeout, Mongo outage, or failed unique-index rebuild makes EnsureIndexes return an error, but startup continues. The service then accepts operations without the room_members, subscriptions, thread_subscriptions, or teams_meetings uniqueness constraints. For example, two concurrent or retried meeting requests can both insert idempotency records and publish duplicate teams_meet_started messages; similarly, redelivered member operations can insert duplicate rows under fresh IDs. Keep the process alive if desired, but do not make affected RPCs or consumers ready until every correctness-critical unique index has been verified, and retry initialization separately from performance indexes. The planned follow-up is not present in this diff, so the prior concern remains valid.

```

</details>

bot-room-service/store_mongo.go
```diff
@@ -31,14 +32,8 @@ func newStoreMongo(db *mongo.Database) *storeMongo {
 // writes, and must be invoked once at startup. The rooms and users access here
 // (including the room-key store) is entirely _id-keyed, so neither needs one.
 func (s *storeMongo) EnsureIndexes(ctx context.Context) error {
```
julianshen, line 34: _🏷️ consistency_ | _🔴 Blocking_ | _`distributed-system`_

**Dependent writers run before owner-managed unique indexes exist.**

WarnMissingIndexes only logs, so startup ordering becomes a correctness dependency. If room-service is unavailable or its index repair fails, bot-room-service still becomes ready without roomId_1_u.account_1. UpsertSubscription identifies rows by roomId and the site-local u._id; two events for the same account carrying different IDs can therefore both insert, producing duplicate subscriptions and leaving one behind when only one ID is deleted. Gate readiness or reject writes until the exact owner-managed index, including its unique option, is verified. Warning-only verification is insufficient, and the announced follow-up is not included here.

<details>

```
In `bot-room-service/store_mongo.go` around line 34: WarnMissingIndexes only logs, so startup ordering becomes a correctness dependency. If room-service is unavailable or its index repair fails, bot-room-service still becomes ready without roomId_1_u.account_1. UpsertSubscription identifies rows by roomId and the site-local u._id; two events for the same account carrying different IDs can therefore both insert, producing duplicate subscriptions and leaving one behind when only one ID is deleted. Gate readiness or reject writes until the exact owner-managed index, including its unique option, is verified. Warning-only verification is insufficient, and the announced follow-up is not included here.

```

</details>

pkg/mongoutil/indexes.go
```diff
@@ -0,0 +1,126 @@
...
+	if err == nil || !isIndexSpecConflict(err) {
+		return err
+	}
+	name := indexNameByKeys(ctx, coll, model.Keys)
+	if name == "" {
+		return err // can't identify the conflicting index; surface the original error
+	}
+	// Best-effort: a peer replica may have already dropped it — the recreate below
+	// is the real check.
+	_ = coll.Indexes().DropOne(ctx, name) //nolint:errcheck
```
julianshen: _🎯 concurrency_ | _🔴 Blocking_ | _`distributed-system`, `performance-and-concurrency`, `tgd-review` (3 rules)_

**Concurrent repairs can drop a peer's newly created unique index.**

The repair sequence is not serialized across replicas. Two rolling-start replicas can both observe the old non-unique account_1 and remember its name. Replica A drops it and successfully creates the unique account_1; replica B then executes this stale DropOne and removes A's completed unique index because the repaired index has the same name. Existing replicas can admit duplicate accounts while B rebuilds; that build then fails with E11000, and B restores a non-unique index while callers merely log the error and continue. Serialize repair per collection/index with a distributed, fenced lock and re-read the specification under that lock, or move destructive repair into a single migration job while all dependent writers are gated.

<details>

- **`performance-and-concurrency`**:65: Two replicas can both observe the same conflicting auto-named index. Replica A can drop it and successfully recreate the unique index, after which replica B executes this stale `DropOne` and removes A's repaired index. This repeats an expensive build on large collections and opens a window where active writers can insert duplicates, causing B's rebuild to fail. Serialize repairs with a distributed per-index lease or a single migration owner, then re-run `CreateOne` and re-read the current specification after acquiring ownership before dropping anything.
- **`tgd-review`**:65: Multiple replicas can observe the same conflict and resolve the same index name. After one replica drops the old index and creates the correct unique index, another replica can execute this stale DropOne and remove the newly repaired index before recreating it. Writes during that gap can introduce duplicates and make the second recreation fail. Serialize repairs, use an atomic MongoDB conversion mechanism, or leave conflicts for coordinated operator remediation rather than performing an unconditional drop.

</details>

<details>

```
In `pkg/mongoutil/indexes.go` around line 65: The repair sequence is not serialized across replicas. Two rolling-start replicas can both observe the old non-unique account_1 and remember its name. Replica A drops it and successfully creates the unique account_1; replica B then executes this stale DropOne and removes A's completed unique index because the repaired index has the same name. Existing replicas can admit duplicate accounts while B rebuilds; that build then fails with E11000, and B restores a non-unique index while callers merely log the error and continue. Serialize repair per collection/index with a distributed, fenced lock and re-read the specification under that lock, or move destructive repair into a single migration job while all dependent writers are gated.

```

</details>

pkg/mongoutil/indexes.go
```diff
@@ -0,0 +1,126 @@
...
+	if _, cerr := coll.Indexes().CreateOne(ctx, model); cerr != nil {
+		// Recreate failed (e.g. E11000: duplicate values block a unique index).
+		// Restore the old index so the collection is never left without one; the
+		// returned error still carries the cause so the caller surfaces its guidance.
```
julianshen: _🏷️ index lifecycle_ | _🔴 Blocking_ | _`mongodb`_

**Index rollback can remove or weaken the previous constraint.**

After dropping the conflicting index, rollback reuses the operation context and reconstructs the old index with only the desired keys and old name. If recreation failed because the context expired, rollback necessarily uses that expired context and leaves the collection without the index. If rollback succeeds, it discards the old index's unique, sparse, partial, collation, TTL, and other options; an old unique index can therefore become non-unique. Capture the complete existing specification before DropOne and restore it with a separate bounded recovery context.

<details>

```
In `pkg/mongoutil/indexes.go` around line 69: After dropping the conflicting index, rollback reuses the operation context and reconstructs the old index with only the desired keys and old name. If recreation failed because the context expired, rollback necessarily uses that expired context and leaves the collection without the index. If rollback succeeds, it discards the old index's unique, sparse, partial, collation, TTL, and other options; an old unique index can therefore become non-unique. Capture the complete existing specification before DropOne and restore it with a separate bounded recovery context.

```

</details>

pkg/mongoutil/indexes.go
```diff
@@ -0,0 +1,126 @@
...
+		if _, rerr := coll.Indexes().CreateOne(ctx, mongo.IndexModel{
```
julianshen: _🎯 correctness_ | _🔴 Blocking_ | _`performance-and-concurrency`, `silent-failure`, `tgd-review` (3 rules)_

**The repair fallback does not restore the old index specification**

The fallback recreates only the old name with the requested keys, discarding the original index's unique, sparse, partial-filter, collation, hidden, and other options. For example, a correctly unique index with a different name can be dropped and then restored as non-unique after a transient recreation failure. It also reuses the failed operation's context, so a deadline or cancellation leaves the collection without any restored index. Capture the complete existing specification before dropping it and restore that specification with an appropriate bounded cleanup context.

<details>

- **`silent-failure`**:70: The rollback recreates only `model.Keys` and the old name, not the old index's unique, sparse, partial-filter, collation, TTL, or other options. A name-only conflict can therefore drop a valid unique index and restore it as non-unique when the desired recreation fails. If rollback itself fails, that error is only logged and omitted from the returned error. Capture the complete existing specification before dropping it, recreate that specification on failure, and return a joined error containing both the repair and rollback failures.
- **`performance-and-concurrency`**:70: The context can expire after `DropOne` succeeds but before the replacement is created. The restoration then reuses that already-canceled context and fails immediately, leaving the collection without either index. This is reachable with room-service's 30-second startup deadline during a large rebuild, and startup now continues afterward. Perform restoration with a separate bounded cleanup context not canceled with the original operation, and keep the owner unready or retry until restoration or repair is confirmed.

</details>

<details>

```
In `pkg/mongoutil/indexes.go` around line 70: The fallback recreates only the old name with the requested keys, discarding the original index's unique, sparse, partial-filter, collation, hidden, and other options. For example, a correctly unique index with a different name can be dropped and then restored as non-unique after a transient recreation failure. It also reuses the failed operation's context, so a deadline or cancellation leaves the collection without any restored index. Capture the complete existing specification before dropping it and restore that specification with an appropriate bounded cleanup context.

```

</details>

media-service/main.go
```diff
@@ -62,7 +62,7 @@ func run() error {
 	defer mongoutil.Disconnect(ctx, mongoClient)
 
 	if err := store.EnsureEmojiIndexes(ctx); err != nil {
-		return fmt.Errorf("ensure emoji indexes: %w", err)
+		slog.Warn("ensure emoji indexes failed; continuing (indexes are best-effort)", "error", err)
```
julianshen, line 65: _🏷️ data-integrity_ | _🔴 Blocking_ | _`silent-failure`_

**Emoji uniqueness failures are converted into successful startup.**

This service solely owns the `(siteId, shortcode)` unique index. If repair fails because of duplicate data, permissions, cancellation, or a MongoDB error, startup now continues and accepts requests without the constraint, allowing multiple emojis to claim the same shortcode. Keep the service unready until the index is verified, or retain fatal startup behavior.

<details>

> ‼️ **IMPORTANT**

```suggestion
return fmt.Errorf("ensure emoji indexes: %w", err)
```

</details>

<details>

```
In `media-service/main.go` around line 65: This service solely owns the `(siteId, shortcode)` unique index. If repair fails because of duplicate data, permissions, cancellation, or a MongoDB error, startup now continues and accepts requests without the constraint, allowing multiple emojis to claim the same shortcode. Keep the service unready until the index is verified, or retain fatal startup behavior.

```

</details>

portal-service/main.go
```diff
@@ -112,7 +112,7 @@ func run() error {
 
 	store := newMongoDirectoryStore(mongoClient.Database(cfg.MongoDB))
 	if err := store.EnsureIndexes(ctx); err != nil {
-		return fmt.Errorf("ensure directory indexes: %w", err)
+		slog.Warn("ensure directory indexes failed; continuing (indexes are best-effort)", "error", err)
```
julianshen, line 115: _🏷️ data-integrity_ | _🔴 Blocking_ | _`silent-failure`_

**Directory index failures are hidden behind successful startup.**

The index is explicitly required to prevent two `hr_employee` documents from publishing different home sites for one account. When creation or repair fails, the service still populates its cache and can become ready with ambiguous directory data. Gate readiness on this owned unique index or fail startup until it is durable.

<details>

> ‼️ **IMPORTANT**

```suggestion
return fmt.Errorf("ensure directory indexes: %w", err)
```

</details>

<details>

```
In `portal-service/main.go` around line 115: The index is explicitly required to prevent two `hr_employee` documents from publishing different home sites for one account. When creation or repair fails, the service still populates its cache and can become ready with ambiguous directory data. Gate readiness on this owned unique index or fail startup until it is durable.

```

</details>

julianshen (commented):
tGD inline review

tcard-service/main.go
```diff
@@ -69,7 +69,7 @@ func run() error {
 
 	store := newMongoCardStore(mongoClient.Database(cfg.MongoDB))
 	if err := store.EnsureIndexes(ctx); err != nil {
-		return fmt.Errorf("ensure cards indexes: %w", err)
+		slog.Warn("ensure cards indexes failed; continuing (indexes are best-effort)", "error", err)
```
julianshen, line 72: _🏷️ data-integrity_ | _🔴 Blocking_ | _`silent-failure`_

**Card uniqueness failures no longer prevent serving writes.**

The `(path, _tcardVersion)` index enforces that only one document claims a card version. Continuing after index creation or repair fails leaves concurrent writes able to create duplicate versions while startup appears successful. Keep readiness false until the index exists, or preserve fatal startup handling.

<details>

> ‼️ **IMPORTANT**

```suggestion
return fmt.Errorf("ensure cards indexes: %w", err)
```

</details>

<details>

```
In `tcard-service/main.go` around line 72: The `(path, _tcardVersion)` index enforces that only one document claims a card version. Continuing after index creation or repair fails leaves concurrent writes able to create duplicate versions while startup appears successful. Keep readiness false until the index exists, or preserve fatal startup handling.

```

</details>

pkg/mongoutil/indexes.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	defer func() { _ = cur.Close(ctx) }()
+
+	have := make(map[string]bool)
+	for cur.Next(ctx) {
+		var idx struct {
+			Name string `bson:"name"`
+		}
+		if err := cur.Decode(&idx); err == nil {
+			have[idx.Name] = true
```
julianshen, line 36: _🏷️ index validation_ | _🟠 Warning_ | _`mongodb`_

**Dependency checks accept an index with the wrong specification.**

WarnMissingIndexes records only index names. A legacy non-unique account_1, or a same-named index with wrong keys/options, is treated as satisfying dependencies that require uniqueness. This is directly reachable in the non-unique account_1 migration state handled elsewhere in this change, and suppresses the only warning while dependent services operate without the expected constraint. Accept expected index models and compare keys plus correctness-relevant options such as unique, partial filter, collation, and hidden state.

<details>

```
In `pkg/mongoutil/indexes.go` around line 35: WarnMissingIndexes records only index names. A legacy non-unique account_1, or a same-named index with wrong keys/options, is treated as satisfying dependencies that require uniqueness. This is directly reachable in the non-unique account_1 migration state handled elsewhere in this change, and suppresses the only warning while dependent services operate without the expected constraint. Accept expected index models and compare keys plus correctness-relevant options such as unique, partial filter, collation, and hidden state.

```

</details>

room-service/store_mongo.go
```diff
@@ -100,167 +100,71 @@ func NewMongoStore(db *mongo.Database, opts ...StoreOption) *MongoStore {
...
+		bson.D{{Key: "u.account", Value: 1}, {Key: "sectionId", Value: 1}, {Key: "sectionOrder", Value: 1}})
+	// Backs getRoomSubscriptions paginated {joinedAt,_id} sort from the index (avoids the 32MB sort limit).
+	create(s.subscriptions, "subscriptions (roomId,joinedAt,_id)",
+		bson.D{{Key: "roomId", Value: 1}, {Key: "joinedAt", Value: 1}, {Key: "_id", Value: 1}})
+	// Backs CountOwners (roomId+roles) index-only.
+	create(s.subscriptions, "subscriptions (roomId,roles)", bson.D{{Key: "roomId", Value: 1}, {Key: "roles", Value: 1}})
+	// room-service owns the thread_subscriptions unique key. Best-effort legacy
+	// (threadRoomId, userId) drop first so the userAccount index creates without a
+	// key conflict; it may not exist (fresh deploy) — ignore all errors.
+	_ = s.threadSubscriptions.Indexes().DropOne(ctx, "threadRoomId_1_userId_1") //nolint:errcheck
```
julianshen: _🏷️ migration_ | _🟠 Warning_ | _`mongodb`_

**A failed legacy-index drop is silently treated as success.**

All errors dropping threadRoomId_1_userId_1 are ignored. If the drop fails for authorization or a transient MongoDB error, the canonical threadRoomId/userAccount index can still be created and EnsureIndexes can return nil while the obsolete unique constraint remains. Because userId is site-local, that stale constraint can reject valid federated subscriptions for different accounts sharing a userId. Ignore only IndexNotFound; collect every other drop error and verify the obsolete index is absent.

<details>

```
In `room-service/store_mongo.go` around line 149: All errors dropping threadRoomId_1_userId_1 are ignored. If the drop fails for authorization or a transient MongoDB error, the canonical threadRoomId/userAccount index can still be created and EnsureIndexes can return nil while the obsolete unique constraint remains. Because userId is site-local, that stale constraint can reject valid federated subscriptions for different accounts sharing a userId. Ignore only IndexNotFound; collect every other drop error and verify the obsolete index is absent.

```

</details>

message-worker/store_mongo.go
```diff
@@ -37,24 +38,16 @@ func newThreadStoreMongo(db *mongo.Database) *threadStoreMongo {
...
-	// The collection or index may not exist (fresh deploy / test container) — ignore all errors.
-	_ = s.threadSubscriptions.Indexes().DropOne(ctx, "threadRoomId_1_userId_1") //nolint:errcheck
-
-	if _, err := s.threadSubscriptions.Indexes().CreateOne(ctx, mongo.IndexModel{
-		Keys:    bson.D{{Key: "threadRoomId", Value: 1}, {Key: "userAccount", Value: 1}},
-		Options: options.Index().SetUnique(true),
-	}); err != nil {
-		return fmt.Errorf("ensure thread_subscriptions (threadRoomId,userAccount) index: %w", err)
-	}
+	// thread_subscriptions.{threadRoomId,userAccount} (unique) is owned by room-service
```
julianshen, line 48: _🏷️ nats-idempotency_ | _🟠 Warning_ | _`nats`_

**The consumer runs without its thread-subscription dedup constraint.**

The message-worker now only warns when `threadRoomId_1_userAccount_1` is absent. Its duplicate-insert test explicitly creates this unique index to exercise `InsertThreadSubscription` deduplication. If room-service has not created the index, a worker crash after insertion but before acknowledging the NATS message allows redelivery to insert the same logical subscription again; both deliveries can then be reported as successful, and the duplicates can prevent the owner from subsequently building the unique index. Keep the NATS consumer unready or do not start consumption until this dependency is verified, while leaving index creation to room-service.

<details>

```
In `message-worker/store_mongo.go` around line 48: The message-worker now only warns when `threadRoomId_1_userAccount_1` is absent. Its duplicate-insert test explicitly creates this unique index to exercise `InsertThreadSubscription` deduplication. If room-service has not created the index, a worker crash after insertion but before acknowledging the NATS message allows redelivery to insert the same logical subscription again; both deliveries can then be reported as successful, and the duplicates can prevent the owner from subsequently building the unique index. Keep the NATS consumer unready or do not start consumption until this dependency is verified, while leaving index creation to room-service.

```

</details>

room-service/main.go
julianshen, line 209: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Index failures leave hot queries unindexed for the process lifetime.**

After a transient failure or the 30-second deadline, the service starts serving traffic without retrying index creation. If indexes such as `(roomId,joinedAt,_id)` are absent, the documented `getRoomSubscriptions` path performs collection scans and in-memory sorts for every page, potentially hitting Mongo's 32 MB sort limit and amplifying database load for large rooms. Preserve non-fatal startup by retrying failed owner indexes in the background with bounded exponential backoff, while keeping readiness false until required query indexes are confirmed.

<details>

```
In `room-service/main.go` around line 209: After a transient failure or the 30-second deadline, the service starts serving traffic without retrying index creation. If indexes such as `(roomId,joinedAt,_id)` are absent, the documented `getRoomSubscriptions` path performs collection scans and in-memory sorts for every page, potentially hitting Mongo's 32 MB sort limit and amplifying database load for large rooms. Preserve non-fatal startup by retrying failed owner indexes in the background with bounded exponential backoff, while keeping readiness false until required query indexes are confirmed.

```

</details>

pkg/mongoutil/indexes.go
chenjr0719: Fixed in `7fa147c4` (narrowed, not a distributed lock). Before the destructive drop, the create is retried once — if a peer already repaired the index, the retry succeeds and we never reach the drop, so we can't drop a peer's freshly-created index. The drop also tolerates `IndexNotFound` (a peer removed it). This closes the common rolling-start race; the tight window where two replicas both re-check before either drops is left to the write-gating follow-up (writers/consumers stay unready until indexes are verified) rather than a distributed lock for a one-time startup convergence.

pkg/mongoutil/indexes.go
chenjr0719: Fixed in `7fa147c4`. Two changes: (1) the old index's **full spec** (unique, sparse, hidden, TTL, partial filter) is captured before the drop and restored from it, so a transient recreate failure can no longer silently weaken the constraint (an old unique index restores as unique). (2) The restore runs on a **fresh bounded context** (`context.WithoutCancel` + a 10s timeout), so a recreate that failed because the caller's context expired doesn't then leave the collection with no index. Collation is the one option not reconstructed — no repaired index in this repo uses it (noted in the doc comment). Test asserts the restored index preserves the old spec.

pkg/mongoutil/indexes.go
chenjr0719: Fixed in `7fa147c4`. Two changes: (1) the old index's **full spec** (unique, sparse, hidden, TTL, partial filter) is captured before the drop and restored from it, so a transient recreate failure can no longer silently weaken the constraint (an old unique index restores as unique). (2) The restore runs on a **fresh bounded context** (`context.WithoutCancel` + a 10s timeout), so a recreate that failed because the caller's context expired doesn't then leave the collection with no index. Collation is the one option not reconstructed — no repaired index in this repo uses it (noted in the doc comment). Test asserts the restored index preserves the old spec.

room-service/store_mongo.go
chenjr0719: Fixed in `7fa147c4`. The legacy `threadRoomId_1_userId_1` drop no longer swallows every error — only `IndexNotFound` is ignored (expected on a fresh deploy); any other drop error is collected into the returned (joined) error so a failed drop of the obsolete unique index surfaces instead of `EnsureIndexes` returning nil.

chenjr0719:
Thanks for the thorough pass. Splitting the findings:

**Fixed in this PR (`7fa147c4`)** — the repair-helper correctness items: concurrent-repair narrowing (retry before the destructive drop + tolerate `IndexNotFound`), faithful restore of the old index's full spec on a fresh context, and collecting (not swallowing) legacy-drop errors. Replied inline on each.

**Deferred to a dedicated readiness-gate follow-up PR** — every "owner/dependent serves writes before its unique index is verified" item (room-service, bot-room-service, media/portal/tcard, the message-worker consumer, the perf-index background retry, and the name-only dependency check). This is the design already agreed on the earlier threads: the service starts and never crashes, but stays **not-ready / defers consuming** until its owned unique indexes are verified — with a background retry. For the NATS/JetStream services the effective lever is deferring subscription/consumption (k8s readiness alone doesn't reroute NATS delivery), which is exactly why it's a focused follow-up rather than a one-line change here.

This PR keeps the log-and-continue + self-heal it set out to deliver; the follow-up adds the write-gating on top.

mliu33 (approved):
Excellent, thanks a lot!

pkg/mongoutil/indexes.go
```diff
@@ -0,0 +1,185 @@
...
+	// A peer replica may have repaired it since the first attempt — retry before the
+	// destructive drop so we never drop a peer's freshly created correct index.
+	if _, err := coll.Indexes().CreateOne(ctx, model); err == nil || !isIndexSpecConflict(err) {
+		return err
+	}
+	old := existingIndexByKeys(ctx, coll, model.Keys)
+	if old.name == "" {
+		return fmt.Errorf("repair index on %s: conflicting index not found", coll.Name())
+	}
+	if err := coll.Indexes().DropOne(ctx, old.name); err != nil && !IsIndexNotFound(err) {
```
julianshen, line 79: _🎯 concurrency_ | _🔴 Blocking_ | _`distributed-system`, `mongodb`, `performance-and-concurrency`, `silent-failure`, `tgd-review` (5 rules)_

**Concurrent replicas can undo a successful index repair.**

Two replicas can both capture the same conflicting auto-named index. Replica A can drop it and recreate the correct index, after which replica B's DropOne(old.name) succeeds by dropping A's newly repaired index because it has the same name. If B's subsequent create is canceled, fails transiently, or encounters duplicates written during this gap, B restores the old conflicting specification, leaving the collection unconstrained despite A reporting success. This is reachable during ordinary multi-replica startup. Serialize the entire recheck/drop/create/restore sequence with a cross-process lock keyed by collection and index keys; the minimal safe alternative is to stop performing destructive automatic repair and return the conflict for controlled migration.

<details>

- **`mongodb`**:75: The repair's list/drop/create sequence is not atomic. Two replicas can both capture the old non-unique index; replica A can replace it with the required unique index, after which replica B's stale `DropOne(old.name)` deletes A's newly created index because it has the same name. If B's subsequent create fails or its startup context expires, the fresh-context restoration recreates the captured non-unique model, leaving the collection downgraded while services continue serving. Serialize destructive repair per collection/index, or move it to a single-run migration; also verify the current spec under a fresh context before restoring so a peer-installed expected index is never replaced. Add a concurrent repair test covering failure after the stale drop.
- **`distributed-system`**:78: Two replicas can both snapshot the same conflicting index. Replica A drops it, creates the required unique index under the same name, and returns success. Replica B then executes DropOne(old.name), which drops A's newly repaired index. If B's context expires, it crashes, or a write inserts duplicates before B recreates the constraint, recreation fails and B restores the captured non-unique model. The cluster is then left without the required uniqueness even though A observed successful repair; long-running callers subsequently log B's error and serve traffic. Serialize destructive repair across replicas with a Mongo-backed lease/fencing mechanism, or perform it as a one-shot migration. A name-based drop after an unfenced snapshot is unsafe.
- **`silent-failure`**:79: Two replicas can both capture the old auto-named index. After replica A drops and recreates `account_1` as unique, replica B's `DropOne(old.name)` can drop A's newly created index because it reuses the same name. If B's subsequent create fails or restores the old non-unique model, A has already returned apparent success although the constraint is gone. Serialize repair per collection/key across replicas, then re-read the current specification under that coordination before dropping it.
- **`tgd-review`**:79: Two replicas can both capture the old index, after which one recreates the correct index before the other executes DropOne by name. The second replica then drops the peer's correct index; if its subsequent CreateOne fails or its context expires, it restores the old conflicting specification under a detached context. The final state can therefore be downgraded even though another replica reported successful repair. Serialize repairs or rework the migration so a stale observation cannot authorize dropping a replacement index, and cover this race in testing.

</details>

<details>

```
In `pkg/mongoutil/indexes.go` around line 79: Two replicas can both capture the same conflicting auto-named index. Replica A can drop it and recreate the correct index, after which replica B's DropOne(old.name) succeeds by dropping A's newly repaired index because it has the same name. If B's subsequent create is canceled, fails transiently, or encounters duplicates written during this gap, B restores the old conflicting specification, leaving the collection unconstrained despite A reporting success. This is reachable during ordinary multi-replica startup. Serialize the entire recheck/drop/create/restore sequence with a cross-process lock keyed by collection and index keys; the minimal safe alternative is to stop performing destructive automatic repair and return the conflict for controlled migration.

```

</details>

admin-service/store_mongo.go
```diff
@@ -30,21 +31,15 @@ func newStoreMongo(db *mongo.Database) *storeMongo {
...
 func (s *storeMongo) EnsureIndexes(ctx context.Context) error {
-	// Matches botplatform's auto-named "account_1" unique index for idempotency on the shared collection.
-	_, err := s.users.Indexes().CreateOne(ctx, mongo.IndexModel{
-		Keys:    bson.D{{Key: "account", Value: 1}},
-		Options: options.Index().SetUnique(true),
-	})
-	if err != nil {
-		return fmt.Errorf("create users account index: %w", err)
-	}
+	// users.account (unique) is owned by user-service; verify + warn only, never create.
```
julianshen, line 34: _🏷️ silent-failure_ | _🔴 Blocking_ | _`silent-failure`_

**User creation succeeds when the account uniqueness index is missing.**

Production now only warns about a missing `account_1` index, while the updated integration test must create that index explicitly to exercise `ErrAccountExists`. If user-service has not created the index or its repair failed, admin-service can insert a second document for the same account without receiving a duplicate-key error and report successful creation. Keep CreateUser unavailable, or keep service readiness false, until the exact unique index is verified.

<details>

```
In `admin-service/store_mongo.go` around line 34: Production now only warns about a missing `account_1` index, while the updated integration test must create that index explicitly to exercise `ErrAccountExists`. If user-service has not created the index or its repair failed, admin-service can insert a second document for the same account without receiving a duplicate-key error and report successful creation. Keep CreateUser unavailable, or keep service readiness false, until the exact unique index is verified.

```

</details>

pkg/mongoutil/indexes.go
```diff
@@ -0,0 +1,185 @@
...
+func isIndexSpecConflict(err error) bool {
+	var se mongo.ServerError
+	return errors.As(err, &se) && (se.HasErrorCode(85) || se.HasErrorCode(86))
+}
+
+// IsIndexNotFound reports whether err is Mongo's IndexNotFound (27), so dropping an
+// index a peer already removed is not treated as a failure.
+func IsIndexNotFound(err error) bool {
+	var se mongo.ServerError
+	return errors.As(err, &se) && se.HasErrorCode(27)
```
julianshen, line 109: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**A missing collection is treated as an index-drop failure.**

MongoDB returns NamespaceNotFound (26) when DropOne targets a collection that does not yet exist. IsIndexNotFound only accepts code 27, so room-service's fresh-database legacy-index cleanup records an error even though the index is necessarily absent. EnsureIndexes then returns that error after successfully creating the new indexes, violating its idempotent fresh-deploy behavior. Treat code 26 as absent too and add a fresh-collection test.

<details>

```
In `pkg/mongoutil/indexes.go` around line 109: MongoDB returns NamespaceNotFound (26) when DropOne targets a collection that does not yet exist. IsIndexNotFound only accepts code 27, so room-service's fresh-database legacy-index cleanup records an error even though the index is necessarily absent. EnsureIndexes then returns that error after successfully creating the new indexes, violating its idempotent fresh-deploy behavior. Treat code 26 as absent too and add a fresh-collection test.

```

</details>

room-service/store_mongo.go
```diff
@@ -100,167 +100,73 @@ func NewMongoStore(db *mongo.Database, opts ...StoreOption) *MongoStore {
...
+		bson.D{{Key: "u.account", Value: 1}, {Key: "sectionId", Value: 1}, {Key: "sectionOrder", Value: 1}})
+	// Backs getRoomSubscriptions paginated {joinedAt,_id} sort from the index (avoids the 32MB sort limit).
+	create(s.subscriptions, "subscriptions (roomId,joinedAt,_id)",
+		bson.D{{Key: "roomId", Value: 1}, {Key: "joinedAt", Value: 1}, {Key: "_id", Value: 1}})
+	// Backs CountOwners (roomId+roles) index-only.
+	create(s.subscriptions, "subscriptions (roomId,roles)", bson.D{{Key: "roomId", Value: 1}, {Key: "roles", Value: 1}})
+	// room-service owns the thread_subscriptions unique key. Best-effort legacy
+	// (threadRoomId, userId) drop first so the userAccount index creates without a
+	// key conflict; it may not exist (fresh deploy) — ignore all errors.
+	if derr := s.threadSubscriptions.Indexes().DropOne(ctx, "threadRoomId_1_userId_1"); derr != nil && !mongoutil.IsIndexNotFound(derr) {
```
julianshen, line 149: _🏷️ data-integrity_ | _🔴 Blocking_ | _`tgd-review`_

**The legacy constraint is dropped before its replacement exists.**

The legacy and canonical indexes have different key specifications, so they can coexist during migration. Dropping threadRoomId_1_userId_1 first means a timeout, duplicate userAccount data, or repair failure leaves the collection without either uniqueness constraint; startup then continues. Ensure the canonical unique index successfully exists before dropping the legacy index.

<details>

```
In `room-service/store_mongo.go` around line 149: The legacy and canonical indexes have different key specifications, so they can coexist during migration. Dropping threadRoomId_1_userId_1 first means a timeout, duplicate userAccount data, or repair failure leaves the collection without either uniqueness constraint; startup then continues. Ensure the canonical unique index successfully exists before dropping the legacy index.

```

</details>

julianshen (commented):
tGD inline review

chenjr0719:
Thanks for the post-merge re-review. All four points are addressed in the follow-up **#334** and the planned readiness-gate PR:

- **NamespaceNotFound(26) treated as a drop failure** (`indexes.go`) — fixed in #334: `IsIndexNotFound` now matches both 26 and 27, so a fresh-DB legacy-index drop is a tolerated no-op. This is exactly what red-mained `main` after the merge; #334 restores green. Added a fresh-collection test.
- **Legacy index dropped before its replacement exists** (`room-service`) — fixed in #334: the canonical `(threadRoomId, userAccount)` index is created first (different keys from the legacy `(threadRoomId, userId)`, so they coexist), and the legacy is dropped only once the replacement exists.
- **Concurrent-replica repair race** (`indexes.go`) — narrowed here (retry-before-drop + tolerate IndexNotFound) and documented as a residual; full serialization / write-safety is handled by the readiness-gate follow-up (writers/consumers stay unready until owned unique indexes are verified), rather than a distributed lock for a one-time startup convergence.
- **admin-service serves without the unique index** — part of the readiness-gate follow-up (owner/dependent gating).

PR #334, opened 2026-08-21
fix(mongo-indexes): tolerate NamespaceNotFound on legacy-index drop (room-service main-red fix)

chenjr0719:
## Summary

Rebased on #338 (merged), which fixed the `sast`/gosec TTL narrowing. This PR now carries **only** the *other* main-red regression from #333 — the `room-service` `test-integration` failure — so #338 + this together restore `main` to green.

## Changes

- **`IsIndexNotFound` also matches NamespaceNotFound (26), not only IndexNotFound (27).** On a fresh
  database the `thread_subscriptions` collection doesn't exist yet, so the best-effort legacy-index
  `DropOne` returns NamespaceNotFound. The stricter drop-error handling in #333 collected it, so
  `room-service`'s `EnsureIndexes` returned a non-nil error and the integration-test setup asserts failed.
  Both codes mean "there is no index to drop"; a real auth/transient drop error still surfaces.
- **Create the canonical `(threadRoomId, userAccount)` index before dropping the legacy `(threadRoomId, userId)` one.**
  They have different keys and coexist, so there's no reason to drop first — and doing so means a failed
  create leaves the collection with neither uniqueness constraint. Drop the legacy only once the replacement exists.
- Fresh-collection test asserting a NamespaceNotFound drop counts as not-found.

## Verification

- `go build`, `go vet -tags integration`, golangci-lint, and `gosec` all clean locally.
- The `room-service` test-integration suite (red on `main` after #333) is the regression coverage — it
  calls `EnsureIndexes` against fresh databases.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved database index migration reliability by creating replacement indexes before removing legacy ones.
  - Preserved existing indexes when replacement creation fails, preventing unintended data access disruptions.
  - Database errors are now reported more accurately when an index or namespace is unavailable.

- **Tests**
  - Added coverage for failed index replacement scenarios, including duplicate data conflicts and preservation of the existing index.

### DIFF
--- a/pkg/mongoutil/indexes.go
+++ b/pkg/mongoutil/indexes.go
@@ -103,13 +103,14 @@ func isIndexSpecConflict(err error) bool {
 	return errors.As(err, &se) && (se.HasErrorCode(85) || se.HasErrorCode(86))
 }
 
-// IsIndexNotFound reports whether err means there was no index to drop — Mongo's
-// IndexNotFound (27) or NamespaceNotFound (26, the collection doesn't exist yet) —
-// so dropping an absent index (fresh deploy, or a peer already removed it) is not
-// treated as a failure.
+// IsIndexNotFound reports whether err is Mongo's IndexNotFound (27), so dropping an
+// index a peer already removed is not treated as a failure. NamespaceNotFound (26)
+// is deliberately NOT folded in: callers drop only after ensuring an index on the
+// same collection, so the collection exists — a 26 there means an external actor
+// dropped it, which must surface rather than pass as "nothing to drop".
 func IsIndexNotFound(err error) bool {
 	var se mongo.ServerError
-	return errors.As(err, &se) && (se.HasErrorCode(26) || se.HasErrorCode(27))
+	return errors.As(err, &se) && se.HasErrorCode(27)
 }
 
 // existingIndex is a pre-existing index captured for a faithful restore.
--- a/pkg/mongoutil/indexes.go
+++ b/pkg/mongoutil/indexes.go
@@ -103,11 +103,13 @@ func isIndexSpecConflict(err error) bool {
 	return errors.As(err, &se) && (se.HasErrorCode(85) || se.HasErrorCode(86))
 }
 
-// IsIndexNotFound reports whether err is Mongo's IndexNotFound (27), so dropping an
-// index a peer already removed is not treated as a failure.
+// IsIndexNotFound reports whether err means there was no index to drop — Mongo's
+// IndexNotFound (27) or NamespaceNotFound (26, the collection doesn't exist yet) —
+// so dropping an absent index (fresh deploy, or a peer already removed it) is not
+// treated as a failure.
 func IsIndexNotFound(err error) bool {
 	var se mongo.ServerError
-	return errors.As(err, &se) && se.HasErrorCode(27)
+	return errors.As(err, &se) && (se.HasErrorCode(26) || se.HasErrorCode(27))
 }
 
 // existingIndex is a pre-existing index captured for a faithful restore.
--- a/room-service/integration_test.go
+++ b/room-service/integration_test.go
@@ -4293,6 +4293,46 @@ func TestEnsureIndexes_ThreadSubsDropsLegacyAndCreatesUnique_Integration(t *test
 		"EnsureIndexes must drop the legacy (threadRoomId, userId) index")
 }
 
+// Create-before-drop: a FAILED canonical create must leave the legacy index intact,
+// so the collection is never left without either uniqueness constraint. Would fail
+// if the order were reverted to drop-legacy-then-create.
+func TestEnsureIndexes_ThreadSubsCanonicalCreateFailure_KeepsLegacy_Integration(t *testing.T) {
+	db := setupMongo(t)
+	store := NewMongoStore(db)
+	ctx := context.Background()
+
+	// Legacy index present, plus duplicate (threadRoomId, userAccount) rows so the
+	// canonical UNIQUE create fails with E11000.
+	_, err := db.Collection("thread_subscriptions").Indexes().CreateOne(ctx, mongo.IndexModel{
+		Keys: bson.D{{Key: "threadRoomId", Value: 1}, {Key: "userId", Value: 1}},
+	})
+	require.NoError(t, err)
+	_, err = db.Collection("thread_subscriptions").InsertMany(ctx, []any{
+		bson.M{"_id": "a", "threadRoomId": "tr1", "userAccount": "dup"},
+		bson.M{"_id": "b", "threadRoomId": "tr1", "userAccount": "dup"},
+	})
+	require.NoError(t, err)
+
+	// EnsureIndexes surfaces the canonical failure...
+	require.Error(t, store.EnsureIndexes(ctx))
+
+	// ...and must NOT have dropped the legacy index (the create ran first and failed).
+	cur, err := db.Collection("thread_subscriptions").Indexes().List(ctx)
+	require.NoError(t, err)
+	var idxs []bson.M
+	require.NoError(t, cur.All(ctx, &idxs))
+	names := make(map[string]bool, len(idxs))
+	for _, ix := range idxs {
+		if n, ok := ix["name"].(string); ok {
+			names[n] = true
+		}
+	}
+	assert.True(t, names["threadRoomId_1_userId_1"],
+		"a failed canonical create must leave the legacy index intact")
+	assert.False(t, names["threadRoomId_1_userAccount_1"],
+		"the canonical unique index must not exist when its create failed")
+}
+
 func TestMongoStore_ClearThreadSubscriptionsForAccount_Integration(t *testing.T) {
 	db := setupMongo(t)
 	store := NewMongoStore(db)
--- a/room-service/integration_test.go
+++ b/room-service/integration_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.mongodb.org/mongo-driver/v2/bson"
 	"go.mongodb.org/mongo-driver/v2/mongo"
+	"go.mongodb.org/mongo-driver/v2/mongo/options"
 	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/trace/noop"
 
@@ -4301,15 +4302,17 @@ func TestEnsureIndexes_ThreadSubsCanonicalCreateFailure_KeepsLegacy_Integration(
 	store := NewMongoStore(db)
 	ctx := context.Background()
 
-	// Legacy index present, plus duplicate (threadRoomId, userAccount) rows so the
-	// canonical UNIQUE create fails with E11000.
+	// Legacy UNIQUE index present (models the real constraint the migration preserves).
 	_, err := db.Collection("thread_subscriptions").Indexes().CreateOne(ctx, mongo.IndexModel{
-		Keys: bson.D{{Key: "threadRoomId", Value: 1}, {Key: "userId", Value: 1}},
+		Keys:    bson.D{{Key: "threadRoomId", Value: 1}, {Key: "userId", Value: 1}},
+		Options: options.Index().SetUnique(true),
 	})
 	require.NoError(t, err)
+	// Distinct (threadRoomId, userId) so both rows coexist under the legacy unique index,
+	// but the SAME (threadRoomId, userAccount) so the canonical UNIQUE create fails E11000.
 	_, err = db.Collection("thread_subscriptions").InsertMany(ctx, []any{
-		bson.M{"_id": "a", "threadRoomId": "tr1", "userAccount": "dup"},
-		bson.M{"_id": "b", "threadRoomId": "tr1", "userAccount": "dup"},
+		bson.M{"_id": "a", "threadRoomId": "tr1", "userId": "u1", "userAccount": "dup"},
+		bson.M{"_id": "b", "threadRoomId": "tr1", "userId": "u2", "userAccount": "dup"},
 	})
 	require.NoError(t, err)
 
--- a/room-service/store_mongo.go
+++ b/room-service/store_mongo.go
@@ -143,14 +143,19 @@ func (s *MongoStore) EnsureIndexes(ctx context.Context) error {
 		bson.D{{Key: "roomId", Value: 1}, {Key: "joinedAt", Value: 1}, {Key: "_id", Value: 1}})
 	// Backs CountOwners (roomId+roles) index-only.
 	create(s.subscriptions, "subscriptions (roomId,roles)", bson.D{{Key: "roomId", Value: 1}, {Key: "roles", Value: 1}})
-	// room-service owns the thread_subscriptions unique key. Best-effort legacy
-	// (threadRoomId, userId) drop first so the userAccount index creates without a
-	// key conflict; it may not exist (fresh deploy) — ignore all errors.
-	if derr := s.threadSubscriptions.Indexes().DropOne(ctx, "threadRoomId_1_userId_1"); derr != nil && !mongoutil.IsIndexNotFound(derr) {
+	// room-service owns the thread_subscriptions unique key. Create the canonical
+	// (threadRoomId, userAccount) index FIRST — it has different keys from the legacy
+	// (threadRoomId, userId) index, so they coexist — then drop the legacy only once
+	// the replacement exists, so a failed create never leaves the collection without
+	// either constraint. IndexNotFound/NamespaceNotFound on the drop is a no-op (fresh deploy).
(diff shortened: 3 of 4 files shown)


julianshen:
**No actionable comments.** ✅

### Related work

- [PR #338](https://github.com/hmchangw/newchat/pull/338) — fix\(mongo-indexes\): bound a captured TTL before narrowing it to int32 (merged)
- [PR #333](https://github.com/hmchangw/newchat/pull/333) — fix\(mongo-indexes\): non-fatal index ensure \+ single-owner shared indexes (merged)

### Local review memory

- `room-service/store_mongo.go:157` — @chenjr0719: Good catch — fixed in ee758343 by reverting to IndexNotFound(27) only. You're right that a 26 in this post-create path is an anomaly (an external actor dropped the collection between the create and the drop) and must surface, not pass… ([thread](https://github.com/hmchangw/newchat/pull/334#discussion_r3827500269))
- `room-service/store_mongo.go:154` — @chenjr0719: Added in ee758343 : TestEnsureIndexes_ThreadSubsCanonicalCreateFailure_KeepsLegacy_Integration — it seeds the legacy (threadRoomId, userId) index plus duplicate (threadRoomId, userAccount) rows so the canonical UNIQUE create fails E… ([thread](https://github.com/hmchangw/newchat/pull/334#discussion_r3827500274))
- `room-service/integration_test.go:4306` — @chenjr0719: Applied in 7809ac77 . The legacy (threadRoomId, userId) index is now created **unique**, and the two rows carry distinct userId ( u1 / u2 ) so they coexist under it — while keeping the same (threadRoomId, userAccount) so the canonic… ([thread](https://github.com/hmchangw/newchat/pull/334#discussion_r3827567294))

<details>

* `pkg/mongoutil/indexes.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

pkg/mongoutil/indexes.go
```diff
@@ -131,11 +133,8 @@ func (e existingIndex) model() mongo.IndexModel {
 	if v, _ := e.spec["hidden"].(bool); v {
 		opts = opts.SetHidden(true)
 	}
-	switch v := e.spec["expireAfterSeconds"].(type) {
-	case int32:
+	if v, ok := e.spec["expireAfterSeconds"].(int32); ok {
 		opts = opts.SetExpireAfterSeconds(v)
```
julianshen: _🏷️ MongoDB TTL indexes_ | _🔴 Blocking_ | _`distributed-system`, `mongodb`, `silent-failure`, `tgd-review` (4 rules)_

**Restored int64 TTL indexes silently lose expiration**

The restore path now applies `expireAfterSeconds` only when `listIndexes` decoded it as `int32`. MongoDB can return this field as BSON int64, and this code previously handled that representation. For such an index, `model()` now recreates a normal non-TTL index, so expired documents remain indefinitely and retention/storage guarantees are violated. Preserve the int64 case for values representable by the driver's int32 option, and reject or explicitly handle out-of-range values rather than silently dropping TTL semantics.

<details>

- **`distributed-system`**:136: Mongo can return `expireAfterSeconds` as BSON int64, especially for indexes created by other clients. After this change, `existingIndex.model()` silently omits that option. Concrete failure path: an existing int64-encoded TTL index is captured and dropped during replacement, creation of the new index then fails, and the recovery path recreates the captured index through `model()`. Recovery succeeds, but the restored index is now non-TTL, so data stops expiring even though the caller observed only the original operation failure. Preserve the int64 case, with bounds validation before converting to the driver's int32 option type.
- **`silent-failure`**:136: When a captured `expireAfterSeconds` value is represented as BSON `int64`—a case explicitly handled before this change—the type assertion now fails and the option is omitted. Index recreation can therefore report success while producing a normal index instead of a TTL index, leaving data past its intended retention period. Preserve the `int64` case (and validate bounds if values are not guaranteed to fit `int32`).
- **`tgd-review`**:136: BSON preserves integer width, so an existing index can expose `expireAfterSeconds` as `int64`. Removing that case causes `model()` to restore such an index without TTL settings, potentially retaining documents indefinitely. Preserve both BSON integer representations; if overflow is a concern, validate the `int64` before conversion rather than ignoring it.

</details>

<details>

```
In `pkg/mongoutil/indexes.go` around line 137: The restore path now applies `expireAfterSeconds` only when `listIndexes` decoded it as `int32`. MongoDB can return this field as BSON int64, and this code previously handled that representation. For such an index, `model()` now recreates a normal non-TTL index, so expired documents remain indefinitely and retention/storage guarantees are violated. Preserve the int64 case for values representable by the driver's int32 option, and reject or explicitly handle out-of-range values rather than silently dropping TTL semantics.

```

</details>

julianshen (commented):
tGD inline review

pkg/mongoutil/indexes.go
chenjr0719: Fixed in `c3619f91`. Restored the int64 case with a `0 <= v <= math.MaxInt32` bounds check — that preserves the TTL when `listIndexes` decodes `expireAfterSeconds` as int64, and the guard also satisfies gosec G115 (the narrowing can no longer overflow). An out-of-range window (> ~68 years) is skipped rather than truncated, since it is not a real retention TTL. (Note: no index this helper repairs currently carries a TTL, but the restore is now faithful for one that does.)

pkg/mongoutil/indexes.go
```diff
@@ -131,11 +133,8 @@ func (e existingIndex) model() mongo.IndexModel {
 	if v, _ := e.spec["hidden"].(bool); v {
 		opts = opts.SetHidden(true)
 	}
-	switch v := e.spec["expireAfterSeconds"].(type) {
-	case int32:
+	if v, ok := e.spec["expireAfterSeconds"].(int32); ok {
```
julianshen: _🏷️ silent-failure_ | _🔴 Blocking_ | _`distributed-system`, `mongodb`, `silent-failure`, `tgd-review` (4 rules)_

**Int64 TTL indexes are silently restored without expiration**

This remains unresolved: `listIndexes` may decode `expireAfterSeconds` as `int64`, which this type assertion now ignores. If index replacement fails after dropping an existing TTL index, recovery can recreate it successfully as a non-TTL index, causing documents to remain indefinitely while hiding the incomplete restoration. Preserve the `int64` case when it fits the driver's `int32` option and surface an error for out-of-range values rather than omitting TTL semantics.

<details>

- **`distributed-system`**:136: The prior concern remains unresolved. MongoDB may return `expireAfterSeconds` as BSON int64. If index replacement fails after dropping the existing index, the recovery path calls `model()`; this assertion then fails, recreating the index without TTL semantics while recovery appears successful. Documents consequently stop expiring. Preserve the int64 representation; valid MongoDB TTL values fit the driver's int32 option.
- **`tgd-review`**:136: BSON preserves integer width, so `expireAfterSeconds` may be decoded as `int64`; the previous implementation explicitly handled that representation. The restore path now omits the TTL option for such indexes, recreating them as ordinary indexes and allowing expired data to remain indefinitely. Preserve the `int64` case, validate that it fits the driver's `int32` option, and add a restoration regression test.
- **`mongodb`**:136: BSON preserves integer width, and `listIndexes` may decode a valid `expireAfterSeconds` as `int64`. During index repair, an existing TTL index is captured, dropped, and potentially restored through `model()`; this change omits the TTL option for that representation, recreating a normal index and retaining expired documents indefinitely. Restore support for both integer representations.

</details>

<details>

```
In `pkg/mongoutil/indexes.go` around line 136: This remains unresolved: `listIndexes` may decode `expireAfterSeconds` as `int64`, which this type assertion now ignores. If index replacement fails after dropping an existing TTL index, recovery can recreate it successfully as a non-TTL index, causing documents to remain indefinitely while hiding the incomplete restoration. Preserve the `int64` case when it fits the driver's `int32` option and surface an error for out-of-range values rather than omitting TTL semantics.

```

</details>

julianshen (commented):
tGD inline review

pkg/mongoutil/indexes.go
chenjr0719: This was against the intermediate commit `49f243c5`; the current head `c3619f91` already restores the bounded int64 case:

```go
case int64:
    if v >= 0 && v <= math.MaxInt32 {
        opts = opts.SetExpireAfterSeconds(int32(v))
    }
```

So an int64-decoded `expireAfterSeconds` in range is preserved (and the bounds check satisfies gosec G115). An out-of-range value (> ~68 years) is skipped rather than surfaced as an error — it is not a real retention window, and note this restore path only runs for an index this helper repairs, none of which carry a TTL. Happy to make out-of-range an explicit error if you prefer, but it is an unreachable degenerate here.

julianshen (commented):
tGD inline review

room-service/store_mongo.go
```diff
@@ -143,14 +143,19 @@ func (s *MongoStore) EnsureIndexes(ctx context.Context) error {
...
+	// (threadRoomId, userAccount) index FIRST — it has different keys from the legacy
+	// (threadRoomId, userId) index, so they coexist — then drop the legacy only once
+	// the replacement exists, so a failed create never leaves the collection without
+	// either constraint. IndexNotFound/NamespaceNotFound on the drop is a no-op (fresh deploy).
+	if cerr := mongoutil.EnsureIndexWithRepair(ctx, s.threadSubscriptions, mongo.IndexModel{
+		Keys:    bson.D{{Key: "threadRoomId", Value: 1}, {Key: "userAccount", Value: 1}},
+		Options: options.Index().SetUnique(true),
+	}); cerr != nil {
+		errs = append(errs, fmt.Errorf("ensure thread_subscriptions (threadRoomId,userAccount) unique index: %w", cerr))
+	} else if derr := s.threadSubscriptions.Indexes().DropOne(ctx, "threadRoomId_1_userId_1"); derr != nil && !mongoutil.IsIndexNotFound(derr) {
```
julianshen, line 157: _🎯 distributed correctness_ | _🟠 Warning_ | _`distributed-system`_

**Namespace loss can be mistaken for a successful index migration**

This call now suppresses NamespaceNotFound (26), but the canonical index was successfully ensured immediately beforehand. A concrete race is: this process creates the canonical unique index; another actor drops the collection; DropOne returns 26 and is ignored; the subsequent index creations recreate the collection without the canonical unique index; EnsureIndexes can then return success and duplicate subscriptions can be accepted. Treat code 26 as an error in this post-create drop path (or re-run/revalidate the canonical index after it); only code 27 is a safe no-op here.

<details>

```
In `room-service/store_mongo.go` around line 156: This call now suppresses NamespaceNotFound (26), but the canonical index was successfully ensured immediately beforehand. A concrete race is: this process creates the canonical unique index; another actor drops the collection; DropOne returns 26 and is ignored; the subsequent index creations recreate the collection without the canonical unique index; EnsureIndexes can then return success and duplicate subscriptions can be accepted. Treat code 26 as an error in this post-create drop path (or re-run/revalidate the canonical index after it); only code 27 is a safe no-op here.

```

</details>

room-service/store_mongo.go
```diff
@@ -143,14 +143,19 @@ func (s *MongoStore) EnsureIndexes(ctx context.Context) error {
...
-	// key conflict; it may not exist (fresh deploy) — ignore all errors.
-	if derr := s.threadSubscriptions.Indexes().DropOne(ctx, "threadRoomId_1_userId_1"); derr != nil && !mongoutil.IsIndexNotFound(derr) {
+	// room-service owns the thread_subscriptions unique key. Create the canonical
+	// (threadRoomId, userAccount) index FIRST — it has different keys from the legacy
+	// (threadRoomId, userId) index, so they coexist — then drop the legacy only once
+	// the replacement exists, so a failed create never leaves the collection without
+	// either constraint. IndexNotFound/NamespaceNotFound on the drop is a no-op (fresh deploy).
+	if cerr := mongoutil.EnsureIndexWithRepair(ctx, s.threadSubscriptions, mongo.IndexModel{
+		Keys:    bson.D{{Key: "threadRoomId", Value: 1}, {Key: "userAccount", Value: 1}},
+		Options: options.Index().SetUnique(true),
```
julianshen, line 154: _🧪 tests_ | _🟠 Warning_ | _`tgd-review`_

**The create-before-drop migration path lacks regression coverage**

The added integration test only verifies that NamespaceNotFound is classified as not-found. It would still pass if this migration were reverted to dropping the legacy index before creating its replacement. Add coverage that starts with the legacy index, forces canonical index creation to fail (for example via duplicate canonical keys), calls EnsureIndexes, and verifies the legacy index remains; also cover successful creation followed by removal of the legacy index.

<details>

```
In `room-service/store_mongo.go` around line 153: The added integration test only verifies that NamespaceNotFound is classified as not-found. It would still pass if this migration were reverted to dropping the legacy index before creating its replacement. Add coverage that starts with the legacy index, forces canonical index creation to fail (for example via duplicate canonical keys), calls EnsureIndexes, and verifies the legacy index remains; also cover successful creation followed by removal of the legacy index.

```

</details>

room-service/store_mongo.go
chenjr0719: Good catch — fixed in `ee758343` by reverting to `IndexNotFound(27)` only. You're right that a 26 in this post-create path is an anomaly (an external actor dropped the collection between the create and the drop) and must surface, not pass as "nothing to drop". And it turns out the 26 tolerance was also redundant: the **reorder** (create the canonical index first) is what actually fixes the original fresh-DB regression — post-create the collection exists, so an absent legacy index is `IndexNotFound (27)`, which was already tolerated. So this PR is now just the ordering fix, with 27-only restored (and a comment documenting why 26 is deliberately excluded).

room-service/store_mongo.go
chenjr0719: Added in `ee758343`: `TestEnsureIndexes_ThreadSubsCanonicalCreateFailure_KeepsLegacy_Integration` — it seeds the legacy `(threadRoomId, userId)` index plus duplicate `(threadRoomId, userAccount)` rows so the canonical UNIQUE create fails E11000, calls `EnsureIndexes`, and asserts the **legacy index remains** (and the canonical is absent). That fails if the order is reverted to drop-legacy-then-create. The success path (create → legacy dropped) stays covered by `TestEnsureIndexes_ThreadSubsDropsLegacyAndCreatesUnique_Integration`.

chenjr0719:
Note: updated in `ee758343` per @julianshen's review — the fix is now the **create-before-drop reorder alone**, and `IsIndexNotFound` is reverted to `IndexNotFound(27)` only (a NamespaceNotFound after a successful create means an external actor dropped the collection and must surface). The PR summary above still mentions the earlier NamespaceNotFound(26) tolerance — that has been superseded; the reorder is what fixes the fresh-DB regression. Added a test that a failed canonical create leaves the legacy index intact.

room-service/integration_test.go
```diff
@@ -4293,6 +4293,46 @@ func TestEnsureIndexes_ThreadSubsDropsLegacyAndCreatesUnique_Integration(t *test
...
+// so the collection is never left without either uniqueness constraint. Would fail
+// if the order were reverted to drop-legacy-then-create.
+func TestEnsureIndexes_ThreadSubsCanonicalCreateFailure_KeepsLegacy_Integration(t *testing.T) {
+	db := setupMongo(t)
+	store := NewMongoStore(db)
+	ctx := context.Background()
+
+	// Legacy index present, plus duplicate (threadRoomId, userAccount) rows so the
+	// canonical UNIQUE create fails with E11000.
+	_, err := db.Collection("thread_subscriptions").Indexes().CreateOne(ctx, mongo.IndexModel{
```
julianshen, line 4306: _🧪 tests_ | _🔵 Suggestion_ | _`mongodb`_

**The migration test does not create the legacy unique index**

The fixture creates `(threadRoomId, userId)` as a non-unique index, so it does not model the uniqueness constraint that the migration promises to preserve. Both documents also omit `userId`; they could not coexist under the real unique compound index because MongoDB indexes both missing values as `null`. Create the legacy index as unique and give the documents distinct legacy IDs while retaining duplicate canonical keys.

<details>

```
In `room-service/integration_test.go` around line 4306: The fixture creates `(threadRoomId, userId)` as a non-unique index, so it does not model the uniqueness constraint that the migration promises to preserve. Both documents also omit `userId`; they could not coexist under the real unique compound index because MongoDB indexes both missing values as `null`. Create the legacy index as unique and give the documents distinct legacy IDs while retaining duplicate canonical keys.

```

</details>

julianshen (commented):
tGD inline review

room-service/integration_test.go
chenjr0719: Applied in `7809ac77`. The legacy `(threadRoomId, userId)` index is now created **unique**, and the two rows carry distinct `userId` (`u1`/`u2`) so they coexist under it — while keeping the same `(threadRoomId, userAccount)` so the canonical unique create still fails E11000. Good call; the non-unique fixture with `userId` omitted didn't model the constraint the migration preserves.

mliu33 (approved):

PR #338, opened 2026-08-21
fix(mongo-indexes): bound a captured TTL before narrowing it to int32

hmchangw:
The `sast` job has been red on `main` since #333. gosec fails the gate on one finding:

```
[pkg/mongoutil/indexes.go:138] - G115 (CWE-190): integer overflow conversion int64 -> int32
    137: 	case int64:
  > 138: 		opts = opts.SetExpireAfterSeconds(int32(v))
```

It is a real defect, not a false positive. `existingIndex.model` rebuilds the index that `EnsureIndexWithRepair` restores after a failed repair, from the spec Mongo handed back — where `expireAfterSeconds` arrives as either `int32` or `int64`. An `int64` outside int32 range wrapped silently, and a wrapped value does not fail the restore: it recreates the index carrying a TTL the original never had, so documents start expiring on a schedule nobody chose. That is the opposite of what the restore path exists to do.

## Changes

- `ttlSeconds` reports anything outside `[0, MaxInt32]` as unusable instead of narrowing it; only an in-range value reaches `SetExpireAfterSeconds`.
- `model` takes a `ctx` and logs the dropped TTL, so a restore that loses one is visible rather than silent. Previously a negative or oversized value either wrapped or was dropped with no trace.
- New `pkg/mongoutil/indexes_test.go` covers both numeric widths, the int32-max boundary, out-of-range and negative values, a non-numeric spec, and the untouched unique / sparse / hidden / partial-filter options.

The test was written first and confirmed failing against the old conversion (the out-of-range case restored a wrapped `-2147483648`); `model` and `ttlSeconds` are at 100% statement coverage.

## Verification

- `gosec -quiet -severity medium -confidence medium -tests=true -exclude-generated -exclude-dir=tools -exclude-dir=testdata ./...` — exit 0, no findings (was 2).
- `make test` (`go test -race ./...`) — exit 0.
- `golangci-lint run ./...` — 0 issues; `go vet -tags integration ./pkg/mongoutil/...` clean.
- semgrep local ruleset (`.semgrep/`) — 0 findings. The registry rulesets (`p/golang`, `p/security-audit`) and govulncheck could not run in this sandbox — `semgrep.dev` and `vuln.go.dev` are blocked by the environment's network policy. Both passed on this base commit in CI and this change adds no dependencies; the `sast` job re-runs them here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01V7FYdt6SF3GRTVwzeS8Thj)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved MongoDB index restoration reliability by safely validating TTL settings.
  - Unsupported, negative, or out-of-range TTL values are now skipped with a warning instead of causing invalid restoration.
  - Valid TTL values are restored accurately.
  - Preserved index names, keys, and configured options while omitting unset options.

### DIFF
--- a/pkg/mongoutil/indexes.go
+++ b/pkg/mongoutil/indexes.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"math"
 	"strings"
 	"time"
 
@@ -84,7 +85,7 @@ func EnsureIndexWithRepair(ctx context.Context, coll *mongo.Collection, model mo
 		// collection keeps an index; the returned error still carries the cause.
 		restoreCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), indexRestoreTimeout)
 		defer cancel()
-		if _, rerr := coll.Indexes().CreateOne(restoreCtx, old.model()); rerr != nil {
+		if _, rerr := coll.Indexes().CreateOne(restoreCtx, old.model(ctx)); rerr != nil {
 			slog.WarnContext(ctx, "mongo: index repair failed and the old index could not be restored",
 				"collection", coll.Name(), "index", old.name, "error", rerr)
 		}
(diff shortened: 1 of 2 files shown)


chenjr0719 (approved):
LGTM — clean and well-targeted. Extracting `ttlSeconds` with the explicit `[0, MaxInt32]` bound is the right call (an out-of-range int64 previously wrapped and restored a TTL the index never had), and the table test covers both widths + the boundary/out-of-range/negative/non-numeric cases at full coverage. The `WarnContext` on a dropped TTL makes the loss visible instead of silent. gosec-clean. Nice fix.

mliu33 (approved):

PR #341, opened 2026-08-21
Remove mention resolution metrics from notification-worker

evelynchang5525:
## Summary

Follow-up to review feedback on #297 (already merged): `recordMentionResults` is unnecessary, and redundant metrics cost work on the hot path. Mention resolution now reports nothing to metrics at all.

## What changed

**Removed entirely:**

- `Handler.recordMentionResults` (`notification-worker/handler.go`) — looped over every looked-up account and issued up to two counter adds per mention-bearing message
- `notification_worker_mention_resolution_total{result="resolved|unresolved|failed"}`, plus the `mentionResult` enum, `allMentionResults`, the `mentionOpts` attribute map, and `RecordMentions`
- `notification_worker_mention_lookup_failures_total`, the `mentionFailures` instrument, and `RecordMentionFailure`

`notificationMetrics` is back to the single `outcomes` counter it had before mention substitution landed, so the mention path adds no metric work to the hot path in any case — success, unresolved, or failure.

## What reports a failure now

A lookup that errors or times out is visible through the existing warn line, `mention name lookup failed, body keeps raw mentions`, which is trace-correlated (`slog.WarnContext`) and carries the requested and resolved counts plus the request ID. The message itself still records as `notification_worker_outcomes_total{result="sent"}`, because the push *is* delivered — with raw `@tokens` in the body.

Noted in the contract doc so operators know to look at logs rather than hunt for a metric that does not exist.

## Testing

- Removed `TestRecordMentionResults_PartitionsTokens` and the `collectMentionResults` / `collectMentionFailures` helpers; no replacement, since there is no metric left to assert on. Behavioural coverage of the fail-open path (partial resolution, unresolved tokens kept as `@token`, push still emitted) is unchanged and still passing.
- `make test` (full suite, `-race`) passes; `make lint` 0 issues; `make sast-gosec` clean; integration tests compile under `-tags integration`.

https://claude.ai/code/session_01HfHNt5BXCQ9fG5bzAKXruh

## Summary by CodeRabbit

* **Monitoring**
  * Mention lookup failures are no longer recorded as dedicated metrics.
  * Warning logs include requested and resolved mention counts for lookup errors and timeouts.
  * Partially resolved names remain substituted, and sent notifications continue to be recorded as successful outcomes.

* **Documentation**
  * Clarified the behavior of partial mention resolution and updated operational guidance to reference warning logs instead of failure metrics.


julianshen:

### Needs clarification

Has every deployed alert using this documented metric been migrated or removed?

`answer clar_pm255wc33aemk46fv7jvhyqa3k: <your answer>`

[Open the question](https://github.com/hmchangw/newchat/pull/341#issuecomment-5384744287)

### Related work

- [PR #297](https://github.com/hmchangw/newchat/pull/297) — Substitute @mentions with display names in push notification bodies (merged)

### Local review memory

- `notification-worker/nats_metrics.go:49` — @evelynchang5525: This removal is deliberate and maintainer-directed, not an oversight — so I'm not restoring the counter, but the alert-migration point deserves a straight answer. **Provenance.** The metric was added in #297 and removed here at @mliu33's r… ([thread](https://github.com/hmchangw/newchat/pull/341#discussion_r3836605485))
- `notification-worker/handler.go:437` — @evelynchang5525: Same finding as the one on nats_metrics.go:49 — answered in full there: the metric removal was requested by @mliu33 in review of #297 and reaffirmed when I proposed keeping a failure-only counter, so it stands as a deliberate decision ra… ([thread](https://github.com/hmchangw/newchat/pull/341#discussion_r3836605488))

<details>

* `notification-worker/nats_metrics_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

notification-worker/nats_metrics.go
```diff
@@ -67,16 +49,9 @@ func newNotificationMetrics(meter metric.Meter) *notificationMetrics {
 		// the service runs blind on this metric rather than not at all.
```
julianshen, line 49: _🏷️ observability_ | _🟠 Warning_ | _`distributed-system`_

**Removing the failure counter makes lookup outages appear healthy.**

Concrete failure path: during or after rollout, the users lookup begins timing out; `resolveBody` fails open and pushes raw `@tokens`, while `notification_worker_outcomes_total` records `sent`. New workers emit no `notification_worker_mention_resolution_total{result="failed"}`, so the previously documented metric alert undercounts failures during mixed-version rollout and then receives no series at all, which common threshold alerts treat as healthy. Per-message warning logs do not preserve the existing metric contract or automatically migrate that alert. Retain the metric until monitoring has been explicitly migrated, or add an equivalent failure counter and alert before removing the old series.

<details>

```
In `notification-worker/nats_metrics.go` around line 49: Concrete failure path: during or after rollout, the users lookup begins timing out; `resolveBody` fails open and pushes raw `@tokens`, while `notification_worker_outcomes_total` records `sent`. New workers emit no `notification_worker_mention_resolution_total{result="failed"}`, so the previously documented metric alert undercounts failures during mixed-version rollout and then receives no series at all, which common threshold alerts treat as healthy. Per-message warning logs do not preserve the existing metric contract or automatically migrate that alert. Retain the metric until monitoring has been explicitly migrated, or add an equivalent failure counter and alert before removing the old series.

```

</details>

notification-worker/handler.go
```diff
@@ -433,8 +433,8 @@ func (h *Handler) resolveBody(ctx context.Context, content string, parsed mentio
 			}
 			resolved, err := h.deps.MentionNames.Resolve(ctx, lookup)
 			names = resolved
-			h.recordMentionResults(ctx, lookup, resolved, err)
 			if err != nil {
+				// Not counted in notificationMetrics: the warn line is the signal.
```
julianshen, line 437: _🏷️ observability_ | _🟠 Warning_ | _`silent-failure`_

**Mention lookup failures are hidden from service metrics**

When MentionNames.Resolve errors or times out, this path only writes a warning, continues with raw @tokens, and records the notification outcome as sent. Removing recordMentionResults and notification_worker_mention_resolution_total therefore makes a users-lookup outage invisible to metric-based failure monitoring while notifications appear fully successful. Users receive degraded message bodies without operators' existing alert signal. Preserve a failure counter for the unresolved remainder, or add an equivalent metric that is incremented here before continuing with the explicitly permitted fail-open delivery.

<details>

```
In `notification-worker/handler.go` around line 437: When MentionNames.Resolve errors or times out, this path only writes a warning, continues with raw @tokens, and records the notification outcome as sent. Removing recordMentionResults and notification_worker_mention_resolution_total therefore makes a users-lookup outage invisible to metric-based failure monitoring while notifications appear fully successful. Users receive degraded message bodies without operators' existing alert signal. Preserve a failure counter for the unresolved remainder, or add an equivalent metric that is incremented here before continuing with the explicitly permitted fail-open delivery.

```

</details>

julianshen (commented):
tGD inline review

notification-worker/nats_metrics.go
evelynchang5525: This removal is deliberate and maintainer-directed, not an oversight — so I'm not restoring the counter, but the alert-migration point deserves a straight answer.

**Provenance.** The metric was added in #297 and removed here at @mliu33's request in review of that PR: *"Let's remove this recordMentionResults as I think this is unnecessary. Let's only record metrics for important failures for now since too many redundant metrics will have some impact to the performance."* My first commit on this branch (`312a0a5`) did exactly what you're asking for — it kept a failure-only counter, `notification_worker_mention_lookup_failures_total`, and dropped just the resolved/unresolved accounting. That was then explicitly declined too ("also don't need to record mention failed results"), which is why `cf241c1` removes it entirely. Re-adding it would reverse a decision the maintainer has now made twice.

**On the alert-migration risk.** The concern is sound in general — a threshold alert on a series that stops being emitted reads as healthy, and that failure mode is worse than never having had the metric. It doesn't apply here on the timeline: `notification_worker_mention_resolution_total` landed in `main` with #297 on 2026-08-21 and is being removed on 2026-08-22. It existed for about a day and, as far as this repo shows, was never referenced by anything but the contract doc paragraph that this PR updates in the same commit. If an alert *was* already built on it, that's the one thing I can't see from here — @mliu33, if so, say the word and I'll re-add the failure counter.

**What remains visible.** A failed or timed-out lookup still emits `mention name lookup failed, body keeps raw mentions` at WARN via `slog.WarnContext`, so it carries traceId/spanId, the room/site/user baggage, the request ID, and both the requested and resolved counts. That is a log-based signal rather than a metric one, which is a real reduction in alertability — the contract doc now says so explicitly, so nobody goes looking for a metric that isn't there.

One correction to the finding as written: "preserve a failure counter for the unresolved remainder" conflates two things. The unresolved remainder is mostly people mentioning accounts that don't exist, which is a normal outcome, not a fault — counting it is what made the original metric noisy enough to be worth removing.

---

notification-worker/handler.go
evelynchang5525: Same finding as the one on `nats_metrics.go:49` — answered in full there: the metric removal was requested by @mliu33 in review of #297 and reaffirmed when I proposed keeping a failure-only counter, so it stands as a deliberate decision rather than an oversight. Failures remain visible through the trace-correlated warn line, and the contract doc says so, so nobody looks for a metric that isn't there.

One detail from this finding is worth correcting, and it turned out to apply to my own docs: *"the unresolved remainder"* is not a failure signal. `Resolve` returns the names it did collect alongside the error, and `resolveBody` still applies them, so a failed lookup can leave some, all, or none of the mentions raw. Counting the remainder as failures would have conflated an outage with people mentioning accounts that simply don't exist.

The doc paragraph I wrote in `cf241c1` overstated this in the same way — it said a failed lookup means the body ships raw `@tokens`. Fixed in `2ddf4bc`, which spells out the partial case and points at the requested-vs-resolved counts on the warn line as the way to tell how much was substituted. Credit to CodeRabbit, which caught the identical wording on the earlier revision of that paragraph.

---

julianshen:
## Needs clarification

Has every deployed alert using this documented metric been migrated or removed?
Reply in this thread, or post: `answer clar_pm255wc33aemk46fv7jvhyqa3k: <your answer>`

docs/notification-worker-downstream-contracts.md
```diff
@@ -81,7 +81,7 @@ Field notes:
...
 
-Operators: `notification_worker_mention_resolution_total{result="resolved|unresolved|failed"}` counts one point per looked-up token. `failed` is non-zero only when the users lookup itself errored or timed out, so alert on it; a rising `unresolved` ratio instead means senders are mentioning unknown accounts.
+Mention resolution emits no metric of its own. A lookup that errors or times out is reported by the `mention name lookup failed, body keeps raw mentions` warn line (trace-correlated, carries both the requested and the resolved count); the message itself still records as `notification_worker_outcomes_total{result="sent"}`, since the push is delivered either way. A failed lookup does **not** imply the whole body is unsubstituted: `Resolve` returns the names it did collect alongside the error and those are still applied, so an affected body may contain raw `@tokens` for some, all, or none of its mentions — compare the two counts on the warn line to tell which.
 
 ### Badge counts (`unreadCounts`)
```
julianshen, line 86: _📝 documentation_ | _🟠 Warning_ | _`tgd-review`_

**Lookup failure does not guarantee a sent outcome.**

The handler fails open after mention lookup errors, but delivery can still subsequently fail; `notifyPublishFailed` and `notifyFailed` remain valid outcomes. Saying the message always records as `sent` overstates the behavior and can mislead operators correlating the warning with outcome metrics. Document that lookup failure itself does not alter the eventual delivery outcome.

<details>

> ‼️ **IMPORTANT**

```suggestion
Mention resolution emits no metric of its own. A lookup that errors or times out is reported by the `mention name lookup failed, body keeps raw mentions` warn line (trace-correlated, carries both the requested and the resolved count); the lookup failure itself does not alter `notification_worker_outcomes_total`, so a successfully delivered push still records as `notification_worker_outcomes_total{result="sent"}` while a later delivery failure records its normal failure outcome. A failed lookup does **not** imply the whole body is unsubstituted: `Resolve` returns the names it did collect alongside the error and those are still applied, so an affected body may contain raw `@tokens` for some, all, or none of its mentions — compare the two counts on the warn line to tell which.
```

</details>

<details>

```
In `docs/notification-worker-downstream-contracts.md` around line 86: The handler fails open after mention lookup errors, but delivery can still subsequently fail; `notifyPublishFailed` and `notifyFailed` remain valid outcomes. Saying the message always records as `sent` overstates the behavior and can mislead operators correlating the warning with outcome metrics. Document that lookup failure itself does not alter the eventual delivery outcome.

```

</details>

julianshen (commented):
tGD inline review

mliu33 (approved):
Thanks!

PR #342, opened 2026-08-21
feat(admin): cross-site account sync with visible failures, per-user resync, cross-site console

nalydadad:
Admin-created and admin-edited accounts now replicate to every site, sync failures are visible and actionable in the console, and the console itself grows cross-site awareness.

## Demo

admin-user-sync-demo

## What's in here

**Two-lane account fanout (admin-service → all sites)**
- Create publishes the durable HR identity bootstrap (`chat.hr.{site}.users.upsert`, zstd — same wire mechanism as `teams-hr-sync`, now sharing `natsutil.NewMsgEncoded`) **plus** a `user_account_updated` snapshot to every remote INBOX. Update/deactivate re-send the whole snapshot (INBOX only).
- Failures never change the HTTP status: they surface as `syncFailures` (per site) / `hrSyncFailed` response fields. The fanout runs on its own budget, severed from client cancellation, capped by the request deadline.
- `inbox-worker` applies snapshots with an `accountUpdatedAt` watermark (`$lte` guard + `$setOnInsert` identity + one-shot non-upsert retry on E11000), so out-of-order arrival, redelivery, and the HR-lane race all converge — covered by an arrival-order permutation test matrix.

**Console (admin-frontend)**
- User list spans every site (Site column); rows homed elsewhere are read-only.
- Per-user **Resync** re-delivers the current state on both lanes — re-delivery, not re-recording: no write, no audit entry; double-confirm before firing.
- Sync notices follow lane coverage (spec R9): INBOX miss + HR ok → "only identity reached, roles/status did not sync"; both missed → full-failure notice; HR-only miss → silent (INBOX snapshot is complete). All notices direct to Resync.
- Permissions tab is deploy-gated by `PERMISSIONS_ENABLED` (**default off at every layer**; enable per environment).

## Deploy notes (read before rollout)

1. **Order**: deploy `inbox-worker` to **all** sites before `admin-service` anywhere — old inbox-workers drop `user_account_updated` events on the floor.
2. **Ops prerequisite**: every site running admin-service needs its `HR-{site}` stream provisioned, and each site's `hr-sync-worker` `SITE_IDS` must include every admin-service site (verified live: a missing stream is exactly the `hrSyncFailed` red path). Dev compose bootstraps this; production streams are IaC-owned.
3. **Authority widening (sign-off)**: replicated accounts make the `admin` role visible at every site, and `room-service`'s `GetUser` is account-only — a platform admin created here is a platform admin everywhere. Same replication also makes admin-created accounts permission-targetable from any site.
4. `PERMISSIONS_ENABLED` now defaults **off** in the dev compose too (was on).

## Verification

- TDD throughout; watermark logic mutation-verified (two independent mutations, each killed by a distinct test subset).
- Live two-site matrix (all four lane states + heal-then-resync): 15/15 checks green, HR-lane true delivery proven with a running `hr-sync-worker` (INBOX severed → replica still lands via the durable feed).
- Three /simplify review rounds (12 agents total) applied; `make lint`, unit + integration suites, and the frontend build are green.

## Known debts

- Coverage: `inbox-worker` 64.1% (structural ceiling ~77% — `main.go` wiring), `admin-service` unit 62.4%; both below the 80% floor pre-date this branch's files and are tracked separately.
- `tools/loadgen` `TestSoakPacing` flake pre-exists (fails at the merge-base too).

## New Features
- Admin user lists now include users from all sites, with remote users shown as read-only.
- Added user resynchronization to retry delivery to connected sites and HR systems.
- Create and edit flows display synchronization failures and affected sites.

## Improvements
- User updates now report detailed synchronization status.
- Improved narrow-screen usability with horizontal scrolling.
- Permissions navigation can be controlled through deployment configuration.

## Documentation
- Updated permissions setup and user API behavior documentation.


julianshen:
## Needs clarification

Can workers for different destination sites connect to the same JetStream domain?
Reply in this thread, or post: `answer clar_okqysvswrwjuulxu6hxkmceg5s: <your answer>`

julianshen:

### Needs clarification

Can workers for different destination sites connect to the same JetStream domain?

`answer clar_okqysvswrwjuulxu6hxkmceg5s: <your answer>`

[Open the question](https://github.com/hmchangw/newchat/pull/342#issuecomment-5369159610)

### Local review memory

- `docs/superpowers/plans/2026-08-20-admin-user-fanout.md:635` — @coderabbitai: _🩺 Stability & Availability_ _🟡 Minor_ _⚡ Quick win_ **Start the HR and INBOX lanes concurrently.** The pseudocode calls publishHRBootstrap before it starts the remote INBOX goroutines. A slow HR publish can consume FanoutTimeout … ([thread](https://github.com/hmchangw/newchat/pull/342#discussion_r3829533527))
- `admin-frontend/src/components/UsersConsole/UserTable/UserTable.jsx:40` — @coderabbitai: _🩺 Stability & Availability_ _🟡 Minor_ _⚡ Quick win_ **Guard user.roles before calling join .** If the API omits roles for a user, user.roles.join throws and the whole table fails to render. The cross-site listing now returns… ([thread](https://github.com/hmchangw/newchat/pull/342#discussion_r3829533478))
- `docs/superpowers/specs/2026-08-19-admin-user-fanout-design.md:57` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟡 Minor_ _⚡ Quick win_ **Make the HR delivery guarantee conditional on topology.** Q6 and the Approach section say that HR reaches every site by construction. The prerequisite states that a shared s… ([thread](https://github.com/hmchangw/newchat/pull/342#discussion_r3829533535))
- `admin-service/handler.go:320` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟠 Major_ _🏗️ Heavy lift_ &lt;details> &lt;summary>🧩 Analysis chain&lt;/summary> 🏁 Script executed: shell #!/bin/bash # Description: Look for an existing per-account write timestamp usable as the snapsh… ([thread](https://github.com/hmchangw/newchat/pull/342#discussion_r3829533498))
- `admin-service/store_mongo.go:118` — @coderabbitai: _🎯 Functional Correctness_ _🟡 Minor_ _⚡ Quick win_ **Add a deterministic sort to the paged search.** SearchUsers pages with SetSkip / SetLimit and no sort, so MongoDB returns no guaranteed order. Page boundaries can repeat or omi… ([thread](https://github.com/hmchangw/newchat/pull/342#discussion_r3829533506))
- `docs/client-api/request-reply.md:279` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟡 Minor_ _⚡ Quick win_ **Update the admin-service Emits statement.** The new rows document HR and INBOX publications for user creation, update, and resync. The following statement still says admin… ([thread](https://github.com/hmchangw/newchat/pull/342#discussion_r3829533515))
- `docs/superpowers/plans/2026-08-20-admin-user-fanout.md:24` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟡 Minor_ _⚡ Quick win_ **Remove the obsolete “docs/client-api.md is NOT touched” constraint.** This cohort changes docs/client-api.md and docs/client-api/request-reply.md . The plan still require… ([thread](https://github.com/hmchangw/newchat/pull/342#discussion_r3829533519))
- `docs/superpowers/specs/2026-08-19-admin-user-fanout-design.md:69` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟡 Minor_ _⚡ Quick win_ **Correct the documented status code for hrSyncFailed .** Q21 says hrSyncFailed is reported in a 200 response. User creation returns 201 ; 200 applies to update and re… ([thread](https://github.com/hmchangw/newchat/pull/342#discussion_r3829533542))
- `docs/superpowers/specs/2026-08-19-admin-user-fanout-design.md:81` — @coderabbitai: _🚀 Performance & Scalability_ _🟡 Minor_ _⚡ Quick win_ **Describe the actual publish count.** The Approach says “up to two JetStream publishes.” A create performs one HR publish plus one INBOX publish per remote site. Describe these a… ([thread](https://github.com/hmchangw/newchat/pull/342#discussion_r3829533551))
- `docs/superpowers/specs/2026-08-19-admin-user-fanout-design.md:121` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟡 Minor_ _⚡ Quick win_ **Include resync in the recovery paths.** These sections describe the next edit or admin change as the recovery mechanism. R8 adds POST /v1/admin/users/:account/resync , whic… ([thread](https://github.com/hmchangw/newchat/pull/342#discussion_r3829533558))

<details>
<summary>📒 Files reviewed (46)</summary>

* `admin-frontend/README.md`
* `admin-frontend/deploy/30-render-config.sh`
* `admin-frontend/deploy/config.js.template`
* `admin-frontend/deploy/docker-compose.yml`
* `admin-frontend/src/api/admin/admin.test.ts`
* `admin-frontend/src/api/admin/index.ts`
* `admin-frontend/src/api/index.ts`
* `admin-frontend/src/components/AppShell/AppShell.jsx`
* `admin-frontend/src/components/AppShell/AppShell.test.jsx`
* `admin-frontend/src/components/UsersConsole/CreateUserForm/CreateUserForm.jsx`
* `admin-frontend/src/components/UsersConsole/CreateUserForm/CreateUserForm.test.jsx`
* `admin-frontend/src/components/UsersConsole/EditUserDialog/EditUserDialog.jsx`
* `admin-frontend/src/components/UsersConsole/EditUserDialog/EditUserDialog.test.jsx`
* `admin-frontend/src/components/UsersConsole/ResyncUserDialog/ResyncUserDialog.jsx`
* `admin-frontend/src/components/UsersConsole/ResyncUserDialog/ResyncUserDialog.test.jsx`
* `admin-frontend/src/components/UsersConsole/ResyncUserDialog/index.jsx`
* `admin-frontend/src/components/UsersConsole/UserTable/UserTable.jsx`
* `admin-frontend/src/components/UsersConsole/UserTable/style.css`
* `admin-frontend/src/components/UsersConsole/UsersPage/UsersPage.jsx`
* `admin-frontend/src/components/UsersConsole/UsersPage/UsersPage.test.jsx`
* `admin-frontend/src/lib/runtimeConfig.js`
* `admin-service/handler.go`
* `admin-service/handler_test.go`
* `admin-service/main.go`
* `admin-service/routes.go`
* `admin-service/store.go`
* `docs/client-api.md`
* `docs/superpowers/plans/2026-08-20-admin-user-fanout.md`
* `docs/superpowers/specs/2026-08-19-admin-user-fanout-design.md`
* `hr-sync-worker/deploy/docker-compose.yml`
* `inbox-worker/handler.go`
* `inbox-worker/main.go`
* `pkg/model/event.go`
* `pkg/model/model_test.go`
* `pkg/natsutil/request_id.go`
* `pkg/natsutil/request_id_test.go`
* `teams-hr-sync/main.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

admin-service/handler.go
```diff
@@ -97,6 +104,121 @@ func toView(u *model.User) userView {
...
+	// them: serialized, a stalled local publish would spend the whole budget and every
+	// healthy peer would then be reported as failed. hrLaneFailed is written only in
+	// that goroutine and read after g.Wait().
+	var g errgroup.Group
+	var hrLaneFailed bool
+	if withHRBootstrap {
+		g.Go(func() error {
+			hrLaneFailed = h.publishHRBootstrap(ctx, u)
+			return nil
+		})
```
julianshen, line 156: _🏷️ nats-ordering_ | _🔴 Blocking_ | _`distributed-system`, `nats`, `performance-and-concurrency`, `silent-failure`, `tgd-review` (5 rules)_

**Publish-time watermarks can make stale snapshots win.**

The `user_account_updated` watermark is generated when fanout starts rather than atomically with the local mutation. A reachable interleaving is: PATCH A commits and pauses; PATCH B commits and publishes its newer state; then A resumes, receives a later `nowMillis()` value, and publishes its stale post-write document. The inbox worker accepts A because its timestamp is newer, regressing names, roles, or active state. Resync has the same issue when it reads before a concurrent PATCH but publishes afterward. Same-millisecond writes can also regress because `$lte` deliberately lets delivery order win. Persist a monotonic per-account revision with each create/update/deactivation, return it in the write projection, and publish that persisted revision for normal fanout and resync; receivers should compare that revision rather than publish time.

<details>

- **`distributed-system`**:163: The watermark is generated after the database mutation rather than atomically with it. Concrete interleaving: PATCH A commits and returns snapshot A, then pauses; PATCH B commits newer snapshot B and publishes it; A resumes, obtains a later `nowMillis()`, and publishes stale snapshot A, which every receiver accepts as newer. Separate replicas with clock skew can also assign a later write a lower timestamp, causing it to be dropped. Equal-millisecond writes remain unsafe because `$lte` lets whichever delivery arrives last win, including a redelivery of the older write. Persist a monotonically increasing account version atomically with each mutation, include that version in every snapshot/resync, and have receivers apply only strictly newer versions while treating equal versions as duplicate no-ops.
- **`silent-failure`**:163: `fanoutUserAccount` generates the watermark when publishing rather than when the represented state was committed. For example, resync can read an old user, a concurrent PATCH can commit and publish the new state, and then resync can publish its stale snapshot with a later timestamp. `inbox-worker` accepts that stale snapshot under its `$lte` watermark, so both HTTP requests report success while remote sites silently regress. The same race exists between concurrent updates or a delayed create fanout and an immediate update. Persist a monotonically increasing account revision atomically with every create/update/deactivation, return it with the projected user, and use that persisted revision for normal fanout and resync instead of `nowMillis()`.
- **`performance-and-concurrency`**:155: `nowMillis()` is sampled after the database operation, so it does not represent write order. For example, PATCH A can commit state V1 and pause before fanout; PATCH B then commits V2 and publishes it; A resumes, assigns a later timestamp, and publishes stale V1. `UpsertUserAccount` accepts V1 because its watermark is newer and regresses remote names, roles, or active state. A resync racing a PATCH has the same failure mode. Persist a unique, monotonically increasing account revision atomically with every mutation, return/read it with the user, and use that revision as the snapshot watermark instead of fanout-time wall-clock milliseconds.
- **`tgd-review`**:155: The watermark is generated only when a request reaches fanout, after its database write and audit. If request A commits and stalls while request B commits and fans out, A can receive the later timestamp and permanently overwrite B's newer state at receivers. Resync has the same read-then-publish race, and clock skew across service instances can also reject newer writes. Persist a monotonic per-account version or write timestamp atomically with create/update/deactivate, return it with the projected document, and reuse it for all snapshot and resync events. Add a concurrent-write/resync test that delays the earlier request's fanout.

</details>

<details>

```
In `admin-service/handler.go` around line 156: The `user_account_updated` watermark is generated when fanout starts rather than atomically with the local mutation. A reachable interleaving is: PATCH A commits and pauses; PATCH B commits and publishes its newer state; then A resumes, receives a later `nowMillis()` value, and publishes its stale post-write document. The inbox worker accepts A because its timestamp is newer, regressing names, roles, or active state. Resync has the same issue when it reads before a concurrent PATCH but publishes afterward. Same-millisecond writes can also regress because `$lte` deliberately lets delivery order win. Persist a monotonic per-account revision with each create/update/deactivation, return it in the write projection, and publish that persisted revision for normal fanout and resync; receivers should compare that revision rather than publish time.

```

</details>

admin-service/handler.go
```diff
@@ -257,7 +410,8 @@ func (h *Handler) createUser(c *gin.Context) {
 		"account": u.Account,
```
julianshen, line 410: _🏷️ atomicity_ | _🔴 Blocking_ | _`distributed-system`_

**A crash after commit can permanently skip all fanout**

The local user write commits before either JetStream publication is durably recorded. If the process crashes after a create commits but before `fanoutUserAccount` publishes or returns the 201, the caller observes a timeout and no HR or INBOX event exists. Retrying the create returns `account_exists`, so this account can remain absent remotely and continue blocking dependent federation work; the response failure fields are never available to direct the admin to resync. Record a fanout/outbox entry in the same Mongo transaction as the user mutation and dispatch it asynchronously with idempotent delivery, or persist an equivalent sync-pending marker that a recovery reconciler drains after crashes.

<details>

```
In `admin-service/handler.go` around line 410: The local user write commits before either JetStream publication is durably recorded. If the process crashes after a create commits but before `fanoutUserAccount` publishes or returns the 201, the caller observes a timeout and no HR or INBOX event exists. Retrying the create returns `account_exists`, so this account can remain absent remotely and continue blocking dependent federation work; the response failure fields are never available to direct the admin to resync. Record a fanout/outbox entry in the same Mongo transaction as the user mutation and dispatch it asynchronously with idempotent delivery, or persist an equivalent sync-pending marker that a recovery reconciler drains after crashes.

```

</details>

admin-service/handler.go
```diff
@@ -97,6 +104,121 @@ func toView(u *model.User) userView {
...
+
+// publishHRBootstrap parks the identity-only users.upsert on the local HR
+// stream (zstd, per the feed's current convention) so the create eventually
+// lands at every site even when a peer is unreachable right now.
+func (h *Handler) publishHRBootstrap(ctx context.Context, u *model.User) bool {
+	users := []model.IUserWithChange{{User: model.User{
+		ID: u.ID, Account: u.Account, SiteID: u.SiteID,
+		EngName: u.EngName, ChineseName: u.ChineseName,
+	}}}
+	data, err := json.Marshal(users)
```
julianshen, line 210: _🏷️ ordering_ | _🟠 Warning_ | _`distributed-system`, `performance-and-concurrency` (2 rules)_

**A delayed HR bootstrap can permanently revert newer identity fields**

The create-time HR event contains the original names and is consumed independently of the versioned INBOX lane. If the HR event is delayed, a later rename snapshot can apply first; when the old HR bootstrap eventually arrives, the existing HR upsert `$set`s the old identity fields without consulting `accountUpdatedAt`. The stale names then remain indefinitely unless another edit or resync occurs, and resync cannot prevent an older parked HR event from arriving afterward. Give the HR bootstrap the same persisted account version and enforce it in the HR consumer, or use bootstrap semantics that only populate identity fields when the account does not already exist.

<details>

- **`performance-and-concurrency`**:204: The durable HR bootstrap carries the create-time identity without a comparable account revision. If its consumer is delayed, a later rename can first arrive through the INBOX snapshot, followed by the older HR bootstrap overwriting those names. No subsequent event is guaranteed, so the regression persists until another edit or manual resync. Make this bootstrap insert-only, or carry and enforce the same persisted account revision on the HR apply path so older identity data cannot replace newer state.

</details>

<details>

```
In `admin-service/handler.go` around line 210: The create-time HR event contains the original names and is consumed independently of the versioned INBOX lane. If the HR event is delayed, a later rename snapshot can apply first; when the old HR bootstrap eventually arrives, the existing HR upsert `$set`s the old identity fields without consulting `accountUpdatedAt`. The stale names then remain indefinitely unless another edit or resync occurs, and resync cannot prevent an older parked HR event from arriving afterward. Give the HR bootstrap the same persisted account version and enforce it in the HR consumer, or use bootstrap semantics that only populate identity fields when the account does not already exist.

```

</details>

inbox-worker/main.go
```diff
@@ -272,6 +272,34 @@ func (s *mongoInboxStore) UpdateUserChatlist(ctx context.Context, account string
...
+	filter := bson.M{"account": e.Account, "$or": bson.A{
+		bson.M{"accountUpdatedAt": bson.M{"$exists": false}},
+		bson.M{"accountUpdatedAt": bson.M{"$lte": updatedAt}},
+	}}
+	update := bson.M{
+		"$setOnInsert": bson.M{"_id": e.ID, "siteId": e.SiteID},
+		"$set": bson.M{"engName": e.EngName, "chineseName": e.ChineseName,
+			"roles": roles, "active": e.Active, "accountUpdatedAt": updatedAt},
+	}
+	_, err := s.userCol.UpdateOne(ctx, filter, update, options.UpdateOne().SetUpsert(true))
```
julianshen, line 355: _🏷️ mongodb-indexes_ | _🟠 Warning_ | _`mongodb`_

**Account upserts proceed without their required unique index.**

The stale-event path intentionally relies on a unique account index: when the watermark filter rejects an existing account, upsert must raise E11000 so the non-upsert retry becomes a no-op. ensureIndexes only warns if account_1 is absent. If that index is missing or non-unique, an out-of-order event inserts a second document with the same account, after which point reads and updates can target arbitrary copies. Verify both the key and unique option before consuming events and fail startup/readiness when the invariant is absent, or provision the index as a mandatory migration before rollout.

<details>

```
In `inbox-worker/main.go` around line 291: The stale-event path intentionally relies on a unique account index: when the watermark filter rejects an existing account, upsert must raise E11000 so the non-upsert retry becomes a no-op. ensureIndexes only warns if account_1 is absent. If that index is missing or non-unique, an out-of-order event inserts a second document with the same account, after which point reads and updates can target arbitrary copies. Verify both the key and unique option before consuming events and fail startup/readiness when the invariant is absent, or provision the index as a mandatory migration before rollout.

```

</details>

julianshen (commented):
tGD inline review

admin-service/handler.go
```diff
@@ -97,6 +104,121 @@ func toView(u *model.User) userView {
...
+	// One goroutine per destination, results in a position-indexed slice: an
+	// unreachable peer burns only its own lane, and the reported order stays the
+	// configured destination order. Same shape as publishPermissionFanout.
+	dests := h.remoteDests
+	failed := make([]bool, len(dests))
+	for i, dest := range dests {
+		g.Go(func() error {
+			data, err := json.Marshal(model.InboxEvent{
+				Type: model.InboxUserAccountUpdated, SiteID: h.cfg.SiteID,
+				DestSiteID: dest, Payload: payload, Timestamp: now,
```
julianshen, line 184: _⚡ memory-exhaustion_ | _🟠 Warning_ | _`oom`_

**Account fanout creates unbounded in-flight goroutines and buffers**

Every create, update, or resync starts one goroutine per `ALL_SITE_IDS` destination. Each goroutine allocates a separate JSON envelope and can remain blocked in synchronous JetStream publication until the fanout deadline; `fanoutCtx` also deliberately ignores client cancellation. During a NATS outage, concurrent resync requests therefore retain roughly requests × destinations goroutines and message buffers for the full budget, with no site-count validation, service-wide concurrency limit, or backpressure. Admin automation retrying during an outage, or a compromised admin token, can cause GC thrashing and process OOM. Validate a practical maximum destination count and use a bounded service-wide publish pool/semaphore; reject or backpressure requests when its in-flight capacity is exhausted.

<details>

```
In `admin-service/handler.go` around line 184: Every create, update, or resync starts one goroutine per `ALL_SITE_IDS` destination. Each goroutine allocates a separate JSON envelope and can remain blocked in synchronous JetStream publication until the fanout deadline; `fanoutCtx` also deliberately ignores client cancellation. During a NATS outage, concurrent resync requests therefore retain roughly requests × destinations goroutines and message buffers for the full budget, with no site-count validation, service-wide concurrency limit, or backpressure. Admin automation retrying during an outage, or a compromised admin token, can cause GC thrashing and process OOM. Validate a practical maximum destination count and use a bounded service-wide publish pool/semaphore; reject or backpressure requests when its in-flight capacity is exhausted.

```

</details>

admin-frontend/src/components/UsersConsole/ResyncUserDialog/ResyncUserDialog.jsx
```diff
@@ -0,0 +1,77 @@
...
+            <>
+              Only this user&apos;s identity reached: {syncResult.syncFailures.join(', ')} —
+              roles and status did not sync. Resync again to deliver them.
+            </>
+          )}
+        </p>
+        <div className="dialog-actions">
+          <button type="button" onClick={onClose}>
+            Done
+          </button>
```
julianshen, line 53: _🏷️ silent-failure_ | _🟠 Warning_ | _`silent-failure`_

**Backdrop close hides an in-flight resync failure.**

After Resync is clicked, the buttons are disabled but the modal remains dismissible through Escape or the backdrop. The request continues after `onClose` unmounts the dialog; if it rejects or returns `syncFailures`, the subsequent error/result state update is invisible, so the administrator never sees that replication failed. Keep the modal non-dismissible while `submitting` is true.

<details>

```
In `admin-frontend/src/components/UsersConsole/ResyncUserDialog/ResyncUserDialog.jsx` around line 53: After Resync is clicked, the buttons are disabled but the modal remains dismissible through Escape or the backdrop. The request continues after `onClose` unmounts the dialog; if it rejects or returns `syncFailures`, the subsequent error/result state update is invisible, so the administrator never sees that replication failed. Keep the modal non-dismissible while `submitting` is true.

```

</details>

julianshen (commented):
tGD inline review

admin-service/handler.go
```diff
@@ -97,6 +104,121 @@ func toView(u *model.User) userView {
...
+	ctx, cancel := h.fanoutCtx(ctx)
+	defer cancel()
+
+	// The HR bootstrap shares the INBOX lanes' errgroup rather than running ahead of
+	// them: serialized, a stalled local publish would spend the whole budget and every
+	// healthy peer would then be reported as failed. hrLaneFailed is written only in
+	// that goroutine and read after g.Wait().
+	var g errgroup.Group
+	var hrLaneFailed bool
+	if withHRBootstrap {
```
mliu33, line 152: Why are we publishing to both hr stream as well as Inbox stream ?

We should just choose one approach. On user creation, we only need to publish to hr Jetstream. No need to do inbox fanout again.

c.c @nalydadad

admin-service/store_mongo.go
```diff
@@ -294,24 +294,33 @@ func (s *storeMongo) UpdateUserPasswordAndRevoke(ctx context.Context, siteID, ac
...
+// live token. Requires a replica set. Returns the post-write doc projected to
+// the fanout fields.
+func (s *storeMongo) DeactivateAndRevoke(ctx context.Context, siteID, account string) (*model.User, error) {
 	filter := bson.M{"account": account, "siteId": siteID}
 
-	return s.withTransaction(ctx, func(ctx context.Context) error {
-		result, err := s.users.UpdateOne(ctx, filter, bson.M{"$set": bson.M{"active": false}})
-		if err != nil {
+	var updated *model.User
+	err := s.withTransaction(ctx, func(ctx context.Context) error {
```
mliu33, line 303: Just curious, Why do we need to use transaction here ?

mliu33 (approved):
Excellent work! Just one comment, thanks!

mliu33:
Also need to rebase and resolve the conflict, thanks!

PR #345, opened 2026-08-22
docs(observability): app-health dashboard + metrics ops guide

chenjr0719:
## Summary

The observability suite already charts the `jetstream_*` exporter gauges, but the
application-side NATS metrics that services emit on `:9090` (the `chat_nats_*` series) are
scraped by Prometheus yet not shown on any dashboard. This adds a dashboard for them plus a
metrics reference/ops guide.

## What's included

- **`grafana/dashboards/nats-app-health.json`** — a new "NATS — Application Health & Terminal
  Failures" dashboard:
  - `chat_nats_consumer_loop_up` — per-consumer liveness (0 = loop down).
  - `chat_nats_terminal_failures_total{reason="max_deliver"}` — retry-exhausted drops. This is
    the one signal the `jetstream_consumer_num_redelivered` gauge cannot show: that gauge falls
    back to 0 whether messages were processed *or* dropped after MaxDeliver, so a drop looks
    like a recovery. This counter distinguishes them.
  - terminal-failure reasons, redelivery rate, and consumer disposition mix (ack/nak/term).
- **`METRICS.md`** — definition + "when high, do X" for every `jetstream_*` and
  `chat_nats_*` / `cache_*` / `room_key_*` / `search_service_*` series; the gauge-vs-counter
  distinction; and a scenario→metric mapping table (given an operational event, which metrics move).

## Verification

- Both dashboard JSON files validate (`json.load`) and follow the same schema (schemaVersion 39,
  datasource uid `prometheus`) as the existing dashboards; the provisioning provider globs the
  dashboards dir, so the new file auto-loads.
- All PromQL targets reference metric names verified present in `pkg/natsmetrics` (e.g.
  `chat_nats_terminal_failures_total`, reason `max_deliver`; `chat_nats_consumer_loop_up`).
- Docs-and-dashboard only — no service code changed.

## Summary by CodeRabbit

* **Documentation**
  * Added a NATS and JetStream metrics reference covering metric interpretation, operational scenarios, dashboards, exporter and application metrics, and observability startup.
  * Clarified redelivery causes and guidance for interpreting gauges, counters, and broker advisories.

* **New Features**
  * Added a Grafana dashboard for NATS application health.
  * Tracks consumer status, retry-exhausted messages, terminal failures, redelivery rates, consumer outcomes, and scrape-target health.
  * Includes five-second refreshes and a default 15-minute view.


tools/observability/grafana/dashboards/nats-app-health.json
```diff
@@ -0,0 +1,328 @@
...
+      "targets": [
+        {
+          "refId": "A",
+          "datasource": {
+            "type": "prometheus",
+            "uid": "prometheus"
+          },
+          "expr": "chat_nats_consumer_loop_up",
+          "legendFormat": "{{consumer}}"
+        }
```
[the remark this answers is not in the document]
chenjr0719: Addressed with a variant: clarified in both the panel description and METRICS.md that `chat_nats_consumer_loop_up=0` is a *graceful* loop failure, while a hard process crash removes the series entirely (it can't emit 0) — so a crash is caught with `absent(chat_nats_consumer_loop_up)` or the Prometheus scrape `up==0`, not the gauge alone. I did not reference `chat_jetstream_consumer_up` since that series is not emitted anywhere in this repo.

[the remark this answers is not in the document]
chenjr0719: Enriched all three legends to `{{consumer}} · {{stream}}/{{site}}` and aggregated the max_deliver + redelivery panels by `(consumer, stream, site)` so `event_type` no longer fragments the series. I left `service_name` out: on these `chat_nats_*` series the labels are `site`/`stream`/`consumer` (+ event_type/outcome/reason) — `service.name` is set as an OTel *resource* attribute (`OTEL_SERVICE_NAME`), which the Prometheus exporter surfaces via `target_info`, not as a per-series label, so `{{service_name}}` would render empty. The `consumer` label is the durable name, which already identifies the owning service.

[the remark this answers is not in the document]
chenjr0719: You are right — I verified `pkg/natsmetrics/prometheus_export_test.go`: the exporter uses `otelprom.WithResourceAsConstantLabels` with an allow-filter that includes `service.name`, and the test asserts every `chat_nats_*` family carries exactly one `service_name` label. My earlier reasoning assumed default otelprom behavior (resource attrs via `target_info`), which the SDK overrides. Restored `service_name` in all three legends and in the panel 2 + panel 4 `sum by` clauses, and documented it as a consumer label in METRICS.md (`38a62e67`). Thanks for the catch.

julianshen:
**No actionable comments.** ✅

### Disputed

- `tools/observability/grafana/dashboards/nats-app-health.json:174` (`distributed-system`) — On the final allowed delivery, the consumer can crash before incrementing this application counter or acknowledging. After AckWait, JetStream exhausts MaxDeliver and emits its advisory, but the application receives no further attempt; the counter remains flat and num_redelivered eventually clears while the message remains unprocessed. Documenting the limitation does not provide an operational signal. Durably capture `$JS.EVENT.ADVISORY.CONSUMER.MAX_DELIVERIES.>`, export persistent per-consumer failure state or counters, and alert on those broker-originated events.
- `tools/observability/grafana/dashboards/nats-app-health.json:368` (`distributed-system`) — If one expected durable never starts, it never creates `chat_nats_consumer_loop_up`; sibling consumers keep the metrics target reachable, so `up` remains 1 and the loop panel has no missing series to display. That durable's backlog can therefore grow without either health panel reporting failure. Define an authoritative expected-durable inventory and add label-scoped absence alerts for each `(service_name, site, stream, consumer)`, treating missing data as failure.
- `tools/observability/grafana/dashboards/nats-app-health.json:368` (`silent-failure`) — If one durable fails to start or its task exits while sibling consumers keep the process alive, its `chat_nats_consumer_loop_up` series disappears but this target-level `up` remains 1. Panel 1 also retains the durable's previous 1 via `lastNotNull`, so work can accumulate unprocessed while both displayed signals remain green. Documenting the limitation does not surface the failure, and no inventory-backed alert is added here. Add alert or recording rules based on an authoritative expected-durable inventory, using label-scoped absence checks and treating missing series as failure.

### Local review memory

- `tools/observability/METRICS.md:9` — @chenjr0719: Thanks — but this one does not hold for the exporter version this suite runs ( prometheus-nats-exporter 0.16.0 with -jsz=all ). I verified against the live Prometheus: the bare names **exist and return data**, and the gnatsd_ -prefixed… ([thread](https://github.com/hmchangw/newchat/pull/345#discussion_r3836120855))
- `tools/observability/grafana/dashboards/nats-app-health.json:368` — @chenjr0719: Agreed on the mechanism — up == 0 is target-level and won't catch a single durable whose series disappears while its process stays alive. I've made that explicit in the panel description and the guide (the chat_nats_consumer_loop_up ro… ([thread](https://github.com/hmchangw/newchat/pull/345#discussion_r3836423730))
- `tools/observability/grafana/dashboards/nats-app-health.json:174` — @chenjr0719: Agreed the broker MaxDeliver advisory is the authoritative signal, and the guide now says so. I have added the same caveat to this panel — it is labeled best-effort with the advisory named as the guarantee. I am deliberately not adding the… ([thread](https://github.com/hmchangw/newchat/pull/345#discussion_r3836758025))

<details>

* `tools/observability/grafana/dashboards/nats-app-health.json`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

tools/observability/grafana/dashboards/nats-app-health.json
```diff
@@ -0,0 +1,328 @@
...
+        "textMode": "value_and_name"
+      },
+      "targets": [
+        {
+          "refId": "A",
+          "datasource": {
+            "type": "prometheus",
+            "uid": "prometheus"
+          },
+          "expr": "chat_nats_consumer_loop_up",
```
julianshen, line 93: _🏷️ observability_ | _🟠 Warning_ | _`distributed-system`, `silent-failure`, `tgd-review` (3 rules)_

**The liveness panel reports crashed consumers as healthy**

If a consumer process crashes or becomes unreachable, Prometheus stops receiving `chat_nats_consumer_loop_up` rather than receiving 0. This query returns earlier samples, and the panel's `lastNotNull` reduction over the 15-minute range keeps displaying the previous green 1; it eventually becomes no-data, never DOWN, while the durable's backlog grows. The description acknowledges this failure mode but the dashboard does not implement the promised detection. Add a separate target/panel and alert based on scrape `up == 0`, plus absence checks against an inventory of expected consumer series; do not use `lastNotNull` as the liveness decision.

<details>

- **`silent-failure`**:111: When a consumer process crashes, `chat_nats_consumer_loop_up` disappears instead of becoming 0, as the panel description acknowledges. This range query is reduced with `lastNotNull`, so the last emitted value of 1 can remain green for the 15-minute dashboard window; afterward it becomes no-data rather than DOWN. Operators can therefore see apparent success while messages accumulate unprocessed. Add an actual scrape-failure/missing-series target or alert (for example using the relevant `up == 0` series), and configure missing data as a red failure state rather than merely documenting that operators should pair the signals.
- **`tgd-review`**:101: This query only reads the application gauge, while the panel reduces the range with `lastNotNull`. If the process crashes, the series disappears rather than emitting 0, so its previous value of 1 can remain green for the entire 15-minute dashboard window. The description acknowledges the limitation but the dashboard does not implement the suggested scrape-health check; add a panel/query based on the relevant Prometheus `up` targets or another inventory-backed absence check. Also correct the scenario in `METRICS.md` that currently says a process crash makes this gauge become 0.

</details>

<details>

```
In `tools/observability/grafana/dashboards/nats-app-health.json` around line 93: If a consumer process crashes or becomes unreachable, Prometheus stops receiving `chat_nats_consumer_loop_up` rather than receiving 0. This query returns earlier samples, and the panel's `lastNotNull` reduction over the 15-minute range keeps displaying the previous green 1; it eventually becomes no-data, never DOWN, while the durable's backlog grows. The description acknowledges this failure mode but the dashboard does not implement the promised detection. Add a separate target/panel and alert based on scrape `up == 0`, plus absence checks against an inventory of expected consumer series; do not use `lastNotNull` as the liveness decision.

```

</details>

tools/observability/METRICS.md
```diff
@@ -0,0 +1,107 @@
...
+| `jetstream_consumer_num_pending` | consumer | **undelivered backlog** (not yet handed to the consumer) | Sustained rise = consumer behind. Scale workers (`MAX_WORKERS`), check handler latency, confirm the loop is alive (`chat_nats_consumer_loop_up`). |
+| `jetstream_consumer_num_ack_pending` | consumer | **delivered-but-unacked** (in-flight); capped by `MaxAckPending` | Pinned at the cap = handlers wedged, or a `MaxAckPending=1` FIFO lane parked on a down dependency. Look at the handler, not the producer. |
+| `jetstream_consumer_num_redelivered` | consumer | messages **currently** in a redelivered state (delivery count > 1) | Any sustained non-zero = a failing/naking handler. Cross-check the `max_deliver` terminal counter — if that's also rising, messages are exhausting retries and being dropped. |
+| `jetstream_consumer_num_waiting` | consumer | parked pull requests | High with `num_pending>0` = pull starvation (workers idle while backlog sits) — usually a stuck/slow handler holding the ack-pending budget. |
+| `jetstream_consumer_delivered_stream_seq` | consumer | last delivered stream seq | Use with the next row. |
```
julianshen, line 59: _🏷️ observability_ | _🟠 Warning_ | _`distributed-system`_

**Stream sequence gaps are not consumer lag**

Subtracting stream sequences can greatly overstate lag for filtered consumers. For example, after one matching message is delivered, 100,000 records on unrelated subjects can separate it from the next matching delivery; while that one delivery is unacked, `delivered_stream_seq - ack_floor_stream_seq` reports a gap near 100,000 although only one consumer message is outstanding. Out-of-order acknowledgements can similarly hold the floor behind already-acked work. Use consumer sequence metrics for the sequence span and `num_ack_pending` for the actual in-flight count.

<details>

> ‼️ **IMPORTANT**

```suggestion
| `jetstream_consumer_delivered_consumer_seq` | consumer | last delivered consumer sequence | Use with the next row. |
| `jetstream_consumer_ack_floor_consumer_seq` | consumer | consumer sequence below which all acked | `delivered − ack_floor` is the outstanding consumer-sequence span; use `num_ack_pending` for the actual in-flight message count. |
```

</details>

<details>

```
In `tools/observability/METRICS.md` around line 48: Subtracting stream sequences can greatly overstate lag for filtered consumers. For example, after one matching message is delivered, 100,000 records on unrelated subjects can separate it from the next matching delivery; while that one delivery is unacked, `delivered_stream_seq - ack_floor_stream_seq` reports a gap near 100,000 although only one consumer message is outstanding. Out-of-order acknowledgements can similarly hold the floor behind already-acked work. Use consumer sequence metrics for the sequence span and `num_ack_pending` for the actual in-flight count.

```

</details>

julianshen, line 59: _🏷️ observability_ | _🟠 Warning_ | _`nats`_

**Ack-floor deltas overstate lag with out-of-order acks**

With concurrent processing, one early unacked message can hold the contiguous ack floor while many later messages are successfully acked. Consequently, `delivered - ack_floor` counts a sequence span rather than outstanding work, and `rate(ack_floor)` stalls and then spikes instead of representing actual ack throughput. Use `num_pending`, `num_ack_pending`, and the application ack counter for those measurements.

<details>

> ‼️ **IMPORTANT**

```suggestion
| `jetstream_consumer_ack_floor_stream_seq` | consumer | highest contiguous acknowledged stream seq | With concurrent or out-of-order processing, `delivered − ack_floor` is a sequence span that can include later messages already acked, and `rate(ack_floor)` advances in bursts. Use `num_pending` for undelivered backlog, `num_ack_pending` for in-flight work, and the application ack counter for ack throughput. |
```

</details>

<details>

```
In `tools/observability/METRICS.md` around line 49: With concurrent processing, one early unacked message can hold the contiguous ack floor while many later messages are successfully acked. Consequently, `delivered - ack_floor` counts a sequence span rather than outstanding work, and `rate(ack_floor)` stalls and then spikes instead of representing actual ack throughput. Use `num_pending`, `num_ack_pending`, and the application ack counter for those measurements.

```

</details>

tools/observability/METRICS.md
```diff
@@ -0,0 +1,107 @@
...
+
+## 4. Scenario → which metrics move
+
+Read the dashboard backwards: match the shape you see to the cause.
+
+| What happened | Metrics that move | How to read it |
+|---|---|---|
+| Consumer can't keep up with ingest | `num_pending` ↑ (sustained), `processing_duration` p99 ↑ | Backlog growing faster than drain. Scale workers / speed up the handler. |
+| A handler is wedged (stuck on a dependency) | `num_ack_pending` → pinned at `MaxAckPending`, `num_pending` ↑, ack rate (`rate(ack_floor)`) → 0 | Delivered work not acking. The producer is fine; the consumer isn't. |
+| A dependency has a brief blip, handler naks + retries | `num_redelivered` ↑ then → 0, `chat_nats_consumer_redeliveries_total` rate ↑, then recovers | Self-heal in progress — **but only if `chat_nats_terminal_failures_total{reason="max_deliver"}` stayed flat.** The gauge returns to 0 whether messages were acked *or* dropped after MaxDeliver, so confirm no `max_deliver` uptick before calling it cleared (next row). |
```
julianshen: _🏷️ jetstream_ | _🟠 Warning_ | _`nats`_

**MaxDeliver stops delivery but does not discard the message**

After a message reaches `MaxDeliver`, JetStream stops delivering it to that consumer and emits an advisory, but the unacknowledged message remains in the stream until retention or explicit removal applies. Calling it discarded or dropped can cause operators to omit recovery/replay and overlook stranded stream storage. Correct the repeated dropped/discarded wording here and in the dashboard, or explicitly document any application code that terminally acknowledges/removes these messages and routes them to recovery storage.

<details>

```
In `tools/observability/METRICS.md` around line 90: After a message reaches `MaxDeliver`, JetStream stops delivering it to that consumer and emits an advisory, but the unacknowledged message remains in the stream until retention or explicit removal applies. Calling it discarded or dropped can cause operators to omit recovery/replay and overlook stranded stream storage. Correct the repeated dropped/discarded wording here and in the dashboard, or explicitly document any application code that terminally acknowledges/removes these messages and routes them to recovery storage.

```

</details>

tools/observability/METRICS.md
```diff
@@ -0,0 +1,107 @@
...
+| A dependency outage lasts longer than the retry budget | `num_redelivered` ↑ then falls to 0 **while** `chat_nats_terminal_failures_total{reason="max_deliver"}` ↑ | **Not healed — dropped.** Messages exhausted MaxDeliver and were discarded (no DLQ). The most important row: the pending/redelivered gauges *look* recovered. |
```
julianshen: _🏷️ lifecycle_ | _🟠 Warning_ | _`nats`_

**A hard crash cannot set the loop gauge to zero**

A crashed process cannot emit the documented `loop_up = 0`; its last value becomes stale and the series eventually disappears. This row contradicts the earlier caveat and can produce a missed page while the durable backlog grows. Distinguish graceful loop termination from process death and monitor scrape health or the absence of each expected consumer identity.

<details>

> ‼️ **IMPORTANT**

```suggestion
| Consumer loop exited / process crashed | Graceful loop exit: `chat_nats_consumer_loop_up` → 0; hard process crash: the series goes stale/disappears; `num_pending` ↑ | Page on gauge 0, scrape `up == 0`, or a missing expected per-consumer series. Nothing is being processed for that durable. |
```

</details>

<details>

```
In `tools/observability/METRICS.md` around line 91: A crashed process cannot emit the documented `loop_up = 0`; its last value becomes stale and the series eventually disappears. This row contradicts the earlier caveat and can produce a missed page while the durable backlog grows. Distinguish graceful loop termination from process death and monitor scrape health or the absence of each expected consumer identity.

```

</details>

tools/observability/METRICS.md
```diff
@@ -0,0 +1,107 @@
+# NATS / JetStream Metrics — Reference & Ops Guide
+
+What each metric means, what to do when it goes high, and which metrics move for
+a given operational event. Companion to the dashboards in `grafana/dashboards/`.
+
+Two exposure paths feed Prometheus:
+
+- **JetStream state** — `prometheus-nats-exporter -jsz=all` reads the NATS monitoring
+  endpoint (`:8222`) and re-exposes `jetstream_*` series. No service code involved.
```
julianshen, line 9: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**JetStream metric names omit the exporter's gnatsd_ prefix**

The standard `prometheus-nats-exporter -jsz=all` exposition uses names such as `gnatsd_jetstream_stream_total_messages` and `gnatsd_jetstream_consumer_num_pending`. The guide consistently documents them as `jetstream_*`, so copied queries will return no data unless an undocumented metric-name rewrite exists. Update the introduction and every JetStream metric reference to the actual exposed names.

<details>

```
In `tools/observability/METRICS.md` around line 9: The standard `prometheus-nats-exporter -jsz=all` exposition uses names such as `gnatsd_jetstream_stream_total_messages` and `gnatsd_jetstream_consumer_num_pending`. The guide consistently documents them as `jetstream_*`, so copied queries will return no data unless an undocumented metric-name rewrite exists. Update the introduction and every JetStream metric reference to the actual exposed names.

```

</details>

julianshen (commented):
tGD inline review

tools/observability/METRICS.md
chenjr0719: Thanks — but this one does not hold for the exporter version this suite runs (`prometheus-nats-exporter` 0.16.0 with `-jsz=all`). I verified against the live Prometheus: the bare names **exist and return data**, and the `gnatsd_`-prefixed variants are **absent**:

- `jetstream_consumer_num_pending` → present · `gnatsd_jetstream_consumer_num_pending` → absent
- same for `num_ack_pending`, `num_redelivered`, `jetstream_stream_total_messages`

The only `gnatsd_`-prefixed family on this exporter is the **varz** endpoint (`gnatsd_varz_*`), a different metric set from the `-jsz` stream/consumer detail these dashboards use. The repo's existing JetStream dashboards already query the bare `jetstream_*` names and render correctly, so the guide matches what the exporter actually exposes here — leaving the names as-is.

The other findings in this pass were fair and are addressed in the latest commit: the `delivered − ack_floor` span-vs-lag correction, the MaxDeliver stops-delivery-not-discard wording, and a real `up == 0` scrape-health panel for the crashed-consumer case the loop gauge can't show.

tools/observability/METRICS.md
```diff
@@ -0,0 +1,109 @@
...
+  spikes while messages are looping and drops back to 0 the moment they clear. A gauge at 0
+  means "nothing stuck right now," **not** "nothing ever went wrong."
+- **Counter (cumulative, `_total` suffix).** Monotonic — only ever counts up, resetting to 0
+  only on process restart. Never read raw; wrap in `rate()` / `increase()[window]`. All
+  `chat_nats_*_total` series are counters — e.g. `increase(chat_nats_consumer_redeliveries_total[1h])`
+  answers "how many redeliveries happened in the last hour."
+
+A redelivery gauge falling to 0 does **not** prove the messages were processed — they may
+have been abandoned after MaxDeliver (the consumer stops delivering them; they stay in the
+stream, unprocessed). To tell healed from abandoned, pair the gauge with the
```
julianshen, line 36: _🏷️ failure-observability_ | _🟠 Warning_ | _`distributed-system`_

**An application counter cannot reliably detect broker-side MaxDeliver**

Concrete path: JetStream delivers a message for its final allowed attempt, then the consumer process crashes before recording `chat_nats_terminal_failures_total` or acknowledging the message. After `AckWait`, JetStream exhausts MaxDeliver and stops delivery, but the restarted application receives no further attempt and therefore cannot repair the missing counter increment. The documented gauge-plus-application-counter check can consequently report no terminal failure while work remains unprocessed. Capture the broker's MaxDeliver advisory in a durable stream and export/alert from that authoritative signal, or reconcile broker state after restart; keep the application counter only as a secondary diagnostic.

<details>

```
In `tools/observability/METRICS.md` around line 34: Concrete path: JetStream delivers a message for its final allowed attempt, then the consumer process crashes before recording `chat_nats_terminal_failures_total` or acknowledging the message. After `AckWait`, JetStream exhausts MaxDeliver and stops delivery, but the restarted application receives no further attempt and therefore cannot repair the missing counter increment. The documented gauge-plus-application-counter check can consequently report no terminal failure while work remains unprocessed. Capture the broker's MaxDeliver advisory in a durable stream and export/alert from that authoritative signal, or reconcile broker state after restart; keep the application counter only as a secondary diagnostic.

```

</details>

tools/observability/grafana/dashboards/nats-app-health.json
```diff
@@ -0,0 +1,395 @@
...
+        }
+      },
+      "targets": [
+        {
+          "refId": "A",
+          "datasource": {
+            "type": "prometheus",
+            "uid": "prometheus"
+          },
+          "expr": "up{job=~\"chat-app-metrics|nats-jetstream\"}",
```
julianshen, line 368: _🏷️ failure-observability_ | _🟠 Warning_ | _`distributed-system`, `tgd-review` (2 rules)_

**Target health misses a failed or never-started durable consumer**

If one expected durable fails during startup or its metric disappears while sibling consumers and the metrics endpoint remain alive, Prometheus still reports `up=1`. The loop-status panel has no inventory against which to recognize the missing series, so that durable can be absent while its backlog grows and both available health signals appear healthy. The guide itself calls for label-scoped `absent(chat_nats_consumer_loop_up{...})` checks, but this dashboard only adds target-level `up`. Add inventory-backed absence queries and alert rules for every expected `(service_name, site, stream, consumer)` identity, with missing data treated as failure.

<details>

- **`tgd-review`**:367: This `up` query detects only an entire scrape target becoming unreachable. If one consumer's `chat_nats_consumer_loop_up` series disappears while sibling consumers keep the process and target alive, `up` remains 1; meanwhile panel 1 retains the old green value via `lastNotNull` and eventually shows no data rather than DOWN. `METRICS.md` explicitly identifies label-scoped `absent(...)` checks per expected durable as necessary. Add inventory-backed absence queries or alerts for each expected durable and treat missing data as failure.

</details>

<details>

```
In `tools/observability/grafana/dashboards/nats-app-health.json` around line 368: If one expected durable fails during startup or its metric disappears while sibling consumers and the metrics endpoint remain alive, Prometheus still reports `up=1`. The loop-status panel has no inventory against which to recognize the missing series, so that durable can be absent while its backlog grows and both available health signals appear healthy. The guide itself calls for label-scoped `absent(chat_nats_consumer_loop_up{...})` checks, but this dashboard only adds target-level `up`. Add inventory-backed absence queries and alert rules for every expected `(service_name, site, stream, consumer)` identity, with missing data treated as failure.

```

</details>

julianshen (commented):
tGD inline review

tools/observability/grafana/dashboards/nats-app-health.json
chenjr0719: Agreed on the mechanism — `up == 0` is target-level and won't catch a single durable whose series disappears while its process stays alive. I've made that explicit in the panel description and the guide (the `chat_nats_consumer_loop_up` row already gives the exact label-scoped `absent(...{service_name,site,stream,consumer})` form).

On implementing it *in this dashboard*: I'm deliberately not baking an inventory into the panel. A per-durable absence check needs a **maintained list of every expected `(service_name, site, stream, consumer)`**, which goes stale the moment a consumer is added/renamed and is exactly the kind of brittle hardcoded inventory a dev dashboard shouldn't own. The correct home is the **alerting layer** — an Alertmanager rule (or a recording rule) over that inventory with missing-data treated as failure — which is a deployment/ops artifact, not a Grafana panel in this repo. So the guide documents the query and names the alerting layer as where it belongs, rather than shipping a list that's wrong by next week. Happy to add the alert rule if/when the expected-durable inventory is defined somewhere authoritative.

The other finding (the app `max_deliver` counter's crash-window gap) was a good catch and is fixed: the guide now marks the counter best-effort and names the broker MaxDeliver advisory as the authoritative signal.

tools/observability/METRICS.md
```diff
@@ -0,0 +1,116 @@
...
+
+## 4. Scenario → which metrics move
+
+Read the dashboard backwards: match the shape you see to the cause.
+
+| What happened | Metrics that move | How to read it |
+|---|---|---|
+| Consumer can't keep up with ingest | `num_pending` ↑ (sustained), `processing_duration` p99 ↑ | Backlog growing faster than drain. Scale workers / speed up the handler. |
+| A handler is wedged (stuck on a dependency) | `num_ack_pending` → pinned at `MaxAckPending`, `num_pending` ↑ | Delivered work not acking. The producer is fine; the consumer isn't. (Use `num_ack_pending`, not the `ack_floor` seq delta — the latter is a sequence span, not the in-flight count.) |
+| A dependency has a brief blip, handler naks + retries | `num_redelivered` ↑ then → 0, `chat_nats_consumer_redeliveries_total` rate ↑, then recovers | Self-heal in progress — **but only if `chat_nats_terminal_failures_total{reason="max_deliver"}` stayed flat.** The gauge returns to 0 whether messages were acked *or* abandoned after MaxDeliver, so confirm no `max_deliver` uptick before calling it cleared (next row). |
```
julianshen: _🏷️ jetstream-reliability_ | _🟠 Warning_ | _`nats`_

**A flat app counter does not prove MaxDeliver did not occur**

The recovery table still instructs operators to call the incident cleared when the best-effort application counter stays flat. On the final allowed delivery, the consumer can crash before recording the counter; after AckWait, JetStream exhausts MaxDeliver and stops delivering without any later application attempt repairing the metric. The dashboard likewise presents this counter as retry-exhausted work. Require captured `$JS.EVENT.ADVISORY.CONSUMER.MAX_DELIVERIES.>` events or broker-state reconciliation before declaring recovery, and label the application panel as secondary evidence.

<details>

> ‼️ **IMPORTANT**

```suggestion
| A dependency has a brief blip, handler naks + retries | `num_redelivered` ↑ then → 0, `chat_nats_consumer_redeliveries_total` rate ↑, then recovers | Possible self-heal in progress. The gauge returns to 0 whether messages were acked *or* abandoned after MaxDeliver. Because the application terminal-failure counter can miss a final-attempt process crash, do not call the incident cleared from a flat counter alone; confirm captured broker MaxDeliver advisories or broker-state reconciliation shows no exhausted messages. |
```

</details>

<details>

```
In `tools/observability/METRICS.md` around line 98: The recovery table still instructs operators to call the incident cleared when the best-effort application counter stays flat. On the final allowed delivery, the consumer can crash before recording the counter; after AckWait, JetStream exhausts MaxDeliver and stops delivering without any later application attempt repairing the metric. The dashboard likewise presents this counter as retry-exhausted work. Require captured `$JS.EVENT.ADVISORY.CONSUMER.MAX_DELIVERIES.>` events or broker-state reconciliation before declaring recovery, and label the application panel as secondary evidence.

```

</details>

tools/observability/METRICS.md
```diff
@@ -0,0 +1,116 @@
...
+---
+
+## 1. Gauge vs counter — read this first
+
+Two shapes of metric behave oppositely; confusing them is the most common misread.
+
+- **Gauge (point-in-time state).** Reflects *now*. Rises during an incident, **returns to 0
+  once resolved**. All `jetstream_*` series are gauges — e.g. `jetstream_consumer_num_redelivered`
+  spikes while messages are looping and drops back to 0 the moment they clear. A gauge at 0
+  means "nothing stuck right now," **not** "nothing ever went wrong."
```
julianshen: _🏷️ observability_ | _🟠 Warning_ | _`tgd-review`_

**Gauges do not necessarily return to zero**

Returning to zero after an incident is not a defining gauge behavior. Metrics listed below such as `jetstream_stream_last_seq` and the acknowledgement-floor sequence normally retain nonzero values, while retained message and byte gauges can have healthy nonzero baselines. Scope the zero-clearing explanation specifically to the redelivery gauge so operators do not apply it to every JetStream gauge.

<details>

> ‼️ **IMPORTANT**

```suggestion
- **Gauge (point-in-time state).** Reflects *now* and may rise or fall; unlike a counter,
  it is not necessarily monotonic or expected to return to 0. The `jetstream_*` state series
  below are gauges. For example, `jetstream_consumer_num_redelivered` rises while messages
  are looping and returns to 0 when they clear. For that metric, 0 means "nothing stuck right
  now," **not** "nothing ever went wrong."
```

</details>

<details>

```
In `tools/observability/METRICS.md` around line 23: Returning to zero after an incident is not a defining gauge behavior. Metrics listed below such as `jetstream_stream_last_seq` and the acknowledgement-floor sequence normally retain nonzero values, while retained message and byte gauges can have healthy nonzero baselines. Scope the zero-clearing explanation specifically to the redelivery gauge so operators do not apply it to every JetStream gauge.

```

</details>

julianshen (commented):
tGD inline review

tools/observability/grafana/dashboards/nats-app-health.json
```diff
@@ -0,0 +1,395 @@
...
+      "fieldConfig": {
+        "defaults": {
+          "unit": "short",
+          "custom": {
+            "drawStyle": "line",
+            "lineWidth": 1,
+            "fillOpacity": 10
+          }
+        },
+        "overrides": []
```
julianshen, line 174: _🏷️ failure-observability_ | _🟠 Warning_ | _`distributed-system`_

**The dashboard can miss stranded MaxDeliver messages**

Concrete path: JetStream performs the final allowed delivery, the consumer crashes before recording this application counter or acknowledging, and after AckWait the broker stops delivery. The message remains unprocessed in the stream, while the application counter never increments and num_redelivered eventually clears. Even recorded increments disappear from this short-window increase query and are lost on process restart. The new documentation accurately acknowledges this gap, but the dashboard still has no authoritative signal. Durably capture $JS.EVENT.ADVISORY.CONSUMER.MAX_DELIVERIES.> and export/query that state, or reconcile broker state after restart.

<details>

```
In `tools/observability/grafana/dashboards/nats-app-health.json` around line 174: Concrete path: JetStream performs the final allowed delivery, the consumer crashes before recording this application counter or acknowledging, and after AckWait the broker stops delivery. The message remains unprocessed in the stream, while the application counter never increments and num_redelivered eventually clears. Even recorded increments disappear from this short-window increase query and are lost on process restart. The new documentation accurately acknowledges this gap, but the dashboard still has no authoritative signal. Durably capture $JS.EVENT.ADVISORY.CONSUMER.MAX_DELIVERIES.> and export/query that state, or reconcile broker state after restart.

```

</details>

julianshen (commented):
tGD inline review

tools/observability/grafana/dashboards/nats-app-health.json
chenjr0719: Agreed the broker MaxDeliver advisory is the authoritative signal, and the guide now says so. I have added the same caveat to this panel — it is labeled best-effort with the advisory named as the guarantee.

I am deliberately not adding the advisory panel to this PR: charting $JS.EVENT.ADVISORY.CONSUMER.MAX_DELIVERIES.> needs a dedicated advisory-consumer plus a metrics exporter — the prometheus-nats-exporter (jsz) does not emit advisory events, so there is no series to query. That is an infrastructure pipeline (a small service + a stream), not a Grafana panel, and it belongs alongside the deployment rather than in this dev-observability dashboard. The guide documents the counter as secondary and names the advisory pipeline as the follow-up for a guarantee, which is the honest state of what this exporter can show. Happy to build the advisory-exporter as its own change if we want the authoritative signal on the board.

julianshen (commented):
tGD inline review

tools/observability/METRICS.md
```diff
@@ -0,0 +1,118 @@
...
+| `jetstream_consumer_num_ack_pending` | consumer | **delivered-but-unacked** (in-flight); capped by `MaxAckPending` | Pinned at the cap = handlers wedged, or a `MaxAckPending=1` FIFO lane parked on a down dependency. Look at the handler, not the producer. |
+| `jetstream_consumer_num_redelivered` | consumer | messages **currently** in a redelivered state (delivery count > 1) | Any sustained non-zero = repeated delivery attempts — a handler sending NAKs, **or** `AckWait` expiring on a slow/hung handler, **or** a consumer crash-looping on the same message. Cross-check the `max_deliver` terminal counter — if that's also rising, messages are exhausting retries and the consumer **stops delivering** them (they stay in the stream until retention; an advisory fires — recoverable, but nothing has processed them). |
+| `jetstream_consumer_num_waiting` | consumer | parked pull requests | High with `num_pending>0` = pull starvation (workers idle while backlog sits) — usually a stuck/slow handler holding the ack-pending budget. |
```
julianshen, line 58: _🏷️ nats-observability_ | _🟠 Warning_ | _`nats`_

**Ack-floor rate is not acknowledgement throughput**

With concurrent processing, one early unacked delivery holds the contiguous ack floor while later deliveries may be successfully acknowledged. The documented `rate()` therefore reports zero during healthy throughput and later spikes when the early message clears. Although the row now explains that the sequence delta is not an in-flight count, it still calls the floor rate “ack progress.” Use the application acknowledgement counter for throughput and describe this metric only as contiguous floor movement.

<details>

> ‼️ **IMPORTANT**

```suggestion
| `jetstream_consumer_ack_floor_stream_seq` | consumer | highest contiguous acked **stream** seq | With concurrent or out-of-order processing, `rate(ack_floor)` measures only contiguous floor movement: one early unacked delivery can make it stall and later spike, so it is not acknowledgement throughput. `delivered − ack_floor` is a **stream-sequence span, NOT the in-flight message count** — for a subject-filtered consumer it counts unrelated messages between matches. Use **`num_ack_pending`** for outstanding work, **`num_pending`** for undelivered backlog, and `rate(chat_nats_consumer_messages_total{outcome="ack"}[$__rate_interval])` for acknowledgement throughput. |
```

</details>

<details>

```
In `tools/observability/METRICS.md` around line 58: With concurrent processing, one early unacked delivery holds the contiguous ack floor while later deliveries may be successfully acknowledged. The documented `rate()` therefore reports zero during healthy throughput and later spikes when the early message clears. Although the row now explains that the sequence delta is not an in-flight count, it still calls the floor rate “ack progress.” Use the application acknowledgement counter for throughput and describe this metric only as contiguous floor movement.

```

</details>

tools/observability/grafana/dashboards/nats-app-health.json
```diff
@@ -0,0 +1,395 @@
...
+        "h": 8,
+        "w": 12,
+        "x": 12,
+        "y": 0
+      },
+      "fieldConfig": {
+        "defaults": {
+          "unit": "short",
+          "custom": {
+            "drawStyle": "line",
```
julianshen, line 117: _🏷️ NATS observability_ | _🟠 Warning_ | _`nats`_

**The panel incorrectly says increase() misses counter resets.**

For the MaxDeliver counter, Prometheus `increase()` compensates for an observed counter reset. A final-attempt increment is lost only if the application crashes before recording it or before Prometheus scrapes it. Claiming that a restart itself makes `increase()` miss the event can cause operators to discard valid pre-restart evidence. Distinguish unsampled increments from observed resets.

<details>

```
In `tools/observability/grafana/dashboards/nats-app-health.json` around line 117: For the MaxDeliver counter, Prometheus `increase()` compensates for an observed counter reset. A final-attempt increment is lost only if the application crashes before recording it or before Prometheus scrapes it. Claiming that a restart itself makes `increase()` miss the event can cause operators to discard valid pre-restart evidence. Distinguish unsampled increments from observed resets.

```

</details>

julianshen (commented):
tGD inline review

mliu33 (approved):

PR #347, opened 2026-08-22
feat: fleet-wide MongoDB pool + request-guard config helpers

hmchangw:
## Summary

Follow-up to #176 (which added the initial history-service pool guards, now merged). That work is generalized into shared, env-tagged config helpers and rolled out to **every long-running service that connects to MongoDB**, so each service has an explicit connection-pool cap, and the request/reply + HTTP services additionally bound in-flight concurrency and per-request duration.

Rebased onto current `main`, so it also reconciles with the read-preference feature main added in the meantime (both are preserved everywhere).

## Shared helpers (`pkg/`)

- **`mongoutil.PoolConfig` + `WithPool`** — `MONGO_MAX_POOL_SIZE` (default **150**) / `MONGO_MIN_POOL_SIZE`, applied via `mongoutil.WithPool(cfg.Pool)`; min is applied only when > 0 so a URI-provided `minPoolSize` survives.
- **`natsrouter.GuardConfig` + `DefaultGuarded`** — `MAX_CONCURRENCY` (default 256, 0 disables) / `REQUEST_TIMEOUT` (default 10s, 0 disables); `DefaultGuarded` wires the standard middleware chain + admission cap + timeout in one call so a service can't apply only half.
- **`ginutil.Timeout` + `TimeoutConfig`** — per-request context deadline for HTTP services.

## Rollout (by transport)

- **Request/reply (natsrouter)** → pool + concurrency cap + timeout: history-service, room-service, room-worker, search-service, bot-message-handler, bot-room-service, user-service, user-presence-service, media-service (NATS side).
- **HTTP (Gin)** → pool + request timeout: botplatform-service, portal-service, tcard-service. **admin-service, upload-service, and media-service's blob routes get the pool cap only** — upload/media stream large up/downloads that a short deadline would cancel; admin-service manages its own longer per-request budget (see below).
- **Workers** → explicit pool cap only (concurrency already bounded by `MAX_WORKERS`): message-worker, broadcast-worker, notification-worker, message-gatekeeper, inbox-worker, search-sync-worker, hr-sync-worker, bot-message-worker, and the teams-* sync services (applied to both read and write clients where split).

## Notable per-service decisions

- **Pool default 150** sits just above a mongod node's ~128 read-ticket ceiling. **history-service overrides to 384** in its deployment (read-heavy, `MAX_CONCURRENCY=256`): 384 is 1.5× the concurrency cap (in-flight handlers never starve at checkout, with burst headroom) and ≈ a 3-node replica set's aggregate read-ticket capacity (3×128). NOTE: the prod manifest must set `MONGO_MAX_POOL_SIZE=384` — the code default is 150.
- **admin-service** takes the pool cap but **no shared request timeout**: its permission handlers pin their own `requestBudget` (just under the 40s HTTP write timeout) and the cross-site permission fanout self-limits to `min(FANOUT_TIMEOUT=30s, request deadline)`. The fleet's 10s `REQUEST_TIMEOUT` default would silently cap that 30s fanout and abort multi-site permission changes early, so admin-service is deliberately timeout-free at the router (guarded by a regression test in `admin-service/router_test.go`).
- **bot-message-handler / bot-room-service** keep a bespoke `MAX_CONCURRENCY=200` default (their historical value), not the shared 256.
- **user-service** keeps its existing `HANDLER_TIMEOUT` as the single per-request deadline and takes only the concurrency cap.
- **user-presence-service** takes pool + timeout but **no admission cap** — its presence routes are fire-and-forget and would be silently dropped under a saturated semaphore.
- **search-service** keeps its existing (non-standard) middleware order rather than `DefaultGuarded`.

## Testing

- Shared helpers: TDD, race-tested, lint-clean.
- Whole repo builds (`go build ./...`); unit tests pass across all changed packages; `golangci-lint` clean.
- No client-facing handler or `pkg/model` struct changed, so `docs/client-api.md` is untouched.

### DIFF
--- a/history-service/internal/config/config.go
+++ b/history-service/internal/config/config.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hmchangw/chat/pkg/atrest"
 	"github.com/hmchangw/chat/pkg/logctx"
 	"github.com/hmchangw/chat/pkg/mongoutil"
+	"github.com/hmchangw/chat/pkg/natsrouter"
 )
 
 // CassandraConfig holds Cassandra connection settings (env prefix: CASSANDRA_).
@@ -21,7 +22,8 @@ type CassandraConfig struct {
 	NumConns int `env:"NUM_CONNS" envDefault:"8"`
 }
 
-// MongoConfig holds MongoDB connection settings (env prefix: MONGO_).
+// MongoConfig holds MongoDB connection settings (env prefix: MONGO_). Pool
+// sizing lives in the shared Config.Pool (mongoutil.PoolConfig).
 type MongoConfig struct {
 	URI      string `env:"URI"      required:"true"`
 	DB       string `env:"DB"       envDefault:"chat"`
@@ -30,16 +32,6 @@ type MongoConfig struct {
 	// ReadPreference is the client-level read preference; secondaryPreferred offloads
 	// history reads. DEK reads pin to primary in code regardless.
 	ReadPreference string `env:"READ_PREFERENCE" envDefault:"secondaryPreferred"`
-	// MaxPoolSize caps connections per server. This is authoritative — it is
-	// always applied to the client, overriding any maxPoolSize in the URI — so
-	// the pool ceiling is explicit rather than the driver default of 100.
-	// Connections are created lazily and are further gated by MAX_CONCURRENCY
-	// (a handler holds at most one at a time), so this is a generous headroom
-	// ceiling, not a count of connections that get opened. Operators must keep
-	// pods × MaxPoolSize under the cluster's connection limit.
-	MaxPoolSize uint64 `env:"MAX_POOL_SIZE" envDefault:"500"`
-	// MinPoolSize keeps a warm connection floor; 0 lets the pool drain to empty.
-	MinPoolSize uint64 `env:"MIN_POOL_SIZE" envDefault:"0"`
 }
 
 // NATSConfig holds NATS connection settings (env prefix: NATS_).
@@ -56,25 +48,22 @@ type Config struct {
 	MetricsAddr             string          `env:"METRICS_ADDR"               envDefault:":9090"`
 	Cassandra               CassandraConfig `envPrefix:"CASSANDRA_"`
 	Mongo                   MongoConfig     `envPrefix:"MONGO_"`
-	NATS                    NATSConfig      `envPrefix:"NATS_"`
-	MessageBucketHours      int             `env:"MESSAGE_BUCKET_HOURS"        envDefault:"360"`
-	MessageReadMaxBuckets   int             `env:"MESSAGE_READ_MAX_BUCKETS"    envDefault:"122"`
-	MessageHistoryFloorDays int             `env:"MESSAGE_HISTORY_FLOOR_DAYS"  envDefault:"730"`
-	LargeRoomThreshold      int             `env:"LARGE_ROOM_THRESHOLD"        envDefault:"500"`
-	MaxPinnedPerRoom        int             `env:"MAX_PINNED_PER_ROOM"         envDefault:"10"`
-	PinEnabled              bool            `env:"PIN_ENABLED"                 envDefault:"true"`
+	Pool                    mongoutil.PoolConfig
+	NATS                    NATSConfig `envPrefix:"NATS_"`
+	MessageBucketHours      int        `env:"MESSAGE_BUCKET_HOURS"        envDefault:"360"`
+	MessageReadMaxBuckets   int        `env:"MESSAGE_READ_MAX_BUCKETS"    envDefault:"122"`
+	MessageHistoryFloorDays int        `env:"MESSAGE_HISTORY_FLOOR_DAYS"  envDefault:"730"`
+	LargeRoomThreshold      int        `env:"LARGE_ROOM_THRESHOLD"        envDefault:"500"`
+	MaxPinnedPerRoom        int        `env:"MAX_PINNED_PER_ROOM"         envDefault:"10"`
+	PinEnabled              bool       `env:"PIN_ENABLED"                 envDefault:"true"`
 
 	// AdminAcctPrefix overrides the platform-admin account prefix (ADMIN_ACCT_PREFIX); keep it identical across services.
 	AdminAcctPrefix string `env:"ADMIN_ACCT_PREFIX" envDefault:"p_admin"`
 
-	// MaxConcurrency caps in-flight request handlers so a burst is shed at the
-	// door (ErrUnavailable) instead of piling unbounded work onto MongoDB. 0
-	// disables the cap (unbounded spawn — the previous behavior).
-	MaxConcurrency int `env:"MAX_CONCURRENCY" envDefault:"256"`
-	// RequestTimeout bounds each request handler so a slow MongoDB/Cassandra op
-	// is cancelled and its pooled connection returned instead of held. 0
-	// disables the per-request deadline.
-	RequestTimeout time.Duration `env:"REQUEST_TIMEOUT" envDefault:"10s"`
+	// Guard bounds in-flight request handlers (MAX_CONCURRENCY) and per-request
+	// duration (REQUEST_TIMEOUT) so a burst or a slow dependency can't saturate
+	// the Mongo pool with unbounded, indefinitely-held work.
+	Guard natsrouter.GuardConfig
 	// MaxResponseBytes caps a paginated reply so it is trimmed to fit rather
 	// than refused by the broker. 0 derives the cap from the broker's
 	// advertised max_payload.
@@ -146,18 +135,11 @@ func validate(cfg *Config) error {
 	if _, err := mongoutil.ParseReadPreference(cfg.Mongo.ReadPreference); err != nil {
 		return fmt.Errorf("MONGO_READ_PREFERENCE: %w", err)
 	}
-	// 0 makes the driver treat the pool as unbounded — reject it so the cap stays explicit.
-	if cfg.Mongo.MaxPoolSize < 1 {
-		return fmt.Errorf("MONGO_MAX_POOL_SIZE must be >= 1, got %d", cfg.Mongo.MaxPoolSize)
+	if err := cfg.Pool.Validate(); err != nil {
+		return err
 	}
-	if cfg.Mongo.MinPoolSize > cfg.Mongo.MaxPoolSize {
-		return fmt.Errorf("MONGO_MIN_POOL_SIZE (%d) must be <= MONGO_MAX_POOL_SIZE (%d)", cfg.Mongo.MinPoolSize, cfg.Mongo.MaxPoolSize)
-	}
-	if cfg.MaxConcurrency < 0 {
-		return fmt.Errorf("MAX_CONCURRENCY must be >= 0, got %d", cfg.MaxConcurrency)
-	}
-	if cfg.RequestTimeout < 0 {
-		return fmt.Errorf("REQUEST_TIMEOUT must be >= 0, got %s", cfg.RequestTimeout)
+	if err := cfg.Guard.Validate(); err != nil {
+		return err
 	}
 	return nil
 }
--- a/pkg/ginutil/timeout.go
+++ b/pkg/ginutil/timeout.go
@@ -1,0 +1,25 @@
+package ginutil
+
+import (
+	"context"
+	"time"
+
+	"github.com/gin-gonic/gin"
+)
+
+// Timeout bounds each request by deriving a context deadline on the request, so
+// a slow downstream op (e.g. a MongoDB query) is cancelled and its pooled
+// connection released rather than held until the pool starves. It complements
+// the http.Server read/write timeouts, which bound the socket but do not cancel
+// the handler's context. A non-positive duration is a no-op.
+func Timeout(d time.Duration) gin.HandlerFunc {
+	if d <= 0 {
+		return func(c *gin.Context) { c.Next() }
+	}
+	return func(c *gin.Context) {
+		ctx, cancel := context.WithTimeout(c.Request.Context(), d)
+		defer cancel()
+		c.Request = c.Request.WithContext(ctx)
+		c.Next()
+	}
+}
--- a/pkg/mongoutil/poolconfig.go
+++ b/pkg/mongoutil/poolconfig.go
@@ -1,0 +1,50 @@
+package mongoutil
+
+import "fmt"
+
+// PoolConfig is an env-tagged MongoDB connection-pool configuration. Add it as a
+// named field to a service's config (e.g. `Pool mongoutil.PoolConfig`), call
+// Validate() during config load, and pass WithPool(cfg.Pool) to Connect — so
+// every service's pool ceiling is explicit and tunable rather than the driver
+// default of 100. The env tags carry the full MONGO_ prefix so the field works
+// whether it sits in a flat config or alongside a MONGO_-prefixed block.
+//
+// The 150 default sits just above a mongod node's ~128 WiredTiger read-ticket
+// ceiling: enough headroom to keep tickets saturated without over-provisioning
+// connections fleet-wide. A high-throughput service whose in-flight handler cap
+// exceeds this (e.g. history-service, MAX_CONCURRENCY=256) should raise
+// MONGO_MAX_POOL_SIZE in its own deployment. Connections are created lazily and,
+// in a request/reply service, further gated by the handler cap, so this is a
+// ceiling, not a count of connections opened. Size pods × MaxPoolSize under the
+// cluster's connection limit — and note a service that opens separate read and
+// write clients applies this cap to each, so its ceiling is 2 × MaxPoolSize.
+type PoolConfig struct {
+	MaxPoolSize uint64 `env:"MONGO_MAX_POOL_SIZE" envDefault:"150"`
+	MinPoolSize uint64 `env:"MONGO_MIN_POOL_SIZE" envDefault:"0"`
+}
+
+// WithPool is a Connect option that applies this pool configuration, so call
+// sites read `Connect(ctx, uri, u, p, mongoutil.WithPool(cfg.Pool), ...)` rather
+// than hand-splicing pool options into the variadic. MaxPoolSize is always
+// applied (authoritative, validated >= 1); MinPoolSize is applied only when > 0,
+// so an unset floor leaves a minPoolSize from the URI (or the driver default)
+// intact rather than forcing it to 0.
+func WithPool(p PoolConfig) Option {
+	return func(c *connectConfig) {
+		c.maxPoolSize = &p.MaxPoolSize
+		if p.MinPoolSize > 0 {
+			c.minPoolSize = &p.MinPoolSize
+		}
+	}
+}
+
+// Validate rejects a zero max (the driver reads 0 as unbounded) and a min above max.
+func (p PoolConfig) Validate() error {
+	if p.MaxPoolSize < 1 {
+		return fmt.Errorf("MONGO_MAX_POOL_SIZE must be >= 1, got %d", p.MaxPoolSize)
+	}
+	if p.MinPoolSize > p.MaxPoolSize {
+		return fmt.Errorf("MONGO_MIN_POOL_SIZE (%d) must be <= MONGO_MAX_POOL_SIZE (%d)", p.MinPoolSize, p.MaxPoolSize)
+	}
+	return nil
+}
--- a/pkg/natsrouter/guard.go
+++ b/pkg/natsrouter/guard.go
@@ -1,0 +1,67 @@
+package natsrouter
+
+import (
+	"fmt"
+	"time"
+
+	o11ynats "github.com/flywindy/o11y/nats"
+)
+
+// GuardConfig is an env-tagged bundle of the two request/reply overload guards:
+// an in-flight handler cap and a per-request timeout. Add it as a named field to
+// a service's config (e.g. `Guard natsrouter.GuardConfig`), call Validate()
+// during config load, pass Options() to New, and Use() the TimeoutMiddleware()
+// after the standard middleware.
+//
+// Together they keep a burst or a slow dependency from spawning unbounded
+// handlers and holding pooled connections: excess requests are shed with
+// ErrUnavailable, and a stalled op is cancelled so its connection is returned.
+type GuardConfig struct {
+	// MaxConcurrency caps concurrent in-flight handlers; 0 disables the cap.
+	MaxConcurrency int `env:"MAX_CONCURRENCY" envDefault:"256"`
+	// RequestTimeout bounds each handler so a slow dependency op is cancelled
+	// and its connection released; 0 disables the deadline.
+	RequestTimeout time.Duration `env:"REQUEST_TIMEOUT" envDefault:"10s"`
+}
+
+// Validate rejects negative values; zero is the documented disable value for both.
+func (g GuardConfig) Validate() error {
+	if g.MaxConcurrency < 0 {
+		return fmt.Errorf("MAX_CONCURRENCY must be >= 0, got %d", g.MaxConcurrency)
+	}
+	if g.RequestTimeout < 0 {
+		return fmt.Errorf("REQUEST_TIMEOUT must be >= 0, got %s", g.RequestTimeout)
+	}
+	return nil
+}
+
+// Options returns the router construction options for this guard (the admission
+// cap when MaxConcurrency > 0, none otherwise).
+func (g GuardConfig) Options() []Option {
+	var opts []Option
+	if g.MaxConcurrency > 0 {
+		opts = append(opts, WithMaxConcurrency(g.MaxConcurrency))
+	}
+	return opts
+}
+
+// TimeoutMiddleware returns the per-request timeout middleware (empty when
+// RequestTimeout is 0), suitable for spreading into Use().
+func (g GuardConfig) TimeoutMiddleware() []HandlerFunc {
+	if g.RequestTimeout > 0 {
+		return []HandlerFunc{HandlerTimeout(g.RequestTimeout)}
+	}
+	return nil
+}
+
+// DefaultGuarded builds a Router with the standard middleware chain (Default:
+// Recovery, RequestID, Logging) plus this guard's admission cap and per-request
+// timeout, wired in the correct order. It exists so a service can't apply only
+// half of the overload protection — the cap (a construction option) and the
+// timeout (post-middleware) are otherwise two separate calls that are easy to
+// forget one of. Services needing a non-standard middleware order keep New.
+func DefaultGuarded(nc *o11ynats.Conn, queue string, g GuardConfig) *Router {
+	r := Default(nc, queue, g.Options()...)
+	r.Use(g.TimeoutMiddleware()...)
+	return r
+}
--- a/room-service/main.go
+++ b/room-service/main.go
@@ -43,11 +43,19 @@ type config struct {
 	MongoPassword     string            `env:"MONGO_PASSWORD"            envDefault:""`
 	// MongoReadPreference routes the store's display/list reads to secondaries; the
 	// client stays on primary for authz/dedup/read-after-write.
-	MongoReadPreference string        `env:"MONGO_READ_PREFERENCE" envDefault:"secondaryPreferred"`
-	MaxRoomSize         int           `env:"MAX_ROOM_SIZE"             envDefault:"1000"`
-	MaxBatchSize        int           `env:"MAX_BATCH_SIZE"            envDefault:"1000"`
-	MemberListTimeout   time.Duration `env:"MEMBER_LIST_TIMEOUT"       envDefault:"5s"`
-	RoomKeyGracePeriod  time.Duration `env:"ROOM_KEY_GRACE_PERIOD"     envDefault:"24h"`
+	MongoReadPreference string `env:"MONGO_READ_PREFERENCE" envDefault:"secondaryPreferred"`
+	// Pool caps the Mongo connection pool (MONGO_MAX_POOL_SIZE/MONGO_MIN_POOL_SIZE)
+	// so a burst can't open unbounded connections. Env tags already carry the
+	// MONGO_ prefix, so this stays a top-level field (never under envPrefix:"MONGO_").
+	Pool mongoutil.PoolConfig
+	// Guard bounds in-flight request handlers (MAX_CONCURRENCY) and per-request
+	// duration (REQUEST_TIMEOUT) so a burst or slow dependency can't saturate the
+	// Mongo pool with unbounded, indefinitely-held work.
+	Guard              natsrouter.GuardConfig
+	MaxRoomSize        int           `env:"MAX_ROOM_SIZE"             envDefault:"1000"`
+	MaxBatchSize       int           `env:"MAX_BATCH_SIZE"            envDefault:"1000"`
+	MemberListTimeout  time.Duration `env:"MEMBER_LIST_TIMEOUT"       envDefault:"5s"`
+	RoomKeyGracePeriod time.Duration `env:"ROOM_KEY_GRACE_PERIOD"     envDefault:"24h"`
 	// RoomKeyRetiredTTL: retention for rotated-out keys; see roomkeystore.WithRetiredKeys for the 2x-cache-TTL rule.
 	RoomKeyRetiredTTL        time.Duration   `env:"ROOM_KEY_RETIRED_TTL"      envDefault:"20m"`
 	HealthAddr               string          `env:"HEALTH_ADDR" envDefault:":8081"`
@@ -94,10 +102,6 @@ type config struct {
 	BadgeCacheTTL time.Duration `env:"BADGE_CACHE_TTL" envDefault:"24h"`
 	// RoomLocalityGrace: post-flip dual-publish window. Must match across all publisher services.
 	RoomLocalityGrace time.Duration `env:"ROOM_LOCALITY_GRACE" envDefault:"168h"`
-	// MaxConcurrency caps in-flight request handlers so a burst is shed at the
-	// door (ErrUnavailable) instead of piling unbounded work onto MongoDB. 0
-	// disables the cap (unbounded spawn).
-	MaxConcurrency int `env:"MAX_CONCURRENCY" envDefault:"256"`
 }
 
 // legacyRoomOrigin maps a site to its legacy origin URL (incl. scheme).
@@ -138,6 +142,14 @@ func main() {
 		os.Exit(1)
 	}
 	logctx.Configure(cfg.DebugLog)
+	if err := cfg.Pool.Validate(); err != nil {
+		slog.Error("invalid mongo pool config", "error", err)
+		os.Exit(1)
+	}
+	if err := cfg.Guard.Validate(); err != nil {
+		slog.Error("invalid guard config", "error", err)
+		os.Exit(1)
+	}
 	if err := model.SetPlatformAdminAccountPrefix(cfg.AdminAcctPrefix); err != nil {
 		slog.Error("invalid ADMIN_ACCT_PREFIX", "error", err)
 		os.Exit(1)
@@ -178,7 +190,8 @@ func main() {
 		os.Exit(1)
 	}
 
-	mongoClient, err := mongoutil.Connect(ctx, cfg.MongoURI, cfg.MongoUsername, cfg.MongoPassword, mongoutil.WithObservability(sdk))
+	mongoClient, err := mongoutil.Connect(ctx, cfg.MongoURI, cfg.MongoUsername, cfg.MongoPassword,
+		mongoutil.WithPool(cfg.Pool), mongoutil.WithObservability(sdk))
 	if err != nil {
 		slog.Error("mongo connect failed", "error", err)
 		os.Exit(1)
@@ -328,14 +341,7 @@ func main() {
 	handler.roomMembersLimit = cfg.RoomMembersLimit
 	handler.roomMembersCallLimit = cfg.RoomMembersCallLimit
 
-	// Bound in-flight handlers so a burst is shed at the door (ErrUnavailable)
-	// instead of piling unbounded work onto MongoDB. MAX_CONCURRENCY=0 disables.
-	routerOpts := []natsrouter.Option{natsrouter.WithSiteID(cfg.SiteID), natsrouter.WithMetrics(publishMetrics)}
-	if cfg.MaxConcurrency > 0 {
-		routerOpts = append(routerOpts, natsrouter.WithMaxConcurrency(cfg.MaxConcurrency))
-	}
-	router := natsrouter.New(nc, "room-service", routerOpts...)
-	router.Use(natsrouter.Recovery(), natsrouter.RequestID(), natsrouter.Logging())
+	router := natsrouter.DefaultGuarded(nc, "room-service", cfg.Guard)
 	handler.Register(router)
 
 	healthStop, err := health.ServeWithPprof(cfg.HealthAddr, 5*time.Second, cfg.PProfEnabled,
--- a/search-service/main.go
+++ b/search-service/main.go
@@ -97,10 +97,6 @@ type Config struct {
 	UserRoomIndex     string `env:"USER_ROOM_INDEX,required,notEmpty"`
 	SpotlightIndex    string `env:"SPOTLIGHT_INDEX,required,notEmpty"`
 	SpotlightOrgIndex string `env:"SPOTLIGHT_ORG_INDEX,required,notEmpty"`
-	// MaxConcurrency caps in-flight request handlers so a burst is shed at the
-	// door (ErrUnavailable) instead of piling unbounded work onto Elasticsearch/
-	// MongoDB. 0 disables the cap (unbounded spawn).
-	MaxConcurrency int `env:"MAX_CONCURRENCY" envDefault:"256"`
 	// ShowTeamsRoom controls whether Teams-migrated rooms/messages (origin
 	// "teams") appear in search results; false hides them (reversible read-time
 	// filter — see pkg/model.OriginTeams).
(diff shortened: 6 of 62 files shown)


julianshen:

### Related work

- [PR #176](https://github.com/hmchangw/newchat/pull/176) — feat: guard MongoDB connection pools across all services (merged)

### Local review memory

- `admin-service/config_test.go:87` — @coderabbitai: _🎯 Functional Correctness_ _🟡 Minor_ _⚡ Quick win_ **Assert the pool validation error.** assert.Error passes when any configuration error occurs. Assert that the error mentions MONGO_MAX_POOL_SIZE ; otherwise this test can pass wi… ([thread](https://github.com/hmchangw/newchat/pull/347#discussion_r3841188789))
- `pkg/ginutil/timeout_test.go:14` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟡 Minor_ _⚡ Quick win_ &lt;details> &lt;summary>🧩 Analysis chain&lt;/summary> 🌐 Web query: For github.com/gin-gonic/gin v1.12.0, does gin.Mode() return the current mode string, and can that string be pas… ([thread](https://github.com/hmchangw/newchat/pull/347#discussion_r3841188815))

<details>
<summary>📒 Files reviewed (62)</summary>

* `admin-service/config.go`
* `admin-service/config_test.go`
* `admin-service/main.go`
* `admin-service/router_test.go`
* `bot-room-service/main.go`
* `botplatform-service/config.go`
* `broadcast-worker/main.go`
* `hr-sync-worker/config.go`
* `hr-sync-worker/main.go`
* `inbox-worker/consumer_config_test.go`
* `inbox-worker/main.go`
* `media-service/config.go`
* `media-service/config_test.go`
* `media-service/main.go`
* `message-worker/main.go`
* `pkg/ginutil/timeout.go`
* `pkg/ginutil/timeout_test.go`
* `pkg/ginutil/timeoutconfig.go`
* `pkg/ginutil/timeoutconfig_test.go`
* `pkg/mongoutil/poolconfig.go`
* `pkg/mongoutil/poolconfig_test.go`
* `pkg/natsrouter/guard.go`
* `pkg/natsrouter/guard_test.go`
* `portal-service/main.go`
* `room-service/main.go`
* `room-worker/main.go`
* `search-service/main.go`
* `search-sync-worker/config_test.go`
* `tcard-service/main.go`
* `teams-chat-member-sync/main.go`
* `teams-chat-member-sync/main_test.go`
* `teams-chat-sync/main.go`
* `teams-chat-sync/main_test.go`
* `teams-hr-sync/config.go`
* `teams-hr-sync/config_test.go`
* `teams-hr-sync/main.go`
* `teams-room-creation/config.go`
* `teams-room-creation/config_test.go`
* `teams-room-creation/main.go`
* `teams-user-sync/config.go`
* `teams-user-sync/config_test.go`
* `teams-user-sync/main.go`
* `upload-service/main.go`
* `user-service/main.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

pkg/mongoutil/poolconfig.go
```diff
@@ -0,0 +1,50 @@
...
+// WithPool is a Connect option that applies this pool configuration, so call
+// sites read `Connect(ctx, uri, u, p, mongoutil.WithPool(cfg.Pool), ...)` rather
+// than hand-splicing pool options into the variadic. MaxPoolSize is always
+// applied (authoritative, validated >= 1); MinPoolSize is applied only when > 0,
+// so an unset floor leaves a minPoolSize from the URI (or the driver default)
+// intact rather than forcing it to 0.
+func WithPool(p PoolConfig) Option {
+	return func(c *connectConfig) {
+		c.maxPoolSize = &p.MaxPoolSize
+		if p.MinPoolSize > 0 {
```
julianshen, line 35: _🏷️ rolling-deploy_ | _🔴 Blocking_ | _`distributed-system`_

**A URI minPoolSize can conflict with the forced maximum.**

WithPool always overrides maxPoolSize but preserves minPoolSize from the URI when MONGO_MIN_POOL_SIZE is zero. An existing valid URI such as `?minPoolSize=200&maxPoolSize=500` therefore becomes min=200/max=150 after this upgrade, despite PoolConfig.Validate seeing only 0/150. Mongo client validation then rejects the configuration, so upgraded replicas fail to connect during rollout. Either make MinPoolSize authoritative too, including zero, or validate the effective URI-derived minimum against the configured maximum before connecting.

<details>

```
In `pkg/mongoutil/poolconfig.go` around line 35: WithPool always overrides maxPoolSize but preserves minPoolSize from the URI when MONGO_MIN_POOL_SIZE is zero. An existing valid URI such as `?minPoolSize=200&maxPoolSize=500` therefore becomes min=200/max=150 after this upgrade, despite PoolConfig.Validate seeing only 0/150. Mongo client validation then rejects the configuration, so upgraded replicas fail to connect during rollout. Either make MinPoolSize authoritative too, including zero, or validate the effective URI-derived minimum against the configured maximum before connecting.

```

</details>

julianshen (commented):
tGD inline review

user-presence-service/main.go
```diff
@@ -128,7 +153,15 @@ func main() {
...
 	handler := NewHandler(store, userDir, peer, publish, cfg.SiteID, cfg.Presence.BatchMax)
 
+	// No admission cap here: Hello/Ping/Activity/Bye are fire-and-forget
+	// (RegisterVoid), and under a saturated concurrency semaphore those are
+	// silently dropped (no reply, no redelivery) — a reconnect storm would lose
+	// presence updates and strand users online/offline. Only the per-request
+	// timeout is applied.
 	router := natsrouter.Default(nc, "user-presence-service", natsrouter.WithSiteID(cfg.SiteID))
+	if cfg.RequestTimeout > 0 {
+		router.Use(natsrouter.HandlerTimeout(cfg.RequestTimeout))
```
julianshen, line 163: _🏷️ delivery-guarantees_ | _🟠 Warning_ | _`distributed-system`_

**The timeout silently loses fire-and-forget presence updates.**

The surrounding comment establishes that these RegisterVoid messages have no reply or redelivery. If Mongo or Valkey takes longer than REQUEST_TIMEOUT, HandlerTimeout cancels a Hello, Activity, or Bye operation; the producer has already completed its fire-and-forget publish, so the failed update is never retried. A lost Hello/Bye can leave presence state incorrect until another event or expiry repairs it. Do not apply a canceling timeout to these routes unless delivery is moved to an acknowledged/durable mechanism with idempotent retries; otherwise remove this timeout from the void handlers.

<details>

```
In `user-presence-service/main.go` around line 163: The surrounding comment establishes that these RegisterVoid messages have no reply or redelivery. If Mongo or Valkey takes longer than REQUEST_TIMEOUT, HandlerTimeout cancels a Hello, Activity, or Bye operation; the producer has already completed its fire-and-forget publish, so the failed update is never retried. A lost Hello/Bye can leave presence state incorrect until another event or expiry repairs it. Do not apply a canceling timeout to these routes unless delivery is moved to an acknowledged/durable mechanism with idempotent retries; otherwise remove this timeout from the void handlers.

```

</details>

room-service/main.go
```diff
@@ -327,14 +344,7 @@ func main() {
...
 
-	// Bound in-flight handlers so a burst is shed at the door (ErrUnavailable)
-	// instead of piling unbounded work onto MongoDB. MAX_CONCURRENCY=0 disables.
-	routerOpts := []natsrouter.Option{natsrouter.WithSiteID(cfg.SiteID), natsrouter.WithMetrics(publishMetrics)}
-	if cfg.MaxConcurrency > 0 {
-		routerOpts = append(routerOpts, natsrouter.WithMaxConcurrency(cfg.MaxConcurrency))
-	}
-	router := natsrouter.New(nc, "room-service", routerOpts...)
-	router.Use(natsrouter.Recovery(), natsrouter.RequestID(), natsrouter.Logging())
+	router := natsrouter.DefaultGuarded(nc, "room-service", cfg.Guard)
```
julianshen, line 344: _🏷️ distributed-observability_ | _🟠 Warning_ | _`distributed-system`, `nats` (2 rules)_

**Guard migrations discard existing router options.**

This replacement drops both `WithSiteID(cfg.SiteID)` and `WithMetrics(publishMetrics)` because DefaultGuarded forwards only GuardConfig.Options. The same regression drops site identity in bot-message-handler, bot-room-service, media-service, room-worker, and search-service, while room-worker also loses its metrics sink. During a mixed-version rollout, requests handled by new replicas therefore lack the previous site attribution, and overload rejections/timeouts in room-service and room-worker can no longer be counted through the configured router metrics. Preserve the existing options when adding guard options, as history-service does, or extend DefaultGuarded to accept additional Option values and update every migrated caller.

<details>

- **`nats`**:347: The prior routers passed WithSiteID(cfg.SiteID), but the new DefaultGuarded helper accepts only GuardConfig and forwards only g.Options(). This drops site identity in bot-message-handler, bot-room-service, media-service, room-service, and room-worker; search-service similarly replaces WithSiteID with cfg.Guard.Options(). Every request handled by these routers now uses the router's empty/default site identity instead of the configured site. Preserve WithSiteID while adding the guard, following history-service's composition pattern, or extend DefaultGuarded to accept additional router options and update these call sites.

</details>

<details>

```
In `room-service/main.go` around line 347: This replacement drops both `WithSiteID(cfg.SiteID)` and `WithMetrics(publishMetrics)` because DefaultGuarded forwards only GuardConfig.Options. The same regression drops site identity in bot-message-handler, bot-room-service, media-service, room-worker, and search-service, while room-worker also loses its metrics sink. During a mixed-version rollout, requests handled by new replicas therefore lack the previous site attribution, and overload rejections/timeouts in room-service and room-worker can no longer be counted through the configured router metrics. Preserve the existing options when adding guard options, as history-service does, or extend DefaultGuarded to accept additional Option values and update every migrated caller.

```

</details>

search-service/main.go
```diff
@@ -233,17 +242,11 @@ func main() {
...
 
-	// Bound in-flight handlers so a burst is shed at the door (ErrUnavailable)
-	// instead of piling unbounded work onto Elasticsearch/MongoDB.
-	// MAX_CONCURRENCY=0 disables.
-	routerOpts := []natsrouter.Option{natsrouter.WithSiteID(cfg.SiteID)}
-	if cfg.MaxConcurrency > 0 {
-		routerOpts = append(routerOpts, natsrouter.WithMaxConcurrency(cfg.MaxConcurrency))
-	}
-	router := natsrouter.New(nc, "search-service", routerOpts...)
+	router := natsrouter.New(nc, "search-service", cfg.Guard.Options()...)
```
julianshen, line 245: _🎯 search correctness_ | _🟠 Warning_ | _`elasticsearch`, `tgd-review` (2 rules)_

**The search router no longer receives its configured site ID.**

Every NATS search request now uses only cfg.Guard.Options(). GuardConfig.Options returns solely WithMaxConcurrency, so the previous WithSiteID(cfg.SiteID) is silently dropped. This removes router-level site identity from site-scoped search requests and can produce incorrect or empty Elasticsearch results where routing/context supplies the site scope. Preserve the site option while adding the guard options.

<details>

> ‼️ **IMPORTANT**

```suggestion
routerOpts := append([]natsrouter.Option{natsrouter.WithSiteID(cfg.SiteID)}, cfg.Guard.Options()...)
	router := natsrouter.New(nc, "search-service", routerOpts...)
```

</details>

<details>

- **`tgd-review`**:245: The previous router options always included `natsrouter.WithSiteID(cfg.SiteID)`, but the replacement passes only guard options. Preserve the site option while adding the guard so request context and site-aware instrumentation retain their prior behavior.

</details>

<details>

```
In `search-service/main.go` around line 245: Every NATS search request now uses only cfg.Guard.Options(). GuardConfig.Options returns solely WithMaxConcurrency, so the previous WithSiteID(cfg.SiteID) is silently dropped. This removes router-level site identity from site-scoped search requests and can produce incorrect or empty Elasticsearch results where routing/context supplies the site scope. Preserve the site option while adding the guard options.

```

</details>

pkg/mongoutil/poolconfig.go
```diff
@@ -0,0 +1,50 @@
...
+			c.minPoolSize = &p.MinPoolSize
+		}
+	}
+}
```
julianshen, line 39: _🏷️ configuration_ | _🟠 Warning_ | _`mongodb`_

**A zero minimum can preserve an incompatible URI pool floor.**

WithPool overrides the URI's maxPoolSize but omits MinPoolSize when the configured value is zero. Because Connect applies the URI first, a URI such as `?minPoolSize=200` combined with the default max of 150 passes PoolConfig.Validate as `{150,0}` but reaches the driver as `{150,200}`, causing connection startup to fail. It also means `MONGO_MIN_POOL_SIZE=0` cannot clear an existing URI floor. Apply the configured minimum, including zero, so validation and the driver's effective pool settings use the same pair.

<details>

> ‼️ **IMPORTANT**

```suggestion
// WithPool is a Connect option that applies this pool configuration, so call
// sites read `Connect(ctx, uri, u, p, mongoutil.WithPool(cfg.Pool), ...)` rather
// than hand-splicing pool options into the variadic. Both limits are applied
// authoritatively so Validate checks the same pair that reaches the driver,
// including an explicit zero minimum.
func WithPool(p PoolConfig) Option {
	return func(c *connectConfig) {
		c.maxPoolSize = &p.MaxPoolSize
		c.minPoolSize = &p.MinPoolSize
	}
}
```

</details>

<details>

```
In `pkg/mongoutil/poolconfig.go` around line 26: WithPool overrides the URI's maxPoolSize but omits MinPoolSize when the configured value is zero. Because Connect applies the URI first, a URI such as `?minPoolSize=200` combined with the default max of 150 passes PoolConfig.Validate as `{150,0}` but reaches the driver as `{150,200}`, causing connection startup to fail. It also means `MONGO_MIN_POOL_SIZE=0` cannot clear an existing URI floor. Apply the configured minimum, including zero, so validation and the driver's effective pool settings use the same pair.

```

</details>

history-service/internal/config/config.go
```diff
@@ -56,25 +48,22 @@ type Config struct {
...
 	Cassandra               CassandraConfig `envPrefix:"CASSANDRA_"`
 	Mongo                   MongoConfig     `envPrefix:"MONGO_"`
-	NATS                    NATSConfig      `envPrefix:"NATS_"`
-	MessageBucketHours      int             `env:"MESSAGE_BUCKET_HOURS"        envDefault:"360"`
-	MessageReadMaxBuckets   int             `env:"MESSAGE_READ_MAX_BUCKETS"    envDefault:"122"`
-	MessageHistoryFloorDays int             `env:"MESSAGE_HISTORY_FLOOR_DAYS"  envDefault:"730"`
-	LargeRoomThreshold      int             `env:"LARGE_ROOM_THRESHOLD"        envDefault:"500"`
-	MaxPinnedPerRoom        int             `env:"MAX_PINNED_PER_ROOM"         envDefault:"10"`
-	PinEnabled              bool            `env:"PIN_ENABLED"                 envDefault:"true"`
+	Pool                    mongoutil.PoolConfig
```
julianshen, line 51: _⚡ performance_ | _🟠 Warning_ | _`mongodb`, `performance-and-concurrency` (2 rules)_

**History's default Mongo pool shrinks from 500 to 150**

Replacing the service-specific pool fields with the shared PoolConfig changes the production fallback from 500 to 150. The compose override does not protect deployments that previously relied on the code default, and this service still admits 256 concurrent handlers. Under concurrent Mongo-heavy requests, at least 106 handlers can block waiting for pool checkout, materially increasing tail latency and reducing throughput. Preserve a history-service default of 500 in its production deployment/configuration, or provide a service-specific default while retaining the shared validation and option wiring.

<details>

- **`mongodb`**:51: The previous history-specific MongoConfig defaulted MONGO_MAX_POOL_SIZE to 500, but the shared PoolConfig defaults it to 150. Only the local Docker Compose file restores 500; any deployment that does not explicitly inject the variable now runs a 150-connection pool while retaining MAX_CONCURRENCY=256 and REQUEST_TIMEOUT=10s. Under the documented high-concurrency history workload, handlers beyond the 150 available checkouts can queue and expire. Preserve a history-specific 500 default or update every production deployment to set MONGO_MAX_POOL_SIZE=500 and add a parsed-default test.

</details>

<details>

```
In `history-service/internal/config/config.go` around line 51: Replacing the service-specific pool fields with the shared PoolConfig changes the production fallback from 500 to 150. The compose override does not protect deployments that previously relied on the code default, and this service still admits 256 concurrent handlers. Under concurrent Mongo-heavy requests, at least 106 handlers can block waiting for pool checkout, materially increasing tail latency and reducing throughput. Preserve a history-service default of 500 in its production deployment/configuration, or provide a service-specific default while retaining the shared validation and option wiring.

```

</details>

room-service/main.go
julianshen, line 344: _🏷️ observability_ | _🟠 Warning_ | _`nats`_

**Room RPC routers no longer install NATS metrics.**

Both room-service and room-worker previously constructed their routers with WithMetrics(publishMetrics). Their new DefaultGuarded calls cannot pass that option, so requests, responses, and overload failures on these RPC subscriptions are no longer recorded by the configured NATS metrics collector. Restore WithMetrics(publishMetrics) while composing the guard options, as history-service does.

<details>

```
In `room-service/main.go` around line 347: Both room-service and room-worker previously constructed their routers with WithMetrics(publishMetrics). Their new DefaultGuarded calls cannot pass that option, so requests, responses, and overload failures on these RPC subscriptions are no longer recorded by the configured NATS metrics collector. Restore WithMetrics(publishMetrics) while composing the guard options, as history-service does.

```

</details>

pkg/ginutil/timeout.go
```diff
@@ -0,0 +1,25 @@
...
+// a slow downstream op (e.g. a MongoDB query) is cancelled and its pooled
+// connection released rather than held until the pool starves. It complements
+// the http.Server read/write timeouts, which bound the socket but do not cancel
+// the handler's context. A non-positive duration is a no-op.
+func Timeout(d time.Duration) gin.HandlerFunc {
+	if d <= 0 {
+		return func(c *gin.Context) { c.Next() }
+	}
+	return func(c *gin.Context) {
+		ctx, cancel := context.WithTimeout(c.Request.Context(), d)
```
julianshen, line 20: _🏷️ error-handling_ | _🟠 Warning_ | _`silent-failure`_

**Expired requests can still return successful responses**

This middleware only cancels the request context; it neither stops the handler nor converts expiration into a timeout response. The new test demonstrates the concrete failure path: the handler waits for Context().Done() and then writes HTTP 200. Any handler that continues after cancellation, or returns without translating context.DeadlineExceeded, can therefore report success despite timed-out or incomplete downstream work. Use an HTTP timeout implementation that buffers the response and emits 504 when the deadline wins, or require every affected handler to check the context before committing a success response.

<details>

```
In `pkg/ginutil/timeout.go` around line 20: This middleware only cancels the request context; it neither stops the handler nor converts expiration into a timeout response. The new test demonstrates the concrete failure path: the handler waits for Context().Done() and then writes HTTP 200. Any handler that continues after cancellation, or returns without translating context.DeadlineExceeded, can therefore report success despite timed-out or incomplete downstream work. Use an HTTP timeout implementation that buffers the response and emits 504 when the deadline wins, or require every affected handler to check the context before committing a success response.

```

</details>

pkg/natsrouter/guard.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// DefaultGuarded builds a Router with the standard middleware chain (Default:
+// Recovery, RequestID, Logging) plus this guard's admission cap and per-request
+// timeout, wired in the correct order. It exists so a service can't apply only
+// half of the overload protection — the cap (a construction option) and the
+// timeout (post-middleware) are otherwise two separate calls that are easy to
+// forget one of. Services needing a non-standard middleware order keep New.
+func DefaultGuarded(nc *o11ynats.Conn, queue string, g GuardConfig) *Router {
+	r := Default(nc, queue, g.Options()...)
```
julianshen, line 64: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**DefaultGuarded migrations discard site and metrics options**

The helper forwards only `g.Options()`, and its migrated callers therefore lose options that were previously installed: `WithSiteID` in bot-message-handler, bot-room-service, media-service, room-service, and room-worker, plus `WithMetrics` in the two room services. This removes the routers' site context and request metrics. Accept additional router options in `DefaultGuarded`, combine them with the guard options, and update each caller to pass its original options.

<details>

```
In `pkg/natsrouter/guard.go` around line 64: The helper forwards only `g.Options()`, and its migrated callers therefore lose options that were previously installed: `WithSiteID` in bot-message-handler, bot-room-service, media-service, room-service, and room-worker, plus `WithMetrics` in the two room services. This removes the routers' site context and request metrics. Accept additional router options in `DefaultGuarded`, combine them with the guard options, and update each caller to pass its original options.

```

</details>

pkg/natsrouter/guard.go
```diff
@@ -0,0 +1,67 @@
...
+	r.Use(g.TimeoutMiddleware()...)
+	return r
+}
```
julianshen, line 67: _🎯 correctness_ | _🔴 Blocking_ | _`distributed-system`, `nats`, `tgd-review` (3 rules)_

**DefaultGuarded drops WithSiteID and WithMetrics that migrated routers previously set.**

DefaultGuarded forwards only g.Options() (which yields solely WithMaxConcurrency). Every migrated caller therefore silently loses options it installed before: WithSiteID(cfg.SiteID) in bot-message-handler/main.go, bot-room-service/main.go, media-service/main.go, room-service/main.go, room-worker/main.go, and search-service/main.go; plus WithMetrics(publishMetrics) in room-service and room-worker. Site attribution disappears from every request handled by these routers, and NATS RPC metrics (requests, replies, overload rejections) stop being recorded in the two room services — a regression unrelated to this PR's stated goal of consolidating pool/guard configuration. history-service shows the intended composition pattern (explicitly appending WithSiteID/WithMetrics alongside cfg.Guard.Options()); the other six services don't follow it because DefaultGuarded gives them nowhere to put extra options. Fix: accept variadic extra options in DefaultGuarded (backward compatible) and update each caller to pass its original WithSiteID/WithMetrics options.

<details>

> ‼️ **IMPORTANT**

```suggestion
func DefaultGuarded(nc *o11ynats.Conn, queue string, g GuardConfig, extra ...Option) *Router {
	opts := append(extra, g.Options()...)
	r := Default(nc, queue, opts...)
	r.Use(g.TimeoutMiddleware()...)
	return r
}
```

</details>

<details>

- **`nats`**:63: DefaultGuarded forwards only g.Options() to Default, so migrating a router to it silently discards every option the caller used to pass. Concretely, in this diff: bot-message-handler/main.go and bot-room-service/main.go lose natsrouter.WithSiteID(cfg.SiteID); media-service/main.go loses WithSiteID; room-service/main.go loses both WithSiteID(cfg.SiteID) and WithMetrics(publishMetrics); room-worker/main.go loses WithSiteID and WithMetrics; search-service/main.go replaces WithSiteID with cfg.Guard.Options() alone. Every request served by these routers now runs without the configured site identity (degrading site-scoped context, logs, and traces) and, in the two room services, NATS request/reply metrics — including overload rejections and timeouts introduced by the guard itself — are no longer recorded, so saturation becomes invisible exactly where the guard sheds load. history-service shows the correct composition (explicitly prepending WithSiteID and WithMetrics around cfg.Guard.Options()); the helper should support the same. Extend DefaultGuarded with variadic Option parameters merged ahead of the guard options, then restore the dropped options at each of the six call sites.

<details>

```text
func DefaultGuarded(nc *o11ynats.Conn, queue string, g GuardConfig, opts ...Option) *Router {
	all := append(append([]Option{}, opts...), g.Options()...)
	r := Default(nc, queue, all...)
	r.Use(g.TimeoutMiddleware()...)
	return r
}
```

</details>

- **`distributed-system`**:60: GuardConfig.Options() returns only WithMaxConcurrency, so every caller that switched to DefaultGuarded silently lost options it previously installed: WithSiteID(cfg.SiteID) in bot-message-handler, bot-room-service, media-service, room-service, and room-worker, plus WithMetrics(publishMetrics) in room-service and room-worker. During a mixed-version rollout, requests handled by new replicas lose site attribution in request context/logging, and overload rejections/timeouts on the room services are no longer counted through the configured NATS metrics — masking exactly the overload behavior this PR is meant to make observable. Extend DefaultGuarded to accept extra router options (as history-service does by composing its own option slice), then update each caller to pass its original options.

<details>

```text
func DefaultGuarded(nc *o11ynats.Conn, queue string, g GuardConfig, extra ...Option) *Router {
	r := Default(nc, queue, append(append([]Option{}, extra...), g.Options()...)...)
	r.Use(g.TimeoutMiddleware()...)
	return r
}
```

</details>


</details>

<details>

```
In `pkg/natsrouter/guard.go` around line 63: DefaultGuarded forwards only g.Options() (which yields solely WithMaxConcurrency). Every migrated caller therefore silently loses options it installed before: WithSiteID(cfg.SiteID) in bot-message-handler/main.go, bot-room-service/main.go, media-service/main.go, room-service/main.go, room-worker/main.go, and search-service/main.go; plus WithMetrics(publishMetrics) in room-service and room-worker. Site attribution disappears from every request handled by these routers, and NATS RPC metrics (requests, replies, overload rejections) stop being recorded in the two room services — a regression unrelated to this PR's stated goal of consolidating pool/guard configuration. history-service shows the intended composition pattern (explicitly appending WithSiteID/WithMetrics alongside cfg.Guard.Options()); the other six services don't follow it because DefaultGuarded gives them nowhere to put extra options. Fix: accept variadic extra options in DefaultGuarded (backward compatible) and update each caller to pass its original WithSiteID/WithMetrics options.

```

</details>

pkg/ginutil/timeout.go
```diff
@@ -0,0 +1,25 @@
...
+// Timeout bounds each request by deriving a context deadline on the request, so
+// a slow downstream op (e.g. a MongoDB query) is cancelled and its pooled
+// connection released rather than held until the pool starves. It complements
+// the http.Server read/write timeouts, which bound the socket but do not cancel
+// the handler's context. A non-positive duration is a no-op.
+func Timeout(d time.Duration) gin.HandlerFunc {
+	if d <= 0 {
+		return func(c *gin.Context) { c.Next() }
+	}
+	return func(c *gin.Context) {
```
julianshen, line 19: _🏷️ observability_ | _🟠 Warning_ | _`distributed-system`_

**Deadline-only timeout lets timed-out handlers still return HTTP 200.**

Timeout only derives a context deadline; it neither aborts the response nor converts expiry into a timeout status. The package's own test demonstrates the failure path: the handler observes Context().Done() and then writes 200. Any handler wired under this middleware (admin-service, botplatform-service, portal-service, tcard-service) that ignores the context error after a downstream Mongo/Valkey op is cancelled can commit a success response over incomplete work — the client cannot distinguish accepted from failed, and retries may not happen. Either emit 504 when the deadline wins before the handler writes, or audit/require every covered handler to translate context.DeadlineExceeded into an error response instead of success.

<details>

```
In `pkg/ginutil/timeout.go` around line 19: Timeout only derives a context deadline; it neither aborts the response nor converts expiry into a timeout status. The package's own test demonstrates the failure path: the handler observes Context().Done() and then writes 200. Any handler wired under this middleware (admin-service, botplatform-service, portal-service, tcard-service) that ignores the context error after a downstream Mongo/Valkey op is cancelled can commit a success response over incomplete work — the client cannot distinguish accepted from failed, and retries may not happen. Either emit 504 when the deadline wins before the handler writes, or audit/require every covered handler to translate context.DeadlineExceeded into an error response instead of success.

```

</details>

user-presence-service/main.go
```diff
@@ -128,7 +153,15 @@ func main() {
 	peer := NewNATSPeerPresenceClient(nc.NatsConn(), cfg.Presence.PeerTimeout)
 	handler := NewHandler(store, userDir, peer, publish, cfg.SiteID, cfg.Presence.BatchMax)
 
+	// No admission cap here: Hello/Ping/Activity/Bye are fire-and-forget
+	// (RegisterVoid), and under a saturated concurrency semaphore those are
+	// silently dropped (no reply, no redelivery) — a reconnect storm would lose
+	// presence updates and strand users online/offline. Only the per-request
+	// timeout is applied.
 	router := natsrouter.Default(nc, "user-presence-service", natsrouter.WithSiteID(cfg.SiteID))
+	if cfg.RequestTimeout > 0 {
```
julianshen, line 162: _🎯 nats-correctness_ | _🟠 Warning_ | _`distributed-system`, `nats`, `silent-failure` (3 rules)_

**HandlerTimeout on fire-and-forget presence routes silently drops updates**

Hello/Ping/Activity/Bye are registered via natsrouter.RegisterVoid: the producers publish fire-and-forget with no inbox, no reply, and no redelivery, so the handler result is the only delivery attempt. Applying HandlerTimeout(cfg.RequestTimeout) gives each handler a cancellable context; if a Mongo/Valkey op exceeds REQUEST_TIMEOUT (default 10s, e.g. during a reconnect storm or secondary lag), the context is cancelled mid-write, the update is discarded, and the producer has already completed its publish — the failure is never propagated anywhere. A dropped Bye strands a user online until the sweep/expiry path repairs it; a dropped Hello delays presence convergence. Note this service deliberately omits the concurrency semaphore for exactly this 'silent drop' reason (see the comment above line 161), yet the timeout reintroduces the same silent-loss failure mode on the time axis. Either exempt these void routes from the canceling timeout (relying on the existing sweep for liveness) or move presence updates to an acknowledged mechanism with idempotent retries before bounding them with a deadline. The fix is not expressible as a one-line replacement because it changes which middleware applies to which registration set.

<details>

- **`silent-failure`**:159: The code comment correctly rejects an admission cap because RegisterVoid messages have no reply or redelivery, but then applies HandlerTimeout to exactly those routes. If a Hello/Ping/Activity/Bye handler's Mongo/Valkey op exceeds REQUEST_TIMEOUT, the context cancellation aborts the operation mid-way; the producer's publish has already completed, nothing replies, nothing retries, and no error is recorded anywhere — the failed update is converted into apparent delivery. A lost Bye keeps a departed user shown online until the sweeper expires them; a lost Hello delays presence until the next ping. Either exclude the void routes from the cancelling timeout (apply it only to request/reply registrations), or make these handlers tolerate cancellation idempotently and log/record the abandoned update rather than discarding it.
- **`distributed-system`**:161: The code itself documents that Hello/Ping/Activity/Bye are RegisterVoid: no reply, no redelivery, producer has already moved on once published. Under load or Mongo/Valkey slowness (>REQUEST_TIMEOUT=10s), HandlerTimeout cancels these handlers mid-operation and the update is lost with no retry and no error surfaced anywhere — precisely the silent-drop class the comment says the admission cap was omitted to avoid. A lost Bye strands a user as online until TTL expiry repairs it; a burst of lost Hellos/Pings during a reconnect storm degrades presence fleet-wide. Since the pool-exhaustion concern only needs the connection released (which cancellation does), either drop the timeout on these void routes or route presence writes through an acknowledged/durable path with idempotent application.

</details>

<details>

```
In `user-presence-service/main.go` around line 162: Hello/Ping/Activity/Bye are registered via natsrouter.RegisterVoid: the producers publish fire-and-forget with no inbox, no reply, and no redelivery, so the handler result is the only delivery attempt. Applying HandlerTimeout(cfg.RequestTimeout) gives each handler a cancellable context; if a Mongo/Valkey op exceeds REQUEST_TIMEOUT (default 10s, e.g. during a reconnect storm or secondary lag), the context is cancelled mid-write, the update is discarded, and the producer has already completed its publish — the failure is never propagated anywhere. A dropped Bye strands a user online until the sweep/expiry path repairs it; a dropped Hello delays presence convergence. Note this service deliberately omits the concurrency semaphore for exactly this 'silent drop' reason (see the comment above line 161), yet the timeout reintroduces the same silent-loss failure mode on the time axis. Either exempt these void routes from the canceling timeout (relying on the existing sweep for liveness) or move presence updates to an acknowledged mechanism with idempotent retries before bounding them with a deadline. The fix is not expressible as a one-line replacement because it changes which middleware applies to which registration set.

```

</details>

pkg/mongoutil/poolconfig.go
julianshen, line 39: _🎯 correctness_ | _🟠 Warning_ | _`distributed-system`, `tgd-review` (2 rules)_

**WithPool forces maxPoolSize but skips zero MinPoolSize, so a URI min can exceed the forced max.**

WithPool always overrides maxPoolSize but omits minPoolSize entirely when the configured MinPoolSize is 0, leaving any minPoolSize in the connection URI in effect. A URI like mongodb://host/?minPoolSize=200&maxPoolSize=500 combined with the shared default PoolConfig{150,0} passes Validate() (which only checks 0 <= 150) but reaches the driver as min=200/max=150, which the Mongo client rejects at construction — upgraded replicas fail to connect during rollout. It also means operators can never clear a URI-supplied floor via MONGO_MIN_POOL_SIZE=0. The pre-refactor history-service wiring (WithMinPoolSize applied unconditionally, including 0) made the configured pair authoritative; restore that symmetry so Validate() checks exactly what reaches the driver. Note this changes the contract asserted by TestWithPool_OmitsMinPoolSizeWhenZero, which must be updated to expect the pointer set to 0.

<details>

> ‼️ **IMPORTANT**

```suggestion
func WithPool(p PoolConfig) Option {
	return func(c *connectConfig) {
		c.maxPoolSize = &p.MaxPoolSize
		c.minPoolSize = &p.MinPoolSize
	}
}
```

</details>

<details>

- **`distributed-system`**:31: WithPool always overrides maxPoolSize but skips minPoolSize when the config value is zero, deliberately leaving a URI-supplied minPoolSize intact. If an existing connection string carries ?minPoolSize=N, Validate() sees {defaultMax,0} and passes, but the driver receives min=N with max forced to 150 (the new default) or whatever MONGO_MAX_POOL_SIZE is set to. When N > max, driver-side validation rejects the client options and every upgraded replica of that service fails to connect during the rolling deploy — an outage introduced purely by adopting PoolConfig. Make the configured pair authoritative end-to-end: apply MinPoolSize even when zero (and update TestWithPool_OmitsMinPoolSizeWhenZero), so Validate checks exactly what reaches the driver and operators can explicitly clear a URI floor with MONGO_MIN_POOL_SIZE=0.

</details>

<details>

```
In `pkg/mongoutil/poolconfig.go` around line 32: WithPool always overrides maxPoolSize but omits minPoolSize entirely when the configured MinPoolSize is 0, leaving any minPoolSize in the connection URI in effect. A URI like mongodb://host/?minPoolSize=200&maxPoolSize=500 combined with the shared default PoolConfig{150,0} passes Validate() (which only checks 0 <= 150) but reaches the driver as min=200/max=150, which the Mongo client rejects at construction — upgraded replicas fail to connect during rollout. It also means operators can never clear a URI-supplied floor via MONGO_MIN_POOL_SIZE=0. The pre-refactor history-service wiring (WithMinPoolSize applied unconditionally, including 0) made the configured pair authoritative; restore that symmetry so Validate() checks exactly what reaches the driver. Note this changes the contract asserted by TestWithPool_OmitsMinPoolSizeWhenZero, which must be updated to expect the pointer set to 0.

```

</details>

pkg/ginutil/timeout.go
julianshen, line 19: _🏷️ silent-failure_ | _🟠 Warning_ | _`silent-failure`_

**Timeout middleware lets an expired request still complete with a success status.**

The middleware only installs a deadline on the request context; it neither stops the handler nor surfaces the expiry. The new test itself demonstrates the masking path: the handler observes Context().Done() and then writes 200. Any handler that proceeds past a cancelled context — or returns an error path that never translates context.DeadlineExceeded into a status — reports success to the client even though the downstream Mongo/query work was aborted mid-flight, i.e. a timeout converted into apparent success. Minimal fix: after c.Next(), if the derived context has expired and the handler wrote nothing, emit 504 instead of letting Gin's implicit 200 stand. (Handlers that already wrote partial output are left untouched.)

<details>

```
In `pkg/ginutil/timeout.go` around line 19: The middleware only installs a deadline on the request context; it neither stops the handler nor surfaces the expiry. The new test itself demonstrates the masking path: the handler observes Context().Done() and then writes 200. Any handler that proceeds past a cancelled context — or returns an error path that never translates context.DeadlineExceeded into a status — reports success to the client even though the downstream Mongo/query work was aborted mid-flight, i.e. a timeout converted into apparent success. Minimal fix: after c.Next(), if the derived context has expired and the handler wrote nothing, emit 504 instead of letting Gin's implicit 200 stand. (Handlers that already wrote partial output are left untouched.)

```

</details>

julianshen (commented):
tGD inline review

julianshen (commented):
tGD inline review

pkg/mongoutil/poolconfig.go
julianshen, line 39: _🏷️ mongodb-configuration_ | _🔴 Blocking_ | _`mongodb`_

**A URI minimum can exceed the forced pool maximum.**

Every Mongo client using WithPool overrides maxPoolSize but preserves a URI-supplied minPoolSize when MONGO_MIN_POOL_SIZE is zero. For example, a previously valid URI with minPoolSize=200 combined with the shared default maximum of 150 passes PoolConfig.Validate but reaches the driver as min=200/max=150, causing client initialization to fail. Apply the configured minimum authoritatively, including zero, and update TestWithPool_OmitsMinPoolSizeWhenZero to expect an explicit zero.

<details>

> ‼️ **IMPORTANT**

```suggestion
func WithPool(p PoolConfig) Option {
	return func(c *connectConfig) {
		c.maxPoolSize = &p.MaxPoolSize
		c.minPoolSize = &p.MinPoolSize
	}
}
```

</details>

<details>

```
In `pkg/mongoutil/poolconfig.go` around line 32: Every Mongo client using WithPool overrides maxPoolSize but preserves a URI-supplied minPoolSize when MONGO_MIN_POOL_SIZE is zero. For example, a previously valid URI with minPoolSize=200 combined with the shared default maximum of 150 passes PoolConfig.Validate but reaches the driver as min=200/max=150, causing client initialization to fail. Apply the configured minimum authoritatively, including zero, and update TestWithPool_OmitsMinPoolSizeWhenZero to expect an explicit zero.

```

</details>

julianshen (commented):
tGD inline review

pkg/mongoutil/poolconfig.go
```diff
@@ -0,0 +1,50 @@
...
+// sites read `Connect(ctx, uri, u, p, mongoutil.WithPool(cfg.Pool), ...)` rather
+// than hand-splicing pool options into the variadic. MaxPoolSize is always
+// applied (authoritative, validated >= 1); MinPoolSize is applied only when > 0,
+// so an unset floor leaves a minPoolSize from the URI (or the driver default)
+// intact rather than forcing it to 0.
+func WithPool(p PoolConfig) Option {
+	return func(c *connectConfig) {
+		c.maxPoolSize = &p.MaxPoolSize
+		if p.MinPoolSize > 0 {
+			c.minPoolSize = &p.MinPoolSize
```
julianshen, line 36: _🏷️ rolling-deploy_ | _🔴 Blocking_ | _`distributed-system`_

**A URI minimum can exceed the forced pool maximum.**

WithPool always overrides maxPoolSize but preserves a URI-supplied minPoolSize when the configured minimum is zero. During rollout, an existing valid URI such as minPoolSize=200&maxPoolSize=500 combines with the new default as min=200/max=150. PoolConfig.Validate sees only 0/150 and passes, but Mongo rejects the effective options, so upgraded replicas cannot start. Apply MinPoolSize authoritatively even when zero, and update the zero-minimum test accordingly.

<details>

```
In `pkg/mongoutil/poolconfig.go` around line 36: WithPool always overrides maxPoolSize but preserves a URI-supplied minPoolSize when the configured minimum is zero. During rollout, an existing valid URI such as minPoolSize=200&maxPoolSize=500 combines with the new default as min=200/max=150. PoolConfig.Validate sees only 0/150 and passes, but Mongo rejects the effective options, so upgraded replicas cannot start. Apply MinPoolSize authoritatively even when zero, and update the zero-minimum test accordingly.

```

</details>

pkg/natsrouter/guard.go
```diff
@@ -0,0 +1,67 @@
...
+	return nil
+}
+
+// DefaultGuarded builds a Router with the standard middleware chain (Default:
+// Recovery, RequestID, Logging) plus this guard's admission cap and per-request
+// timeout, wired in the correct order. It exists so a service can't apply only
+// half of the overload protection — the cap (a construction option) and the
+// timeout (post-middleware) are otherwise two separate calls that are easy to
+// forget one of. Services needing a non-standard middleware order keep New.
+func DefaultGuarded(nc *o11ynats.Conn, queue string, g GuardConfig) *Router {
```
julianshen, line 63: _🏷️ nats-operational-safety_ | _🟠 Warning_ | _`nats`_

**Guard migrations drop site identity and NATS metrics options**

`DefaultGuarded` forwards only `g.Options()`, which contains only the concurrency option. Consequently, migrated routers in bot-message-handler, bot-room-service, media-service, room-service, and room-worker lose their previous `WithSiteID(cfg.SiteID)` option; room-service and room-worker also lose `WithMetrics(publishMetrics)`. Search-service makes the equivalent mistake by replacing its site option with `cfg.Guard.Options()`. Requests therefore run with the router's empty/default site identity, and overload rejections or timeouts in the room routers are no longer recorded by their configured metrics sink. Extend `DefaultGuarded` to accept additional `Option` values, restore each caller's previous site/metrics options, and prepend `WithSiteID` to the search router's guard options.

<details>

```
In `pkg/natsrouter/guard.go` around line 63: `DefaultGuarded` forwards only `g.Options()`, which contains only the concurrency option. Consequently, migrated routers in bot-message-handler, bot-room-service, media-service, room-service, and room-worker lose their previous `WithSiteID(cfg.SiteID)` option; room-service and room-worker also lose `WithMetrics(publishMetrics)`. Search-service makes the equivalent mistake by replacing its site option with `cfg.Guard.Options()`. Requests therefore run with the router's empty/default site identity, and overload rejections or timeouts in the room routers are no longer recorded by their configured metrics sink. Extend `DefaultGuarded` to accept additional `Option` values, restore each caller's previous site/metrics options, and prepend `WithSiteID` to the search router's guard options.

```

</details>

PR #350, opened 2026-08-22
fix(natsrouter): reply response_too_large instead of timing out on oversize

mliu33:
## Problem

`docs/client-api.md` §6 already promises that **any** RPC whose reply exceeds the transport `max_payload` comes back as `internal` / `response_too_large`. The live path did not deliver it.

Every client-facing request/reply goes through `pkg/natsrouter` (confirmed — no raw `QueueSubscribe` request/reply handlers remain outside the router). Its `Context.respond` logged `"reply failed"` and returned, so the caller got a **bare timeout** — indistinguishable from a dead service, even though the two call for opposite recoveries: narrow the request vs. retry as-is.

The only implementation of the fallback sat in `natsutil.ReplyJSON`, which has had **no production caller** since the router migration. Meanwhile `room-service`'s proactive size guard returned its oversize error with **no `reason` at all**, so even the one path that caught the condition early missed the documented branch.

Three separate spellings of the reason string existed, none of them a registered constant.

## Change

| File | Change |
|---|---|
| `pkg/natsrouter/context.go` | **The fix.** `respond()` retries once with a compact envelope on `nats.ErrMaxPayload`. Sent through a new `send()` rather than `respond()` so a failing fallback cannot recurse. |
| `pkg/errcode/codes_platform.go` | New `ResponseTooLarge` reason |
| `pkg/errcode/errnats/reply.go` | `IsOversize`, `OversizeEnvelope` (built once at init, returned as a fresh copy), plus the fallback in `Reply`/`ReplyQuiet` |
| `pkg/natsutil/reply.go`, `room-service/helper.go`, `tools/loadgen/soak_rpc.go` | Point the three literals at the shared constant |
| `docs/client-api.md` | Widened one sentence: the fallback now covers the error lane too, not only success bodies |

### Why reactive rather than a pre-marshal size check

`nats.go` refuses an oversize publish **synchronously, before anything reaches the wire**, so the reply subject is still live and a compact envelope still gets through. The broker's limit also counts the trace headers the o11y facade injects, which a byte count over the response alone would miss.

## Verification

Built test-first; each test was watched failing for the right reason before the implementation existed.

**End-to-end** (`pkg/natsrouter/oversize_integration_test.go`, `//go:build integration`) — real NATS container capped at 4 KiB `max_payload`, real router route, real client request:

| | Without the fix | With the fix |
|---|---|---|
| oversize success body | `FAIL` — "the caller must get an answer, not a timeout" @ **5.61s** | `PASS` @ **0.49s** |
| oversize error envelope | `FAIL` — same @ **5.46s** | `PASS` @ **0.46s** |
| reply that fits | `PASS` (unaffected) | `PASS` (unaffected) |

The "reply that fits" case is asserted separately so the fallback cannot mask successful responses. Inline `testcontainers` rather than `testutil.NATS(t)` because the shared broker is process-wide at the 1 MiB default and lowering it would affect every other test; the container ref is stored and `t.Cleanup(Terminate)` registered.

Reverting the fix also reproduced the equivalent 2.01s timeout in the `errnats` unit tests.

**Gates:**
- `make lint` → 0 issues
- `make test` (all 112 packages, `-race`) → exit 0, 0 failures
- `make test-integration` for `pkg/natsrouter`, `pkg/errcode`, `pkg/natsutil`, `room-service` → all ok
- `make sast` → gosec **PASS**, govulncheck **PASS**, semgrep **PASS**

**Coverage:** `IsOversize`, `OversizeEnvelope`, `respondOversize` at **100%**; packages at 91.3% / 88.9% / 87.9%, all clear of the 80% floor.

## Notes for reviewers

- No handler registration, request/reply struct, or event struct changed, so no new `docs/client-api.md` rows are needed — the docs were already the promise; this makes the code keep it. The one doc edit is a precision fix.
- `natsutil.ReplyJSON` is left in place (still no production callers) rather than deleted — removing it is a separate cleanup.

---
_Generated by [Claude Code](https://claude.ai/code/session_012vGY47HWtztTUL36YcEHpz)_

### DIFF
--- a/docs/client-api.md
+++ b/docs/client-api.md
@@ -6718,7 +6718,7 @@ Every error response — NATS reply subjects, JetStream async results, and HTTP 
 > **Malformed request bodies.** Any room request/reply RPC whose payload is not valid JSON for its schema is rejected uniformly with `code: bad_request` and the message `"invalid request payload"` — the transport layer rejects it before the handler runs. Treat this as a generic encoding error; do not pattern-match the message text.
 
 > [!IMPORTANT]
-> **Oversize replies.** If a successful response would exceed the transport's maximum payload size, the reply is returned as `code: internal` with `reason: response_too_large` instead of the success body. This is most likely on large history reads (e.g. Load History / Load Next / Load Surrounding / Get Thread Messages with a high `limit`); the client should retry with a smaller `limit`. Branch on `reason` (`response_too_large`), not the message text.
+> **Oversize replies.** If a reply would exceed the transport's maximum payload size, it is returned as `code: internal` with `reason: response_too_large` instead. This covers both a success body that is too large and an error envelope that is too large; either way the caller gets an answer rather than a timeout. This is most likely on large history reads (e.g. Load History / Load Next / Load Surrounding / Get Thread Messages with a high `limit`); the client should retry with a smaller `limit`. Branch on `reason` (`response_too_large`), not the message text.
 
 ### `reason` catalog (present today)
 
--- a/pkg/errcode/codes_platform.go
+++ b/pkg/errcode/codes_platform.go
@@ -25,4 +25,12 @@ const (
 	// NatsRequestTimeout marks a request that was delivered but not answered
 	// within the caller's timeout. Retryable.
 	NatsRequestTimeout Reason = "upstream_timeout"
+
+	// ResponseTooLarge marks a reply the transport refused because the
+	// marshalled response exceeds the broker's max_payload. Without the
+	// fallback envelope that carries this reason the request would simply
+	// time out, so the client cannot tell "too big" from "server down" —
+	// the two call for opposite recoveries. Retry with a narrower request
+	// (a smaller limit or page), never with the same one.
+	ResponseTooLarge Reason = "response_too_large"
 )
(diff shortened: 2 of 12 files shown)


chenjr0719 (approved):
Approving. This closes a real gap cleanly — an oversize reply was a bare timeout, indistinguishable from a dead service, when the two call for opposite client recoveries (narrow the request vs retry as-is). The fix is well-factored: the fallback is centralized in errnats (IsOversize/OversizeEnvelope), the retry goes through send() rather than respond() so a failing fallback cannot recurse, OversizeEnvelope() returns a genuine per-call copy via bytes.Clone (no shared-buffer race), and the three scattered reason literals now converge on the registered ResponseTooLarge constant — with room-service’\s proactive guard gaining the previously-absent reason to match the documented contract. Test coverage is thorough: unit plus a real-NATS 4 KiB-cap integration test over both the success-body and error-envelope lanes, and a no-double-fallback assertion.

One non-blocking observation for a possible fast-follow: the oversize-fallback reply is metered as RequestInternal, the same bucket as unrelated server failures. There is a publish-side payload_too_large outcome but no request-side response_too_large result, so dashboards cannot separate "client asked for too much" from "server broke." A dedicated RequestResult would recover that signal — out of scope here, just flagging.

PR #352, opened 2026-08-22
perf(user-service): count active subscriptions without the rooms join (option A: move the Del- check onto the subscription)

hmchangw:
**One of two alternative approaches to the same problem.** The sibling PR removes the `Del-` filter entirely. Pick one; they are mutually exclusive and both touch `CountActiveSubscriptions`.

## The problem

`CountActiveSubscriptions` ran an aggregation whose only reason to touch `rooms` was the soft-delete (`^Del-`) filter on `room.name`. For that one field it:

- did an indexed `rooms` lookup **per active subscription** — unbounded, unlike the unread path which caps at `MAX_SUBSCRIPTION_LIMIT` *before* joining;
- transported **eleven** room fields per row, including the E2E key blob (`encKey.priv`);
- computed **twelve** `$addFields` (the `__sortKey` among them) that `$count` then discarded.

All to answer a question about one string.

## The approach

`subscriptions.name` is already a denormalized copy of the room name, kept in step by `UpdateSubscriptionNamesForRoom` — room-worker for in-stack renames, inbox-worker for oplog and federated ones. Reading the marker from there makes every predicate local to the subscription, so the count becomes a plain `CountDocuments`: no join, no per-row enrichment, no per-subscription fan-out.

`GetActiveSubscriptions` shares `activeSubscriptionFilter` and so picks up the same predicate. Its rooms join stays (it needs `lastMsgAt`) but no longer re-applies the regex, and because the filter now runs in the leading `$match` ahead of the cap, a capped page is exact rather than possibly short.

## Behavior change

A cross-site subscription whose home site soft-deleted the room is now excluded from the count. The join could never see that — there was no local room doc to join — so the count kept it while the list dropped it later, after the `GetRoomsInfo` lookup. Both now agree, which is what `docs/client-api.md` already described.

## This is gated on a data property, not a code property

The rewrite assumes the denormalized name is faithful. That is a fact about the data, so this PR ships the probe that checks it: `tools/del-name-audit`, run via `make audit-del-names`. It is read-only (`ConnectRead`, secondaryPreferred) and checks both directions, because either alone misleads:

| Direction | Failure means |
|---|---|
| Every subscription of a `Del-` room carries the marker | the filter **keeps** a deleted room |
| No `Del-`-named subscription points at a live room | the filter **drops** a live room |

Results split by `roomType`, because `sub.name` is the room name only for `channel` — for `dm` it is the counterpart account, so a DM-only divergence would otherwise average away against the channel totals. Cross-site rows (no local room doc) report as *unverifiable* rather than as false positives, since they are excluded under both the join and the name filter. Exit 1 when the data does not support the rewrite, so it can gate a rollout.

**Run it per site before merging.** The subscription set is site-scoped and renames federate, so one site agreeing does not prove another does.

## Indexes

Unchanged. `{u.account, roomType}` still serves the leading equality, and a wider index would trade write amplification on a hot collection for a read this change has already made cheap.

## Testing

- Unit tests, `golangci-lint` (0 issues), `go vet -tags integration`, and `go build ./...` all pass.
- **Integration tests have not been run** — the authoring environment had no Docker. Fixtures and expected counts changed in `user-service/mongorepo/subscriptions_test.go`, so `make test-integration SERVICE=user-service` is the gate before merging.
- The probe's unit coverage is 54.5%; everything except the Mongo layer and CLI wiring is at 100%, with the remainder covered by integration tests that likewise have not run.
- The probe's implementation was written before its tests, against the repo's TDD rule. The tests were validated by mutating the implementation seven ways — all seven were caught. The `user-service` changes were done Red-first.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NZJaxhkaWPi9UEoeFBKZ3N)_

## Summary by CodeRabbit

* **New Features**
  * Added a read-only audit tool to verify consistency between soft-deleted rooms and subscription names.
  * Added configurable scan limits, JSON output, timeouts, and safety verdicts.
  * Added a Make target for running the audit.

* **Bug Fixes**
  * Improved subscription list and count handling for soft-deleted and cross-site rooms.
  * Preserved valid subscriptions when room records are unavailable.

* **Documentation**
  * Clarified soft-deleted room behavior and documented the audit tool.

* **Tests**
  * Added comprehensive unit and integration coverage for auditing and subscription filtering.


julianshen:
**2 findings · 2 inline comments posted** — 🟠 1 warning · 🔵 1 suggestion

### Local review memory

- `Makefile:524` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟡 Minor_ _⚡ Quick win_ **Limit the $lookup removal claim to counting.** subscription.list still uses the rooms $lookup for enrichment and lastMsgAt . The audit validates the deleted-name fil… ([thread](https://github.com/hmchangw/newchat/pull/352#discussion_r3835749314))
- `tools/del-name-audit/main.go:120` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟠 Major_ _🏗️ Heavy lift_ **Do not report an unverifiable Del- subscription as SAFE.** The old rooms join preserves a subscription when its room is absent locally. roomsEnrichStages(true) keeps… ([thread](https://github.com/hmchangw/newchat/pull/352#discussion_r3835749317))
- `tools/del-name-audit/README.md:12` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟡 Minor_ _⚡ Quick win_ **Describe the previous implementation in the past tense.** Line 8 states that CountActiveSubscriptions currently joins subscriptions to rooms . The current implementati… ([thread](https://github.com/hmchangw/newchat/pull/352#discussion_r3835749319))
- `tools/del-name-audit/store_mongo.go:62` — @coderabbitai: _🩺 Stability & Availability_ _🟡 Minor_ _⚡ Quick win_ **Wrap the batch query error with this function's context.** Line 62 returns a bare error. Add context that identifies the room-batch subscription query. &lt;details> &lt;summary>Propose… ([thread](https://github.com/hmchangw/newchat/pull/352#discussion_r3835749321))
- `user-service/service/service.go:28` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟡 Minor_ _⚡ Quick win_ **Do not document this as an exact-size page.** GetActiveSubscriptions returns up to limit , not exactly limit , when fewer active subscriptions exist. The integration tes… ([thread](https://github.com/hmchangw/newchat/pull/352#discussion_r3836262265))

<details>

* `.gitignore`
* `Makefile`
* `docs/client-api.md`
* `tools/del-name-audit/README.md`
* `tools/del-name-audit/audit.go`
* `tools/del-name-audit/audit_test.go`
* `tools/del-name-audit/integration_test.go`
* `tools/del-name-audit/main.go`
* `tools/del-name-audit/main_test.go`
* `tools/del-name-audit/store_mongo.go`
* `user-service/mongorepo/subscriptions_count_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -436,46 +446,58 @@ func activeSubscriptionFilter(account string) bson.M {
...
 		"open": bson.M{"$ne": false},
+		// Soft-deleted (^Del-) rooms, read from the subscription's OWN denormalized
+		// name rather than through a rooms $lookup. UpdateSubscriptionNamesForRoom
+		// keeps that copy in step (room-worker for in-stack renames, inbox-worker for
+		// oplog and federated ones), so the join bought nothing the sub doesn't already
+		// carry — see tools/del-name-audit for the data check behind this. A regex
+		// literal under $not also matches docs with no name at all, so a nameless sub
+		// still counts, as it did when the join left room.name null.
+		"name": bson.M{"$not": bson.Regex{Pattern: deletedRoomNameRegex}},
 		"$or": bson.A{
```
julianshen, line 457: _🎯 correctness_ | _🔴 Blocking_ | _`distributed-system`, `tgd-review` (2 rules)_

**DM and botDM deletions cannot be detected from subscription names**

The new predicate applies to every active room type, but the audit README and TestAudit_PerRoomTypeSplit explicitly establish that subscriptions.name is the counterpart account for dm and the bot name for botDM, not the room name. A DM subscription named "bob" therefore still passes this filter after its room is renamed to "Del-alice-bob", so CountActiveSubscriptions and GetActiveSubscriptions continue returning the deleted room; the previous lookup rejected it using rooms.name. Retain the authoritative room lookup for these types, or introduce a dedicated deletion marker with a reliable propagation contract before using CountDocuments.

<details>

- **`tgd-review`**:456: This predicate applies to channel, dm, and botDM rows, but the new audit README states that subscriptions.name is the counterpart account for dm and the bot name for botDM, not the room name. TestAudit_PerRoomTypeSplit likewise models a deleted DM whose subscriptions remain named alice/bob and correctly classifies that as unsafe. Such locally deleted DM/botDM rooms were excluded by the former room lookup but will now be counted and returned. Retain the joined room-name predicate for these types, or enforce and migrate a writer invariant that puts the deletion marker on every affected subscription before switching the read path.

</details>

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 457: The new predicate applies to every active room type, but the audit README and TestAudit_PerRoomTypeSplit explicitly establish that subscriptions.name is the counterpart account for dm and the bot name for botDM, not the room name. A DM subscription named "bob" therefore still passes this filter after its room is renamed to "Del-alice-bob", so CountActiveSubscriptions and GetActiveSubscriptions continue returning the deleted room; the previous lookup rejected it using rooms.name. Retain the authoritative room lookup for these types, or introduce a dedicated deletion marker with a reliable propagation contract before using CountDocuments.

```

</details>

tools/del-name-audit/audit.go
```diff
@@ -0,0 +1,179 @@
...
+	Unverifiable    int                   `json:"unverifiable"`
+	Samples         []mismatch            `json:"samples"`
+}
+
+// safeToMoveFilter reports whether the data supports replacing user-service's
+// rooms $lookup with a subscription-name filter: every subscription of a Del-
+// room must carry the prefix (no Disagree), and no Del--named subscription may
+// point at a live room (no FalsePositive). Unverifiable rows (no local room doc)
+// are cross-site and are excluded either way, so they do not block.
+func (r *report) safeToMoveFilter() bool {
```
julianshen, line 68: _🏷️ distributed-consistency_ | _🔴 Blocking_ | _`distributed-system`, `silent-failure`, `tgd-review` (3 rules)_

**Unverifiable cross-site rows are incorrectly accepted as safe**

safeToMoveFilter ignores Unverifiable, although a missing local room is precisely where the old and new predicates differ: roomsEnrichStages(true) keeps a missing join result, while the new subscription-name predicate drops a Del--named row. Thus the included cross-site fixture changes from counted to excluded while the audit reports SAFE, without checking the authoritative remote room. During delayed cross-site rename delivery, the marker may also be stale relative to that room. Resolve these rooms through their owning site, or make any Unverifiable row produce a non-safe/inconclusive verdict.

<details>

- **`tgd-review`**:68: A Del--named subscription with no local room is not equivalent under the old and new filters. roomsEnrichStages(true) explicitly keeps missing/cross-site rooms because the joined room.name is absent, while the new subscription-name predicate drops the row. If its remote room is live and the local Del- marker is stale, the rewrite is a false positive, but this audit increments Unverifiable and still returns SAFE. Treat any unverifiable row as inconclusive/unsafe unless the authoritative remote room is resolved, and update the conflicting README and tests.
- **`silent-failure`**:68: A `Del-`-named subscription with no local room increments `Unverifiable`, but this predicate ignores that count. The previous room lookup kept missing-room rows, while the new subscription-name filter drops them. Therefore a stale `Del-` marker for a live cross-site room can be excluded from active counts even though this audit exits successfully. Treat unverifiable rows as unsafe/inconclusive unless the authoritative remote room is resolved, and update the corresponding tests and documentation.

</details>

<details>

```
In `tools/del-name-audit/audit.go` around line 68: safeToMoveFilter ignores Unverifiable, although a missing local room is precisely where the old and new predicates differ: roomsEnrichStages(true) keeps a missing join result, while the new subscription-name predicate drops a Del--named row. Thus the included cross-site fixture changes from counted to excluded while the audit reports SAFE, without checking the authoritative remote room. During delayed cross-site rename delivery, the marker may also be stale relative to that room. Resolve these rooms through their owning site, or make any Unverifiable row produce a non-safe/inconclusive verdict.

```

</details>

tools/del-name-audit/store_mongo.go
```diff
@@ -0,0 +1,141 @@
...
+		subs:  db.Collection(subscriptionsCollection),
+	}
+}
+
+// delRooms returns rooms whose name carries the soft-delete marker, capped by
+// limit (0 = uncapped).
+func (s *mongoStore) delRooms(ctx context.Context, limit int) ([]roomRef, error) {
+	opt := options.Find().SetProjection(bson.M{"_id": 1, "name": 1, "type": 1})
+	if limit > 0 {
+		opt = opt.SetLimit(int64(limit))
```
julianshen, line 43: _🎯 correctness_ | _🔴 Blocking_ | _`distributed-system`, `mongodb` (2 rules)_

**Capped scans can still emit a SAFE verdict.**

`--rooms` and `--subs` apply MongoDB limits, but the final verdict does not record that data was omitted. For example, a mismatch outside the first 5,000 rooms makes `--rooms 5000` exit 0 and authorize the rewrite. Treat any capped run as non-authoritative, or query total counts and return SAFE only when the limits did not truncate either direction.

<details>

- **`distributed-system`**:118: roomLimit and subLimit silently make the safety checks incomplete, but the report has no incomplete state. With --rooms 5000, a stale subscription belonging to an omitted Del- room is never checked; with --subs, an omitted Del--named subscription over a live room is never checked. In either case safeToMoveFilter can return true and exit 0. Query one extra row or obtain totals to detect truncation, record it in the report, and never emit SAFE for a capped/incomplete scan.

</details>

<details>

```
In `tools/del-name-audit/store_mongo.go` around line 43: `--rooms` and `--subs` apply MongoDB limits, but the final verdict does not record that data was omitted. For example, a mismatch outside the first 5,000 rooms makes `--rooms 5000` exit 0 and authorize the rewrite. Treat any capped run as non-authoritative, or query total counts and return SAFE only when the limits did not truncate either direction.

```

</details>

julianshen (commented):
tGD inline review

tools/del-name-audit/audit.go
julianshen, line 68: _🎯 correctness_ | _🔴 Blocking_ | _`mongodb`_

**Missing local rooms are incorrectly treated as proof of safety.**

The old rooms lookup keeps a subscription when no local room document exists, while the new subscription-name filter drops it when its name starts with `Del-`. Therefore an unverifiable row may be a stale marker for a live cross-site room, and `Unverifiable > 0` cannot support an equivalence verdict. Make such runs inconclusive/unsafe unless the authoritative remote room is resolved, and update the corresponding documentation and tests.

<details>

```
In `tools/del-name-audit/audit.go` around line 68: The old rooms lookup keeps a subscription when no local room document exists, while the new subscription-name filter drops it when its name starts with `Del-`. Therefore an unverifiable row may be a stale marker for a live cross-site room, and `Unverifiable > 0` cannot support an equivalence verdict. Make such runs inconclusive/unsafe unless the authoritative remote room is resolved, and update the corresponding documentation and tests.

```

</details>

tools/del-name-audit/store_mongo.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+		for _, room := range batch {
+			out[room.ID] = room
+		}
+	}
+	return out, nil
+}
+
+// collect runs the four read phases and returns the audit input.
+func collect(ctx context.Context, s *mongoStore, roomLimit, subLimit, sampleCap int) (auditInput, error) {
```
julianshen, line 113: _🏷️ consistency_ | _🔴 Blocking_ | _`mongodb`_

**The audit reads no consistent MongoDB snapshot.**

The four sequential finds do not share a session or snapshot. A room can be live during `delRooms`, then be renamed to `Del-` while its subscription fan-out is delayed; `delNamedSubs` also misses the still-live subscription name, so the run reports SAFE despite a mismatch existing at completion. Reads may additionally land on different secondaries. Run against quiesced writers or perform the checks within one consistent MongoDB snapshot/single aggregation before issuing a SAFE verdict.

<details>

```
In `tools/del-name-audit/store_mongo.go` around line 113: The four sequential finds do not share a session or snapshot. A room can be live during `delRooms`, then be renamed to `Del-` while its subscription fan-out is delayed; `delNamedSubs` also misses the still-live subscription name, so the run reports SAFE despite a mismatch existing at completion. Reads may additionally land on different secondaries. Run against quiesced writers or perform the checks within one consistent MongoDB snapshot/single aggregation before issuing a SAFE verdict.

```

</details>

tools/del-name-audit/audit.go
```diff
@@ -0,0 +1,179 @@
...
+	ByRoomType      map[string]*typeStats `json:"byRoomType"`
+	DelNamedSubs    int                   `json:"delNamedSubs"`
+	FalsePositive   int                   `json:"falsePositive"`
+	Unverifiable    int                   `json:"unverifiable"`
+	Samples         []mismatch            `json:"samples"`
+}
+
+// safeToMoveFilter reports whether the data supports replacing user-service's
+// rooms $lookup with a subscription-name filter: every subscription of a Del-
+// room must carry the prefix (no Disagree), and no Del--named subscription may
```
julianshen, line 65: _🏷️ nats/cross-site-replication_ | _🔴 Blocking_ | _`nats`_

**Cross-site rename state is not authoritatively audited.**

The inbox-worker applies federated room renames asynchronously through the OUTBOX/NATS lane, but the audit cannot verify either direction for rooms without a local document. If a delete event is delayed during reconnect/retry, the remote subscription retains a live name; its room is absent from `delRooms` and its name is absent from `delNamedSubs`, so the audit reports SAFE while the new count filter keeps a deleted room. Conversely, a delayed restore leaves a `Del-` name that is classified as `Unverifiable` and also reported SAFE, causing a live room to be dropped. Resolve cross-site subscriptions against their authoritative home sites or a replicated event version/watermark before returning SAFE; merely requiring `Unverifiable == 0` fixes only the restore direction. Until then, retain the authoritative filtering path for cross-site rows.

<details>

```
In `tools/del-name-audit/audit.go` around line 65: The inbox-worker applies federated room renames asynchronously through the OUTBOX/NATS lane, but the audit cannot verify either direction for rooms without a local document. If a delete event is delayed during reconnect/retry, the remote subscription retains a live name; its room is absent from `delRooms` and its name is absent from `delNamedSubs`, so the audit reports SAFE while the new count filter keeps a deleted room. Conversely, a delayed restore leaves a `Del-` name that is classified as `Unverifiable` and also reported SAFE, causing a live room to be dropped. Resolve cross-site subscriptions against their authoritative home sites or a replicated event version/watermark before returning SAFE; merely requiring `Unverifiable == 0` fixes only the restore direction. Until then, retain the authoritative filtering path for cross-site rows.

```

</details>

tools/del-name-audit/main.go
```diff
@@ -0,0 +1,200 @@
...
+		slog.Error("collect audit input", "error", err)
+		return 2
+	}
+	rep := audit(&in)
+
+	if err := render(os.Stdout, &rep, f.asJSON); err != nil {
+		slog.Error("render report", "error", err)
+		return 2
+	}
+	if !rep.safeToMoveFilter() {
```
julianshen, line 113: _🏷️ silent-failure_ | _🔴 Blocking_ | _`silent-failure`, `tgd-review` (2 rules)_

**A capped partial scan can exit successfully as SAFE.**

When `--rooms` or `--subs` is positive, the queries intentionally inspect only a subset, but the verdict considers only mismatches found in that subset. A clean capped scan therefore reaches the unconditional SAFE message and exit code 0 even when omitted rows contain disagreements or false positives. This is reachable through the documented `--rooms 5000 --json` example. Reject nonzero limits for a conclusive verdict, or represent the result as partial/inconclusive and return a nonzero exit code; reserve exit 0 for uncapped scans.

<details>

- **`tgd-review`**:106: The --rooms and --subs limits are passed directly to MongoDB, but the verdict only checks mismatches among returned documents. A capped arbitrary subset can contain no mismatch while omitted rooms or subscriptions violate the invariant, causing exit 0 and a SAFE report. Fetch one extra row to detect truncation and report an inconclusive result, or never emit SAFE when a nonzero cap may have truncated either direction.

</details>

<details>

```
In `tools/del-name-audit/main.go` around line 113: When `--rooms` or `--subs` is positive, the queries intentionally inspect only a subset, but the verdict considers only mismatches found in that subset. A clean capped scan therefore reaches the unconditional SAFE message and exit code 0 even when omitted rows contain disagreements or false positives. This is reachable through the documented `--rooms 5000 --json` example. Reject nonzero limits for a conclusive verdict, or represent the result as partial/inconclusive and return a nonzero exit code; reserve exit 0 for uncapped scans.

```

</details>

tools/del-name-audit/store_mongo.go
```diff
@@ -0,0 +1,141 @@
...
+	rooms, err := s.delRooms(ctx, roomLimit)
+	if err != nil {
+		return auditInput{}, fmt.Errorf("collect Del- rooms: %w", err)
+	}
```
julianshen, line 117: _🏷️ consistency_ | _🟠 Warning_ | _`distributed-system`_

**The multi-query audit can miss a concurrent partial rename**

collect performs four independent reads with no common snapshot. For example, delRooms can read before room R is renamed to Del-R; the rename can then commit while its subscription fan-out is delayed; subsForRooms never examines R because it was absent from the first result, and delNamedSubs does not find the still-live subscription name. The audit consequently returns SAFE even though the inconsistency exists when it finishes. Run all phases against one snapshot transaction, require an immutable restored dump or quiesced writers, or otherwise detect concurrent mutations and return inconclusive.

<details>

```
In `tools/del-name-audit/store_mongo.go` around line 117: collect performs four independent reads with no common snapshot. For example, delRooms can read before room R is renamed to Del-R; the rename can then commit while its subscription fan-out is delayed; subsForRooms never examines R because it was absent from the first result, and delNamedSubs does not find the still-live subscription name. The audit consequently returns SAFE even though the inconsistency exists when it finishes. Run all phases against one snapshot transaction, require an immutable restored dump or quiesced writers, or otherwise detect concurrent mutations and return inconclusive.

```

</details>

docs/client-api.md
```diff
@@ -5188,7 +5188,8 @@ Results are **paginated** by `offset`/`limit` (offset-based): the server returns
...
+  - **List paths** (`subscription.list`, `subscription.getChannels`): the subscription is **dropped**. Local rooms are filtered in the Mongo query; cross-site rooms are dropped after the per-site `GetRoomsInfo` lookup reveals the `Del-` name — this happens post-pagination, so a page can be shorter than `limit` (`hasMore` is computed from the database page, before the cross-site drop).
+  - **`subscription.count`**: the subscription is **dropped**, local and cross-site alike, in the Mongo query — the filter reads the `Del-` prefix from the subscription's own denormalized room name, which a rename fans out to every subscriber (including across sites). No `GetRoomsInfo` lookup and no pagination are involved, so unlike the list paths the count needs no post-filtering pass.
   - **Single-item lookups** (`subscription.getDM`, `subscription.getByRoomID`): the subscription is **kept with no `room` object** — the row is returned so the caller knows the subscription exists, but the deleted room is omitted.
```
julianshen, line 5193: _🏷️ distributed-consistency_ | _🟠 Warning_ | _`distributed-system`_

**Cross-site deletion filtering is only eventually consistent**

The count now depends solely on a federated subscription-name update. If the owning site commits a soft deletion while the OUTBOX event is delayed by a broker or network partition, the remote site's subscription retains its live name and CountActiveSubscriptions continues counting the deleted room until delivery. The documentation currently states that local and cross-site rows are dropped alike without describing this consistency window. Either document the eventual-consistency contract and expected recovery behavior, or consult authoritative remote state when immediate filtering is required.

<details>

```
In `docs/client-api.md` around line 5193: The count now depends solely on a federated subscription-name update. If the owning site commits a soft deletion while the OUTBOX event is delayed by a broker or network partition, the remote site's subscription retains its live name and CountActiveSubscriptions continues counting the deleted room until delivery. The documentation currently states that local and cross-site rows are dropped alike without describing this consistency window. Either document the eventual-consistency contract and expected recovery behavior, or consult authoritative remote state when immediate filtering is required.

```

</details>

user-service/mongorepo/subscriptions.go
julianshen, line 457: _🏷️ atomicity_ | _🟠 Warning_ | _`mongodb`_

**The count trusts a non-atomic denormalized rename.**

The count and active-list match now treat `subscriptions.name` as authoritative, although room renames fan out across many subscription documents and cross-site updates pass through outbox/inbox workers. During delayed or failed fan-out, a deleted room with an old subscription name is counted; after a restore, a stale `Del-` name is omitted. A one-time audit cannot enforce this future multi-document invariant. Retain the authoritative local-room filter, or establish and migrate a durable deletion field with idempotent writer/reconciliation guarantees and explicitly define the cross-site consistency window.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 457: The count and active-list match now treat `subscriptions.name` as authoritative, although room renames fan out across many subscription documents and cross-site updates pass through outbox/inbox workers. During delayed or failed fan-out, a deleted room with an old subscription name is counted; after a restore, a stale `Del-` name is omitted. A one-time audit cannot enforce this future multi-document invariant. Retain the authoritative local-room filter, or establish and migrate a durable deletion field with idempotent writer/reconciliation guarantees and explicitly define the cross-site consistency window.

```

</details>

tools/del-name-audit/store_mongo.go
```diff
@@ -0,0 +1,141 @@
...
+// limit (0 = uncapped).
+func (s *mongoStore) delRooms(ctx context.Context, limit int) ([]roomRef, error) {
+	opt := options.Find().SetProjection(bson.M{"_id": 1, "name": 1, "type": 1})
+	if limit > 0 {
+		opt = opt.SetLimit(int64(limit))
+	}
+	cur, err := s.rooms.Find(ctx, bson.M{"name": delNameRegex}, opt)
+	if err != nil {
+		return nil, fmt.Errorf("find Del- rooms: %w", err)
+	}
```
julianshen, line 48: _⚡ memory-exhaustion_ | _🟠 Warning_ | _`mongodb`, `oom` (2 rules)_

**The default full audit retains the entire matching dataset in memory.**

Running the documented default command on a large deployment passes zero limits, so `delRooms` and `findSubs` decode complete result sets with `cur.All`. `subsForRooms` only chunks the `$in` query and then appends every returned subscription to one slice; a 500-room batch can itself have unbounded subscriber fan-out. `collect` simultaneously retains all deleted rooms, all subscriptions for those rooms, all `Del-`-named subscriptions, distinct-ID copies, and the resolved-room map, while `audit` allocates additional maps over them. Production-scale or user-amplified room/subscription cardinality can therefore cause severe GC thrashing or OOM the audit process/container. Stream Mongo cursors and update report counters incrementally, processing room IDs and reverse lookups in bounded batches and releasing each batch before fetching the next; also enforce a hard maximum or require explicit limits as a guardrail.

<details>

- **`mongodb`**:50: The default uncapped audit uses `Cursor.All` for rooms and subscriptions and then appends every batch into process-wide slices. The 500-ID chunks only bound each `$in` document; they do not bound results, so a deployment with many deleted rooms or subscribers can exhaust memory before the timeout. Stream and classify cursor results incrementally, or aggregate counts server-side while retaining only the bounded samples.

</details>

<details>

```
In `tools/del-name-audit/store_mongo.go` around line 48: Running the documented default command on a large deployment passes zero limits, so `delRooms` and `findSubs` decode complete result sets with `cur.All`. `subsForRooms` only chunks the `$in` query and then appends every returned subscription to one slice; a 500-room batch can itself have unbounded subscriber fan-out. `collect` simultaneously retains all deleted rooms, all subscriptions for those rooms, all `Del-`-named subscriptions, distinct-ID copies, and the resolved-room map, while `audit` allocates additional maps over them. Production-scale or user-amplified room/subscription cardinality can therefore cause severe GC thrashing or OOM the audit process/container. Stream Mongo cursors and update report counters incrementally, processing room IDs and reverse lookups in bounded batches and releasing each batch before fetching the next; also enforce a hard maximum or require explicit limits as a guardrail.

```

</details>

tools/del-name-audit/main.go
```diff
@@ -0,0 +1,200 @@
...
+	cfg, err := parseConfig(envMap())
+	if err != nil {
+		slog.Error("invalid configuration", "error", err)
+		return 2
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), cfg.Timeout)
+	defer cancel()
+
+	// ConnectRead: secondaryPreferred. The audit is a bulk scan and must not add
```
julianshen, line 87: _🏷️ operations_ | _🟠 Warning_ | _`mongodb`_

**secondaryPreferred can move the bulk scan onto the primary.**

The surrounding contract says this bulk audit must not load the primary, but MongoDB's `secondaryPreferred` mode falls back to the primary when no eligible secondary is available. An uncapped scan can therefore hit the primary during replica-set degradation. Use strict `secondary` read preference when primary avoidance is required, or correct the operational contract and require an explicitly targeted secondary/restored dump.

<details>

```
In `tools/del-name-audit/main.go` around line 87: The surrounding contract says this bulk audit must not load the primary, but MongoDB's `secondaryPreferred` mode falls back to the primary when no eligible secondary is available. An uncapped scan can therefore hit the primary during replica-set degradation. Use strict `secondary` read preference when primary avoidance is required, or correct the operational contract and require an explicitly targeted secondary/restored dump.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -436,46 +446,58 @@ func activeSubscriptionFilter(account string) bson.M {
...
 			bson.M{"roomType": bson.M{"$in": bson.A{"dm", "channel"}}},
 			bson.M{"roomType": "botDM", "isSubscribed": true},
 		}}
 }
```
julianshen, line 462: _🎯 concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Counts race with asynchronous subscription-name fan-out.**

The room rename and the denormalized-name propagation are not observed atomically, especially for the documented inbox/outbox cross-site path. A count can run after a room becomes `Del-` but before this subscriber's `name` is updated, so the new predicate still includes the deleted room; restoring a room creates the inverse undercount window. Delayed or failed fan-out can extend this beyond a transient discrepancy. Retain the authoritative room filter until propagation has a durable, monitored consistency guarantee, or explicitly provide atomic/ordered visibility before relying solely on the subscription name.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 462: The room rename and the denormalized-name propagation are not observed atomically, especially for the documented inbox/outbox cross-site path. A count can run after a room becomes `Del-` but before this subscriber's `name` is updated, so the new predicate still includes the deleted room; restoring a room creates the inverse undercount window. Delayed or failed fan-out can extend this beyond a transient discrepancy. Retain the authoritative room filter until propagation has a durable, monitored consistency guarantee, or explicitly provide atomic/ordered visibility before relying solely on the subscription name.

```

</details>

tools/del-name-audit/store_mongo.go
```diff
@@ -0,0 +1,141 @@
...
+
+// subsForRooms returns every subscription pointing at one of roomIDs.
+func (s *mongoStore) subsForRooms(ctx context.Context, roomIDs []string) ([]subRef, error) {
+	var out []subRef
+	for _, ids := range chunk(roomIDs, inChunkSize) {
+		batch, err := s.findSubs(ctx, bson.M{"roomId": bson.M{"$in": ids}}, 0)
+		if err != nil {
+			return nil, err
+		}
+		out = append(out, batch...)
```
julianshen, line 64: _🏷️ resource-exhaustion_ | _🟠 Warning_ | _`performance-and-concurrency`_

**The default full audit retains every matching document.**

With the default zero limits, each `findSubs` call materializes its cursor with `cur.All`, and this append retains every subscription across all room batches. `collect` then simultaneously holds all deleted rooms, all subscriptions for those rooms, all `Del-`-named subscriptions, and the resolved-room map. One deleted high-membership room can therefore consume unbounded memory despite the 500-ID query chunks; the room cap does not cap its subscribers. On a large deployment this can OOM the audit process and sustain load on the target secondary without producing a verdict. Stream cursors and update report counters incrementally, resolving the reverse scan in bounded batches so memory remains proportional to batch and sample sizes.

<details>

```
In `tools/del-name-audit/store_mongo.go` around line 64: With the default zero limits, each `findSubs` call materializes its cursor with `cur.All`, and this append retains every subscription across all room batches. `collect` then simultaneously holds all deleted rooms, all subscriptions for those rooms, all `Del-`-named subscriptions, and the resolved-room map. One deleted high-membership room can therefore consume unbounded memory despite the 500-ID query chunks; the room cap does not cap its subscribers. On a large deployment this can OOM the audit process and sustain load on the target secondary without producing a verdict. Stream cursors and update report counters incrementally, resolving the reverse scan in bounded batches so memory remains proportional to batch and sample sizes.

```

</details>

tools/del-name-audit/store_mongo.go
julianshen, line 113: _🎯 concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**The audit can report SAFE from inconsistent read phases.**

The four dependent queries run sequentially without a common snapshot, and `secondaryPreferred` can also expose different replication positions. For example, room X can be stale at phase one and fixed before its subscriptions are read, while room Y is renamed after phase one and its delayed fan-out leaves an unprefixed subscription; X appears agreeing, Y is absent from both scans, and the audit reports SAFE although at least one mismatch existed throughout. A false SAFE verdict can authorize the production rewrite against unsafe data. Run all phases at one Mongo snapshot/cluster time, or require and enforce execution against an immutable restored snapshot.

<details>

```
In `tools/del-name-audit/store_mongo.go` around line 113: The four dependent queries run sequentially without a common snapshot, and `secondaryPreferred` can also expose different replication positions. For example, room X can be stale at phase one and fixed before its subscriptions are read, while room Y is renamed after phase one and its delayed fan-out leaves an unprefixed subscription; X appears agreeing, Y is absent from both scans, and the audit reports SAFE although at least one mismatch existed throughout. A false SAFE verdict can authorize the production rewrite against unsafe data. Run all phases at one Mongo snapshot/cluster time, or require and enforce execution against an immutable restored snapshot.

```

</details>

Makefile
```diff
@@ -516,3 +516,13 @@ seed-reset:
 
 seed-dry-run:
 	go run ./tools/seed-sample-data --dry-run
+
+# --- Del- name audit ----------------------------------------------------------
+# Read-only probe: does every subscription of a soft-deleted ("Del-") room carry
+# the marker on its own denormalized name? Answers whether user-service can drop
```
julianshen, line 522: _📝 documentation_ | _🔵 Suggestion_ | _`tgd-review`_

**The audit overstates which rooms lookups can be removed**

The audit only validates moving the deleted-name predicate onto subscriptions. List paths still require the rooms lookup for enrichment and lastMsgAt, and this diff retains that lookup. Limit this claim to CountActiveSubscriptions and the active-path predicate change; make the same clarification in tools/del-name-audit/README.md and main.go.

<details>

```
In `Makefile` around line 522: The audit only validates moving the deleted-name predicate onto subscriptions. List paths still require the rooms lookup for enrichment and lastMsgAt, and this diff retains that lookup. Limit this claim to CountActiveSubscriptions and the active-path predicate change; make the same clarification in tools/del-name-audit/README.md and main.go.

```

</details>

tools/del-name-audit/store_mongo.go
```diff
@@ -0,0 +1,141 @@
...
+	opt := options.Find().SetProjection(bson.M{"_id": 1, "name": 1, "type": 1})
+	if limit > 0 {
+		opt = opt.SetLimit(int64(limit))
+	}
+	cur, err := s.rooms.Find(ctx, bson.M{"name": delNameRegex}, opt)
+	if err != nil {
+		return nil, fmt.Errorf("find Del- rooms: %w", err)
+	}
+	var out []roomRef
+	if err := cur.All(ctx, &out); err != nil {
```
julianshen, line 50: _🏷️ resource-exhaustion_ | _🟠 Warning_ | _`tgd-review`_

**The uncapped audit materializes the full dataset**

The default zero limits call `Cursor.All`, and `collect` simultaneously retains all deleted rooms, all subscriptions for those rooms, all `Del-`-named subscriptions, distinct-ID copies, and the resolved-room map. The 500-ID chunks only bound query documents; one high-membership room can still produce an unbounded result and exhaust process memory. Stream and classify cursors incrementally or aggregate counts server-side while retaining only bounded samples.

<details>

```
In `tools/del-name-audit/store_mongo.go` around line 50: The default zero limits call `Cursor.All`, and `collect` simultaneously retains all deleted rooms, all subscriptions for those rooms, all `Del-`-named subscriptions, distinct-ID copies, and the resolved-room map. The 500-ID chunks only bound query documents; one high-membership room can still produce an unbounded result and exhaust process memory. Stream and classify cursors incrementally or aggregate counts server-side while retaining only bounded samples.

```

</details>

user-service/service/service.go
```diff
@@ -22,10 +22,10 @@ type SubscriptionRepository interface {
 	GetDMSubscription(ctx context.Context, account, target string) (*model.EnrichedDMSubscription, error)
 	GetSubscriptionByRoomID(ctx context.Context, account, roomID string) (*model.EnrichedSubscription, error)
 	CountActiveSubscriptions(ctx context.Context, account string) (int, error)
-	// GetActiveSubscriptions returns up to limit active subscriptions. The cap
-	// is applied before the deleted-room filter, so a page whose capped slice
-	// contains soft-deleted rooms comes back slightly short of limit — fine for
-	// the unread count (its only consumer), not a general pagination surface.
+	// GetActiveSubscriptions returns up to limit active subscriptions. The
```
julianshen, line 25: _📝 documentation_ | _🔵 Suggestion_ | _`tgd-review`_

**The interface promises an exact-size result incorrectly**

Filtering before the cap prevents downstream deletion filtering from shortening a capped set, but the method still returns fewer than `limit` when fewer active subscriptions exist, and a non-positive limit is uncapped. Document those actual semantics rather than promising an exact-size page.

<details>

```
In `user-service/service/service.go` around line 25: Filtering before the cap prevents downstream deletion filtering from shortening a capped set, but the method still returns fewer than `limit` when fewer active subscriptions exist, and a non-positive limit is uncapped. Document those actual semantics rather than promising an exact-size page.

```

</details>

julianshen (commented):
tGD inline review

hmchangw:
Closing in favour of #353. Both PRs solved the same problem — `CountActiveSubscriptions` doing an indexed `rooms` lookup per active subscription to read one field — and the review here established that this one solves it on a foundation that does not hold.

Thanks for the detailed review; it changed the outcome. Two of the findings are things this PR got factually wrong, and I verified both against `main` rather than taking them on trust:

**The audit tool's verdict logic is unsound.** The old filter is `$match {room.name: {$not: /^Del-/}}` applied *after* `$unwind` with `preserveNullAndEmptyArrays` — a subscription whose room has no local document has no `room.name`, so `$not` matches and the row is **kept**. The proposed subscription-name filter **drops** that same row when the name carries the marker. An `Unverifiable` row is therefore precisely where the two predicates diverge, not a case where they agree. `safeToMoveFilter` ignored `Unverifiable` entirely, and `tools/del-name-audit/README.md` asserted that cross-site rows "are excluded under both the join and the name filter", which is simply not true of the join. The tool built to gate this PR could return SAFE for data where the rewrite changes behaviour. Same class of hole for capped runs: `--rooms N` truncates the scan and the verdict never records it.

**DM and botDM deletions cannot be detected from `sub.name`.** The filter applies to every active room type, but the denormalised name is the room name only for channels — for `dm` it is the counterpart account and for `botDM` the bot name. The audit's own per-`roomType` split exists to surface exactly this, and it argues against the filter it was meant to validate.

Behind both is the point raised on the consistency threads: `subscriptions.name` is maintained by an asynchronous multi-document fan-out (room-worker locally, inbox-worker for oplog and federated renames). Treating it as authoritative for a read-path predicate means a delayed or failed fan-out is a wrong badge count, and a one-time audit cannot enforce a continuing invariant. That is a real cost this PR was buying with no corresponding benefit over the alternative.

#353 reaches the same `CountDocuments` without any of it: it removes the name-based filtering rather than relocating it, so nothing depends on the propagation contract. #354 closes the migration path that replicated `Del-` names from the legacy source, which makes #353's premise substantially true — with the caveat, noted there, that `room-service`'s rename RPC still validates only length and can accept a `Del-` prefixed name.

The rollout and consistency concerns raised here are not fully answered by #353 either — the mixed-version window is tracked on that PR and is still open.

The `tools/del-name-audit` probe on this branch is not carried over. If the denormalisation question is revisited, the tool is worth resurrecting from this branch's history with the `Unverifiable`/capped-scan verdict fixed first.

---

PR #353, opened 2026-08-22
perf(user-service): drop the Del- soft-delete filtering entirely

hmchangw:
Supersedes #352, which took the other approach (relocating the `Del-` filter onto the subscription's denormalized name) and was closed after review — see [the closing note](https://github.com/hmchangw/newchat/pull/352#issuecomment-5384570671) for why that foundation did not hold.

## The problem

`CountActiveSubscriptions` ran an aggregation whose only reason to touch `rooms` was the soft-delete (`^Del-`) filter on `room.name`. For that one field it:

- did an indexed `rooms` lookup **per active subscription** — unbounded, unlike the unread path which caps at `MAX_SUBSCRIPTION_LIMIT` *before* joining;
- transported **eleven** room fields per row, including the E2E key blob (`encKey.priv`);
- computed eleven `$addFields` (the `__sortKey` among them) that `$count` then discarded.

## The approach

Remove the name-based filtering rather than relocate it. Nothing then depends on a propagation contract — which is the distinction from #352, where the read path would have leaned on an asynchronous multi-document fan-out staying in step.

**1. The count.** Every predicate becomes local to the subscription, so it is a plain `CountDocuments`. `GetActiveSubscriptions` drops the same filter so the unread count keeps selecting the same rows as the total.

**2. The rest of it.** The first commit alone would leave the invariant half-applied — the list paths would still run the regex, and a client folding its badge from the list would disagree with the server if such a room appeared. So:

- `roomsEnrichStages` loses its `dropDeleted` parameter and the regex `$match`; the rooms join is purely additive on every path that uses it.
- `roomMatchStages` no longer constrains the room name inside its `$lookup`. It still drops subs whose room has no local doc — member matching is inherently local, and that has nothing to do with the name.
- `enrichWithRoomInfoAndLastMsg` loses its `dropDeleted` parameter and the post-enrichment removal pass; `removeIndices` went with it. `enrichCrossSite` and `applyRoomInfo` no longer report rows to drop, so both lose their return values.
- `buildLocalRoom` no longer suppresses the room object for a `Del-` name.
- The unread count excludes a cross-site room only when the remote site reports it not-found, never by its name.

**3. Cleanup pass.** The active set had grown two representations — a flat filter for the count, two pipeline stages for the unread path — which are contractually required to select identical rows. Collapsed to one `activeFilter`; `originFilterStage` folded into `originFilter`; dead scaffolding removed (`SubscriptionRepo.siteID`, unreachable nil-guards, the vestigial return value).

What still removes or blanks a row is unchanged, and is about a room being **absent**, not about what it is called.

## Dependency — satisfied

**#354 has merged** (`bd6f596`), and `main` is merged into this branch. It stops `oplog-collections-transformer` replicating `Del-` names from the legacy source, which is what makes this PR's premise true.

One caveat neither PR closes: `room-service`'s rename RPC (`handler.go:1960-1963`) validates only trim/non-empty/≤100 runes, so a channel owner can still set a `Del-` prefixed name through the live app. After #354 such a name is a name someone typed, never a deletion — precisely the case this PR handles correctly and #352 would have mishandled.

## Rollout — verified clear

The mixed-version concern raised in review (old replicas suppress `Del-` rooms while new ones show them) is **vacuous**: a data check by the repository owner confirms **no `Del-` prefixed rows exist in the destination databases**. The predicate this PR removes matches nothing, so no request can observe a difference and rollout order does not matter.

Worth re-running `countDocuments({name: /^Del-/})` on `rooms` and `subscriptions` per site immediately before rollout, since that check is a point-in-time snapshot.

## Behavior change

A room named `Del-anything` is now returned like any other, on every endpoint. Documented in `docs/client-api.md`.

## Open review thread

- 🟠 **Missing local rooms yield a fabricated empty room object** — verified as **pre-existing**: on `main`, `buildLocalRoom` returned nil only for the `Del-` prefix, so an absent local room already produced a non-nil room with empty metadata. Not a regression here. The underlying contract point is fair — an unmatched lookup should leave `Room` nil — but it needs a match marker carried out of `roomsEnrichStages` and is a client-visible change, so it is tracked as a follow-up rather than folded in.

## Indexes

Unchanged. `{u.account, roomType}` still serves the leading equality, and a wider index would trade write amplification on a hot collection for a read this change has already made cheap.

## Testing

CI green on the merged base: lint, test, sast, and **`test-integration (user-service)`** — the integration suite ran against a real MongoDB, covering the rewritten fixtures, the changed counts, the `NewSubscriptionRepo` signature change, and the member-matching and `getByRoomID` assertions.

- Coverage on the changed service functions: `enrichWithRoomInfoAndLastMsg`, `applyRoomInfo`, `buildLocalRoom`, `activeFilter`, `originFilter` at 100%; `enrichCrossSite` 97.2%.
- The service-layer work was Red-first. Two table cases (`hasUnread`/`hasGroupMention` over a `Del-Eng` room) flipped from `false` to `true` — the honest consequence of the room object now being built, so the timestamps decide.

## Summary by CodeRabbit

* **Bug Fixes**
  * Subscription lists and counts no longer exclude rooms based on `Del-` name prefixes.
  * Subscription pages now remain filled to the requested limit.
  * Unresolved or cross-site rooms are retained without room details where appropriate.
  * Unread and mention indicators now reflect room availability and timestamps consistently.
  * Active subscription counts and results now apply consistent state and room-type filtering.

* **Documentation**
  * Clarified that active subscription results contain the requested number of entries.


julianshen:
## Needs clarification

Is rollout gated on purging every legacy \`Del-\*\` room and disabling all old room-service writers?
Reply in this thread, or post: `answer clar_a7yo7nelednqwuyzh5ebcd7uur: <your answer>`

julianshen:
**2 findings · 2 inline comments posted** — 🔴 1 blocking · 🔵 1 suggestion

### Needs clarification

Will all room-service producers and existing `Del-` tombstones be migrated before this user-service version is deployed?

`answer clar_t6yqlmflovm7js7zouec5etu4l: <your answer>`

[Open the question](https://github.com/hmchangw/newchat/pull/353#issuecomment-5384903277)

### Related work

- [PR #352](https://github.com/hmchangw/newchat/pull/352) — perf\(user-service\): count active subscriptions without the rooms join \(option A: move the Del- check onto the subscription\) (closed)
- [PR #354](https://github.com/hmchangw/newchat/pull/354) — Skip soft-deleted rooms and subscriptions in CDC migration (merged)

### Local review memory

- `docs/client-api.md:5525` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟡 Minor_ _⚡ Quick win_ **Document the open:false exclusion.** activeSubscriptionFilter excludes subscriptions where open is false . The active-set definition lists only mute state, room type,… ([thread](https://github.com/hmchangw/newchat/pull/353#discussion_r3837667771))

<details>
<summary>📒 Files reviewed (12)</summary>

* `docs/client-api.md`
* `user-service/main.go`
* `user-service/mongorepo/subscriptions_count_test.go`
* `user-service/mongorepo/subscriptions_origin_test.go`
* `user-service/service/enrich_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

user-service/service/subscriptions.go
```diff
@@ -421,17 +377,11 @@ func buildLocalRoom(sub *model.EnrichedSubscription) *model.SubscriptionRoom {
...
 // are authoritative subscription state; room-service only supplies room data.
 //
-// Returns true when the cross-site room is soft-deleted ("Del-" name prefix),
-// signalling the caller to DROP the subscription entirely — the same exclusion the
-// Mongo query applies to locally-deleted rooms. A not-found or degraded room
-// returns false and is kept with no room object.
-func applyRoomInfo(sub *model.Subscription, info *model.RoomInfo) bool {
+// A not-found or degraded room is left with no room object; the subscription is
+// kept either way.
+func applyRoomInfo(sub *model.Subscription, info *model.RoomInfo) {
```
julianshen, line 379: _🏷️ rolling-deploy_ | _🔴 Blocking_ | _`distributed-system`_

**Legacy soft-deletes reappear during mixed-version rollout**

The removed code explicitly identifies `Del-` renames as room-service's legacy soft-delete protocol, but this PR changes only user-service consumers and provides no producer cutoff or data migration. Concrete timeline: an old or unchanged room-service replica renames a room to `Del-X`; the write succeeds; a new user-service replica receives `Found:true, Name:"Del-X"` and `applyRoomInfo` now exposes it as an ordinary room, while an old user-service replica still suppresses it. Counts, unread badges, and lists can therefore resurrect deleted rooms and vary by serving replica. Keep backward-compatible suppression until every producer is incapable of writing the legacy representation and existing rows are migrated, or introduce an explicit tombstone field understood by both versions and gate rollout on it.

<details>

```
In `user-service/service/subscriptions.go` around line 382: The removed code explicitly identifies `Del-` renames as room-service's legacy soft-delete protocol, but this PR changes only user-service consumers and provides no producer cutoff or data migration. Concrete timeline: an old or unchanged room-service replica renames a room to `Del-X`; the write succeeds; a new user-service replica receives `Found:true, Name:"Del-X"` and `applyRoomInfo` now exposes it as an ordinary room, while an old user-service replica still suppresses it. Counts, unread badges, and lists can therefore resurrect deleted rooms and vary by serving replica. Keep backward-compatible suppression until every producer is incapable of writing the legacy representation and existing rows are migrated, or introduce an explicit tombstone field understood by both versions and gate rollout on it.

```

</details>

user-service/service/subscriptions.go
```diff
@@ -392,10 +352,6 @@ const roomKeySecretLen = 32
 // the wire room object both carry *time.Time, so LastMsgAt/LastMentionAllAt pass
 // through unconverted.
 func buildLocalRoom(sub *model.EnrichedSubscription) *model.SubscriptionRoom {
-	// A soft-deleted room (name "Del-...") is surfaced with no room object.
-	if strings.HasPrefix(sub.RoomName, deletedRoomNamePrefix) {
-		return nil
-	}
 	room := &model.SubscriptionRoom{
```
julianshen, line 352: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**Missing local rooms still produce fabricated room objects**

The lookup preserves subscriptions with no matching room and leaves RoomName at its zero value, but enrichLocal still calls buildLocalRoom for every subscription whose SiteID is local. Since buildLocalRoom now unconditionally constructs a SubscriptionRoom, an unresolved local room is returned as a non-nil room with empty/default metadata, contradicting the new documented behavior that unresolved rooms have no room object. Preserve a lookup-match marker and skip local room construction when no room document matched; add coverage for a missing local room through the service layer.

<details>

```
In `user-service/service/subscriptions.go` around line 355: The lookup preserves subscriptions with no matching room and leaves RoomName at its zero value, but enrichLocal still calls buildLocalRoom for every subscription whose SiteID is local. Since buildLocalRoom now unconditionally constructs a SubscriptionRoom, an unresolved local room is returned as a non-nil room with empty/default metadata, contradicting the new documented behavior that unresolved rooms have no room object. Preserve a lookup-match marker and skip local room construction when no room document matched; add coverage for a missing local room through the service layer.

```

</details>

docs/client-api.md
```diff
@@ -5187,8 +5187,7 @@ Results are **paginated** by `offset`/`limit` (offset-based): the server returns
...
-  - **List paths** (`subscription.list`, `subscription.getChannels`) and `subscription.count`: the subscription is **dropped**. Local rooms are filtered in the Mongo query; cross-site rooms are dropped after the per-site `GetRoomsInfo` lookup reveals the `Del-` name — this happens post-pagination, so a page can be shorter than `limit` (`hasMore` is computed from the database page, before the cross-site drop).
+- **Room names carry no special meaning.** No endpoint filters, drops, or suppresses a subscription based on its room's name — the `Del-` prefix that legacy soft-deletes used is not interpreted anywhere. The deployment guarantees no soft-delete rename reaches the database, so a room named `Del-anything` is returned like any other room. Enrichment is purely additive: every subscription matched by a query is returned, and a room the server cannot resolve (not found, or a degraded cross-site RPC) simply arrives with **no `room` object** rather than removing the row. One consequence: a `subscription.list` page is no longer shortened after pagination, so a full page is always exactly `limit` rows.
```
julianshen: _📝 documentation_ | _🟠 Warning_ | _`tgd-review`_

**The deleted-room documentation is internally contradictory**

This paragraph says the Del- prefix is never interpreted and such rooms are returned normally, but the immediately following nested bullet still says single-item lookups omit the room object for a deleted room. The implementation and updated tests now enrich Del-prefixed rooms like any other room, so remove or rewrite that stale single-item bullet.

<details>

```
In `docs/client-api.md` around line 5190: This paragraph says the Del- prefix is never interpreted and such rooms are returned normally, but the immediately following nested bullet still says single-item lookups omit the room object for a deleted room. The implementation and updated tests now enrich Del-prefixed rooms like any other room, so remove or rewrite that stale single-item bullet.

```

</details>

julianshen (commented):
tGD inline review

julianshen:
## Needs clarification

Have all legacy \`Del-\` writers been disabled and existing tombstones migrated at every site?
Reply in this thread, or post: `answer clar_wzgqbmwygd6oqlmb6mdnvmiklt: <your answer>`

user-service/service/subscriptions.go
```diff
@@ -203,56 +190,30 @@ func (s *UserService) enrichWithRoomInfoAndLastMsg(c *natsrouter.Context, subs [
...
-		kept = append(kept, subs[i])
-	}
-	return kept
 }
 
 // enrichLocal builds sub.Room for LOCAL subs entirely from the $lookup baseline —
 // room metadata plus the E2E key projected from the room's encKey sub-document —
 // so it needs no separate key store read.
 func (s *UserService) enrichLocal(subs []model.EnrichedSubscription, localIdx []int) {
 	for _, j := range localIdx {
```
julianshen, line 203: _🏷️ masked-failure_ | _🟠 Warning_ | _`silent-failure`_

**Missing local rooms are masked by fabricated room objects**

When the Mongo lookup finds no local room, the subscription is preserved with zero-value enrichment fields. `enrichLocal` now unconditionally calls `buildLocalRoom`, which always returns a non-nil room, converting that unresolved lookup into apparent success with empty/default metadata. Clients therefore cannot distinguish a missing room from a valid but blank room. Preserve a lookup-match marker and leave `Room` nil when no local room document matched; add service-level coverage for a missing local room.

<details>

```
In `user-service/service/subscriptions.go` around line 203: When the Mongo lookup finds no local room, the subscription is preserved with zero-value enrichment fields. `enrichLocal` now unconditionally calls `buildLocalRoom`, which always returns a non-nil room, converting that unresolved lookup into apparent success with empty/default metadata. Clients therefore cannot distinguish a missing room from a valid but blank room. Preserve a lookup-match marker and leave `Room` nil when no local room document matched; add service-level coverage for a missing local room.

```

</details>

julianshen (commented):
tGD inline review

julianshen:
## Needs clarification

Will all room-service producers and existing \`Del-\` tombstones be migrated before this user-service version is deployed?
Reply in this thread, or post: `answer clar_t6yqlmflovm7js7zouec5etu4l: <your answer>`

user-service/mongorepo/subscriptions.go
```diff
@@ -72,13 +67,12 @@ func (r *SubscriptionRepo) EnsureIndexes(ctx context.Context) error {
...
-// roomsEnrichStages builds the shared rooms-join + enrichment. When dropDeleted is true
-// it drops local soft-deleted (^Del-) rooms — the list, count, and active paths all
-// pass true. A missing/cross-site room has no room.name so it is kept either way. The
+// roomsEnrichStages builds the shared rooms-join + enrichment. It never removes a
+// subscription: a missing or cross-site room simply yields no room fields. The
 // rooms-type activity window is applied separately by the caller on the room's
 // lastMsgAt (surfaced here).
-func roomsEnrichStages(dropDeleted bool) bson.A {
-	stages := bson.A{
+func roomsEnrichStages() bson.A {
```
julianshen, line 74: _🏷️ migration-compatibility_ | _🔴 Blocking_ | _`mongodb`_

**Legacy soft-deleted room documents are exposed as active rooms.**

The changed Mongo pipelines remove every `^Del-` check, and `CountActiveSubscriptions` no longer consults `rooms` at all. Existing `rooms` documents produced by the legacy rename protocol, or writes from an older room-service replica during a rolling deployment, will therefore reappear in lists, counts, and unread processing. Old and new user-service replicas will also return inconsistent results for the same Mongo data. Keep backward-compatible suppression until legacy producers are retired and existing documents are migrated, or introduce an explicit tombstone field with a rollout-compatible migration.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 74: The changed Mongo pipelines remove every `^Del-` check, and `CountActiveSubscriptions` no longer consults `rooms` at all. Existing `rooms` documents produced by the legacy rename protocol, or writes from an older room-service replica during a rolling deployment, will therefore reappear in lists, counts, and unread processing. Old and new user-service replicas will also return inconsistent results for the same Mongo data. Keep backward-compatible suppression until legacy producers are retired and existing documents are migrated, or introduce an explicit tombstone field with a rollout-compatible migration.

```

</details>

julianshen (commented):
tGD inline review

user-service/service/service.go
```diff
@@ -22,10 +22,9 @@ type SubscriptionRepository interface {
 	GetDMSubscription(ctx context.Context, account, target string) (*model.EnrichedDMSubscription, error)
 	GetSubscriptionByRoomID(ctx context.Context, account, roomID string) (*model.EnrichedSubscription, error)
 	CountActiveSubscriptions(ctx context.Context, account string) (int, error)
-	// GetActiveSubscriptions returns up to limit active subscriptions. The cap
-	// is applied before the deleted-room filter, so a page whose capped slice
-	// contains soft-deleted rooms comes back slightly short of limit — fine for
-	// the unread count (its only consumer), not a general pagination surface.
+	// GetActiveSubscriptions returns up to limit active subscriptions — exactly
```
julianshen: _📝 documentation_ | _🔵 Suggestion_ | _`tgd-review`_

**The repository contract incorrectly promises exactly limit rows**

The implementation returns fewer than `limit` when the account has fewer matching active subscriptions. Although downstream enrichment no longer shortens a capped result, the interface comment's unconditional “exactly limit” promise is still false at the end of the result set.

<details>

```
In `user-service/service/service.go` around line 25: The implementation returns fewer than `limit` when the account has fewer matching active subscriptions. Although downstream enrichment no longer shortens a capped result, the interface comment's unconditional “exactly limit” promise is still false at the end of the result set.

```

</details>

user-service/service/subscriptions.go
hmchangw: Checked this against `main` rather than the diff, and the fabricated-room behaviour is **pre-existing** — this PR does not introduce it.

On `main`, `buildLocalRoom` returned `nil` on exactly one condition:

```go
if strings.HasPrefix(sub.RoomName, deletedRoomNamePrefix) {
    return nil
}
```

A subscription whose local room document is missing has `RoomName == ""`, which does not match that prefix, so `main` already built a non-nil `SubscriptionRoom` with empty metadata for it. The `Room != nil` guards in `enrichLocal` were therefore already unreachable for the missing-room case; removing them changed nothing observable. What this PR changed is the *other* branch: a `Del-` prefixed name now gets a room object instead of `nil`.

So the two behaviours differ as follows, before and after:

| local room | `main` | this PR |
|---|---|---|
| missing | non-nil, empty metadata | non-nil, empty metadata |
| `Del-` prefixed | `nil` | non-nil, real metadata |

The underlying point stands on its own merits: clients cannot distinguish "no local room document" from "a real room with blank fields", and the enrichment contract would be cleaner if an unmatched lookup left `Room` nil. That needs a match marker carried out of `roomsEnrichStages` (the `$unwind` preserves nulls, so the absence is currently indistinguishable downstream from a room with zero values), plus service-level coverage. It is a behaviour change for the list and single-item endpoints — rows that today carry an empty room object would start carrying none — so I would rather not fold it into a perf PR that is otherwise behaviour-preserving for missing rooms.

Tracked as a follow-up. Happy to open it now if you would prefer it before this lands.

---

user-service/service/subscriptions.go
hmchangw: Same finding as [the later thread on `enrichLocal`](https://github.com/hmchangw/newchat/pull/353#discussion_r3837740895) — answered in full there.

Short version: verified against `main`, and the fabricated empty room for a missing local room is **pre-existing**, not introduced here. `buildLocalRoom` returned `nil` only for the `Del-` prefix, and a missing room has `RoomName == ""`, which never matched it. The real change is that a `Del-` prefixed name now gets a room object rather than `nil`.

The underlying contract point is fair and tracked as a follow-up; it needs a lookup-match marker out of `roomsEnrichStages` and is a client-visible change, so not folded into this PR.

---

user-service/service/subscriptions.go
hmchangw: Agreed this is real, and it is the one thread I am not proposing to close with a code change — it needs a rollout decision.

Two things narrowed it since you wrote this:

**#354 merged** (`bd6f596`), which stops `oplog-collections-transformer` replicating `Del-` names from the legacy source. That was the producer that mattered: it copied `sr.displayName()` verbatim into `room.Name` and into the `room_renamed` payload, so a legacy soft-delete propagated into both `rooms.name` and every `subscriptions.name`. That path is now closed.

**What remains is not a deletion protocol.** I traced the live write side: `room-service/handler.go:1960-1963` validates a rename only for trim / non-empty / ≤100 runes, so a channel owner can still set a `Del-` prefixed name through the app — but after #354 that is a name someone typed, not a soft delete. Nothing in this repo has ever *originated* the prefix as a deletion; the two textual matches in Go code on `main` were both comments.

So the mixed-version window you describe still exists, with a smaller consequence than "deleted rooms resurrect": during a rolling deploy, a room a user happened to name `Del-…` is visible on new replicas and suppressed on old ones, converging once rollout completes. Genuinely legacy-soft-deleted rooms are only at issue for rows migrated *before* #354 landed.

Your suggested alternatives both have real costs worth naming. Keeping backward-compatible suppression indefinitely means the read path keeps interpreting a name, which is the thing this PR removes. An explicit tombstone field is the right long-term shape, but it needs a producer, a backfill, and a propagation contract — and #352 was closed precisely because leaning the read path on an asynchronous denormalised field turned out to be the weaker foundation.

What I would want before merging, and cannot determine from here:

1. Whether any `Del-` prefixed rows already exist in the target databases (pre-#354 migrated data) — a `count({name: /^Del-/})` per site on `rooms` and `subscriptions` settles it.
2. If the count is zero, the mixed-version window is vacuous and this can land on any rollout order.
3. If it is non-zero, those rows want cleaning up before rollout, and that is a data task rather than a code one.

Leaving this open pending that check rather than resolving it.

---

docs/client-api.md
```diff
@@ -5522,7 +5520,7 @@ Same shape as `subscription.list` — a (here, at most one) list:
 
 Returns the count of active subscriptions, optionally filtered to unread rooms only.
 
-**Active set:** an active subscription is a non-muted DM or channel, **or** a botDM that is both non-muted **and** subscribed (`isSubscribed: true`). Unsubscribed botDMs and muted rooms of any type are excluded from the count. Rooms that have been soft-deleted (renamed with a `Del-` prefix) are also excluded.
+**Active set:** an active subscription is a non-muted DM or channel, **or** a botDM that is both non-muted **and** subscribed (`isSubscribed: true`). Unsubscribed botDMs and muted rooms of any type are excluded from the count. The count reads subscription state only — it consults no room document, and no room name is filtered.
```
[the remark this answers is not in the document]
hmchangw: Fixed in `a95bc98`. Verified against `activeSubscriptionFilter` first — it does apply `"open": {"$ne": false}`, and the definition omitted it.

The definition now reads:

> an active subscription is a non-muted, **open** DM or channel, **or** a botDM that is non-muted, open **and** subscribed (`isSubscribed: true`). Excluded from the count: unsubscribed botDMs, muted rooms of any type, and rooms the user has closed (`open: false`) — closed rooms are hidden from `subscription.list`, so counting them would put the badge and the list permanently out of step. A missing `open` field (legacy documents) counts as open.

Added the missing-field case too, since `$ne: false` passes legacy documents that have no `open` field at all, and a client would otherwise have to guess.

Same commit corrects a second overstatement in the same area that a reviewer caught on the sibling PR: `GetActiveSubscriptions` was documented as returning "exactly `limit`" when it returns *up to* `limit`, and is uncapped for `limit <= 0`.

---

user-service/service/subscriptions.go
hmchangw: Resolved by a data check: **no `Del-` prefixed rows exist in the destination databases** as of now. Reported by the repository owner, who ran the check — I have no database access from this session, so I am relaying rather than attesting to it.

That makes the scenario in this thread vacuous rather than mitigated. The predicate this PR removes matches nothing, so no request — to an old replica or a new one — can observe a difference. There is no set of rooms to resurrect, and rollout order does not matter.

For the record, the two ways such a row could have existed, and why neither does:

- **Legacy migration.** #354 (`bd6f596`, merged) stops `oplog-collections-transformer` importing soft-deleted records: a room already `Del-` at migration time is never imported, and a later rename-into-`Del-` is skipped so the destination keeps the live name. Rows migrated *before* #354 landed could have carried the prefix — the check above says none did.
- **Live rename.** `room-service/handler.go:1960-1963` validates a rename only for trim / non-empty / ≤100 runes, so a `Del-` prefixed name is still accepted. But after #354 that is a name a user typed, never a deletion marker.

One residual case, stated for completeness rather than as an objection: if someone renames a room to a `Del-` prefixed name *during* the deploy window, that room would be visible on new replicas and hidden on old ones until rollout completes. It is a user-chosen name, the divergence is cosmetic, and it converges on its own — this PR's behaviour (show it) is the correct one of the two.

Worth re-running the count immediately before rollout, since the check is a point-in-time snapshot.

---

mliu33 (approved):
Thanks a lot!

PR #354, opened 2026-08-22
Skip soft-deleted rooms and subscriptions in CDC migration

hmchangw:
## Summary

The legacy source app has no room-delete operation and no delete flag: "deleting" a room renames it — and the denormalized name on every subscription to it — to `Del-`+name. This branch keeps those records out of the migration.

**Invariant: a room or subscription carrying the `Del-` marker must never appear in the destination MongoDB — no doc, no rename, no denormalized copy.**

## Changes

- **`isSoftDeletedRecord`** (`classify.go`): reports whether a source room or subscription carries the marker on `name` **or** `fname`. Only the exact prefix counts — `delta`, `del-general` and `team-Del-old` are live rooms.
- **Room lane** (`rooms.go`): any op (`insert` / `replace` / `update`) on a `Del-` room is skipped with `migration.ErrSkipped`, metered `room_soft_deleted`. Room-type classification runs first so a `Del-` livechat/voip/group-DM room keeps metering its own exclusion reason.
- **Subscription lane** (`subscriptions.go`): same, metered `subscription_soft_deleted`. Skipping the field events is required rather than merely tidy — `inbox-worker` treats a missing subscription as a redelivery signal (`UpdateSubscriptionRoles`/mute/favorite/open), so `role_updated` and friends would Nak-retry to exhaustion against a subscription whose insert this guard skipped.
- **Docs**: `CDC_COVERAGE.md` states the invariant, the per-op table, and the residuals; `SOURCE_DATA.md` records the `Del-` convention and the open questions it rests on.

## Why the invariant is provable, not just asserted

Every name this transformer can write into Mongo arrives via a published event:

| Event | Writes | Guarded by |
|---|---|---|
| `room_sync` | `rooms.name` | room lane |
| `room_renamed` | `subscriptions.name`, for every sub in the room | room lane |
| `member_added` | the new subscription's name | subscription lane |

`invariant_test.go` therefore proves the rule by asserting that no `Del-` source doc — across every doc shape, both lanes, every op, and six update-delta shapes — publishes anything at all: **99 cases**. The `company_room_members` and `company_thread_subscriptions` lanes carry no name field and cannot violate it.

## Known consequences (documented, not coded around)

- **A room soft-deleted after the CDC checkpoint keeps its original, unprefixed name** and stays visible to its members. Hiding it would mean writing the very document the invariant forbids, and the destination has no room-delete path (no `inbox-worker` handler, no room-delete RPC). Removal is an ops backfill: query the **source** for `Del-` prefixed names and reconcile against the destination `rooms` collection by `_id`. Building a destination delete path is separate, deliberately-scoped work.
- **DMs are not exempt.** For `t:"d"` the name fields hold the peer's username and display name, so a user literally named `Del-…` costs their DM. That is the deliberate price of a total invariant — type-scoping the check would let a `Del-` named DM document into Mongo.
- The `…_events_skipped_total{reason=room_soft_deleted}` counter reports skipped **event volume** only — it counts every event on a dead room, not unique rooms, and carries no room id. The per-skip log line carries `roomId` for tracing an individual record.

## Verification

`make lint` 0 issues · `gosec` clean · `go test -race ./data-migration/...` green · CI green on the head commit, including `sast` and `test-integration (data-migration)`.

https://claude.ai/code/session_013vDkcvPJaoXQdEkK9rupXC

### DIFF
--- a/data-migration/CDC_COVERAGE.md
+++ b/data-migration/CDC_COVERAGE.md
@@ -31,14 +31,14 @@ The connector forwards raw change-stream events with **no `updateLookup`** and *
 | # | Source event | Op + payload | Source lookup (by `_id`) | Current-system facts | Handling / impact |
 |---|---|---|---|---|---|
 | **Rooms** |
-| 1 | Room create | `insert` — full doc | in payload | `t` ∈ `c,p,d,l,v`; `prid`⇒discussion; `teamId`/`teamMain`; `d` can have >2 users | ✅ map → `room_sync` (skip `l`,`v`,group-DM, soft-deleted) |
-| 2 | Room replace | `replace` — full doc | not needed | whole-doc rewrite; can cross type/exclusion boundary; **no delta** to tell which fields changed | ✅ re-classify → `room_renamed` + `room_restricted` + `room_sync` (skip soft-deleted; conservative — field events are idempotent + guarded; subs' denormalized name/visibility must not go stale) |
-| 3 | Room change | `update` — changed fields only | full current doc | — | ✅ re-read doc → `room_renamed` / `room_restricted` / `room_sync` (skip soft-deleted) |
+| 1 | Room create | `insert` — full doc | in payload | `t` ∈ `c,p,d,l,v`; `prid`⇒discussion; `teamId`/`teamMain`; `d` can have >2 users | ✅ map → `room_sync` (skip `l`,`v`,group-DM, already-soft-deleted) |
+| 2 | Room replace | `replace` — full doc | not needed | whole-doc rewrite; can cross type/exclusion boundary; **no delta** to tell which fields changed | ✅ re-classify → `room_renamed` + `room_restricted` + `room_sync` (skip already-soft-deleted; conservative — field events are idempotent + guarded; subs' denormalized name/visibility must not go stale) |
+| 3 | Room change | `update` — changed fields only | full current doc | — | ✅ re-read doc → `room_renamed` / `room_restricted` / `room_sync`; a rename INTO `Del-` **is** the delete and is applied, other changes on a `Del-` room are skipped |
 | 4 | Room delete | `delete` — `_id` only | nothing — doc gone | app has no room-delete operation | ❌ skip (no app deletion; un-actionable) |
 | **Subscriptions** |
-| 5 | Sub create | `insert` — full doc | in payload | `u`, `rid`, `roles[]`, `open`, `f`, `disableNotifications`, `ls`/`lr`, `alert` | ✅ `member_added` + state events (skip soft-deleted) |
-| 6 | Sub replace | `replace` — full doc | not needed | whole-doc rewrite | ✅ re-classify → `member_added` + state (skip soft-deleted) |
-| 7 | Sub change (incl. leave/rejoin) | `update` — changed fields only | full current doc | leaving sets `open:false` (not a row delete) | ✅ re-read doc → `open`-toggle → `member_added`/`member_removed`; mute/fav/role/read → matching event (skip soft-deleted) |
+| 5 | Sub create | `insert` — full doc | in payload | `u`, `rid`, `roles[]`, `open`, `f`, `disableNotifications`, `ls`/`lr`, `alert` | ✅ `member_added` + state events (skip subs to soft-deleted rooms) |
+| 6 | Sub replace | `replace` — full doc | not needed | whole-doc rewrite | ✅ re-classify → `member_added` + state (skip subs to soft-deleted rooms) |
+| 7 | Sub change (incl. leave/rejoin) | `update` — changed fields only | full current doc | leaving sets `open:false` (not a row delete) | ✅ re-read doc → `open`-toggle → `member_added`/`member_removed`; mute/fav/role/read → matching event (skip subs to soft-deleted rooms) |
 | 8 | Sub delete (true row removal) | `delete` — `_id` only | nothing — doc gone | destination subs key by generated `UUIDv7`, not source `_id`; removal needs `(roomID, account)` | ❌ skip (un-actionable; rare — leave is `open:false`) |
 | **Thread subscriptions** |
 | 9 | Follow / first reply | `insert` — full doc | in payload | keyed `(u._id, parentMessage._id)`; carries `rid`, `lastSeenAt`, `unreadMention` | ✅ resolve thread-room+user → `thread_subscription_upserted` |
@@ -56,17 +56,37 @@ The connector forwards raw change-stream events with **no `updateLookup`** and *
 
 ### Soft-deleted rooms (`Del-` prefix)
 
-The source "deletes" a room by renaming it — and every subscription's denormalized name — to
-`Del-`+name; there is no delete flag and no row removal. Rooms and subscriptions whose `name` **or**
-`fname` carries that exact prefix are **never imported**, on any op (`insert`/`replace`/`update`),
-metered as `room_soft_deleted` / `subscription_soft_deleted`. Only the exact `Del-` prefix counts
-(`delta`, `del-general`, `team-Del-old` are live rooms) — the same marker `user-service` filters
-subscription reads on.
+The source has no room-delete operation and no delete flag: "deleting" a room renames it — and the
+denormalized name on every subscription to it — to `Del-`+name. The destination honors the same
+marker (`user-service/mongorepo/subscriptions.go:21` filters `^Del-` rooms out of
+`subscription.list` / `getChannels` / `count`), but nothing in the new stack ever *writes* it, so
+this migration is its only producer. That makes the rename both the deletion **event** and the
+deletion **state**, handled differently per op:
 
-**Consequence:** a room soft-deleted *after* the CDC checkpoint arrives as a rename-into-`Del-`
-update, which is skipped like any other soft-deleted record — the destination room (imported while
-it was live) stays as it was. The new stack has no room deletion to apply, so this is the same
-end-state as row 4 (`Room delete` → skip).
+| Source event | Meaning | Handling |
+|---|---|---|
+| `insert` / `replace` of a `Del-` doc | born deleted — never existed downstream | ❌ skip (`room_soft_deleted`) |
+| `update` whose delta touches `name`/`fname`, doc now `Del-` | the deletion itself | ✅ apply → `room_renamed` + `room_sync`; the destination room takes the prefixed name and the `^Del-` filter hides it |
+| `update` touching anything else on a `Del-` doc | churn on a dead room | ❌ skip (`room_soft_deleted`) |
+| Any subscription op on a `Del-` sub | sub to a deleted room | ❌ skip (`subscription_soft_deleted`) |
+
+Only the exact prefix counts — `delta`, `del-general` and `team-Del-old` are live rooms.
+
+**Subscriptions never carry the deletion.** It rides the room lane: `room_renamed` →
+`UpdateSubscriptionNamesForRoom` already rewrites the name on *every* sub in the room, so subs
+imported while the room was live stay consistent. Letting sub-lane field events through instead
+would be actively harmful — `UpdateSubscriptionRoles`/mute/favorite/open treat a missing
+subscription as an error so the event redelivers until `member_added` lands
+(`inbox-worker/main.go:118`), which never comes for a sub whose insert this guard skipped.
+
+**DMs (`t:"d"`) are exempt on both lanes.** There `name`/`fname` hold the peer's username and
+display name, not a room name, so the prefix would match a user rather than a deletion. A
+soft-deleted DM that slips through is still hidden downstream — the `^Del-` room filter is
+type-agnostic.
+
+**Residual:** `UpsertRoom` upserts (`inbox-worker/main.go:85`), so applying the deletion rename for
+a room whose `insert` was skipped *creates* a hidden `Del-` room doc rather than no-op'ing.
+Invisible to users; avoiding it would cost a room-existence read per event.
 
 ## Direct-transfer collections (oplog-direct-transfer)
 
--- a/data-migration/CDC_COVERAGE.md
+++ b/data-migration/CDC_COVERAGE.md
@@ -31,14 +31,14 @@ The connector forwards raw change-stream events with **no `updateLookup`** and *
 | # | Source event | Op + payload | Source lookup (by `_id`) | Current-system facts | Handling / impact |
 |---|---|---|---|---|---|
 | **Rooms** |
-| 1 | Room create | `insert` — full doc | in payload | `t` ∈ `c,p,d,l,v`; `prid`⇒discussion; `teamId`/`teamMain`; `d` can have >2 users | ✅ map → `room_sync` (skip `l`,`v`,group-DM) |
-| 2 | Room replace | `replace` — full doc | not needed | whole-doc rewrite; can cross type/exclusion boundary; **no delta** to tell which fields changed | ✅ re-classify → `room_renamed` + `room_restricted` + `room_sync` (conservative — field events are idempotent + guarded; subs' denormalized name/visibility must not go stale) |
-| 3 | Room change | `update` — changed fields only | full current doc | — | ✅ re-read doc → `room_renamed` / `room_restricted` / `room_sync` |
+| 1 | Room create | `insert` — full doc | in payload | `t` ∈ `c,p,d,l,v`; `prid`⇒discussion; `teamId`/`teamMain`; `d` can have >2 users | ✅ map → `room_sync` (skip `l`,`v`,group-DM, soft-deleted) |
+| 2 | Room replace | `replace` — full doc | not needed | whole-doc rewrite; can cross type/exclusion boundary; **no delta** to tell which fields changed | ✅ re-classify → `room_renamed` + `room_restricted` + `room_sync` (skip soft-deleted; conservative — field events are idempotent + guarded; subs' denormalized name/visibility must not go stale) |
+| 3 | Room change | `update` — changed fields only | full current doc | — | ✅ re-read doc → `room_renamed` / `room_restricted` / `room_sync` (skip soft-deleted) |
 | 4 | Room delete | `delete` — `_id` only | nothing — doc gone | app has no room-delete operation | ❌ skip (no app deletion; un-actionable) |
 | **Subscriptions** |
-| 5 | Sub create | `insert` — full doc | in payload | `u`, `rid`, `roles[]`, `open`, `f`, `disableNotifications`, `ls`/`lr`, `alert` | ✅ `member_added` + state events |
-| 6 | Sub replace | `replace` — full doc | not needed | whole-doc rewrite | ✅ re-classify → `member_added` + state |
-| 7 | Sub change (incl. leave/rejoin) | `update` — changed fields only | full current doc | leaving sets `open:false` (not a row delete) | ✅ re-read doc → `open`-toggle → `member_added`/`member_removed`; mute/fav/role/read → matching event |
+| 5 | Sub create | `insert` — full doc | in payload | `u`, `rid`, `roles[]`, `open`, `f`, `disableNotifications`, `ls`/`lr`, `alert` | ✅ `member_added` + state events (skip soft-deleted) |
+| 6 | Sub replace | `replace` — full doc | not needed | whole-doc rewrite | ✅ re-classify → `member_added` + state (skip soft-deleted) |
+| 7 | Sub change (incl. leave/rejoin) | `update` — changed fields only | full current doc | leaving sets `open:false` (not a row delete) | ✅ re-read doc → `open`-toggle → `member_added`/`member_removed`; mute/fav/role/read → matching event (skip soft-deleted) |
 | 8 | Sub delete (true row removal) | `delete` — `_id` only | nothing — doc gone | destination subs key by generated `UUIDv7`, not source `_id`; removal needs `(roomID, account)` | ❌ skip (un-actionable; rare — leave is `open:false`) |
 | **Thread subscriptions** |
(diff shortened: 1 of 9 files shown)


julianshen:
**No actionable comments.** ✅

### Disputed

- `data-migration/oplog-collections-transformer/rooms.go:99` (`nats`) — Concrete lifecycle: a live room is first published via `room_sync`, then the source renames it to `Del-...`; this guard publishes neither `room_sync` nor `room_renamed` and returns `migration.ErrSkipped`, whose normal consumer handling terminally acknowledges the delivery. The JetStream checkpoint can therefore advance while Mongo retains the live room and dependent subscriptions indefinitely. The discussion confirms this diagnosis; documenting a later source-driven backfill does not make the consumed deletion durable. Before returning `ErrSkipped`, publish an idempotent room tombstone handled by the destination, or durably persist a reconciliation record keyed by `sr.ID`; propagate persistence/publish failures so the message is retried rather than acknowledged.
- `data-migration/oplog-collections-transformer/rooms.go:99` (`tgd-review`) — When a previously migrated live room is renamed to `Del-...`, this guard publishes neither `room_sync` nor `room_renamed`, and the subscription guard suppresses the related updates. The destination therefore retains the original unprefixed room and subscriptions, so the documented `^Del-` filter does not hide the deleted room. The documentation acknowledges this, but an unspecified ops backfill does not prevent the user-visible regression when CDC later consumes another deletion. Implement a durable room deletion/tombstone path before advancing the checkpoint, or retain the existing `Del-` rename behavior until such a path is available; add a live-room-then-delete lifecycle test against destination state.

### Local review memory

- `data-migration/oplog-collections-transformer/rooms.go:99` — @hmchangw: Correct diagnosis, and it is a known, documented consequence rather than an oversight. Grouping the three related findings here (this one, the silent-failure one at rooms.go:106, and the observability one at rooms.go:103). A room importe… ([thread](https://github.com/hmchangw/newchat/pull/354#discussion_r3836511045))

<details>

* `data-migration/CDC_COVERAGE.md`
* `data-migration/SOURCE_DATA.md`
* `data-migration/oplog-collections-transformer/classify.go`
* `data-migration/oplog-collections-transformer/invariant_test.go`
* `data-migration/oplog-collections-transformer/rooms.go`
* `data-migration/oplog-collections-transformer/rooms_test.go`
* `data-migration/oplog-collections-transformer/subscriptions.go`
* `data-migration/oplog-collections-transformer/subscriptions_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

data-migration/oplog-collections-transformer/subscriptions.go
```diff
@@ -106,6 +106,16 @@ func (h *handler) handleSubscription(ctx context.Context, ev oplogEvent) error {
 		return fmt.Errorf("%w: decode source subscription: %v", migration.ErrPoison, uerr) //nolint:errorlint // intentional single-%w sentinel wrap; decode err is informational only
 	}
 
+	if isSoftDeletedName(ss.Name, ss.FName) {
```
julianshen: _🎯 correctness_ | _🟠 Warning_ | _`cassandra`_

**Soft-delete updates leave existing Cassandra memberships active**

The affected downstream mutation is `member_removed`. A live subscription can first emit `member_added` and create destination membership/subscription rows; when the source later soft-deletes its room, the subscription is renamed to `Del-*`. This return occurs before `handleSubscriptionUpdate`, so update or replace events emit neither `member_removed` nor another Cassandra cleanup mutation. The source hides these subscriptions, but the destination retains the old membership and denormalized room data. Skip prefixed inserts, but translate prefixed updates/replaces for previously imported subscriptions into an idempotent removal, and test the live-import followed by soft-delete sequence.

<details>

```
In `data-migration/oplog-collections-transformer/subscriptions.go` around line 109: The affected downstream mutation is `member_removed`. A live subscription can first emit `member_added` and create destination membership/subscription rows; when the source later soft-deletes its room, the subscription is renamed to `Del-*`. This return occurs before `handleSubscriptionUpdate`, so update or replace events emit neither `member_removed` nor another Cassandra cleanup mutation. The source hides these subscriptions, but the destination retains the old membership and denormalized room data. Skip prefixed inserts, but translate prefixed updates/replaces for previously imported subscriptions into an idempotent removal, and test the live-import followed by soft-delete sequence.

```

</details>

julianshen (commented):
tGD inline review

julianshen:
## Needs clarification

Can a \`Del-\`-prefixed room ever be restored after migration begins?
Reply in this thread, or post: `answer clar_5c4zwy7cr6ovwqf3mzuzjo7dbe: <your answer>`

data-migration/oplog-collections-transformer/rooms.go
```diff
@@ -94,6 +96,16 @@ func (h *handler) handleRoom(ctx context.Context, ev oplogEvent) error {
 		return migration.ErrSkipped
 	}
 
+	if isSoftDeletedRecord(sr.Name, sr.FName) {
```
julianshen, line 99: _🏷️ consistency_ | _🔴 Blocking_ | _`mongodb`_

**Soft-deletes leave live Mongo documents indefinitely.**

A room first migrated while live already has `rooms` and `subscriptions` documents under its original name. When the source later renames it to `Del-...`, this guard skips `room_sync` and `room_renamed`, while the subscription guard skips the corresponding subscription updates. Mongo therefore retains the unprefixed documents, and the documented `^Del-` query filter will not hide them, so members can continue seeing a deleted room. Add an idempotent soft-delete event carrying the room ID and destination handling that removes or tombstones the room and its dependent documents; alternatively, emit the existing `Del-` rename if hiding rather than physical removal is acceptable. Cover the live-insert-then-delete lifecycle and assert the final Mongo state.

<details>

```
In `data-migration/oplog-collections-transformer/rooms.go` around line 99: A room first migrated while live already has `rooms` and `subscriptions` documents under its original name. When the source later renames it to `Del-...`, this guard skips `room_sync` and `room_renamed`, while the subscription guard skips the corresponding subscription updates. Mongo therefore retains the unprefixed documents, and the documented `^Del-` query filter will not hide them, so members can continue seeing a deleted room. Add an idempotent soft-delete event carrying the room ID and destination handling that removes or tombstones the room and its dependent documents; alternatively, emit the existing `Del-` rename if hiding rather than physical removal is acceptable. Cover the live-insert-then-delete lifecycle and assert the final Mongo state.

```

</details>

data-migration/oplog-collections-transformer/classify.go
```diff
@@ -1,6 +1,32 @@
...
+
+// dmSourceType is the source room type for a direct message.
+const dmSourceType = "d"
+
+// isSoftDeletedRecord reports whether a source room or subscription doc carries the soft-delete
+// marker on any name field. The invariant it enforces is absolute — a "Del-" prefixed room or
+// subscription must never reach the destination Mongo — so the check is type-agnostic and covers
+// both fields: DMs are included even though their name fields hold the peer's username and display
+// name, which costs a live DM whose peer is literally named "Del-…" but keeps the invariant total.
+func isSoftDeletedRecord(names ...string) bool {
```
julianshen: _🏷️ data-modeling_ | _🔴 Blocking_ | _`mongodb`_

**DM peer names are incorrectly treated as room deletion markers.**

For `t:"d"`, the changed documentation states that `name` and `fname` contain the peer's username and display name, not a room deletion field. The type-agnostic helper therefore drops a valid DM whenever either peer name begins with `Del-`; both its room and subscription documents will be absent from Mongo. Do not infer DM deletion from these user-controlled fields. Use an authoritative deletion signal confirmed from the source model, or scope the prefix check to room types and fields where it semantically represents a room soft-delete.

<details>

```
In `data-migration/oplog-collections-transformer/classify.go` around line 22: For `t:"d"`, the changed documentation states that `name` and `fname` contain the peer's username and display name, not a room deletion field. The type-agnostic helper therefore drops a valid DM whenever either peer name begins with `Del-`; both its room and subscription documents will be absent from Mongo. Do not infer DM deletion from these user-controlled fields. Use an authoritative deletion signal confirmed from the source model, or scope the prefix check to room types and fields where it semantically represents a room soft-delete.

```

</details>

data-migration/oplog-collections-transformer/rooms.go
```diff
@@ -94,6 +96,16 @@ func (h *handler) handleRoom(ctx context.Context, ev oplogEvent) error {
...
+		// Soft-deleted at source: skipped on every op, and before the delta is even decoded. Nothing
+		// about this room may reach Mongo — not the doc, not a rename of it — because room_sync would
+		// write the "Del-" name onto rooms.name and room_renamed onto every subscription in the room.
+		slog.DebugContext(ctx, "skip soft-deleted room",
+			"op", ev.Op, "t", sr.T, "eventId", ev.EventID, "request_id", natsutil.RequestIDFromContext(ctx))
+		h.metrics.onSkipped(ctx, "room_soft_deleted")
+		return migration.ErrSkipped
```
julianshen, line 106: _🏷️ data consistency_ | _🔴 Blocking_ | _`silent-failure`_

**Soft deletes are acknowledged without durable remediation**

Returning migration.ErrSkipped publishes nothing, so a room imported while live and later renamed with the Del- prefix remains visible in the destination under its old name; the added tests explicitly require no event to be published. The documentation delegates cleanup to an ops backfill, but this path only increments a counter, which does not retain the affected room ID. Once the CDC checkpoint advances, the deletion cannot be recovered reliably. Persist an idempotent remediation/tombstone record keyed by sr.ID before treating the event as skipped, or implement a destination deletion event and propagate any failure.

<details>

```
In `data-migration/oplog-collections-transformer/rooms.go` around line 106: Returning migration.ErrSkipped publishes nothing, so a room imported while live and later renamed with the Del- prefix remains visible in the destination under its old name; the added tests explicitly require no event to be published. The documentation delegates cleanup to an ops backfill, but this path only increments a counter, which does not retain the affected room ID. Once the CDC checkpoint advances, the deletion cannot be recovered reliably. Persist an idempotent remediation/tombstone record keyed by sr.ID before treating the event as skipped, or implement a destination deletion event and propagate any failure.

```

</details>

data-migration/oplog-collections-transformer/rooms.go
```diff
@@ -94,6 +96,16 @@ func (h *handler) handleRoom(ctx context.Context, ev oplogEvent) error {
 		return migration.ErrSkipped
 	}
 
+	if isSoftDeletedRecord(sr.Name, sr.FName) {
+		// Soft-deleted at source: skipped on every op, and before the delta is even decoded. Nothing
+		// about this room may reach Mongo — not the doc, not a rename of it — because room_sync would
+		// write the "Del-" name onto rooms.name and room_renamed onto every subscription in the room.
+		slog.DebugContext(ctx, "skip soft-deleted room",
```
julianshen, line 102: _🏷️ observability_ | _🟠 Warning_ | _`distributed-system`_

**Skipped deletions do not retain the room ID for the promised backfill**

A live room can be migrated, then soft-deleted; this path publishes no tombstone and records only a reason counter plus a debug log containing op, type, and event ID—not the room ID. The destination deliberately retains the stale room, but the documented ops backfill cannot derive affected IDs from `...events_skipped_total`, and the consumed event may no longer be available. Persist a reconciliation record keyed by room ID, or emit a production-retained structured audit event containing the room ID; do not use an unbounded ID metric label.

<details>

```
In `data-migration/oplog-collections-transformer/rooms.go` around line 103: A live room can be migrated, then soft-deleted; this path publishes no tombstone and records only a reason counter plus a debug log containing op, type, and event ID—not the room ID. The destination deliberately retains the stale room, but the documented ops backfill cannot derive affected IDs from `...events_skipped_total`, and the consumed event may no longer be available. Persist a reconciliation record keyed by room ID, or emit a production-retained structured audit event containing the room ID; do not use an unbounded ID metric label.

```

</details>

data-migration/CDC_COVERAGE.md
```diff
@@ -54,6 +54,49 @@ The connector forwards raw change-stream events with **no `updateLookup`** and *
...
+another casing.
+
+**Rooms deleted after the CDC checkpoint keep their live name.** A room imported while live and
+soft-deleted later arrives as a rename into `Del-`, which is skipped like everything else about a
+dead room — so the destination keeps the room under its **original, unprefixed** name. The invariant
+holds (no `Del-` doc exists), but that room stays visible to its members. Applying the rename would
+hide it via user-service's `^Del-` filter, at the cost of writing exactly the doc this invariant
+forbids. **Removing such rooms is an ops backfill, not a CDC behaviour** — the destination has no
+room-delete path, and the affected ids are exactly those counted by
+`…_events_skipped_total{reason=room_soft_deleted}`.
```
julianshen: _📝 documentation_ | _🔵 Suggestion_ | _`tgd-review`_

**The skipped-event counter cannot identify affected room IDs**

This counter increments once per skipped event, so repeated updates to one dead room are counted multiple times, and its labels do not expose room IDs. It therefore cannot identify or accurately count the unique rooms needed by the stated ops backfill.

<details>

> ‼️ **IMPORTANT**

```suggestion
forbids. **Removing such rooms is an ops backfill, not a CDC behaviour** — the destination has no
room-delete path. Identify affected room IDs by querying the source for `name` or `fname` with the
`Del-` prefix; `…_events_skipped_total{reason=room_soft_deleted}` reports skipped event volume only.
```

</details>

<details>

```
In `data-migration/CDC_COVERAGE.md` around line 92: This counter increments once per skipped event, so repeated updates to one dead room are counted multiple times, and its labels do not expose room IDs. It therefore cannot identify or accurately count the unique rooms needed by the stated ops backfill.

```

</details>

julianshen (commented):
tGD inline review

hmchangw:
answer clar_5c4zwy7cr6ovwqf3mzuzjo7dbe: Unknown — and the branch deliberately does not assume either way.

The source app has no documented un-delete: `SOURCE_DATA.md` §3 now records the `Del-` rename as the only deletion mechanism and lists "is it ever undone (restore)?" as an open question for the source team, alongside which field(s) carry the prefix and whether DM docs are renamed at all.

What the code does under that uncertainty: a restore is indistinguishable from an ordinary rename in the change stream. The connector runs without `fullDocumentBeforeChange`, so an update carries only the new value — there is no way to tell "renamed out of `Del-`" from "renamed" without prior state. If a restore did occur, the room would sync normally, but its subscriptions would not come back: their inserts were skipped while the room was dead, and the restore reaches the subscription lane as a `name`/`fname` delta, which carries no membership. Rebuilding membership on every rename-shaped subscription update would emit a `member_added` storm on every legitimate rename, so that is not a trade the CDC path should make silently.

This is documented as a residual in `CDC_COVERAGE.md` rather than coded around. If the source team confirms restore is real, the repair is a targeted backfill of the affected rooms, not a change to this guard.

---

data-migration/oplog-collections-transformer/classify.go
hmchangw: Not changing this — the type-agnostic check is deliberate, and scoping it by room type is the option that was explicitly rejected.

The requirement this branch implements is absolute: a room or subscription carrying the `Del-` marker must never appear in the destination MongoDB. Exempting `t:"d"` breaks exactly that. A DM room doc maps its destination `rooms.name` from `fname` (falling back to `name`), so a `Del-` prefixed DM would be written into Mongo verbatim — the precise outcome the invariant forbids — and `member_added` would carry the same string onto the subscription.

You are right about the cost, and it is documented rather than hidden (`CDC_COVERAGE.md`, "DMs are not exempt"): a live DM whose peer is literally named `Del-…` is dropped. The trade was made knowingly — a dropped DM is a bounded, identifiable loss, while a `Del-` document in Mongo violates the stated invariant.

On "use an authoritative deletion signal confirmed from the source model": there isn't one. The source has no delete flag and no row removal — the rename *is* the deletion. `SOURCE_DATA.md` §3 now records that, and flags which field(s) carry the prefix and whether DM docs are renamed at all as open questions for the source team. If they confirm DMs are never renamed, narrowing the check becomes safe and this comment is the place to revisit it.

---

data-migration/oplog-collections-transformer/rooms.go
hmchangw: Correct diagnosis, and it is a known, documented consequence rather than an oversight. Grouping the three related findings here (this one, the `silent-failure` one at rooms.go:106, and the observability one at rooms.go:103).

A room imported while live and soft-deleted afterwards keeps its original unprefixed name in Mongo and stays visible to its members. The invariant still holds — no `Del-` document exists — but the room lingers.

Why the suggested remedies aren't taken here:

- **"emit the existing `Del-` rename if hiding is acceptable"** — that is what the previous commit did, and it was reverted on instruction. It writes the marker onto `rooms.name` and, via `room_renamed`, onto every subscription in the room: exactly the documents this branch exists to keep out of Mongo.
- **"add a soft-delete event that removes or tombstones the room and its dependents"** — agreed that this is the real fix, and it is out of scope for this PR. The destination has no room-delete path at all: `inbox-worker` has no handler, and there is no room-delete RPC anywhere in the new stack (`DeleteRoomEvent` in `pkg/model` is message deletion). Adding one is a destination-side change spanning a new inbox event type, a handler, and the dependent-document cleanup you describe — a separate, deliberately-decided piece of work, not a small fix to this guard.

What did change from this review: the reconciliation claim was wrong and is fixed in `41ba19f`. `…_events_skipped_total{reason=room_soft_deleted}` counts events, not unique rooms, and carries no room id, so it cannot key a backfill — `CDC_COVERAGE.md` now says to query the **source** for `Del-` prefixed names and reconcile against the destination `rooms` collection by `_id`. Both skip logs now carry `roomId` so an individual record stays traceable.

Leaving this thread open: the destination delete path is a real gap and should stay visible, not be resolved away.

---

data-migration/CDC_COVERAGE.md
hmchangw: Agreed and fixed in `41ba19f` — the claim was wrong, for exactly the reason you give.

`CDC_COVERAGE.md` now says to identify the affected rooms by querying the **source** for docs whose `name` or `fname` carries the prefix, then reconcile against the destination `rooms` collection by `_id`. It states explicitly that `…_events_skipped_total{reason=room_soft_deleted}` reports skipped **event volume** only — it counts every event on a dead room, not unique rooms, and carries no room id (an unbounded metric label).

Both skip logs also now carry `roomId`, so an individual skipped record stays traceable even though the counter can't enumerate them.

---

data-migration/oplog-collections-transformer/subscriptions.go
hmchangw: This one went without a reply on the thread itself — answering it directly, since the membership angle isn't covered by the grouped response on rooms.go:99.

The diagnosis is right: a subscription imported while its room was live, then soft-deleted at source, keeps its destination `subscriptions` row and its `room_members` row. No `member_removed` is emitted, because the guard returns before `handleSubscriptionUpdate`.

Skipping those field events is deliberate, and not only for the invariant. `inbox-worker` treats a missing subscription as a redelivery signal — `UpdateSubscriptionRoles`, mute, favorite and open all call `naksIfSubscriptionMissing` (`inbox-worker/main.go:118`). So letting update/replace through for a `Del-` subscription would Nak-retry to `MaxDeliver` against every subscription whose insert this same guard skipped, which is the common case during a backfill. The two populations — "imported while live" and "never imported" — are indistinguishable from the event alone.

Translating a prefixed update into an idempotent removal needs the same destination machinery as the room case: an inbox event type and handler that don't exist yet (there is no room- or membership-delete path in the new stack). That is deliberately scoped as separate work, and it's the same gap tracked on the rooms.go:99 thread. Cassandra is not implicated — subscriptions and room members are Mongo-only; Cassandra holds message history.

One correction to the record: `CDC_COVERAGE.md` now notes that `company_room_members` and `company_thread_subscriptions` carry no room name, so a name-based guard structurally cannot cover them — filtering those lanes would need an rid-keyed "was this room skipped" check.

---

data-migration/oplog-collections-transformer/rooms.go
hmchangw: Partly fixed in `41ba19f`: both skip logs now carry `roomId` (`rooms.go`, `subscriptions.go`), so a skipped record is traceable to a specific room rather than only to a reason counter. Your point about not using an unbounded id metric label is respected — the counter is unchanged.

What is **not** done: a persisted reconciliation record. That is the same gap as the rooms.go:99 thread and needs destination machinery that does not exist yet. The interim answer, now stated in `CDC_COVERAGE.md`, is that the authoritative list of affected rooms comes from the **source** — query for docs whose `name`/`fname` carries the prefix and reconcile against the destination `rooms` collection by `_id` — which does not depend on the CDC checkpoint still holding the events.

Leaving open, since the durable-record ask is unresolved.

---

data-migration/oplog-collections-transformer/rooms.go
hmchangw: Answered in full on the [rooms.go:99 thread](https://github.com/hmchangw/newchat/pull/354#discussion_r3837673250) — flagging it here so this thread isn't silent.

Short version: the diagnosis is correct and the consequence is documented, not overlooked. Emitting the `Del-` rename is the option that was explicitly reverted (it writes the marker into `rooms.name` and every subscription in the room — the documents this branch exists to keep out of Mongo). A destination deletion event is the real fix and is separate work: no room-delete path exists anywhere in the new stack today.

What did land from this finding: `roomId` on both skip logs, and a corrected reconciliation procedure in `CDC_COVERAGE.md` that keys off the **source**, so it survives the CDC checkpoint advancing.

Leaving open — the destination delete path is a real gap and should stay visible.

---

mliu33 (approved):
Thanks! They might need to make these changes in house

PR #355, opened 2026-08-22
feat(broadcast-worker): federate room mention badges to remote sites

mliu33:
## Problem

`broadcast-worker` flags `hasMention=true` on mentioned accounts' subscriptions, but only in the **origin** site's Mongo. A federated user reads their subscription list from **their** home site, so a mention raised at another site never reaches them. `room-worker` already solves this shape for `member_added`; this gives room mentions the same lane.

## Approach

Group the mentionees by home site, then publish one OUTBOX event per destination for `outbox-worker` to forward into that site's INBOX — the established federation path.

```
broadcast-worker ──> OUTBOX ──> outbox-worker ──> INBOX ──> inbox-worker ──> Mongo
                (subscription_mention, concurrent lane)
```

`resolved.Participants` already carries each mentionee's `SiteID` from the existing `FindUsersByAccounts` call, so the new-message path adds **no extra Mongo round-trip**. The edit path needs one lookup, and it runs after the client broadcast so it stays off the user-visible latency path.

### Contract

| File | Change |
|---|---|
| `pkg/model/event.go` | `InboxSubscriptionMention = "subscription_mention"` + `SubscriptionMentionEvent` |
| `pkg/outbox/outbox.go` | added to `ConcurrentEventTypes` |

Dedup ID: `natsutil.InboxDedupID(ctx, destSiteID, seed)` → `{requestID}:{destSiteID}`, the same helper every `room-worker` federation site uses. `message-gatekeeper` rejects a submission whose `requestId` is not a UUID and publishes canonical via `natsutil.NewMsg`, so the request ID rides the canonical event as `X-Request-ID` and is identical on every redelivery. An edit is its own canonical event with its own request ID, so it can collide neither with the original send nor with another edit in the same millisecond.

Per-destination publishes run concurrently under `maxSiteFanout = 8`, matching `user-service` and `search-service`.

`Accounts` carries only the accounts homed at the destination — no site learns mention identities it has no business seeing.

## ⚠️ Deploy ordering — required

There is **no feature flag**; `broadcast-worker` starts producing `subscription_mention` the moment it rolls out. Deploy in this order:

1. **`inbox-worker` on every destination site.** An older one hits its unknown-event branch, returns nil and **Acks** — the badge is discarded silently.
2. **`outbox-worker`.** Its per-peer consumer filters are built from `ConcurrentEventTypes` at startup; a stale one leaves the new subject matching no lane, so events park in OUTBOX until retention deletes them, with a green publish metric.
3. **`broadcast-worker`.**

`TestIntegration_SubscriptionMentionForwardedViaConcurrentLane` pins the outbox-worker half so the filter can't silently drift.

## Accepted trade-offs

Both are deliberate and recorded in the design doc, not oversights.

**OUTBOX publish is best-effort.** It runs after the client fan-out and a failure is logged, never returned. Propagating it would mean a canonical redelivery re-broadcasts the message to every client in the room to save one remote badge. Revisit in a broader broadcast-worker refactor.

**A mention can overtake its `member_added`.** The concurrent and ordered lanes are separate consumers, so a mention can arrive before the subscription exists, match nothing, and lose that badge. Moving to `OrderedEventTypes` would park high-volume badges behind a `MaxAckPending=1` per-peer FIFO; NAKing on zero-match would make every mention of a non-member burn `MaxDeliver`. The user still receives the message, and later mentions badge normally. See "Accepted limitation" in the design doc.

## Not changed

- **Thread replies.** `handleThreadCreated` deliberately skips room-level badges, and `message-worker` already federates the thread subscription's mention.
- **`@all`.** Rides `UpdateRoomLastMessage`, a room-document concern.
- **Mention/read race.** A delayed read can clear a newer mention — pre-existing, inherent to `hasMention` having no companion timestamp, and equally reachable single-site today. Fixing it touches `room-service`, `user-service` and the local write plus a migration.

## Testing

Written TDD, red before green.

| Package | Coverage |
|---|---|
| `pkg/model` | round-trip |
| `pkg/outbox` | new type on the concurrent lane; partition stays disjoint |
| `broadcast-worker` | all-local, mixed sites, unresolved mentionee, `@all`-only, publish error swallowed, federation disabled, edit path, lookup failure still badges locally |
| `inbox-worker` | dispatch; malformed, blank-room, empty-accounts and zero-`mentionedAt` payloads all drop as `errcode.Permanent`; store error |
| `inbox-worker` (integration) | badges an unread and a stale-read sub; skips one that read past the message and a non-subscriber |
| `broadcast-worker` (integration) | producer leg against real Mongo + JetStream — correct subject, dedup ID and per-site account list; the local mentionee is badged and not federated |
| `outbox-worker` (integration) | relay leg — forwarded to the destination INBOX verbatim |

The producer-leg test was mutation-checked: flipping the remote user to the local site makes it fail.

Gates: `make lint` (0 issues), `make test` (race), `make test-integration` for all touched services, `make sast` (gosec/govulncheck/semgrep PASS).

## Docs

`docs/client-api.md` — noted the cross-site relay on `hasMention`, and corrected an existing line crediting `message-worker` with the room-level write (that's `broadcast-worker`; `message-worker` owns the thread one). Refreshed the OUTBOX producer lists in `CLAUDE.md` (three places) and the `pkg/outbox` package doc.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DWWLhH9X8FtkXAKNkhwm6o)_

## Summary by CodeRabbit

* **New Features**
  * Added cross-site mention badge support for room subscriptions.
  * Mentions update local badges and relay to the appropriate remote site.
  * Added remote room-activity synchronization for accurate cross-site room ordering.
  * Added thread-specific event delivery, including encrypted thread updates.
  * Added subscription mention events to the OUTBOX/INBOX delivery flow.

* **Bug Fixes**
  * Improved routing and deduplication for federated mention events.
  * Invalid or failed remote events no longer disrupt local processing.

* **Documentation**
  * Documented federation metadata, payload-size limits, thread events, and mention delivery.


mliu33:
Addressing the two items in CodeRabbit's merge-risk note — one fixed, one declined.

**Fixed — silent accept of a malformed federated event** (`1b06c75`).

`handleSubscriptionMention` unmarshalled a payload with a blank `roomId` successfully, then passed it to `UpdateMany`, which matched nothing and returned no error. The event was Acked with no write and no log. Our own producer can't emit one (`roomID` is always set), but a federated payload isn't ours to trust, and no redelivery can repair it. It's now logged and dropped, with the origin site named so the offending producer is identifiable:

```go
if e.RoomID == "" || len(e.Accounts) == 0 {
    slog.WarnContext(ctx, "dropping subscription_mention with no room or accounts",
        "room_id", e.RoomID, "accounts", len(e.Accounts), "origin_site", evt.SiteID)
    return nil
}
```

Two test cases added: blank room ID, and empty account list.

**Declined — "producers send a `timestamp` the consumer ignores."**

This is the repo's convention, not a contract mismatch. `CLAUDE.md` requires it:

> Every NATS event struct in `pkg/model` must include a `Timestamp int64` field […] set at the publish site.

It's the event-level publish time, deliberately distinct from the domain timestamp (`mentionedAt` here), and it's carried-but-unread by the consumers of `subscription_read`, `thread_unread_added`, and the other inbox payloads too. Removing it would make `SubscriptionMentionEvent` the one event struct that breaks the rule.

On the unmarshal path more broadly: I kept plain `fmt.Errorf` rather than `errcode.Permanent`, matching the 16-of-18 majority in this file. Worth a follow-up to make the whole file consistent one way or the other, but not something to change for one new handler.

---

mliu33:
The two inline findings landed after my previous comment. Both were valid and both are fixed in `2221d1c`.

**Correction to my earlier comment.** I read the merge-risk note's "producers send a `timestamp` field the consumer ignores" as a complaint about the `Timestamp` field and declined it. That was a misread — the inline comment makes the actual point clear: the **spec and plan documents still said `MessageCreatedAt`** while the code says `MentionedAt`. I renamed the field during a review pass and updated the code and tests but not the design docs I'd written earlier. A producer following those docs really would have sent a field the consumer ignores. My decline was wrong; the finding was right.

Fixed:
- Spec updated to `MentionedAt` with corrected semantics, and both dedup-ID formats replaced with the single implemented one, `mention:{roomID}:{msgID}:{mentionedAtMillis}:{destSiteID}`.
- Plan snippets updated for the rename and dedup format, plus a note at the top that the review pass also renamed the publisher field, made the per-site map lazily allocated, and gated the edit-path lookup on federation — the plan records the plan; the spec and code are authoritative.

**Missing `mentionedAt` — the real hole.** My previous commit guarded `roomId` and `accounts` but not the timestamp, and CodeRabbit is right that this is the worst of the three: a zero `mentionedAt` becomes 1970, so the store's `lastSeenAt $not/$gte` guard skips *every* subscription with any later read position, and the handler still returns success. Silent, and it looks like a working badge path. Now folded into the same drop:

```go
if e.RoomID == "" || len(e.Accounts) == 0 || e.MentionedAt <= 0 {
    slog.WarnContext(ctx, "dropping malformed subscription_mention", ...)
    return nil
}
```

One deliberate difference from the suggested diff: it returns `nil` (drop) rather than an error. Returning an error NAKs, and none of these three conditions can be repaired by redelivery — the event would just burn `MaxDeliver` and dead-letter. Dropping with a warning that names the origin site surfaces the producer bug without the retry churn. Test added for the omitted-`mentionedAt` case, alongside the blank-room and empty-accounts ones.

Note for whoever reviews next: CodeRabbit hit its hourly review limit before it could look at `1b06c75` or `2221d1c`, so its merge-risk banner still reflects `fd37b`.

---

julianshen:
**5 findings · 4 unique issues · 4 inline comments posted** — 🔴 2 blocking · 🟠 3 warning

### Disputed

- `broadcast-worker/main.go:241` (`distributed-system`) — A new broadcast-worker can enqueue `subscription_mention` while a destination still runs the previous inbox-worker. The updated outbox-worker forwards it, but the old inbox-worker's unknown-event branch returns nil, so the message is acknowledged and cannot be replayed after that worker is upgraded. The stated manual deployment order does not protect canary overlap or rollback. Keep production disabled behind a default-off capability flag until every destination consumer supports the event, or make unsupported events remain replayable.
- `broadcast-worker/main.go:241` (`mongodb`) — The producer enables `subscription_mention` publishing unconditionally. If any destination still runs the previous inbox-worker, its unknown-event branch returns nil, so the event is acknowledged without executing `SetSubscriptionMentions` against the `subscriptions` collection. Upgrading that worker later cannot replay the event, leaving `hasMention` false permanently. The discussed inbox-worker → outbox-worker → broadcast-worker deployment order avoids this only when followed across every site, but the code does not enforce it. Restore a default-off producer feature gate until all destinations support the event, or add a version/capability mechanism that keeps unsupported events replayable.
- `broadcast-worker/main.go:242` (`silent-failure`) — This enables `subscription_mention` production immediately, without capability negotiation or a rollout gate. During a mixed-version deployment or rollback, an older destination inbox-worker handles this unknown type through its default branch, returns nil, and therefore ACKs the event. The event is permanently removed before a compatible worker can process it, leaving the remote user's mention badge unset. The documented deployment order reduces the risk but does not enforce compatibility or protect rollbacks. Add a default-off producer flag enabled only after every destination supports the event, or introduce capability/version negotiation so unsupported events remain replayable.
- `inbox-worker/handler.go:184` (`tgd-review`) — The acknowledged consumer-first deployment order protects the planned forward rollout, but the protocol remains incompatible with older inbox-worker binaries: their unknown-event branch returns nil, causing `subscription_mention` to be ACKed without applying it. A rollback or ordering mistake after producers activate therefore loses badges irrecoverably. Retain a default-off producer gate until all destinations advertise support, or otherwise make unsupported events replayable.

### Local review memory

- `inbox-worker/handler.go:184` — @mliu33: Same finding as broadcast-worker/main.go:242 , answered there — the short version: A flag was added for exactly this ( MENTION_FEDERATION_ENABLED , default-off) and then removed at the owner's direction in ec1cf18 . The hazard you descri… ([thread](https://github.com/hmchangw/newchat/pull/355#discussion_r3837651684))
- `docs/superpowers/plans/2026-08-22-subscription-mention-fanout.md:738` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟡 Minor_ _⚡ Quick win_ &lt;details> &lt;summary>🧩 Analysis chain&lt;/summary> 🏁 Script executed: shell #!/bin/bash set -euo pipefail fd -e go xargs rg -n -C 8 \ 'func InboxDedupID InboxDedupID\( Handl… ([thread](https://github.com/hmchangw/newchat/pull/355#discussion_r3837712526))

<details>
<summary>📒 Files reviewed (20)</summary>

* `CLAUDE.md`
* `broadcast-worker/main.go`
* `broadcast-worker/testhelpers_test.go`
* `docs/client-api.md`
* `docs/superpowers/plans/2026-08-22-subscription-mention-fanout.md`
* `docs/superpowers/specs/2026-08-22-subscription-mention-fanout-design.md`
* `inbox-worker/handler.go`
* `inbox-worker/main.go`
* `outbox-worker/consumer_config_test.go`
* `outbox-worker/integration_test.go`
* `pkg/model/event.go`
* `pkg/model/model_test.go`
* `pkg/outbox/outbox.go`
* `pkg/outbox/outbox_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

pkg/outbox/outbox.go
```diff
@@ -40,6 +40,9 @@ var ConcurrentEventTypes = []model.InboxEventType{
 	// Idempotent + commutes with add/remove (a delete just makes it a no-op), so
 	// it rides the concurrent lane.
 	model.InboxMemberJoinedAtRefreshed,
+	// broadcast-worker: room-level mention badge. The destination write is a
+	// guarded idempotent $set, so duplicates and out-of-order applies converge.
+	model.InboxSubscriptionMention,
```
julianshen, line 45: _🏷️ mongodb-consistency_ | _🔴 Blocking_ | _`distributed-system`, `mongodb`, `nats`, `performance-and-concurrency`, `silent-failure`, `tgd-review` (6 rules)_

**The concurrent lane can apply a mention before subscription creation.**

`subscription_mention` and `member_added` use separate consumers, so a mention can reach the destination before the ordered membership event even when `member_added` has the earlier OUTBOX sequence. `SetSubscriptionMentions` then performs an `UpdateMany` matching zero documents, treats that as success, and the inbox message is acknowledged; the later membership insert creates an unbadged subscription, permanently losing the current mention. Put this event in the same ordered FIFO lane as `member_added`/`member_removed`, or introduce a durable pending-badge mechanism that is reconciled when the subscription is created.

<details>

- **`silent-failure`**:45: `subscription_mention` does not commute with `member_added`. Even when `member_added` has the earlier OUTBOX sequence, the separate concurrent consumer can deliver the mention first. The destination's `UpdateMany` then matches zero subscriptions, but its result is ignored and the handler returns success, so the event is ACKed. When the ordered lane later creates the subscription, `hasMention` remains false and the badge is permanently lost. Move this event type to `OrderedEventTypes` so it shares the FIFO lane with membership creation, and update the consumer tests accordingly.
- **`nats`**:45: `subscription_mention` is consumed independently from the ordered `member_added` lane. If a remote member is added and then immediately mentioned, the concurrent consumer can apply and acknowledge the mention first; `SetSubscriptionMentions` matches no subscription and returns success. The ordered consumer then creates the subscription without the badge, permanently losing this mention. Put this event in the same per-destination ordered lane as membership events, or provide a durable mechanism that defers a no-match until the preceding add is applied.
- **`distributed-system`**:45: The new event uses a concurrent consumer separate from the ordered `member_added` lane. A concrete interleaving is: `member_added` is sequence N but has not yet been applied; `subscription_mention` is sequence N+1 and its concurrent consumer applies first; `UpdateMany` matches no subscription and returns success; inbox-worker ACKs the mention; the ordered consumer then creates the subscription without the badge. Move `InboxSubscriptionMention` into `OrderedEventTypes` with membership events so an earlier member creation is applied first.
- **`tgd-review`**:45: `subscription_mention` is consumed independently from the ordered `member_added` lane, so it can reach the destination before that user's subscription exists. `SetSubscriptionMentions` then matches zero rows and returns success, permanently losing the badge for this mention. This is a normal race when a newly added remote member is mentioned soon after joining. Put this event in the same ordered per-destination lane as `member_added` (and update the lane tests), or otherwise retry when the prerequisite subscription has not arrived.
- **`performance-and-concurrency`**:45: A `member_added` event may already be queued in the ordered lane while this mention is fetched concurrently and reaches the destination first. `SetSubscriptionMentions` then matches no subscription and the event is acknowledged permanently; when `member_added` later creates the subscription, the current mention badge is lost. Route mention events through ordering shared with membership creation, or otherwise preserve a causally pending badge until the corresponding subscription exists.

</details>

<details>

```
In `pkg/outbox/outbox.go` around line 45: `subscription_mention` and `member_added` use separate consumers, so a mention can reach the destination before the ordered membership event even when `member_added` has the earlier OUTBOX sequence. `SetSubscriptionMentions` then performs an `UpdateMany` matching zero documents, treats that as success, and the inbox message is acknowledged; the later membership insert creates an unbadged subscription, permanently losing the current mention. Put this event in the same ordered FIFO lane as `member_added`/`member_removed`, or introduce a durable pending-badge mechanism that is reconciled when the subscription is created.

```

</details>

broadcast-worker/handler.go
```diff
@@ -234,16 +253,22 @@ func (h *Handler) handleCreated(ctx context.Context, evt *model.MessageEvent) er
...
+			return err
+		}
 	default:
 		slog.WarnContext(ctx, "unknown room type, skipping fan-out",
 			"type", meta.Type,
 			"room_id", meta.ID,
 			"request_id", natsutil.RequestIDFromContext(ctx))
 		return nil
 	}
+	h.federateMentions(ctx, meta.ID, msg.ID, resolved.Participants, msg.CreatedAt)
```
julianshen, line 281: _🏷️ consistency_ | _🔴 Blocking_ | _`distributed-system`, `mongodb` (2 rules)_

**OUTBOX enqueue failures permanently lose remote mention badges.**

The created-message path broadcasts to clients before calling `federateMentions`, whose publish errors are only logged; the edit path has the same ordering. If local JetStream is unavailable or times out, `HandleMessage` still returns nil, so the canonical event is acknowledged without an OUTBOX event and the destination site's `subscriptions.hasMention` update never occurs. Enqueue the stable-ID OUTBOX events before client publication and propagate enqueue failures so canonical redelivery retries them; successful destinations will deduplicate on retry.

<details>

- **`distributed-system`**:445: If the local JetStream publish fails after the Mongo badge and client broadcast succeed, this branch only logs the error. The handler then returns nil, so the canonical message is ACKed even though no durable OUTBOX record exists and outbox-worker has nothing to retry. Partial per-site success is also possible. Propagate enqueue failures and perform the OUTBOX publishes before the client broadcast; canonical redelivery can then retry safely using the stable message IDs and idempotent destination write.

</details>

<details>

```
In `broadcast-worker/handler.go` around line 270: The created-message path broadcasts to clients before calling `federateMentions`, whose publish errors are only logged; the edit path has the same ordering. If local JetStream is unavailable or times out, `HandleMessage` still returns nil, so the canonical event is acknowledged without an OUTBOX event and the destination site's `subscriptions.hasMention` update never occurs. Enqueue the stable-ID OUTBOX events before client publication and propagate enqueue failures so canonical redelivery retries them; successful destinations will deduplicate on retry.

```

</details>

broadcast-worker/handler.go
```diff
@@ -347,19 +378,77 @@ func (h *Handler) handleUpdated(ctx context.Context, evt *model.MessageEvent) er
...
+			slog.ErrorContext(ctx, "marshal subscription_mention failed",
+				"error", err, "room_id", roomID, "dest_site", destSiteID,
+				"request_id", natsutil.RequestIDFromContext(ctx))
+			continue
+		}
+		// at keeps a later edit's badge from being deduped against the original send.
+		dedupID := fmt.Sprintf("mention:%s:%s:%d:%s", roomID, msgID, at.UnixMilli(), destSiteID)
+		if err := outbox.Publish(ctx, h.publish, h.siteID, roomID, destSiteID,
+			model.InboxSubscriptionMention, payload, dedupID, now); err != nil {
+			slog.ErrorContext(ctx, "federate subscription_mention failed",
```
julianshen: _🏷️ nats-delivery_ | _🔴 Blocking_ | _`nats`_

**OUTBOX publish failures are acknowledged as successful processing.**

When the local OUTBOX publish fails due to a missing stream, timeout, or JetStream transition, `federateMentions` only logs the error. Both created and edited message handlers then return nil, allowing the canonical message to be acknowledged even though no durable relay exists. The remote badge is therefore permanently lost, defeating the OUTBOX retry guarantee. Return and propagate publish failures, handling partial destination success through the stable message IDs and idempotent destination write. Enqueue before the core-NATS client fan-out, or otherwise persist retry state, so a retry does not require acknowledging an undurable relay.

<details>

```
In `broadcast-worker/handler.go` around line 447: When the local OUTBOX publish fails due to a missing stream, timeout, or JetStream transition, `federateMentions` only logs the error. Both created and edited message handlers then return nil, allowing the canonical message to be acknowledged even though no durable relay exists. The remote badge is therefore permanently lost, defeating the OUTBOX retry guarantee. Return and propagate publish failures, handling partial destination success through the stable message IDs and idempotent destination write. Enqueue before the core-NATS client fan-out, or otherwise persist retry state, so a retry does not require acknowledging an undurable relay.

```

</details>

broadcast-worker/handler.go
```diff
@@ -347,19 +378,77 @@ func (h *Handler) handleUpdated(ctx context.Context, evt *model.MessageEvent) er
...
+			Timestamp:   now,
+		})
+		if err != nil {
+			slog.ErrorContext(ctx, "marshal subscription_mention failed",
+				"error", err, "room_id", roomID, "dest_site", destSiteID,
+				"request_id", natsutil.RequestIDFromContext(ctx))
+			continue
+		}
+		// at keeps a later edit's badge from being deduped against the original send.
+		dedupID := fmt.Sprintf("mention:%s:%s:%d:%s", roomID, msgID, at.UnixMilli(), destSiteID)
```
julianshen: _🏷️ idempotency_ | _🟠 Warning_ | _`distributed-system`_

**Millisecond timestamps can deduplicate distinct edits.**

The deduplication key identifies an edit only by `at.UnixMilli()`. Two edits to the same message and destination within one millisecond, or a creation and immediate edit sharing that millisecond, produce the same NATS message ID. JetStream then accepts the later publish as a duplicate while discarding its different account payload, so newly mentioned users miss their badges. Include a stable edit revision/event ID, or a canonical hash of the destination account set, in the deduplication key.

<details>

```
In `broadcast-worker/handler.go` around line 444: The deduplication key identifies an edit only by `at.UnixMilli()`. Two edits to the same message and destination within one millisecond, or a creation and immediate edit sharing that millisecond, produce the same NATS message ID. JetStream then accepts the later publish as a duplicate while discarding its different account payload, so newly mentioned users miss their badges. Include a stable edit revision/event ID, or a canonical hash of the destination account set, in the deduplication key.

```

</details>

broadcast-worker/main.go
```diff
@@ -224,9 +224,22 @@ func main() {
...
+			return fmt.Errorf("publish jetstream message to %s with msgID %s: %w", subj, msgID, err)
+		}
+		return nil
+	}
+
 	parentFetcher := newHistoryParentFetcher(nc, publishMetrics)
 	handler := NewHandler(coalescer, us, publisher, keyProvider, parentFetcher, cfg.Encryption.Enabled, roomRouteMode,
-		withBroadcastMetrics(domainMetrics), withThreadViewSubject(cfg.ThreadViewSubjectEnabled))
+		withBroadcastMetrics(domainMetrics), withOutboxFederation(cfg.SiteID, outboxPublish),
+		withThreadViewSubject(cfg.ThreadViewSubjectEnabled))
```
julianshen, line 266: _🏷️ compatibility_ | _🟠 Warning_ | _`distributed-system`_

**Mixed-version inbox workers can ACK-drop the new event.**

Federation is enabled unconditionally when a new broadcast-worker starts. During a rolling or cross-site deployment, outbox-worker can forward `subscription_mention` to a destination still running an old inbox-worker; that version falls through the existing unknown-event branch and returns nil, causing the delivery to be ACKed and permanently lost. Gate production of this event behind a default-off rollout flag and enable it only after all destination inbox fleets understand the type, or otherwise provide a versioned consumer/deployment protocol that prevents old workers from consuming it.

<details>

```
In `broadcast-worker/main.go` around line 242: Federation is enabled unconditionally when a new broadcast-worker starts. During a rolling or cross-site deployment, outbox-worker can forward `subscription_mention` to a destination still running an old inbox-worker; that version falls through the existing unknown-event branch and returns nil, causing the delivery to be ACKed and permanently lost. Gate production of this event behind a default-off rollout flag and enable it only after all destination inbox fleets understand the type, or otherwise provide a versioned consumer/deployment protocol that prevents old workers from consuming it.

```

</details>

inbox-worker/main.go
```diff
@@ -462,6 +462,27 @@ func (s *mongoInboxStore) UpdateSubscriptionRead(ctx context.Context, roomID, ac
...
+// guard is $not/$gte rather than $lt so a never-read subscription (missing
+// lastSeenAt) still matches — plain $lt would skip the missing field (#467).
+func (s *mongoInboxStore) SetSubscriptionMentions(ctx context.Context, roomID string, accounts []string, msgCreatedAt time.Time) error {
+	_, err := s.subCol.UpdateMany(ctx,
+		bson.M{
+			"roomId":     roomID,
+			"u.account":  bson.M{"$in": accounts},
+			"lastSeenAt": bson.M{"$not": bson.M{"$gte": msgCreatedAt}},
+		},
+		bson.M{"$set": bson.M{"hasMention": true}},
```
julianshen, line 543: _🎯 concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Mention and read updates can lose a newer unread badge**

This update records only `hasMention=true`; `MentionedAt` is used in the filter but is not persisted. A read can compute an older read timestamp, then race with a newer mention: the mention executes first and sets the badge, after which the delayed read advances `lastSeenAt` to its older timestamp and clears `hasMention`. The final state incorrectly hides an unread mention. Persist the latest mention timestamp and make the read update clear the badge only when its read timestamp is at least that timestamp, using an atomic conditional update.

<details>

```
In `inbox-worker/main.go` around line 480: This update records only `hasMention=true`; `MentionedAt` is used in the filter but is not persisted. A read can compute an older read timestamp, then race with a newer mention: the mention executes first and sets the badge, after which the delayed read advances `lastSeenAt` to its older timestamp and clears `hasMention`. The final state incorrectly hides an unread mention. Persist the latest mention timestamp and make the read update clear the badge only when its read timestamp is at least that timestamp, using an atomic conditional update.

```

</details>

broadcast-worker/main.go
```diff
@@ -224,9 +224,22 @@ func main() {
 		slog.Info("room-key cache enabled", "size", cfg.RoomKeyCacheSize, "ttl", cfg.RoomKeyCacheTTL)
 	}
 
+	// JetStream publish for the OUTBOX mention relay; the client fan-out stays on
+	// core NATS (publisher above).
+	outboxPublish := func(ctx context.Context, subj string, data []byte, msgID string) error {
+		_, err := js.PublishMsg(ctx, natsutil.NewMsg(ctx, subj, data), jetstream.WithMsgID(msgID))
```
julianshen, line 254: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Per-destination publishes serialize the broadcast worker**

`federateMentions` invokes this synchronous `PublishMsg` sequentially for every destination site. A message mentioning users across D sites therefore waits for D JetStream acknowledgements; during an OUTBOX leader or NATS impairment, the waits compound and keep the canonical message unacknowledged, reducing worker throughput and potentially causing redelivery and duplicate client broadcasts. Use bounded concurrent or bounded-pending asynchronous publishes under one short overall deadline so latency is capped without creating unbounded goroutines.

<details>

```
In `broadcast-worker/main.go` around line 230: `federateMentions` invokes this synchronous `PublishMsg` sequentially for every destination site. A message mentioning users across D sites therefore waits for D JetStream acknowledgements; during an OUTBOX leader or NATS impairment, the waits compound and keep the canonical message unacknowledged, reducing worker throughput and potentially causing redelivery and duplicate client broadcasts. Use bounded concurrent or bounded-pending asynchronous publishes under one short overall deadline so latency is capped without creating unbounded goroutines.

```

</details>

julianshen (commented):
tGD inline review

broadcast-worker/handler.go
mliu33: Swallowing the OUTBOX enqueue error is deliberate, not an oversight — the ordering was chosen so a failed relay can't block or duplicate the client fan-out.

Publishing before the broadcast and propagating the error means a canonical redelivery re-runs `publishChannelEvent`/`publishMutation`, so every client in the room sees the message twice to save a badge on one remote user. We took the opposite trade: the message is never duplicated, and a mention badge is best-effort at the enqueue boundary. It's called out under "Design decision worth a reviewer's eye" in the PR body.

You're right that this narrows what OUTBOX buys us: the durable hop is protected, the hop *into* the buffer isn't. We're accepting that for now and will revisit it as part of a broader broadcast-worker refactor rather than reshaping the publish ordering in this PR.

Two things that reduce the exposure in the meantime: the publish only happens for messages that actually mention a remote-homed user, and a failure is logged with the room and destination site, so the gap is visible rather than silent.

---

pkg/outbox/outbox.go
mliu33: Same decision as the enqueue-failure thread — the swallow is deliberate so a failed relay can't NAK the canonical message and re-broadcast it to every client in the room. Details in my reply on `handler.go:270`; this is on the list for the broadcast-worker refactor rather than this PR.

---

broadcast-worker/handler.go
mliu33: Same decision as the enqueue-failure thread on `handler.go:270` — the swallow is deliberate so a failed relay can't NAK the canonical message and re-broadcast it to every client in the room. Rationale and the accepted trade-off are in that reply; revisiting it belongs to the broadcast-worker refactor rather than this PR.

---

pkg/outbox/outbox.go
mliu33: Correction: my previous reply on this thread was meant for the `handler.go:447` durability finding and was posted here by mistake. Please disregard it — it does not answer this finding. Sorry for the noise.

On the actual finding: agreed, and it is the most serious thing in this review. My design doc explicitly waved this away ("a mention arriving before its `member_added` matches no subscription and is a silent no-op"), which was wrong — "add a remote user to a channel, then @-mention them in the welcome message" is a normal flow, not a corner case, and the badge is lost permanently.

I don't think moving the type into `OrderedEventTypes` is the right fix, though. That lane is `MaxAckPending=1` per destination, and mentions are far higher volume than membership changes, so every mention badge would queue behind a single in-flight FIFO slot per peer — and a down peer parks the lane.

`inbox-worker` already has an idiomatic answer to exactly this race: `naksIfSubscriptionMissing` (`main.go:128`), used by `UpdateSubscriptionMute`/`Favorite`/`Open`/`Roles`, whose comments say "a genuinely missing sub returns an error (Nak) so the event redelivers until `member_added` lands". `SetSubscriptionMentions` needs the same treatment, with the one wrinkle that a zero `MatchedCount` here is ambiguous: it means either "no subscription yet" (retry) or "the account already read past the message" (correct no-op). `subscriptionExists` disambiguates that, same as the existing call sites.

I'm raising the approach with the PR owner before implementing, since it is a behavioral change beyond the fixes just pushed. Not resolving this thread.

---

inbox-worker/main.go
mliu33: Real, but pre-existing rather than introduced here, so out of scope for this PR.

`hasMention` has always been a bare boolean with no companion timestamp: `broadcast-worker` sets it on the origin side and a read clears it, with no ordering guard between the two. The same delayed-read-clears-newer-mention interleaving is reachable today on a single site, without any federation involved. This PR replicates the existing write to the mentionee's home site; it doesn't change the read/write protocol.

Fixing it properly means persisting a `mentionedAt` on the subscription and making every read-clear conditional on it — which touches `room-service`'s read path, `user-service`, and the local origin write as well as this one, and needs a migration story for existing docs. That is its own change, and doing it only on the federated path would leave the local path inconsistent with it.

Filing as follow-up rather than expanding this PR.

---

broadcast-worker/main.go
mliu33: Fixed in `882900a`. Added `MENTION_FEDERATION_ENABLED`, default-off:

```go
// MentionFederationEnabled: gates cross-site mention badges. Off until every
// destination site's inbox-worker knows subscription_mention — an older one
// Acks it as unknown and drops it.
MentionFederationEnabled bool `env:"MENTION_FEDERATION_ENABLED" envDefault:"false"`
```

`withOutboxFederation` is only wired when it's set, so a default deploy produces no `subscription_mention` at all. Both dev compose stacks and the two-site test stack set it on; a config test pins the default.

This also completes the deploy note in the PR body, which only covered `outbox-worker` (its consumer filters are built from `ConcurrentEventTypes` at startup). You're right that the destination `inbox-worker` fleet is the other half — and the worse half, since a stale `outbox-worker` merely parks the event while a stale `inbox-worker` Acks and discards it.

---

broadcast-worker/handler.go
mliu33: Fixed in `882900a`. A sorted fnv-1a digest of the destination's account set now participates in the key:

```go
dedupID := fmt.Sprintf("mention:%s:%s:%d:%s:%s",
    roomID, msgID, at.UnixMilli(), accountsDigest(accounts), destSiteID)
```

`accountsDigest` clones before sorting (the slice feeds the published payload) and writes a `0x00` separator so `("ab","c")` can't collide with `("a","bc")`. A unit test pins order-independence, set-sensitivity, and non-mutation.

One deliberate choice worth noting: I kept the key payload-derived rather than switching to `natsutil.InboxDedupID`, which is what the other federating services use. That helper bases the ID on `X-Request-ID`, and this consume path stamps with `StampRequestID` — the lenient variant that **mints a fresh UUID** when the header is absent. `RequireRequestID`'s own doc says paths deriving `Nats-Msg-Id` components must use the strict variant, precisely because minting breaks dedup. A request-ID-based key would therefore differ per redelivery here, whereas `msgID` + timestamp + digest is stable unconditionally.

---

broadcast-worker/main.go
mliu33: Fixed in `882900a`. Per-destination publishes now run concurrently under a bounded semaphore, matching the existing per-site fan-outs in `user-service/service/subscriptions.go:270` and `search-service/enrich.go:163`:

```go
const maxSiteFanout = 8
...
var wg sync.WaitGroup
sem := make(chan struct{}, maxSiteFanout)
for destSiteID, accounts := range accountsBySite {
    ...
    wg.Add(1)
    sem <- struct{}{}   // acquire before spawning: bounds goroutine COUNT, not just concurrency
    go func() { defer wg.Done(); defer func() { <-sem }(); ... }()
}
wg.Wait()
```

Same const value and the same acquire-before-spawn ordering, so a wide federation can't spawn one parked goroutine per site. Worst case is now `ceil(N/8)` acks instead of `N`.

I deliberately did not add a `ctx.Err()` re-check after parking on the semaphore, even though `user-service` has one: there each site is a request-scoped RPC that's worthless once the client is gone, whereas here bailing on cancellation would drop a badge the local Mongo write already committed to. `search-service`'s fan-out has no re-check either. No extra deadline: `js.PublishMsg` carries JetStream's own PubAckWait.

---

pkg/outbox/outbox.go
mliu33: Owner decision: **accepting this race with the current Ack behavior.** Not fixing it in this PR.

The finding stands as described — a mention on the concurrent lane can reach the destination before the `member_added` on the ordered lane, match nothing, Ack, and lose that badge. What tipped it was the cost of both remedies:

- **`OrderedEventTypes`** puts high-volume mention badges behind the per-destination `MaxAckPending=1` FIFO shared with membership events, so one unreachable peer parks every pending mention to that site.
- **NAK on zero-match** (via `subscriptionExists`, as `UpdateSubscriptionMute`/`Favorite`/`Open`/`Roles` do) fixes the race, but then every mention of a non-member — a typo'd `@name`, a user removed between send and delivery — burns the full `MaxDeliver` chain before dead-lettering. That's a common case paying for a rare one.

Blast radius when it does hit: the affected user still receives the message; only that one badge is missed, and any subsequent mention badges normally.

Recorded as an explicit "Accepted limitation" section in `docs/superpowers/specs/2026-08-22-subscription-mention-fanout-design.md` (`ec1cf18`) — including the two rejected alternatives — so the decision is discoverable in the repo rather than only in this thread. Worth revisiting if it shows up in practice.

---

broadcast-worker/main.go
mliu33: Update — reversing course on this one. The owner has decided against the feature flag, so `MENTION_FEDERATION_ENABLED` is removed in `ec1cf18` and federation is unconditional again.

The hazard you identified is real and unchanged: a destination site still running an `inbox-worker` that predates `subscription_mention` will hit its unknown-event branch, return nil, and Ack — discarding the badge silently. Removing the gate means that's now prevented **operationally rather than in code**, by deploy order:

1. `inbox-worker` on every destination site
2. `outbox-worker` (its per-peer consumer filters are built from `ConcurrentEventTypes` at startup)
3. `broadcast-worker` — starts producing immediately on rollout

I'm making that sequence explicit in the PR description rather than leaving it as a code comment, since it's the only safeguard now.

---

inbox-worker/handler.go
```diff
@@ -177,6 +181,8 @@ func (h *Handler) HandleEvent(ctx context.Context, data []byte) error {
 		return h.handleUserChatlistUpdated(ctx, &evt)
 	case model.InboxSubscriptionSectionMoved:
 		return h.handleSubscriptionSectionMoved(ctx, &evt)
+	case model.InboxSubscriptionMention:
```
julianshen, line 243: _🏷️ protocol-compatibility_ | _🔴 Blocking_ | _`distributed-system`_

**Old inbox workers ACK-drop the new event during rolling deploys.**

The producer is enabled unconditionally, but an older destination inbox-worker falls through its unknown-event default and returns nil. During a mixed-version rollout, a new origin can therefore emit `subscription_mention`, outbox-worker can forward it, and the old destination consumer acknowledges it permanently before gaining this case. The user's badge is then unrecoverable. Gate publishing until inbox-worker support is deployed to every site, or use a rollout-safe protocol where unsupported events remain replayable.

<details>

```
In `inbox-worker/handler.go` around line 184: The producer is enabled unconditionally, but an older destination inbox-worker falls through its unknown-event default and returns nil. During a mixed-version rollout, a new origin can therefore emit `subscription_mention`, outbox-worker can forward it, and the old destination consumer acknowledges it permanently before gaining this case. The user's badge is then unrecoverable. Gate publishing until inbox-worker support is deployed to every site, or use a rollout-safe protocol where unsupported events remain replayable.

```

</details>

broadcast-worker/handler.go
```diff
@@ -347,19 +385,104 @@ func (h *Handler) handleUpdated(ctx context.Context, evt *model.MessageEvent) er
...
+				"error", err, "room_id", roomID, "dest_site", destSiteID,
+				"request_id", natsutil.RequestIDFromContext(ctx))
+			continue
+		}
+		// Payload-derived, so it is redelivery-stable even though this consume path
+		// mints a request ID when the header is absent (StampRequestID, not Require).
+		// at separates an edit from the send; the digest separates same-ms edits.
+		dedupID := fmt.Sprintf("mention:%s:%s:%d:%s:%s", roomID, msgID, at.UnixMilli(), accountsDigest(accounts), destSiteID)
+		wg.Add(1)
+		sem <- struct{}{}
```
julianshen: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Fan-out latency remains uncapped across destination batches.**

The semaphore limits concurrency to eight, but the handler still synchronously waits for every batch before acknowledging the canonical message. With D remote sites and an impaired OUTBOX stream, latency grows as `ceil(D/8)` JetStream publish timeouts; repeated remote-mention messages can therefore collapse worker throughput, and sufficiently wide fan-out can exceed the consumer AckWait and trigger duplicate client broadcasts. Use one short overall fan-out deadline, including cancellation-aware semaphore acquisition, or bounded-pending asynchronous publication so handler latency does not scale with the number of timeout batches.

<details>

```
In `broadcast-worker/handler.go` around line 459: The semaphore limits concurrency to eight, but the handler still synchronously waits for every batch before acknowledging the canonical message. With D remote sites and an impaired OUTBOX stream, latency grows as `ceil(D/8)` JetStream publish timeouts; repeated remote-mention messages can therefore collapse worker throughput, and sufficiently wide fan-out can exceed the consumer AckWait and trigger duplicate client broadcasts. Use one short overall fan-out deadline, including cancellation-aware semaphore acquisition, or bounded-pending asynchronous publication so handler latency does not scale with the number of timeout batches.

```

</details>

julianshen (commented):
tGD inline review

broadcast-worker/main.go
```diff
@@ -224,9 +224,22 @@ func main() {
...
+		destination, operation := natsmetrics.PublishLabelsFromSubject(subj)
+		publishMetrics.Attempt(ctx, destination, operation, err)
```
mliu33: Remove this metrics. Only track failure count when failing to publish

broadcast-worker/handler.go
```diff
@@ -347,19 +385,104 @@ func (h *Handler) handleUpdated(ctx context.Context, evt *model.MessageEvent) er
...
+			}
+		}()
 	}
-	return h.store.SetSubscriptionMentions(ctx, roomID, parsed.Accounts, *msg.EditedAt)
+	wg.Wait()
+}
+
+// accountsDigest is a stable fingerprint of a destination's account set, sorted so
+// the same set always yields the same digest regardless of mention order.
+func accountsDigest(accounts []string) string {
```
mliu33: Don't calculate this for dedupeId, just use simple requestID

mliu33: Done in `093e315`. `accountsDigest` is deleted; the key is now the standard helper:

```go
seed := fmt.Sprintf("%s:%s:%d", roomID, msgID, at.UnixMilli())
dedupID := natsutil.InboxDedupID(ctx, destSiteID, seed)
```

I need to correct what I argued on this PR earlier. I declined a request-ID key on the grounds that this consume path uses `StampRequestID`, which mints a fresh UUID when the header is absent, so the ID wouldn't survive a redelivery. That reasoning was wrong for this path: `message-gatekeeper` rejects any submission whose payload `requestId` isn't a valid UUID (`handler.go:302`), stamps it onto ctx, and publishes canonical through `natsutil.NewMsg` — so the canonical event always carries `X-Request-ID` and every redelivery re-reads the same value. The mint branch is unreachable here.

It also subsumes the millisecond-collision finding the digest was added for: an edit is its own canonical event with its own request ID, so two same-ms edits get different keys without hashing anything.

Tests now stamp a request ID on the handler context the way the consume loop does, and assert the `{requestID}:{destSiteID}` form; the integration test pins it end-to-end through OUTBOX.

---

broadcast-worker/main.go
mliu33: Done in `093e315` — the metric is now recorded only on a failed publish:

```go
_, err := js.PublishMsg(ctx, natsutil.NewMsg(ctx, subj, data), jetstream.WithMsgID(msgID))
if err != nil {
    destination, operation := natsmetrics.PublishLabelsFromSubject(subj)
    publishMetrics.Attempt(ctx, destination, operation, err)
    return fmt.Errorf("publish jetstream message to %s with msgID %s: %w", subj, msgID, err)
}
return nil
```

`Attempt` already carries an `outcome` label, so this drops the `outcome="success"` series for this publisher and keeps the failure counts.

---

inbox-worker/handler.go
mliu33: Same finding as `broadcast-worker/main.go:242`, answered there — the short version:

A flag was added for exactly this (`MENTION_FEDERATION_ENABLED`, default-off) and then removed at the owner's direction in `ec1cf18`. The hazard you describe is unchanged and real; it is now prevented operationally rather than in code, by deploy order:

1. `inbox-worker` on every destination site
2. `outbox-worker` — its per-peer consumer filters are built from `ConcurrentEventTypes` at startup
3. `broadcast-worker` — starts producing immediately on rollout

That sequence is now a required section in the PR description, since it's the only safeguard.

---

broadcast-worker/handler.go
```diff
@@ -347,19 +382,90 @@ func (h *Handler) handleUpdated(ctx context.Context, evt *model.MessageEvent) er
...
+		}
+		// at separates an edit from the send; an edit is its own canonical event
+		// carrying its own request ID, so two same-ms edits can't collide.
+		seed := fmt.Sprintf("%s:%s:%d", roomID, msgID, at.UnixMilli())
+		dedupID := natsutil.InboxDedupID(ctx, destSiteID, seed)
+		wg.Add(1)
+		sem <- struct{}{}
+		go func() {
+			defer wg.Done()
+			defer func() { <-sem }()
```
julianshen, line 480: _🏷️ reliability_ | _🔴 Blocking_ | _`distributed-system`, `nats` (2 rules)_

**OUTBOX enqueue failures are reported as successful processing.**

When publishing `chat.outbox.{siteID}.{destSiteID}.subscription_mention` fails because JetStream is unavailable or times out, this branch only logs the error. `federateMentions` then returns normally, so the canonical message is acknowledged although no durable OUTBOX record exists and `outbox-worker` has nothing to retry. The remote badge is permanently lost. Make enqueue failure part of a retry-safe boundary: publish before the non-idempotent client fan-out and propagate any failure, using the existing message ID to deduplicate partial retries, or persist the pending relay in a durable local outbox.

<details>

- **`distributed-system`**:459: The local badge and client broadcast have already succeeded when `outbox.Publish` runs. If JetStream rejects or times out this publish, the error is only logged; `federateMentions` returns and the canonical message is acknowledged, leaving no durable relay to retry. The destination badge therefore remains false after a transient origin-side NATS failure. Make relay creation retryable without replaying the client side effect—for example, enqueue all destination events before client publication, propagate failures, and rely on the stable message IDs to deduplicate destinations already accepted.

</details>

<details>

```
In `broadcast-worker/handler.go` around line 459: When publishing `chat.outbox.{siteID}.{destSiteID}.subscription_mention` fails because JetStream is unavailable or times out, this branch only logs the error. `federateMentions` then returns normally, so the canonical message is acknowledged although no durable OUTBOX record exists and `outbox-worker` has nothing to retry. The remote badge is permanently lost. Make enqueue failure part of a retry-safe boundary: publish before the non-idempotent client fan-out and propagate any failure, using the existing message ID to deduplicate partial retries, or persist the pending relay in a durable local outbox.

```

</details>

pkg/outbox/outbox.go
julianshen, line 45: _🏷️ ordering_ | _🔴 Blocking_ | _`nats`_

**Mention updates do not commute with an older read update.**

The concurrent-lane claim is unsafe because `SetSubscriptionMentions` stores only `hasMention=true`, while reads clear that boolean without any persisted mention timestamp. A reachable interleaving is: the subscription has `lastSeenAt=T0`; a read at T1 is delayed; a later mention at T2 is delivered first and sets the badge; then the T1 read advances `lastSeenAt` and clears the badge even though T2 remains unread. Persist the latest mention timestamp and clear only when the read timestamp covers it, or place all relevant read and mention mutations on one ordering lane. Add an out-of-order read/mention test.

<details>

```
In `pkg/outbox/outbox.go` around line 45: The concurrent-lane claim is unsafe because `SetSubscriptionMentions` stores only `hasMention=true`, while reads clear that boolean without any persisted mention timestamp. A reachable interleaving is: the subscription has `lastSeenAt=T0`; a read at T1 is delayed; a later mention at T2 is delivered first and sets the badge; then the T1 read advances `lastSeenAt` and clears the badge even though T2 remains unread. Persist the latest mention timestamp and clear only when the read timestamp covers it, or place all relevant read and mention mutations on one ordering lane. Add an out-of-order read/mention test.

```

</details>

inbox-worker/main.go
```diff
@@ -462,6 +462,27 @@ func (s *mongoInboxStore) UpdateSubscriptionRead(ctx context.Context, roomID, ac
...
+	// SetSubscriptionMentions filters on (roomId, u.account); without this index
+	// the federated badge write collscans the shared subscriptions collection.
+	mongoutil.WarnMissingIndexes(ctx, s.subCol, "roomId_1_u.account_1")
+}
+
+// SetSubscriptionMentions flags the accounts' subscriptions as mentioned. The
+// guard is $not/$gte rather than $lt so a never-read subscription (missing
+// lastSeenAt) still matches — plain $lt would skip the missing field (#467).
+func (s *mongoInboxStore) SetSubscriptionMentions(ctx context.Context, roomID string, accounts []string, msgCreatedAt time.Time) error {
+	_, err := s.subCol.UpdateMany(ctx,
```
julianshen, line 537: _🏷️ ordering_ | _🟠 Warning_ | _`distributed-system`_

**Concurrent mentions can overtake subscription creation and be lost.**

`subscription_mention` uses a different concurrent OUTBOX lane from ordered `member_added`. If the mention reaches the destination first, `UpdateMany` matches zero subscriptions, discards the result, and returns success, so inbox-worker acknowledges the event. A later `member_added` then creates an unbadged subscription and the first mention is unrecoverable. Preserve the unmatched event for reconciliation when membership arrives, or serialize this event with membership creation rather than treating a zero match as successfully processed.

<details>

```
In `inbox-worker/main.go` around line 474: `subscription_mention` uses a different concurrent OUTBOX lane from ordered `member_added`. If the mention reaches the destination first, `UpdateMany` matches zero subscriptions, discards the result, and returns success, so inbox-worker acknowledges the event. A later `member_added` then creates an unbadged subscription and the first mention is unrecoverable. Preserve the unmatched event for reconciliation when membership arrives, or serialize this event with membership creation rather than treating a zero match as successfully processed.

```

</details>

broadcast-worker/handler.go
```diff
@@ -347,19 +382,90 @@ func (h *Handler) handleUpdated(ctx context.Context, evt *model.MessageEvent) er
...
+			if err := outbox.Publish(ctx, h.publish, h.siteID, roomID, destSiteID,
+				model.InboxSubscriptionMention, payload, dedupID, now); err != nil {
+				slog.ErrorContext(ctx, "federate subscription_mention failed",
+					"error", err, "room_id", roomID, "dest_site", destSiteID, "accounts", len(accounts),
+					"request_id", natsutil.RequestIDFromContext(ctx))
+			}
+		}()
+	}
```
julianshen, line 488: _🏷️ backpressure_ | _🟠 Warning_ | _`nats`_

**Fan-out wait time scales with destination timeout batches.**

The per-message semaphore limits concurrency to eight, but `wg.Wait()` still blocks canonical-message acknowledgement until every destination publish finishes. With more than eight remote sites and impaired JetStream publishing, processing takes `ceil(destinations/8)` timeout batches. This can exceed the consumer AckWait and redeliver a message whose client broadcast already succeeded, producing duplicate broadcasts and reducing worker throughput. Apply one short overall fan-out deadline, make semaphore acquisition cancellation-aware, or use a globally bounded asynchronous dispatcher.

<details>

```
In `broadcast-worker/handler.go` around line 467: The per-message semaphore limits concurrency to eight, but `wg.Wait()` still blocks canonical-message acknowledgement until every destination publish finishes. With more than eight remote sites and impaired JetStream publishing, processing takes `ceil(destinations/8)` timeout batches. This can exceed the consumer AckWait and redeliver a message whose client broadcast already succeeded, producing duplicate broadcasts and reducing worker throughput. Apply one short overall fan-out deadline, make semaphore acquisition cancellation-aware, or use a globally bounded asynchronous dispatcher.

```

</details>

julianshen (commented):
tGD inline review

inbox-worker/main.go
mliu33: Flagging a factual error in the bot's reply above, so this finding isn't withdrawn on a wrong premise.

It states that `UpdateSubscriptionRead` "updates only `lastSeenAt` and `alert`. It does not clear `Subscription.hasMention`." It does:

```go
// room-service/store_mongo.go:1021
bson.M{"$set": bson.M{"lastSeenAt": lastSeenAt, "alert": false, "hasMention": false}},
```

Its own doc comment says so (`store_mongo.go:1007-1008`, "clearing alert and hasMention"), and `room-service/store_mongo.go:1775` and `:1834` clear it on the bulk paths too. `threadReadUpdate` on `thread_subscriptions` is a *second* clear site, not the only one.

Worse for the finding's validity: that update carries **no `$lt` guard** on `lastSeenAt`, unlike the guarded writes in `inbox-worker`. So a delayed read computed at T1 unconditionally overwrites a mention set at T2 > T1 — exactly the interleaving @julianshen described.

My position doesn't change — it's pre-existing, reachable single-site today without any federation, and fixing it means persisting a `mentionedAt` on the subscription plus making every read-clear conditional on it, across `room-service`, `user-service` and the local origin write, with a migration. Out of scope here. But it should be recorded as a real deferred issue, not dismissed as not applicable.

---

chenjr0719 (approved):
LGTM. The mention-badge federation reads cleanly and the deterministic room-mention id keeps redeliveries idempotent. Understood and on board with the design tradeoff here: the badge is best-effort at the enqueue boundary — prioritizing never double-broadcasting the client message on a redelivery over guaranteeing the badge — with the new-member-mention lane race, the post-fan-out enqueue, and the deploy-order dependency (inbox -> outbox -> broadcast) as conscious accepts. Approving for that contract.

PR #356, opened 2026-08-22
feat(pagefit): trim paginated replies to the payload budget instead of failing

mliu33:
Follow-up to #350. That PR made an oversize reply *report* `response_too_large` instead of timing out; this one makes the page **fit in the first place**, so the frontend stops seeing the error at all on the RPCs in scope.

Today the frontend recovers by halving `size` and retrying — a wasted round trip per halving, on exactly the busiest rooms. `history-service` has no size guard at all: it marshals whatever Cassandra returned and hands it to the transport.

## Scope: only where trimming is provably safe

Three RPCs locate their next page from **returned content**, so a shorter page still resumes exactly where it left off:

| RPC | Service | Next page located by | Trim |
|---|---|---|---|
| Load History | history | client re-asks `before` = oldest returned `createdAt` | drop from tail |
| Thread List | user | composite value cursor `{lastMsgAt, threadRoomId}` | drop from tail |
| Load Surrounding | history | `moreBefore` / `moreAfter`, client pages outward by timestamp | drop from both ends |

**Deliberately excluded:** `LoadNextMessages`, `ListPinnedMessages`, `GetThreadMessages`. Their `nextCursor` is Cassandra's opaque driver `PageState`, which marks the position after the last row *the driver fetched* and cannot be re-derived from row content — trimming without re-deriving it would **silently skip the dropped rows**. They keep #350's fallback, and the frontend's `size/2` retry stays live for them. `search-service` (`offset`/`size`) is out for the same reason.

## Design

**`pkg/pagefit`** owns byte arithmetic only — no knowledge of messages, threads or NATS:

```go
func Fit[T any](items []T, b Budget, envelope int) (kept []T, oversize bool, err error)
func FitWindow[T any](items []T, pivot int, b Budget, envelope int) (lo, hi int, oversize bool, err error)
func Resolve(override, brokerMaxPayload int64, reserve int) Budget
```

`Fit` returns only what the caller cannot derive: "rows were dropped" is `len(kept) < len(items)`, so it is not a return value. A sizing failure is an `error`, not a magic width — an unencodable row surfaces as `internal` rather than being blanked and returned as success.

Budget comes from `nc.NatsConn().MaxPayload()` less a reserve, mirroring `room-service/main.go:318`. `MAX_RESPONSE_BYTES` overrides it, but **only downward** — an override above what the broker accepts would size replies the wire then refuses. `0` derives from the broker, and a non-positive result disables trimming so it can be switched off without a rollback.

## The constraint that shaped the implementation: millisecond ties

Cassandra `TIMESTAMP` is millisecond-precision, and Load History resumes with a **strict** `created_at < before`. A trim landing in the middle of a millisecond therefore excludes every row sharing that timestamp from the next query — silent, permanent data loss on exactly the high-rate rooms trimming exists to serve. Load Surrounding had the same hole at both edges.

The cut now lands on a timestamp boundary. When the kept prefix ends inside a run that continues past it, the cut retreats to the run's start. When the run starts at index 0 — the whole page sits inside one millisecond, so there is nothing to cut — the run is returned intact and its rows are blanked to fit. If even the blanked run overflows, it is still returned whole and the transport's `response_too_large` surfaces: an explicit error beats a silent skip. `docs/client-api.md` documents this as the one case where Load History and Load Surrounding can still return `response_too_large`.

**One ordering constraint worth reviewer attention:** in Thread List, `enrichThreadPage` runs after the count-trim and *adds* bytes (display names, HR records). The byte-trim therefore runs **after** enrichment, with the cursor re-derived from the post-trim last item. Trimming first would overflow again — pinned by a test that only exceeds the budget once enriched.

**Single-row backstop.** A row that alone exceeds the budget is blanked, not dropped or errored, so pagination always advances past it: `msg`, `mentions`, `attachments`, `card`, `cardAction`, `quotedParentMessage`, `reactions`, `sysMsgData` cleared, plus `encPayload`/`encMeta` — leaving the encrypted body would defeat the strip in exactly the E2E rooms most likely to overflow. Identifiers, `sender`, `createdAt` and `type` are kept for placeholder rendering, and `truncated: true` marks the row. The same blanking is applied across a tied timestamp run, which is `truncated`'s second cause.

`type` is deliberately retained, unlike `redactUnavailablePins`, which clears it so a pre-access system message cannot leak event details. That reasoning does not apply here: the caller is authorised to see this row, it is merely too large to ship inside a page.

This is a backstop, not a hot path — `content` caps at 20 KB and attachments at 1 × 8 KB, so a row's realistic ceiling is ~50 KB against a 128 KiB budget; only `reactions` is unbounded.

## Performance

The trim runs on every read, so the common case must stay cheap. Sizing is one counting pass: `sum(rows)` plus separators and brackets **is** the array length, so measuring rows and measuring the whole slice are the same work. One encoder writes into a length counter, so no encoded bytes are ever retained.

| 100 rows × 512 B | ns/op | B/op | allocs/op |
|---|---|---|---|
| page fits (the common case) | 60,894 | 2,505 | 102 |
| page trimmed (rare) | 62,613 | 2,504 | 102 |

An earlier revision short-circuited the fitting case on a whole-slice length check (595 B / 2 allocs) and fell back to per-row sizing only when trimming. Collapsing to the single pass costs ~2.5 KB and 100 allocations on the common path, and measures faster on both (77 µs → 61 µs fitting, 143 µs → 63 µs trimming). Cost is now uniform and the trimmed path is no longer a cliff.

## Testing

TDD throughout — every test watched failing for the right reason before the implementation existed.

- **`pkg/pagefit`** — 98.7% coverage: empty, exact fit, one byte over, all-oversize, window symmetry, pivot at either edge and one past the end, disabled budget, sizing failure, override-cannot-raise, plus two benchmarks.
- **Size arithmetic pinned directly** — per-row widths and the derived total are asserted equal to `json.Marshal` byte-for-byte across HTML metacharacters, unicode, `RawMessage` and nested containers, and the trim result is asserted both within budget and maximal across a sweep of budgets. The arithmetic is derived, not measured, so it gets its own tests.
- **Millisecond ties** — a trim landing inside a run retreats to its boundary; a page wholly inside one millisecond is returned whole and blanked; the oversize blanking cannot be skipped when the tie path fires.
- **Pagination continuity** (the test that guards the actual risk) — trim page 1, feed its own `before` back in, assert page 2 resumes at the first dropped row and every row appears exactly once.
- **End-to-end** against real Cassandra: seed 40 rows, walk the room across multiple trimmed pages, assert the union equals the seeded set with each ID appearing exactly once. A second integration test seeds rows sharing a millisecond and walks the same assertion.
- **Thread List** — cursor derived from the last *kept* item, the enrichment-ordering test above, and an oversize row blanked rather than passed through.

**Gates:** `make lint` 0 issues · `make test` `-race`, 0 failures · `make test-integration` for `pkg/pagefit`, `history-service`, `user-service` all ok · `make sast` gosec/govulncheck/semgrep all PASS.

## Defects found and fixed during review

- **Mid-millisecond trims lost rows** (blocking, from review) — the boundary cut above.
- **`MAX_RESPONSE_BYTES` could raise the ceiling** above the broker's `max_payload`, sizing replies the wire would refuse. `Resolve` now only lowers.
- **An unencodable row was charged an oversize width**, so an encoding fault was blanked, marked `truncated`, and returned as success. Sizing failures are errors now. The width sentinel this replaced had its own bug: `MaxInt32` *ties* against a large budget, and a tie read as fitting, so the row was kept and the reply overflowed anyway. Sentinel and bug are both gone.
- **Thread List discarded `Fit`'s `oversize` signal** on the claim it could not happen — untrue with a small configured budget. Such a row now has its bodies dropped and is marked `truncated`, so pagination advances.

## Known limitations

- **`unavailableSites` is outside the envelope estimate.** Thread List charges a fixed 256-byte envelope, which does not include the variable cross-site failure list. With many simultaneously-unreachable peers the response can exceed the budget by the size of that list; the router's `response_too_large` still catches it.
- **Trimming is post-fetch, so discarded rows were still read.** A trimmed page re-reads its dropped rows on the next request — roughly a `limit/kept` multiplier on Cassandra and, for Thread List, on the Mongo read and cross-site fan-out. This is a regression only against a path that had no trimming at all; correctness is unaffected, since every dropped row is re-delivered. An adaptive fetch cap belongs in a follow-up PR with its own benchmarks rather than bolted onto this one, because it reopens the query path this PR deliberately leaves alone.

## Client contract

`truncated` is a new field on a client-facing struct, so `docs/client-api.md` and both derived views are updated in this PR. The documented rules: **`limit` is a maximum, never a guarantee** — the more-flag is authoritative, page until it clears, and never treat a short page as the end; and the one residual `response_too_large` case on Load History / Load Surrounding, where a smaller-`limit` retry is the wrong move because it drops the rest of the tied group.

Spec and plan are committed under `docs/superpowers/`.

---
_Generated by [Claude Code](https://claude.ai/code/session_012vGY47HWtztTUL36YcEHpz)_

## Summary by CodeRabbit

* **New Features**
  * Paginated history, surrounding-message, and thread-list responses now automatically fit configured transport-size limits.
  * Oversized message content is safely blanked and marked as truncated while preserving identifiers and pagination metadata.
  * Responses indicate when results were size-limited, with continuation flags and cursors maintained for reliable pagination.
  * Added configuration to enable or disable page trimming and set response-size limits.

* **Documentation**
  * Updated client API documentation with new truncation, size-limit, short-page, and retry behavior.


julianshen:
**7 findings · 3 unique issues · 3 inline comments posted** — 🟠 7 warning

### Disputed

- `history-service/internal/service/utils.go:221` (`tgd-review`) — `timestampRun` can inspect only the rows already returned by Cassandra. If more than the requested limit share one `createdAt`, the repository page itself ends mid-group; the client then resumes with strict `created_at < before` and permanently skips tied rows beyond that page. The integration test uses only three rows per timestamp, below the limit, so it does not cover this case. Use a complete `(createdAt, messageId)` cursor or make the repository fetch the complete boundary group. The same limitation affects surrounding windows.

### Related work

- [PR #350](https://github.com/hmchangw/newchat/pull/350) — fix\(natsrouter\): reply response\_too\_large instead of timing out on oversize (merged)

### Local review memory

- `history-service/internal/service/messages.go:99` — @mliu33: Confirmed and fixed in d6f789b. You're right, and the schema proves it: messages_by_room is PRIMARY KEY ((room_id, bucket), created_at, message_id) — message_id is there precisely because created_at is not unique — while the resume… ([thread](https://github.com/hmchangw/newchat/pull/356#discussion_r3836735818))
- `docs/client-api.md:6419` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟡 Minor_ _⚡ Quick win_ **Use content for truncated broadcast messages.** Both locations define the broadcast body as content but document msg as the cleared field. - docs/client-api.md#L6419-… ([thread](https://github.com/hmchangw/newchat/pull/356#discussion_r3838022656))
- `docs/client-api.md:6729` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟡 Minor_ _⚡ Quick win_ **Keep the admonition blockquote contiguous.** markdownlint-cli2 reports MD028 for blank lines inside this blockquote at Lines 6728-6729. Remove the blank separator or pref… ([thread](https://github.com/hmchangw/newchat/pull/356#discussion_r3838022659))

<details>

* `docs/client-api.md`
* `docs/client-api/events.md`
* `docs/superpowers/plans/2026-08-22-page-size-budget-trimming.md`
* `docs/superpowers/specs/2026-08-22-page-size-budget-trimming-design.md`
* `history-service/internal/service/pagefit_integration_test.go`
* `history-service/internal/service/pagefit_test.go`
* `history-service/internal/service/pin.go`
* `history-service/internal/service/utils.go`
* `pkg/model/cassandra/message.go`
* `pkg/model/cassandra/message_test.go`
* `pkg/model/threadlist.go`
* `pkg/pagefit/pagefit.go`
* `pkg/pagefit/pagefit_test.go`
* `user-service/main.go`
* `user-service/service/threadfit_test.go`
* `user-service/service/threads.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

history-service/internal/service/messages.go
```diff
@@ -93,12 +96,15 @@ func (s *HistoryService) LoadHistory(c *natsrouter.Context, req models.LoadHisto
 
 	redactUnavailableQuotes(page.Data, accessSince)
 	setDecodedAttachments(c, page.Data)
+	// Trim last: both passes above change encoded size. Rows are DESC, so
```
julianshen, line 99: _🏷️ cassandra-pagination_ | _🔴 Blocking_ | _`cassandra`, `distributed-system`, `performance-and-concurrency`, `tgd-review` (4 rules)_

**Timestamp-only resume can skip trimmed Cassandra rows**

Load History trims an arbitrary prefix and tells clients to resume with the oldest kept row's `createdAt`; the test confirms the next Cassandra query uses a strict `created_at < before` bound. Cassandra timestamps have millisecond precision, so multiple messages can share that value. If the trim boundary splits such a group, every dropped row with the same timestamp as the last kept row is excluded from the next query and is permanently skipped. The surrounding-window trim at line 388 has the same problem at either edge. Use a cursor containing the complete clustering position, such as `(createdAt, messageId)`, and apply a matching tuple bound in the `messages_by_room` queries. Add Cassandra integration coverage with multiple rows sharing one `created_at` and a budget that splits them.

<details>

- **`tgd-review`**:100: The prefix may end between messages with the same `CreatedAt`, but the next request uses only the oldest kept timestamp as `before`. The continuity test confirms the repository then selects rows strictly older via `m.CreatedAt.Before(before)`, so any dropped row sharing that timestamp is permanently skipped. Converting the cursor through `UnixMilli` can similarly lose rows that differ only below millisecond precision. Use a cursor with a stable tie-breaker such as `(createdAt, messageID)`, or otherwise ensure trimming cannot split a timestamp group while preserving forward progress, and add equal-timestamp coverage.
- **`distributed-system`**:99: The trim assumes `createdAt` uniquely identifies the resume position, but the client resumes with only the oldest kept timestamp and the tests model the repository boundary as strictly older. If two messages share a millisecond and trimming keeps one while dropping the other, the next `before` excludes the dropped equal-time message permanently. The same boundary risk applies when trimming either end of Load Surrounding. Use a composite cursor such as `(createdAt, messageId)` with matching tie-break queries, or avoid trimming until the repository can resume within an equal-timestamp group.
- **`cassandra`**:99: `GetMessagesBetweenDesc` still fetches up to the requested limit before `fitPage` discards rows. The next request resumes at the last kept timestamp, so all dropped rows are fetched and decoded again. With `limit=100`, roughly 20 KB messages, and the default approximately 124 KiB budget, only about six rows may be returned while nearly 100 are read on each request; walking history repeatedly reads the same partition rows. If oversized rows are blanked one at a time, the amplification can approach 100 Cassandra row reads per delivered row. Apply a payload-aware cap to the limit passed into the Cassandra repository, or introduce a resumable cursor/fetch strategy that does not discard already-read rows. Add a repository-level test asserting bounded rows read across a trimmed walk.
- **`performance-and-concurrency`**:101: Trimming occurs only after Cassandra has returned and processed the full requested limit, while the next `before` advances only past the kept prefix. If 2 of 100 large rows fit, the next request fetches 98 of those discarded rows again; under a continuing walk, each row can be read, redacted, decoded, and JSON-sized roughly 50 times. This materially amplifies Cassandra and CPU load on the large-room path this change targets. Thread List has the same issue with cross-site fan-out and enrichment. Update pagination clients to carry the effective returned page length into subsequent requests, or add a server-provided effective limit/adaptive fetch cap so dropped rows are not repeatedly fetched at the original limit.

</details>

<details>

```
In `history-service/internal/service/messages.go` around line 99: Load History trims an arbitrary prefix and tells clients to resume with the oldest kept row's `createdAt`; the test confirms the next Cassandra query uses a strict `created_at < before` bound. Cassandra timestamps have millisecond precision, so multiple messages can share that value. If the trim boundary splits such a group, every dropped row with the same timestamp as the last kept row is excluded from the next query and is permanently skipped. The surrounding-window trim at line 388 has the same problem at either edge. Use a cursor containing the complete clustering position, such as `(createdAt, messageId)`, and apply a matching tuple bound in the `messages_by_room` queries. Add Cassandra integration coverage with multiple rows sharing one `created_at` and a budget that splits them.

```

</details>

pkg/pagefit/pagefit.go
```diff
@@ -0,0 +1,166 @@
...
+		return true
+	}
+	// A slice that will not encode can never be said to fit — comparing a
+	// sentinel width here would call it fitting at a large enough budget.
+	n, ok := encodedLen(items)
+	return ok && n <= b.max-envelope
+}
+
+// widths marshals each row once so the trim loops can scan sizes without
+// re-encoding. Only reached once the whole slice has already overflowed.
```
julianshen: _⚡ memory-exhaustion_ | _🔴 Blocking_ | _`oom`_

**Size checking materializes the entire oversized page in memory.**

`encodedLen(items)` is not actually streaming: `encoding/json.Encoder.Encode` first builds the complete JSON value in an internal `encodeState` buffer before calling `lenCounter.Write`. When the page exceeds the budget, `Fit` then calls `widths`, which marshals every row again before trimming. A reachable history request containing a message with very high reaction cardinality therefore retains the decoded Cassandra page while allocating full-page JSON and per-row JSON buffers; concurrent reads can cause GC thrashing, container eviction, or process OOM before `blankOversize` runs. Use a genuinely budget-short-circuiting size calculation that stops once the remaining budget is exceeded, and stop sizing rows after the fitting prefix/window is known. For message rows, field-wise capped accounting or another encoder that can abort during traversal is needed; a limited writer around `encoding/json.Encoder` is insufficient because `Encode` buffers first.

<details>

```
In `pkg/pagefit/pagefit.go` around line 136: `encodedLen(items)` is not actually streaming: `encoding/json.Encoder.Encode` first builds the complete JSON value in an internal `encodeState` buffer before calling `lenCounter.Write`. When the page exceeds the budget, `Fit` then calls `widths`, which marshals every row again before trimming. A reachable history request containing a message with very high reaction cardinality therefore retains the decoded Cassandra page while allocating full-page JSON and per-row JSON buffers; concurrent reads can cause GC thrashing, container eviction, or process OOM before `blankOversize` runs. Use a genuinely budget-short-circuiting size calculation that stops once the remaining budget is exceeded, and stop sizing rows after the fitting prefix/window is known. For message rows, field-wise capped accounting or another encoder that can abort during traversal is needed; a limited writer around `encoding/json.Encoder` is insufficient because `Encode` buffers first.

```

</details>

pkg/pagefit/pagefit.go
```diff
@@ -0,0 +1,166 @@
...
+func widths[T any](items []T) []int {
+	out := make([]int, len(items))
+	for i := range items {
+		data, err := json.Marshal(items[i])
+		if err != nil {
+			out[i] = unmeasurable
```
julianshen: _🏷️ error-handling_ | _🔴 Blocking_ | _`silent-failure`_

**Marshal failures are misreported as oversized rows.**

`widths` converts every `json.Marshal` error into the same sentinel used for an oversized row. `Fit`/`FitWindow` then return `oversize=true`, and the history-service callers blank the message, set `truncated:true`, and return success. Thus an encoding failure—for example malformed raw JSON in a message field—can be permanently skipped during pagination and falsely presented as transport truncation. Return marshal errors separately from size results and propagate them from the handlers; only successful encodings whose measured size exceeds the budget should be blanked.

<details>

```
In `pkg/pagefit/pagefit.go` around line 142: `widths` converts every `json.Marshal` error into the same sentinel used for an oversized row. `Fit`/`FitWindow` then return `oversize=true`, and the history-service callers blank the message, set `truncated:true`, and return success. Thus an encoding failure—for example malformed raw JSON in a message field—can be permanently skipped during pagination and falsely presented as transport truncation. Return marshal errors separately from size results and propagate them from the handlers; only successful encodings whose measured size exceeds the budget should be blanked.

```

</details>

user-service/service/threads.go
```diff
@@ -110,6 +114,20 @@ func (s *UserService) ListUserThreads(c *natsrouter.Context, req model.ThreadLis
...
 
+	// Trim after enrichment — it adds bytes, so a page measured before it can
+	// still overflow. The cursor below then resumes at the last kept row.
+	//
+	// The oversize return is ignored: a lone row too large to send has no
+	// blanked form here (unlike a message, a thread item carries no truncated
+	// flag), and one cannot occur in practice — its parent body inherits the
+	// 20 KB content cap. Such a row falls through to the router's
+	// response_too_large reply rather than being silently degraded.
+	kept, dropped, _ := pagefit.Fit(merged, s.pageBudget, threadListEnvelope)
```
julianshen: _🎯 correctness_ | _🔴 Blocking_ | _`tgd-review`_

**A lone oversized thread row is still sent unchanged.**

`pagefit.Fit` reports `oversize=true` when its mandatory retained row still exceeds the budget, but this call discards that result. The claim that this cannot occur is not valid: the broker or `MAX_RESPONSE_BYTES` can impose a budget below 20 KB, enrichment adds data, and the forwarded parent message can contain fields such as unbounded reactions in addition to its capped body. Such a page still becomes `response_too_large`, contradicting the documented Thread List behavior and the forward-progress goal. Handle the oversized item by producing a defined truncated representation, or explicitly remove Thread List from the single-row guarantee and update the API contract.

<details>

```
In `user-service/service/threads.go` around line 125: `pagefit.Fit` reports `oversize=true` when its mandatory retained row still exceeds the budget, but this call discards that result. The claim that this cannot occur is not valid: the broker or `MAX_RESPONSE_BYTES` can impose a budget below 20 KB, enrichment adds data, and the forwarded parent message can contain fields such as unbounded reactions in addition to its capped body. Such a page still becomes `response_too_large`, contradicting the documented Thread List behavior and the forward-progress goal. Handle the oversized item by producing a defined truncated representation, or explicitly remove Thread List from the single-row guarantee and update the API contract.

```

</details>

user-service/service/threads.go
```diff
@@ -110,6 +114,20 @@ func (s *UserService) ListUserThreads(c *natsrouter.Context, req model.ThreadLis
...
+	if dropped {
```
julianshen: _🏷️ failure-handling_ | _🟠 Warning_ | _`distributed-system`_

**A single oversized thread still dead-ends pagination.**

`pagefit.Fit` reports `oversize=true` while retaining one row, but this handler discards that signal. A thread whose parent message fits the history-service reply but exceeds the user-service budget after enrichment remains unchanged; if the final payload crosses the broker limit, NATS returns `response_too_large`. Retrying with `limit=1` cannot make progress because the same row is returned again. Add a bounded representation for an oversized thread item, such as blanking its parent message with an explicit truncation marker, and verify the assembled response fits.

<details>

```
In `user-service/service/threads.go` around line 126: `pagefit.Fit` reports `oversize=true` while retaining one row, but this handler discards that signal. A thread whose parent message fits the history-service reply but exceeds the user-service budget after enrichment remains unchanged; if the final payload crosses the broker limit, NATS returns `response_too_large`. Retrying with `limit=1` cannot make progress because the same row is returned again. Add a bounded representation for an oversized thread item, such as blanking its parent message with an explicit truncation marker, and verify the assembled response fits.

```

</details>

julianshen: _🏷️ partial-failure_ | _🟠 Warning_ | _`distributed-system`_

**Failed-site metadata is not included in exact payload sizing.**

Only `merged` is measured, while the final response also contains the dynamically sized `UnavailableSites`; the fixed 256-byte envelope cannot bound that list. During a broad cross-site outage, failed site IDs accumulate after fan-out, so the fitter can retain an item prefix that appears safe but the assembled response exceeds `max_payload`. The caller then receives `response_too_large` instead of the available threads and failure metadata. Size the complete `ThreadListResponse` after setting the cursor and unavailable sites, then trim items until the exact encoded response fits; define bounded behavior if the non-item envelope alone is too large.

<details>

```
In `user-service/service/threads.go` around line 126: Only `merged` is measured, while the final response also contains the dynamically sized `UnavailableSites`; the fixed 256-byte envelope cannot bound that list. During a broad cross-site outage, failed site IDs accumulate after fan-out, so the fitter can retain an item prefix that appears safe but the assembled response exceeds `max_payload`. The caller then receives `response_too_large` instead of the available threads and failure metadata. Size the complete `ThreadListResponse` after setting the cursor and unavailable sites, then trim items until the exact encoded response fits; define bounded behavior if the non-item envelope alone is too large.

```

</details>

history-service/cmd/main.go
```diff
@@ -199,6 +200,8 @@ func main() {
 	}
```
julianshen, line 201: _🏷️ failover_ | _🟠 Warning_ | _`distributed-system`, `nats` (2 rules)_

**The payload budget becomes stale after broker reconnection.**

The service snapshots `MaxPayload()` once at startup and stores an immutable budget; user-service uses the same pattern. If a network partition or rolling broker reconfiguration reconnects the client from a server advertising a larger cap to one with a smaller cap, handlers continue producing pages sized for the old server and the new broker rejects them. Resolve the effective cap from the current connection on requests, or update an atomic budget from reconnect/discovery callbacks; an explicit override should still be capped by the currently advertised broker maximum.

<details>

- **`nats`**:203: The service snapshots nc.NatsConn().MaxPayload() once at startup and stores a static budget. If NATS reconnects to a server advertising a lower max_payload, such as during a rolling configuration change, the connection enforces the new limit while handlers continue sizing against the old one. Replies between those limits then fail with ErrMaxPayload until restart. The user-service startup has the same pattern. Update an atomic budget from reconnect callbacks or resolve against the connection's current MaxPayload for each request.

</details>

<details>

```
In `history-service/cmd/main.go` around line 201: The service snapshots `MaxPayload()` once at startup and stores an immutable budget; user-service uses the same pattern. If a network partition or rolling broker reconfiguration reconnects the client from a server advertising a larger cap to one with a smaller cap, handlers continue producing pages sized for the old server and the new broker rejects them. Resolve the effective cap from the current connection on requests, or update an atomic budget from reconnect/discovery callbacks; an explicit override should still be capped by the currently advertised broker maximum.

```

</details>

pkg/pagefit/pagefit.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// Resolve picks the startup ceiling: a positive override (MAX_RESPONSE_BYTES)
+// wins, else the broker's max_payload. Both are reduced by reserve.
+func Resolve(override, brokerMaxPayload int64, reserve int) Budget {
+	if override > 0 {
+		return NewBudget(override, reserve)
+	}
+	return NewBudget(brokerMaxPayload, reserve)
+}
```
julianshen, line 50: _🎯 nats-correctness_ | _🟠 Warning_ | _`nats`_

**The configured override can exceed the broker's max_payload.**

Resolve lets any positive MAX_RESPONSE_BYTES replace the broker-advertised limit. If it is configured above max_payload, handlers can construct replies larger than the active NATS connection accepts, causing ErrMaxPayload and a response_too_large fallback. Treat the override as an additional cap and never allow it to raise the broker limit.

<details>

> ‼️ **IMPORTANT**

```suggestion
func Resolve(override, brokerMaxPayload int64, reserve int) Budget {
	if override > 0 && (brokerMaxPayload <= 0 || override < brokerMaxPayload) {
		return NewBudget(override, reserve)
	}
	return NewBudget(brokerMaxPayload, reserve)
}
```

</details>

<details>

```
In `pkg/pagefit/pagefit.go` around line 44: Resolve lets any positive MAX_RESPONSE_BYTES replace the broker-advertised limit. If it is configured above max_payload, handlers can construct replies larger than the active NATS connection accepts, causing ErrMaxPayload and a response_too_large fallback. Treat the override as an additional cap and never allow it to raise the broker limit.

```

</details>

julianshen (commented):
tGD inline review

user-service/service/threads.go
julianshen: _🎯 nats-correctness_ | _🟠 Warning_ | _`nats`_

**Fixed envelope estimates do not guarantee the reply fits NATS.**

Fit measures only merged items and assumes every other response field costs 256 bytes. The final ThreadListResponse also contains variable UnavailableSites and a cursor derived from potentially variable IDs; neither is measured. With enough unavailable sites or a large cursor, the item prefix can consume the calculated budget while the serialized reply exceeds max_payload, so NATS rejects it and the caller receives response_too_large. Build and measure the actual response, including cursor and unavailable sites, and shrink until its encoded body fits the budget; apply the same exact final-response check to the history replies.

<details>

```
In `user-service/service/threads.go` around line 125: Fit measures only merged items and assumes every other response field costs 256 bytes. The final ThreadListResponse also contains variable UnavailableSites and a cursor derived from potentially variable IDs; neither is measured. With enough unavailable sites or a large cursor, the item prefix can consume the calculated budget while the serialized reply exceeds max_payload, so NATS rejects it and the caller receives response_too_large. Build and measure the actual response, including cursor and unavailable sites, and shrink until its encoded body fits the budget; apply the same exact final-response check to the history replies.

```

</details>

history-service/internal/service/pagefit_integration_test.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+// Walking a room whose pages are trimmed must yield every message exactly once.
+// A cursor derived from the wrong row would show up here as a gap or a repeat,
+// which no unit test over mocked rows can prove against real Cassandra ordering.
+func TestLoadHistory_TrimmedPagination_Integration(t *testing.T) {
+	session := setupCassandra(t)
+	repo := cassrepo.NewRepository(session, msgbucket.New(24*time.Hour), 365, nil)
+
+	// A budget that admits only a handful of these rows, so the walk takes
```
julianshen, line 29: _🧪 tests_ | _🟠 Warning_ | _`nats`_

**The integration test never sends a reply through NATS.**

Despite claiming to validate a real low-max_payload broker, this test only starts Cassandra, injects an arbitrary 8 KiB Budget, and calls LoadHistory directly. It cannot detect ErrMaxPayload, actual response-envelope/header overhead, response_too_large fallback behavior, or startup MaxPayload wiring. Start NATS with a low max_payload, register the real router handler, issue NATS requests, and assert every page is a successful reply rather than the fallback envelope.

<details>

```
In `history-service/internal/service/pagefit_integration_test.go` around line 29: Despite claiming to validate a real low-max_payload broker, this test only starts Cassandra, injects an arbitrary 8 KiB Budget, and calls LoadHistory directly. It cannot detect ErrMaxPayload, actual response-envelope/header overhead, response_too_large fallback behavior, or startup MaxPayload wiring. Start NATS with a low max_payload, register the real router handler, issue NATS requests, and assert every page is a successful reply rather than the fallback envelope.

```

</details>

user-service/service/threads.go
julianshen: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Ignoring oversize makes Thread List retries unable to progress**

`pagefit.Fit` explicitly reports when the first retained item alone exceeds the budget, but this result is ignored. With a low valid `MAX_RESPONSE_BYTES`/broker cap or a sufficiently large parent message, the router rejects the retained row as `response_too_large`. Halving `limit` cannot fix a single oversized first row, so each retry repeats the cross-site history fan-out and enrichment before failing again, and pagination never advances. Add a truncated/blanked representation for an oversized thread item, clear its heavy parent content, and recheck the assembled response; alternatively reject such a budget at startup with a non-retryable configuration error.

<details>

```
In `user-service/service/threads.go` around line 125: `pagefit.Fit` explicitly reports when the first retained item alone exceeds the budget, but this result is ignored. With a low valid `MAX_RESPONSE_BYTES`/broker cap or a sufficiently large parent message, the router rejects the retained row as `response_too_large`. Halving `limit` cannot fix a single oversized first row, so each retry repeats the cross-site history fan-out and enrichment before failing again, and pagination never advances. Add a truncated/blanked representation for an oversized thread item, clear its heavy parent content, and recheck the assembled response; alternatively reject such a budget at startup with a non-retryable configuration error.

```

</details>

user-service/service/threads.go
```diff
@@ -11,11 +11,15 @@ import (
 	"github.com/hmchangw/chat/pkg/model"
 	"github.com/hmchangw/chat/pkg/natsrouter"
 	"github.com/hmchangw/chat/pkg/natsutil"
+	"github.com/hmchangw/chat/pkg/pagefit"
 )
 
 const (
 	defaultThreadListLimit = 20
 	maxThreadListLimit     = 100
```
julianshen, line 20: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**A fixed envelope estimate cannot guarantee the reply fits.**

`threadListEnvelope` claims to cover `unavailableSites` and `nextCursor`, but both are created outside the measured item slice and their encoded lengths vary. In particular, `unavailableSites` grows with failed configured sites, while the cursor embeds values from the retained row. Consequently, `pagefit.Fit` can declare the items fitting while the assembled response still exceeds the payload ceiling. The integration test invokes the service directly, so it cannot detect this transport-level overflow. Measure the actual provisional response and trim until it fits, or calculate the envelope from the actual unavailable-site list and generated cursor before choosing the prefix.

<details>

```
In `user-service/service/threads.go` around line 19: `threadListEnvelope` claims to cover `unavailableSites` and `nextCursor`, but both are created outside the measured item slice and their encoded lengths vary. In particular, `unavailableSites` grows with failed configured sites, while the cursor embeds values from the retained row. Consequently, `pagefit.Fit` can declare the items fitting while the assembled response still exceeds the payload ceiling. The integration test invokes the service directly, so it cannot detect this transport-level overflow. Measure the actual provisional response and trim until it fits, or calculate the envelope from the actual unavailable-site list and generated cursor before choosing the prefix.

```

</details>

docs/client-api.md
```diff
@@ -2889,6 +2889,7 @@ Used by every history-service method that returns messages. Mirrors the Cassandr
 | `threadRoomId` | string | Optional. The thread room ID when this is a thread message. |
 | `pinnedAt` | string | Optional. RFC 3339. With the `messages_by_room` `pinned_at` mirror, room-timeline history loads now return this on pinned rows too (previously only `pin.list` and point lookups carried it). |
 | `pinnedBy` | [MessageParticipant](#messageparticipant) | Optional. |
+| `truncated` | boolean | Optional. `true` when the server blanked this row because it alone exceeded the transport's `max_payload`. `msg`, `mentions`, `attachments`, `card`, `cardAction`, `quotedParentMessage`, `reactions` and `sysMsgData` are cleared; identifiers, `sender`, `createdAt` and `type` are retained so a placeholder can be rendered. Absent on every ordinary row. |
```
julianshen: _📝 documentation_ | _🟠 Warning_ | _`tgd-review`_

**The truncated-row contract omits cleared encrypted payload fields.**

`blankOversize` also clears `EncPayload` and `EncMeta`, but both `truncated` field descriptions enumerate the cleared fields without mentioning `encPayload` or `encMeta`. This is observable for encrypted messages and may lead clients to expect ciphertext to remain available. Add both fields to the cleared-field lists here and in the broadcast-message table.

<details>

```
In `docs/client-api.md` around line 2892: `blankOversize` also clears `EncPayload` and `EncMeta`, but both `truncated` field descriptions enumerate the cleared fields without mentioning `encPayload` or `encMeta`. This is observable for encrypted messages and may lead clients to expect ciphertext to remain available. Add both fields to the cleared-field lists here and in the broadcast-message table.

```

</details>

history-service/internal/service/messages.go
mliu33: Confirmed and fixed in d6f789b. You're right, and the schema proves it: `messages_by_room` is `PRIMARY KEY ((room_id, bucket), created_at, message_id)` — `message_id` is there precisely because `created_at` is not unique — while the resume bound is a strict timestamp-only `created_at < ?`.

Rather than change the cursor format, the trim now never ends mid-millisecond. `timestampRun` finds the maximal run sharing a row's `createdAt`, and the cut retreats to a run boundary, so every dropped row is strictly older than the resume bound. Load Surrounding applies the same rule at both edges, expanding instead of retreating when the pivot's own run would be split.

The residual you'd expect — a single millisecond holding more than a budget's worth of rows — is handled rather than ignored: that run is returned whole with its rows blanked (`truncated: true`), so no row is ever skipped and pagination still advances. It also logs a warning if the blanked run still exceeds the budget.

Added the Cassandra coverage you asked for: `TestLoadHistory_EqualTimestampsSurviveTrimming_Integration` seeds 8 groups of 3 rows sharing one millisecond with a budget that would otherwise cut through them, walks the room, and asserts every id appears exactly once.

On the read-amplification sub-finding — that one is real and not fixed here. Trimming still happens after the fetch, so dropped rows are re-read on the next page. Fixing it properly means a payload-aware cap inside the repository fetch, which is a larger change than this PR; tracking it as a follow-up rather than pretending otherwise.

---

pkg/pagefit/pagefit.go
mliu33: Correct on both counts, and fixed in d6f789b. `Encoder.Encode` does buffer the whole value internally, so calling the old helper "streaming" was wrong, and `widths` then re-marshalled every row on top of it.

Both passes are now one. `rowSizes` runs a single `json.Encoder` over a counting writer, encoding one row at a time and deriving each row's width from the counter delta; the array total is computed arithmetically from the row sums, separators and brackets, so no separate whole-slice encode happens at all. Peak transient memory is now one row's encoding rather than the full page plus per-row buffers, and no `[]byte` is retained.

Not done: the budget short-circuit you suggested — `rowSizes` still measures every row rather than stopping once the budget is blown. With peak memory now bounded by the largest single row that's a CPU optimisation rather than an OOM guard, so I left it out to keep the change small.

Measured on 100 rows: the fitting path is 595 B / 2 allocs, the trimmed path 61 KB / 203 allocs.

---

pkg/pagefit/pagefit.go
mliu33: Agreed — an encoding failure being laundered into `truncated: true` and a 200 was the worst shape this could have. Fixed in d6f789b.

The sentinel is gone entirely. `rowSizes` returns `([]int, int, error)`, and `Fit`/`FitWindow` now return an `error` alongside the size decision, so a marshal failure propagates instead of resolving to `oversize`. The history handlers wrap it (`fmt.Errorf("fit history page: %w", err)`) and it collapses to `internal` at the boundary; `ListUserThreads` does the same. Only a successful encode whose measured size exceeds the budget reaches `blankOversize`.

Covered by `TestFit_MarshalFailureIsAnErrorNotAnOversizeRow`, the `FitWindow` equivalent, and `TestListUserThreads_SizingFailureSurfacesAsAnError`, which feeds malformed `json.RawMessage` through the real handler and asserts an error rather than a trimmed page.

---

user-service/service/threads.go
mliu33: You were right to push back on the "cannot occur" claim — it rested on the 20 KB content cap while ignoring that `MAX_RESPONSE_BYTES` or the broker can set a budget below it, and that enrichment and unbounded reactions add on top. Fixed in d6f789b rather than documented away.

`ListUserThreads` now consumes the `oversize` signal. `ThreadListItem` gained a `Truncated bool` field, and `blankOversizeThread` drops `parentMessage` and `lastMessage` while keeping the identifiers and `lastMsgAt` — so the row still fits, the cursor still derives from it, and pagination advances instead of dead-ending on a `response_too_large` that halving `limit` could never clear.

`TestListUserThreads_LoneOversizeItemIsBlankedNotPassedThrough` drives a single 8 KB row against a 512-byte budget and asserts the bodies are gone, the identifiers and sort key survive, and the encoded page actually fits.

The API contract is updated to match rather than left claiming the old behaviour: `docs/client-api.md` documents `truncated` on ThreadListItem and states the forward-progress guarantee for Thread List.

---

user-service/service/threads.go
mliu33: Fixed in d6f789b, along the lines you suggested. `blankOversizeThread` clears `parentMessage` and `lastMessage` and sets an explicit `truncated: true` marker on the row, so the assembled reply fits and the `limit=1` retry loop you describe can't happen — the same row now comes back small enough to deliver, and the cursor advances past it.

Covered by `TestListUserThreads_LoneOversizeItemIsBlankedNotPassedThrough`, which asserts the encoded page fits the budget rather than only that the bodies were dropped.

One part of your suggestion isn't done: "verify the assembled response fits". Sizing still measures the items against a fixed 256-byte envelope allowance rather than encoding the final `ThreadListResponse`. That's the same gap as your `unavailableSites`/cursor threads, and it's deliberately deferred to a follow-up rather than fixed here.

---

pkg/pagefit/pagefit.go
mliu33: Applied your suggestion verbatim in d6f789b — the override now only ever lowers the ceiling:

```go
if override > 0 && (brokerMaxPayload <= 0 || override < brokerMaxPayload) {
	return NewBudget(override, reserve)
}
return NewBudget(brokerMaxPayload, reserve)
```

Pinned by `TestResolve_OverrideNeverExceedsTheBrokerCap`, covering an override below the broker cap (honoured), above it (clamped to the broker), equal to it, and the no-broker-cap-known case where the override stands alone.

---

user-service/service/threads.go
mliu33: Fixed in d6f789b — `ListUserThreads` no longer discards the `oversize` result. `blankOversizeThread` clears the row's `parentMessage` and `lastMessage` and marks it `truncated: true`, so the retained row fits and the repeated cross-site fan-out you describe can't recur: the retry that would have returned the same oversized row now gets a deliverable one.

I did not take the alternative of rejecting a small budget at startup. A low `MAX_RESPONSE_BYTES` is a legitimate operational choice, and failing to boot on it seemed worse than degrading one row; `Resolve` also now clamps the override to the broker cap so the budget can't be set *above* what the wire accepts.

---

docs/client-api.md
mliu33: Fixed in d6f789b. `encPayload` and `encMeta` are now listed among the cleared fields in both `truncated` descriptions in `docs/client-api.md` and in the broadcast-message table in `docs/client-api/events.md`.

While there I also corrected the descriptions to cover the second way a row can now come back truncated — sharing a `createdAt` millisecond with an oversized row, where the whole run is returned together and blanked rather than split. That case didn't exist when you reviewed; it's the fix for your timestamp-tie finding.

---

history-service/internal/service/messages.go
```diff
@@ -93,12 +96,18 @@ func (s *HistoryService) LoadHistory(c *natsrouter.Context, req models.LoadHisto
...
+	// dropping the tail leaves the client's next before = oldest kept createdAt.
```
julianshen, line 100: _⚡ cassandra-performance_ | _🟠 Warning_ | _`cassandra`_

**Post-fetch trimming repeatedly reads discarded Cassandra rows**

`fitPage` runs after `GetMessagesBetweenDesc` has fetched up to the requested 100 rows from `messages_by_room`. With valid approximately 20 KiB messages and the default approximately 124 KiB budget, only about six rows may be returned; the next timestamp request then rereads roughly 94 previously discarded rows. A sequence of individually oversize rows can approach 100 Cassandra reads and JSON decodes per delivered row. Move the payload-aware cutoff into repository iteration or adapt the Cassandra fetch/page size while retaining a complete clustering-key cursor, so fetched-but-unreturned rows are not repeatedly scanned. Add a repository-level pagination test that counts rows read across a trimmed walk.

<details>

```
In `history-service/internal/service/messages.go` around line 100: `fitPage` runs after `GetMessagesBetweenDesc` has fetched up to the requested 100 rows from `messages_by_room`. With valid approximately 20 KiB messages and the default approximately 124 KiB budget, only about six rows may be returned; the next timestamp request then rereads roughly 94 previously discarded rows. A sequence of individually oversize rows can approach 100 Cassandra reads and JSON decodes per delivered row. Move the payload-aware cutoff into repository iteration or adapt the Cassandra fetch/page size while retaining a complete clustering-key cursor, so fetched-but-unreturned rows are not repeatedly scanned. Add a repository-level pagination test that counts rows read across a trimmed walk.

```

</details>

user-service/service/threads.go
```diff
@@ -110,6 +115,21 @@ func (s *UserService) ListUserThreads(c *natsrouter.Context, req model.ThreadLis
 	// botDM rows swap their bot account for the app's display name.
 	s.enrichThreadPage(c, merged)
 
+	// Trim after enrichment — it adds bytes, so a page measured before it can
+	// still overflow. The cursor below then resumes at the last kept row.
+	kept, oversize, err := pagefit.Fit(merged, s.pageBudget, threadListEnvelope)
```
julianshen, line 120: _🏷️ failure-handling_ | _🟠 Warning_ | _`distributed-system`_

**Variable response metadata is excluded from thread-page sizing**

`pagefit.Fit` charges a fixed 256-byte envelope, but `UnavailableSites` is assembled from failed fan-out targets and can exceed that estimate. During a multi-site outage, items can consume nearly the full calculated budget and the subsequently encoded unavailable-site list pushes the actual reply over `max_payload`; retries may continue failing even with one item if metadata dominates. Build the provisional response with its actual cursor and unavailable-site list, then trim until its encoded size fits, or calculate the exact non-item envelope before fitting.

<details>

```
In `user-service/service/threads.go` around line 120: `pagefit.Fit` charges a fixed 256-byte envelope, but `UnavailableSites` is assembled from failed fan-out targets and can exceed that estimate. During a multi-site outage, items can consume nearly the full calculated budget and the subsequently encoded unavailable-site list pushes the actual reply over `max_payload`; retries may continue failing even with one item if metadata dominates. Build the provisional response with its actual cursor and unavailable-site list, then trim until its encoded size fits, or calculate the exact non-item envelope before fitting.

```

</details>

user-service/service/threads.go
```diff
@@ -110,6 +115,21 @@ func (s *UserService) ListUserThreads(c *natsrouter.Context, req model.ThreadLis
 	// botDM rows swap their bot account for the app's display name.
 	s.enrichThreadPage(c, merged)
```
julianshen, line 117: _🏷️ mongodb-pagination_ | _🟠 Warning_ | _`mongodb`_

**Post-fetch trimming repeatedly re-reads dropped MongoDB rows**

The per-site Thread List queries and enrichment complete before `pagefit.Fit` discards the tail. The cursor then advances only to the last retained item, so the next cross-site request re-fetches the discarded thread rows from the BSON-backed leaf data path. With `limit=100` and large parent/last-message bodies that allow only a few items per reply, each page can query and decode nearly the same 100 rows at every site, causing O(limit/kept) MongoDB read amplification and repeated high-cardinality fan-out. This was acknowledged in the prior discussion and remains unfixed. Minimally, carry the effective retained limit in the opaque cursor and cap subsequent per-site fetches to that value (plus the existing lookahead), or make leaf cursor iteration payload-aware so it stops reading once enough candidates have been assembled.

<details>

```
In `user-service/service/threads.go` around line 117: The per-site Thread List queries and enrichment complete before `pagefit.Fit` discards the tail. The cursor then advances only to the last retained item, so the next cross-site request re-fetches the discarded thread rows from the BSON-backed leaf data path. With `limit=100` and large parent/last-message bodies that allow only a few items per reply, each page can query and decode nearly the same 100 rows at every site, causing O(limit/kept) MongoDB read amplification and repeated high-cardinality fan-out. This was acknowledged in the prior discussion and remains unfixed. Minimally, carry the effective retained limit in the opaque cursor and cap subsequent per-site fetches to that value (plus the existing lookahead), or make leaf cursor iteration payload-aware so it stops reading once enough candidates have been assembled.

```

</details>

user-service/service/threads.go
```diff
@@ -110,6 +115,21 @@ func (s *UserService) ListUserThreads(c *natsrouter.Context, req model.ThreadLis
...
 
+	// Trim after enrichment — it adds bytes, so a page measured before it can
```
julianshen, line 118: _⚡ performance_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Thread trimming repeats cross-site fetch and enrichment work.**

Each request first obtains and enriches the full merged page, then retains only the budget-fitting prefix. If five of 100 large thread rows fit, the next cursor advances by five, so subsequent cross-site fan-out fetches most of the discarded rows again and enriches them again before another trim. Across many sites this multiplies network, history-store hydration, and directory lookup work by roughly `limit/kept`. Build the page incrementally within the byte budget, or carry an adaptive per-site fetch size/continuation state in the cursor so later requests do not repeatedly retrieve and enrich already-discarded rows.

<details>

```
In `user-service/service/threads.go` around line 118: Each request first obtains and enriches the full merged page, then retains only the budget-fitting prefix. If five of 100 large thread rows fit, the next cursor advances by five, so subsequent cross-site fan-out fetches most of the discarded rows again and enriches them again before another trim. Across many sites this multiplies network, history-store hydration, and directory lookup work by roughly `limit/kept`. Build the page incrementally within the byte budget, or carry an adaptive per-site fetch size/continuation state in the cursor so later requests do not repeatedly retrieve and enrich already-discarded rows.

```

</details>

julianshen (commented):
tGD inline review

history-service/internal/service/utils.go
```diff
@@ -168,3 +171,141 @@ func canModify(msg *models.Message, account string) bool {
...
+	start, end := timestampRun(msgs, len(kept))
+	n := start
+	if n == 0 {
+		// The kept prefix sits inside one millisecond that continues past it.
+		// Nothing can be cut without losing rows, so keep the run whole and
+		// shrink its rows instead.
+		n = end
+		blankRange(msgs[:n])
+		s.warnIfStillOversize(ctx, msgs[:n], envelope)
+	}
```
julianshen, line 230: _🎯 correctness_ | _🟠 Warning_ | _`tgd-review`_

**An unfittable timestamp group is knowingly returned over budget.**

When a tied timestamp group still exceeds the budget after blanking, this path only logs and returns the entire group. The transport will then emit `response_too_large`, contradicting the documented claim that Load History pages are sized to fit and that timestamp-sharing rows are returned as truncated placeholders. Use a cursor capable of splitting the group safely, or document this fallback explicitly instead of promising these RPCs always fit.

<details>

```
In `history-service/internal/service/utils.go` around line 226: When a tied timestamp group still exceeds the budget after blanking, this path only logs and returns the entire group. The transport will then emit `response_too_large`, contradicting the documented claim that Load History pages are sized to fit and that timestamp-sharing rows are returned as truncated placeholders. Use a cursor capable of splitting the group safely, or document this fallback explicitly instead of promising these RPCs always fit.

```

</details>

user-service/service/threads.go
```diff
@@ -110,6 +115,21 @@ func (s *UserService) ListUserThreads(c *natsrouter.Context, req model.ThreadLis
...
+	// still overflow. The cursor below then resumes at the last kept row.
+	kept, oversize, err := pagefit.Fit(merged, s.pageBudget, threadListEnvelope)
+	if err != nil {
```
julianshen, line 121: _⚡ memory_ | _🟠 Warning_ | _`oom`_

**Dropped thread bodies remain reachable after trimming**

`kept` is a subslice of `merged`, so assigning it back does not release the backing-array elements beyond its length. Up to 99 discarded `ThreadListItem` values can therefore keep large `ParentMessage` and `LastMessage` buffers reachable while the response is serialized. With the configured concurrent request fan-out, large thread bodies can multiply this otherwise-dead memory into GC thrashing or container eviction. Clear the discarded tail before shortening the slice.

<details>

```
In `user-service/service/threads.go` around line 121: `kept` is a subslice of `merged`, so assigning it back does not release the backing-array elements beyond its length. Up to 99 discarded `ThreadListItem` values can therefore keep large `ParentMessage` and `LastMessage` buffers reachable while the response is serialized. With the configured concurrent request fan-out, large thread bodies can multiply this otherwise-dead memory into GC thrashing or container eviction. Clear the discarded tail before shortening the slice.

```

</details>

julianshen (commented):
tGD inline review

history-service/cmd/main.go
```diff
@@ -199,6 +200,8 @@ func main() {
...
 
 	pub := publisher.New(js, publisher.WithMetrics(publishMetrics))
+	opts = append(opts, service.WithPageBudget(
```
julianshen: _🏷️ nats-reconnect_ | _🟠 Warning_ | _`nats`_

**Reconnects can leave the budget above the active NATS limit.**

The service snapshots `MaxPayload()` into an immutable budget at startup; user-service does the same. If the connection later reconnects to a server advertising a lower `max_payload`, replies between the new limit and the stored limit are rejected by NATS. Supply the current connection limit when fitting each reply, or update a shared atomic budget from reconnect/INFO callbacks, always capping the configured override by the current broker limit.

<details>

```
In `history-service/cmd/main.go` around line 203: The service snapshots `MaxPayload()` into an immutable budget at startup; user-service does the same. If the connection later reconnects to a server advertising a lower `max_payload`, replies between the new limit and the stored limit are rejected by NATS. Supply the current connection limit when fitting each reply, or update a shared atomic budget from reconnect/INFO callbacks, always capping the configured override by the current broker limit.

```

</details>

pkg/pagefit/pagefit.go
```diff
@@ -0,0 +1,167 @@
...
+// misconfigured reserve must not empty every reply.
+func NewBudget(brokerMaxPayload int64, reserve int) Budget {
+	if brokerMaxPayload <= 0 || reserve < 0 {
+		return Budget{}
+	}
+	if brokerMaxPayload > math.MaxInt32 {
+		brokerMaxPayload = math.MaxInt32
+	}
+	max := int(brokerMaxPayload) - reserve
+	if max <= 0 {
```
julianshen, line 32: _🏷️ nats-configuration_ | _🟠 Warning_ | _`nats`_

**Low NATS payload limits silently disable response trimming.**

`NewBudget` returns a disabled budget when the reserve equals or exceeds the cap. With `max_payload=4096` and `DefaultReserve=4096`, for example, every `Fit` call keeps the full page and NATS rejects oversized replies—the opposite of the safety feature. Reject this configuration at startup or use a cap-aware reserve that still produces a positive payload budget; add coverage through the startup `Resolve` path.

<details>

```
In `pkg/pagefit/pagefit.go` around line 32: `NewBudget` returns a disabled budget when the reserve equals or exceeds the cap. With `max_payload=4096` and `DefaultReserve=4096`, for example, every `Fit` call keeps the full page and NATS rejects oversized replies—the opposite of the safety feature. Reject this configuration at startup or use a cap-aware reserve that still produces a positive payload budget; add coverage through the startup `Resolve` path.

```

</details>

history-service/internal/service/pagefit_integration_test.go
```diff
@@ -0,0 +1,152 @@
...
+	// several trimmed pages rather than one.
+	svc := New(repo, alwaysSubscribedRepo{}, stubRoomRepo{}, &recordingPublisher{}, nil, nil, nil, nil,
```
julianshen, line 31: _🧪 tests_ | _🟠 Warning_ | _`nats`_

**The integration test never crosses the NATS request/reply transport.**

This test injects an arbitrary budget and calls `LoadHistory` directly. It cannot detect actual `max_payload` rejection, response-envelope overhead, `response_too_large` fallback behavior, or incorrect startup/reconnect wiring. Start NATS with a low payload limit, register the real router handler, walk pages using NATS requests, and assert every reply is a successful correlated response rather than an oversize error envelope.

<details>

```
In `history-service/internal/service/pagefit_integration_test.go` around line 31: This test injects an arbitrary budget and calls `LoadHistory` directly. It cannot detect actual `max_payload` rejection, response-envelope overhead, `response_too_large` fallback behavior, or incorrect startup/reconnect wiring. Start NATS with a low payload limit, register the real router handler, walk pages using NATS requests, and assert every reply is a successful correlated response rather than an oversize error envelope.

```

</details>

history-service/internal/service/utils.go
```diff
@@ -168,3 +171,141 @@ func canModify(msg *models.Message, account string) bool {
...
+		return kept, false, nil
+	}
+	start, end := timestampRun(msgs, len(kept))
+	n := start
+	if n == 0 {
+		// The kept prefix sits inside one millisecond that continues past it.
+		// Nothing can be cut without losing rows, so keep the run whole and
+		// shrink its rows instead.
+		n = end
+		blankRange(msgs[:n])
```
julianshen, line 228: _⚡ memory_ | _🟠 Warning_ | _`oom`_

**Trimmed history pages retain discarded rich-content buffers.**

`fitPage` returns a prefix of the original message slice without clearing its tail, and `LoadSurroundingMessages` similarly returns `messages[lo:hi]`. Go keeps and scans the entire backing array, so discarded message bodies, attachments, decoded attachments, reactions, and encrypted payloads remain reachable while the response is serialized. For a 100-row history request where only about six 20 KiB rows fit, roughly 94 discarded rich rows remain live per request; concurrent large-room reads can therefore drive substantial GC pressure or container eviction. Clear `msgs[n:]` before returning from `fitPage`, and clear `messages[:lo]` plus `messages[hi:]` before retaining the surrounding sub-slice.

<details>

```
In `history-service/internal/service/utils.go` around line 224: `fitPage` returns a prefix of the original message slice without clearing its tail, and `LoadSurroundingMessages` similarly returns `messages[lo:hi]`. Go keeps and scans the entire backing array, so discarded message bodies, attachments, decoded attachments, reactions, and encrypted payloads remain reachable while the response is serialized. For a 100-row history request where only about six 20 KiB rows fit, roughly 94 discarded rich rows remain live per request; concurrent large-room reads can therefore drive substantial GC pressure or container eviction. Clear `msgs[n:]` before returning from `fitPage`, and clear `messages[:lo]` plus `messages[hi:]` before retaining the surrounding sub-slice.

```

</details>

user-service/service/threads.go
```diff
@@ -110,6 +115,21 @@ func (s *UserService) ListUserThreads(c *natsrouter.Context, req model.ThreadLis
...
+		return nil, fmt.Errorf("fit thread list page: %w", err)
+	}
+	hasNext = hasNext || len(kept) < len(merged)
```
julianshen: _⚡ memory_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Trimmed thread slices retain all dropped message bodies.**

kept is a subslice of merged, so its response retains the entire backing array and all ParentMessage/LastMessage buffers in the discarded tail while the router serializes the reply. Under concurrent large thread requests, up to 99 otherwise-dead bodies per request remain GC-reachable, increasing heap pressure and eviction risk. Clear the tail before shortening the slice.

<details>

```
In `user-service/service/threads.go` around line 124: kept is a subslice of merged, so its response retains the entire backing array and all ParentMessage/LastMessage buffers in the discarded tail while the router serializes the reply. Under concurrent large thread requests, up to 99 otherwise-dead bodies per request remain GC-reachable, increasing heap pressure and eviction risk. Clear the tail before shortening the slice.

```

</details>

user-service/service/threads.go
```diff
@@ -110,6 +115,21 @@ func (s *UserService) ListUserThreads(c *natsrouter.Context, req model.ThreadLis
...
+	merged = kept
```
julianshen, line 130: _⚡ memory_ | _🔵 Suggestion_ | _`tgd-review`_

**Discarded thread bodies remain reachable through the backing array.**

`kept` is a subslice of `merged`, so the response retains the original backing array and its discarded `ParentMessage` and `LastMessage` references through serialization. Clearing the discarded tail before shortening the slice lets those potentially large buffers be collected sooner under concurrent load.

<details>

```
In `user-service/service/threads.go` around line 125: `kept` is a subslice of `merged`, so the response retains the original backing array and its discarded `ParentMessage` and `LastMessage` references through serialization. Clearing the discarded tail before shortening the slice lets those potentially large buffers be collected sooner under concurrent load.

```

</details>

docs/client-api.md
```diff
@@ -2889,6 +2889,7 @@ Used by every history-service method that returns messages. Mirrors the Cassandr
 | `threadRoomId` | string | Optional. The thread room ID when this is a thread message. |
 | `pinnedAt` | string | Optional. RFC 3339. With the `messages_by_room` `pinned_at` mirror, room-timeline history loads now return this on pinned rows too (previously only `pin.list` and point lookups carried it). |
 | `pinnedBy` | [MessageParticipant](#messageparticipant) | Optional. |
+| `truncated` | boolean | Optional. `true` when the server blanked this row to make the page fit — either because the row alone exceeded the transport's `max_payload`, or because it shares a `createdAt` millisecond with such a row and the whole group had to be returned together. `msg`, `mentions`, `attachments`, `card`, `cardAction`, `quotedParentMessage`, `reactions`, `sysMsgData`, `encPayload` and `encMeta` are cleared; identifiers, `sender`, `createdAt` and `type` are retained so a placeholder can be rendered. Absent on every ordinary row. |
```
julianshen, line 2892: _📝 documentation_ | _🔵 Suggestion_ | _`tgd-review`_

**The documented reason for truncating timestamp peers is inaccurate.**

The description says a timestamp peer is blanked only when it shares a millisecond with a row that alone exceeded the budget. In `fitPage` and `fitWindow`, a group of individually deliverable rows is also blanked when the group collectively cannot be split at the timestamp boundary. Document that collective case and mirror the correction in the broadcast/events schema and model comment.

<details>

```
In `docs/client-api.md` around line 2892: The description says a timestamp peer is blanked only when it shares a millisecond with a row that alone exceeded the budget. In `fitPage` and `fitWindow`, a group of individually deliverable rows is also blanked when the group collectively cannot be split at the timestamp boundary. Document that collective case and mirror the correction in the broadcast/events schema and model comment.

```

</details>

julianshen (commented):
tGD inline review

chenjr0719 (approved):
Re-reviewed at the current head after the fix pass — the four blocking findings from the first round are all resolved, verified in code + tests:

- **Timestamp-tie skip:** the trim now cuts back to a whole-timestamp-run boundary (never splits a same-createdAt group), so resuming with before = oldest-kept createdAt cannot skip a same-millisecond row — with tests covering both Load History and Load Surrounding (both edges) plus the unsplittable-run-blanked case.
- **Size-calc memory:** row sizing now encodes one row at a time into a discarding counter — peak memory is the largest single row, not the whole page.
- **Marshal errors:** Fit now returns an error and the handlers propagate it as an error envelope, instead of blanking the row as truncated-success.
- **Thread List oversize:** the oversize result is handled (lone row blanked, encrypted body cleared) with a distinct sizeLimited flag, so a too-large row still makes forward progress.

CI is green. Remaining items are non-blocking (the trim-after-fetch read amplification is a documented perf tradeoff; the integration test injects the budget rather than crossing real NATS max_payload — a test-depth gap worth a follow-up, not a blocker). Nice, thorough fix pass. LGTM.

hmchangw (approved):
Looks great. Thanks!

history-service/cmd/main.go
```diff
@@ -199,6 +200,14 @@ func main() {
 	}
 
 	pub := publisher.New(js, publisher.WithMetrics(publishMetrics))
+	// A zero Budget disables trimming, so the toggle needs no handler branch.
+	pageBudget := pagefit.Budget{}
```
julianshen, line 204: _🏷️ failover_ | _🟠 Warning_ | _`distributed-system`, `nats`, `performance-and-concurrency`, `tgd-review` (4 rules)_

**The payload budget remains stale after broker reconnection.**

The service snapshots `MaxPayload()` once during startup; user-service does the same. If a partition or rolling NATS reconfiguration reconnects the client to a server with a lower `max_payload`, the connection starts enforcing the new limit while handlers continue sizing replies against the old one. Replies between those limits then become `response_too_large` until restart. Keep an atomic effective budget updated from reconnect/discovery events, or resolve it from the current connection for each request, still capping any override by the current broker maximum.

<details>

- **`performance-and-concurrency`**:204: The service snapshots `MaxPayload()` at startup into an immutable budget; user-service does the same. If a rolling broker reconfiguration reconnects the client from a larger limit to a smaller one, concurrent handlers continue producing responses sized for the old limit, and the broker rejects them until restart, provoking retries and latency. Update an atomic budget from reconnect/discovery callbacks or resolve against the current connection limit per request, always capping an explicit override by the currently advertised maximum.
- **`nats`**:204: The service snapshots `MaxPayload()` during startup; `user-service/main.go` does the same. If reconnection moves the client from a server advertising a larger limit to one with a smaller limit, handlers still retain replies that fit only the old limit, and publishing them fails with `ErrMaxPayload`/`response_too_large`. Inject a budget provider based on the connection's current `MaxPayload()`, or atomically refresh the effective budget on reconnect while still capping any override to the current broker limit.
- **`tgd-review`**:203: The service snapshots `MaxPayload()` once at startup. After reconnecting to a server with a lower limit, handlers still size responses against the old value and the broker can reject them. User-service has the same startup snapshot. Update an atomic budget from connection callbacks or resolve against the current connection limit when sizing replies.

</details>

<details>

```
In `history-service/cmd/main.go` around line 204: The service snapshots `MaxPayload()` once during startup; user-service does the same. If a partition or rolling NATS reconfiguration reconnects the client to a server with a lower `max_payload`, the connection starts enforcing the new limit while handlers continue sizing replies against the old one. Replies between those limits then become `response_too_large` until restart. Keep an atomic effective budget updated from reconnect/discovery events, or resolve it from the current connection for each request, still capping any override by the current broker maximum.

```

</details>

user-service/service/threads.go
```diff
@@ -110,7 +115,29 @@ func (s *UserService) ListUserThreads(c *natsrouter.Context, req model.ThreadLis
...
+	// still overflow. The cursor below then resumes at the last kept row.
+	kept, oversize, err := pagefit.Fit(merged, s.pageBudget, threadListEnvelope)
+	if err != nil {
+		return nil, fmt.Errorf("fit thread list page: %w", err)
+	}
+	// Dropped rows only: a blanked row is truncated, never sizeLimited, since
+	// it is still oversize at half the limit. That holds here because a lone
+	// oversize row leaves len(kept) == len(merged) — do not weaken the test to
+	// something the blanking path can also satisfy.
+	sizeLimited := len(kept) < len(merged)
```
julianshen, line 128: _⚡ memory_ | _🟠 Warning_ | _`oom`, `performance-and-concurrency` (2 rules)_

**Dropped thread bodies remain reachable after page trimming.**

A multi-site request can enrich up to 100 rows with large `ParentMessage` and `LastMessage` buffers, then retain only a few under the byte budget. Because `kept` shares `merged`'s backing array, the discarded raw-message buffers remain reachable while the response is serialized. At the default 256 concurrent handlers, repeated large thread-list requests can produce substantial GC pressure or container eviction. Clear the discarded elements before shortening the slice.

<details>

- **`performance-and-concurrency`**:125: `kept` is a subslice of `merged`, so the response keeps the original backing array and the discarded `ParentMessage` and `LastMessage` buffers reachable through serialization. With 100 large hydrated rows and high request concurrency, this can materially increase live heap and GC pressure. Clear `merged[len(kept):]` before shortening the slice; optionally cap the returned slice capacity as well.

</details>

<details>

```
In `user-service/service/threads.go` around line 128: A multi-site request can enrich up to 100 rows with large `ParentMessage` and `LastMessage` buffers, then retain only a few under the byte budget. Because `kept` shares `merged`'s backing array, the discarded raw-message buffers remain reachable while the response is serialized. At the default 256 concurrent handlers, repeated large thread-list requests can produce substantial GC pressure or container eviction. Clear the discarded elements before shortening the slice.

```

</details>

docs/client-api.md
```diff
@@ -6718,7 +6724,21 @@ Every error response — NATS reply subjects, JetStream async results, and HTTP
...
+> **Oversize replies.** If a reply would exceed the transport's maximum payload size, it is returned as `code: internal` with `reason: response_too_large` instead. This covers both a success body that is too large and an error envelope that is too large; either way the caller gets an answer rather than a timeout. This is most likely on large history reads (e.g. Load History / Load Next / Load Surrounding / Get Thread Messages with a high `limit`); the client should retry with a smaller `limit`. Branch on `reason` (`response_too_large`), not the message text. Which RPCs can raise it, and where the smaller-`limit` retry is the wrong move, is set out under **A page may be shorter than `limit`** below.
+
+
```
julianshen, line 6729: _📝 documentation_ | _🟠 Warning_ | _`tgd-review`_

**Blank lines inside the blockquote fail markdown lint.**

The two unquoted blank lines split the adjacent admonition block and trigger markdownlint MD028. Keep the separator inside the blockquote.

<details>

> ‼️ **IMPORTANT**

```suggestion
>
```

</details>

<details>

```
In `docs/client-api.md` around line 6728: The two unquoted blank lines split the adjacent admonition block and trigger markdownlint MD028. Keep the separator inside the blockquote.

```

</details>

julianshen (commented):
tGD inline review

PR #349, opened 2026-08-22
feat(broadcast-worker): deliver thread replies to open thread panels

mliu33:
## Problem

A user who opens a thread panel but has never replied is not a thread follower, so `broadcast-worker` never delivers new replies to them. The panel silently goes stale until it is closed and reopened.

`channelThreadFanOut` builds the channel recipient set as *reply sender ∪ parent author ∪ `thread_rooms.replyAccounts` ∪ history-gated @-mentions*, then publishes once per recipient on `chat.user.{account}.event.room`. A viewer is in none of those sets.

The root cause is that one set does two jobs: **notification interest** (who wants to be told later) and **delivery interest** (who is looking right now). Followership answers the first question correctly and the second one wrong.

## Approach

Publish channel thread events a second time onto a thread-scoped subject derived from the room subject. A client subscribes for exactly as long as the panel is open, so NATS's own subscription table becomes the viewer registry.

```
chat.room.{roomID}.thread.{parentMessageID}.event         # cross-site rooms
chat.local.room.{roomID}.thread.{parentMessageID}.event   # same-site rooms
```

The per-follower fan-out is unchanged. This is the shape Slack described moving to ("clients subscribe to the events relevant in the current view, and change subscriptions when users switch views") and that Discord implements as guild/thread subscriptions.

Why this shape over the alternatives considered:

- **No new state.** NATS tracks interest and reaps it on disconnect — no TTL, no heartbeat, no sweeper, no cross-site replication. A Valkey viewer registry would have needed all four plus a new federation lane.
- **No new hot-path I/O.** `handleThreadCreated` already holds `roommetacache.Meta`, which already carries `CrossSite`/`CrossSiteAt`; routing reuses `roomRouteGlobals`, so a same-site room stays on the leaf-filtered local namespace and never touches a gateway.
- **No auth change.** Client JWTs already grant `sub` on `chat.room.>` and `chat.local.room.>`.
- **Zero cost when unwatched.** A publish with no interest is dropped at the server.

Rejected: auto-following on panel open (corrupts push targeting, the Threads tab, and `thread_rooms.minUserLastSeenAt` — an accidental subscriber who never reads pins the read floor and breaks thread read receipts); broadcasting every reply room-wide (a 5,000-member channel would receive and decrypt every reply in a three-person thread).

Channel rooms only. DM and botDM thread replies already reach every member.

## Changes

[...]

PR #357, opened 2026-08-23
feat(broadcast-worker): per-thread viewer lane for non-follower thread delivery

yenta:
## Problem

Opening a channel thread you have never participated in gives you a static transcript. The reply count ticks up; the replies never arrive.

Channel thread replies fan out **per-account** to a recipient set built by `threadFanOutAccounts` — the reply author, the parent author, thread followers (`thread_rooms.replyAccounts`), and history-gated @-mentioned accounts. Everyone else in the room receives only `ThreadMetadataUpdatedEvent`, a content-free badge. Nothing makes a viewer a follower: there is no follow RPC, `Get Thread Messages` emits nothing, and `Mark Thread as Read` is a no-op for a non-subscriber.

The same gap applied to edits and deletes, so a viewer could sit on content that had since been corrected or removed.

## Approach

Channel thread creates, edits, and deletes are now **also** published to a new per-thread subject:

```
chat.room.{roomID}.thread.{parentMessageId}.event
chat.local.room.{roomID}.thread.{parentMessageId}.event   (same-site rooms)
```

A client subscribes while a thread pane is open and unsubscribes when it closes. That is the entire lifecycle — no RPC, no server-side state, no TTL, no heartbeat. Subscribing does **not** make you a follower: no thread subscription is created, no unread accrues, and the thread does not enter your thread list.

`ThreadEventTargets` delegates to the existing `roomRouteGlobals`, so the lane honours `ROOM_SUBJECT_MODE`, the sticky `crossSite` flag, and the post-flip locality grace window identically to `.event` and `.event.member`.

The pre-existing per-account follower lane is **unchanged and still plaintext** — this is purely additive. DM and BotDM rooms are excluded; they already deliver thread events to every member.

Design doc: `docs/superpowers/specs/2026-08-23-thread-viewer-lane-design.md`

## Confidentiality

This is the part worth reviewing carefully. Client JWTs carry a wildcard `chat.room.>` subscribe grant, so room-lane confidentiality comes from the payload, not the subject. Thread replies were previously plaintext precisely because the per-account subject is JWT-scoped to one recipient; moving content onto the room namespace requires the payload-level lock.

- Create and edit lane copies go through `encryptRoomEvent` / `encryptEditedContent` when `ENCRYPTION_ENABLED=true`, clearing the corresponding plaintext field.
- An encryption failure returns before any publish — no path falls through to plaintext on the lane.
- Deletes carry no message content (`DeleteRoomEvent`, with `PreviewMessage` nil for `tshow=false` replies), so one payload serves both lanes with no encryption step.
- `mentions` / `mentionAll` are deliberately **retained** on the lane copy — the frontend renders mention styling from that resolved participant list, and both fields are already plaintext on `chat.room.{roomID}.event` for every ordinary message.

With `ENCRYPTION_ENABLED=false` thread content is plaintext on an openly-subscribable lane — the same exposure ordinary channel messages already have in that configuration, but a reduction for thread content specifically. A warning comment now sits beside that env var in both deploy compose files.

## Client-visible changes

- **Deduplicate by `message.id`.** A follower who opens the thread receives each event on both lanes. Required regardless — broadcast-worker consumes MESSAGES-CANONICAL with `AckExplicitPolicy`, so redelivery already made delivery at-least-once.
- **The lane carries only `tshow=false` replies.** `tshow=true` thread replies and their edits/deletes bypass the thread handlers entirely and ride `chat.room.{roomID}.event` as ordinary `new_message` / `message_edited` / `message_deleted`. A pane rendered from the lane alone would silently drop them. Documented in §4.2.
- No new event type, no new model struct, no NATS permission change — the subject sits under `roomBase`, inside the existing grant.

## Changes

| Area | What |
|---|---|
| `pkg/subject` | `ThreadEvent`, `ThreadEventTargets` |
| `broadcast-worker/nats_metrics.go` | new `thread_lane` room kind — the lane is one publish to an unobservable audience, so it cannot share `thread`, whose fanout/deliveries pair is documented as directly comparable |
| `broadcast-worker/handler.go` | `publishThreadLaneEvent` + wiring into all three thread handlers |
| `.semgrep/room-subject.yml` | `subject.ThreadEvent` added to the routing guard |
| `docs/client-api.md` + both derived views | new subject, two-lane delivery, dedupe requirement, subscribe lifecycle |
| deploy compose files | encryption warning beside `ENCRYPTION_ENABLED` |

Also removes the empty-fan-out early return in the create and edit handlers: a bot-authored reply produces an empty per-account set, and a viewer with the pane open must still receive it.

## Testing

TDD throughout. `make test` green across every package; `make lint` clean; `gosec` and the repo's own offline semgrep rules clean.

Coverage includes both lanes per handler, encryption on and off, the encrypt-failure branch asserting **nothing** reaches the lane, failure-swallow at all three publish sites, empty fan-out still publishing to the lane, and DM/BotDM publishing nothing to it. Six pre-existing count assertions were bumped, each paired with a new assertion naming the lane record; no existing DM/BotDM test was modified — that is the regression guarantee.

> [!IMPORTANT]
> **The registry-backed SAST checks could not run in the authoring environment** — the sandbox network policy returns 403 for `semgrep.dev` and `vuln.go.dev`, so `make sast-semgrep` and `make sast-vuln` could not fetch their configs. The repo's own `.semgrep` rules and `gosec` were run and are clean. CI's blocking `sast` job needs to cover the rest before merge.

https://claude.ai/code/session_01V2KmtQP7RtDj5zGzPrqvSw

---
_Generated by [Claude Code](https://claude.ai/code/session_01V2KmtQP7RtDj5zGzPrqvSw)_

## Summary by CodeRabbit

- **New Features**
  - Added live per-thread event delivery for channel thread viewers, including new, edited, and deleted replies.
  - Thread updates now remain available even when there are no per-account recipients.
  - Added support for encrypted thread-lane messages and appropriate routing across local and global streams.

- **Documentation**
  - Documented subscription behavior, encryption, duplicate-event handling, and thread viewer lifecycle.
  - Clarified that direct room replies and private-message threads use existing delivery paths.

- **Bug Fixes**
  - Prevented thread-lane delivery failures from interrupting standard message delivery.

### DIFF
--- a/.semgrep/room-subject.yml
+++ b/.semgrep/room-subject.yml
@@ -7,7 +7,7 @@ rules:
       publishRoomEvent / publishMemberEvent helper, which fans out over
       subject.RoomEventTargets / RoomMemberEventTargets(roomID, crossSite,
       crossSiteAt, mode, now). Passing a subject built inline by subject.RoomEvent /
-      RoomMsgStream / RoomMetadataUpdate / RoomMemberEvent to a publish call skips the
+      RoomMsgStream / RoomMetadataUpdate / RoomMemberEvent / ThreadEvent to a publish call skips the
       ROOM_SUBJECT_MODE global/local routing, so same-site rooms silently miss
       delivery once local mode is enabled (and dual mode loses its global
       safety-net publish). See pkg/subject.RoomEventTargets and each service's
@@ -32,3 +32,6 @@ rules:
           - pattern: $X.Publish($CTX, subject.RoomMemberEvent(...), ...)
           - pattern: $X.publishCore($CTX, subject.RoomMemberEvent(...), ...)
           - pattern: $X.publish($CTX, subject.RoomMemberEvent(...), ...)
+          - pattern: $X.Publish($CTX, subject.ThreadEvent(...), ...)
+          - pattern: $X.publishCore($CTX, subject.ThreadEvent(...), ...)
+          - pattern: $X.publish($CTX, subject.ThreadEvent(...), ...)
--- a/broadcast-worker/deploy/bot/docker-compose.yml
+++ b/broadcast-worker/deploy/bot/docker-compose.yml
@@ -36,6 +36,9 @@ services:
       # true like the user variants — false means verify-and-exit, and the bot
       # streams' creator (bot-message-handler) is not in compose.services.yaml.
       - BOOTSTRAP_STREAMS=${BOOTSTRAP_STREAMS:-true}
+      # Off by default: channel and thread message content (including the thread
+      # viewer lane) goes plaintext on chat.room.> subjects, which any authenticated
+      # user's JWT can subscribe to.
       - ENCRYPTION_ENABLED=${ENCRYPTION_ENABLED:-false}
       # Dev-only X-Debug verbose logging. DEBUG_LOG_PAYLOADS gates full
       # request/reply/consumed body capture (pair with the X-Debug-Payload header).
--- a/broadcast-worker/deploy/user/docker-compose.yml
+++ b/broadcast-worker/deploy/user/docker-compose.yml
@@ -33,6 +33,9 @@ services:
       - VALKEY_ADDRS=${VALKEY_ADDRS:-valkey:6379}
       - ROOM_META_L2_TTL=${ROOM_META_L2_TTL:-15m}
       - BOOTSTRAP_STREAMS=${BOOTSTRAP_STREAMS:-true}
+      # Off by default: channel and thread message content (including the thread
+      # viewer lane) goes plaintext on chat.room.> subjects, which any authenticated
+      # user's JWT can subscribe to.
       - ENCRYPTION_ENABLED=${ENCRYPTION_ENABLED:-false}
       # Dev-only X-Debug verbose logging. DEBUG_LOG_PAYLOADS gates full
       # request/reply/consumed body capture (pair with the X-Debug-Payload header).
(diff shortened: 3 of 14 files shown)


yenta:
Closing — this duplicates #349, which already delivers this feature on `main`.

This branch was cut from `5e6f365`, which predates #349's merge, so it was designed and built against a stale `main`. The two implementations produce a **byte-identical subject string**:

```go
// #349 (RoomThreadEvent)  roomBase(roomID, global) + ".thread." + parentMessageID + ".event"
// this PR (ThreadEvent)   roomBase(roomID, global) + ".thread." + parentMessageID + ".event"
```

Merging would have shipped both: `pkg/subject/subject.go` and `nats_metrics.go` both auto-merge **cleanly**, leaving two builders for one subject and two metric paths, so every viewer would receive every thread event twice. The duplication is in the files git does *not* flag as conflicting.

#349 is also the better implementation — it has a kill switch (`THREAD_VIEW_SUBJECT_ENABLED`), integration tests, the frontend wiring, and documentation covering edge cases this PR never considered (subscribe-before-fetch, placeholder vs. decrypted body precedence, `editedAt` ordering, per-thread handler serialization).

## One finding worth keeping

The review work here did surface a real gap in `main` that #349 did not cover:

**`.semgrep/room-subject.yml` does not guard `subject.RoomThreadEvent`.** The `room-subject-publish-must-route` rule enumerates the subject builders that must not be passed inline to a publish call, precisely so a caller cannot bypass `ROOM_SUBJECT_MODE` global/local routing. `RoomThreadEvent` is missing from that list, so `x.Publish(ctx, subject.RoomThreadEvent(...), ...)` would pass SAST today — and a same-site room would silently miss thread delivery once local mode is enabled.

The fix is three `pattern-either` entries (`Publish` / `publishCore` / `publish` × `subject.RoomThreadEvent`) plus adding the builder to the rule's `message:` text, mirroring the existing entries.

Branch `claude/thread-metadata-event-5065r0` is left in place for reference.

---

PR #339, opened 2026-08-21
fix(loadgen): bound what one operation costs the reconciler, and say when a run cannot pay it

flywindy:
Follow-up to #330. That PR bounded retained recipient evidence and stopped the memory growth; this PR fixes the reconciliation deficit that left `loadgen_failure_inflight` pinned near `SEND_RATE x RECONCILE_DEADLINE` and made fault-window results predominantly `unverified`.

## What changed

- Added deadline-aware exponential backoff for answered-but-pending history and search probes, including a final authoritative probe at the deadline. Unanswered dependency calls keep the flat retry interval.
- Modelled reconciliation capacity from the query-mode observers actually enabled. Event-mode recipient claims no longer consume read slots or one callback each; free claims are drained in a bounded burst and refunded credit cannot accumulate.
- Added `loadgen_failure_reconcile_claims_total{outcome}` with `advanced`, `retried`, `unavailable`, `deferred`, `idle`, and `failed`, plus `loadgen_failure_reconcile_lag_seconds`. The bundled dashboard exposes claim outcomes, p99 lag, and startup invalidations.
- A run below the capacity floor still starts, but is durably marked inconclusive with `reconcile_capacity`. A deadline outside the resolvable lag range is similarly marked with `reconcile_lag_range`. A search settle window that cannot finish before the deadline remains a startup error.
- Added a versioned `invalidated` WAL record. Replay and compaction preserve every distinct invalidation in order while `InvalidReason` retains the first cause.
- Pending invalidations must settle before later evidence is accepted. Started and invalidated records wait for the group-commit fsync barrier, startup refuses an invalidation the journal did not durably accept, and shutdown refuses to close over outstanding durability debt.
- Added a durability watchdog. A debt that survives a full grace interval cancels the workload with a distinct cause and exits with code 2; an external cancellation keeps its existing exit semantics.

## Capacity and operational impact

No configuration change is required for the current deployment:

```text
demand   = SEND_RATE 55 x 1 query step          = 55 claims/s
capacity = READ_RATE 110 x RECONCILE_SHARE 0.75 = 82.5 claims/s
```

The recipient observer is event-mode, so it is not part of the query-step count. Because those local finalizations no longer consume read actions, the workload now delivers more of the configured read traffic to the system under test; comparisons with older runs must account for that baseline change.

The search observer remains refused at startup until soak payloads contain a per-message searchable marker. Its reconciliation scheduling and accounting paths are covered here so they are correct before it is enabled.

## Ledger epoch and deployment

[...]

PR #359, opened 2026-08-23
fix(loadgen): keep the soak generating traffic and telling the truth through a MongoDB outage

flywindy:
The MongoDB failure campaign runs a soak workload while the MongoDB team performs the faults. For that to produce evidence, loadgen has to stay alive through the outage and has to keep saying "I could not look" rather than "it was lost". Today it does neither.

## The one that would have ended the campaign

`runSoakHeartbeat` returned on the **first** `TouchHeartbeat` error. That error is not `context.Canceled`, so the goroutine set `fatalErr` and called `cancel()` — which cancels the run context every lane dispatches on.

```
soak_workload.go   runSoakHeartbeat returns on first error
                   → fatalErr = err; cancel()
soak_store.go      TouchHeartbeat is a Mongo UpdateOne
                   HeartbeatInterval defaults to 30s
```

MongoDB goes down, and roughly 30 seconds later the entire workload stops. The campaign would have observed nothing for the whole outage.

## What this branch changes

**1. Survive transient heartbeat outages**

`TouchHeartbeat` had two failures that were indistinguishable by type: an `UpdateOne` error (Mongo unreachable — state unknown) and `MatchedCount == 0` (the manifest is authoritatively gone or no longer running). Swallowing both keeps loadgen hammering a torn-down topology; failing on both is what kills the campaign.

`errSoakRunNotActive` separates them. Only authoritative proof that the run is inactive is fatal; every other storage error is unknown state, so the loop goes degraded and retries every 5s with a 5s per-attempt timeout, no overlap. There is deliberately no allowlist of "connection errors" — stepdown, server selection, not-primary and timeout take too many shapes to enumerate, and missing one would resurrect the original bug.

**2. Stop before the lease it holds can go stale**

Surviving an outage creates the opposite hazard: seed and teardown read `SOAK_HEARTBEAT_STALE_AFTER` as the active-run lease, so a workload that retries forever can still be dispatching when teardown decides the run was abandoned and starts deleting topology.

The heartbeat loop now carries the lease. It stops at `lastPersistedHeartbeat + staleAfter - oneHeartbeatInterval`, returns `errSoakHeartbeatLeaseAtRisk`, and that cancels the run. Each attempt's timeout is clamped to the remaining budget so no single attempt can straddle the boundary, and startup rejects a lease that cannot hold the margin (`staleAfter >= 2*interval + 5s`, `errSoakHeartbeatLeaseInvalid`).

Deciding to stop is not the same as having stopped, so on that path — and only that path — the lane drain is bounded to half the margin, leaving the other half for the rest of shutdown. If lanes are still in flight when it expires, loadgen logs which ones, records a `lease_abort` ledger invalidation so the interval reads as inconclusive rather than as ordinary `unverified`, and exits without the graceful cleanup it can no longer afford. Ordinary shutdown keeps the full unbounded drain.

**3. Bound room-state reconciliation**

The lane that reads MongoDB had no timeout of its own, so a probe rode the driver's 30s server-selection default. At the room reconcile rate that is ~300 probes in flight against a global in-flight budget of 256 — an outage would stall every other lane, not just this one.

`Verify` now takes one 5s budget covering the room-service RPC *and* the authoritative primary read, with a nested 2s cap on the RPC phase so a hung service cannot consume the fallback's budget. `ProbeQuarantine` and the room-create ownership write share the same bound. Every timeout still resolves to `unknown`/`unverified` — no timeout can manufacture an `absent` or a `missing_after_deadline`.

The room lane also picks up the backoff #339 introduced for the message lane, and schedules from the time verification *completed* rather than the time it was claimed. The distinction #339 drew is preserved in the names — `releaseProbe` backs off a pending effect, `releaseFailedCall` keeps the flat interval for a call that failed, capped at the operation deadline.

**4. Make the lease writes durable**

The fence is only as strong as the write behind it. Nothing in `pkg/mongoutil` or the soak store set a write concern, so manifest writes inherited the deployment default — majority on most deployments, but not on a PSA topology or a URI carrying `w:1`, where an acknowledged heartbeat can be rolled back by a failover. All three manifest writes now go through one collection handle carrying majority with journalling, so a write added later cannot quietly drop the concern.

**5. Expose Mongo and heartbeat health**

Without these, "the service cannot reach Mongo" and "loadgen cannot reach Mongo" look identical, and the second one has to invalidate the interval rather than indict the system.

| Metric | |
|---|---|
| `loadgen_mongo_up` | Latest bounded `Ping` to the primary, 5s cadence, 2s timeout |
| `loadgen_mongo_probe_timestamp_seconds` | Updated on every probe regardless of outcome |
| `loadgen_mongo_probe_attempts_total{outcome}` | `success` / `error` — a blip between two scrapes is not erased |
| `loadgen_soak_heartbeat_degraded` | |
| `loadgen_soak_heartbeat_success_timestamp_seconds` | |
| `loadgen_soak_heartbeat_attempts_total{outcome}` | `success` / `error` / `not_active` |

Each pair is emitted by one collector that loads its atomic snapshot once per scrape, so the two halves cannot disagree about which probe they describe. The timestamps are the point: a single gauge written by a stuck goroutine freezes at its last healthy value, which is exactly the condition it exists to expose. The counters carry what the gauges cannot — an outage shorter than the scrape interval. Transition logs mark degraded and recovered without one line per probe.

**6. Exercise a real outage**

A dedicated Mongo container on an explicitly bound host port, stopped and started from outside the process. A spike confirmed the default dynamic mapping does not survive `stop`/`start`, which would have made the test fail for the wrong reason; a fixed binding holds, and the test asserts that as an explicit precondition so a regression reports itself clearly instead of looking like a recovery failure.

The scenario walks healthy → stop → degraded → recovery → inactive-run-fatal, and the recovery assertions run against **the same `mongo.Client`** throughout. The client keeps the driver's 30s server-selection default so the elapsed-time assertion actually proves the 5s cap is doing the work.

## Campaign configuration this requires

Chart defaults are unchanged; the campaign applies overrides, documented in `deploy/k8s/README.md`:

- `SOAK_HEARTBEAT_STALE_AFTER=30m` — the destructive-operation lease, sized well beyond the planned ~2m outage. With the default 30s interval the workload stops dispatching about 29m30s after its last persisted heartbeat.
- `SOAK_ROOM_READ_RATE=30`, `SOAK_ROOM_RECONCILE_READ_SHARE=0.5` — 15/s of reconcile capacity against 8.05/s of steady demand, which drains a 2-minute backlog in ~139s instead of ~495s. This only holds while the outage stays well inside `SOAK_RECONCILE_DEADLINE=10m`; anything that ages past the deadline expires `unverified` no matter what the read rate is.

There is no teardown force flag. The README documents the escape hatch — a teardown-only lease override, to be used only after confirming no loadgen Pod exists and the heartbeat has stopped advancing.

## Verification

Run on the current head:

```
make fmt                          pass
make lint                         0 issues
make test SERVICE=tools/loadgen   pass (-race, 271.8s)
```

CI covers `test`, `test-integration (tools)`, `lint` and `sast`. `make coverage-loadgen-soak` and `make validate-loadgen-k8s` were **not** re-run against the last three commits — the environment they were pushed from has neither Docker nor Helm. They passed on the earlier head, and the commits since then add one production helper plus tests.

## Scope

Not in this PR, and each is a deliberate decision rather than an oversight:

- **A server-side heartbeat timestamp.** The stop boundary uses the timestamp loadgen persisted and its own clock, while seed and teardown judge staleness against theirs, so the margin also has to cover the difference between those clocks. One heartbeat interval assumes they agree to within that interval, which ordinary NTP beats by orders of magnitude. Removing the assumption means a `$$NOW` write, a committed timestamp returned to the loop, a monotonic local timer, and matching changes in seed and teardown — the lease-fencing work this branch is deliberately not doing. The assumption is now written down where the margin is configured.
- **A fencing token checked by destructive operations**, so a stale heartbeat is not on its own sufficient authority to delete.
- **Room-create ownership reporting.** When the ownership write fails after the operation deadline the lane logs, counts `untracked{reason="ownership"}`, and still records `good`. That predates this branch, and `good` is the honest verdict about the system: the room was created and read back. The lost fact is loadgen's own bookkeeping, and the dashboard contract already treats `untracked` as invalidating absence claims.
- **`SOAK_LEDGER_EXPIRE_BATCH`**, which needs WAL characterization first and addresses a risk that predates this work.
- **A `values-mongo-campaign.yaml`** collecting the overrides above so they cannot be forgotten at deploy time.

This validates connectivity loss and recovery against a single node. It does **not** validate multi-node primary election or majority loss — those need the staging replica set.

## Summary by CodeRabbit

- **New Features**
  - Added MongoDB availability and soak-heartbeat health metrics, including freshness, recovery, and attempt outcomes.
  - Added Grafana dashboard visibility for MongoDB and heartbeat validity.
  - Improved recovery with bounded verification, retries, and backoff during storage outages.
  - Added lease-aware shutdown and evidence invalidation for incomplete operations.

- **Documentation**
  - Added guidance for configuring and safely tearing down MongoDB outage campaigns.
  - Expanded observability contracts with health, freshness, and correctness indicators.

- **Tests**
  - Added coverage for outage recovery, heartbeat transitions, probing, retries, deadlines, and reconciliation.


flywindy:
Addressed the review findings on the rebased head `ffc38710`:

- Kept the pre-call timestamp only for `ClaimDueLanes`, then refreshed time after `Verify` and after `claimCreatedRoom` for deadline checks, retry scheduling, and observation timestamps.
- Added regressions that advance the injected clock by five seconds during verification and ownership persistence. Both failed before the fix and now pass.
- Corrected the MongoDB document's out-of-scope reference from section 5 to section 6.
- Updated the PR verification section with the post-rebase results.

Validation: full `tools/loadgen` race suite, soak coverage (82.72% overall / 90.26% core), uncached external Mongo stop/start integration, Helm/K8s validation, and repository lint all pass. A fresh exact-head review found no new actionable issues. CodeRabbit could not run a new review because its review limit is temporarily exhausted; the post-push message contains no new finding.

flywindy:
@coderabbitai review

julianshen:
**11 findings · 6 unique issues · 6 inline comments posted** — 🔴 8 blocking · 🟠 3 warning

### Disputed

- `tools/loadgen/soak_workload.go:858` (`cassandra`) — The affected Cassandra mutation is the optional teardown truncation, which can race ongoing message inserts and history reads. `TouchHeartbeat` persists a loadgen-supplied timestamp, this loop computes `stopAt` using the loadgen clock, and seed/teardown evaluates staleness with another process's clock. With the documented 30-minute lease and 30-second margin, a teardown node 45 seconds ahead can declare the heartbeat stale after 29m15s of real time while loadgen continues dispatching until 29m30s. Teardown may therefore truncate live Cassandra data for 15 seconds. Use a Mongo server-generated committed heartbeat timestamp and evaluate expiry against the same server clock or an atomic server-side predicate; use monotonic elapsed time for the local conservative stop timer. Documenting an NTP assumption does not enforce the destructive-operation fence.
- `tools/loadgen/soak_workload.go:858` (`distributed-system`) — The persisted heartbeat and `stopAt` both use loadgen's wall clock, while seed or teardown evaluates staleness using another process's clock. With the documented 30-minute lease and 30-second shutdown margin, a loadgen node 45 seconds behind writes a heartbeat already 45 seconds old to teardown. Teardown therefore authorizes destructive cleanup after 29m15s of real time, while loadgen continues dispatching until 29m30s, creating a 15-second live-teardown window; a backward clock adjustment creates the same interleaving. The README documents this assumption but does not enforce it, so the violation remains. Use a Mongo server-generated heartbeat timestamp and server-clock expiry predicate, return the committed lease time to the loop, and use monotonic elapsed time for the local conservative stop timer.
- `tools/loadgen/soak_store.go:254` (`mongodb`) — `TouchHeartbeat` persists the loadgen node's wall-clock value, while seed/teardown compare it with another node's clock. With the documented 30-minute lease and 30-second shutdown margin, a loadgen clock 45 seconds behind lets teardown declare the manifest stale after 29m15s, while loadgen continues dispatching until 29m30s. Majority journaling and `$max` do not reconcile these clock domains, so teardown can delete topology for a live workload. Use a Mongo server-generated heartbeat timestamp and evaluate expiry against the same server clock or an atomic server-side predicate; return the committed timestamp and use monotonic elapsed time for the local shutdown timer.
- `tools/loadgen/soak_main.go:492` (`tgd-review`) — On the lease-risk path where lanes drain successfully, this defer receives a fresh half-interval duration rather than the remaining time to an absolute shutdown deadline. The lane drain may already consume half an interval, and later-registered defers shut down observers, the ledger, and metrics before this drain starts. NATS can therefore flush traffic after the manifest becomes stale. If the drain then times out, its `lease_abort` callback runs only after the ledger and metrics endpoint have closed, so the invalidation cannot be durably recorded or scraped. Carry an absolute lease deadline through cleanup and perform the bounded NATS drain and bounded invalidation before closing the ledger and metrics server.
- `tools/loadgen/soak_workload.go:858` (`tgd-review`) — The stop boundary is derived from a client-generated persisted timestamp and loadgen's wall clock, while teardown evaluates staleness using another process's clock. With the documented 30-second margin, a loadgen node 45 seconds behind lets teardown classify the manifest as stale 15 seconds before loadgen stops dispatching, authorizing destructive cleanup against a live workload. Documentation of the clock-synchronization assumption does not enforce the lease invariant. Use a Mongo server-generated timestamp and server-clock expiry predicate, returning the committed time to a monotonic local shutdown timer.

### Related work

- [PR #339](https://github.com/hmchangw/newchat/pull/339) — fix\(loadgen\): bound what one operation costs the reconciler, and say when a run cannot pay it (merged)

### 🔗 Findings that share one root cause (2)

- **The deferred NATS drain can outlive the lease and lose invalidation**
  - `tools/loadgen/soak_main.go:494` (`distributed-system`) — The deferred NATS drain can outlive the lease and lose invalidation
  - `tools/loadgen/soak_main.go:494` (`cassandra`) — The bounded NATS drain runs after its deadline and ledger close.
  - `tools/loadgen/soak_main.go:1220` (`performance-and-concurrency`) — Lease-risk cleanup still uses a reset, non-global timeout.
  - `tools/loadgen/soak_main.go:496` (`silent-failure`) — NATS failures are invalidated after the ledger is closed.
  - `tools/loadgen/soak_main.go:592` (`nats`) — NATS loss is invalidated after the durable ledger is closed.
  - `tools/loadgen/failure_nats.go:489` (`silent-failure`) — A stalled invalidation prevents the NATS connection from closing.
- **The NATS drain budget still starts too late.**
  - `tools/loadgen/soak_main.go:492` (`nats`) — The NATS drain budget still starts too late.
  - `tools/loadgen/failure_nats.go:480` (`nats`) — A drain-timeout close is treated as a successful flush.

### Local review memory

- `tools/loadgen/soak_workload.go:893` — @flywindy: Fixed in ac91245 . Confirmed the premise first: neither pkg/mongoutil nor the soak store sets a write concern anywhere, so all three manifest writes inherited the deployment default — and your PSA caveat is the exact case where that def… ([thread](https://github.com/hmchangw/newchat/pull/359#discussion_r3843005773))
- `tools/loadgen/soak_workload.go:858` — @flywindy: The mechanism is real and your arithmetic checks out: stopAt is computed from the timestamp loadgen persisted and its own clock, teardown judges staleness against its own, so the margin has to cover the difference between them as well as… ([thread](https://github.com/hmchangw/newchat/pull/359#discussion_r3843005758))
- `tools/loadgen/soak_roommember.go:427` — @flywindy: The fall-through is exactly as you describe, but two things argue against changing it here. It predates this branch. git show 44f11d5:tools/loadgen/soak_roommember.go has the same structure byte for byte — claim error past the deadline l… ([thread](https://github.com/hmchangw/newchat/pull/359#discussion_r3843005764))
- `tools/loadgen/soak_store.go:248` — @flywindy: Fixed in 93b6cf1 . TouchHeartbeat now writes through $max , so a late arrival carrying an older timestamp changes nothing and the write order stops mattering. Worth naming why this one bites harder than it looks: the whole point of thi… ([thread](https://github.com/hmchangw/newchat/pull/359#discussion_r3847280490))
- `tools/loadgen/soak_main.go:491` — @flywindy: Fixed in 845b400 . You are right that fastLeaseAbort only engages when the lane drain overruns, so the common case — lanes drain in time — fell through to ordinary cleanup and nc.Drain() took natsutil's 10s against a 15s margin that e… ([thread](https://github.com/hmchangw/newchat/pull/359#discussion_r3844963599))
- `tools/loadgen/soak_workload.go:351` — @flywindy: Checked this against the code and the premise does not hold: heartbeat failures do not go through setFatal . setFatal has exactly one caller — the lane action at soak_workload.go:407 , which is where the StopOnActionError policy belo… ([thread](https://github.com/hmchangw/newchat/pull/359#discussion_r3844963596))
- `tools/loadgen/soak_workload.go:332` — @flywindy: Same as the :351 thread — the premise does not hold. setFatal is called from one place, the lane action at soak_workload.go:407 ; the heartbeat goroutine sets fatalErr and calls cancel() directly, gated only on the error not being… ([thread](https://github.com/hmchangw/newchat/pull/359#discussion_r3844963603))

<details>
<summary>📒 Files reviewed (31)</summary>

* `Makefile`
* `docs/load-testing/failure/mongodb.md`
* `docs/load-testing/loadgen/dashboard-contract.md`
* `tools/loadgen/README.md`
* `tools/loadgen/deploy/grafana/dashboards/loadtest.json`
* `tools/loadgen/deploy/k8s/README.md`
* `tools/loadgen/failure_dashboard_contract_test.go`
* `tools/loadgen/failure_nats.go`
* `tools/loadgen/failure_nats_test.go`
* `tools/loadgen/metrics.go`
* `tools/loadgen/mongo_outage_recovery_integration_test.go`
* `tools/loadgen/soak_config_test.go`
* `tools/loadgen/soak_mongo_probe.go`
* `tools/loadgen/soak_mongo_probe_test.go`
* `tools/loadgen/soak_reconcile_backoff.go`
* `tools/loadgen/soak_roommember.go`
* `tools/loadgen/soak_roomops_test.go`
* `tools/loadgen/soak_roomverify_test.go`
* `tools/loadgen/soak_store_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

tools/loadgen/soak_workload.go
```diff
@@ -610,27 +641,128 @@ func stopSoakRun(
...
+			recordSoakHeartbeatAttempt(
+				observer, soakHeartbeatNotActive, false, completedAt,
+			)
+			return fmt.Errorf("update soak heartbeat: %w", err)
+		case err != nil:
+			recordSoakHeartbeatAttempt(observer, soakHeartbeatError, true, completedAt)
+			if !degraded {
+				degraded = true
+				degradedAt = completedAt
+				slog.Warn("soak heartbeat entered degraded state",
```
julianshen, line 910: _🏷️ lease-safety_ | _🔴 Blocking_ | _`distributed-system`, `performance-and-concurrency`, `tgd-review` (3 rules)_

**Heartbeat retries let an unfenced workload outlive its lease.**

On every transient heartbeat error this loop retries indefinitely while the workload continues dispatching. A concrete failure path is: loadgen loses Mongo connectivity while services and a teardown job remain connected; its last heartbeat ages past `SOAK_HEARTBEAT_STALE_AFTER`; teardown therefore treats the running manifest as abandoned and starts deleting topology; loadgen continues issuing work and may later recover concurrently with teardown. The documentation's larger campaign lease only postpones this race and cannot make an unexpectedly long partition safe. Pass the lease duration into this loop and propagate an error that cancels the workload before the last successful heartbeat can become stale, leaving a safety margin for shutdown. A fencing token checked by destructive operations would provide stronger protection.

<details>

- **`performance-and-concurrency`**:722: Every heartbeat error now retries indefinitely while the workload continues dispatching. If Mongo remains unavailable longer than `SOAK_HEARTBEAT_STALE_AFTER`, the persisted running manifest becomes stale and can authorize teardown even though loadgen is still active—the exact live-workload/teardown race described in the updated Mongo campaign documentation. A prolonged outage or unrecoverable authentication/configuration error therefore permits topology deletion under active traffic. Pass the configured lease duration into the heartbeat loop and return a fatal error (so `Run` cancels the workload) before the last successful persisted heartbeat reaches that age, with a small shutdown margin. Campaigns can still use the documented longer lease to tolerate planned outages.
- **`tgd-review`**:712: Every non-`errSoakRunNotActive` failure now retries indefinitely while workload dispatch continues. Once the manifest's last successful heartbeat exceeds `heartbeatStaleAfter`, seed/teardown considers the run stale even though this process may still be active—the exact destructive-operation race documented in `docs/load-testing/failure/mongodb.md`. A failed campaign recovery or permanent error such as lost authorization can therefore outlive even the 30-minute overlay. Pass the configured lease into heartbeat supervision and terminate/cancel the workload before the last successful manifest heartbeat becomes stale; add a test covering an unrecovered failure crossing that deadline.

</details>

<details>

```
In `tools/loadgen/soak_workload.go` around line 720: On every transient heartbeat error this loop retries indefinitely while the workload continues dispatching. A concrete failure path is: loadgen loses Mongo connectivity while services and a teardown job remain connected; its last heartbeat ages past `SOAK_HEARTBEAT_STALE_AFTER`; teardown therefore treats the running manifest as abandoned and starts deleting topology; loadgen continues issuing work and may later recover concurrently with teardown. The documentation's larger campaign lease only postpones this race and cannot make an unexpectedly long partition safe. Pass the lease duration into this loop and propagate an error that cancels the workload before the last successful heartbeat can become stale, leaving a safety margin for shutdown. A fencing token checked by destructive operations would provide stronger protection.

```

</details>

tools/loadgen/soak_mongo_probe.go
```diff
@@ -0,0 +1,173 @@
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case _, ok := <-ticks:
+			if !ok {
+				return nil
+			}
+			_ = probe.Probe(ctx)
+		}
+	}
```
julianshen, line 120: _🎯 correctness_ | _🔴 Blocking_ | _`silent-failure`_

**Mongo probe failures disappear before the next scrape.**

The loop discards each Probe error, while the snapshot retains only the latest outcome. With 5-second probes and 30-second scrapes, a failed Ping followed by a successful Ping between scrapes is overwritten without any counter, timestamp, or log preserving the failure. The dashboard then sees only a fresh successful probe and can treat absence evidence from the blind interval as conclusive. Export a failure/attempt counter or last-failure timestamp and include failures over the evaluated window in the validity gate.

<details>

```
In `tools/loadgen/soak_mongo_probe.go` around line 95: The loop discards each Probe error, while the snapshot retains only the latest outcome. With 5-second probes and 30-second scrapes, a failed Ping followed by a successful Ping between scrapes is overwritten without any counter, timestamp, or log preserving the failure. The dashboard then sees only a fresh successful probe and can treat absence evidence from the blind interval as conclusive. Export a failure/attempt counter or last-failure timestamp and include failures over the evaluated window in the validity gate.

```

</details>

tools/loadgen/soak_roommember.go
```diff
@@ -417,27 +420,33 @@ func (l *soakRoomLanes) Reconcile(
 		// budget. Hold the operation open and try again while the deadline
 		// allows it.
```
julianshen, line 421: _🏷️ failure-bounds_ | _🟠 Warning_ | _`distributed-system`_

**The ownership write bypasses the new reconciliation timeout.**

The verifier and quarantine reads are now explicitly bounded, but the subsequent `AppendOwnedRooms` path still receives the long-lived reconciliation context. If room-service reports a created room while loadgen's Mongo path is partitioned, reconciliation reaches this call and can remain blocked for the driver's server-selection or socket timeout. During that interval the claim cannot be released and the worker cannot process other due operations, allowing the backlog to grow and operations to expire without evidence. Bound each ownership-write attempt as well, then use the existing flat retry path on timeout.

<details>

```
In `tools/loadgen/soak_roommember.go` around line 421: The verifier and quarantine reads are now explicitly bounded, but the subsequent `AppendOwnedRooms` path still receives the long-lived reconciliation context. If room-service reports a created room while loadgen's Mongo path is partitioned, reconciliation reaches this call and can remain blocked for the driver's server-selection or socket timeout. During that interval the claim cannot be released and the worker cannot process other due operations, allowing the backlog to grow and operations to expire without evidence. Bound each ownership-write attempt as well, then use the existing flat retry path on timeout.

```

</details>

julianshen (commented):
tGD inline review

tools/loadgen/metrics.go
```diff
@@ -375,6 +382,41 @@ func NewMetrics() *Metrics {
...
+		m.soakHeartbeatSuccessTimestamp,
+	)
+	m.SoakHeartbeatAttempts = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "loadgen_soak_heartbeat_attempts_total",
+			Help: "Soak manifest heartbeat attempts by bounded outcome (success, error, not_active).",
+		},
+		[]string{"outcome"},
+	)
+	m.MongoUp = prometheus.NewGaugeFunc(
```
julianshen: _🏷️ consistency_ | _🟠 Warning_ | _`mongodb`_

**Mongo validity gauges can expose different probe attempts.**

`loadgen_mongo_up` and `loadgen_mongo_probe_timestamp_seconds` are separate GaugeFunc collectors, and each independently loads the probe snapshot. Prometheus may collect them concurrently or with a probe completion between calls. If a failed Ping lands after `MongoUp` reads the previous successful snapshot but before `MongoProbeTimestamp` reads the new snapshot, one scrape reports `up=1` with a fresh failure timestamp. The documented gate then treats absence evidence as valid despite the latest Ping failing. The heartbeat degraded/timestamp pair has the same transition race. Use custom collectors that load each atomic snapshot once in `Collect` and emit both related gauges from that single snapshot.

<details>

```
In `tools/loadgen/metrics.go` around line 406: `loadgen_mongo_up` and `loadgen_mongo_probe_timestamp_seconds` are separate GaugeFunc collectors, and each independently loads the probe snapshot. Prometheus may collect them concurrently or with a probe completion between calls. If a failed Ping lands after `MongoUp` reads the previous successful snapshot but before `MongoProbeTimestamp` reads the new snapshot, one scrape reports `up=1` with a fresh failure timestamp. The documented gate then treats absence evidence as valid despite the latest Ping failing. The heartbeat degraded/timestamp pair has the same transition race. Use custom collectors that load each atomic snapshot once in `Collect` and emit both related gauges from that single snapshot.

```

</details>

tools/loadgen/soak_roommember.go
```diff
@@ -417,27 +420,33 @@ func (l *soakRoomLanes) Reconcile(
...
 		if operation.OperationType == failureOperationRoomCreate {
-			if err := l.claimCreatedRoom(ctx, &operation); err != nil {
-				if now.Before(operation.Deadline) {
-					return true, l.release(operation.ID, now)
+			claimErr := l.claimCreatedRoom(ctx, &operation)
+			completedAt = l.now().UTC()
+			if claimErr != nil {
+				if completedAt.Before(operation.Deadline) {
+					return true, l.releaseFailedCall(operation.ID, completedAt)
```
julianshen: _🏷️ reliability_ | _🟠 Warning_ | _`silent-failure`_

**Ownership retries can be scheduled beyond the operation deadline.**

When AppendOwnedRooms fails just before the deadline, this branch returns success after releaseFailedCall schedules the next attempt at completedAt + RetryInterval without capping it at the operation deadline. As the new deadline-walk test documents, work scheduled past the deadline can expire unverified instead of being claimed again. That skips the after-deadline log and countUntracked path, leaving the created room absent from teardown ownership with the failure hidden. Pass the deadline to the flat-retry helper and schedule the next attempt at min(completedAt + RetryInterval, operation.Deadline).

<details>

```
In `tools/loadgen/soak_roommember.go` around line 427: When AppendOwnedRooms fails just before the deadline, this branch returns success after releaseFailedCall schedules the next attempt at completedAt + RetryInterval without capping it at the operation deadline. As the new deadline-walk test documents, work scheduled past the deadline can expire unverified instead of being claimed again. That skips the after-deadline log and countUntracked path, leaving the created room absent from teardown ownership with the failure hidden. Pass the deadline to the flat-retry helper and schedule the next attempt at min(completedAt + RetryInterval, operation.Deadline).

```

</details>

tools/loadgen/soak_workload.go
```diff
@@ -610,27 +655,180 @@ func stopSoakRun(
...
+			"soak heartbeat lease requires stale threshold at least twice the shutdown margin plus the attempt timeout",
+		)
+	}
+	if lastSuccessAt.IsZero() {
+		return fmt.Errorf("soak heartbeat lease requires the last persisted heartbeat time")
+	}
+	lastSuccessAt = lastSuccessAt.UTC()
+	degraded := false
+	var degradedAt time.Time
+	for healthyTicks != nil || degraded {
```
julianshen, line 858: _🏷️ distributed-consistency_ | _🔴 Blocking_ | _`distributed-system`_

**Client wall clocks can violate the destructive lease margin**

The stop boundary is calculated from a client-supplied heartbeat timestamp and the loadgen process's `now()`, while seed or teardown evaluates staleness in another process. With the documented 30-minute lease and 30-second margin, if the loadgen node is 45 seconds behind the teardown node, its heartbeat is persisted 45 seconds old. Teardown can therefore consider it stale after 29m15s of real time, but loadgen does not stop dispatching until 29m30s, leaving a 15-second interval where destructive teardown is authorized against a live workload. A backward clock adjustment creates the same failure. Use a Mongo server-generated heartbeat time and evaluate expiry against the same server clock or an atomic server-side predicate; return the committed timestamp to the heartbeat loop and use monotonic elapsed time for its conservative stop timer. Add a test with differing worker and teardown clock offsets.

<details>

```
In `tools/loadgen/soak_workload.go` around line 724: The stop boundary is calculated from a client-supplied heartbeat timestamp and the loadgen process's `now()`, while seed or teardown evaluates staleness in another process. With the documented 30-minute lease and 30-second margin, if the loadgen node is 45 seconds behind the teardown node, its heartbeat is persisted 45 seconds old. Teardown can therefore consider it stale after 29m15s of real time, but loadgen does not stop dispatching until 29m30s, leaving a 15-second interval where destructive teardown is authorized against a live workload. A backward clock adjustment creates the same failure. Use a Mongo server-generated heartbeat time and evaluate expiry against the same server clock or an atomic server-side predicate; return the committed timestamp to the heartbeat loop and use monotonic elapsed time for its conservative stop timer. Add a test with differing worker and teardown clock offsets.

```

</details>

tools/loadgen/soak_roommember.go
```diff
@@ -417,27 +420,37 @@ func (l *soakRoomLanes) Reconcile(
...
 		// allows it.
 		if operation.OperationType == failureOperationRoomCreate {
-			if err := l.claimCreatedRoom(ctx, &operation); err != nil {
-				if now.Before(operation.Deadline) {
-					return true, l.release(operation.ID, now)
+			claimCtx, cancelClaim := context.WithTimeout(ctx, soakRoomStateTimeout)
+			claimErr := l.claimCreatedRoom(claimCtx, &operation)
+			cancelClaim()
+			completedAt = l.now().UTC()
+			if claimErr != nil {
```
julianshen, line 427: _🏷️ error-handling_ | _🔴 Blocking_ | _`silent-failure`_

**A failed ownership write can still terminalize the room create as good.**

If room verification succeeds shortly before the deadline but `claimCreatedRoom` times out or fails after it, `completedAt.Before(operation.Deadline)` is false. The code only logs and increments `untracked`, then falls through to `observe(...failureObservationGood...)`, which removes the operation and returns success. The ownership record is the durable input used for teardown, so the created room can be omitted from cleanup while the ledger reports a good result. Do not emit `good` unless ownership persistence succeeds; preserve the operation for durable retry or surface a terminal unverified/fatal failure, and add a test where the ownership call crosses the deadline.

<details>

```
In `tools/loadgen/soak_roommember.go` around line 427: If room verification succeeds shortly before the deadline but `claimCreatedRoom` times out or fails after it, `completedAt.Before(operation.Deadline)` is false. The code only logs and increments `untracked`, then falls through to `observe(...failureObservationGood...)`, which removes the operation and returns success. The ownership record is the durable input used for teardown, so the created room can be omitted from cleanup while the ledger reports a good result. Do not emit `good` unless ownership persistence succeeds; preserve the operation for durable retry or surface a terminal unverified/fatal failure, and add a test where the ownership call crosses the deadline.

```

</details>

tools/loadgen/soak_workload.go
```diff
@@ -610,27 +655,180 @@ func stopSoakRun(
...
+			}
+		}
+
+		stopAt := lastSuccessAt.Add(staleAfter - shutdownMargin)
+		remaining := stopAt.Sub(now().UTC())
+		if remaining <= 0 {
+			return soakHeartbeatLeaseError(lastSuccessAt, stopAt)
+		}
+		attemptCtx, cancel := context.WithTimeout(ctx, min(attemptTimeout, remaining))
+		err := store.TouchHeartbeat(attemptCtx, runID, at.UTC())
```
julianshen, line 893: _🏷️ durability_ | _🟠 Warning_ | _`mongodb`_

**Heartbeat acknowledgments do not guarantee a durable lease.**

The loop advances `lastSuccessAt` whenever `TouchHeartbeat` returns nil, but the manifest update relies on inherited MongoDB write concern. With `w:1`—including the implicit default for some PSA replica sets—the old primary can acknowledge the heartbeat and then lose it during failover. Loadgen would continue dispatching using its advanced in-memory lease while seed or teardown reads an older heartbeat from the new primary and may classify the live run as abandoned. Configure all manifest lease writes, including initial activation and `TouchHeartbeat`, with explicit majority and journaling guarantees; retain the bounded context and retry behavior. The new stop/start integration test cannot detect this because it restarts the same node rather than causing a replica-set election and rollback.

<details>

```
In `tools/loadgen/soak_workload.go` around line 759: The loop advances `lastSuccessAt` whenever `TouchHeartbeat` returns nil, but the manifest update relies on inherited MongoDB write concern. With `w:1`—including the implicit default for some PSA replica sets—the old primary can acknowledge the heartbeat and then lose it during failover. Loadgen would continue dispatching using its advanced in-memory lease while seed or teardown reads an older heartbeat from the new primary and may classify the live run as abandoned. Configure all manifest lease writes, including initial activation and `TouchHeartbeat`, with explicit majority and journaling guarantees; retain the bounded context and retry behavior. The new stop/start integration test cannot detect this because it restarts the same node rather than causing a replica-set election and rollback.

```

</details>

julianshen (commented):
tGD inline review

tools/loadgen/soak_workload.go
```diff
@@ -303,6 +348,14 @@ func (w *soakWorkload) Run(
 			w.store,
 			w.cfg.RunID,
 			heartbeatTicker.C,
+			soakHeartbeatAttemptTimeout,
```
julianshen, line 351: _🎯 concurrency_ | _🔴 Blocking_ | _`distributed-system`, `performance-and-concurrency`, `tgd-review` (3 rules)_

**Lease-risk heartbeat failures do not stop the workload.**

`soakWorkloadConfigFrom` leaves `StopOnActionError` false in production, while the heartbeat goroutine routes `runSoakHeartbeat` failures through `setFatal`, which returns immediately when that flag is false. Consequently, after Mongo remains unavailable until `errSoakHeartbeatLeaseAtRisk`, the heartbeat goroutine exits without recording the error or canceling `runCtx`; lane dispatch continues until the configured duration or an external signal, even after teardown may treat the manifest as stale. The new lease-risk test can hang on this same path. Keep `StopOnActionError` limited to lane-action failures, and route every non-cancellation heartbeat error through an unconditional fatal setter that records the error and cancels the workload.

<details>

- **`distributed-system`**:363: `soakWorkloadConfigFrom` leaves `StopOnActionError` false, yet heartbeat failures are propagated through `setFatal`, which discards every error when that flag is false. If Mongo remains unavailable until `runSoakHeartbeat` returns `errSoakHeartbeatLeaseAtRisk`, the heartbeat goroutine exits without cancelling `runCtx`; continuous-mode lanes keep dispatching after the persisted heartbeat becomes stale, allowing teardown to delete a live run's topology. Use an unconditional lifecycle-failure path that records the heartbeat error and cancels the workload, while retaining `StopOnActionError` only for lane action failures. Add a test with `StopOnActionError == false` that verifies lease risk still stops dispatch.
- **`tgd-review`**:353: `runSoakHeartbeat` reports `errSoakHeartbeatLeaseAtRisk` through `setFatal`, but `setFatal` returns unless `w.cfg.StopOnActionError` is true. `soakWorkloadConfigFrom` never sets that field, so the production workload keeps dispatching after the heartbeat goroutine exits instead of reaching the lease-abort path. The new lease-risk test also leaves the field false and therefore cannot complete as written. Lifecycle failures must cancel the workload independently of the action-error policy.

</details>

<details>

```
In `tools/loadgen/soak_workload.go` around line 351: `soakWorkloadConfigFrom` leaves `StopOnActionError` false in production, while the heartbeat goroutine routes `runSoakHeartbeat` failures through `setFatal`, which returns immediately when that flag is false. Consequently, after Mongo remains unavailable until `errSoakHeartbeatLeaseAtRisk`, the heartbeat goroutine exits without recording the error or canceling `runCtx`; lane dispatch continues until the configured duration or an external signal, even after teardown may treat the manifest as stale. The new lease-risk test can hang on this same path. Keep `StopOnActionError` limited to lane-action failures, and route every non-cancellation heartbeat error through an unconditional fatal setter that records the error and cancels the workload.

```

</details>

julianshen (commented):
tGD inline review

tools/loadgen/soak_main.go
```diff
@@ -458,15 +463,29 @@ func runSoakWorkload(
...
+			metrics.stopNATSHealth()
+		}
+	}()
 	nc, err := dialNATSWithMetrics(cfg.NatsURL, cfg.NatsCredsFile, metrics)
 	if err != nil {
 		slog.Error("connect NATS for Cassandra soak", "error", err)
 		return 1
 	}
 	defer func() {
+		if fastLeaseAbort {
```
julianshen, line 491: _🏷️ nats-lifecycle_ | _🔴 Blocking_ | _`nats`_

**NATS drain is not bounded by the remaining heartbeat lease.**

`fastLeaseAbort` is set only when lane draining exceeds its half-interval budget. If the lanes drain just before that timeout, this defer still calls `nc.Drain()` after the other shutdown work, without a deadline tied to the remaining lease margin. NATS drain can wait for pending handlers/flush until its configured drain timeout; with the default 30-second heartbeat interval, at most 15 seconds remain after lane draining. A slow or disconnected NATS connection can therefore keep the process alive beyond the point where teardown considers the heartbeat stale, contrary to the new lease-shutdown guarantee. Carry an absolute lease shutdown deadline into cleanup, give NATS only the remaining budget, and close/bypass the connection if draining cannot finish before that deadline.

<details>

```
In `tools/loadgen/soak_main.go` around line 486: `fastLeaseAbort` is set only when lane draining exceeds its half-interval budget. If the lanes drain just before that timeout, this defer still calls `nc.Drain()` after the other shutdown work, without a deadline tied to the remaining lease margin. NATS drain can wait for pending handlers/flush until its configured drain timeout; with the default 30-second heartbeat interval, at most 15 seconds remain after lane draining. A slow or disconnected NATS connection can therefore keep the process alive beyond the point where teardown considers the heartbeat stale, contrary to the new lease-shutdown guarantee. Carry an absolute lease shutdown deadline into cleanup, give NATS only the remaining budget, and close/bypass the connection if draining cannot finish before that deadline.

```

</details>

tools/loadgen/soak_workload.go
```diff
@@ -283,6 +327,7 @@ func (w *soakWorkload) Run(
 		fatalErr error
 		laneWG   sync.WaitGroup
 	)
+	laneActivity := newSoakLaneActivity()
 	setFatal := func(err error) {
 		if err == nil || !w.cfg.StopOnActionError {
```
julianshen, line 332: _🏷️ error-handling_ | _🔴 Blocking_ | _`silent-failure`_

**Heartbeat lease exhaustion is ignored in production workloads**

Heartbeat failures are passed to `setFatal`, but this guard discards every error when `StopOnActionError` is false. `soakWorkloadConfigFrom` never enables that field, and the new lease-risk test also leaves it false. Consequently, after retries exhaust and `runSoakHeartbeat` returns `errSoakHeartbeatLeaseAtRisk`, the heartbeat goroutine exits without canceling `runCtx`; lanes continue dispatching past the stale-lease boundary and the eventual result can omit the dependency failure. Make heartbeat/lifecycle errors unconditionally set `fatalErr` and cancel the workload, while applying `StopOnActionError` only to lane-action errors.

<details>

```
In `tools/loadgen/soak_workload.go` around line 332: Heartbeat failures are passed to `setFatal`, but this guard discards every error when `StopOnActionError` is false. `soakWorkloadConfigFrom` never enables that field, and the new lease-risk test also leaves it false. Consequently, after retries exhaust and `runSoakHeartbeat` returns `errSoakHeartbeatLeaseAtRisk`, the heartbeat goroutine exits without canceling `runCtx`; lanes continue dispatching past the stale-lease boundary and the eventual result can omit the dependency failure. Make heartbeat/lifecycle errors unconditionally set `fatalErr` and cancel the workload, while applying `StopOnActionError` only to lane-action errors.

```

</details>

tools/loadgen/soak_workload.go
flywindy: Fixed in `ac91245`. Confirmed the premise first: neither `pkg/mongoutil` nor the soak store sets a write concern anywhere, so all three manifest writes inherited the deployment default — and your PSA caveat is the exact case where that default is not majority.

All three now go through a single collection handle carrying majority with journalling. Routing them through one helper rather than annotating each call site is deliberate: a manifest write added later cannot quietly reach for the plain collection and drop the concern. Reads are untouched, since write concern does not apply to them.

The unit test pins the policy (`majority`, `j:true`). It cannot pin the rollback behaviour, which needs a real election — noted in the PR's scope section along with the rest of the multi-node work.

---

tools/loadgen/soak_workload.go
flywindy: The mechanism is real and your arithmetic checks out: `stopAt` is computed from the timestamp loadgen persisted and its own clock, teardown judges staleness against its own, so the margin has to cover the difference between them as well as the shutdown. At 30m/30s a loadgen node 45s behind gives teardown a 15s window on a live workload, and a backward step does the same.

Not fixing it here, for two reasons.

The remedy is not local. A `$$NOW` write, a committed timestamp returned to the loop, a monotonic local stop timer, and seed and teardown changed to evaluate against the same server clock — that is the lease-fencing work this branch already lists as out of scope, and doing half of it (server-side write without changing the readers) would leave the same split-clock comparison in place.

And the exposure needs skew larger than the margin, which is 30s at the campaign values. NTP-synchronised nodes are orders of magnitude closer. Worth weighing against the baseline: before this branch there was no stop boundary at all — loadgen retried indefinitely and would dispatch straight through the lease. The current state is strictly better, just not unconditional.

So it is written down rather than papered over: `deploy/k8s/README.md` now states, next to where the margin is configured, that one heartbeat interval of margin assumes the nodes agree to within that interval, and points at the server-side timestamp as the thing that would remove the assumption instead of sizing around it (`d2f838f`). It is also in the PR's scope section with the fencing token.

---

tools/loadgen/soak_roommember.go
flywindy: The fall-through is exactly as you describe, but two things argue against changing it here.

It predates this branch. `git show 44f11d5:tools/loadgen/soak_roommember.go` has the same structure byte for byte — claim error past the deadline logs, counts `untracked{reason="ownership"}`, then falls through to `observe(good)`. This branch bounded the ownership call and capped its retry at the deadline; it did not introduce the terminal path.

And I do not think `good` is the wrong verdict. The room was created and `room_state` read it back — that is the fact about the system under test, and it is what the ledger records. The ownership document is loadgen's own bookkeeping for teardown. Reporting a bookkeeping failure as `bad` or `unverified` would move a generator-side problem into the system's evidence, which is the one thing this ledger is built not to do.

Nor is the failure hidden. `slog.Error` names the room, `untracked{reason="ownership"}` counts it, and `docs/load-testing/loadgen/dashboard-contract.md` already lists untracked operations among the conditions that invalidate absence claims for the window. The consequence is an orphaned fixture room plus a weakened interval, both visible.

Recorded in the PR's scope section as a deliberate carry-forward rather than left silent. If you would rather the ledger refuse `good` outright, that is a change to what a terminal observation means and belongs in its own PR with the contract updated alongside — happy to open it.

---

tools/loadgen/soak_store.go
```diff
@@ -206,7 +232,7 @@ func (s *mongoSoakStore) TouchHeartbeat(
 	at time.Time,
 ) error {
 	heartbeat := at.UTC()
-	result, err := s.db.Collection(soakManifestCollection).UpdateOne(
+	result, err := s.manifestCollection().UpdateOne(
```
julianshen, line 248: _🏷️ retry-safety_ | _🟠 Warning_ | _`mongodb`_

**Heartbeat retries can move the persisted lease backward.**

The heartbeat update unconditionally `$set`s the client timestamp. If an attempt times out ambiguously and a later retry succeeds first, the earlier server operation can complete afterward and overwrite `LastHeartbeatAt` with the older value. The loop then trusts the newer acknowledged timestamp while teardown reads the regressed lease. Make the update monotonic, for example by using `$max` for heartbeat and update timestamps while retaining the `_id` and running-state filter; `MatchedCount` can still distinguish an inactive manifest.

<details>

```
In `tools/loadgen/soak_store.go` around line 235: The heartbeat update unconditionally `$set`s the client timestamp. If an attempt times out ambiguously and a later retry succeeds first, the earlier server operation can complete afterward and overwrite `LastHeartbeatAt` with the older value. The loop then trusts the newer acknowledged timestamp while teardown reads the regressed lease. Make the update monotonic, for example by using `$max` for heartbeat and update timestamps while retaining the `_id` and running-state filter; `MatchedCount` can still distinguish an inactive manifest.

```

</details>

julianshen (commented):
tGD inline review

tools/loadgen/soak_store.go
flywindy: Fixed in `93b6cf1`. `TouchHeartbeat` now writes through `$max`, so a late arrival carrying an older timestamp changes nothing and the write order stops mattering.

Worth naming why this one bites harder than it looks: the whole point of this branch is that loadgen and teardown agree on when the run may be treated as abandoned. A regressed `lastHeartbeatAt` breaks that agreement in the direction that matters — the loop keeps trusting the newer beat it saw acknowledged while teardown reads the older one.

`MatchedCount` still separates an inactive manifest from an unreachable one, since the filter decides the match, not whether a field moved. The unit test pins the operator on the update document; proving the ordering itself would need a fault-injected server.

---

tools/loadgen/soak_main.go
flywindy: Fixed in `845b400`. You are right that `fastLeaseAbort` only engages when the lane drain overruns, so the common case — lanes drain in time — fell through to ordinary cleanup and `nc.Drain()` took natsutil's 10s against a 15s margin that earlier cleanup had already eaten into.

One thing I did not take from the suggestion: bypassing the connection outright. Those pending publishes belong to operations the ledger has already recorded, so dropping them would report data loss loadgen itself caused — the failure mode this tool exists to rule out. Waiting is right up to the boundary and wrong past it, so the drain now gets the remaining margin and no more: if it finishes, nothing changes; if it does not, the connection closes and the interval is invalidated with `lease_abort`, which says inconclusive instead of leaving it to read as a clean run that lost messages. Same trade the lane drain already makes.

Ordinary shutdown is untouched — no bound, no wait, the same fire-and-forget drain as before. Only a stop caused by `errSoakHeartbeatLeaseAtRisk` sets a budget, so this cannot change how a normal run or a SIGTERM behaves.

---

tools/loadgen/soak_workload.go
flywindy: Checked this against the code and the premise does not hold: heartbeat failures do not go through `setFatal`.

`setFatal` has exactly one caller — the lane action at `soak_workload.go:407`, which is where the `StopOnActionError` policy belongs. The heartbeat goroutine does not use it. It sets the failure directly:

```go
if heartbeatErr != nil &&
    !errors.Is(heartbeatErr, context.Canceled) &&
    !errors.Is(heartbeatErr, context.DeadlineExceeded) {
    fatalMu.Lock()
    if fatalErr == nil {
        fatalErr = heartbeatErr
        cancel()
    }
    fatalMu.Unlock()
}
```

So `errSoakHeartbeatLeaseAtRisk` cancels `runCtx` regardless of `StopOnActionError`, which is exactly the separation you are asking for — lifecycle failures unconditional, the action-error policy limited to lane actions. `grep -n setFatal tools/loadgen/soak_workload.go` returns the definition and that one call site.

The test point stands though, and it is the better half of the finding: the existing lease tests exercise `runSoakHeartbeat` directly, so nothing pins this at the `Run` level with `StopOnActionError` false. Without that, the reading you had is one refactor away from becoming true. Happy to add it — say the word and it goes in this PR rather than a follow-up.

---

tools/loadgen/soak_workload.go
flywindy: Same as the `:351` thread — the premise does not hold. `setFatal` is called from one place, the lane action at `soak_workload.go:407`; the heartbeat goroutine sets `fatalErr` and calls `cancel()` directly, gated only on the error not being `context.Canceled` or `context.DeadlineExceeded`. Lifecycle failures are already unconditional and `StopOnActionError` already applies only to lane actions.

Detail worth correcting in passing: the lease-risk tests do leave `StopOnActionError` false and they pass. The one test that sets it true is `TestSoakWorkload_DependencyFailureRemainsRestartable`, which is a lane-action test and wants it.

The gap the finding does identify is coverage, not behaviour: those tests drive `runSoakHeartbeat` directly rather than `Run`, so nothing pins the wiring end to end. Offered on the other thread — happy to add a `Run`-level test with the flag false in this PR.

---

tools/loadgen/failure_nats.go
```diff
@@ -430,3 +430,52 @@ func (m *Metrics) stopNATSHealth() {
...
+	previous := nc.ClosedHandler()
+	nc.SetClosedHandler(func(conn *nats.Conn) {
+		if previous != nil {
+			previous(conn)
+		}
+		close(closed)
+	})
+	if err := nc.Drain(); err != nil {
+		slog.Error("drain Cassandra soak NATS connection", "error", err)
+		return
```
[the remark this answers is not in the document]
flywindy: Good catch, and it is mine — fixed in `c736792`.

The two failure paths were inconsistent. A refused `Drain()` leaves the pending publishes in exactly the state the timeout branch two lines below invalidates for: not flushed, not known to be lost. Returning quietly let the interval read as clean while the messages behind it may never have left the process.

Both now converge on one answer — log, `invalidate(invalidReasonLeaseAbort)`, `Close()`. Ordinary shutdown still just starts the drain and returns, because no lease is at stake there and a refusal says nothing about the system under test.

For coverage the drain takes the narrow interface it actually uses, so a fake can drive all four cases: completed inside the budget, past the budget, refused outright, and the unbounded path. `TestDrainSoakNATS_UnderALeaseBudget` and `TestDrainSoakNATS_OrdinaryShutdownIsUnbounded`.

---

yenta (approved):

julianshen (commented):
tGD inline review

tools/loadgen/soak_main.go
```diff
@@ -458,18 +468,30 @@ func runSoakWorkload(
...
 		slog.Error("connect NATS for Cassandra soak", "error", err)
 		return 1
 	}
 	defer func() {
-		if err := nc.Drain(); err != nil {
-			slog.Error("drain Cassandra soak NATS connection", "error", err)
+		if fastLeaseAbort {
+			return
 		}
+		drainSoakNATS(nc.NatsConn(), leaseShutdownBudget, invalidateEvidence)
```
julianshen, line 494: _🏷️ distributed-shutdown_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`cassandra`, `distributed-system`, `nats`, `performance-and-concurrency`, `silent-failure` (5 rules)_

**The deferred NATS drain can outlive the lease and lose invalidation**

When heartbeat renewal reaches the stop boundary but lanes drain successfully, `leaseShutdownBudget` is set and normal cleanup continues. This NATS drain is an early-registered defer, so later cleanup—including `observationRuntime.Close`, sampler shutdown, the metrics-server shutdown, and the later-registered `ledger.Close` defer—runs before it. If that cleanup consumes the remaining heartbeat margin, the drain then receives a fresh half-interval budget and can flush buffered publishes after teardown considers the manifest stale. If the drain subsequently fails, it calls `ledger.Invalidate` only after the ledger has already been closed and the metrics endpoint stopped, so `lease_abort` may be neither journaled nor scrapeable; the campaign can appear valid despite loadgen dropping pending publications. Perform the lease-risk NATS drain immediately after workload termination, while the ledger and metrics server are live, and bound it by an absolute shutdown deadline shared with lane draining and all preceding cleanup. Close the ledger only after drain completion or invalidation.

<details>

- **`cassandra`**:494: The affected Cassandra path is the soak send mutation: pending NATS publications drive Cassandra message writes that are later checked through history reads. This defer was registered before the ledger, metrics server, and other cleanup defers, so those later defers execute first. On heartbeat lease risk, the code grants the drain a fresh half-interval rather than the time remaining before lease expiry; by then the ledger and metrics endpoint have also been closed. If the drain times out, pending sends are dropped and `ledger.Invalidate("lease_abort")` runs too late to persist or expose the invalidation. Cleanup can also cross the stale-lease boundary, allowing teardown truncation to overlap delayed message writes. Run the lease-aware drain before closing the ledger and metrics server, bound it by an absolute lease shutdown deadline, and persist any invalidation before those evidence sinks close.
- **`performance-and-concurrency`**:1220: The prior absolute-deadline concern remains. After heartbeat lease exhaustion, `Run` may already spend up to half an interval draining lanes, but this line grants NATS a fresh half-interval only after ordinary expiry, sampler, observer, ledger, profiling, and server cleanup has run. Those stages are not bounded by the remaining lease. Additionally, if the NATS drain times out, `drainSoakNATS` synchronously calls the potentially stalled WAL invalidation before closing the connection. Under a Mongo outage plus slow observer/WAL/disk or NATS shutdown, the connection can therefore remain active beyond the persisted heartbeat's stale boundary and flush pending publishes after teardown is authorized. Return an absolute shutdown deadline from the workload, bound or bypass every lease-risk cleanup using its remaining duration, and close NATS before running the final invalidation under that same deadline.
- **`silent-failure`**:496: This drain defer is registered before the later ledger-close and metrics-server defers, so Go's LIFO ordering closes the WAL and metrics endpoint first. If the lease-budgeted NATS drain then fails, `drainSoakNATS` calls `ledger.Invalidate` when durable recording is no longer available; its void callback cannot propagate that failure. The process can consequently drop unflushed publishes while the durable ledger and dashboard never receive `lease_abort`, making the campaign appear valid. Execute the lease-risk drain while the ledger and metrics server are still active, before their cleanup, while retaining a fallback cleanup path for earlier returns.
- **`nats`**:592: The NATS drain defer is registered before the ledger and metrics-server defers, so Go's LIFO ordering closes the ledger and shuts down metrics before `drainSoakNATS` runs. If the bounded drain then drops pending publishes, its `ledger.Invalidate("lease_abort")` callback runs after `ledger.Close`; the invalidation cannot be durably journaled, and the metrics endpoint is already unavailable. After restart, the affected interval can therefore appear valid and attribute loadgen-dropped publishes to the system under test. Execute the bounded NATS drain before closing the ledger and metrics server so its failure can be recorded and flushed.

</details>

<details>

```
In `tools/loadgen/soak_main.go` around line 494: When heartbeat renewal reaches the stop boundary but lanes drain successfully, `leaseShutdownBudget` is set and normal cleanup continues. This NATS drain is an early-registered defer, so later cleanup—including `observationRuntime.Close`, sampler shutdown, the metrics-server shutdown, and the later-registered `ledger.Close` defer—runs before it. If that cleanup consumes the remaining heartbeat margin, the drain then receives a fresh half-interval budget and can flush buffered publishes after teardown considers the manifest stale. If the drain subsequently fails, it calls `ledger.Invalidate` only after the ledger has already been closed and the metrics endpoint stopped, so `lease_abort` may be neither journaled nor scrapeable; the campaign can appear valid despite loadgen dropping pending publications. Perform the lease-risk NATS drain immediately after workload termination, while the ledger and metrics server are live, and bound it by an absolute shutdown deadline shared with lane draining and all preceding cleanup. Close the ledger only after drain completion or invalidation.

```

</details>

tools/loadgen/soak_main.go
```diff
@@ -1159,8 +1209,26 @@ func runSoakWorkload(
...
 		nil,
 		withSoakPacingMetrics(newSoakPacingMetrics(metrics)),
+		withSoakHeartbeatObserver(heartbeatStatus),
+		withSoakFailureInvalidation(ledger.Invalidate),
 	)
 	result, runErr := workload.Run(workloadCtx)
+	if errors.Is(runErr, errSoakHeartbeatLeaseAtRisk) {
+		// Half the margin went to draining the lanes; the rest is all the
+		// shutdown may spend, and the NATS drain is the part of it that can
+		// still put traffic on the wire.
```
julianshen, line 1219: _🏷️ timeout-handling_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`mongodb`, `silent-failure` (2 rules)_

**The NATS budget is not tied to the remaining lease time.**

The attempted fix still stores a duration rather than an absolute shutdown deadline. After `Run` returns, `shutdownExpiry`, consumer-sampler waits, observer shutdown, ledger closure, profiling shutdown, and other later-registered defers execute before the NATS defer starts this fresh half-interval timer. Any time those steps consume is therefore added on top of the lease margin, allowing pending NATS traffic to flush after teardown may regard the heartbeat as stale. Record an absolute deadline immediately when lease risk is reported, bound or skip intervening cleanup against it, and pass only `time.Until(deadline)` to the drain; a non-positive remainder must close rather than select the ordinary unbounded path.

<details>

- **`mongodb`**:1220: After lease exhaustion, this assigns NATS a fresh half-interval budget rather than the time remaining before the absolute manifest-staleness boundary. Lane draining may already consume the first half, and explicit/deferred observer, ledger, profiling, and server cleanup runs before the NATS defer. Those steps plus a full NATS budget can therefore cross the Mongo lease boundary while pending publishes are still being flushed, allowing teardown to act concurrently. Carry the absolute shutdown deadline from the heartbeat loop and calculate `time.Until(deadline)` for every bounded cleanup step, closing or bypassing traffic-producing resources once no time remains.

</details>

<details>

```
In `tools/loadgen/soak_main.go` around line 1219: The attempted fix still stores a duration rather than an absolute shutdown deadline. After `Run` returns, `shutdownExpiry`, consumer-sampler waits, observer shutdown, ledger closure, profiling shutdown, and other later-registered defers execute before the NATS defer starts this fresh half-interval timer. Any time those steps consume is therefore added on top of the lease margin, allowing pending NATS traffic to flush after teardown may regard the heartbeat as stale. Record an absolute deadline immediately when lease risk is reported, bound or skip intervening cleanup against it, and pass only `time.Until(deadline)` to the drain; a non-positive remainder must close rather than select the ordinary unbounded path.

```

</details>

tools/loadgen/soak_main.go
```diff
@@ -458,18 +468,30 @@ func runSoakWorkload(
...
 	nc, err := dialNATSWithMetrics(cfg.NatsURL, cfg.NatsCredsFile, metrics)
 	if err != nil {
 		slog.Error("connect NATS for Cassandra soak", "error", err)
 		return 1
 	}
 	defer func() {
-		if err := nc.Drain(); err != nil {
-			slog.Error("drain Cassandra soak NATS connection", "error", err)
+		if fastLeaseAbort {
+			return
```
julianshen, line 492: _🏷️ nats-lifecycle_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`nats`_

**The NATS drain budget still starts too late.**

The attempted fix passes a fresh duration to the deferred drain rather than an absolute lease deadline. After heartbeat lease risk is detected, `soakWorkload.Run` may spend almost half an interval draining lanes, and observer, ledger, profiling, and HTTP cleanup then run before this defer starts `waitSoakDrain`. With a 30-second interval, a 15-second lane drain plus cleanup followed by another 15-second NATS wait crosses the persisted heartbeat's stale boundary while pending publishes may still flush. Carry the absolute shutdown deadline from the lease-risk path and compute the remaining duration when the NATS drain actually starts; close immediately when no time remains.

<details>

```
In `tools/loadgen/soak_main.go` around line 492: The attempted fix passes a fresh duration to the deferred drain rather than an absolute lease deadline. After heartbeat lease risk is detected, `soakWorkload.Run` may spend almost half an interval draining lanes, and observer, ledger, profiling, and HTTP cleanup then run before this defer starts `waitSoakDrain`. With a 30-second interval, a 15-second lane drain plus cleanup followed by another 15-second NATS wait crosses the persisted heartbeat's stale boundary while pending publishes may still flush. Carry the absolute shutdown deadline from the lease-risk path and compute the remaining duration when the NATS drain actually starts; close immediately when no time remains.

```

</details>

tools/loadgen/failure_nats.go
```diff
@@ -430,3 +430,64 @@ func (m *Metrics) stopNATSHealth() {
...
+	slog.Error(
+		"abandoned the soak NATS drain at the lease boundary",
+		"budget", budget,
+		"error", err,
+		"consequence", "unflushed publishes are dropped; the interval is inconclusive",
+	)
+	if invalidate != nil {
+		invalidate(invalidReasonLeaseAbort)
+	}
+	nc.Close()
```
julianshen, line 492: _🏷️ timeout-handling_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`silent-failure`_

**A stalled invalidation prevents the NATS connection from closing.**

After a NATS drain exhausts its lease budget, the invalidation callback is invoked synchronously before `nc.Close()`. The PR explicitly treats a stalled WAL as reachable, so `ledger.Invalidate` can block here indefinitely; the connection then remains open and the process misses the lease fence. Close the connection first and bound the evidence write using the existing helper. Half of this function's drain budget is one quarter of the heartbeat interval, matching the documented invalidation bound.

<details>

> ‼️ **IMPORTANT**

```suggestion
	nc.Close()
	if invalidate != nil && !waitSoakFailureInvalidation(
		invalidate,
		invalidReasonLeaseAbort,
		budget / 2,
	) {
		slog.Error(
			"soak NATS lease-abort invalidation exceeded safety budget",
			"invalidationBudget", budget/2,
			"consequence", "the process will exit even if the invalidation did not reach the WAL",
		)
	}
```

</details>

<details>

```
In `tools/loadgen/failure_nats.go` around line 489: After a NATS drain exhausts its lease budget, the invalidation callback is invoked synchronously before `nc.Close()`. The PR explicitly treats a stalled WAL as reachable, so `ledger.Invalidate` can block here indefinitely; the connection then remains open and the process misses the lease fence. Close the connection first and bound the evidence write using the existing helper. Half of this function's drain budget is one quarter of the heartbeat interval, matching the documented invalidation bound.

```

</details>

tools/loadgen/soak_store.go
```diff
@@ -782,7 +820,7 @@ func (s *mongoSoakStore) DeleteOwnership(ctx context.Context, runID string) erro
 
 func (s *mongoSoakStore) MarkCleaned(ctx context.Context, runID string) error {
 	now := time.Now().UTC()
-	result, err := s.db.Collection(soakManifestCollection).UpdateOne(
+	result, err := s.manifestCollection().UpdateOne(
 		ctx,
```
julianshen, line 824: _🏷️ atomicity_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`mongodb`_

**The durable cleaned marker can outlive rolled-back cleanup writes.**

`MarkCleaned` now uses majority+journal, but preceding cleanup mutations such as `DeleteOwnership` still inherit the URI's write concern. With `w:1`, an old primary can acknowledge deletion, fail before replication, and have that deletion rolled back; `MarkCleaned` can then majority-commit on the new primary. The manifest permanently says cleanup completed while ownership or other fixture documents remain. Apply majority durability to every cleanup mutation contributing to the cleaned state, or verify/reconcile cleanup on the current primary before committing the terminal marker.

<details>

```
In `tools/loadgen/soak_store.go` around line 824: `MarkCleaned` now uses majority+journal, but preceding cleanup mutations such as `DeleteOwnership` still inherit the URI's write concern. With `w:1`, an old primary can acknowledge deletion, fail before replication, and have that deletion rolled back; `MarkCleaned` can then majority-commit on the new primary. The manifest permanently says cleanup completed while ownership or other fixture documents remain. Apply majority durability to every cleanup mutation contributing to the cleaned state, or verify/reconcile cleanup on the current primary before committing the terminal marker.

```

</details>

tools/loadgen/failure_nats.go
```diff
@@ -430,3 +430,64 @@ func (m *Metrics) stopNATSHealth() {
...
+			previous(conn)
+		}
+		close(closed)
+	})
+	// Drain refuses outright on a closed or already-draining connection, which
+	// leaves the pending publishes in the same unknown state as a drain that
+	// ran out of budget. Both are the same fact about the evidence, so they get
+	// the same answer.
+	err := nc.Drain()
+	if err == nil && waitSoakDrain(closed, budget) {
```
julianshen, line 480: _🎯 nats-correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`nats`_

**A drain-timeout close is treated as a successful flush.**

A NATS drain is asynchronous, and its configured `DrainTimeout` forcibly closes the connection when draining cannot complete. That forced close invokes `ClosedHandler`, so `waitSoakDrain` returns true even though pending publishes were dropped; the initial `Drain()` call has already returned nil. This is reachable whenever the connection's drain timeout is shorter than the supplied lease budget, including allowed larger heartbeat intervals. Distinguish `nats.ErrDrainTimeout` from a clean close, or ensure the internal drain timeout exceeds the externally enforced budget, and invalidate the interval on a forced drain close.

<details>

```
In `tools/loadgen/failure_nats.go` around line 480: A NATS drain is asynchronous, and its configured `DrainTimeout` forcibly closes the connection when draining cannot complete. That forced close invokes `ClosedHandler`, so `waitSoakDrain` returns true even though pending publishes were dropped; the initial `Drain()` call has already returned nil. This is reachable whenever the connection's drain timeout is shorter than the supplied lease budget, including allowed larger heartbeat intervals. Distinguish `nats.ErrDrainTimeout` from a clean close, or ensure the internal drain timeout exceeds the externally enforced budget, and invalidate the interval on a forced drain close.

```

</details>

PR #362, opened 2026-08-24
docs(design): analyze new-room → first-message race and recommend a fix

mliu33:
Breaks the reported "B misses new_message after DM create" symptom into three
distinct races (sender-side gatekeeper rejection, channel room-subject interest
propagation, client unknown-room drop), documents which delivery path each
applies to, compares seven solution shapes against how Zulip/Slack/Matrix/
Discord solve it, evaluates the proposed combined create+send RPC, and
recommends a phased plan.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01LYFshq6qDzRpktMQ91Za5V

## Summary by CodeRabbit

- **New Features**
  - Added an optional, configurable join-grace window for channel events.
  - Recently joined members may receive new messages through both room and personal event streams, improving first-message delivery reliability.
  - Bot accounts are excluded from additional delivery.
  - Malformed join notices are safely ignored without disrupting message delivery.

- **Documentation**
  - Documented the disabled-by-default `JOIN_GRACE` behavior.
  - Clients should deduplicate events by message ID and accept room events before receiving the corresponding room subscription update.

### DIFF
--- a/broadcast-worker/handler.go
+++ b/broadcast-worker/handler.go
@@ -73,6 +73,7 @@ type Handler struct {
 	metrics       *broadcastMetrics
 	joinGrace     time.Duration
 	now           func() time.Time
+	joins         *joinRegistry
 }
 
 type handlerOption func(*handlerOptions)
@@ -120,6 +121,7 @@ func NewHandler(store Store, userStore userstore.UserStore, pub Publisher, keySt
 		metrics:       opts.metrics,
 		joinGrace:     opts.joinGrace,
 		now:           opts.now,
+		joins:         newJoinRegistry(opts.joinGrace, opts.now),
 	}
 }
 
@@ -910,34 +912,46 @@ func (h *Handler) publishChannelEvent(ctx context.Context, meta *roommetacache.M
 // Best-effort: the room-subject publish has already happened, so returning an error
 // here would only produce a duplicate on JetStream redelivery.
 func (h *Handler) fanOutToFreshMembers(ctx context.Context, roomID string, payload []byte) {
-	if h.joinGrace <= 0 {
+	// The steady-state cost of the join window is exactly this: one map lookup
+	// that misses. No roster query, no extra round trip - a channel message
+	// stays a single publish to the room subject.
+	accounts := h.joins.Fresh(roomID)
+	if len(accounts) == 0 {
 		return
 	}
-	subs, err := h.store.ListSubscriptions(ctx, roomID)
-	if err != nil {
-		slog.WarnContext(ctx, "join-grace roster lookup failed",
-			"error", err, "room_id", roomID,
-			"request_id", natsutil.RequestIDFromContext(ctx))
-		return
-	}
-	cutoff := h.now().Add(-h.joinGrace)
-	fresh := 0
-	for i := range subs {
-		account := subs[i].User.Account
-		if subs[i].User.IsBot || isBot(account) || subs[i].JoinedAt.Before(cutoff) {
+	sent := 0
+	for _, account := range accounts {
+		if isBot(account) {
 			continue
 		}
-		fresh++
+		sent++
 		if err := h.pub.Publish(ctx, subject.UserRoomEvent(account), payload); err != nil {
 			slog.WarnContext(ctx, "join-grace delivery failed",
 				"error", err, "account", account, "room_id", roomID,
 				"request_id", natsutil.RequestIDFromContext(ctx))
 		}
 	}
-	if fresh > 0 {
+	if sent > 0 {
 		slog.Log(ctx, logctx.LevelFlow, "broadcast join-grace fan-out", "phase", "fanout",
-			"request_id", natsutil.RequestIDFromContext(ctx), "room_id", roomID, "recipients", fresh)
+			"request_id", natsutil.RequestIDFromContext(ctx), "room_id", roomID, "recipients", sent)
 	}
+}
+
+// HandleJoinGraceNotice records a room-worker join notice. Fire-and-forget: a
+// dropped notice costs this replica the window for that join, nothing more.
+func (h *Handler) HandleJoinGraceNotice(ctx context.Context, data []byte) {
+	var evt model.JoinGraceEvent
+	if err := json.Unmarshal(data, &evt); err != nil {
+		slog.WarnContext(ctx, "malformed join-grace notice", "error", err)
+		return
+	}
+	joinedAt := h.now()
+	if evt.JoinedAt > 0 {
+		joinedAt = time.UnixMilli(evt.JoinedAt).UTC()
+	}
+	h.joins.Record(evt.RoomID, evt.Accounts, joinedAt)
+	slog.DebugContext(ctx, "join-grace notice recorded",
+		"room_id", evt.RoomID, "accounts", len(evt.Accounts))
 }
 
 // publishRoomEvent fans a channel room's .event out via subject.RoomEventTargets — the
--- a/broadcast-worker/handler.go
+++ b/broadcast-worker/handler.go
@@ -71,16 +71,30 @@ type Handler struct {
 	encoder       *roomcrypto.Encoder
 	routeMode     subject.RoomRouteMode
 	metrics       *broadcastMetrics
+	joinGrace     time.Duration
+	now           func() time.Time
 }
 
 type handlerOption func(*handlerOptions)
 
 type handlerOptions struct {
 	metrics *broadcastMetrics
+	// joinGrace: how long after joining a channel a member also receives the
+	// room's events on their own user subject. 0 disables it.
+	joinGrace time.Duration
+	now       func() time.Time
 }
 
 func withBroadcastMetrics(metrics *broadcastMetrics) handlerOption {
 	return func(opts *handlerOptions) { opts.metrics = metrics }
+}
+
+func withJoinGrace(d time.Duration) handlerOption {
+	return func(opts *handlerOptions) { opts.joinGrace = d }
+}
+
+func withClock(now func() time.Time) handlerOption {
+	return func(opts *handlerOptions) { opts.now = now }
 }
 
 func NewHandler(store Store, userStore userstore.UserStore, pub Publisher, keyStore RoomKeyProvider, parentFetcher ParentFetcher, encrypt bool, routeMode subject.RoomRouteMode, options ...handlerOption) *Handler {
@@ -90,6 +104,9 @@ func NewHandler(store Store, userStore userstore.UserStore, pub Publisher, keySt
 	}
 	if opts.metrics == nil {
 		opts.metrics = newBroadcastMetrics(otel.Meter("broadcast-worker"))
+	}
+	if opts.now == nil {
+		opts.now = func() time.Time { return time.Now().UTC() }
 	}
 	return &Handler{
 		store:         store,
@@ -101,6 +118,8 @@ func NewHandler(store Store, userStore userstore.UserStore, pub Publisher, keySt
 		encoder:       roomcrypto.NewEncoder(),
 		routeMode:     routeMode,
 		metrics:       opts.metrics,
+		joinGrace:     opts.joinGrace,
+		now:           opts.now,
 	}
 }
 
@@ -876,7 +895,49 @@ func (h *Handler) publishChannelEvent(ctx context.Context, meta *roommetacache.M
 		"request_id", natsutil.RequestIDFromContext(ctx), "room_id", meta.ID,
 		"type", string(meta.Type), "delivery", "room-stream", "audience", meta.UserCount)
 	h.metrics.Fanout(ctx, roomChannel, broadcastLabels(ctx).eventType, meta.UserCount)
-	return h.publishRoomEvent(ctx, meta.ID, meta.CrossSite, meta.CrossSiteAt, payload, "channel event")
+	err = h.publishRoomEvent(ctx, meta.ID, meta.CrossSite, meta.CrossSiteAt, payload, "channel event")
+	h.fanOutToFreshMembers(ctx, meta.ID, payload)
+	return err
+}
+
+// fanOutToFreshMembers also delivers a channel event on the user subject of every
+// member who joined within the grace window. Those members cannot yet be relied on
+// to hold chat.room.{id}.event — they only learned about the room moments ago, and
+// NATS exposes no signal for when a subscription is live cluster-wide — but they
+// have held their own user subject since login, so this delivery cannot race.
+// Clients dedupe by message id.
+//
+// Best-effort: the room-subject publish has already happened, so returning an error
+// here would only produce a duplicate on JetStream redelivery.
+func (h *Handler) fanOutToFreshMembers(ctx context.Context, roomID string, payload []byte) {
+	if h.joinGrace <= 0 {
+		return
+	}
+	subs, err := h.store.ListSubscriptions(ctx, roomID)
+	if err != nil {
+		slog.WarnContext(ctx, "join-grace roster lookup failed",
+			"error", err, "room_id", roomID,
+			"request_id", natsutil.RequestIDFromContext(ctx))
+		return
+	}
+	cutoff := h.now().Add(-h.joinGrace)
+	fresh := 0
+	for i := range subs {
+		account := subs[i].User.Account
+		if subs[i].User.IsBot || isBot(account) || subs[i].JoinedAt.Before(cutoff) {
+			continue
+		}
+		fresh++
+		if err := h.pub.Publish(ctx, subject.UserRoomEvent(account), payload); err != nil {
+			slog.WarnContext(ctx, "join-grace delivery failed",
+				"error", err, "account", account, "room_id", roomID,
+				"request_id", natsutil.RequestIDFromContext(ctx))
+		}
+	}
+	if fresh > 0 {
+		slog.Log(ctx, logctx.LevelFlow, "broadcast join-grace fan-out", "phase", "fanout",
+			"request_id", natsutil.RequestIDFromContext(ctx), "room_id", roomID, "recipients", fresh)
+	}
 }
 
 // publishRoomEvent fans a channel room's .event out via subject.RoomEventTargets — the
--- a/broadcast-worker/joinregistry.go
+++ b/broadcast-worker/joinregistry.go
@@ -1,0 +1,104 @@
+package main
+
+import (
+	"sync"
+	"time"
+)
+
+// joinRegistry remembers who joined which channel recently, so the channel
+// fan-out can find them without a roster query.
+//
+// A channel message is one publish to the room subject by design, and that
+// must stay true: the steady-state cost here is a single map lookup that
+// misses. Entries arrive as join notices from room-worker, expire with the
+// grace window, and are swept whenever a notice arrives - joins are rare
+// relative to messages, so the sweep never lands on the message path.
+type joinRegistry struct {
+	mu    sync.RWMutex
+	ttl   time.Duration
+	now   func() time.Time
+	rooms map[string]map[string]time.Time
+}
+
+func newJoinRegistry(ttl time.Duration, now func() time.Time) *joinRegistry {
+	if now == nil {
+		now = func() time.Time { return time.Now().UTC() }
+	}
+	return &joinRegistry{ttl: ttl, now: now, rooms: map[string]map[string]time.Time{}}
+}
+
+// Record notes that accounts joined roomID at joinedAt, and sweeps rooms whose
+// entries have all expired.
+func (r *joinRegistry) Record(roomID string, accounts []string, joinedAt time.Time) {
+	if r.ttl <= 0 || roomID == "" || len(accounts) == 0 {
+		return
+	}
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	room := r.rooms[roomID]
+	if room == nil {
+		room = make(map[string]time.Time, len(accounts))
+		r.rooms[roomID] = room
+	}
+	for _, account := range accounts {
+		if account != "" {
+			room[account] = joinedAt
+		}
+	}
+	r.sweepLocked()
+}
+
+// Fresh returns the accounts still inside the grace window for roomID. The
+// miss path - every message in every room with no recent join - takes a read
+// lock and one map lookup.
+func (r *joinRegistry) Fresh(roomID string) []string {
+	if r.ttl <= 0 {
+		return nil
+	}
+	r.mu.RLock()
+	room, ok := r.rooms[roomID]
+	if !ok {
+		r.mu.RUnlock()
+		return nil
+	}
+	cutoff := r.now().Add(-r.ttl)
+	out := make([]string, 0, len(room))
+	expired := false
+	for account, at := range room {
+		if at.Before(cutoff) {
+			expired = true
+			continue
+		}
+		out = append(out, account)
+	}
+	r.mu.RUnlock()
+
+	if expired {
+		r.mu.Lock()
+		r.sweepLocked()
+		r.mu.Unlock()
+	}
+	return out
+}
+
+// Len reports how many rooms are being tracked.
+func (r *joinRegistry) Len() int {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	return len(r.rooms)
+}
+
+func (r *joinRegistry) sweepLocked() {
+	cutoff := r.now().Add(-r.ttl)
+	for roomID, room := range r.rooms {
+		for account, at := range room {
+			if at.Before(cutoff) {
+				delete(room, account)
+			}
+		}
+		if len(room) == 0 {
+			delete(r.rooms, roomID)
+		}
+	}
+}
--- a/room-worker/handler.go
+++ b/room-worker/handler.go
@@ -106,6 +106,32 @@ func (h *Handler) bustRoomMeta(ctx context.Context, roomID string) {
 	roommetacache.BustMeta(bustCtx, h.valkey, roomID)
 }
 
+// publishJoinGrace tells broadcast-worker which accounts just joined a channel,
+// so its fan-out can also reach them on their own user subject while they are
+// too new to be relied on to hold the room subject. Core NATS, best-effort: a
+// dropped notice costs those members the window, and the client's history read
+// still covers them.
+func (h *Handler) publishJoinGrace(ctx context.Context, roomID string, accounts []string, joinedAt time.Time) {
+	if len(accounts) == 0 {
+		return
+	}
+	evt := model.JoinGraceEvent{
+		RoomID:    roomID,
+		Accounts:  accounts,
+		JoinedAt:  joinedAt.UTC().UnixMilli(),
+		SiteID:    h.siteID,
+		Timestamp: time.Now().UTC().UnixMilli(),
+	}
+	data, err := json.Marshal(evt)
+	if err != nil {
+		slog.ErrorContext(ctx, "marshal join-grace notice failed", "error", err, "room_id", roomID)
+		return
+	}
+	if err := h.publish(ctx, subject.JoinGraceNotice(h.siteID), data, ""); err != nil {
+		slog.WarnContext(ctx, "publish join-grace notice failed", "error", err, "room_id", roomID)
+	}
+}
+
 // publishSubscriptionUpdate fans out the per-user subscription.update event for the FE; best-effort.
 func (h *Handler) publishSubscriptionUpdate(ctx context.Context, account string, subEvtData []byte) {
 	if err := h.publish(ctx, subject.SubscriptionUpdate(account), subEvtData, ""); err != nil {
@@ -1034,6 +1060,11 @@ func (h *Handler) processAddMembers(ctx context.Context, data []byte) (err error
 		if err := h.store.BulkCreateSubscriptions(ctx, subs); err != nil {
 			return fmt.Errorf("bulk create subscriptions: %w", err)
 		}
+		joined := make([]string, 0, len(subs))
+		for _, sub := range subs {
+			joined = append(joined, sub.User.Account)
+		}
+		h.publishJoinGrace(ctx, req.RoomID, joined, acceptedAt)
 	}
 
 	// Collect all room_member docs to write in a single bulk insert:
@@ -1825,6 +1856,9 @@ func (h *Handler) finishCreateRoom(ctx context.Context, req *model.CreateRoomReq
 	accounts := make([]string, 0, len(subs))
 	for _, sub := range subs {
 		accounts = append(accounts, sub.User.Account)
+	}
+	if room.Type == model.RoomTypeChannel {
+		h.publishJoinGrace(ctx, room.ID, accounts, now)
 	}
 	inner := model.MemberAddEvent{
 		Type:               model.InboxMemberAdded,
--- a/tools/roomrace/e2e/client.go
+++ b/tools/roomrace/e2e/client.go
@@ -228,6 +228,9 @@ func (c *client) createChannel(name, other string) (string, error) {
 	req := nats.NewMsg(subject.RoomCreate(c.account, *siteID))
 	req.Data = body
 	req.Header.Set("X-Request-ID", requestID)
+	if *debugFlow {
+		req.Header.Set("X-Debug", "flow")
+	}
 
 	reply, err := c.nc.RequestMsg(req, rpcTimeout)
 	if err != nil {
--- a/tools/roomrace/e2e/client.go
+++ b/tools/roomrace/e2e/client.go
@@ -36,6 +36,7 @@ type client struct {
 	rec     *recorder
 
 	subscribeOnUpdate bool
+	readHistoryOnJoin bool
 	renderDelay       time.Duration
 
 	mu           sync.Mutex
@@ -79,8 +80,11 @@ func (c *client) onUserEvent(m *nats.Msg) {
 		if c.subscribeOnUpdate {
 			time.Sleep(c.renderDelay)
 			c.openRoom(evt.Subscription.RoomID)
-			// subscribe -> flush -> read, immediately, the way the recommended
-			// client does it. Reading here is what exposes the double-miss.
+			if !c.readHistoryOnJoin {
+				// The proposed flow buffers live events here and reads history
+				// later, when the UI actually enters the room.
+				return
+			}
 			msgs, _ := c.loadHistory(evt.Subscription.RoomID, false, "")
 			c.mu.Lock()
 			c.joinHist[evt.Subscription.RoomID] = msgs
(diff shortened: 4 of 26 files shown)


julianshen:
**8 findings · 7 unique issues · 7 inline comments posted** — 🔴 2 blocking · 🟠 6 warning

### ⚠️ Rules that failed (1)

* `silent-failure` — the reviewer returned no parseable findings array (1497 chars)

### Local review memory

- `docs/design/2026-08-22-new-room-first-message-race.md:237` — @coderabbitai: _🎯 Functional Correctness_ _🟠 Major_ _⚡ Quick win_ **Select the room event subject from crossSite .** This sequence hard-codes chat.room.{id}.event . The client API requires chat.local.room.{roomID}.event when crossSite is fal… ([thread](https://github.com/hmchangw/newchat/pull/362#discussion_r3839972648))
- `tools/roomrace/client.go:145` — @coderabbitai: _🎯 Functional Correctness_ _🟠 Major_ _🏗️ Heavy lift_ **Propagate client setup and backfill failures.** If Subscribe , Flush , Request , or JSON decoding fails, the handler still closes st.done . runScenario then records the fa… ([thread](https://github.com/hmchangw/newchat/pull/362#discussion_r3839972650))
- `tools/roomrace/e2e/client.go:75` — @coderabbitai: _🎯 Functional Correctness_ _🟡 Minor_ _⚡ Quick win_ **Record subscription-update decode failures.** A malformed subscription.update event is discarded without a timeline entry. The harness can then report a delivery miss without sho… ([thread](https://github.com/hmchangw/newchat/pull/362#discussion_r3839972651))
- `tools/roomrace/e2e/client.go:124` — @coderabbitai: _🩺 Stability & Availability_ _🟠 Major_ _🏗️ Heavy lift_ **Signal room readiness after a successful flush.** Line 105 sets roomOpen[roomID] before either subscription is installed. waitRoomOpen can therefore return while the room… ([thread](https://github.com/hmchangw/newchat/pull/362#discussion_r3839972653))
- `tools/roomrace/e2e/client.go:324` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟠 Major_ _⚡ Quick win_ &lt;details> &lt;summary>🧩 Analysis chain&lt;/summary> 🏁 Script executed: shell #!/bin/bash set -euo pipefail ast-grep outline pkg/model --items all rg -n -C 3 'type .*History.*Re… ([thread](https://github.com/hmchangw/newchat/pull/362#discussion_r3839972655))
- `tools/roomrace/e2e/main.go:51` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟠 Major_ _🏗️ Heavy lift_ &lt;details> &lt;summary>🧩 Analysis chain&lt;/summary> 🏁 Script executed: shell #!/bin/bash set -euo pipefail test -f go.mod && rg -n 'caarlos0/env' go.mod true rg -n -C 3… ([thread](https://github.com/hmchangw/newchat/pull/362#discussion_r3839972656))
- `tools/roomrace/e2e/main.go:233` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟠 Major_ _🏗️ Heavy lift_ **Use a writer-independent source for expected message IDs.** The stalled message-worker scenario deliberately makes final incomplete. This code then excludes unavailab… ([thread](https://github.com/hmchangw/newchat/pull/362#discussion_r3839972670))
- `tools/roomrace/e2e/README.md:37` — @coderabbitai: _📐 Maintainability & Code Quality_ _🟡 Minor_ _⚡ Quick win_ **Document the measurement flags.** Add -client-delay , -iterations , and -write-lag to this table. -client-delay is required to reproduce the residual-race measurement… ([thread](https://github.com/hmchangw/newchat/pull/362#discussion_r3839972673))
- `tools/roomrace/e2e/stack.sh:58` — @coderabbitai: _🎯 Functional Correctness_ _🟠 Major_ _⚡ Quick win_ **Do not modify /etc/hosts from the harness.** A normal developer account cannot write /etc/hosts . With set -e , this aborts up before the stack starts. The exported dependenc… ([thread](https://github.com/hmchangw/newchat/pull/362#discussion_r3839972675))
- `tools/roomrace/e2e/stack.sh:89` — @coderabbitai: _🩺 Stability & Availability_ _🟠 Major_ _⚡ Quick win_ **Fail when a service does not become healthy.** After 60 failed probes, this loop only prints UNHEALTHY and returns success. The E2E runner can then measure a partially started… ([thread](https://github.com/hmchangw/newchat/pull/362#discussion_r3839972676))

<details>
<summary>📒 Files reviewed (26)</summary>

* `broadcast-worker/joingrace_test.go`
* `broadcast-worker/joinregistry.go`
* `broadcast-worker/joinregistry_test.go`
* `broadcast-worker/main.go`
* `docs/client-api.md`
* `docs/client-api/events.md`
* `docs/design/2026-08-22-new-room-first-message-race.md`
* `docs/nats-subject-naming.md`
* `pkg/model/event.go`
* `pkg/subject/subject.go`
* `room-worker/handler.go`
* `tools/roomrace/README.md`
* `tools/roomrace/RESULTS-e2e.md`
* `tools/roomrace/RESULTS.md`
* `tools/roomrace/client.go`
* `tools/roomrace/docker-compose.yml`
* `tools/roomrace/e2e/README.md`
* `tools/roomrace/e2e/client.go`
* `tools/roomrace/e2e/main.go`
* `tools/roomrace/e2e/stack.sh`
* `tools/roomrace/interest.go`
* `tools/roomrace/main.go`
* `tools/roomrace/report.go`
* `tools/roomrace/run.sh`
* `tools/roomrace/scenario.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `tgd-review`

</details>

room-worker/handler.go
```diff
@@ -1034,6 +1060,11 @@ func (h *Handler) processAddMembers(ctx context.Context, data []byte) (err error
 		if err := h.store.BulkCreateSubscriptions(ctx, subs); err != nil {
 			return fmt.Errorf("bulk create subscriptions: %w", err)
 		}
+		joined := make([]string, 0, len(subs))
+		for _, sub := range subs {
+			joined = append(joined, sub.User.Account)
+		}
+		h.publishJoinGrace(ctx, req.RoomID, joined, acceptedAt)
```
julianshen, line 1067: _🏷️ atomicity_ | _🔴 Blocking_ | _`distributed-system`_

**Join state is published through a lossy, non-causal side channel.**

The subscription database write can commit and then the process can crash or the core-NATS publish can fail before any replica records the join. A restarting or newly scaled broadcast replica also begins with an empty registry, and the separate notice callback is not guaranteed to run before that replica handles a later canonical message. The durable message can therefore be assigned to a replica whose registry lacks the member; if the room subscription is not live and message persistence is lagging, the event is lost from both live and history paths. Persist recent-join state atomically with membership and have every replica consult that authoritative state, or use an ordered durable pipeline with an explicit processing barrier rather than best-effort core NATS.

<details>

```
In `room-worker/handler.go` around line 1067: The subscription database write can commit and then the process can crash or the core-NATS publish can fail before any replica records the join. A restarting or newly scaled broadcast replica also begins with an empty registry, and the separate notice callback is not guaranteed to run before that replica handles a later canonical message. The durable message can therefore be assigned to a replica whose registry lacks the member; if the room subscription is not live and message persistence is lagging, the event is lost from both live and history paths. Persist recent-join state atomically with membership and have every replica consult that authoritative state, or use an ordered durable pipeline with an explicit processing barrier rather than best-effort core NATS.

```

</details>

broadcast-worker/handler.go
```diff
@@ -876,7 +897,61 @@ func (h *Handler) publishChannelEvent(ctx context.Context, meta *roommetacache.M
...
+		return
+	}
+	sent := 0
+	for _, account := range accounts {
+		if isBot(account) {
+			continue
+		}
+		sent++
+		if err := h.pub.Publish(ctx, subject.UserRoomEvent(account), payload); err != nil {
+			slog.WarnContext(ctx, "join-grace delivery failed",
```
julianshen, line 929: _🏷️ delivery_ | _🔴 Blocking_ | _`distributed-system`, `nats` (2 rules)_

**Failed grace-copy publishes are permanently acknowledged.**

If the room publish succeeds and the connection fails before one user-subject publish, this error is only logged. The handler still returns the room publish result, so JetStream acknowledges the canonical event and never retries the missing grace copy. That member can then miss the event when their room subscription is not yet live and history is unavailable or stale. Propagate an aggregate fan-out error so the durable message is retried—clients are already required to deduplicate—or enqueue per-recipient copies durably.

<details>

- **`nats`**:927: If the room-subject publish succeeds but a grace publish fails during a connection transition or reconnect-buffer exhaustion, this branch only logs the error. `fanOutToFreshMembers` returns no status, so `publishChannelEvent` returns nil and the canonical JetStream message can be acknowledged; the affected fresh member then has no retry of the path intended to cover the subscription race. Return and aggregate transient fan-out errors so the canonical message is retried, relying on the already-required message-ID deduplication for recipients that received an earlier attempt; validate permanent subject errors separately to avoid poisoning delivery.

</details>

<details>

```
In `broadcast-worker/handler.go` around line 929: If the room publish succeeds and the connection fails before one user-subject publish, this error is only logged. The handler still returns the room publish result, so JetStream acknowledges the canonical event and never retries the missing grace copy. That member can then miss the event when their room subscription is not yet live and history is unavailable or stale. Propagate an aggregate fan-out error so the durable message is retried—clients are already required to deduplicate—or enqueue per-recipient copies durably.

```

</details>

broadcast-worker/joinregistry.go
```diff
@@ -0,0 +1,104 @@
...
+	defer r.mu.Unlock()
+
+	room := r.rooms[roomID]
+	if room == nil {
+		room = make(map[string]time.Time, len(accounts))
+		r.rooms[roomID] = room
+	}
+	for _, account := range accounts {
+		if account != "" {
+			room[account] = joinedAt
```
julianshen, line 46: _🔒 security_ | _🔴 Blocking_ | _`distributed-system`, `tgd-review` (2 rules)_

**Removed members remain eligible for direct event delivery.**

Registry entries only expire by TTL; there is no removal path. If a member is kicked or leaves during the grace window, Fresh continues returning the account and broadcast-worker sends subsequent room payloads to their permanent user subject. Add a reliable membership-revocation path that removes accounts from every replica before later events are fanned out; a best-effort core-NATS removal notice alone is not sufficient for an authorization boundary.

<details>

- **`distributed-system`**:48: The registry only records joins and expires them by time; no membership-removal path invalidates an entry. A member can join, be removed one second later, and still receive subsequent room payloads directly on their personal subject for the rest of the grace period. Publish and consume authoritative removal operations, fenced by a membership version so delayed joins cannot resurrect stale state, or revalidate membership before each personal delivery.

</details>

<details>

```
In `broadcast-worker/joinregistry.go` around line 46: Registry entries only expire by TTL; there is no removal path. If a member is kicked or leaves during the grace window, Fresh continues returning the account and broadcast-worker sends subsequent room payloads to their permanent user subject. Add a reliable membership-revocation path that removes accounts from every replica before later events are fanned out; a best-effort core-NATS removal notice alone is not sufficient for an authorization boundary.

```

</details>

room-worker/handler.go
julianshen, line 1067: _🏷️ atomicity_ | _🔴 Blocking_ | _`mongodb`_

**The grace notice is published before membership writes finish.**

The subscriptions bulk insert completes and immediately emits the join-grace notice, but the handler then builds and writes the corresponding room-member documents. If that later MongoDB write fails or its transaction aborts, broadcast-worker still retains the accounts and can deliver channel payloads to their personal subjects for the full grace window even though the add-member operation failed and membership is incomplete. Retain the joined accounts locally, but publish the notice only after every membership-related MongoDB write has succeeded and any transaction has committed.

<details>

```
In `room-worker/handler.go` around line 1067: The subscriptions bulk insert completes and immediately emits the join-grace notice, but the handler then builds and writes the corresponding room-member documents. If that later MongoDB write fails or its transaction aborts, broadcast-worker still retains the accounts and can deliver channel payloads to their personal subjects for the full grace window even though the add-member operation failed and membership is incomplete. Retain the joined accounts locally, but publish the notice only after every membership-related MongoDB write has succeeded and any transaction has committed.

```

</details>

broadcast-worker/handler.go
```diff
@@ -876,7 +897,61 @@ func (h *Handler) publishChannelEvent(ctx context.Context, meta *roommetacache.M
...
+// NATS exposes no signal for when a subscription is live cluster-wide — but they
+// have held their own user subject since login, so this delivery cannot race.
+// Clients dedupe by message id.
+//
+// Best-effort: the room-subject publish has already happened, so returning an error
+// here would only produce a duplicate on JetStream redelivery.
+func (h *Handler) fanOutToFreshMembers(ctx context.Context, roomID string, payload []byte) {
+	// The steady-state cost of the join window is exactly this: one map lookup
+	// that misses. No roster query, no extra round trip - a channel message
+	// stays a single publish to the room subject.
```
julianshen, line 917: _🔒 security_ | _🔴 Blocking_ | _`nats`_

**Removed members keep receiving room events during the grace window.**

`Fresh` selects recipients solely by join age, and the new registry has no removal/invalidation path or authoritative membership check. If a member is added and then kicked or leaves within `JOIN_GRACE`, subsequent channel messages are still published to `chat.user.{account}.event.room`, bypassing their removal from the room subscription for the rest of the TTL. Validate fresh candidates against current membership before publishing, or add a durable, ordered removal mechanism that purges them before later room fan-out.

<details>

```
In `broadcast-worker/handler.go` around line 917: `Fresh` selects recipients solely by join age, and the new registry has no removal/invalidation path or authoritative membership check. If a member is added and then kicked or leaves within `JOIN_GRACE`, subsequent channel messages are still published to `chat.user.{account}.event.room`, bypassing their removal from the room subscription for the rest of the TTL. Validate fresh candidates against current membership before publishing, or add a durable, ordered removal mechanism that purges them before later room fan-out.

```

</details>

room-worker/handler.go
```diff
@@ -106,6 +106,32 @@ func (h *Handler) bustRoomMeta(ctx context.Context, roomID string) {
...
+	}
+	evt := model.JoinGraceEvent{
+		RoomID:    roomID,
+		Accounts:  accounts,
+		JoinedAt:  joinedAt.UTC().UnixMilli(),
+		SiteID:    h.siteID,
+		Timestamp: time.Now().UTC().UnixMilli(),
+	}
+	data, err := json.Marshal(evt)
+	if err != nil {
```
julianshen, line 126: _🎯 NATS correctness_ | _🔴 Blocking_ | _`nats`_

**Core notices cannot guarantee grace state before event consumption.**

`chat.server.joingrace.{siteID}` is a fire-and-forget Core NATS notice, while channel events reach competing broadcast-worker replicas through an independent JetStream consumer. A canonical event can be handled before that replica's notice callback records the join; a newly started or reconnecting replica can also consume events after having missed the notice entirely because Core NATS provides no replay. In either path the grace copy is skipped, and the documented combination of an inactive room subscription and a lagging message-worker loses the message from both live and history paths. Commit join state to shared/durable storage before emitting dependent canonical events, or use an ordered durable mechanism that every replica catches up before it begins consuming those events.

<details>

```
In `room-worker/handler.go` around line 126: `chat.server.joingrace.{siteID}` is a fire-and-forget Core NATS notice, while channel events reach competing broadcast-worker replicas through an independent JetStream consumer. A canonical event can be handled before that replica's notice callback records the join; a newly started or reconnecting replica can also consume events after having missed the notice entirely because Core NATS provides no replay. In either path the grace copy is skipped, and the documented combination of an inactive room subscription and a lagging message-worker loses the message from both live and history paths. Commit join state to shared/durable storage before emitting dependent canonical events, or use an ordered durable mechanism that every replica catches up before it begins consuming those events.

```

</details>

julianshen (commented):
tGD inline review

tools/roomrace/e2e/stack.sh
```diff
@@ -0,0 +1,105 @@
...
+  mkdir -p "$BIN" "$LOGS"
+
+  grep -qE "^127\.0\.0\.1 .*valkey" /etc/hosts || \
+    echo "127.0.0.1 valkey mongodb cassandra nats" >> /etc/hosts
+
+  [ -f "$ROOT/docker-local/backend.creds" ] || "$ROOT/docker-local/setup.sh"
+
+  docker compose -f "$ROOT/docker-local/compose.deps.yaml" up -d --wait nats mongodb cassandra valkey
+  # A warm Cassandra volume makes the init report "already exists"; that is fine.
+  docker compose -f "$ROOT/docker-local/compose.deps.yaml" --profile init run --rm cassandra-init || true
```
julianshen, line 64: _🏷️ cassandra-initialization_ | _🟠 Warning_ | _`cassandra`_

**Cassandra initialization failures are silently ignored.**

The unconditional `|| true` suppresses every `cassandra-init` failure, not just harmless “already exists” responses. A CQL syntax error, connectivity failure, or incomplete schema migration therefore lets seeding and services start against a missing or incompatible Cassandra schema, producing invalid E2E results. Make the initialization idempotent with `IF NOT EXISTS` and fail on other errors, or explicitly recognize only the expected already-exists outcome.

<details>

```
In `tools/roomrace/e2e/stack.sh` around line 64: The unconditional `|| true` suppresses every `cassandra-init` failure, not just harmless “already exists” responses. A CQL syntax error, connectivity failure, or incomplete schema migration therefore lets seeding and services start against a missing or incompatible Cassandra schema, producing invalid E2E results. Make the initialization idempotent with `IF NOT EXISTS` and fail on other errors, or explicitly recognize only the expected already-exists outcome.

```

</details>

broadcast-worker/handler.go
```diff
@@ -876,7 +897,61 @@ func (h *Handler) publishChannelEvent(ctx context.Context, meta *roommetacache.M
...
+	var evt model.JoinGraceEvent
+	if err := json.Unmarshal(data, &evt); err != nil {
+		slog.WarnContext(ctx, "malformed join-grace notice", "error", err)
+		return
+	}
+	joinedAt := h.now()
+	if evt.JoinedAt > 0 {
+		joinedAt = time.UnixMilli(evt.JoinedAt).UTC()
+	}
+	h.joins.Record(evt.RoomID, evt.Accounts, joinedAt)
```
julianshen, line 952: _🏷️ time-consistency_ | _🟠 Warning_ | _`distributed-system`_

**Cross-host clock skew can expire a new join immediately.**

Expiry compares broadcast-worker's clock with `JoinedAt` produced by room-worker. If the producer clock is behind by at least `JOIN_GRACE`, a newly received notice is immediately discarded; a clock ahead retains it beyond the configured window. Base this process-local TTL on receipt time from the same clock used for expiry. Keeping a monotonic component in the production clock would also avoid wall-clock adjustments changing TTL behavior.

<details>

> ‼️ **IMPORTANT**

```suggestion
joinedAt := h.now()
```

</details>

<details>

```
In `broadcast-worker/handler.go` around line 949: Expiry compares broadcast-worker's clock with `JoinedAt` produced by room-worker. If the producer clock is behind by at least `JOIN_GRACE`, a newly received notice is immediately discarded; a clock ahead retains it beyond the configured window. Base this process-local TTL on receipt time from the same clock used for expiry. Keeping a monotonic component in the production clock would also avoid wall-clock adjustments changing TTL behavior.

```

</details>

broadcast-worker/joinregistry.go
```diff
@@ -0,0 +1,104 @@
...
+		return
+	}
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	room := r.rooms[roomID]
+	if room == nil {
+		room = make(map[string]time.Time, len(accounts))
+		r.rooms[roomID] = room
+	}
```
julianshen, line 43: _⚡ memory-exhaustion_ | _🟠 Warning_ | _`oom`_

**The join registry has no cardinality limit.**

With JOIN_GRACE enabled, every room creation or add-members operation inserts each unique room/account pair into this in-memory map on every broadcast-worker replica. Although normal entries expire by age, there is no per-room or global capacity bound, so a burst of large membership batches or many unique room joins within one grace window can grow memory to the full burst cardinality on every replica, causing GC thrashing, container eviction, or process OOM. A future-dated JoinedAt also defeats the intended age bound because it is accepted without validation. Add limits on accounts per notice and total registry entries with eviction/drop behavior, and reject or clamp timestamps outside a small allowed range around the local clock.

<details>

```
In `broadcast-worker/joinregistry.go` around line 43: With JOIN_GRACE enabled, every room creation or add-members operation inserts each unique room/account pair into this in-memory map on every broadcast-worker replica. Although normal entries expire by age, there is no per-room or global capacity bound, so a burst of large membership batches or many unique room joins within one grace window can grow memory to the full burst cardinality on every replica, causing GC thrashing, container eviction, or process OOM. A future-dated JoinedAt also defeats the intended age bound because it is accepted without validation. Add limits on accounts per notice and total registry entries with eviction/drop behavior, and reject or clamp timestamps outside a small allowed range around the local clock.

```

</details>

broadcast-worker/joinregistry.go
```diff
@@ -0,0 +1,104 @@
...
+	for _, account := range accounts {
+		if account != "" {
+			room[account] = joinedAt
+		}
+	}
+	r.sweepLocked()
```
julianshen, line 49: _🎯 performance/concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**Every join globally sweeps the registry under the message-path lock.**

With J joins per second and grace duration T, the registry contains roughly J*T entries, yet every Record scans all of them while holding the global write lock. Join bursts therefore produce quadratic work and block every concurrent Fresh call used by channel-message handlers. Fresh adds another herd: multiple handlers can observe an expired entry under the read lock, then each queue for the write lock and repeat the full sweep. This can stall broadcast processing and backlog the serial join-notice callback. Record should update only its target room; expiration should be amortized with a single periodic/bucketed sweeper or limited to the target room with coalesced cleanup.

<details>

```
In `broadcast-worker/joinregistry.go` around line 49: With J joins per second and grace duration T, the registry contains roughly J*T entries, yet every Record scans all of them while holding the global write lock. Join bursts therefore produce quadratic work and block every concurrent Fresh call used by channel-message handlers. Fresh adds another herd: multiple handlers can observe an expired entry under the read lock, then each queue for the write lock and repeat the full sweep. This can stall broadcast processing and backlog the serial join-notice callback. Record should update only its target room; expiration should be amortized with a single periodic/bucketed sweeper or limited to the target room with coalesced cleanup.

```

</details>

broadcast-worker/joinregistry.go
julianshen, line 46: _🎯 concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**An older notice can overwrite a newer join timestamp.**

Join notices from separate room-worker replicas have no global ordering. If a member leaves and rejoins quickly, the newer notice can be recorded first and a delayed older notice can then overwrite it. The account consequently expires according to the old join and can miss grace-window delivery after the actual rejoin. Keep timestamps monotonic per room/account by retaining the later value.

<details>

```
In `broadcast-worker/joinregistry.go` around line 46: Join notices from separate room-worker replicas have no global ordering. If a member leaves and rejoins quickly, the newer notice can be recorded first and a delayed older notice can then overwrite it. The account consequently expires according to the old join and can miss grace-window delivery after the actual rejoin. Keep timestamps monotonic per room/account by retaining the later value.

```

</details>

tools/roomrace/e2e/client.go
```diff
@@ -0,0 +1,345 @@
...
+	deadline := time.Now().Add(3 * time.Second)
+	tries := 0
+	var last []histMsg
+	for {
+		tries++
+		req := map[string]any{"limit": 50}
+		if *hint || forceHint {
+			// A hint the client already holds: the room is at least as fresh as
+			// the event it just received. Without it the scan ceiling is the room
+			// doc's lastMsgAt, which the just-sent message has not reached yet.
```
julianshen, line 321: _🎯 correctness_ | _🟠 Warning_ | _`silent-failure`_

**History failures are converted into valid-looking missing data.**

`history` logs request and JSON-decoding failures but returns the previous result, while callers have no error to propagate. A timeout, cancellation, or malformed reply therefore becomes empty or partial history; `runOne`, `probe`, and write-lag measurement continue successfully and can report apparent message loss or lag caused solely by the failed RPC. Return an error from `history`/`loadHistory` and propagate it through every measurement entry point so infrastructure failures abort the run rather than alter its results.

<details>

```
In `tools/roomrace/e2e/client.go` around line 314: `history` logs request and JSON-decoding failures but returns the previous result, while callers have no error to propagate. A timeout, cancellation, or malformed reply therefore becomes empty or partial history; `runOne`, `probe`, and write-lag measurement continue successfully and can report apparent message loss or lag caused solely by the failed RPC. Return an error from `history`/`loadHistory` and propagate it through every measurement entry point so infrastructure failures abort the run rather than alter its results.

```

</details>

broadcast-worker/handler.go
```diff
@@ -876,7 +897,61 @@ func (h *Handler) publishChannelEvent(ctx context.Context, meta *roommetacache.M
...
+	if len(accounts) == 0 {
+		return
+	}
+	sent := 0
+	for _, account := range accounts {
+		if isBot(account) {
+			continue
+		}
+		sent++
+		if err := h.pub.Publish(ctx, subject.UserRoomEvent(account), payload); err != nil {
```
julianshen, line 928: _🏷️ delivery_ | _🔴 Blocking_ | _`distributed-system`, `tgd-review` (2 rules)_

**Failed grace-copy publishes are permanently acknowledged.**

If the room publish succeeds but this per-user publish fails during a connection transition, the failure is only logged. publishChannelEvent consequently returns the successful room-publish result, allowing the canonical JetStream event to be acknowledged. A fresh member whose room subscription is not yet active receives no retry, and a lagging writer leaves no history fallback. Aggregate and return transient fan-out errors so the canonical event is retried with message-ID deduplication, or enqueue each recipient copy durably.

<details>

- **`tgd-review`**:928: A failed user-subject publish is only logged, and publishChannelEvent returns the room-subject result. If the room publish succeeds, the canonical event can therefore be acknowledged without retrying the failed grace copy. Return an aggregate fan-out error so the durable event is retried, relying on the documented message-ID deduplication, or enqueue recipient copies durably.

</details>

<details>

```
In `broadcast-worker/handler.go` around line 928: If the room publish succeeds but this per-user publish fails during a connection transition, the failure is only logged. publishChannelEvent consequently returns the successful room-publish result, allowing the canonical JetStream event to be acknowledged. A fresh member whose room subscription is not yet active receives no retry, and a lagging writer leaves no history fallback. Aggregate and return transient fan-out errors so the canonical event is retried with message-ID deduplication, or enqueue each recipient copy durably.

```

</details>

broadcast-worker/handler.go
```diff
@@ -876,7 +897,61 @@ func (h *Handler) publishChannelEvent(ctx context.Context, meta *roommetacache.M
...
+}
+
+// HandleJoinGraceNotice records a room-worker join notice. Fire-and-forget: a
+// dropped notice costs this replica the window for that join, nothing more.
+func (h *Handler) HandleJoinGraceNotice(ctx context.Context, data []byte) {
+	var evt model.JoinGraceEvent
+	if err := json.Unmarshal(data, &evt); err != nil {
+		slog.WarnContext(ctx, "malformed join-grace notice", "error", err)
+		return
+	}
```
julianshen, line 947: _🏷️ time-consistency_ | _🟠 Warning_ | _`distributed-system`_

**Cross-host clocks determine whether a new join is fresh.**

Expiry compares broadcast-worker's clock with JoinedAt generated by room-worker. If the producer clock trails by JOIN_GRACE, a newly received notice is swept immediately; if it leads, delivery continues beyond the configured window. Base the process-local TTL on receipt time using a monotonic clock, or validate and clamp the producer timestamp against a bounded skew policy.

<details>

```
In `broadcast-worker/handler.go` around line 947: Expiry compares broadcast-worker's clock with JoinedAt generated by room-worker. If the producer clock trails by JOIN_GRACE, a newly received notice is swept immediately; if it leads, delivery continues beyond the configured window. Base the process-local TTL on receipt time using a monotonic clock, or validate and clamp the producer timestamp against a bounded skew policy.

```

</details>

broadcast-worker/handler.go
```diff
@@ -876,7 +897,61 @@ func (h *Handler) publishChannelEvent(ctx context.Context, meta *roommetacache.M
 		"request_id", natsutil.RequestIDFromContext(ctx), "room_id", meta.ID,
 		"type", string(meta.Type), "delivery", "room-stream", "audience", meta.UserCount)
 	h.metrics.Fanout(ctx, roomChannel, broadcastLabels(ctx).eventType, meta.UserCount)
-	return h.publishRoomEvent(ctx, meta.ID, meta.CrossSite, meta.CrossSiteAt, payload, "channel event")
+	err = h.publishRoomEvent(ctx, meta.ID, meta.CrossSite, meta.CrossSiteAt, payload, "channel event")
```
julianshen, line 900: _🎯 performance-and-concurrency_ | _🟠 Warning_ | _`performance-and-concurrency`_

**A failed room publish still triggers the full grace fan-out.**

When the room-subject publish fails, execution still performs one sequential Publish per fresh member. With N recent joiners during a NATS disconnect or backpressure event, each canonical delivery produces N additional failing calls and warning logs before returning the room error; JetStream then retries the event and repeats this amplification across worker slots. Return immediately on the primary publish error. The durable redelivery can perform the grace fan-out after the room publish succeeds.

<details>

```
In `broadcast-worker/handler.go` around line 900: When the room-subject publish fails, execution still performs one sequential Publish per fresh member. With N recent joiners during a NATS disconnect or backpressure event, each canonical delivery produces N additional failing calls and warning logs before returning the room error; JetStream then retries the event and repeats this amplification across worker slots. Return immediately on the primary publish error. The durable redelivery can perform the grace fan-out after the room publish succeeds.

```

</details>

broadcast-worker/handler.go
```diff
@@ -876,7 +897,61 @@ func (h *Handler) publishChannelEvent(ctx context.Context, meta *roommetacache.M
...
+
+// HandleJoinGraceNotice records a room-worker join notice. Fire-and-forget: a
+// dropped notice costs this replica the window for that join, nothing more.
+func (h *Handler) HandleJoinGraceNotice(ctx context.Context, data []byte) {
+	var evt model.JoinGraceEvent
+	if err := json.Unmarshal(data, &evt); err != nil {
+		slog.WarnContext(ctx, "malformed join-grace notice", "error", err)
+		return
+	}
+	joinedAt := h.now()
```
julianshen, line 948: _🏷️ time-consistency_ | _🟠 Warning_ | _`tgd-review`_

**Producer clock skew changes the effective grace window.**

Expiry compares broadcast-worker's clock against JoinedAt generated on a room-worker host. Sufficient negative skew expires a newly received join immediately, while positive skew retains access beyond JOIN_GRACE. Base the process-local TTL on receipt time using the same clock as expiry, or validate and clamp remote timestamps.

<details>

```
In `broadcast-worker/handler.go` around line 948: Expiry compares broadcast-worker's clock against JoinedAt generated on a room-worker host. Sufficient negative skew expires a newly received join immediately, while positive skew retains access beyond JOIN_GRACE. Base the process-local TTL on receipt time using the same clock as expiry, or validate and clamp remote timestamps.

```

</details>

tools/roomrace/e2e/main.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+
+	a.openRoom(roomID)
+	b.waitRoomOpen(roomID, 3*time.Second)
+
+	aHist, _ := a.loadHistory(roomID, *backfill, msgID)
+	rec.add(*userA, "history", fmt.Sprintf("room-open msg.history returned %d", len(aHist)), "", "")
+	bHist, _ := b.loadHistory(roomID, *backfill, msgID)
+	rec.add(*userB, "history", fmt.Sprintf("room-open msg.history returned %d", len(bHist)), "", "")
+	// When the client joined via subscription.update it already read history
```
julianshen, line 215: _🧪 tests_ | _🟠 Warning_ | _`tgd-review`_

**The stalled-writer test omits unavailable expected messages.**

The expected set is built from a history read whose incompleteness is the condition being measured. When message-worker is stalled, missing system messages are absent from final and therefore never counted as lost; only msgID is forcibly included. Obtain expected message IDs from a writer-independent source or collect them after recovery before calculating results.

<details>

```
In `tools/roomrace/e2e/main.go` around line 215: The expected set is built from a history read whose incompleteness is the condition being measured. When message-worker is stalled, missing system messages are absent from final and therefore never counted as lost; only msgID is forcibly included. Obtain expected message IDs from a writer-independent source or collect them after recovery before calculating results.

```

</details>

tools/roomrace/interest.go
```diff
@@ -0,0 +1,111 @@
...
+	windows := make([]time.Duration, 0, samples)
+	flushes := make([]time.Duration, 0, samples)
+
+	for i := range samples {
+		subj := fmt.Sprintf("roomrace.interest.%d.%d", time.Now().UnixNano(), i)
+		stop := make(chan struct{})
+		done := make(chan struct{})
+		go func() {
+			defer close(done)
+			buf := make([]byte, 8)
```
julianshen, line 41: _🧪 tests_ | _🟠 Warning_ | _`tgd-review`_

**The interest timer starts before Subscribe returns.**

The publisher goroutine starts before subscription creation, and t0 is captured before Subscribe rather than after it returns. The resulting duration does not measure the documented post-Subscribe interest-propagation interval. Create the subscription first, capture t0 after successful return, and then release a gated publisher.

<details>

```
In `tools/roomrace/interest.go` around line 41: The publisher goroutine starts before subscription creation, and t0 is captured before Subscribe rather than after it returns. The resulting duration does not measure the documented post-Subscribe interest-propagation interval. Create the subscription first, capture t0 after successful return, and then release a gated publisher.

```

</details>

tools/roomrace/e2e/stack.sh
```diff
@@ -0,0 +1,105 @@
...
+    message-worker) echo 18085 ;;
+    history-service) echo 18086 ;;
+    inbox-worker) echo 18087 ;;
+  esac
+}
+
+up() {
+  mkdir -p "$BIN" "$LOGS"
+
+  grep -qE "^127\.0\.0\.1 .*valkey" /etc/hosts || \
```
julianshen, line 57: _🏷️ usability_ | _🟠 Warning_ | _`tgd-review`_

**The harness requires write access to /etc/hosts.**

A normal developer account cannot append to /etc/hosts, so `set -e` aborts startup. The exported dependency addresses already use localhost. Remove this implicit privileged modification or document any genuinely required host setup as an explicit prerequisite.

<details>

```
In `tools/roomrace/e2e/stack.sh` around line 57: A normal developer account cannot append to /etc/hosts, so `set -e` aborts startup. The exported dependency addresses already use localhost. Remove this implicit privileged modification or document any genuinely required host setup as an explicit prerequisite.

```

</details>

julianshen (commented):
tGD inline review

PR #281, opened 2026-08-15
fix: keep room key versions resolvable when members are removed in quick succession

hmchangw:
## Problem

Removing several room members in quick succession can produce broadcast messages that **no client can ever decrypt**. Two independent defects combine, and neither fix substitutes for the other.

**A cached key can outlive the store's ability to explain it.** `broadcast-worker` caches room keys in an in-process LRU (`ROOM_KEY_CACHE_TTL`, 10m) and stamps the cached version into every envelope. The store retained exactly **one** previous version, so two removals inside one cache TTL take it from `v` to `v+2`: the cache still serves `v`, the previous slot now holds `v+1`, and the client's `key.get` repair RPC returns `errRoomKeyAbsent`.

**Fan-out labelled keys with a version the store had not assigned.** Both `room-worker` and `bot-room-service` computed `predictedVersion = currentPair.Version + 1` and fanned it to survivors *before* calling `Rotate`. `room-worker` runs ROOMS messages on 100 concurrent goroutines with no per-room serialization, so two simultaneous removals both predict `v+1` while the store assigns `v+1` and `v+2`. Survivors hold bytes under a version belonging to a different key, and the frontend short-circuits refetches for versions it already holds — so that state never recovers.

Impact is partial, which is why it reads as intermittent: clients that already held the version keep it in an untrimmed session map and still decrypt; clients that joined later or reloaded cannot. Live broadcast only — Cassandra history is not encrypted with this key.

### Retention: a separate collection, not a wider slot

The obvious fix is to keep more previous versions. *Where* is the question that drives everything else.

Retention must be bounded by **time**, because the thing it has to outlast — a cache entry — is bounded by time. Keeping the last N versions bounds it by **count**, and N is a guess: rotations per cache window are driven by how many members an admin removes. A count-capped ring is a probabilistic fix dressed as a deterministic one.

A per-version TTL gives the right bound, and MongoDB offers one — but only at document granularity. That rules out a field of the room document (a TTL index there deletes the whole room) and points at a separate collection: **`retired_room_keys`**, one document per `(roomID, version)`, `_id = {roomID}:{version}`, expired by a TTL index on `expiresAt` with `expireAfterSeconds: 0`.

Rejected alternatives:

| Option | Why not |
|---|---|
| Expiry-pruned ring inside `encKey` | Grows a document many services read for unrelated reasons; overloads the 24h grace period with a second meaning |
[...]

PR #363, opened 2026-08-24
fix(user-service): compile badge_test against the ActiveSubscription return type

chenjr0719:
## Summary

`main` currently fails the `ci / test` (and `lint`) job because `user-service/service/badge_test.go` does not compile: two `DoAndReturn` closures still declare the pre-rename return type `[]model.EnrichedSubscription`, while the real `SubscriptionRepository.GetActiveSubscriptions` signature returns `[]models.ActiveSubscription` (`user-service/service/service.go`, the mongorepo impl, and the generated mock all agree). This is a leftover from the type rename — 17 references were updated, these two closures were missed — so `go test ./user-service/service/...` fails to build and reds the whole `make test` run for every branch that rebases on `main`.

## Changes

- `user-service/service/badge_test.go`: change the two `GetActiveSubscriptions` `DoAndReturn` closure return types from `[]model.EnrichedSubscription` to `[]models.ActiveSubscription` (4 lines) so the test compiles against the current repository signature. Test-only; no production code changes.

## Verification

- `go vet ./user-service/service/` — passes (previously failed to compile).
- `go test ./user-service/service/ -run Badge` — passes.
- `golangci-lint` on `./user-service/...` — clean.

## Summary by CodeRabbit

* **Tests**
  * Updated badge budget-exhaustion test scenarios to use the current subscription representation.
  * Preserved existing test behavior and coverage.

### DIFF
--- a/user-service/service/badge_test.go
+++ b/user-service/service/badge_test.go
@@ -221,9 +221,9 @@ func TestBadgeCountBatch_BudgetSpentMidBatch_StopsAndRemainingAbsent(t *testing.
 	// Account "a" computes, then the shared budget expires (simulating the 15s
 	// handler deadline elapsing part-way through the batch).
 	subs.EXPECT().GetActiveSubscriptions(gomock.Any(), "a", gomock.Any()).
-		DoAndReturn(func(_ context.Context, _ string, _ int) ([]model.EnrichedSubscription, error) {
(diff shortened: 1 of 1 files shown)


chenjr0719:
Superseded — #281 (just merged) already carries this badge_test fix, so main compiles green again. Closing as redundant. Thanks!

PR #371, opened 2026-08-25
fix(user-service): emit subscription.update on setAppSubscription un/re-subscribe

### LINKED ISSUE #370: [Fix] setAppSubscription flips isSubscribed without emitting subscription.update — app list doesn't sync across devices
> **Verified against:** `main`@e7930903 — code read after a fresh `git pull` (code is source of truth over docs).

## Summary
`subscription.setAppSubscription {subscribed:false}` unsubscribes a user's app (flips the botDM subscription's `isSubscribed` to `false`) but publishes **no** `subscription.update` event. For a user with two active clients (e.g. mobile + desktop), the device that made the call updates, but the **other device gets no signal** — its `apps.list` view stays stale until a manual refetch. Expected: both devices reflect the change in real time, exactly like every other subscription change.

## Reproduce steps
- **Preconditions:** a user (`alice`) with an active botDM subscription to an app (`isSubscribed=true`); `site-local`; an app doc whose `assistant.name` matches the bot and `assistant.enabled=true`.
- **Steps:**
  1. Open two client connections for the same account (dev1, dev2). Both subscribe to `chat.user.{account}.event.subscription.update`.
  2. Confirm `apps.list` shows the app with `isSubscribed=true`.
  3. From dev1, request `chat.user.{account}.request.user.{siteId}.subscription.setAppSubscription` with body `{"appId":"…","subscribed":false}` → reply `{"success":true}`.
  4. Watch both devices' `subscription.update` stream and re-query `apps.list`.
- **Expected vs actual:** expected a `subscription.update` (action removed/updated) on **both** devices so each flips/prunes the app; actual — **neither** device receives any event, though `apps.list` now returns `isSubscribed:false`.
- **Evidence (verified on our dev stack, two-connection wire test):** before `isSubscribed=true` → after `isSubscribed=false`; device1 events = 0, device2 events = 0. Reproduced twice.

## Root cause  (verified, file:line)
- `user-service/service/apps.go:11-49` — `SetAppSubscription`: on the **unsubscribe** (and **reactivate**) branch it calls `SetAppSubscribed(...)` and returns `{success:true}` with **no event publish**.
- `user-service/mongorepo/subscriptions.go:819` — `SetAppSubscribed`: a bare `UpdateOne` `$set{isSubscribed, muted}`. Pure persistence, no fan-out.
- By contrast `subscription.update` (`chat.user.{account}.event.subscription.update`, `pkg/subject/subject.go:260`; `model.SubscriptionUpdateEvent`/`SubscriptionRemovedEvent`, `pkg/model/event.go:76,523`) **is** published on every other subscription change — `room-worker/handler.go:436,649,2191`, `room-service/handler.go:837`. The app-subscription path is the sole omission.
- **Asymmetry:** first-time subscribe (no existing sub) routes through `CreateDMRoom`, which **does** emit → both devices sync; only **unsubscribe** and **reactivate** (existing sub → the silent `UpdateOne`) skip it.

## Message-path trace
1. Client request struct + docs/client-api.md — `SetAppSubscriptionRequest{appId, subscribed}` (`user-service/models/app.go:6`); docs entry for the RPC needs the now-emitted event noted. **change**
2. Gatekeeper mapping (request → canonical) — **N/A** (user-service request/reply, not the message gatekeeper).
3. Canonical model / event structs (pkg/model) — reuse the existing `SubscriptionUpdateEvent` / `SubscriptionRemovedEvent` (`pkg/model/event.go`). **no change** (or a minimal appId field if the FE needs it).
4. Workers (message-worker / broadcast-worker / notification-worker) — **N/A** (handled in user-service; no message/broadcast worker on this path).
5. Storage tables (Cassandra / Mongo) — **no change**; the existing `subscriptions` doc `isSubscribed`/`muted` is already written.
6. Read / history path (history-service) — **no change**; `apps.list` (`ListApps`) already derives `isSubscribed` from the sub.
7. Client-facing events (Leg-2 fan-out) — **the fix**: publish `subscription.update` to `chat.user.{account}.event.subscription.update` on unsubscribe + reactivate so both devices react. **change**

## Fix
**Primary (recommended):** in `SetAppSubscription`, after a successful `SetAppSubscribed` on **both** the unsubscribe and reactivate branches, publish a `subscription.update` to the user's own subject (`subject.UserEventSubscriptionUpdate(account)`) carrying the app subscription's new state (action `removed` for `subscribed:false`, `updated` for reactivate) — mirroring room-service/room-worker's existing publish. Smallest change that makes the two silent paths consistent with the rest of the system; no schema change, and the event type is already defined and already consumed by clients.

Alternatives considered:
- (a) Publish from `SetAppSubscribed` (the repo) — rejected: the repo layer is pure persistence; publishing belongs in the service/handler per the codebase convention that stores don't fan out.
- (b) Rely on the FE polling `apps.list` — rejected: inconsistent with every other sub change and doesn't deliver real-time multi-device sync.

## Contracts & docs
- docs/client-api.md updated same PR? **Y** — note that `subscription.setAppSubscription` now emits `subscription.update`, and update the derived events view.
- Schema change? **N** (no Cassandra DDL, no Mongo index change).

## Tests
- TDD red→green: unit test on `SetAppSubscription` asserting a publish occurs on the unsubscribe **and** reactivate branches (inject the publish fn per the repo's test convention) — red first, then fix.
- **e2e (dev stack):** required (multi-device event path) — drive `setAppSubscription{subscribed:false}` and assert `subscription.update` now reaches a second connection. The failing-state harness already exists.

## Acceptance criteria
- `setAppSubscription{subscribed:false}` publishes `subscription.update` to the caller's own subject; a second active client receives it and flips the app's `isSubscribed` without a refetch.
- Reactivate (`subscribed:true` on an existing sub) likewise emits.
- First-time subscribe behavior unchanged (already emits via room creation).
- Unit + e2e cover the regression.

- Exact event shape for the app case (reuse `SubscriptionRemovedEvent` with the botDM roomId vs a `SubscriptionUpdateEvent` carrying appInfo). Recommend matching the existing botDM `subscription.update` shape so no new FE handler is needed — FE to confirm which it consumes.

## PR / merge
- pr-self-audit is run as a pre-review gate; its result is **not** included in the PR body.
- Source repo `hmchangw/newchat`: external review + external merge — open/rebase/CI-green/request-review, do not solo-merge.

chenjr0719:
## Summary

`subscription.setAppSubscription` flipped `isSubscribed` in Mongo but published no
`subscription.update`, so a user's other devices never learned an app was un/re-subscribed
— `apps.list` stayed stale until a manual refetch. Every other subscription change already
fans out `subscription.update`; this path was the sole omission. Fixes #370.

This publishes the event on the two affected branches. First-time subscribe already emits
via room creation, so it is left unchanged.

## What's included

- Publish `subscription.update` on **unsubscribe** (`action: "removed"`, the lean
  `SubscriptionRemovedEvent` room-worker already uses) and on **reactivate**
  (`action: "added"`, the same shape botDM first-subscribe emits — the other devices
  dropped the room on the prior `removed`, so re-adding is correct and needs no new FE
  handler).
- Docs: `docs/client-api.md` (`setAppSubscription` now lists a Triggered-events section)
  and the `docs/client-api/events.md` `subscription.update` view.
- Tests covering both branches plus the no-existing-subscription no-op.

## Changes

- `user-service/service/apps.go` — fetch the existing botDM subscription up front (both
  branches need it: the removed event carries the room id; each keys off existence), then
  publish over the same core-NATS client subject the sibling `settings.update` /
  `chatlist.update` fanouts use. The publish is best-effort — the Mongo write stays the
  source of truth, a publish failure is logged and clients reconcile on next refetch.
- `user-service/service/apps_test.go` — assert the publish + payload (action, room id,
  `appInfo`, cleared `muted`) on both branches; add the no-subscription no-op case.
- `docs/client-api.md`, `docs/client-api/events.md` — doc-ratchet.

The subscription repository stays pure (no fan-out from the data layer) — the handler owns
the event, matching the write-owner-emits convention.

## Verification

- `make test SERVICE=user-service` — green (all packages), 13 `SetAppSubscription` cases pass.
- Lint clean on the touched service.
- Red→green: the publish/branch assertions fail against the pre-fix handler and pass after.

🤖 Generated with Claude Code

## Summary by CodeRabbit

- **New Features**
  - Unsubscribing from an app subscription now emits a removal update.
  - Reactivating a subscription emits an added update and restores the subscription.
  - Attempting to unsubscribe without an existing subscription succeeds without creating an update.
  - Reapplying an already active subscription does not emit a duplicate update.
  - New subscriptions continue to emit the standard room-creation update.

- **Documentation**
  - Updated API documentation to describe subscription update events for subscribing, unsubscribing, and reactivating.

### DIFF
--- a/user-service/service/apps.go
+++ b/user-service/service/apps.go
@@ -1,11 +1,16 @@
 package service
 
 import (
+	"encoding/json"
 	"fmt"
+	"log/slog"
+	"time"
 
 	"github.com/hmchangw/chat/pkg/errcode"
 	"github.com/hmchangw/chat/pkg/model"
 	"github.com/hmchangw/chat/pkg/natsrouter"
+	"github.com/hmchangw/chat/pkg/natsutil"
+	"github.com/hmchangw/chat/pkg/subject"
 	"github.com/hmchangw/chat/user-service/models"
 )
 
@@ -27,17 +32,24 @@ func (s *UserService) SetAppSubscription(c *natsrouter.Context, req models.SetAp
 	}
 	botName := app.Assistant.Name
 
-	if !req.Subscribed {
-		if err := s.subs.SetAppSubscribed(c, account, botName, false, true); err != nil {
-			return nil, fmt.Errorf("unsubscribe app: %w", err)
-		}
-		return &models.OKResponse{Success: true}, nil
-	}
+	// Both branches need this: the removed event carries the room id, and each keys off existence.
 	existing, err := s.subs.GetAppSubscription(c, account, botName)
 	if err != nil {
 		return nil, fmt.Errorf("get app subscription: %w", err)
 	}
+
+	if !req.Subscribed {
+		if existing == nil {
+			return &models.OKResponse{Success: true}, nil // nothing subscribed: no write, no event
+		}
+		if err := s.subs.SetAppSubscribed(c, account, botName, false, true); err != nil {
+			return nil, fmt.Errorf("unsubscribe app: %w", err)
+		}
+		s.publishAppSubscriptionRemoved(c, account, existing)
+		return &models.OKResponse{Success: true}, nil
+	}
 	if existing == nil {
+		// First subscribe emits via room creation — nothing to publish here.
 		if _, err := s.rooms.CreateDMRoom(c, account, botName, model.RoomTypeBotDM); err != nil {
 			return nil, fmt.Errorf("create botDM room: %w", err)
 		}
@@ -46,7 +58,52 @@ func (s *UserService) SetAppSubscription(c *natsrouter.Context, req models.SetAp
 	if err := s.subs.SetAppSubscribed(c, account, botName, true, false); err != nil {
 		return nil, fmt.Errorf("reactivate app: %w", err)
 	}
+	s.publishAppSubscriptionReactivated(c, account, existing, app)
 	return &models.OKResponse{Success: true}, nil
+}
+
+// Tell the user's other devices to drop the botDM (mirrors room-worker's removed event; best-effort).
+func (s *UserService) publishAppSubscriptionRemoved(c *natsrouter.Context, account string, sub *model.Subscription) {
+	evt := model.SubscriptionRemovedEvent{
+		UserID: sub.User.ID,
+		Subscription: model.RemovedSubscriptionRef{
+			RoomID:   sub.RoomID,
+			RoomType: model.RoomTypeBotDM,
+			U:        model.SubscriptionUser{ID: sub.User.ID, Account: account},
+		},
+		Action:    "removed",
+		Timestamp: time.Now().UTC().UnixMilli(),
+	}
+	data, _ := json.Marshal(evt) // primitives only; cannot fail
+	s.publishSubscriptionUpdate(c, account, data)
+}
+
+// Tell the user's other devices to re-add the botDM (mirrors room-worker's added shape; best-effort).
+func (s *UserService) publishAppSubscriptionReactivated(c *natsrouter.Context, account string, sub *model.Subscription, app *model.App) {
+	subCopy := *sub
+	subCopy.IsSubscribed = true
+	subCopy.Muted = false
+	roomName := app.Name
+	if roomName == "" {
+		roomName = app.Assistant.Name
+	}
+	evt := model.SubscriptionUpdateEvent{
+		UserID:       sub.User.ID,
+		Subscription: subCopy,
+		Action:       "added",
+		RoomName:     roomName,
+		AppInfo:      &model.CounterpartAppInfo{ID: app.ID, Name: app.Name, AssistantName: app.Assistant.Name},
+		Timestamp:    time.Now().UTC().UnixMilli(),
+	}
+	data, _ := json.Marshal(evt) // primitives only; cannot fail
+	s.publishSubscriptionUpdate(c, account, data)
+}
+
+// Best-effort core-NATS fan-out of a botDM subscription.update to the user's own subject.
+func (s *UserService) publishSubscriptionUpdate(c *natsrouter.Context, account string, data []byte) {
+	if err := s.clientPub.Publish(c, subject.SubscriptionUpdate(account), data); err != nil {
+		slog.WarnContext(c, "publish app subscription.update event", "error", err, "account", account, "request_id", natsutil.RequestIDFromContext(c))
+	}
 }
 
 func (s *UserService) ListApps(c *natsrouter.Context, req models.AppsListRequest) (*models.AppsListResponse, error) {
--- a/user-service/service/apps.go
+++ b/user-service/service/apps.go
@@ -86,10 +86,15 @@ func (s *UserService) publishAppSubscriptionReactivated(c *natsrouter.Context, a
 	subCopy := *sub
 	subCopy.IsSubscribed = true
 	subCopy.Muted = false
-	// Hydrate the non-persisted Room the "added" contract carries (GetAppSubscription decodes persisted fields only).
-	if infos, err := s.rooms.GetRoomsMeta(c, sub.SiteID, []string{sub.RoomID}); err == nil && len(infos) > 0 {
-		applyRoomInfo(&subCopy, &infos[0])
+	// The "added" contract requires a hydrated Room (GetAppSubscription decodes persisted fields only).
+	// If the lookup fails, skip the publish rather than emit a Room-less event — clients reconcile on next apps.list.
+	infos, err := s.rooms.GetRoomsMeta(c, sub.SiteID, []string{sub.RoomID})
+	if err != nil || len(infos) == 0 || !infos[0].Found {
+		slog.WarnContext(c, "skip app subscription.update: room metadata unavailable",
+			"error", err, "account", account, "room_id", sub.RoomID, "request_id", natsutil.RequestIDFromContext(c))
+		return
 	}
+	applyRoomInfo(&subCopy, &infos[0])
 	roomName := app.Name
 	if roomName == "" {
 		roomName = app.Assistant.Name
--- a/user-service/service/apps.go
+++ b/user-service/service/apps.go
@@ -55,10 +55,13 @@ func (s *UserService) SetAppSubscription(c *natsrouter.Context, req models.SetAp
 		}
 		return &models.OKResponse{Success: true}, nil
 	}
+	wasSubscribed := existing.IsSubscribed
 	if err := s.subs.SetAppSubscribed(c, account, botName, true, false); err != nil {
 		return nil, fmt.Errorf("reactivate app: %w", err)
 	}
-	s.publishAppSubscriptionReactivated(c, account, existing, app)
+	if !wasSubscribed { // emit only on a real unsubscribed→subscribed transition, not an idempotent re-subscribe
+		s.publishAppSubscriptionReactivated(c, account, existing, app)
+	}
 	return &models.OKResponse{Success: true}, nil
 }
 
@@ -83,6 +86,10 @@ func (s *UserService) publishAppSubscriptionReactivated(c *natsrouter.Context, a
 	subCopy := *sub
 	subCopy.IsSubscribed = true
 	subCopy.Muted = false
+	// Hydrate the non-persisted Room the "added" contract carries (GetAppSubscription decodes persisted fields only).
+	if infos, err := s.rooms.GetRoomsMeta(c, sub.SiteID, []string{sub.RoomID}); err == nil && len(infos) > 0 {
+		applyRoomInfo(&subCopy, &infos[0])
+	}
 	roomName := app.Name
 	if roomName == "" {
 		roomName = app.Assistant.Name
--- a/docs/client-api.md
+++ b/docs/client-api.md
@@ -5610,6 +5610,14 @@ PUT-like idempotent endpoint to subscribe or unsubscribe the calling user from a
 { "success": true }
 ```
 
+##### Triggered events — success path
+
+**`chat.user.{account}.event.subscription.update`** — emitted once for the requester (best-effort, core NATS) so the user's other devices reconcile the botDM without a refetch. Bot accounts receive it on their **encoded** per-user subject (dots→underscores), matching their NATS JWT scope.
+
+- **Unsubscribe** (`subscribed: false` on an existing subscription) → `action: "removed"`. Payload is the dedicated `SubscriptionRemovedEvent` (`subscription` carries only `roomId`, `roomType: "botDM"`, and `u`). No event fires when there was no subscription to remove.
+- **Reactivate** (`subscribed: true` on an existing, previously-unsubscribed subscription) → `action: "added"`, the same event the first-time subscribe path emits, so the FE re-adds the botDM it dropped on the prior `removed`. See the [subscription.update schema](#subscriptionupdate-event); the embedded `Subscription` reflects `isSubscribed: true` / `muted: false`, and `appInfo` carries the bot app's identity.
+- **First-time subscribe** (`subscribed: true`, no existing DM room) → the `action: "added"` event is emitted by room-service as part of botDM room creation, not by this handler.
+
 ##### Error response
 
 | Condition | `code` | `reason` | Notes |
--- a/docs/client-api/events.md
+++ b/docs/client-api/events.md
@@ -204,7 +204,8 @@ fields are sent.
 
 **Triggered by:** Add Members (`added`), Remove Member (`removed`), Update Member Role
 (`role_updated`), Toggle Mute (`mute_toggled`), Toggle Favorite (`favorite_toggled`),
-Open Room (`opened`), Mark Messages Read (`read`) — see [request-reply.md](request-reply.md).
+Open Room (`opened`), Mark Messages Read (`read`), Set App Subscription (`removed` on
+unsubscribe, `added` on reactivate) — see [request-reply.md](request-reply.md).
 
 ---
 
--- a/user-service/service/apps_test.go
+++ b/user-service/service/apps_test.go
@@ -126,6 +126,19 @@ func TestSetAppSubscription_Reactivate_ClearsMuted(t *testing.T) {
 	assert.NotZero(t, got.Timestamp)
 }
 
+// Room hydration failure on reactivate → skip the publish (no Room-less event), still success.
+func TestSetAppSubscription_Reactivate_RoomMetaUnavailable_NoEvent(t *testing.T) {
+	svc, subs, _, apps, rooms, _, _ := newSvc(t)
+	apps.EXPECT().GetApp(gomock.Any(), "app1").Return(appWith(true), nil)
+	subs.EXPECT().GetAppSubscription(gomock.Any(), "alice", "helper.bot").Return(appSub(false), nil)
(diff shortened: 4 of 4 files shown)


user-service/service/apps.go
```diff
@@ -46,9 +58,54 @@ func (s *UserService) SetAppSubscription(c *natsrouter.Context, req models.SetAp
 	if err := s.subs.SetAppSubscribed(c, account, botName, true, false); err != nil {
 		return nil, fmt.Errorf("reactivate app: %w", err)
 	}
+	s.publishAppSubscriptionReactivated(c, account, existing, app)
```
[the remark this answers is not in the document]
chenjr0719: Good catch — applied in e97e1bc2. The reactivate branch now captures the prior `IsSubscribed` and publishes `added` only on a real unsubscribed→subscribed transition; an idempotent re-subscribe writes but emits nothing. Added a test asserting the already-active case emits no event.

user-service/service/apps.go
```diff
@@ -46,9 +58,54 @@ func (s *UserService) SetAppSubscription(c *natsrouter.Context, req models.SetAp
...
+	roomName := app.Name
+	if roomName == "" {
+		roomName = app.Assistant.Name
+	}
+	evt := model.SubscriptionUpdateEvent{
+		UserID:       sub.User.ID,
+		Subscription: subCopy,
+		Action:       "added",
+		RoomName:     roomName,
+		AppInfo:      &model.CounterpartAppInfo{ID: app.ID, Name: app.Name, AssistantName: app.Assistant.Name},
```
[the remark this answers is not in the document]
chenjr0719: Confirmed and fixed in e97e1bc2. The reactivate `added` event now hydrates `subscription.room` via `GetRoomsMeta` + `applyRoomInfo` (the same mapper `subscription.list` uses), so it matches the `added`-event contract; the reactivation test asserts the event carries the room.

julianshen:
**No actionable comments.** ✅

### Disputed

- `user-service/service/apps.go:36` (`distributed-system`) — The read, update, metadata lookup, and publish are not ordered as one transition. Starting unsubscribed, reactivation can read false and commit true, then pause in GetRoomsMeta; an unsubscribe then reads true, commits false, and publishes removed; when reactivation resumes, it publishes added with a later timestamp. The database is unsubscribed while devices display the room as active. Conversely, starting active, both requests can read true; unsubscribe can commit false, subscribe can commit true but suppress added based on its stale snapshot, and unsubscribe can then publish removed, leaving devices removed while the database is active. Best-effort delivery elsewhere does not prevent these successfully delivered stale events. Atomically assign a monotonically increasing subscription revision during each conditional transition and have consumers reject older revisions, or serialize delivery through an ordered transactional outbox.
- `user-service/service/apps.go:36` (`nats`) — For `chat.user.{account}.event.subscription.update`, request A can read an inactive subscription, commit reactivation, and block in `GetRoomsMeta`. Request B then reads the active state, commits unsubscribe, and publishes `removed`; when A resumes, it publishes `added`. The database remains unsubscribed while devices display the room as active. A corresponding stale-read interleaving can suppress `added` while the database ends active. Make transitions conditional and atomic, assign a monotonic subscription revision, and have consumers reject stale events, or deliver transitions through a transactionally written ordered outbox.
- `user-service/service/apps.go:92` (`nats`) — `SetAppSubscribed` commits before event preparation and publication. If `GetRoomsMeta` fails, this branch skips publishing; if the later core-NATS `Publish` fails, line 116 only logs the error. Both paths still return API success. A retry then reads `IsSubscribed == true`, so the `wasSubscribed` guard suppresses another `added` event and affected devices remain stale until an unrelated refetch. Persist an event intent atomically with the subscription transition and retry it after metadata or NATS recovery; prefetching metadata before the write can separately eliminate the metadata-failure window.
- `user-service/service/apps.go:36` (`performance-and-concurrency`) — The read, update, and publish are unordered across requests. Starting active, unsubscribe and subscribe can both read `IsSubscribed=true`; unsubscribe writes false, subscribe writes true but suppresses `added` from its stale snapshot, then unsubscribe publishes `removed`, leaving the database active while devices hide the room. Starting inactive, reactivation can write true and pause in `GetRoomsMeta`; unsubscribe then writes false and publishes `removed`, after which reactivation publishes `added`, leaving devices active while the database is inactive. The shared best-effort delivery contract does not address successfully delivered but reordered stale events, and the metadata call widens that window. Make transitions conditional/atomic, assign a monotonic revision with each committed update, and have consumers reject stale revisions or publish through a per-subscription ordered outbox.
- `user-service/service/apps.go:36` (`tgd-review`) — The separate read, write, and publish steps remain unordered. Starting inactive, reactivation can write `true` and pause in `GetRoomsMeta`; an unsubscribe can then write `false` and publish `removed`, after which reactivation publishes `added`, leaving clients subscribed while the database is not. Starting active, both requests can read `true`, causing reactivation to suppress `added` while a delayed unsubscribe publishes `removed` even if the final database state is active. Existing emitters having the same weakness does not remove it from this newly added path. Use an atomic conditional transition with a monotonically increasing revision that consumers can reject when stale, or persist events transactionally and deliver them in transition order.

### Related work

- [Issue #370](https://github.com/hmchangw/newchat/issues/370) — \[Fix\] setAppSubscription flips isSubscribed without emitting subscription.update — app list doesn't sync across devices (open)

### Local review memory

- `user-service/service/apps.go:36` — @chenjr0719: Thanks — addressed the error-path sibling finding in e9b7c656 : on a room-metadata lookup failure the reactivate path now skips the publish (logs it) instead of emitting a Room -less event, with a test. On the concurrent-toggle race: thi… ([thread](https://github.com/hmchangw/newchat/pull/371#discussion_r3850693633))
- `user-service/service/apps.go:116` — @chenjr0719: This is the documented behavior of the shared subscription.update path: room-service’s own publishSubscriptionUpdate states it is best-effort — a publish failure is logged, not returned, because the DB write is the source of truth and… ([thread](https://github.com/hmchangw/newchat/pull/371#discussion_r3850693638))

<details>

* `docs/client-api.md`
* `docs/client-api/events.md`
* `user-service/service/apps.go`
* `user-service/service/apps_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

user-service/service/apps.go
```diff
@@ -27,28 +32,87 @@ func (s *UserService) SetAppSubscription(c *natsrouter.Context, req models.SetAp
 	}
 	botName := app.Assistant.Name
 
+	// Both branches need this: the removed event carries the room id, and each keys off existence.
+	existing, err := s.subs.GetAppSubscription(c, account, botName)
```
julianshen, line 36: _🎯 concurrency_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`, `mongodb`, `nats`, `performance-and-concurrency`, `silent-failure`, `tgd-review` (6 rules)_

**Concurrent toggles can leave devices with the opposite subscription state**

The read, update, and publish are not serialized. Starting active, concurrent unsubscribe and subscribe calls can both read `IsSubscribed=true`; unsubscribe writes false, subscribe writes true but suppresses its `added` event from the stale snapshot, then unsubscribe publishes `removed`. The database ends active while other devices drop the room. Starting inactive, reactivation can write true and block in `GetRoomsMeta`; an unsubscribe then writes false and publishes `removed`, after which reactivation publishes `added`, leaving devices active while the database is inactive. Make each account/bot transition atomic and emit events in committed transition order, such as a conditional update returning a monotonic revision combined with an ordered outbox or consumer-side revision rejection.

<details>
<summary>Also reported by 6 other rules</summary>

- **`distributed-system`**:103: Starting unsubscribed, request A writes `true` and then pauses before publishing, for example in `GetRoomsMeta`. Request B subsequently reads the active subscription, writes `false`, and publishes `removed`. When A resumes, it publishes `added`, leaving the database unsubscribed while other devices process `removed` then `added` and display the botDM as active. The timestamp cannot fence this because A generates it after the delay, making the stale event appear newer. Include a monotonically increasing revision assigned atomically with each subscription update and have consumers reject lower revisions, or serialize durable outbox delivery by subscription revision.
- **`mongodb`**:35: The subscription document is read, updated in a separate MongoDB operation, and then published to NATS without a shared ordering mechanism. For example, an unsubscribe can update the document to false, a reactivate can then read false, update it to true, and publish `added`, after which the delayed unsubscribe publishes `removed`. MongoDB's final state is subscribed, but other devices end in the removed state. Concurrent requests can also both observe the same old value and emit duplicate transition events. Make the state change a conditional `findOneAndUpdate` that returns the committed document and atomically increments a per-subscription revision, then include that revision in events so consumers discard stale deliveries; alternatively, persist an ordered outbox entry in the same transaction and publish through a per-subscription ordered relay.
- **`distributed-system`**:90: The reactivation has already been committed when `GetRoomsMeta` runs. If room-service is unavailable or returns no metadata, the error is ignored and the code publishes `added` with the unhydrated `Room` that the comment says is not persisted. Consumers expecting the added-event contract cannot reconstruct the botDM, while the caller receives success; retrying also emits nothing because the subscription is now active. Fetch and validate the required metadata before changing the subscription, or persist a recoverable outbox job that retries enrichment and publication.
- **`silent-failure`**:90: If GetRoomsMeta fails or returns no matching room, this condition silently skips hydration and still publishes an "added" event. The subscription's non-persisted Room therefore remains nil, contradicting the required event shape asserted by the new test, and another device may fail to restore the botDM despite receiving an apparently successful transition. Preserve the documented best-effort request behavior, but log the metadata failure and skip publishing rather than emitting a partial event.

<details>

```text
	infos, err := s.rooms.GetRoomsMeta(c, sub.SiteID, []string{sub.RoomID})
	if err != nil {
		slog.WarnContext(c, "hydrate app subscription.update event", "error", err, "account", account, "request_id", natsutil.RequestIDFromContext(c))
		return
	}
	if len(infos) == 0 || !infos[0].Found {
		slog.WarnContext(c, "room metadata not found for app subscription.update event", "account", account, "room_id", sub.RoomID, "request_id", natsutil.RequestIDFromContext(c))
		return
	}
	applyRoomInfo(&subCopy, &infos[0])
```

</details>

- **`tgd-review`**:90: When GetRoomsMeta returns an error or no results, execution still publishes the `added` event with `Subscription.Room` unset. The surrounding comment and new test establish that the added-event contract requires a hydrated room so clients can re-add the botDM; publishing the incomplete payload can leave other devices stale. Handle lookup failure explicitly—at minimum log and skip this best-effort publication—or construct a valid fallback, and add an error-path test.
- **`nats`**:90: After the subscription is reactivated, a GetRoomsMeta error, empty response, or Found=false result leaves subCopy.Room unset, but the code still publishes the subscription.update event. This violates the added-event shape asserted by the new test and can cause other devices to reject or mishandle the reactivation. Since delivery is explicitly best-effort, minimally log and skip publishing rather than send a malformed event; also cover the failure path in a test.

<details>

```text
	infos, err := s.rooms.GetRoomsMeta(c, sub.SiteID, []string{sub.RoomID})
	if err != nil || len(infos) == 0 || !infos[0].Found {
		slog.WarnContext(c, "hydrate app subscription.update event", "error", err, "account", account, "room_id", sub.RoomID, "request_id", natsutil.RequestIDFromContext(c))
		return
	}
	applyRoomInfo(&subCopy, &infos[0])
```

</details>


</details>

<details>

```
In `user-service/service/apps.go` around line 36: The read, update, and publish are not serialized. Starting active, concurrent unsubscribe and subscribe calls can both read `IsSubscribed=true`; unsubscribe writes false, subscribe writes true but suppresses its `added` event from the stale snapshot, then unsubscribe publishes `removed`. The database ends active while other devices drop the room. Starting inactive, reactivation can write true and block in `GetRoomsMeta`; an unsubscribe then writes false and publishes `removed`, after which reactivation publishes `added`, leaving devices active while the database is inactive. Make each account/bot transition atomic and emit events in committed transition order, such as a conditional update returning a monotonic revision combined with an ordered outbox or consumer-side revision rejection.

```

</details>

user-service/service/apps.go
```diff
@@ -27,28 +32,87 @@ func (s *UserService) SetAppSubscription(c *natsrouter.Context, req models.SetAp
...
+		AppInfo:      &model.CounterpartAppInfo{ID: app.ID, Name: app.Name, AssistantName: app.Assistant.Name},
+		Timestamp:    time.Now().UTC().UnixMilli(),
+	}
+	data, _ := json.Marshal(evt) // primitives only; cannot fail
+	s.publishSubscriptionUpdate(c, account, data)
+}
+
+// Best-effort core-NATS fan-out of a botDM subscription.update to the user's own subject.
+func (s *UserService) publishSubscriptionUpdate(c *natsrouter.Context, account string, data []byte) {
+	if err := s.clientPub.Publish(c, subject.SubscriptionUpdate(account), data); err != nil {
```
julianshen, line 116: _🏷️ delivery_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`_

**A failed publish permanently loses a successful reactivation.**

The database transition completes before this best-effort publish, and a NATS outage, cancellation, or process failure causes the error to be logged while the API still returns success. Other devices therefore remain unsubscribed. A later retry reads `IsSubscribed == true`, so the `wasSubscribed` guard suppresses the replacement event and cannot repair the loss. Persist the event atomically with the subscription update in an outbox, then publish with acknowledgement and retry until delivery succeeds.

<details>

```
In `user-service/service/apps.go` around line 111: The database transition completes before this best-effort publish, and a NATS outage, cancellation, or process failure causes the error to be logged while the API still returns success. Other devices therefore remain unsubscribed. A later retry reads `IsSubscribed == true`, so the `wasSubscribed` guard suppresses the replacement event and cannot repair the loss. Persist the event atomically with the subscription update in an outbox, then publish with acknowledgement and retry until delivery succeeds.

```

</details>

julianshen (commented):
tGD inline review

user-service/service/apps.go
chenjr0719: Thanks — addressed the error-path sibling finding in `e9b7c656`: on a room-metadata lookup failure the reactivate path now skips the publish (logs it) instead of emitting a `Room`-less event, with a test.

On the concurrent-toggle race: this describes the existing best-effort `subscription.update` contract shared by every emitter, not something this PR introduces. The sibling toggle events go read→write→best-effort-local-publish with no atomicity or revision — see `room-service/handler.go` `mute_toggled`/`favorite_toggled` — and there is no revision field on the subscription anywhere. Making subscription.update atomically-ordered (a per-subscription revision consumers reject stale deliveries against, or an ordered outbox) is a worthwhile hardening, but a system-wide change across all emitters; scoping it to only this one path would diverge from the rest. Proposing that as a separate change rather than folding it into this targeted fix.

user-service/service/apps.go
chenjr0719: This is the documented behavior of the shared `subscription.update` path: room-service’s own `publishSubscriptionUpdate` states it is best-effort — a publish failure is logged, not returned, because the DB write is the source of truth and clients reconcile on next refetch. This handler follows that same contract. A durable outbox with acked retry would be the right hardening for that contract, but applied uniformly to every `subscription.update` emitter, not just this path — so I’d propose it separately rather than in this bug fix.

mliu33 (approved):
Awesome! Thanks

PR #372, opened 2026-08-25
fix(upload-service): stream file uploads instead of buffering them in memory

nalydadad:
## Summary

upload-service buffered every upload fully in memory across three independent layers, so a handful of concurrent large uploads could OOM the pod. This PR makes the upload path stream end-to-end: peak heap for a 100 MiB upload drops from **~341 MiB to ~6 MiB**, three concurrent 40 MiB files add ~0, and a 2 GiB body streams at ~5 MiB peak.

Each buffering layer was found by measurement (pprof / heap sampling), not inference:

1. **pkg/drive (resty)** — resty materializes any `io.Reader` body in memory to make it replayable (`handleMultipart` pooled buffer, then `getBodyCopy` → `io.ReadAll`). The bulk upload now builds a streamed multipart body (`io.MultiReader` chain: in-memory envelope parts + `io.LimitReader` over each file) with an exact precomputed `Content-Length`, sent through the underlying `*http.Client` — still constructed via `restyutil`, so the shared transport, OTel instrumentation, and timeout are preserved. The wire format was verified part-by-part against what resty produced (part order, boundary, header order, sniffed Content-Type, filename escaping); the new escaping additionally covers CR/LF in filenames, closing a header-injection gap resty's escaper leaves open (now pinned by a dedicated test). `url.PathEscape(groupID)` replicates what resty's `SetPathParam` did internally.
2. **Gin** — the default `MaxMultipartMemory` (32 MiB) kept large chunks resident (~97 MiB peak per in-flight upload). Now a 1 MiB const: parts spill to the OS temp dir instead (a deliberate RAM→disk trade; ephemeral storage must fit the concurrent-upload total — documented at the const).
3. **Image preview — removed** — `imagePreview` fully decoded every uploaded image into pixels (width × height × 4 bytes, up to ~200 MiB at the old 50 MP cap) to build a 32×32 blurred JPEG. The preview is unused by the frontend, so it is no longer generated: the upload response keeps `imageDimensions` (now read from the image *header* via `image.DecodeConfig` — a few hundred bytes, never the pixels) and simply omits `imagePreview`. The 50 MP decompression-bomb cap goes with it (it existed solely to bound the decode). `golang.org/x/image` is no longer needed and leaves `go.mod`.

## API change (client-facing)

- `POST /api/v1/file/rooms/:roomId/upload/file` no longer returns `imagePreview` in the attachment. `imageDimensions` is still returned.
- `Attachment.ImagePreview` stays on the model: messages stored before this change keep round-tripping their preview through history, search, and delivery (round-trip pinned by existing model tests). `docs/client-api.md` updated accordingly.

## Testing

- Heap-ceiling tests assert the streaming for real (ticker-sampled `runtime.ReadMemStats`): a 3×40 MiB bulk upload must stay under 32 MiB in `pkg/drive`, and a 100 MiB upload through the real Gin router + routes + handler must stay under 16 MiB end-to-end.
- Exact-Content-Length (no chunked fallback), sniff-boundary file sizes (0/1/511/512/513 bytes), unreadable/unmeasurable files (fail before any byte is sent), dimension rewind semantics (every exit leaves the reader at the start; non-image MIME never touches it), and transport/JSON-decode error paths.
- Header-injection pin: a hostile filename embedding `"` + CR/LF must arrive percent-encoded, must not override the sniffed part Content-Type, and must not split the file part — verified the test fails when the CR escaping is removed.
- Blank-import guard: `_ "image/jpeg"` / `_ "image/png"` in `dimensions.go` are what teach `image.DecodeConfig` those formats. The dimension tests embed real fixtures via `go:embed` (no test file imports the format packages), so deleting either blank import fails the suite instead of silently costing every upload of that format its dimensions.
- `pkg/drive` coverage 90.4%; the shared fixtures (`PeakHeapDuring`, `ZeroReader`) are hoisted to `pkg/testutil` with their own tests.
- `make lint`: 0 issues. `make sast` (gosec + govulncheck + semgrep): clean.
- Full `make test` on the author's machine fails only on the pre-existing `tools/loadgen` FMA-contraction failure specific to arm64 — unrelated to this change and passing on amd64.

## Deliberately out of scope

- Drive upload client keeps its existing 30s timeout (vs the 5m download client) — with true streaming, slow-network large uploads will hit it; worth revisiting separately.
- `UploadGroupImages` still takes no `context.Context` (pre-existing).
- 2 GiB uploads remain gated by `FILE_UPLOAD_MAX_FILE_SIZE` (100 MiB default) and the 30s server `ReadTimeout`.
- GIF/WebP uploads get no `imageDimensions` (only JPEG/PNG decoders are registered — pre-existing behavior, unchanged).

## Summary by CodeRabbit

* **Performance**
  * Large file uploads now stream without loading entire files into memory.
  * Multipart uploads use accurate content lengths and temporary storage for larger parts.

* **Changes**
  * Newly uploaded images no longer include generated blurred previews; previews remain available for older attachments.
  * Image dimensions continue to be detected without loading full image data.

* **Reliability**
  * Improved upload error reporting and protection against unsafe filenames.
  * Added safeguards for unreadable files, malformed responses, and rewind failures.

* **Testing**
  * Expanded coverage for streaming, memory usage, content lengths, boundaries, and error handling.


pkg/drive/uploader.go
```diff
@@ -21,21 +23,29 @@ type MultipartFile struct {
...
+// OOM this path exists to avoid. Building it through restyutil anyway preserves
+// the shared transport, OTel instrumentation and timeout — but not resty's
+// request/response log hooks; upload failures are logged once at the caller's
+// errhttp boundary instead.
 func NewClient(cfg *Config) *Client {
 	// #nosec G402 -- internal Drive over a private network; TLS verification is intentionally skipped per deployment.
 	insecure := &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true, MinVersion: tls.VersionTLS13}}
 	return &Client{
-		uploadClient:   restyutil.New(cfg.URL, restyutil.WithTransport(insecure)),
+		uploadHTTP:     restyutil.New(cfg.URL, restyutil.WithTransport(insecure)).GetClient(),
```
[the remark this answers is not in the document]
nalydadad: Verified against the resty v2.17.2 source before rewording. In `createHTTPRequest`, an `io.Reader` body goes through `http.NewRequest`, which sets `GetBody` only for `*bytes.Buffer`/`*bytes.Reader`/`*strings.Reader`; every other reader hits `getBodyCopy` → `io.ReadAll`. So the buffering claim holds for any genuinely streaming body — the three escapes are already fully in memory — but "always" was still imprecise as written.

92b49716 narrows the doc to "materializes any io.Reader body it cannot natively replay" and names the direct `*http.Client` an explicit exception to the resty-for-outbound-HTTP guideline. Kept the documented-exception option over a resty streaming path: `RawRequest`/`GetBody` cannot be pre-set from outside the library, so no such path exists in v2 without buffering.

nalydadad:
Review feedback addressed in 92b49716:

- **Status error snippet** (`pkg/drive/uploader.go`): the bulk-upload error now carries a 512-byte bounded prefix of the Drive response body, with a test asserting it surfaces. Bounded, so no full body ever reaches a log.
- **`ZeroReader.ReadAt` contract** (`pkg/testutil/memtest.go`): a short read now returns `io.EOF` alongside the count, matching `bytes.Reader`; negative offsets error instead of ErrUnexpectedEOF, which is the stdlib convention for out-of-range offsets.
- **testify + `t.Logf`** (both streaming tests): converted. One deliberate deviation from the suggestion: inside the `httptest` handlers the assertions use `assert`, not `require` — the handler runs off the test goroutine, where `FailNow` is not allowed.

Skipped:

- **Ephemeral-storage budget/alerting** (`upload-service/main.go`): valid operational point, but pod resource requests, `TMPDIR`, and alerts live in the deployment configuration, which is not in this repo. The disk-for-RAM trade and sizing rule (`MAX_ATTACHMENTS × FILE_UPLOAD_MAX_FILE_SIZE` per in-flight request) are documented in the `maxMultipartMemory` comment and the PR description for whoever owns that config.

julianshen:

### Local review memory

- `pkg/drive/uploader_stream_test.go:57` — @coderabbitai: _🩺 Stability & Availability_ _🟡 Minor_ _⚡ Quick win_ **Synchronize values returned by httptest handlers.** The handlers run in separate goroutines. Returning from UploadGroupImages does not provide a Go memory synchronization poi… ([thread](https://github.com/hmchangw/newchat/pull/372#discussion_r3851868234))

<details>
<summary>📒 Files reviewed (23)</summary>

* `docs/client-api.md`
* `go.mod`
* `go.sum`
* `pkg/drive/multipart.go`
* `pkg/drive/uploader.go`
* `pkg/drive/uploader_stream_test.go`
* `pkg/drive/uploader_test.go`
* `pkg/model/cassandra/attachment.go`
* `pkg/testutil/memtest.go`
* `pkg/testutil/memtest_test.go`
* `upload-service/attachment.go`
* `upload-service/attachment_test.go`
* `upload-service/dimensions.go`
* `upload-service/dimensions_test.go`
* `upload-service/handler.go`
* `upload-service/handler_setcookie_test.go`
* `upload-service/handler_test.go`
* `upload-service/main.go`
* `upload-service/preview.go`
* `upload-service/preview_test.go`
* `upload-service/testdata/32x32.jpg`
* `upload-service/testdata/64x48.png`
* `upload-service/upload_stream_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

pkg/drive/uploader.go
```diff
@@ -58,30 +70,52 @@ func (c *Client) GetBaseURLFromRoomOrigin(origin string) string {
...
 	}
-	if resp.IsError() {
-		return nil, fmt.Errorf("drive bulk upload returned status %d", resp.StatusCode())
+	defer resp.Body.Close()
+
+	if resp.StatusCode >= http.StatusBadRequest {
+		// Bounded snippet so a Drive rejection is diagnosable without resty's
+		// response logging, and without ever logging a full body. A read error
+		// here is irrelevant next to the status and intentionally dropped.
+		snippet, _ := io.ReadAll(io.LimitReader(resp.Body, 512))
```
julianshen, line 113: _⚡ performance_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`, `performance-and-concurrency` (2 rules)_

**Large error bodies prevent HTTP connection reuse.**

When Drive repeatedly returns an error body larger than 512 bytes, LimitReader stops before EOF and the deferred Close discards the connection instead of returning it to the transport pool. During an outage or rejection spike, every failed upload therefore opens a new connection and incurs another TLS handshake, amplifying latency and socket load. Drain the remainder to io.Discard after capturing the bounded diagnostic snippet; the HTTP client timeout still bounds the read.

<details>

> ‼️ **IMPORTANT**

```suggestion
		snippet, _ := io.ReadAll(io.LimitReader(resp.Body, 512))
		_, _ = io.Copy(io.Discard, resp.Body)
```

</details>

<details>

- **`distributed-system`**:107: When Drive or a proxy returns an error body larger than 512 bytes, only the snippet is consumed before Close. With HTTP/1.x the transport cannot reuse that connection, so repeated downstream rejections create a new TCP/TLS connection per upload, amplifying an outage through handshake latency and socket churn. Drain the body to EOF in constant memory on every return path while retaining only the bounded diagnostic snippet.

<details>

```text
	defer func() {
		_, _ = io.Copy(io.Discard, resp.Body)
		_ = resp.Body.Close()
	}()
```

</details>


</details>

<details>

```
In `pkg/drive/uploader.go` around line 113: When Drive repeatedly returns an error body larger than 512 bytes, LimitReader stops before EOF and the deferred Close discards the connection instead of returning it to the transport pool. During an outage or rejection spike, every failed upload therefore opens a new connection and incurs another TLS handshake, amplifying latency and socket load. Drain the remainder to io.Discard after capturing the bounded diagnostic snippet; the HTTP client timeout still bounds the read.

```

</details>

upload-service/main.go
```diff
@@ -152,6 +160,7 @@ func run() error {
 
 	gin.SetMode(gin.ReleaseMode)
 	r := gin.New()
+	r.MaxMultipartMemory = maxMultipartMemory
```
julianshen, line 163: _🏷️ backpressure_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`_

**Spooling uploads has no aggregate disk backpressure.**

The 1 MiB threshold moves each larger in-flight upload to temporary disk, but it does not bound total spool usage or concurrent uploads. If Drive becomes slow, each handler retains its complete temp file while the outbound upload waits; concurrent requests can therefore fill ephemeral storage, after which otherwise valid uploads fail across the instance. Add admission control before multipart parsing, with a concurrency or byte budget derived from available spool space and maximum request size, and reject excess work with 429/503.

<details>

```
In `upload-service/main.go` around line 163: The 1 MiB threshold moves each larger in-flight upload to temporary disk, but it does not bound total spool usage or concurrent uploads. If Drive becomes slow, each handler retains its complete temp file while the outbound upload waits; concurrent requests can therefore fill ephemeral storage, after which otherwise valid uploads fail across the instance. Add admission control before multipart parsing, with a concurrency or byte budget derived from available spool space and maximum request size, and reject excess work with 429/503.

```

</details>

pkg/drive/uploader.go
```diff
@@ -58,30 +70,52 @@ func (c *Client) GetBaseURLFromRoomOrigin(origin string) string {
...
+		return nil, fmt.Errorf("drive bulk upload returned status %d: %s", resp.StatusCode, snippet)
+	}
+	var result []UploadGroupImageResponse
+	if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
+		return nil, fmt.Errorf("decode drive bulk upload response: %w", err)
```
julianshen, line 118: _🏷️ error-handling_ | _🟠 Warning_ | _⚡ Quick fix_ | _`silent-failure`_

**Response transfer failures after the first JSON value are ignored.**

json.Decoder.Decode returns once one complete value is parsed; it does not require the response body to reach EOF. If Drive or a proxy sends valid result JSON but the declared response is truncated or the connection resets afterward, Decode can return nil while the body retains the transfer error, and this method reports upload success. Consume the decoder through EOF and propagate any read error or unexpected trailing data before returning the result.

<details>

> ‼️ **IMPORTANT**

```suggestion
	var result []UploadGroupImageResponse
	decoder := json.NewDecoder(resp.Body)
	if err := decoder.Decode(&result); err != nil {
		return nil, fmt.Errorf("decode drive bulk upload response: %w", err)
	}
	var trailing json.RawMessage
	if err := decoder.Decode(&trailing); err == nil {
		return nil, fmt.Errorf("decode drive bulk upload response: unexpected trailing JSON")
	} else if err != io.EOF {
		return nil, fmt.Errorf("finish reading drive bulk upload response: %w", err)
	}
	return result, nil
```

</details>

<details>

```
In `pkg/drive/uploader.go` around line 114: json.Decoder.Decode returns once one complete value is parsed; it does not require the response body to reach EOF. If Drive or a proxy sends valid result JSON but the declared response is truncated or the connection resets afterward, Decode can return nil while the body retains the transfer error, and this method reports upload success. Consume the decoder through EOF and propagate any read error or unexpected trailing data before returning the result.

```

</details>

upload-service/upload_stream_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/hmchangw/chat/pkg/testutil"
+)
+
+// uploadStack stands up the real router against a stub Drive backend that drains
+// whatever it is sent, and returns the live upload URL.
+func uploadStack(t *testing.T) string {
+	t.Helper()
+	driveSrv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, _ = io.Copy(io.Discard, r.Body)
+		w.Header().Set("Content-Type", "application/json")
```
julianshen, line 27: _🧪 tests_ | _🟠 Warning_ | _⚡ Quick fix_ | _`silent-failure`_

**The fake Drive acknowledges truncated request bodies.**

The stub discards both io.Copy's byte count and error and always returns a successful upload result. A regression that sends an empty or truncated but otherwise readable multipart body can therefore still produce the asserted 200 status, so this streaming test does not verify its stated requirement that the complete file reaches Drive. Check the copy error and either parse/count the uploaded file in the stub or return the count to the test and compare it with fileSize before emitting success.

<details>

```
In `upload-service/upload_stream_test.go` around line 27: The stub discards both io.Copy's byte count and error and always returns a successful upload result. A regression that sends an empty or truncated but otherwise readable multipart body can therefore still produce the asserted 200 status, so this streaming test does not verify its stated requirement that the complete file reaches Drive. Check the copy error and either parse/count the uploaded file in the stub or return the count to the test and compare it with fileSize before emitting success.

```

</details>

pkg/testutil/memtest.go
```diff
@@ -0,0 +1,96 @@
...
+	return z.pos, nil
+}
+
+func (z *ZeroReader) Close() error { return nil }
+
+// PeakHeapDuring reports the maximum HeapAlloc observed while fn runs. It
+// samples on a ticker because runtime.ReadMemStats stops the world; a busy
+// loop would starve the very work being measured.
+func PeakHeapDuring(fn func()) uint64 {
+	runtime.GC()
```
julianshen, line 72: _🧪 tests_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**Memory assertions include unrelated baseline heap usage.**

PeakHeapDuring returns the process's absolute HeapAlloc after GC, while the new streaming tests compare it with fixed 16 MiB and 32 MiB limits as though it represented allocation growth caused by the upload. Live package state, runtime/version differences, or allocations retained by earlier tests can therefore fail these tests even when the upload remains fully streamed. Capture HeapAlloc immediately after GC and report the maximum increase over that baseline (then adjust the helper's documentation and test messages accordingly).

<details>

```
In `pkg/testutil/memtest.go` around line 72: PeakHeapDuring returns the process's absolute HeapAlloc after GC, while the new streaming tests compare it with fixed 16 MiB and 32 MiB limits as though it represented allocation growth caused by the upload. Live package state, runtime/version differences, or allocations retained by earlier tests can therefore fail these tests even when the upload remains fully streamed. Capture HeapAlloc immediately after GC and report the maximum increase over that baseline (then adjust the helper's documentation and test messages accordingly).

```

</details>

julianshen (commented):
tGD inline review

julianshen (commented):
tGD inline review

pkg/drive/uploader.go
```diff
@@ -58,30 +70,52 @@ func (c *Client) GetBaseURLFromRoomOrigin(origin string) string {
...
+// upload size.
 func (c *Client) UploadGroupImages(userID, username, email, groupID, origin string, files []MultipartFile) ([]UploadGroupImageResponse, error) {
-	req := c.uploadClient.R().SetHeader("api-token", c.apiToken)
-	formData := map[string]string{
-		"userId":   userID,
-		"userName": username,
-		"email":    email,
+	fields := []formField{
+		{"userId", userID},
+		{"userName", username},
```
julianshen, line 78: _⚡ memory-exhaustion_ | _🟠 Warning_ | _⚡ Quick fix_ | _`oom`_

**Oversized filenames are retained in the streamed request envelope.**

An authenticated uploader can supply a multi-megabyte multipart filename, which Gin's multipart parser can accept within its metadata allowance even with MaxMultipartMemory set to 1 MiB. The filename is then written once as a form value and again, potentially expanded by escaping, as a part header. buildStreamedBody copies and retains those envelope bytes for the entire outbound upload. Concurrent uploads with oversized filenames, especially while Drive is slow, can therefore cause tens of MiB of retained heap per request, leading to GC thrashing or container OOM. Reject filenames above a small hard limit before assembling the envelope.

<details>

```
In `pkg/drive/uploader.go` around line 78: An authenticated uploader can supply a multi-megabyte multipart filename, which Gin's multipart parser can accept within its metadata allowance even with MaxMultipartMemory set to 1 MiB. The filename is then written once as a form value and again, potentially expanded by escaping, as a part header. buildStreamedBody copies and retains those envelope bytes for the entire outbound upload. Concurrent uploads with oversized filenames, especially while Drive is slow, can therefore cause tens of MiB of retained heap per request, leading to GC thrashing or container OOM. Reject filenames above a small hard limit before assembling the envelope.

```

</details>

pkg/drive/uploader.go
```diff
@@ -58,30 +70,52 @@ func (c *Client) GetBaseURLFromRoomOrigin(origin string) string {
...
+		// response logging, and without ever logging a full body. A read error
+		// here is irrelevant next to the status and intentionally dropped.
+		snippet, _ := io.ReadAll(io.LimitReader(resp.Body, 512))
+		return nil, fmt.Errorf("drive bulk upload returned status %d: %s", resp.StatusCode, snippet)
+	}
+	var result []UploadGroupImageResponse
+	if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
+		return nil, fmt.Errorf("decode drive bulk upload response: %w", err)
 	}
 	return result, nil
```
julianshen, line 120: _🏷️ error-handling_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**Response transfer failures after valid JSON are ignored.**

If Drive sends one valid JSON value and then the connection fails before the declared response ends, the first Decode can return nil without reading the transfer error. The upload is then reported successful despite an incomplete response. Decode through EOF and reject transfer errors or trailing JSON.

<details>

> ‼️ **IMPORTANT**

```suggestion
	var result []UploadGroupImageResponse
	decoder := json.NewDecoder(resp.Body)
	if err := decoder.Decode(&result); err != nil {
		return nil, fmt.Errorf("decode drive bulk upload response: %w", err)
	}
	var trailing json.RawMessage
	if err := decoder.Decode(&trailing); err == nil {
		return nil, fmt.Errorf("decode drive bulk upload response: unexpected trailing JSON")
	} else if err != io.EOF {
		return nil, fmt.Errorf("finish reading drive bulk upload response: %w", err)
	}
	return result, nil
```

</details>

<details>

```
In `pkg/drive/uploader.go` around line 116: If Drive sends one valid JSON value and then the connection fails before the declared response ends, the first Decode can return nil without reading the transfer error. The upload is then reported successful despite an incomplete response. Decode through EOF and reject transfer errors or trailing JSON.

```

</details>

pkg/drive/uploader_stream_test.go
```diff
@@ -0,0 +1,239 @@
...
+	files := make([]MultipartFile, fileCount)
+	for i := range files {
+		files[i] = MultipartFile{File: &testutil.ZeroReader{N: fileSize}, Filename: fmt.Sprintf("big%d.bin", i)}
+	}
+
+	var err error
+	peak := testutil.PeakHeapDuring(func() {
+		_, err = c.UploadGroupImages("alice", "Alice", "a@x.com", "r1", "site-x", files)
+	})
+	require.NoError(t, err)
```
julianshen, line 51: _🧪 tests_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**Handler results are shared without synchronization.**

The httptest handlers write variables that the test goroutine later reads, but completing an HTTP exchange is not a documented Go memory synchronization point. This affects got here, declared/received/te around lines 80-82, got around lines 176-177, and raw/boundary around lines 212-238, and can be reported by the race detector. Transfer each handler result through a buffered channel or protect it with a mutex.

<details>

```
In `pkg/drive/uploader_stream_test.go` around line 51: The httptest handlers write variables that the test goroutine later reads, but completing an HTTP exchange is not a documented Go memory synchronization point. This affects got here, declared/received/te around lines 80-82, got around lines 176-177, and raw/boundary around lines 212-238, and can be reported by the race detector. Transfer each handler result through a buffered channel or protect it with a mutex.

```

</details>

pkg/drive/multipart.go
```diff
@@ -0,0 +1,115 @@
...
+	return size, nil
+}
+
+// partHeader builds a file part's MIME header in mime/multipart's own format.
+// Concatenated rather than %q-formatted: the values are already escaped by
+// quoteEscaper, and Go quoting would escape them a second time.
+func partHeader(field, filename, contentType string) textproto.MIMEHeader {
+	h := make(textproto.MIMEHeader)
+	h.Set("Content-Disposition", `form-data; name="`+quoteEscaper.Replace(field)+
+		`"; filename="`+quoteEscaper.Replace(filename)+`"`)
```
mliu33, line 112: Don't we need to url encode the filename here ? For example, filename could be in Chinese characters.

mliu33 (approved):
Excellent work, thanks! Just one comment

pkg/drive/multipart.go
nalydadad: Verified against the wire format and the pre-#372 resty path — the filename must stay raw UTF-8 here, not URL-encoded:

- RFC 7578 §4.2 forbids percent/RFC 5987 encoding of the `multipart/form-data` filename; raw UTF-8 inside `filename="…"` is the standard form (what browsers send too), and Go's server-side parser returns it as-is.
- resty v2.17.2 (`util.go` `createMultipartHeader`) escaped only `\` and `"` and passed every other byte through, so Chinese filenames already crossed the wire as raw UTF-8 before #372 — `quoteEscaper` preserves that byte-for-byte and only adds CR/LF hardening.
- URL-encoding would regress behavior: Drive would receive and store the literal percent-encoded string (`%E4%B8%AD%E6%96%87….png`) as the name.

Pinned in [#380](https://github.com/hmchangw/newchat/pull/380): `TestClient_UploadGroupImages_UnicodeFilename` asserts a Chinese filename arrives intact in both the part header and the `files[N].fileName` field, and fails if `url.PathEscape` is applied.

---

PR #373, opened 2026-08-25
Fix preview staleness on re-login and resync

hmchangw:
## Summary
Fixes a regression where cached previews from a previous session would persist in the sidebar after re-login, and ensures that encrypted message placeholders are not overwritten by unencrypted wire previews during resync.

## Key Changes

- **Refactored preview comparison logic** into a new `storedPreviewWins()` helper that consolidates two conditions:
  - Encrypted placeholders for the same message ID are preserved (Fix 3)
  - Stored previews newer than incoming wire previews are kept (Fix 7)

- **Changed BUCKETS_LOADED behavior** from fill-if-absent to always compare incoming previews:
  - Previously only seeded rooms with no prior preview, allowing stale cached entries to persist
  - Now compares all incoming previews against stored ones and replaces them unless the stored version wins
  - Prevents re-login from showing the previous session's messages in the sidebar

- **Added reference stability optimization** via `samePreview()` check:
  - Avoids unnecessary object mutations when a resync carries the same preview
  - Prevents sidebar re-renders on no-op syncs

- **Updated MESSAGE_RECEIVED** to use the consolidated `storedPreviewWins()` logic for consistency

- **Added comprehensive test coverage**:
  - Test for preserving live messages that arrive before BUCKETS_LOADED resolves
  - Test for replacing stale cached previews with fresher server ones
  - Test for keeping encrypted placeholders when the same message is resync'd
  - Test for reference stability when nothing changes
  - Integration test verifying the re-login scenario end-to-end

## Implementation Details

The fix addresses the hydration regression where `hydrateFromCache` replays BUCKETS_LOADED with the previous session's cached `previewMessage`. By the time the network bootstrap lands, every room already has a stale entry. The old fill-if-absent logic would skip updating these rooms, leaving the sidebar showing outdated messages.

The new approach compares timestamps (`createdAt`) to determine which preview is fresher, with special handling for encrypted messages to avoid flipping rows that the timeline deliberately refuses to render.

https://claude.ai/code/session_01DwBSL6i3tBedQAS8E6tAhb

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved room preview synchronization when loading rooms.
  * Stale cached previews are now replaced by newer messages.
  * Preserved newer live previews and encrypted message placeholders.
  * Ensured rooms without existing previews still display their latest available preview.
  * Prevented deleted messages from remaining pinned in the room list.
  * Restored room previews across sessions while removing invalid or outdated cached entries.
  * Improved preview handling when cached messages are deleted while offline.

### DIFF
--- a/chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx
+++ b/chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx
@@ -214,8 +214,15 @@ function hydrateFromCache(user: Nats['user']): RoomEventsState {
     subscriptions: cached.subscriptions,
     rooms,
   })
-  if (!cached.chatlist) return seeded as RoomEventsState
-  return roomEventsReducer(seeded, {
+  // Overlay the previews the last session ended on — the cached subscriptions
+  // only carry the previewMessage its bootstrap saw, so a message that arrived
+  // afterwards lives here.
+  const withPreviews = roomEventsReducer(seeded, {
+    type: 'PREVIEWS_HYDRATED',
+    previews: cached.previews,
+  })
+  if (!cached.chatlist) return withPreviews as RoomEventsState
+  return roomEventsReducer(withPreviews, {
     type: 'CHATLIST_LOADED',
     chatlist: cached.chatlist,
   }) as RoomEventsState
@@ -271,22 +278,23 @@ export function RoomEventsProvider({ children }: { children: ReactNode }) {
     seedKeys,
   )
 
-  // Write-through to the browser cache. Keyed by reference on the five
-  // persisted slices: message traffic churns roomState / summaries /
-  // msgRecvSeq but never these, so this stays quiet during a busy session.
+  // Write-through to the browser cache, keyed by reference on the persisted
+  // slices. `previews` is the one that message traffic churns, and the trailing
+  // debounce is what keeps a busy room from writing per message: the timer
+  // restarts on each change, so a burst costs one write once it settles.
   //
   // The identity check is the teardown guard. `RESET` returns the
   // `initialState` OBJECT, so `=== initialState.subscriptions` is an exact
   // "we've been reset" test — without it, logout (and StrictMode's
   // double-mount) would persist the empty post-RESET state over a good cache.
-  const { subscriptions, favoriteIds, appIds, channelDmIds, chatlist } = state
+  const { subscriptions, favoriteIds, appIds, channelDmIds, chatlist, previews } = state
   useEffect(() => {
     if (subscriptions === initialState.subscriptions) return undefined
     const timer = setTimeout(() => {
-      saveSubscriptionCache(user, { subscriptions, favoriteIds, appIds, channelDmIds, chatlist })
+      saveSubscriptionCache(user, { subscriptions, favoriteIds, appIds, channelDmIds, chatlist, previews })
     }, CACHE_WRITE_DEBOUNCE_MS)
     return () => clearTimeout(timer)
-  }, [user, subscriptions, favoriteIds, appIds, channelDmIds, chatlist])
+  }, [user, subscriptions, favoriteIds, appIds, channelDmIds, chatlist, previews])
 
   const loadHistory = useCallback(
     async (roomId: string) => {
--- a/chat-frontend/src/context/RoomEventsContext/reducer.js
+++ b/chat-frontend/src/context/RoomEventsContext/reducer.js
@@ -210,6 +210,23 @@ function storedPreviewWins(cur, next) {
   return isOlderPreview(next.createdAt, cur.createdAt)
 }
 
+// Same question for a `subscription.list` row, which is the server's read-time
+// resolve rather than one event's view — so plain recency is the wrong test.
+// A delete never rolls `room.lastMsgAt` back, so a stored preview at or before
+// it is inside the window the row already accounts for and the row supersedes
+// it; only a message that POSTDATES the row (one that arrived live after it was
+// resolved) survives. Recency alone would pin a message deleted while this
+// client was away: the survivor the server returns is older, forever.
+// Falls back to the recency rule when either timestamp is missing.
+function rowPreviewWins(cur, next, rowLastMsgAt) {
+  if (!cur) return false
+  if (cur.encrypted && cur.messageId === next.messageId) return true
+  const curT = Date.parse(cur.createdAt ?? '')
+  const rowT = Date.parse(rowLastMsgAt ?? '')
+  if (Number.isNaN(curT) || Number.isNaN(rowT)) return isOlderPreview(next.createdAt, cur.createdAt)
+  return curT > rowT
+}
+
 /**
  * Apply the server's per-user subscription record onto a summary.
  *
@@ -363,11 +380,11 @@ export function roomEventsReducer(state, action) {
     }
     case 'BUCKETS_LOADED': {
       const subs = action.subscriptions ?? {}
-      // Take the server's preview unless what's stored wins on recency (a live
-      // message that landed before this fetch resolved) or is an encrypted
-      // placeholder for the same message — NOT a fill-if-absent seed: hydration
-      // replays this action with the previous session's cached previewMessage,
-      // so on a re-login every room already has a stale entry to replace.
+      // Take the server's preview unless what's stored postdates the row or is
+      // an encrypted placeholder for the same message (see rowPreviewWins) —
+      // NOT a fill-if-absent seed: hydration replays this action with the
+      // previous session's cached previewMessage, so on a re-login every room
+      // already has a stale entry to replace.
       // Shared by both paths: a degraded bootstrap still carries real previews
       // for the buckets it did reach.
       let previews = state.previews
@@ -377,7 +394,7 @@ export function roomEventsReducer(state, action) {
         const cur = previews[roomId]
         // samePreview keeps the map reference stable on a no-op resync, which
         // would otherwise re-render every row in the sidebar.
-        if (storedPreviewWins(cur, preview) || samePreview(cur, preview)) continue
+        if (rowPreviewWins(cur, preview, sub?.room?.lastMsgAt) || samePreview(cur, preview)) continue
         if (previews === state.previews) previews = { ...state.previews }
         previews[roomId] = preview
       }
--- a/chat-frontend/src/context/RoomEventsContext/reducer.js
+++ b/chat-frontend/src/context/RoomEventsContext/reducer.js
@@ -448,6 +448,21 @@ export function roomEventsReducer(state, action) {
         previews,
       }
     }
+    case 'PREVIEWS_HYDRATED': {
+      // The previous session's previews, overlaid on what BUCKETS_LOADED just
+      // seeded from the cached subscriptions. Live messages update previews but
+      // never the subscription's previewMessage, so either side can be the
+      // fresher one — the same take-if-newer rule picks the winner.
+      let previews = state.previews
+      for (const [roomId, preview] of Object.entries(action.previews ?? {})) {
+        if (!preview?.messageId) continue
+        const cur = previews[roomId]
+        if (storedPreviewWins(cur, preview) || samePreview(cur, preview)) continue
+        if (previews === state.previews) previews = { ...state.previews }
+        previews[roomId] = preview
+      }
+      return previews === state.previews ? state : { ...state, previews }
+    }
     case 'SUBSCRIPTION_UPSERTED': {
       // Upsert a single subscription record (live delta from
       // `subscription.update` events). Spreads the new fields on top
--- a/chat-frontend/src/context/RoomEventsContext/reducer.js
+++ b/chat-frontend/src/context/RoomEventsContext/reducer.js
@@ -210,23 +210,6 @@ function storedPreviewWins(cur, next) {
   return isOlderPreview(next.createdAt, cur.createdAt)
 }
 
-// Same question against a `subscription.list` row, which is the server's
-// read-time resolve rather than one event's view — so plain recency is the
-// wrong test. A delete never rolls `room.lastMsgAt` back, so a stored preview
-// at or before it is inside the window the row already accounts for and the
-// row supersedes it; only a stored preview that POSTDATES the row (a message
-// that arrived live after it was resolved) wins. Recency alone would pin a
-// message deleted while this client was away: the survivor the server returns
-// is older, forever. Falls back to recency when either timestamp is missing.
-function storedBeatsRow(cur, next, rowLastMsgAt) {
-  if (!cur) return false
-  if (cur.encrypted && cur.messageId === next.messageId) return true
-  const curT = Date.parse(cur.createdAt ?? '')
-  const rowT = Date.parse(rowLastMsgAt ?? '')
-  if (Number.isNaN(curT) || Number.isNaN(rowT)) return isOlderPreview(next.createdAt, cur.createdAt)
-  return curT > rowT
-}
-
 /**
  * Apply the server's per-user subscription record onto a summary.
  *
@@ -380,11 +363,11 @@ export function roomEventsReducer(state, action) {
     }
     case 'BUCKETS_LOADED': {
       const subs = action.subscriptions ?? {}
-      // Take the server's preview unless what's stored postdates the row or is
-      // an encrypted placeholder for the same message (see storedBeatsRow) —
-      // NOT a fill-if-absent seed: hydration replays this action with the
-      // previous session's cached previewMessage, so on a re-login every room
-      // already has a stale entry to replace.
+      // Take the server's preview unless what's stored wins on recency (a live
+      // message that landed before this fetch resolved) or is an encrypted
+      // placeholder for the same message — NOT a fill-if-absent seed: hydration
+      // replays this action with the previous session's cached previewMessage,
+      // so on a re-login every room already has a stale entry to replace.
       // Shared by both paths: a degraded bootstrap still carries real previews
       // for the buckets it did reach.
       let previews = state.previews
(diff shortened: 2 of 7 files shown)


julianshen:
**4 findings · 2 unique issues · 2 inline comments posted** — 🔴 2 blocking · 🟠 2 warning

<details>

* `chat-frontend/src/context/RoomEventsContext/RoomEventsContext.cache.test.jsx`
* `chat-frontend/src/lib/subscriptionCache.js`
* `chat-frontend/src/lib/subscriptionCache.test.js`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

chat-frontend/src/context/RoomEventsContext/reducer.js
```diff
@@ -354,17 +363,21 @@ export function roomEventsReducer(state, action) {
...
+      // replays this action with the previous session's cached previewMessage,
+      // so on a re-login every room already has a stale entry to replace.
       // Shared by both paths: a degraded bootstrap still carries real previews
       // for the buckets it did reach.
       let previews = state.previews
       for (const [roomId, sub] of Object.entries(subs)) {
-        if (previews[roomId]) continue
         const preview = previewFromWire(sub?.room?.previewMessage)
         if (!preview) continue
+        const cur = previews[roomId]
```
julianshen, line 377: _🏷️ consistency_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`_

**A fresh bootstrap can preserve a deleted cached preview.**

Concrete timeline: session A caches newest message m2 at 12:00; while the client is offline, m2 is deleted and the server's current preview rolls back to surviving message m1 at 11:00; on reload, hydration restores m2, then the fresh subscription bootstrap supplies m1. `storedPreviewWins` treats m1 as stale solely because its timestamp is earlier and skips the authoritative replacement. Because the delete broadcast was missed while offline, no later event is guaranteed to repair the row, so the sidebar can indefinitely display deleted content. Distinguish previous-session hydrated previews from current-session live updates: the network bootstrap should replace hydrated state, while only live events received after that bootstrap began should be protected from an older snapshot (for example with provenance or a bootstrap generation/sequence).

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 377: Concrete timeline: session A caches newest message m2 at 12:00; while the client is offline, m2 is deleted and the server's current preview rolls back to surviving message m1 at 11:00; on reload, hydration restores m2, then the fresh subscription bootstrap supplies m1. `storedPreviewWins` treats m1 as stale solely because its timestamp is earlier and skips the authoritative replacement. Because the delete broadcast was missed while offline, no later event is guaranteed to repair the row, so the sidebar can indefinitely display deleted content. Distinguish previous-session hydrated previews from current-session live updates: the network bootstrap should replace hydrated state, while only live events received after that bootstrap began should be protected from an older snapshot (for example with provenance or a bootstrap generation/sequence).

```

</details>

chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx
```diff
@@ -271,22 +278,23 @@ export function RoomEventsProvider({ children }: { children: ReactNode }) {
...
-  // persisted slices: message traffic churns roomState / summaries /
-  // msgRecvSeq but never these, so this stays quiet during a busy session.
+  // Write-through to the browser cache, keyed by reference on the persisted
+  // slices. `previews` is the one that message traffic churns, and the trailing
+  // debounce is what keeps a busy room from writing per message: the timer
+  // restarts on each change, so a burst costs one write once it settles.
   //
   // The identity check is the teardown guard. `RESET` returns the
   // `initialState` OBJECT, so `=== initialState.subscriptions` is an exact
   // "we've been reset" test — without it, logout (and StrictMode's
```
julianshen, line 288: _⚡ performance_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`, `performance-and-concurrency` (2 rules)_

**Preview traffic can starve or amplify whole-cache writes**

This pure trailing debounce restarts whenever `previews` changes. A stream with gaps below `CACHE_WRITE_DEBOUNCE_MS` continually cancels the timer, so closing or reloading the tab can leave the newest previews unpersisted. With gaps just above the delay, every message instead rebuilds, serializes, and synchronously writes the entire subscriptions/previews record to localStorage; large accounts can therefore incur repeated main-thread stalls, with quota failures adding a trim and second write. Replace this with bounded periodic throttling of the latest dirty snapshot (a minimum write interval plus a maximum wait), and flush pending state on `pagehide` where possible.

<details>

- **`distributed-system`**:291: Every preview update cancels and restarts this trailing timer. If messages arrive more frequently than `CACHE_WRITE_DEBOUNCE_MS`, no write occurs; logout or unmount then cancels the final timer, leaving the previous session cache unchanged. On the next login, especially when the bootstrap service is unavailable or times out, the warm sidebar shows stale previews despite this change's persistence contract. Use a debounce with a bounded maximum wait or periodic throttled checkpoints, and explicitly flush the latest coherent snapshot on logout/pagehide.

</details>

<details>

```
In `chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx` around line 288: This pure trailing debounce restarts whenever `previews` changes. A stream with gaps below `CACHE_WRITE_DEBOUNCE_MS` continually cancels the timer, so closing or reloading the tab can leave the newest previews unpersisted. With gaps just above the delay, every message instead rebuilds, serializes, and synchronously writes the entire subscriptions/previews record to localStorage; large accounts can therefore incur repeated main-thread stalls, with quota failures adding a trim and second write. Replace this with bounded periodic throttling of the latest dirty snapshot (a minimum write interval plus a maximum wait), and flush pending state on `pagehide` where possible.

```

</details>

chat-frontend/src/context/RoomEventsContext/reducer.js
```diff
@@ -354,17 +363,21 @@ export function roomEventsReducer(state, action) {
...
+        // samePreview keeps the map reference stable on a no-op resync, which
+        // would otherwise re-render every row in the sidebar.
```
julianshen, line 379: _🏷️ nats-ordering_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`nats`_

**A stale bootstrap reply can overwrite a newer live preview mutation**

The new timestamp-based replacement allows an in-flight `subscription.list` reply to undo a NATS event already reduced locally. For example, the request snapshots message m1, then an edit broadcast updates m1's preview text while retaining its original `createdAt`; when `BUCKETS_LOADED` arrives, `storedPreviewWins` returns false because the timestamps are equal and `samePreview` returns false because the text differs, so the stale text replaces the edit. A delete has the same problem: its live event can roll the preview back to an older survivor, after which the delayed bootstrap's newer timestamp resurrects the deleted message. Preserve previews touched by live events after the bootstrap request began, such as by recording a request-generation/event-sequence baseline and preventing that response from replacing rooms mutated since that baseline.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 379: The new timestamp-based replacement allows an in-flight `subscription.list` reply to undo a NATS event already reduced locally. For example, the request snapshots message m1, then an edit broadcast updates m1's preview text while retaining its original `createdAt`; when `BUCKETS_LOADED` arrives, `storedPreviewWins` returns false because the timestamps are equal and `samePreview` returns false because the text differs, so the stale text replaces the edit. A delete has the same problem: its live event can roll the preview back to an older survivor, after which the delayed bootstrap's newer timestamp resurrects the deleted message. Preserve previews touched by live events after the bootstrap request began, such as by recording a request-generation/event-sequence baseline and preventing that response from replacing rooms mutated since that baseline.

```

</details>

chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx
```diff
@@ -271,22 +278,23 @@ export function RoomEventsProvider({ children }: { children: ReactNode }) {
...
+  const { subscriptions, favoriteIds, appIds, channelDmIds, chatlist, previews } = state
   useEffect(() => {
     if (subscriptions === initialState.subscriptions) return undefined
     const timer = setTimeout(() => {
-      saveSubscriptionCache(user, { subscriptions, favoriteIds, appIds, channelDmIds, chatlist })
+      saveSubscriptionCache(user, { subscriptions, favoriteIds, appIds, channelDmIds, chatlist, previews })
     }, CACHE_WRITE_DEBOUNCE_MS)
     return () => clearTimeout(timer)
-  }, [user, subscriptions, favoriteIds, appIds, channelDmIds, chatlist])
+  }, [user, subscriptions, favoriteIds, appIds, channelDmIds, chatlist, previews])
```
julianshen, line 297: _🏷️ reliability_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`silent-failure`_

**Debounced preview writes can be cancelled without ever persisting.**

Every preview mutation restarts this trailing timer. Continuous traffic can therefore postpone the write indefinitely, and logout, unmount, or page closure cancels the final timer. A deletion followed by an immediate logout leaves the old preview in localStorage, so the next hydration can resurrect deleted message text; the new deletion regression test only covers a cache write that already completed. Add a maximum wait or synchronously flush the latest non-reset snapshot during teardown/pagehide.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx` around line 297: Every preview mutation restarts this trailing timer. Continuous traffic can therefore postpone the write indefinitely, and logout, unmount, or page closure cancels the final timer. A deletion followed by an immediate logout leaves the old preview in localStorage, so the next hydration can resurrect deleted message text; the new deletion regression test only covers a cache write that already completed. Add a maximum wait or synchronously flush the latest non-reset snapshot during teardown/pagehide.

```

</details>

chat-frontend/src/lib/subscriptionCache.js
```diff
@@ -48,6 +54,32 @@ function withoutPrivateKey(sub) {
...
 }
 
+/** Keep each room's sidebar snippet, minus any encrypted placeholder. That
+ *  placeholder records "this client could not decrypt it THEN"; after a reload
+ *  the room key is seeded from `subscription.list` and the message may well
+ *  render, so it must not outlive the session. */
+function persistablePreviews(previews) {
+  const out = {}
+  for (const [roomId, preview] of Object.entries(previews ?? {})) {
+    if (!preview?.messageId || preview.encrypted) continue
```
julianshen, line 59: _🎯 correctness_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`silent-failure`_

**Filtered encrypted previews are mistaken for deleted previews.**

This filter makes the persisted preview map partial, but PREVIEWS_HYDRATED treats every absent room as an authoritative deletion and replaces the previews seeded from cached subscriptions. After an encrypted preview is saved and the app reloads, hydration first seeds the subscription preview and then removes it because this function omitted the room; if bootstrap fails, the sidebar remains blank despite having a cached fallback. Preserve an explicit marker for filtered encrypted entries and retain the seeded preview for that marker, while representing actual deletions distinctly.

<details>

```
In `chat-frontend/src/lib/subscriptionCache.js` around line 64: This filter makes the persisted preview map partial, but PREVIEWS_HYDRATED treats every absent room as an authoritative deletion and replaces the previews seeded from cached subscriptions. After an encrypted preview is saved and the app reloads, hydration first seeds the subscription preview and then removes it because this function omitted the room; if bootstrap fails, the sidebar remains blank despite having a cached fallback. Preserve an explicit marker for filtered encrypted entries and retain the seeded preview for that marker, while representing actual deletions distinctly.

```

</details>

chat-frontend/src/context/RoomEventsContext/reducer.js
julianshen, line 379: _🎯 correctness_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`tgd-review`_

**Newer cached previews survive server-side deletion rollbacks**

If the cached newest message is deleted while the user is offline, the next bootstrap returns either an older surviving preview or no preview. `storedPreviewWins` rejects the older survivor, while the preceding `if (!preview) continue` also preserves the cached entry when the room is now empty, so the sidebar can display the deleted message indefinitely. Reconciliation must distinguish hydrated cache entries from previews received live after the bootstrap request began: the server snapshot should replace or clear hydrated entries even when its timestamp moves backward, while only post-request live updates should win the fetch race. Add coverage for both an older survivor and a null server preview.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 379: If the cached newest message is deleted while the user is offline, the next bootstrap returns either an older surviving preview or no preview. `storedPreviewWins` rejects the older survivor, while the preceding `if (!preview) continue` also preserves the cached entry when the room is now empty, so the sidebar can display the deleted message indefinitely. Reconciliation must distinguish hydrated cache entries from previews received live after the bootstrap request began: the server snapshot should replace or clear hydrated entries even when its timestamp moves backward, while only post-request live updates should win the fetch race. Add coverage for both an older survivor and a null server preview.

```

</details>

julianshen (commented):
tGD inline review

julianshen (commented):
tGD inline review

chat-frontend/src/context/RoomEventsContext/reducer.js
```diff
@@ -354,17 +380,21 @@ export function roomEventsReducer(state, action) {
...
+        if (rowPreviewWins(cur, preview, sub?.room?.lastMsgAt) || samePreview(cur, preview)) continue
```
julianshen: _🏷️ distributed-consistency_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`, `nats`, `tgd-review` (3 rules)_

**Bootstrap replies can resurrect deleted previews**

Concrete race: subscription.list snapshots newest message m2, then a live delete rolls the local preview back to m1. When the delayed BUCKETS_LOADED response arrives, rowPreviewWins sees m1.createdAt <= row.lastMsgAt and replaces it with the stale m2 snapshot. No later event is guaranteed, so deleted content remains visible. Similarly, when an offline deletion leaves the room empty, `if (!preview) continue` preserves cached m2. Track per-room live mutations or a bootstrap generation: authoritative responses should replace or clear hydrated entries, but must not overwrite rooms changed after the request began.

<details>

- **`tgd-review`**:392: When an offline deletion leaves a room empty, the fresh row has no preview and this `continue` preserves the deleted cached preview indefinitely. The same merge also lacks provenance for concurrent events: a live edit with the same `createdAt`, or a live deletion that rolls back to an older survivor, can be overwritten by a delayed bootstrap response. Track which previews came from hydration versus live events since the request began, explicitly clear authoritative empty rows, and protect rooms mutated after the request snapshot.
- **`nats`**:397: The `subscription.list` request can snapshot message m2, then a live `ROOM_PREVIEW_UPDATED` NATS delete event can roll the local preview back to survivor m1 before the reply arrives. When that delayed reply dispatches `BUCKETS_LOADED`, `rowPreviewWins` returns false because m1 predates the row's `lastMsgAt`, so the stale m2 snapshot replaces m1 and resurrects deleted content. An edit retaining the same `createdAt` is similarly overwritten. Record a request-generation or event-sequence baseline and prevent each bootstrap response from replacing rooms mutated by live events since its request began, while still allowing it to replace previews hydrated from a previous session.

</details>

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 397: Concrete race: subscription.list snapshots newest message m2, then a live delete rolls the local preview back to m1. When the delayed BUCKETS_LOADED response arrives, rowPreviewWins sees m1.createdAt <= row.lastMsgAt and replaces it with the stale m2 snapshot. No later event is guaranteed, so deleted content remains visible. Similarly, when an offline deletion leaves the room empty, `if (!preview) continue` preserves cached m2. Track per-room live mutations or a bootstrap generation: authoritative responses should replace or clear hydrated entries, but must not overwrite rooms changed after the request began.

```

</details>

chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx
```diff
@@ -271,22 +278,23 @@ export function RoomEventsProvider({ children }: { children: ReactNode }) {
...
+  // debounce is what keeps a busy room from writing per message: the timer
+  // restarts on each change, so a burst costs one write once it settles.
   //
   // The identity check is the teardown guard. `RESET` returns the
   // `initialState` OBJECT, so `=== initialState.subscriptions` is an exact
   // "we've been reset" test — without it, logout (and StrictMode's
   // double-mount) would persist the empty post-RESET state over a good cache.
-  const { subscriptions, favoriteIds, appIds, channelDmIds, chatlist } = state
+  const { subscriptions, favoriteIds, appIds, channelDmIds, chatlist, previews } = state
   useEffect(() => {
```
julianshen, line 291: _🏷️ reliability_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`tgd-review`_

**Trailing debounce can discard the latest preview snapshot**

Every preview change cancels and restarts this timer. Continuous traffic can postpone persistence indefinitely, while logout, unmount, or page closure cancels the pending final write. For example, deleting a message and immediately logging out leaves its old text cached for the next hydration. Use a bounded throttle or maximum wait and flush the latest non-reset snapshot during an appropriate teardown/pagehide path; add coverage for teardown before the debounce expires.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx` around line 291: Every preview change cancels and restarts this timer. Continuous traffic can postpone persistence indefinitely, while logout, unmount, or page closure cancels the pending final write. For example, deleting a message and immediately logging out leaves its old text cached for the next hydration. Use a bounded throttle or maximum wait and flush the latest non-reset snapshot during an appropriate teardown/pagehide path; add coverage for teardown before the debounce expires.

```

</details>

PR #375, opened 2026-08-25
feat(media-service): add DEFAULT_AVATAR_ENABLED toggle for the generated default avatar

chenjr0719:
## Summary

Add a config toggle to enable/disable the **generated default avatar** — the on-the-fly "initials" SVG media-service synthesizes when no custom image exists. When disabled, the avatar endpoints return `404 not_found` instead of the SVG, so the client renders its own fallback (the same `` contract already documented for the employee-photo `404` case). Default **on**, preserving current behavior.

## What's included

- New env flag `DEFAULT_AVATAR_ENABLED` (default `true`).
- A single guard in `serveDefault` — the one chokepoint every no-custom-image path routes through (rooms, DM/botDM, users without an `employeeId`, bots, and the blob-missing fallback) — so the toggle covers all of them consistently.
- Docs: the avatar section of `docs/client-api.md` documents the disabled `404` behavior.
- Unit tests: disabled → `404` for room / user / bot; enabled → SVG (contrast).

## Changes

- `media-service/config.go` — add `DefaultAvatarEnabled bool` (`DEFAULT_AVATAR_ENABLED`, default `true`).
- `media-service/handler.go` — `serveDefault` returns `errcode.NotFound("avatar not found")` (via `errhttp.Write`) when the flag is off, before synthesizing the SVG. No route or storage change.
- `media-service/avatar_toggle_test.go` — disabled-path tests (room/user/bot → 404) + an enabled-path contrast test.
- `media-service/handler_test.go` — set the flag in the shared test config so existing default-SVG tests keep their current behavior.
- `docs/client-api.md` — avatar section note.

## Verification

- `go test ./media-service/...` — green (full package).
- Lint clean on the touched service.
- Enabled (default) path unchanged: SVG served with the existing `ETag`/`Cache-Control`/`304` behavior. Disabled path returns `404 not_found` across rooms, users-without-`employeeId`, and bots.

## Summary by CodeRabbit

* **New Features**
  * Added a configuration option to enable or disable server-generated default avatar initials.
  * When disabled, avatar requests return a cacheable `404`, allowing clients to display their own fallback image.
  * Default avatar generation remains enabled by default.
  * Avatar responses include cache controls so configuration changes take effect after the cache period or an explicit purge.

* **Documentation**
  * Documented the configuration option, caching behavior, and client fallback handling.

* **Tests**
  * Added coverage for enabled and disabled behavior across user, bot, and room avatars.

### DIFF
--- a/docs/client-api.md
+++ b/docs/client-api.md
@@ -6875,6 +6875,8 @@ HTTP endpoints served by `media-service` — the three GETs are public, the two 
 
 **Default image:** when no custom image exists (and for users with no `employeeId`), the service returns a deterministic SVG "initials" avatar (`Content-Type: image/svg+xml`) generated on the fly — never stored. The SVG is cacheable: it carries a stable `ETag` and `Cache-Control: public, max-age=<cfg>`.
 
+**Default-avatar toggle:** the generated default is gated by `DEFAULT_AVATAR_ENABLED` (default `true`). When disabled, every avatar endpoint returns `404 not_found` instead of synthesizing the SVG (for rooms, users with no `employeeId`, and bots), and the client MUST render its own fallback on image-load failure — the same `<img onerror>` contract as the employee-photo `404` case below.
+
 **Frontend-default contract for user avatars:** after a `307` to the employee-photo host, a user who has an `employeeId` but no actual photo on that host receives a `404` from the external service. The client MUST render its own fallback on image-load failure (`<img onerror>`). The server-side default (initials SVG) only covers users with no `employeeId`, bots, and rooms.
 
 ### GET /api/v1/avatar/:accountName
--- a/docs/client-api.md
+++ b/docs/client-api.md
@@ -6875,7 +6875,7 @@ HTTP endpoints served by `media-service` — the three GETs are public, the two 
 
 **Default image:** when no custom image exists (and for users with no `employeeId`), the service returns a deterministic SVG "initials" avatar (`Content-Type: image/svg+xml`) generated on the fly — never stored. The SVG is cacheable: it carries a stable `ETag` and `Cache-Control: public, max-age=<cfg>`.
 
-**Default-avatar toggle:** the generated default is gated by `DEFAULT_AVATAR_ENABLED` (default `true`). When disabled, every avatar endpoint returns `404 not_found` instead of synthesizing the SVG (for rooms, users with no `employeeId`, and bots), and the client MUST render its own fallback on image-load failure — the same `<img onerror>` contract as the employee-photo `404` case below.
+**Default-avatar toggle:** the generated default is gated by `DEFAULT_AVATAR_ENABLED` (default `true`). When disabled, every avatar endpoint returns `404 not_found` instead of synthesizing the SVG (for rooms, users with no `employeeId`, and bots), and the client MUST render its own fallback on image-load failure — the same `<img onerror>` contract as the employee-photo `404` case below. Both the SVG and the disabled `404` carry the same `Cache-Control: public, max-age=<cfg>`, so **a toggle change propagates to shared/browser caches within that window** (default 6h); purge those caches to apply it immediately.
 
 **Frontend-default contract for user avatars:** after a `307` to the employee-photo host, a user who has an `employeeId` but no actual photo on that host receives a `404` from the external service. The client MUST render its own fallback on image-load failure (`<img onerror>`). The server-side default (initials SVG) only covers users with no `employeeId`, bots, and rooms.
 
--- a/media-service/config.go
+++ b/media-service/config.go
@@ -77,6 +77,11 @@ type config struct {
 	MaxUploadBytes     int64 `env:"MAX_UPLOAD_BYTES" envDefault:"1048576"`
 	CacheMaxAgeSeconds int   `env:"CACHE_MAX_AGE_SECONDS" envDefault:"21600"`
 
+	// DefaultAvatarEnabled gates the generated "initials" default avatar. When
+	// false, the avatar endpoints return 404 instead of synthesizing an SVG, so
+	// the client renders its own fallback.
+	DefaultAvatarEnabled bool `env:"DEFAULT_AVATAR_ENABLED" envDefault:"true"`
+
 	// Custom-emoji upload limits. Bytes cap the raw body; dimension caps the
 	// decoded width AND height independently.
 	EmojiMaxUploadBytes int64 `env:"EMOJI_MAX_UPLOAD_BYTES" envDefault:"262144"`
--- a/media-service/config_test.go
+++ b/media-service/config_test.go
@@ -93,6 +93,32 @@ func TestConfig_MaxConcurrency(t *testing.T) {
 	})
 }
 
+func TestConfig_DefaultAvatarEnabled(t *testing.T) {
+	t.Setenv("SITE_ID", "s1")
+	t.Setenv("CLUSTER_DOMAINS", `[{"siteID":"s1","domain":"http://localhost:8080"}]`)
+	t.Setenv("EMPLOYEE_PHOTO_BASE_URL", "https://photos.example.com")
+	t.Setenv("MONGO_URI", "mongodb://localhost:27017")
+	t.Setenv("MINIO_ENDPOINT", "localhost:9000")
+	t.Setenv("MINIO_ACCESS_KEY", "k")
+	t.Setenv("MINIO_SECRET_KEY", "s")
+	t.Setenv("NATS_URL", "nats://localhost:4222")
+	t.Setenv("BOTPLATFORM_URL", "https://botplatform.example.com")
+
+	t.Run("default true", func(t *testing.T) {
+		require.NoError(t, os.Unsetenv("DEFAULT_AVATAR_ENABLED"))
+		cfg, err := env.ParseAs[config]()
+		require.NoError(t, err)
+		assert.True(t, cfg.DefaultAvatarEnabled)
+	})
+
+	t.Run("override false", func(t *testing.T) {
+		t.Setenv("DEFAULT_AVATAR_ENABLED", "false")
+		cfg, err := env.ParseAs[config]()
+		require.NoError(t, err)
+		assert.False(t, cfg.DefaultAvatarEnabled)
+	})
+}
+
 // TestConfig_BotplatformURLRequired guards the tag that keeps the write endpoints
 // from ever being served anonymously: absent and empty must both refuse to start.
 func TestConfig_BotplatformURLRequired(t *testing.T) {
--- a/media-service/config_test.go
+++ b/media-service/config_test.go
@@ -105,7 +105,13 @@ func TestConfig_DefaultAvatarEnabled(t *testing.T) {
 	t.Setenv("BOTPLATFORM_URL", "https://botplatform.example.com")
 
 	t.Run("default true", func(t *testing.T) {
+		orig, had := os.LookupEnv("DEFAULT_AVATAR_ENABLED")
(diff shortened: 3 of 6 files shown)


media-service/config_test.go
```diff
@@ -93,6 +93,32 @@ func TestConfig_MaxConcurrency(t *testing.T) {
...
+	t.Setenv("EMPLOYEE_PHOTO_BASE_URL", "https://photos.example.com")
+	t.Setenv("MONGO_URI", "mongodb://localhost:27017")
+	t.Setenv("MINIO_ENDPOINT", "localhost:9000")
+	t.Setenv("MINIO_ACCESS_KEY", "k")
+	t.Setenv("MINIO_SECRET_KEY", "s")
+	t.Setenv("NATS_URL", "nats://localhost:4222")
+	t.Setenv("BOTPLATFORM_URL", "https://botplatform.example.com")
+
+	t.Run("default true", func(t *testing.T) {
+		require.NoError(t, os.Unsetenv("DEFAULT_AVATAR_ENABLED"))
```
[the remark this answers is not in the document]
chenjr0719: Applied in e1b5591a — the default-value subtest now captures the original `DEFAULT_AVATAR_ENABLED` with `os.LookupEnv` and restores it via `t.Cleanup`, so it no longer leaks unset state to later tests.

media-service/handler.go
```diff
@@ -48,6 +49,12 @@ func (h *handler) setImageCacheHeaders(c *gin.Context, etag string) {
 
 func (h *handler) serveDefault(c *gin.Context, kind, seed, name string) {
 	c.Set("media_kind", kind)
+	if !h.cfg.DefaultAvatarEnabled {
+		// Default generated avatar disabled: 404 so the client renders its own fallback.
+		c.Set("media_outcome", "disabled")
+		errhttp.Write(c.Request.Context(), c, errcode.NotFound("avatar not found"))
+		return
+	}
```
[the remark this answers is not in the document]
chenjr0719: Leaving this as-is, with reasoning: the 304-before-blob-check lives in `serveStored`, which this PR does not touch — it is the existing caching behavior. That 304 fires only when the client`\s `If-None-Match` matches a *stored* avatar record`\s ETag, i.e. the client already holds a valid cached custom image; that request never reaches `serveDefault`. The toggle governs the generated default (`serveDefault`), not validation of an already-cached custom image. The case you describe (stored record + missing blob + matching `If-None-Match`) is a pre-existing storage-inconsistency edge, and moving the ETag check after `blobs.Get` would defeat the 304 optimization on every cached hit. If the missing-blob inconsistency is worth hardening, that would be a separate change to `serveStored` rather than part of this toggle.

julianshen:
**No actionable comments.** ✅

### Local review memory

- `media-service/handler.go:51` — @chenjr0719: Applied in 331eb27f — the disabled 404 now carries the same bounded Cache-Control: public, max-age=<cfg> the SVG uses, so a disabled deployment no longer re-runs the EmployeeID/BotSite/RoomSite lookup on every render. Asserted the header… ([thread](https://github.com/hmchangw/newchat/pull/375#discussion_r3852023942))

<details>

* `docs/client-api.md`
* `media-service/avatar_toggle_test.go`
* `media-service/config.go`
* `media-service/config_test.go`
* `media-service/handler.go`
* `media-service/handler_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

docs/client-api.md
```diff
@@ -6875,6 +6875,8 @@ HTTP endpoints served by `media-service` — the three GETs are public, the two
 
 **Default image:** when no custom image exists (and for users with no `employeeId`), the service returns a deterministic SVG "initials" avatar (`Content-Type: image/svg+xml`) generated on the fly — never stored. The SVG is cacheable: it carries a stable `ETag` and `Cache-Control: public, max-age=<cfg>`.
 
+**Default-avatar toggle:** the generated default is gated by `DEFAULT_AVATAR_ENABLED` (default `true`). When disabled, every avatar endpoint returns `404 not_found` instead of synthesizing the SVG (for rooms, users with no `employeeId`, and bots), and the client MUST render its own fallback on image-load failure — the same `<img onerror>` contract as the employee-photo `404` case below.
```
julianshen: _🏷️ distributed consistency_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`_

**Cached SVGs can outlive disabling the default-avatar toggle.**

An enabled replica returns an SVG with `Cache-Control: public, max-age=<cfg>` (six hours by default). If a browser or CDN caches that response and the deployment then sets `DEFAULT_AVATAR_ENABLED=false`, subsequent image loads can be satisfied from the cache without reaching `serveDefault`, so clients continue showing the generated avatar rather than their fallback. During a rolling deployment, an old or still-enabled replica can refresh that cached 200 even after newer replicas return 404. Either purge shared caches as part of the transition and document the bounded propagation delay, or introduce a cache-generation/versioned URL (or mandatory revalidation) so changing the flag invalidates prior responses.

<details>

```
In `docs/client-api.md` around line 6878: An enabled replica returns an SVG with `Cache-Control: public, max-age=<cfg>` (six hours by default). If a browser or CDN caches that response and the deployment then sets `DEFAULT_AVATAR_ENABLED=false`, subsequent image loads can be satisfied from the cache without reaching `serveDefault`, so clients continue showing the generated avatar rather than their fallback. During a rolling deployment, an old or still-enabled replica can refresh that cached 200 even after newer replicas return 404. Either purge shared caches as part of the transition and document the bounded propagation delay, or introduce a cache-generation/versioned URL (or mandatory revalidation) so changing the flag invalidates prior responses.

```

</details>

media-service/handler.go
```diff
@@ -48,6 +49,12 @@ func (h *handler) setImageCacheHeaders(c *gin.Context, etag string) {
...
 	c.Set("media_outcome", "default")
```
julianshen: _🏷️ distributed caching_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`_

**The disabled 404 can remain cached after the toggle is re-enabled.**

The new 404 has no explicit cache policy, and 404 is heuristically cacheable under HTTP semantics; shared caches may also apply negative-caching policies. A cache can therefore store the disabled response, the service can later restart with the flag enabled, and clients can continue receiving the cached 404 without reaching the now-enabled handler. Mark the feature-toggle 404 as `no-store` and assert that header in the disabled-path tests.

<details>

> ‼️ **IMPORTANT**

```suggestion
	if !h.cfg.DefaultAvatarEnabled {
		// Default generated avatar disabled: 404 so the client renders its own fallback.
		c.Set("media_outcome", "disabled")
		c.Header("Cache-Control", "no-store")
		errhttp.Write(c.Request.Context(), c, errcode.NotFound("avatar not found"))
		return
	}
```

</details>

<details>

```
In `media-service/handler.go` around line 53: The new 404 has no explicit cache policy, and 404 is heuristically cacheable under HTTP semantics; shared caches may also apply negative-caching policies. A cache can therefore store the disabled response, the service can later restart with the flag enabled, and clients can continue receiving the cached 404 without reaching the now-enabled handler. Mark the feature-toggle 404 as `no-store` and assert that header in the disabled-path tests.

```

</details>

media-service/handler.go
```diff
@@ -48,6 +49,12 @@ func (h *handler) setImageCacheHeaders(c *gin.Context, etag string) {
 
 func (h *handler) serveDefault(c *gin.Context, kind, seed, name string) {
 	c.Set("media_kind", kind)
```
julianshen, line 51: _⚡ performance_ | _🟠 Warning_ | _⚡ Quick fix_ | _`performance-and-concurrency`_

**Disabled avatars lose explicit caching and amplify datastore load**

When the toggle is off, each missing user, bot, or room avatar still performs its EmployeeID/BotSite/RoomSite lookup before reaching this branch, but the resulting 404 omits the explicit Cache-Control/ETag used by generated defaults. Clients and proxies that do not assign heuristic freshness to 404s will repeat that lookup on every image render or page load, materially increasing service and datastore traffic for installations with many default avatars. Add an explicit, bounded negative-cache policy to this deterministic 404 response, ideally with an ETag distinct from the enabled SVG, and test the response headers.

<details>

```
In `media-service/handler.go` around line 51: When the toggle is off, each missing user, bot, or room avatar still performs its EmployeeID/BotSite/RoomSite lookup before reaching this branch, but the resulting 404 omits the explicit Cache-Control/ETag used by generated defaults. Clients and proxies that do not assign heuristic freshness to 404s will repeat that lookup on every image render or page load, materially increasing service and datastore traffic for installations with many default avatars. Add an explicit, bounded negative-cache policy to this deterministic 404 response, ideally with an ETag distinct from the enabled SVG, and test the response headers.

```

</details>

julianshen (commented):
tGD inline review

media-service/handler.go
chenjr0719: Applied in 331eb27f — the disabled 404 now carries the same bounded `Cache-Control: public, max-age=<cfg>` the SVG uses, so a disabled deployment no longer re-runs the EmployeeID/BotSite/RoomSite lookup on every render. Asserted the header in the disabled-path test. (Kept it to Cache-Control; skipped a distinct ETag as it adds little for a deterministic empty body.)

media-service/handler.go
chenjr0719: Addressed in 331eb27f by making the 404`\s caching explicit rather than heuristic — but with a bounded `max-age` instead of `no-store`. `no-store` conflicts with the performance finding on this same response (an uncached 404 re-hits the datastore on every avatar render for a disabled deployment). Since this endpoint is cache-heavy by design and the flag is deploy-time config, an explicit bounded negative-cache (predictable, not heuristic) is the better balance: a toggle change propagates within `max-age` — the same bounded window the enabled SVG already has — and that delay is now documented, with cache-purge as the immediate-apply path.

docs/client-api.md
chenjr0719: Documented in 331eb27f: the doc now states both the SVG and the disabled 404 share the same `Cache-Control: public, max-age=<cfg>`, so a toggle change propagates to shared/browser caches within that window (default 6h), and to purge those caches to apply it immediately. Chose the documented-bounded-delay option over versioned URLs / cache-generation, since this is a deploy-time toggle rather than a per-request concern and the whole avatar surface is intentionally long-cached.

media-service/config.go
```diff
@@ -77,6 +77,11 @@ type config struct {
 	MaxUploadBytes     int64 `env:"MAX_UPLOAD_BYTES" envDefault:"1048576"`
 	CacheMaxAgeSeconds int   `env:"CACHE_MAX_AGE_SECONDS" envDefault:"21600"`
 
+	// DefaultAvatarEnabled gates the generated "initials" default avatar. When
+	// false, the avatar endpoints return 404 instead of synthesizing an SVG, so
+	// the client renders its own fallback.
+	DefaultAvatarEnabled bool `env:"DEFAULT_AVATAR_ENABLED" envDefault:"true"`
```
mliu33, line 83: This will apply all the rooms and bot. Is this what we want ?

mliu33 (approved):

media-service/config.go
chenjr0719: Good point. Right now it is a single global toggle, so yes — disabling it drops the generated default for rooms, users-without-a-photo, and bots alike. On reflection per-subject control reads cleaner (you may want it off for rooms but keep it for bots, etc.). I will follow up with a change that splits this into per-subject toggles rather than one global flag.

PR #378, opened 2026-08-25
feat(auth): scope the client request/reply inbox to the caller's account

hmchangw:
Self-contained: it carries the template change, the auth-service work that enforces and proves it, the client change, and the docs. The only thing split out to [#381](https://github.com/hmchangw/newchat/pull/381) is the unrelated **backend-to-backend** reply prefix (`chat.server.response`), which touches `pkg/natsutil` / `pkg/subject` and nothing here depends on.

## Problem

Every client JWT is scoped by one signing-key template (`docker-local/setup.sh`). Its user-scoped grants were already per-account — `chat.user.{{tag(account)}}.>` — but the inbox grants were not:

```
--allow-sub "_INBOX.>"
--allow-pub "_INBOX.>"
```

`_INBOX` is not special to nats-server; it is a client-library convention over an ordinary subject. A wildcard grant on it handed every authenticated user the whole reply namespace, in both directions:

- **Read.** Any user could `SUB _INBOX.>` and receive every other user's RPC replies — `subscription.list`, room metadata, history pages, and the `key.get` responses carrying room key material.
- **Write.** Any user could publish to another user's live inbox subject and land a forged reply. `nc.request` resolves on the first message to arrive, so a spoofed reply wins if it beats the real responder.

Neither failed visibly: nothing errored, nothing logged, and the grant looked correct beside the per-account grants next to it.

## What changed

Clients set `inboxPrefix` to `chat.user.{account}`, so replies land inside the namespace the template **already** grants. Both `_INBOX` grants are deleted as dead.

| Area | Change |
|---|---|
| `docker-local/setup.sh` | Both `_INBOX` grants deleted; no grant added. **This is the security fix.** |
| `auth-service/handler.go` | `canonicalAccount()` returns the account exactly as the JWT tag carries it, on all three auth paths (sso / session / dev), so clients never re-derive the server's normalisation |
| `auth-service/permissions_integration_test.go` | Operator-mode fixture that executes the real template and proves the boundary |
| `auth-service/setup_script_test.go` | Bidirectional drift guard binding `setup.sh` to the tested grant set, plus a guard on the `sh -c` payload quoting |
| `chat-frontend` | `userInboxPrefix()`, three connect sites, and a `nc.status()` observer that tears the link down on a denied inbox |
| docs | `client-api.md` §2.1 and derived view, plus four other docs whose grant claims had gone stale |

No backend service changes: the `backend` user keeps `--allow-pub ">"`, so every responder replies exactly as before.

### Why this subject, and what it costs

Putting replies inside `chat.user.{account}.>` means **a client can adopt the new prefix against an unchanged server** — that grant is live in production today. Clients migrate on their own schedule; deleting the `_INBOX` grants afterwards is a cleanup with no client in the loop.

The trade: the user namespace is granted for publish as well as subscribe, so a client can publish into **its own** inbox. Cross-user forgery stays blocked — another account's publish grant never reaches this one. `InboxPublishDenied` is therefore replaced by `CrossUserInboxPublishDenied`, asserting the direction that still holds. The give-back is permanent by construction: a deny rule needs a subject to match on, and a bare prefix yields an opaque nuid with no distinct segment.

## Verification

Nothing in the repo previously ran nats-server in operator/JWT mode, so the template had never been executed by any test — the pre-existing `chat.user.{{tag(account)}}.>` grant was equally unverified. This PR closes that gap: operator/account/user NKeys built in Go, a server on a memory resolver, scoped JWTs minted per account tag.

The boundary tests were written red-first against the *old* wildcard grant, so the failure they first produced was the vulnerability itself.

| Test | Asserts |
|---|---|
| `InboxIsolation` | bob cannot subscribe to alice's inbox |
| `RequestReplyRoundTrip` | a reply flows back with no dedicated inbox grant |
| `CrossUserInboxPublishDenied` | bob cannot publish into alice's inbox |
| `UserNamespace` | alice's own namespaces work (asserted via `ErrTimeout`, so an unrelated error cannot pass); bob's are denied both ways |
| `SetupScriptMatchesTemplate` | `setup.sh` and the Go constants agree **in both directions** |
| `SetupScriptShCPayloadIsSingleQuoted` | no apostrophe truncates the `sh -c` payload |
| `CanonicalAccount_MatchesJWTTag` | the account returned to clients equals the JWT tag value, including for `İUser` / `ΟΣ` |

Both `setup.sh` guards live in an untagged file so `make test` always runs them — behind the integration tag they would never fire on a PR that changed `setup.sh` alone, which is exactly what they exist to catch.

Local: `make test` and `make test-integration` for auth-service (uncached), `make lint` (0 issues), frontend 1072 tests, `npm run typecheck`.

## Rollout

Two independent steps, in either order, with no window to coordinate:

1. **Ship the clients.** The new prefix works against the current production template unchanged.
2. **Delete the `_INBOX` grants** once telemetry shows no client still uses the default prefix.

Clients that have not migrated keep working until step 2. Reviewers have raised that long-lived old tabs make step 2's timing non-trivial — that thread is open and is a scheduling decision, not a code change.

## Follow-up (pre-existing, not addressed here)

Several services issue `nc.Request` on `chat.user.{account}.request.…msg.get` — a subject the account's own client may subscribe to. A client can receive such a request and use its one-shot `allow-pub-response` grant to reply into a *service's* inbox, racing the real responder. That predates this change and is out of its scope.

Design spec (with an amendment recording the subject change) and implementation plan are under `docs/superpowers/`. The plan is marked SUPERSEDED — its checklists target the earlier `_INBOX` design and would undo this change if followed.

## Summary by CodeRabbit

- **New Features**
  - Added account-scoped request/reply channels to isolate client traffic.
  - Preserved authenticated account values exactly, including capitalization and non-ASCII characters.
  - Added connection handling for permission violations, including error reporting and safe disconnection.

- **Bug Fixes**
  - Improved consistency between authentication responses, client connections, and message subjects.

- **Documentation**
  - Updated API, messaging, authentication, traffic, and avatar guidance for the revised behavior.

- **Tests**
  - Added coverage for inbox isolation, request/reply flows, permission enforcement, account formatting, and setup validation.


julianshen:

### Related work

- [PR #381](https://github.com/hmchangw/newchat/pull/381) — feat\(natsutil\): put server-to-server replies on chat.server.response (open)

<details>

* `auth-service/handler.go`
* `auth-service/handler_test.go`
* `auth-service/permissions_integration_test.go`
* `auth-service/setup_script_test.go`
* `chat-frontend/scripts/liveStack.smoke.mjs`
* `chat-frontend/smoke-test.mjs`
* `chat-frontend/src/api/_transport/inbox.js`
* `chat-frontend/src/api/_transport/inbox.test.js`
* `chat-frontend/src/api/_transport/subjects.test.js`
* `chat-frontend/src/api/_transport/subjects.ts`
* `chat-frontend/src/context/NatsContext/NatsContext.jsx`
* `chat-frontend/src/context/NatsContext/NatsContext.test.jsx`
* `docker-local/setup.sh`
* `docs/client-api.md`
* `docs/nats-subject-naming.md`
* `docs/nats-traffic-estimation.md`
* `docs/specs/botplatform/auth.md`
* `docs/specs/botplatform/traffic-isolation.md`
* `docs/superpowers/plans/2026-08-25-per-user-inbox-prefix.md`
* `docs/superpowers/specs/2026-08-25-per-user-inbox-prefix-design.md`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

julianshen (commented):
tGD inline review

chat-frontend/src/api/_transport/inbox.js
```diff
@@ -0,0 +1,17 @@
...
+// account while auth-service's own validation (handler.go:189, via
+// subject.IsValidAccountToken) still admits uppercase. A verbatim prefix
+// would fail in the worst way available: the connection succeeds and every
+// reply is then silently denied.
+//
+// Note the namespace is granted for publish as well as subscribe, so a client
+// can publish to its own inbox. That is self-spoofing only — another user's
+// publish grant is scoped to their own account — and the bare prefix leaves
+// no distinct segment to hang a deny rule on.
+export const userInboxPrefix = (account) => `chat.user.${account.toLowerCase()}`
```
julianshen: _🏷️ identity_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`, `tgd-review` (2 rules)_

**JavaScript and Go lowercase Unicode accounts differently.**

Accounts are validated as NATS subject tokens, which permits non-ASCII tokens, but this independently reproduces jwt/v2's Go lowercasing in JavaScript. For an account such as `İUser`, Go's `strings.ToLower` produces `iuser` for the JWT tag while JavaScript produces `i\u0307user`. The connection therefore succeeds with a grant for `chat.user.iuser.>`, but nats.ws subscribes under `chat.user.i\u0307user`; NATS denies the mux subscription and every RPC fails. Return the exact normalized tag value from auth-service and use it for the prefix, or restrict account identifiers to an explicitly shared ASCII normalization contract.

<details>

- **`tgd-review`**:17: `subject.IsValidAccountToken` permits Unicode account tokens, but JavaScript's full Unicode lowercasing is not equivalent to the rune-by-rune lowercasing used by Go's `strings.ToLower` in `jwt.TagList.Add`. For example, `ΟΣ` becomes `ος` in JavaScript but `οσ` in Go, so the generated inbox falls outside the JWT grant and every request times out. Define one canonical account representation at the authentication boundary—such as enforcing an ASCII account grammar or returning the exact normalized tag value—instead of independently lowercasing it in the client.

</details>

<details>

```
In `chat-frontend/src/api/_transport/inbox.js` around line 17: Accounts are validated as NATS subject tokens, which permits non-ASCII tokens, but this independently reproduces jwt/v2's Go lowercasing in JavaScript. For an account such as `İUser`, Go's `strings.ToLower` produces `iuser` for the JWT tag while JavaScript produces `i\u0307user`. The connection therefore succeeds with a grant for `chat.user.iuser.>`, but nats.ws subscribes under `chat.user.i\u0307user`; NATS denies the mux subscription and every RPC fails. Return the exact normalized tag value from auth-service and use it for the prefix, or restrict account identifiers to an explicitly shared ASCII normalization contract.

```

</details>

chat-frontend/src/context/NatsContext/NatsContext.jsx
```diff
@@ -196,6 +197,7 @@ export function NatsProvider({ children }) {
       const nc = await natsConnect({
         servers: portal.natsUrl,
         authenticator,
+        inboxPrefix: userInboxPrefix(userInfo.account),
```
julianshen: _🏷️ backpressure_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`, `nats`, `performance-and-concurrency` (3 rules)_

**The inbox prefix fans each reply out to personal-event subscribers.**

Every client already subscribes to `chat.user.{account}.>`. After this change, a backend reply also matches that wildcard in addition to the requester's mux subscription; NATS delivers once per matching subscription rather than letting the request mux suppress the wildcard delivery. With multiple sessions for one account, one RPC reply is therefore sent to every session, and twice to the requesting connection. Large replies such as the documented 128 KB subscription pages amplify broker traffic and enqueue unsolicited payloads in background tabs. Put request inboxes outside the personal-event wildcard (for example a scoped `_INBOX.{account}` namespace with a staged grant rollout), or replace the broad event subscription with non-overlapping subjects.

<details>

- **`nats`**:200: For account `alice`, nats.ws creates its request mux subscription under `chat.user.alice.<nuid>.*`, while the documented baseline subscription is `chat.user.alice.>`. Core NATS fan-out delivers each reply to both matching subscription IDs; the library does not route it exclusively to the request promise. Consequently every synchronous reply—including responses documented as reaching 128 KB—is also queued and dispatched through the personal-event handler, increasing pending bytes and exposing opaque RPC payloads to that handler. Use a disjoint namespace such as `_INBOX.<account>` with a scoped subscribe-only grant, or replace the broad personal wildcard with subscriptions that do not overlap request inboxes.
- **`performance-and-concurrency`**:200: Every RPC reply now matches both the request mux subscription (`chat.user.<account>.<nuid>.*`) and the client's documented baseline subscription (`chat.user.<account>.>`). NATS delivers once per matching subscription, so each reply is sent and dispatched twice; under normal RPC load this doubles reply WebSocket traffic and processing, including responses documented as reaching 128 KB. Move inboxes to a per-account namespace outside the baseline wildcard, such as `_INBOX.<account>` with a scoped subscribe grant, or narrow the baseline subscriptions so replies match only the mux.

</details>

<details>

```
In `chat-frontend/src/context/NatsContext/NatsContext.jsx` around line 200: Every client already subscribes to `chat.user.{account}.>`. After this change, a backend reply also matches that wildcard in addition to the requester's mux subscription; NATS delivers once per matching subscription rather than letting the request mux suppress the wildcard delivery. With multiple sessions for one account, one RPC reply is therefore sent to every session, and twice to the requesting connection. Large replies such as the documented 128 KB subscription pages amplify broker traffic and enqueue unsolicited payloads in background tabs. Put request inboxes outside the personal-event wildcard (for example a scoped `_INBOX.{account}` namespace with a staged grant rollout), or replace the broad event subscription with non-overlapping subjects.

```

</details>

chat-frontend/src/context/NatsContext/NatsContext.jsx
```diff
@@ -220,6 +222,28 @@ export function NatsProvider({ children }) {
...
+      // other permission error under it. That namespace is granted for both
+      // pub and sub, so nothing else should ever land here.
+      const inboxPrefix = userInboxPrefix(userInfo.account)
+      if (typeof nc.status === 'function') {
+        void (async () => {
+          for await (const s of nc.status()) {
+            if (myGen !== connectGenRef.current) return
+            const subject = s?.permissionContext?.subject
+            if (subject && subject.startsWith(`${inboxPrefix}.`)) {
+              setError('Reply inbox permission denied — reload the page to reconnect.')
```
julianshen: _🎯 concurrency_ | _🟠 Warning_ | _⚡ Quick fix_ | _`performance-and-concurrency`_

**Permission failures leave the mux retry cycle running.**

On a prefix/template mismatch, the stated nats.ws behavior repeatedly recreates the denied mux subscription. This handler only updates React state, leaving the connection, status loop, and request producers active, so a partial rollout can make every affected client continually generate denied subscriptions and permission events. At scale that creates a thundering herd of control traffic while RPCs continue failing. After the first matching error, close the connection or otherwise stop request production, mark it disconnected, and return from the status loop.

<details>

```
In `chat-frontend/src/context/NatsContext/NatsContext.jsx` around line 242: On a prefix/template mismatch, the stated nats.ws behavior repeatedly recreates the denied mux subscription. This handler only updates React state, leaving the connection, status loop, and request producers active, so a partial rollout can make every affected client continually generate denied subscriptions and permission events. At scale that creates a thundering herd of control traffic while RPCs continue failing. After the first matching error, close the connection or otherwise stop request production, mark it disconnected, and return from the status loop.

```

</details>

auth-service/permissions_integration_test.go
```diff
@@ -0,0 +1,374 @@
...
+				ContainerFilePath: "/etc/nats/nats.conf",
+				FileMode:          0o644,
+			}},
+			Cmd:        []string{"-c", "/etc/nats/nats.conf"},
+			WaitingFor: wait.ForLog("Server is ready").WithStartupTimeout(60 * time.Second),
+		},
+		Started: true,
+	})
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = container.Terminate(context.Background()) })
```
julianshen: _🧹 cleanup_ | _🟠 Warning_ | _⚡ Quick fix_ | _`silent-failure`_

**Container termination failures are silently ignored.**

If `Terminate` fails, the integration test still reports success while leaving the NATS container running. Repeated failures can leak containers and exhaust CI resources. Record the cleanup failure on the test.

<details>

> ‼️ **IMPORTANT**

```suggestion
	t.Cleanup(func() {
		if err := container.Terminate(context.Background()); err != nil {
			t.Errorf("terminate NATS container: %v", err)
		}
	})
```

</details>

<details>

```
In `auth-service/permissions_integration_test.go` around line 143: If `Terminate` fails, the integration test still reports success while leaving the NATS container running. Repeated failures can leak containers and exhaust CI resources. Record the cleanup failure on the test.

```

</details>

auth-service/permissions_integration_test.go
```diff
@@ -0,0 +1,374 @@
...
+	o := startOperatorNATS(t)
+
+	t.Run("own namespace allowed", func(t *testing.T) {
+		alice := o.connectScoped(t, "alice")
+
+		sub, err := alice.SubscribeSync("chat.user.alice.event")
+		require.NoError(t, err)
+		_, err = sub.NextMsg(500 * time.Millisecond)
+		require.NotErrorIs(t, err, nats.ErrPermissionViolation)
+
```
julianshen: _🧪 tests_ | _🟠 Warning_ | _⚡ Quick fix_ | _`silent-failure`_

**Allowed-subscription checks accept unrelated failures.**

Each of the four checks at lines 329, 334, 339, and 344 passes for every error except `ErrPermissionViolation`. For example, if that client disconnects after `SubscribeSync`, `NextMsg` can return `nats.ErrConnectionClosed` and the grant is still reported as allowed; later subtests use new connections, so the suite can remain green. Require the expected `nats.ErrTimeout` at all four sites, or publish and receive a probe message after flushing the subscription.

<details>

```
In `auth-service/permissions_integration_test.go` around line 329: Each of the four checks at lines 329, 334, 339, and 344 passes for every error except `ErrPermissionViolation`. For example, if that client disconnects after `SubscribeSync`, `NextMsg` can return `nats.ErrConnectionClosed` and the grant is still reported as allowed; later subtests use new connections, so the suite can remain green. Require the expected `nats.ErrTimeout` at all four sites, or publish and receive a probe message after flushing the subscription.

```

</details>

docs/nats-subject-naming.md
```diff
@@ -23,6 +23,8 @@ All subjects are dot-delimited and organized into four namespaces:
 
 On connect, every client subscribes to `chat.user.{account}.>`. This single wildcard captures all personal events:
 
+> **Note:** it also matches the client's own request/reply replies, since `inboxPrefix` is `chat.user.{account}`. A client that dispatches this wildcard into an event handler must ignore subjects it did not expect — the NATS client library delivers replies to the requesting call, not to this subscription, but both match the same pattern.
```
julianshen: _📝 documentation_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**Overlapping NATS subscriptions both receive each reply.**

NATS does not route a matching message exclusively to the request mux. It delivers a copy to every matching non-queue subscription, so the request mux and the baseline `chat.user.{account}.>` subscription both receive each RPC reply. The frontend's wildcard handler must therefore safely discard generated reply subjects; the current integration test does not model this overlap.

<details>

> ‼️ **IMPORTANT**

```suggestion
> **Note:** it also matches the client's own request/reply replies, since `inboxPrefix` is `chat.user.{account}`. NATS delivers a copy to every matching subscription, so both the request mux and this wildcard receive each reply; a wildcard handler must ignore reply subjects it does not recognise.
```

</details>

<details>

```
In `docs/nats-subject-naming.md` around line 26: NATS does not route a matching message exclusively to the request mux. It delivers a copy to every matching non-queue subscription, so the request mux and the baseline `chat.user.{account}.>` subscription both receive each RPC reply. The frontend's wildcard handler must therefore safely discard generated reply subjects; the current integration test does not model this overlap.

```

</details>

docs/client-api.md
```diff
@@ -6256,7 +6254,7 @@ See [Error envelope](#6-error-envelope-reference). The reply carries the `{ code
 
 - `{siteID}` must be the room's **origin `siteID`** (the site that owns the room), not the caller's own site.
 
-This RPC uses the **publish + async-reply** pattern, not the standard NATS request/reply. The client publishes to the `msg.send` subject (no `_INBOX.>` reply expected). `message-gatekeeper` validates the request, publishes the canonical message to `MESSAGES-CANONICAL`, and replies to `chat.user.{account}.response.{requestID}` with the persisted `Message` (or an error envelope on failure).
+This RPC uses the **publish + async-reply** pattern, not the standard NATS request/reply. The client publishes to the `msg.send` subject (no `chat.user.{account}.>` reply expected). `message-gatekeeper` validates the request, publishes the canonical message to `MESSAGES-CANONICAL`, and replies to `chat.user.{account}.response.{requestID}` with the persisted `Message` (or an error envelope on failure).
```
julianshen: _📝 documentation_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**The async-reply documentation denies its actual reply namespace.**

This now says `msg.send` expects no `chat.user.{account}.>` reply, but the same sentence identifies its async reply as `chat.user.{account}.response.{requestID}`, which is inside that wildcard. Reword this as “no synchronous request/reply response is expected” and make the same correction in `docs/client-api/request-reply.md` around line 2263.

<details>

```
In `docs/client-api.md` around line 6257: This now says `msg.send` expects no `chat.user.{account}.>` reply, but the same sentence identifies its async reply as `chat.user.{account}.response.{requestID}`, which is inside that wildcard. Reword this as “no synchronous request/reply response is expected” and make the same correction in `docs/client-api/request-reply.md` around line 2263.

```

</details>

docs/superpowers/plans/2026-08-25-per-user-inbox-prefix.md
```diff
@@ -0,0 +1,815 @@
+# Per-User NATS Inbox Prefix Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Scope the NATS request/reply inbox to the calling user's account so no client can read or forge another user's RPC replies.
+
+**Architecture:** Clients pass `inboxPrefix: "_INBOX.<account>"` at connect time; the scoped signing-key template grants `--allow-sub "_INBOX.{{tag(account)}}.>"` and no inbox publish at all. Backend responders are unaffected — the `backend` user keeps `--allow-pub ">"`. A new operator-mode integration test executes the real template and proves both the new inbox boundary and the pre-existing `chat.user.*` boundary.
```
julianshen, line 20: _📝 documentation_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**The implementation plan still prescribes the rejected inbox design.**

The newly committed plan presents an executable checklist that directs workers to use `_INBOX.<account>` and re-add `_INBOX.{{tag(account)}}.>` to the template, while this implementation and the design amendment use `chat.user.{account}` and remove both `_INBOX` grants. Following this plan would undo the security change. Mark the entire plan prominently as superseded or update/remove its obsolete tasks; also correct the related “subscribe-only inbox” wording in `docs/specs/botplatform/traffic-isolation.md`.

<details>

```
In `docs/superpowers/plans/2026-08-25-per-user-inbox-prefix.md` around line 7: The newly committed plan presents an executable checklist that directs workers to use `_INBOX.<account>` and re-add `_INBOX.{{tag(account)}}.>` to the template, while this implementation and the design amendment use `chat.user.{account}` and remove both `_INBOX` grants. Following this plan would undo the security change. Mark the entire plan prominently as superseded or update/remove its obsolete tasks; also correct the related “subscribe-only inbox” wording in `docs/specs/botplatform/traffic-isolation.md`.

```

</details>

docker-local/setup.sh
```diff
@@ -53,14 +53,19 @@ docker run --rm \
     # Scoped signing key that auth-service uses to sign user JWTs. The role
     # template mirrors the grants auth-service used to inline per JWT, keyed
     # off the account:<account> tag every user JWT now carries.
+    #
+    # There is no _INBOX grant: clients set their request/reply inbox prefix to
+    # chat.user.{account}, so replies ride the user namespace already granted
+    # below. One user therefore cannot read or forge another's replies, since
```
julianshen: _🏷️ deployment_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`nats`, `tgd-review` (2 rules)_

**The apostrophe breaks the quoted NATS setup command.**

This comment is inside the single-quoted shell payload passed to `sh -c`. The apostrophe in `another's` terminates that payload early, causing shell parsing to fail before the scoped signing-key policy can be applied. Remove the apostrophe while preserving the comment.

<details>

> ‼️ **IMPORTANT**

```suggestion
    # below. One user therefore cannot read or forge replies for another user, since
```

</details>

<details>

- **`tgd-review`**:59: This comment is inside the single-quoted shell program passed to the container. The apostrophe in “another's” terminates that quote, so setup.sh cannot reliably parse or apply the signing-key policy.

<details>

```text
    # below. One user therefore cannot read or forge replies for another user, since
```

</details>


</details>

<details>

```
In `docker-local/setup.sh` around line 59: This comment is inside the single-quoted shell payload passed to `sh -c`. The apostrophe in `another's` terminates that payload early, causing shell parsing to fail before the scoped signing-key policy can be applied. Remove the apostrophe while preserving the comment.

```

</details>

docker-local/setup.sh
```diff
@@ -53,14 +53,19 @@ docker run --rm \
     # Scoped signing key that auth-service uses to sign user JWTs. The role
     # template mirrors the grants auth-service used to inline per JWT, keyed
     # off the account:<account> tag every user JWT now carries.
+    #
+    # There is no _INBOX grant: clients set their request/reply inbox prefix to
```
julianshen: _🏷️ protocol-compatibility_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`_

**Removing _INBOX breaks mixed-version clients after reconnect.**

After this template is applied, a pre-change browser tab that reconnects still creates its mux under `_INBOX`. Its request publish remains allowed, so a mutating service can commit the operation, but the reply subscription is denied and the caller observes failure or timeout. This is especially unsafe for documented non-idempotent toggle RPCs if callers retry. Client-first deployment does not eliminate long-lived old tabs. Retain compatibility until old clients are forcibly refreshed or rejected before publishing, and remove the wildcard grant only after that population is drained.

<details>

```
In `docker-local/setup.sh` around line 57: After this template is applied, a pre-change browser tab that reconnects still creates its mux under `_INBOX`. Its request publish remains allowed, so a mutating service can commit the operation, but the reply subscription is denied and the caller observes failure or timeout. This is especially unsafe for documented non-idempotent toggle RPCs if callers retry. Client-first deployment does not eliminate long-lived old tabs. Retain compatibility until old clients are forcibly refreshed or rejected before publishing, and remove the wildcard grant only after that population is drained.

```

</details>

julianshen (commented):
tGD inline review

julianshen (commented):
tGD inline review

chat-frontend/src/context/NatsContext/NatsContext.jsx
```diff
@@ -193,9 +237,11 @@ export function NatsProvider({ children }) {
       })
```
julianshen, line 238: _🏷️ protocol-compatibility_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`_

**New clients are incompatible with old auth-service replicas.**

During a frontend-first or overlapping rollout, a mixed-case user such as `HMChang` can authenticate against an old auth-service replica. That replica returns raw `HMChang`, while its JWT TagList normalizes the grant to `chat.user.hmchang.>`; the new frontend uses the returned value verbatim and creates `chat.user.HMChang` subjects. The first request or mux subscription is therefore denied and the permission watcher drains the connection. Fully deploy and drain old auth-service replicas before releasing this frontend, and document/enforce that ordering, or introduce a backward-compatible canonical NATS-account field/versioned auth contract that the client can use across mixed versions.

<details>

```
In `chat-frontend/src/context/NatsContext/NatsContext.jsx` around line 238: During a frontend-first or overlapping rollout, a mixed-case user such as `HMChang` can authenticate against an old auth-service replica. That replica returns raw `HMChang`, while its JWT TagList normalizes the grant to `chat.user.hmchang.>`; the new frontend uses the returned value verbatim and creates `chat.user.HMChang` subjects. The first request or mux subscription is therefore denied and the permission watcher drains the connection. Fully deploy and drain old auth-service replicas before releasing this frontend, and document/enforce that ordering, or introduce a backward-compatible canonical NATS-account field/versioned auth contract that the client can use across mixed versions.

```

</details>

chat-frontend/src/context/NatsContext/NatsContext.jsx
```diff
@@ -63,6 +64,49 @@ function tracedFetch(method, url, init = {}) {
...
+
+  void (async () => {
+    for await (const s of statusIter) {
+      if (genRef.current !== gen) break
+      const subject = s?.permissionContext?.subject
+      if (subject && subject.startsWith(`${inboxPrefix}.`)) {
+        setError('Reply inbox permission denied — reload the page to reconnect.')
+        setConnected(false)
+        statusIter.stop?.()
+        nc.drain().catch(() => {})
```
julianshen, line 101: _🏷️ error-handling_ | _🟠 Warning_ | _⚡ Quick fix_ | _`silent-failure`_

**A failed drain leaves the denied connection running silently.**

After an inbox permission violation, the provider marks the session disconnected and stops observing status before calling `drain()`. If that promise rejects, the empty catch hides the failure and never force-closes the connection, so NATS may remain connected and continue recreating denied inbox subscriptions while the UI reports it as disconnected. Fall back to `close()` when draining fails, surface a failure of that fallback, and add a rejected-drain test.

<details>

> ‼️ **IMPORTANT**

```suggestion
        void nc.drain().catch(async (drainErr) => {
          try {
            await nc.close()
          } catch (closeErr) {
            setError(`Reply inbox disconnect failed: ${closeErr?.message || drainErr?.message || 'unknown error'}`)
          }
        })
```

</details>

<details>

```
In `chat-frontend/src/context/NatsContext/NatsContext.jsx` around line 101: After an inbox permission violation, the provider marks the session disconnected and stops observing status before calling `drain()`. If that promise rejects, the empty catch hides the failure and never force-closes the connection, so NATS may remain connected and continue recreating denied inbox subscriptions while the UI reports it as disconnected. Fall back to `close()` when draining fails, surface a failure of that fallback, and add a rejected-drain test.

```

</details>

docs/client-api/request-reply.md
```diff
@@ -2260,7 +2260,7 @@ kill-switch off (`forbidden`, reason `emoji_delete_disabled`), store failure (`i
 **Subject:** `chat.user.{account}.room.{roomID}.{siteID}.msg.send`
 **Async reply:** `chat.user.{account}.response.{requestID}` (subscribe to `chat.user.{account}.>` to receive it)
 
-Publish + async-reply pattern — no `_INBOX.>` reply. Covers plain message, thread reply,
+Publish + async-reply pattern — no `chat.user.{account}.>` reply. Covers plain message, thread reply,
```
julianshen, line 2263: _📝 documentation_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**The async-reply description contradicts its documented subject.**

The preceding line says the async reply is `chat.user.{account}.response.{requestID}`, which is inside `chat.user.{account}.>`, but this line says there is no such reply. Clarify that there is no synchronous NATS request/reply response; the async reply still uses this namespace.

<details>

> ‼️ **IMPORTANT**

```suggestion
Publish + async-reply pattern — no synchronous NATS request/reply response. Covers plain message, thread reply,
```

</details>

<details>

```
In `docs/client-api/request-reply.md` around line 2263: The preceding line says the async reply is `chat.user.{account}.response.{requestID}`, which is inside `chat.user.{account}.>`, but this line says there is no such reply. Clarify that there is no synchronous NATS request/reply response; the async reply still uses this namespace.

```

</details>

PR #381, opened 2026-08-26
feat(natsutil): put server-to-server replies on chat.server.response

hmchangw:
Split out of [#378](https://github.com/hmchangw/newchat/pull/378) so the client-side inbox change can land on its own. **Draft, because the deployment question below is unresolved** — the code is complete and green.

## What this does

`natsutil.Connect` sets `subject.ServerInboxPrefix` (`chat.server.response`) as every backend connection's inbox prefix, so a server-to-server reply lands on `chat.server.response.<nuid>.<token>` instead of the library default `_INBOX.<nuid>.<token>`.

One chokepoint covers the fleet: every production service dials through `natsutil.Connect`. Only diagnostic tools (`tools/nats-debug`, `tools/cdc-verify`) and `pkg/testutil` construct their own connections, and they are deliberately left alone.

## Why

Backend replies are currently indistinguishable from client reply traffic by subject. Putting them in their own namespace makes them greppable in traffic inspection and, more importantly, makes them *scopeable* — a future per-service backend credential can grant a reply namespace, which is impossible while everything shares `_INBOX`.

There is no security change today: `backend.creds` grants `>` in both directions, so nothing is newly permitted or denied.

## Verification

`TestConnectSetsServerInboxPrefix` asserts the option is stored **and** that a real round trip produces a reply subject under the prefix — the option alone would not prove nats.go builds the subject from it.

Checked for collisions before committing: every server-to-server subject is a specific `chat.server.request.…`, and nothing in the repo subscribes to a `chat.server` wildcard, so the new namespace cannot be captured by an existing subscriber. nats.go creates the request mux lazily, so a service that only consumes JetStream pays nothing.

Local: `make test SERVICE=pkg/subject`, `make test-integration SERVICE=pkg/natsutil`, `make lint` (0 issues).

## The open question — why this is a draft

**Any NATS account or gateway permission that allow-lists `_INBOX.>` for backend traffic will drop every backend reply the moment this deploys.** Dev and CI cannot catch it, because `backend.creds` grants `>`.

This is not hypothetical: `docs/specs/botplatform/traffic-isolation.md` carries a gateway config in exactly that shape, allow-listing `_INBOX.>` for cross-site traffic.

Before merging, one of:

1. **Confirm the production account and gateway permissions** allow `chat.server.response.>` — then this merges as-is.
2. **Put the prefix behind an env toggle** so it can be rolled out and reverted without a redeploy. Say the word and I will add it.
3. **Leave it parked** until the permission side is ready.

Option 1 is the cleanest if the answer is easy to get; the change is otherwise low-risk and self-contained.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RNz8HWf47Tyg2sqqzDbUeb)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -265,6 +265,7 @@ All commands are wrapped in the root Makefile. Always use `make` targets — nev
 ### NATS Subject Naming
 - Dot-delimited hierarchical subjects — use `pkg/subject` builders, never raw `fmt.Sprintf`
 - User-scoped: `chat.user.{account}.…`
+- Server-to-server replies: `chat.server.response.…` — `natsutil.Connect` sets `subject.ServerInboxPrefix` as every backend connection's inbox prefix, so backend request/reply never uses the library default `_INBOX`. Any NATS account or gateway permission that allow-lists `_INBOX.>` for backend traffic must allow `chat.server.response.>` instead.
 - Room-scoped: `chat.room.{roomID}.…`
 - MESSAGES-CANONICAL: `chat.msg.canonical.{siteID}.created` (`.edited`, `.deleted` for future)
 - Inbox (cross-site, remote-origin): `chat.inbox.{destSiteID}.external.{eventType}` — published directly into the destination site's INBOX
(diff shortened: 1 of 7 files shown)


hmchangw:
We decided not to do this.

PR #382, opened 2026-08-26
Exclude system messages from unread counts and sidebar ordering

hmchangw:
## Problem

QA: system events (`room_created`, `members_added`, `member_removed`, `member_left`, `room_renamed`, `room_restricted`, `teams_meet_started`) mark rooms unread and re-sort the sidebar for every member.

Root cause: they ride MESSAGES-CANONICAL like user messages, and broadcast-worker bumps `rooms.lastMsgAt` unconditionally — one field doing three unrelated jobs: unread, sidebar ordering, and history-service's Cassandra read ceiling. That third job is why "just stop bumping it" doesn't work: `walkBounds` sets `ceiling = lastMsgAt`, so a system message above the ceiling would render live and then vanish on reload.

## Fix

Two timestamps in the database, **one on the wire** (design: `docs/superpowers/specs/2026-08-26-system-message-unread-ordering-design.md`).

### Storage

- **`rooms.lastMsgAt` — unchanged, every message.** Still the history read ceiling, the `messageRead` floor-recompute guard, and the cross-site activity payload.
- **`rooms.lastUserMsgAt` — new, non-system messages only.** Internal: Mongo, `model.Room`, `model.EnrichedSubscription`, and the server-to-server `rooms.info` RPC.

### Wire

**The client sees one activity timestamp.** `SubscriptionRoom.lastMsgAt` carries `lastUserMsgAt ?? lastMsgAt`, resolved server-side at the three sites that build it — user-service `buildLocalRoom` and `applyRoomInfo`, room-worker `subscriptionRoomFor`. No new field, no fallback rule, no adoption step: the field clients already read now means what they always assumed it meant, so an **old client is fixed by this without changing a line**. `TestSubscriptionRoom_NeverExposesLastUserMsgAt` pins the struct by reflection against re-exposing the internal field.

The only addition to the client contract is **`RoomEvent.systemMsg`** — a plaintext marker on system `new_message` events, needed because encrypted channels seal `message.type` inside `encryptedMessage`. Clients must not advance unread or ordering from a flagged event.

### The freeze

On a system-only write window the store pins `lastUserMsgAt` once, in a pipeline `$set` evaluated against the pre-update document:

```
$ifNull($lastUserMsgAt,
        $cond($lastMsgAt is null-or-missing,
              $ifNull($createdAt, $$REMOVE),
              $$REMOVE))
```

- **Keeps an existing value** — a member who has read the room is never re-flagged by a later system event.
- **Pins `createdAt` only for a room that has never carried a message.** This keeps the reference at-or-below the creator's pre-read `lastSeenAt` (both are the same instant, and the comparison is strict), so the creator's own brand-new room does not light up.
- **Writes nothing for a room that already has a `lastMsgAt`.** That timestamp is *unclassified* — pre-deploy it was bumped by system messages too — so promoting it would persist system activity as user activity, permanently. Readers coalesce to `lastMsgAt`, which is exactly the pre-field behaviour.

No migration and no backfill.

### Newly added members still see the room as unread

Their subscription has no `lastSeenAt`, which the read rule already treats as "never opened" — zero per-recipient machinery. `preRead` is applied only to the creator/initiator, never to invited members.

### chat-frontend

Takes `room.lastMsgAt` verbatim (no fallback chain), and the reducer appends system messages to the timeline without bumping `lastMsgAt`/`unreadCount` or re-sorting — both the live and historical-buffer branches.

### Docs

`docs/client-api.md` + both derived views: `hasUnread`, ordering and `updatedWithinDays` simplify to one field; the room-object `lastUserMsgAt` entry is **removed** rather than added; `systemMsg` documented on `new_message`; `RoomEvent.lastMsgAt` distinguished from the room object's (it is the individual message's own time, system messages included — gate on `systemMsg` before folding it into a summary); and the `meta.lastMsgAt` history hint marked **omit this**, since no client-visible field carries full activity any more.

## Verification

- `make lint` — 0 issues; `make test` — 116 packages green with the race detector; `go vet -tags integration ./...` clean; `make sast-gosec` clean.
- chat-frontend: 1105/1105 tests, `npm run typecheck` clean.
- **CI green on the current head**, including `sast` (govulncheck + semgrep) and all five `test-integration` jobs.
- All new code TDD'd — failing tests first, then the minimum implementation.

## Notes for reviewers

- **Rollout order: broadcast-worker writers first and fully drained, then room-service at every site, then the user-service readers.** Old and new writers share the MESSAGES-CANONICAL queue group, so an overlap can leave `lastUserMsgAt` stale until the room's next user message reaches an upgraded writer. Rolling the writer back after the field is written has the same effect and does not self-heal.
- **Legacy rooms are not repaired.** A room whose newest pre-deploy event was a system message keeps reading against that timestamp until its next user message. That is today's behaviour, not a regression — the freeze simply declines to make it permanent. Repairing it would need a per-room backfill from history, which this design set out to avoid.
- Frontend `ROOM_METADATA_UPDATED` still patches ordering ungated — deliberate: no server publisher of that event exists today; gate it alongside any future publisher.
- **Known open, not addressed here:**
  1. room-worker does not persist `Room.LastUserMsgAt = Room.CreatedAt` at creation, so a **cold** `subscription.list` landing between room creation and broadcast-worker's first write sees a fresh room as read. A connected client is unaffected — the `subscription.update` *added* event carries the resolved reference.
  2. `lastUserMsgAt` is written unguarded, so an out-of-order coalescer flush can regress it.
- Deferred nits: extend the badge-projection integration guard's seed; one stale doc comment on `AggregateSubscriptions`; mirror tests for the non-flagged encrypted placeholder path.

## Summary by CodeRabbit

* **New Features**
  * Added separate tracking for the latest user message, distinct from system activity.
  * System messages remain visible in timelines without affecting unread counts, sidebar ordering, mentions, or room activity indicators.
  * Added system-message markers to room events, including encrypted channels.
  * Sidebar ordering and subscription filtering now prioritize user activity, with compatibility fallbacks.

* **Bug Fixes**
  * Corrected unread styling when unread status is present without an unread count badge.

* **Documentation**
  * Updated client API documentation for activity timestamps, system-message markers, and reply handling.


julianshen:
**6 findings · 4 unique issues · 3 inline comments posted** — 🔴 4 blocking · 🟠 2 warning

### 💬 Additional comments (1)


**`user-service/service/subscriptions.go:601`**

_🎯 cassandra-correctness_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`cassandra`_

**Reusing lastMsgAt for user activity can hide Cassandra rows**

`buildLocalRoom` now serializes `LastUserMsgAt` under the existing `SubscriptionRoom.lastMsgAt` field; `applyRoomInfo` does the same for cross-site rooms. Concrete path: a user message is stored at T1, a system message is stored in the Cassandra timeline at T2, and a previously deployed client echoes the room's now-T1 `lastMsgAt` as the documented history `meta.lastMsgAt` hint. The unchanged history endpoint accepts that timestamp, skips its MongoDB bounds lookup, and caps Load History/Load Next at T1, so the T2 system row disappears after reload. Updating documentation cannot protect existing clients. Preserve full activity in the existing wire `lastMsgAt` and expose user activity separately, or make history-service ignore/version this hint before changing its semantics.

### Disputed

- `user-service/service/subscriptions.go:256` (`mongodb`) — A new worker can write `rooms.lastUserMsgAt=T1`, after which an overlapping old worker processes user message T2 but updates only `lastMsgAt`. Once readers unconditionally prefer the non-null new field, T2 no longer affects unread state, ordering, or activity-window filtering. Draining old workers before deploying readers prevents further stale documents but does not repair documents created during the overlap, contrary to the rollout note. Use a non-overlapping writer cutover, repair rooms touched during overlap, or add a version/fence that lets readers detect incomplete writer coverage; rollback must also be prevented or handled.

- **Reusing lastMsgAt for user activity can hide Cassandra rows**
  - `user-service/service/subscriptions.go:601` (`cassandra`) — Reusing lastMsgAt for user activity can hide Cassandra rows
  - `broadcast-worker/store_mongo.go:157` (`performance-and-concurrency`) — Concurrent flushes can regress lastUserMsgAt.
  - `broadcast-worker/store_mongo.go:184` (`mongodb`) — Missing lastMsgAt is not proof that a room is new.
  - `broadcast-worker/store_mongo.go:184` (`tgd-review`) — Any first system event can mark an empty legacy room unread

### Local review memory

- `chat-frontend/src/context/RoomEventsContext/RoomEventsContext.tsx:122` — @coderabbitai: _🎯 Functional Correctness_ _🟠 Major_ _🏗️ Heavy lift_ **Keep full activity and user-message activity separate.** lastMsgAt must retain all-message activity. The current reducer freezes it for system events, while the unread predica… ([thread](https://github.com/hmchangw/newchat/pull/382#discussion_r3861034847))
- `user-service/service/subscriptions.go:637` — @hmchangw: Confirmed, and this is a distinct window from the writer overlap — thanks, I had only reasoned about broadcast-worker. The asymmetry that makes it worse: the writer-overlap window self-heals once the room's next user message reaches an upg… ([thread](https://github.com/hmchangw/newchat/pull/382#discussion_r3864406938))
- `broadcast-worker/coalescer.go:156` — @mliu33: Why not just update the original lastMsgAt only if it's a user message ? ([thread](https://github.com/hmchangw/newchat/pull/382#discussion_r3867873630))
- `broadcast-worker/coalescer.go:156` — @hmchangw: TL;DR: **System messages disappear from the timeline on reload — intermittently** Because lastMsgAt isn't only the unread/ordering input — it's also history-service's Cassandra read ceiling, and system messages have to stay under it. hi… ([thread](https://github.com/hmchangw/newchat/pull/382#discussion_r3867873630))

<details>
<summary>📒 Files reviewed (42)</summary>

* `broadcast-worker/lastusermsg_test.go`
* `broadcast-worker/preview_test.go`
* `chat-frontend/src/api/fetchSidebarBuckets/index.test.ts`
* `chat-frontend/src/api/fetchSidebarBuckets/index.ts`
* `chat-frontend/src/components/MainApp/Sidebar/RoomList/RoomList.jsx`
* `chat-frontend/src/components/MainApp/Sidebar/RoomList/RoomList.test.jsx`
* `chat-frontend/src/context/RoomEventsContext/selectUnread.js`
* `chat-frontend/src/context/RoomEventsContext/selectUnread.test.js`
* `docs/client-api.md`
* `docs/client-api/events.md`
* `docs/superpowers/plans/2026-08-26-system-message-unread-ordering.md`
* `docs/superpowers/specs/2026-08-26-system-message-unread-ordering-design.md`
* `pkg/model/event.go`
* `pkg/model/model_test.go`
* `pkg/model/room.go`
* `pkg/model/subscription.go`
* `pkg/timeutil/timeutil.go`
* `pkg/timeutil/timeutil_test.go`
* `room-service/handler.go`
* `room-worker/handler.go`
* `user-service/models/subscription.go`
* `user-service/mongorepo/lastusermsg_test.go`
* `user-service/mongorepo/sortkeycache.go`
* `user-service/mongorepo/subscriptions_bench_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

chat-frontend/src/api/fetchSidebarBuckets/index.ts
```diff
@@ -223,7 +223,10 @@ export function subToRoom(sub: DMSubscription, fallbackSiteId: string): Room {
     userCount: sub.room?.userCount ?? 0,
     appCount: sub.room?.appCount ?? 0,
     lastMsgId: sub.room?.lastMsgId ?? '',
-    lastMsgAt: sub.room?.lastMsgAt ?? undefined,
+    // lastMsgAt here means "user activity" client-side: seed from
+    // lastUserMsgAt (last NON-system message) and fall back to lastMsgAt
+    // for rooms that predate the field.
+    lastMsgAt: sub.room?.lastUserMsgAt ?? sub.room?.lastMsgAt ?? undefined,
```
julianshen: _🎯 Cassandra history correctness_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`cassandra`, `distributed-system`, `silent-failure`, `tgd-review` (4 rules)_

**User-activity time is incorrectly reused as the Cassandra history ceiling**

The history-service Cassandra bucket/range query uses `meta.lastMsgAt` as its scan start and upper timeline bound; `docs/client-api.md` explicitly requires the all-message `lastMsgAt` because system messages are persisted in history. This mapping instead stores `lastUserMsgAt` in the summary's `lastMsgAt`. For a user message at T1 followed by a system message at T2, opening the room from this summary supplies T1, so the Cassandra read is bounded before T2 and the system message disappears after reload. A new room containing only a membership system message is even bounded at `createdAt`. Preserve the full `sub.room.lastMsgAt` separately, add a user-activity field for sidebar sorting/unread checks, and ensure history requests pass only the full timestamp. Add a reload/history regression covering a newer system message.

<details>

- **`silent-failure`**:229: `subToRoom` overwrites the room summary's `lastMsgAt` with the earlier `lastUserMsgAt`. The documented history path passes this summary value as `meta.lastMsgAt`, which the server accepts as the timeline ceiling without fetching Mongo. After a user message followed by a system message, opening or reloading the room can therefore return a successful but partial history response that omits the newer system message, violating the requirement that system messages remain visible after reload. Keep the full-activity `lastMsgAt` intact, carry user activity in a separate `lastUserMsgAt` field, and use only that separate field for unread state and sidebar ordering.
- **`tgd-review`**:229: When a system message is newer than the last user message, this assignment replaces the room's all-message `lastMsgAt` with the older `lastUserMsgAt`. The history API documentation in this change says clients pass the summary's `lastMsgAt` as a sufficient timeline ceiling and that this hint must include system messages. A subsequent history request can therefore skip the Mongo lookup and stop before the newer system message, making it disappear after reload. Preserve both timestamps in the client model: keep `lastMsgAt` as full activity, store `lastUserMsgAt` separately, and use the latter only for unread and sidebar sorting.
- **`distributed-system`**:229: Concrete path: a room has a user message at t1 and a system message at t2. The subscription payload correctly carries lastUserMsgAt=t1 and lastMsgAt=t2, but subToRoom stores only t1 in Room.lastMsgAt and discards t2. The documented history path uses the room summary's lastMsgAt as a valid client hint and may skip the Mongo lookup; history then bounds its Cassandra scan at t1, so the system message at t2 disappears after reload despite appearing live. Preserve both timestamps in frontend state: lastMsgAt must remain the all-message history ceiling, while a separate lastUserMsgAt should drive unread and sidebar sorting.

</details>

<details>

```
In `chat-frontend/src/api/fetchSidebarBuckets/index.ts` around line 229: The history-service Cassandra bucket/range query uses `meta.lastMsgAt` as its scan start and upper timeline bound; `docs/client-api.md` explicitly requires the all-message `lastMsgAt` because system messages are persisted in history. This mapping instead stores `lastUserMsgAt` in the summary's `lastMsgAt`. For a user message at T1 followed by a system message at T2, opening the room from this summary supplies T1, so the Cassandra read is bounded before T2 and the system message disappears after reload. A new room containing only a membership system message is even bounded at `createdAt`. Preserve the full `sub.room.lastMsgAt` separately, add a user-activity field for sidebar sorting/unread checks, and ensure history requests pass only the full timestamp. Add a reload/history regression covering a newer system message.

```

</details>

broadcast-worker/store_mongo.go
```diff
@@ -145,7 +153,11 @@ func (m *mongoStore) lastMessageUpdate(u *roomLastMsgUpdate) any {
 	if !u.lastMentionAllAt.IsZero() {
 		fields["lastMentionAllAt"] = u.lastMentionAllAt
 	}
```
julianshen, line 155: _🏷️ rolling-deploy_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`, `nats` (2 rules)_

**Mixed-version writers can permanently stale lastUserMsgAt**

During a rolling broadcast-worker deployment, a new replica can process user message U1 and create lastUserMsgAt=t1. A still-running old replica can then process later user message U2 and update lastMsgAt/lastMsgId without updating the new field, leaving lastUserMsgAt=t1. Once new readers prefer the non-nil field, U2 does not mark the room unread or move it in the sidebar, and the stale state survives after the rollout until another user message happens to reach a new writer. Reader/writer deployment order alone cannot repair this interleaving. Before enabling readers, fully drain old writers and backfill/repair rooms touched during the rollout, or add a version/fencing value tied to the last-message tuple so readers can detect legacy writes and safely fall back.

<details>

- **`nats`**:155: The new field is unsafe during a rolling deployment of the consumers competing on MESSAGES-CANONICAL. If a new worker first writes `lastUserMsgAt=T1`, then an old worker receives a later user message and writes only `lastMsgAt=T2`, new readers continue preferring stale T1. A member whose read position lies between T1 and T2 is then reported as read, and the room does not reorder, until another user message happens to be processed by a new worker. Draining all old canonical-message subscribers before enabling this writer and then the readers, plus preventing rollback, is the minimal operational fix; otherwise introduce a versioned/backfilled cutover that lets readers distinguish incomplete writer coverage.

</details>

<details>

```
In `broadcast-worker/store_mongo.go` around line 155: During a rolling broadcast-worker deployment, a new replica can process user message U1 and create lastUserMsgAt=t1. A still-running old replica can then process later user message U2 and update lastMsgAt/lastMsgId without updating the new field, leaving lastUserMsgAt=t1. Once new readers prefer the non-nil field, U2 does not mark the room unread or move it in the sidebar, and the stale state survives after the rollout until another user message happens to reach a new writer. Reader/writer deployment order alone cannot repair this interleaving. Before enabling readers, fully drain old writers and backfill/repair rooms touched during the rollout, or add a version/fencing value tied to the last-message tuple so readers can detect legacy writes and safely fall back.

```

</details>

broadcast-worker/store_mongo.go
```diff
@@ -145,7 +153,11 @@ func (m *mongoStore) lastMessageUpdate(u *roomLastMsgUpdate) any {
...
-	if !m.previews {
+	systemOnly := u.userAt.IsZero()
+	if !systemOnly {
+		fields["lastUserMsgAt"] = u.userAt
```
julianshen, line 158: _🏷️ mongodb-atomicity_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`mongodb`_

**Out-of-order flushes can regress lastUserMsgAt.**

The rooms UpdateOne unconditionally places u.userAt in $set. newerRow only orders messages within one coalescer window, so a delayed message or redelivery in a later window can write t1 after another worker already wrote t2. A subscriber whose lastSeenAt is between t1 and t2 is then incorrectly reported as read, and the room moves backward in ordering. Make lastUserMsgAt monotonic atomically in MongoDB—for example, use $max in the plain update and an equivalent conditional expression in pipeline updates—and test writes applied in t2-then-t1 order.

<details>

```
In `broadcast-worker/store_mongo.go` around line 158: The rooms UpdateOne unconditionally places u.userAt in $set. newerRow only orders messages within one coalescer window, so a delayed message or redelivery in a later window can write t1 after another worker already wrote t2. A subscriber whose lastSeenAt is between t1 and t2 is then incorrectly reported as read, and the room moves backward in ordering. Make lastUserMsgAt monotonic atomically in MongoDB—for example, use $max in the plain update and an equivalent conditional expression in pipeline updates—and test writes applied in t2-then-t1 order.

```

</details>

user-service/mongorepo/subscriptions.go
```diff
@@ -356,26 +360,38 @@ func listWindowCutoff(listType string, withinDays *int) *time.Time {
...
 
+// effectiveUserAt is the ordering/window position: the last USER message,
+// falling back to lastMsgAt for rooms that predate lastUserMsgAt. Never
+// createdAt — that stays buildListRows' final undated fallback only.
+func effectiveUserAt(k roomSortKey) *time.Time {
+	return timeutil.Coalesce(k.LastUserMsgAt, k.LastMsgAt)
+}
+
 // buildListRows drops rows that shouldn't appear — rooms soft-deleted here, and
 // rooms outside the activity window — and works out where the rest sort. A room
```
julianshen, line 379: _🏷️ mongodb-compatibility_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`mongodb`_

**Mixed-version writers leave lastUserMsgAt permanently stale.**

The read path treats any non-nil lastUserMsgAt as authoritative. During an overlapping broadcast-worker rollout, a new worker can initialize it at t1, after which an old worker can process a later user message at t2 while updating only lastMsgAt. New readers continue selecting stale t1, so users with lastSeenAt between t1 and t2 miss unread state and the room remains incorrectly ordered or excluded by updatedWithinDays. The stale source document persists beyond the sort-key cache TTL. Use a coordinated non-overlapping writer cutover and repair documents touched during overlap, or introduce a versioned/two-phase compatibility mechanism that lets readers determine whether lastUserMsgAt is authoritative.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 379: The read path treats any non-nil lastUserMsgAt as authoritative. During an overlapping broadcast-worker rollout, a new worker can initialize it at t1, after which an old worker can process a later user message at t2 while updating only lastMsgAt. New readers continue selecting stale t1, so users with lastSeenAt between t1 and t2 miss unread state and the room remains incorrectly ordered or excluded by updatedWithinDays. The stale source document persists beyond the sort-key cache TTL. Use a coordinated non-overlapping writer cutover and repair documents touched during overlap, or introduce a versioned/two-phase compatibility mechanism that lets readers determine whether lastUserMsgAt is authoritative.

```

</details>

chat-frontend/src/context/RoomEventsContext/reducer.js
```diff
@@ -627,6 +641,13 @@ export function roomEventsReducer(state, action) {
           : { ...state.previews, [roomId]: nextPreview }
       const prev = state.roomState[roomId] ?? emptyRoomState()
       const isActive = state.activeRoomId === roomId
+      // Derived once for both buffer modes, so the "what counts as room
+      // activity" rule lives in one place rather than a copy per branch.
+      const nextLastMsgAt = isSystemEvent
```
julianshen, line 649: _🎯 nats-correctness_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`nats`_

**System events discard the NATS history-ceiling timestamp**

On the `new_message` consume path, a system event keeps `prev.lastMsgAt` instead of retaining `evt.lastMsgAt`. The bootstrap path similarly stores `lastUserMsgAt` under `lastMsgAt`. After a user message at T1 and a system message at T2, the event appears live but the client retains only T1. The documented history RPC accepts this valid timestamp as `meta.lastMsgAt`, skips its Mongo lookup, and uses it to bound the timeline read, so the T2 system message can disappear after reload. Keep separate full-activity and user-activity fields client-side: always advance `lastMsgAt` from the RoomEvent, advance `lastUserMsgAt` only for non-system events, and use only `lastUserMsgAt ?? lastMsgAt` for unread and sidebar ordering.

<details>

```
In `chat-frontend/src/context/RoomEventsContext/reducer.js` around line 646: On the `new_message` consume path, a system event keeps `prev.lastMsgAt` instead of retaining `evt.lastMsgAt`. The bootstrap path similarly stores `lastUserMsgAt` under `lastMsgAt`. After a user message at T1 and a system message at T2, the event appears live but the client retains only T1. The documented history RPC accepts this valid timestamp as `meta.lastMsgAt`, skips its Mongo lookup, and uses it to bound the timeline read, so the T2 system message can disappear after reload. Keep separate full-activity and user-activity fields client-side: always advance `lastMsgAt` from the RoomEvent, advance `lastUserMsgAt` only for non-system events, and use only `lastUserMsgAt ?? lastMsgAt` for unread and sidebar ordering.

```

</details>

user-service/service/subscriptions.go
```diff
@@ -253,7 +253,7 @@ func (s *UserService) enrichLocal(subs []model.EnrichedSubscription, localIdx []
 		subs[j].Room = room
 		// hasUnread / hasGroupMention are computed at read time: room activity (resp.
 		// an @all mention) newer than lastSeenAt.
-		subs[j].HasUnread = unread(subs[j].LastSeenAt, timeutil.TimeToMillis(room.LastMsgAt))
+		subs[j].HasUnread = unread(subs[j].LastSeenAt, timeutil.TimeToMillis(timeutil.Coalesce(subs[j].LastUserMsgAt, room.LastMsgAt)))
```
julianshen, line 256: _🎯 deployment-correctness_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`tgd-review`_

**Mixed-version writers can leave user activity permanently stale.**

The unconditional preference for a non-nil `LastUserMsgAt` is unsafe during a reader-first rolling deployment. Once a new broadcast worker creates that field, an old worker can process a later ordinary user message and update only `lastMsgAt`; new readers will continue choosing the stale `lastUserMsgAt`, so the real message does not affect unread state or ordering until another new writer happens to process a user message for that room. This contradicts the claimed either-order rollout safety. Require the writer fleet to be fully upgraded before deploying these readers, or introduce a cutover/version signal that prevents readers from trusting the field while old writers remain.

<details>

```
In `user-service/service/subscriptions.go` around line 256: The unconditional preference for a non-nil `LastUserMsgAt` is unsafe during a reader-first rolling deployment. Once a new broadcast worker creates that field, an old worker can process a later ordinary user message and update only `lastMsgAt`; new readers will continue choosing the stale `lastUserMsgAt`, so the real message does not affect unread state or ordering until another new writer happens to process a user message for that room. This contradicts the claimed either-order rollout safety. Require the writer fleet to be fully upgraded before deploying these readers, or introduce a cutover/version signal that prevents readers from trusting the field while old writers remain.

```

</details>

chat-frontend/src/api/types.ts
```diff
@@ -43,6 +43,11 @@ export interface SubscriptionRoom {
   appCount?: number
   lastMsgAt?: string | null
   lastMsgId?: string
+  /** Last NON-system message time — what unread/sidebar ordering are
+   *  computed from; absent for rooms with no user messages since the
+   *  field shipped (clients fall back to lastMsgAt). Mirrors
+   *  model.SubscriptionRoom.LastUserMsgAt. */
+  lastUserMsgAt?: string | null
```
julianshen: _📝 documentation_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`tgd-review`_

**The field is not always a non-system message timestamp.**

`room-worker/subscriptionRoomFor` deliberately supplies `createdAt` when an added-member payload has no message timestamps, and the system-only Mongo freeze persists the same fallback. The public type and API documentation currently describe this value exclusively as the last non-system message time. Document it as a user-activity reference and also update the corresponding client API field description so consumers do not infer that a real message exists at this timestamp.

<details>

> ‼️ **IMPORTANT**

```suggestion
  /** User-activity reference for unread/sidebar ordering. Usually the last
   * non-system message time; a fresh added-member payload may use createdAt
   * when no message timestamps exist. Absent on legacy rooms, where clients
   * fall back to lastMsgAt. Mirrors model.SubscriptionRoom.LastUserMsgAt. */
  lastUserMsgAt?: string | null
```

</details>

<details>

```
In `chat-frontend/src/api/types.ts` around line 46: `room-worker/subscriptionRoomFor` deliberately supplies `createdAt` when an added-member payload has no message timestamps, and the system-only Mongo freeze persists the same fallback. The public type and API documentation currently describe this value exclusively as the last non-system message time. Document it as a user-activity reference and also update the corresponding client API field description so consumers do not infer that a real message exists at this timestamp.

```

</details>

julianshen (commented):
tGD inline review

hmchangw:
Branch rebased onto `main` and reorganized into 9 logical commits (one per component, dependency-ordered, each verified to compile independently). Review-finding triage below.

## Fixed

- **Preview leak on the encrypted path** — a flagged system event could store `"[encrypted message]"` as the room's sidebar snippet. `previewFromMessage` only inspects the message, and an encrypted system event's synthesized placeholder carries no `type`/`sysMsgData`, so preview now keys off the same `isSystemEvent` gate that already covers unread and ordering. Test added.
- **Mention flags on system events** — both buffer-mode branches applied `evt.hasMention`/`evt.mentionAll` unconditionally, so a system message whose body parses as an `@mention` raised the badge and could be copied into the summary by a later ordinary message. Now preserved for system events. Tests added (including a control that a normal message still sets them).
- **Docs accuracy** — the unread-count section still described the pre-change comparison (`lastMsgAt` vs `lastSeenAt`); corrected to `lastUserMsgAt ?? lastMsgAt`. `lastUserMsgAt` is now documented as a *user-activity reference* rather than always a real message time, since a fresh `subscription.update` `added` payload carries the room's `createdAt`.
- **Separately, from QA:** the creator's own room went bold after switching away. Every subscription was created without `lastSeenAt`, and the actor exemption existed only server-side (applied asynchronously by broadcast-worker once the creation system message flowed), which the client's copy never saw. The actor is now marked pre-read at creation for channels, DMs, botDMs and self-DMs; invited members keep no `lastSeenAt`, which is what makes their room unread.

## Not applied

**Frontend history-ceiling separation** — the design point is right: client-side `lastMsgAt` now carries user-activity semantics, so it shouldn't double as the full-activity history ceiling. Verified this is **latent, not live**: this frontend never sends the `meta` hint on any history request (no `meta` in any `api/` operation, no `lastMsgAt` in any request payload), so no history read can currently be truncated. It's a real trap for whoever adds the hint later, but it's a multi-file semantic split, so raising it here with the patch rather than landing it unreviewed inside this PR:

- `RoomSummary`: add `lastUserMsgAt`, keep `lastMsgAt` as full activity.
- `subToRoom`: pass both through raw instead of collapsing.
- Reducer `MESSAGE_RECEIVED`: always advance `lastMsgAt`; advance `lastUserMsgAt` only for non-system events.
- `isRoomUnread` / `sortByLastMsgDesc`: compare `lastUserMsgAt ?? lastMsgAt`.
- Add a regression: user message at T1, system message at T2, reload — assert the ceiling is T2.

**Monotonic `$max` on `lastUserMsgAt`** — deliberately matching the existing `lastMsgAt` convention, which is unguarded by design (`store_mongo.go`: "a bad future one would freeze ordering"). Guarding one and not its sibling would be inconsistent, and within a worker the coalescer already takes the max per flush window. Worth revisiting for both fields together if out-of-order flushes prove real in practice.

**Mixed-writer rollout** — accurate: an old writer updates only `lastMsgAt`, and readers prefer a non-nil `lastUserMsgAt`, so a stale value can persist through a mixed deploy until a new writer touches the room. The spec documents the rollback case; I'll extend it to cover rolling deploys.

**BSON tags on `RoomEvent.SystemMsg`** — skipped: `RoomEvent` is a wire event, and its sibling flags (`HasMention`, `MentionAll`, `Mentions`) carry no bson tags either. (The `RoomInfo` half was withdrawn in the follow-up analysis — agreed, it's a JSON RPC DTO.)

**"Move DB-backed tests to an integration target"** — not applicable: `user-service/mongorepo/subscriptions_test.go` already begins with `//go:build integration`, so those tests never run under `make test`.

Gates on the pushed head: `make lint` 0 issues, full Go suite green with `-race`, `gosec` pass, frontend 1105 tests + typecheck pass.

---

user-service/mongorepo/subscriptions_test.go
```diff
@@ -926,3 +926,64 @@ func TestGetActiveSubscriptions_ProjectsBadgeFields_Integration(t *testing.T) {
...
+	page, err := r.FindChannelsByMembers(ctx, "alice", []string{"bob"}, mongoutil.OffsetPageRequest{Offset: 0, Limit: 100})
+	require.NoError(t, err)
+	require.Len(t, page.Data, 1)
+
+	enriched := page.Data[0]
+	require.NotNil(t, enriched.LastUserMsgAt, "enriched subscription must have lastUserMsgAt from room")
+	// WithinDuration, not Equal: BSON datetimes are millisecond-precision, so the
+	// round-tripped value loses the seed's sub-millisecond digits.
+	assert.WithinDuration(t, lastUser, *enriched.LastUserMsgAt, time.Millisecond, "lastUserMsgAt must match the room's value")
+}
```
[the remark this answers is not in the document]
hmchangw: Verified against the file — these tests are already in the integration target.

`user-service/mongorepo/subscriptions_test.go` carries `//go:build integration` on line 1, and `newTestSubscriptionRepo` is defined in `setup_test.go`, which is equally tagged. Both tests therefore compile and run only under `make test-integration`; `make test` never builds them (which is also why the unit `test` job is green without Docker).

No change needed.

---

pkg/model/event.go
```diff
@@ -379,6 +379,10 @@ type RoomEvent struct {
 	MentionAll bool          `json:"mentionAll,omitempty"`
 
 	HasMention bool `json:"hasMention,omitempty"`
+	// SystemMsg marks a system-message new_message (IsSystemMessageType) in
+	// plaintext, so clients can exclude it from unread/ordering even when the
+	// message itself is sealed inside EncryptedMessage.
+	SystemMsg bool `json:"systemMsg,omitempty"`
```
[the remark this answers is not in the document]
hmchangw: Declining — `RoomEvent` is a NATS wire event, never persisted, and its payload fields are json-only by the struct's own convention.

Of the fields in `RoomEvent`, only `Timestamp` and `EventTimestamp` carry bson tags. `RoomName`, `RoomType`, `SiteID`, `UserCount`, `LastMsgAt`, `LastMsgID`, `Mentions`, `MentionAll` and `HasMention` are all json-only. `SystemMsg` sits directly alongside `HasMention` and matches it.

The repo guideline targets model structs that round-trip through Mongo. Tagging only the new field here would break the struct's internal consistency without any decoder to serve.

---

pkg/model/room.go
```diff
@@ -90,14 +96,16 @@ type RoomsInfoBatchRequest struct {
...
+	RoomID    string `json:"roomId"`
+	Found     bool   `json:"found"`
+	SiteID    string `json:"siteId,omitempty"`
+	Name      string `json:"name,omitempty"`
+	UserCount int    `json:"userCount,omitempty"`
+	AppCount  int    `json:"appCount,omitempty"`
+	LastMsgAt *int64 `json:"lastMsgAt,omitempty"`
+	LastMsgID string `json:"lastMsgId,omitempty"`
+	// LastUserMsgAt mirrors Room.LastUserMsgAt (epoch ms); nil when absent.
+	LastUserMsgAt     *int64  `json:"lastUserMsgAt,omitempty"`
```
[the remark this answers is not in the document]
hmchangw: Declining — `RoomInfo` is the `rooms.info` RPC DTO, not a persisted model. It is marshalled with `encoding/json` for the reply payload and never round-trips through bson.

Every sibling timestamp on the struct is json-only: `LastMsgAt`, `LastMentionAllAt`, `MinUserLastSeenAt`. `LastUserMsgAt` matches them. (`CrossSite` is the lone bson-tagged field and is vestigial — it is not read from Mongo through this type either.)

The persisted counterpart, `Room.LastUserMsgAt` in the same file, does carry both tags, which is where the guideline applies.

---

docs/client-api.md
```diff
@@ -5597,11 +5605,11 @@ Returns the count of active subscriptions, optionally filtered to unread rooms o
...
+**Caching:** when the server-side badge cache is enabled, the unread count may be served from the account's maintained unread-room set instead of being recomputed. The set is bumped on every **notifiable** message — the same gate that decides whether a message pushes, so system messages (rename, member changes, …) never enter it — and a read removes exactly the room read (a room with unread followed threads stays counted); mute, unmute, thread-read and membership changes invalidate it. The set is re-derived from MongoDB whenever it has gone unverified for longer than the server's badge marker TTL, which is therefore the upper bound on how stale this count can be. Same response schema and staleness bounds as the badge counts carried on push notifications (this count is not capped for display, unlike push counts).
```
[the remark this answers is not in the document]
hmchangw: Declining — there is no contradiction to reconcile.

Line 5586 is the **Active set** definition: which subscriptions are eligible to be counted (non-muted, open, botDMs only when subscribed). It says nothing about where the data is read from, and does not claim the count avoids the room document. Grepping the file for such a claim returns nothing.

The section this finding sits on is the only place `subscription.count` describes its data sources, and it already states both: the local `$lookup` room baseline and the cross-site `GetRoomsMeta` RPC.

The other four findings in this review were valid and are fixed in `a790df9`.

---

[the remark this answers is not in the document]
hmchangw: You're right and my decline was wrong — retracting it.

The sentence is exactly where you said, at the end of line 5588. My grep for it truncated each match at 250 characters, so the tail of that long line never reached me, and I then asserted the claim wasn't in the file. That assertion was false; apologies for the noise.

The statement was written for the active-set membership rule but reads as covering the whole RPC, which is the inconsistency you identified. Fixed in `40af13c` by scoping it and pointing at the unread-count section:

> Membership in the active set is decided from subscription state alone — no room document is consulted, and no room name is filtered. That holds for the whole request when `unread` is absent or `false`; `unread: true` then narrows the set using each room's activity (the local `$lookup` baseline, and `GetRoomsMeta` for cross-site rows) — see **Unread count behavior** below.

---

user-service/service/subscriptions.go
```diff
@@ -637,7 +639,7 @@ func applyRoomInfo(sub *model.Subscription, info *model.RoomInfo) {
 	sub.Room = room
```
julianshen, line 637: _🏷️ protocol-compatibility_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`_

**Mixed-version cross-site RPCs restore system-message unread.**

A new user-service can call an older room-service that omits LastUserMsgAt even when the origin room document contains it. Coalesce then selects LastMsgAt, so a system event at T2 after user activity at T1 marks cross-site members unread and reorders the room until the remote room-service is upgraded. Sequence room-service upgrades across all sites before enabling the new user-service readers, or version the RoomInfo response so absence caused by an old peer can be distinguished from a legacy room that genuinely needs the fallback.

<details>

```
In `user-service/service/subscriptions.go` around line 639: A new user-service can call an older room-service that omits LastUserMsgAt even when the origin room document contains it. Coalesce then selects LastMsgAt, so a system event at T2 after user activity at T1 marks cross-site members unread and reorders the room until the remote room-service is upgraded. Sequence room-service upgrades across all sites before enabling the new user-service readers, or version the RoomInfo response so absence caused by an old peer can be distinguished from a legacy room that genuinely needs the fallback.

```

</details>

docs/client-api.md
```diff
@@ -148,7 +148,8 @@ Each rung includes the ones below it. The header propagates across every service
...
+- **Async reply on `chat.user.{account}.response.{requestID}`** — carries every deferred result: the `msg.send` reply (§4) and the [AsyncJobResult](#asyncjobresult) terminal of the two-phase room operations (Create Room, Add Members, Remove Member, Remove Org, Update Member Role, Rename Room). The request ID comes from a different place per flow: `msg.send` reads `requestId` from its own JSON payload, while a room operation reuses the request's `X-Request-ID` and publishes its `AsyncJobResult` **only when that header was present** — a client that omits it gets the synchronous accept/reject and never the outcome. The client must hold a subscription matching that subject to receive **any** of them — the `chat.user.{account}.>` wildcard is one way, but a narrower subscription works and is what the frontend uses. A client that skips this subscription never learns a room operation's outcome.
```
julianshen: _🏷️ silent-failure_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`silent-failure`_

**Accepted room jobs can fail without any terminal result**

The documented protocol accepts two-phase room operations without `X-Request-ID` but then publishes no `AsyncJobResult`. If the deferred operation later fails, the caller receives only the synchronous acceptance and has no channel for learning the failure, leaving the operation apparently successful or indefinitely pending. Make `X-Request-ID` mandatory for these operations and synchronously reject requests that omit it, or provide another guaranteed terminal response mechanism.

<details>

```
In `docs/client-api.md` around line 151: The documented protocol accepts two-phase room operations without `X-Request-ID` but then publishes no `AsyncJobResult`. If the deferred operation later fails, the caller receives only the synchronous acceptance and has no channel for learning the failure, leaving the operation apparently successful or indefinitely pending. Make `X-Request-ID` mandatory for these operations and synchronously reject requests that omit it, or provide another guaranteed terminal response mechanism.

```

</details>

julianshen (commented):
tGD inline review

user-service/service/subscriptions.go
hmchangw: Confirmed, and this is a distinct window from the writer overlap — thanks, I had only reasoned about broadcast-worker.

The asymmetry that makes it worse: the writer-overlap window self-heals once the room's next user message reaches an upgraded writer. This one does not heal through traffic at all. `rooms.info` is a reader's *only* source of a cross-site room's `lastUserMsgAt`, so while one peer lags, every read of that site's rooms coalesces to `lastMsgAt` and system events keep marking cross-site members unread. It clears only when that room-service is upgraded.

Documented in `2973c16`. The required order is now **broadcast-worker (drained) → room-service at every site → user-service readers → client**, with the mechanism spelled out in the spec's Rollout section.

On versioning the `RoomInfo` response so an old peer is distinguishable from a legacy room: agreed that field-absence alone cannot tell them apart, and I've noted that in the spec. I've stopped short of implementing it here — it's a cross-site protocol addition, and it interacts with an open design question on this PR (whether to keep `lastUserMsgAt` additive at all, or redefine `lastMsgAt` as user-only with a server-private history ceiling). If that goes the other way, the versioning question changes shape. Holding for the author's call rather than landing a wire change unreviewed.

---

docs/client-api.md
hmchangw: You're right that the hole exists, and tracing it corrected a statement I'd pushed earlier in this branch.

I had written that a room operation publishes its `AsyncJobResult` "only when the `X-Request-ID` header was present". That isn't what happens. room-service builds its router with `natsrouter.Default` (`room-service/main.go:344` → `Default` → `RequestID()`), which is the **lenient** middleware: `natsutil.StampRequestID` → `idgen.ResolveRequestID` **mints** a UUID when the inbound value is missing or malformed. `RequireRequestID` exists and its doc comment claims room-service uses it, but it has no production call site — `grep` finds it only inside `pkg/natsrouter` and `pkg/natsutil`.

So the result *is* published — to `chat.user.{account}.response.{server-minted-id}`. Net effect for a client that omits the header:

- Subscribed to its own `chat.user.{account}.response.{requestID}` → never sees the reply. A job that later fails looks indefinitely pending. That's your silent-failure case.
- Subscribed to the `chat.user.{account}.>` wildcard → does receive it, but can only correlate on the payload's `operation`/`roomId`, not the subject.

Fixed the documentation in `2973c16`: the section now states the minting behaviour, names the unguessable subject, and tells clients to always send the header.

**I have not made it mandatory server-side**, and I'd argue that shouldn't ride on this PR. Switching those routes to `RequireRequestID` turns a currently-accepted request into a `BadRequest`/`request_id_required` for any client omitting the header — a protocol break with a client-migration story, in a PR whose subject is system-message unread accounting. It also deserves its own decision on scope: the same lenient middleware covers every `natsrouter.Default` service, so room ops are unlikely to be the only affected two-phase path.

Happy to open that as a separate PR if you want it — the change itself is small (swap the middleware on the two-phase routes, add the reason to the client-API error table), it's the rollout that needs thought.

---

docs/client-api.md
```diff
@@ -5144,7 +5147,13 @@ reorder fans a few-KB event no matter how many chats a section holds.
...
 
 **sortMode**: `"custom"` (sort the section's chats by `subscription.sectionOrder`) or
-`"mostRecent"` (sort by last message — the default and fallback).
+`"mostRecent"` (sort by last **user** activity — the default and fallback; same key as
+[`subscription.list`](#subscriptionlist): `room.lastUserMsgAt` descending, falling back to
+`room.lastMsgAt` for rooms that predate it. Sorting on `lastMsgAt` alone would let a system
+message — a rename, a member change — resurface a dormant room above one with newer real
+conversation. A room with neither timestamp needs no third client-side rung: the server
+already orders it by its `createdAt`, and an `added` payload carries that value in
+`lastUserMsgAt`.)
```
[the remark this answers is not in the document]
hmchangw: Valid — this is an over-claim in text I added two commits ago, and it doesn't survive contact with the reference comparator.

`chat-frontend/src/lib/chatlist.js`:

```js
function lastMsgTime(summary) {
  return summary.lastMsgAt ? new Date(summary.lastMsgAt).getTime() : 0
}
export function sortByLastMsgDesc(rooms) {
  return [...rooms].sort((a, b) => lastMsgTime(b) - lastMsgTime(a))
}
```

A missing timestamp maps to `0`, so a room the server placed at the *top* of a section by its `createdAt` sinks to the bottom of the client's list. My "needs no third client-side rung" was wrong in the direction that matters.

I did not reinstate a `createdAt` rung, because the room object genuinely doesn't expose one — that was the earlier finding this text came from. Fixed the other way in `d26fe34`: the contract now says preserve the server's relative position for such rooms, sort **stably**, and treat two rooms with no activity timestamp as equal, so the server's `createdAt` order survives client-side grouping.

Worth noting how narrow the exposure is, which the doc now states: an `added` payload stamps `lastUserMsgAt`, and a room's own `room_created` system message freezes the field to `createdAt`. The gap is only a room that predates the field and has seen no message since.

---

julianshen (commented):
tGD inline review

broadcast-worker/store_mongo.go
```diff
@@ -155,6 +167,19 @@ func (m *mongoStore) lastMessageUpdate(u *roomLastMsgUpdate) any {
 			fields[k] = bson.M{"$literal": s}
 		}
 	}
+	if systemOnly {
+		// Sticky freeze: pin lastUserMsgAt ONCE to the room's pre-system
+		// position. Every $set expression reads the pre-update document, so
+		// "$lastMsgAt" here is the position BEFORE this write advances it; a
+		// brand-new room (no lastMsgAt yet) pins to its createdAt, which is
```
julianshen: _🎯 correctness_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`tgd-review`_

**The freeze can persist an earlier system message as user activity**

Consider a legacy room whose last real user message was T1 but whose current `lastMsgAt` is a pre-deployment system message at T2. Its first upgraded system-only write at T3 finds no `lastUserMsgAt` and freezes T2 into that field. A member read through T1 is then incorrectly marked unread and the room sorts at T2; subsequent system writes preserve the wrong value indefinitely. Contrary to the design's self-healing claim, the room repairs only after another user message. The initial value must come from an accurate user-message backfill/history lookup or another cutover mechanism that does not infer user activity from an unclassified legacy `lastMsgAt`.

<details>

```
In `broadcast-worker/store_mongo.go` around line 174: Consider a legacy room whose last real user message was T1 but whose current `lastMsgAt` is a pre-deployment system message at T2. Its first upgraded system-only write at T3 finds no `lastUserMsgAt` and freezes T2 into that field. A member read through T1 is then incorrectly marked unread and the room sorts at T2; subsequent system writes preserve the wrong value indefinitely. Contrary to the design's self-healing claim, the room repairs only after another user message. The initial value must come from an accurate user-message backfill/history lookup or another cutover mechanism that does not infer user activity from an unclassified legacy `lastMsgAt`.

```

</details>

room-worker/handler.go
```diff
@@ -1440,6 +1473,7 @@ func subscriptionRoomFor(room *model.Room, pair *roomkeystore.VersionedKeyPair)
 		AppCount:          room.AppCount,
 		LastMsgAt:         room.LastMsgAt,
 		LastMsgID:         room.LastMsgID,
+		LastUserMsgAt:     timeutil.Coalesce(room.LastUserMsgAt, timeutil.Coalesce(room.LastMsgAt, &room.CreatedAt)),
```
julianshen: _🏷️ mongodb-atomicity_ | _🟠 Warning_ | _⚡ Quick fix_ | _`mongodb`_

**Brand-new rooms can appear read before the async freeze**

This seeds `LastUserMsgAt` only in the `subscription.update` payload; it does not persist the value in the `rooms` document. Room/subscription creation and broadcast-worker's later system-message update are separate operations. If an invited member calls `subscription.list` after their subscription becomes visible but before broadcast-worker freezes the room, a fresh room still has neither `lastUserMsgAt` nor `lastMsgAt`, so user-service computes `hasUnread=false`. Persist `lastUserMsgAt=createdAt` when inserting a new room, before exposing its subscriptions, rather than relying solely on the asynchronous Mongo update. Add an integration test that lists the invited member's subscription before broadcast-worker flushes.

<details>

```
In `room-worker/handler.go` around line 1476: This seeds `LastUserMsgAt` only in the `subscription.update` payload; it does not persist the value in the `rooms` document. Room/subscription creation and broadcast-worker's later system-message update are separate operations. If an invited member calls `subscription.list` after their subscription becomes visible but before broadcast-worker freezes the room, a fresh room still has neither `lastUserMsgAt` nor `lastMsgAt`, so user-service computes `hasUnread=false`. Persist `lastUserMsgAt=createdAt` when inserting a new room, before exposing its subscriptions, rather than relying solely on the asynchronous Mongo update. Add an integration test that lists the invited member's subscription before broadcast-worker flushes.

```

</details>

broadcast-worker/store_mongo.go
hmchangw: Verified — you're right, and the design doc's claim was the thing that was wrong.

The freeze is `$ifNull(lastUserMsgAt, lastMsgAt, createdAt, $$REMOVE)`. Before this change `lastMsgAt` was bumped by system messages too, so it is an **unclassified** activity time, not a user-message time. Your T1/T2/T3 trace is exactly right: the first post-deploy system-only write pins T2, a member read through T1 is flagged unread, and the room sorts at T2 until its next user message overwrites the field. The spec said such rooms "self-heal on their first post-deploy system event" — they don't.

One scoping correction I'd offer, because it changes how urgent this is: **it isn't a regression.** Today's unread already compares `lastSeenAt` against that same `lastMsgAt` = T2, so that member reads as unread on this room right now, before the PR. The freeze preserves today's wrong answer rather than introducing a new one. What's untrue is the "no migration needed, it heals itself" framing — the affected population is repaired only by its next user message, and a room that goes quiet keeps the stale reference indefinitely.

Scope of the affected set: legacy rooms whose *newest pre-deploy event was a system message*. Rooms created after the cutover are unaffected — their own `room_created` message freezes `createdAt`, which is a correct floor.

Corrected in `fb7262f`, with the mechanism, the non-regression note, and the remaining limitation all stated.

On the fix itself: an accurate initial value needs a per-room user-message backfill from Cassandra history, which is precisely the migration this design was built to avoid. That's a real trade to make, not one I'll make unilaterally — it's the author's call whether to accept the limitation as scoped above or to add the backfill. Flagging it to them now. If the answer is backfill, I'd want it as its own commit with a dry-run mode, not folded into this one.

---

room-worker/handler.go
hmchangw: Confirmed against the code — the race is real and the window resolves the wrong way.

`unread()` in `user-service/service/subscriptions.go:647` is explicit about it:

```go
// nil ms ⇒ false, nil lastSeen with ms set ⇒ true.
func unread(lastSeen *time.Time, ms *int64) bool {
	if ms == nil {
		return false
	}
	...
```

So a fresh room carrying neither `lastUserMsgAt` nor `lastMsgAt` in the `rooms` document yields `hasUnread=false`, even for a member with no `lastSeenAt`. `subscriptionRoomFor` stamps the value onto the `subscription.update` payload only, so a client that is live at creation gets the right answer while a client doing a cold `subscription.list` inside the window gets the wrong one — and that window is exactly requirement 2, the newly added member.

The fix you describe is right: persist the reference in the room document at insert, not only on the event. Concretely that means seeding `LastUserMsgAt: createdAt` on the `model.Room` literals at room construction — there are four (`handler.go:1418`, `:1560`, `:2038`, `teamsroomcreate.go:60`) — plus the integration test you name, listing the invited member's subscription before broadcast-worker flushes.

I have not pushed it, for one reason: there is an open question on this PR about whether to keep `lastUserMsgAt` additive at all, or redefine `lastMsgAt` as user-only with a server-private history ceiling. Under that alternative the seed is still needed but lands on a different field, so writing it now risks writing it twice. It's queued and small either way — I'll implement it as soon as that's settled, or immediately if the author would rather not wait.

---

broadcast-worker/store_mongo.go
hmchangw: Following up on my own reply — I framed the choice as "accept the limitation or add a backfill". There's a third option, and having worked it through I think it's strictly better than both.

**Don't promote an existing `lastMsgAt` at all. Only pin `createdAt`, and only for a room that has never had a message.**

The current chain is:

```go
$ifNull: [$lastUserMsgAt, $lastMsgAt, $createdAt, $$REMOVE]
```

The `$lastMsgAt` rung is the entire problem: it is the one that launders an unclassified legacy timestamp into a field that claims to mean user activity. Drop it, and pin `createdAt` only when `lastMsgAt` is absent. Then:

| Room | Freeze writes | Reader sees | Requirement 2 |
|---|---|---|---|
| New (no `lastMsgAt`) | `createdAt` — a correct floor | `createdAt` | ✅ member with no `lastSeenAt` reads unread |
| Legacy (has `lastMsgAt`) | nothing — field stays absent | coalesces to `lastMsgAt`, i.e. today's behaviour | ✅ newly added member still unread, because `lastMsgAt` is non-nil |
| Any room, after a real user message | the real timestamp, and the freeze preserves it from then on | correct | ✅ |

So the field is only ever written from something that genuinely *is* user activity, or from `createdAt` where that is definitionally correct. No wrong value is ever persisted, and no backfill is needed. Legacy rooms simply defer to `lastMsgAt` until their next user message classifies them — which is exactly the "explicitly exclude unclassified rooms from the new contract until they are classified" that CodeRabbit's parallel finding proposes, achieved by omission rather than by a migration.

What it does *not* do: repair the T1/T2 case you identified. A member who read through T1 still reads unread against T2. But they read unread against T2 today too, so nothing regresses — and critically, the wrong value is no longer *frozen in*, so the room self-corrects on its next user message instead of being pinned to the wrong answer by later system writes. That is the "self-healing" property my spec claimed and did not have.

Cost is one expression plus tests. I have not pushed it: it changes the mechanism this PR is built on, in response to a blocking finding, so it's the author's call rather than mine. Flagging it to them with a recommendation to take it.

---

broadcast-worker/coalescer.go
```diff
@@ -144,6 +153,10 @@ func (c *coalescingStore) UpdateRoomLastMessage(_ context.Context, upd roomLastM
 		cur.msgID = upd.MsgID
 		cur.at = upd.At
 	}
+	if !upd.SystemMsg && newerRow(upd.At, upd.MsgID, cur.userAt, cur.userMsgID) {
```
mliu33, line 156: Why not just update the original lastMsgAt only if it's a user message ?

broadcast-worker/store_mongo.go
hmchangw: Implemented in 262f3c0 — the `$lastMsgAt` rung is gone.

```
$ifNull($lastUserMsgAt,
        $cond($lastMsgAt is null-or-missing,
              $ifNull($createdAt, $$REMOVE),
              $$REMOVE))
```

- brand-new room (no `lastMsgAt`) → pins `createdAt`, a correct floor; requirement 2 still holds (no `lastSeenAt` + non-nil reference ⇒ unread).
- legacy room → writes nothing. Readers coalesce to `lastMsgAt`, which is exactly the pre-field behavior, and a newly added member is still unread because `lastMsgAt` is non-nil. The room's next user message writes the first accurate value the field ever holds.
- after any user message → the real timestamp, kept by the stickiness from then on.

So no wrong value is persisted and no backfill is needed. What it does not do is repair the T1/T2 member — they read as unread today regardless — but that is now just the fallback, not something frozen in.

Tests: the unit assertion on the update-document shape, plus a new integration case for the legacy room (`lastMsgAt` set, field never written) asserting a system-only write leaves `lastUserMsgAt` unset. The two existing freeze integration tests are unaffected. Spec updated, and room-worker's `subscriptionRoomFor` comment corrected — the value it computes still equals what a reader resolves, but it is no longer what the freeze itself persists.

Leaving this thread open for you to confirm the blocking finding is closed.

---

broadcast-worker/coalescer.go
hmchangw: TL;DR: **System messages disappear from the timeline on reload — intermittently**

Because `lastMsgAt` isn't only the unread/ordering input — it's also history-service's Cassandra read ceiling, and system messages have to stay under it.

`history-service/internal/service/room_times.go:42-46`:

```go
func (s *HistoryService) walkBounds(lastMsgAt, createdAt, now time.Time) (ceiling, floor time.Time) {
	ceiling = lastMsgAt
```

and `history-service/internal/service/messages.go:50-52`:

```go
// Cap before at lastMsgAt+1ms so year-dead rooms become 1-bucket reads instead of walking from now.
if !lastMsgAt.IsZero() && before.After(lastMsgAt) {
	before = lastMsgAt.Add(time.Millisecond)
}
```

So if a system message didn't advance `lastMsgAt`, any system message newer than the room's last user message would sit *above* the ceiling and never come back from the walk. "alice added bob" would appear live and then vanish on reload. That's the whole reason this PR adds a second field instead of just narrowing the existing one — it's the first thing ruled out in the design doc (`docs/superpowers/specs/2026-08-26-system-message-unread-ordering-design.md`, "Alternatives considered": *stop bumping `lastMsgAt`; move history's ceiling to server-clock bounds*).

`lastMsgAt` has two other consumers that also want full activity: the `meta.lastMsgAt` client hint that lets a client skip the room-times lookup, and the preview freshness key (identity-checked against `lastMsgId`).

The alternative that *is* still open is the inverse — redefine `lastMsgAt` as user-only and give history-service a separate, server-private ceiling field. Same two timestamps, opposite naming. I raised it as "Option 1" and it's still an open question on this PR; it's a larger change (every reader and the client contract move) but it leaves the field whose name says "last message" meaning the last *user* message, which is arguably the more honest end state.

---

user-service/service/subscriptions.go
```diff
@@ -594,7 +594,7 @@ func buildLocalRoom(sub *model.EnrichedSubscription) *model.SubscriptionRoom {
 		CrossSite:         sub.CrossSite,
```
julianshen, line 594: _🎯 cassandra-correctness_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`cassandra`, `tgd-review` (2 rules)_

**Legacy history hints can hide trailing system messages**

This changes the client-visible `SubscriptionRoom.lastMsgAt` from the full-activity timestamp to `lastUserMsgAt`. A client deployed against the previous documented contract can receive user activity at T1 followed by a system message at T2, then send T1 as `meta.lastMsgAt` to Load History/Load Next. The history service accepts that valid hint, skips its MongoDB bounds lookup, and caps the Cassandra timeline bucket query at T1, so the T2 system row appears live but disappears after reload. Updating documentation does not protect existing or third-party clients. Preserve the full timestamp in `lastMsgAt` and expose user activity separately, or version/disable the history hint server-side so user-activity timestamps cannot bound Cassandra reads.

<details>

- **`tgd-review`**:597: `buildLocalRoom` now serializes user activity under the existing `room.lastMsgAt` field. Previously deployed clients were explicitly allowed to echo that field as history `meta.lastMsgAt`; the updated API documentation confirms that a valid hint skips the Mongo lookup and bounds the system-inclusive timeline. After user message T1 and system message T2, such a client receives T1, sends it as the ceiling, and omits T2 on reload. Updating documentation or the new frontend cannot protect existing clients. Preserve full activity in the existing field and add a separate user-activity field, or version/disable history hints before changing the field's semantics.

</details>

<details>

```
In `user-service/service/subscriptions.go` around line 594: This changes the client-visible `SubscriptionRoom.lastMsgAt` from the full-activity timestamp to `lastUserMsgAt`. A client deployed against the previous documented contract can receive user activity at T1 followed by a system message at T2, then send T1 as `meta.lastMsgAt` to Load History/Load Next. The history service accepts that valid hint, skips its MongoDB bounds lookup, and caps the Cassandra timeline bucket query at T1, so the T2 system row appears live but disappears after reload. Updating documentation does not protect existing or third-party clients. Preserve the full timestamp in `lastMsgAt` and expose user activity separately, or version/disable the history hint server-side so user-activity timestamps cannot bound Cassandra reads.

```

</details>

broadcast-worker/store_mongo.go
```diff
@@ -145,7 +153,11 @@ func (m *mongoStore) lastMessageUpdate(u *roomLastMsgUpdate) any {
 	if !u.lastMentionAllAt.IsZero() {
 		fields["lastMentionAllAt"] = u.lastMentionAllAt
 	}
-	if !m.previews {
+	systemOnly := u.userAt.IsZero()
```
julianshen, line 156: _🎯 concurrency_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`performance-and-concurrency`, `tgd-review` (2 rules)_

**Concurrent flushes can regress lastUserMsgAt**

Each coalescer orders messages only within its own pending map. If replica A buffers user message T1 and replica B buffers a later message T2 for the same room, B can flush first and set lastUserMsgAt=T2, after which A unconditionally overwrites it with T1. A member whose lastSeenAt lies between T1 and T2 is then incorrectly considered caught up, and the room remains misordered or excluded from activity windows until another user message repairs it. Make lastUserMsgAt monotonic atomically in MongoDB, using $max for plain updates and an equivalent max expression for pipeline updates, and test T2-then-T1 write order.

<details>

- **`tgd-review`**:184: The system-only pipeline pins `lastUserMsgAt` to `createdAt` whenever the pre-update document lacks `lastMsgAt`, regardless of the system-event type. A pre-field room with no recorded messages can therefore receive a rename or membership event and suddenly become unread for every existing member whose `lastSeenAt` is nil. That violates the exception being limited to newly added members. The write needs to distinguish genuine room creation/new-member initialization from an arbitrary system event; supporting only the newly added member may require a per-subscription floor rather than this room-wide inference.

</details>

<details>

```
In `broadcast-worker/store_mongo.go` around line 156: Each coalescer orders messages only within its own pending map. If replica A buffers user message T1 and replica B buffers a later message T2 for the same room, B can flush first and set lastUserMsgAt=T2, after which A unconditionally overwrites it with T1. A member whose lastSeenAt lies between T1 and T2 is then incorrectly considered caught up, and the room remains misordered or excluded from activity windows until another user message repairs it. Make lastUserMsgAt monotonic atomically in MongoDB, using $max for plain updates and an equivalent max expression for pipeline updates, and test T2-then-T1 write order.

```

</details>

room-worker/handler.go
```diff
@@ -1432,13 +1458,23 @@ func (h *Handler) createSelfDM(ctx context.Context, roomID string, requester *mo
...
+	// carried a message, and otherwise leaves the field absent for the reader to
+	// coalesce to lastMsgAt.
 	sr := &model.SubscriptionRoom{
 		SiteID:            room.SiteID,
 		Name:              room.Name,
 		CrossSite:         room.CrossSite,
 		UserCount:         room.UserCount,
 		AppCount:          room.AppCount,
-		LastMsgAt:         room.LastMsgAt,
+		LastMsgAt:         timeutil.Coalesce(room.LastUserMsgAt, timeutil.Coalesce(room.LastMsgAt, &room.CreatedAt)),
```
julianshen, line 1477: _🏷️ distributed-consistency_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`_

**New-room unread state depends on an asynchronous freeze**

The createdAt fallback is only placed in the ephemeral subscription.update payload; the persisted room still starts with both lastUserMsgAt and lastMsgAt unset. Concrete timeline: room-worker commits the room and subscriptions, the invited member misses the added event or calls subscription.list before broadcast-worker processes room_created, and user-service coalesces nil with nil, returning hasUnread=false and no activity timestamp. A later system event does not repair that client state because the reducer deliberately ignores system events for unread activity; delayed or lost canonical processing also leaves the database incorrect. Initialize rooms.lastUserMsgAt to rooms.createdAt in the durable room-creation write before subscriptions become visible. This matches the existing system-only freeze while making creation and its unread floor atomic.

<details>

```
In `room-worker/handler.go` around line 1477: The createdAt fallback is only placed in the ephemeral subscription.update payload; the persisted room still starts with both lastUserMsgAt and lastMsgAt unset. Concrete timeline: room-worker commits the room and subscriptions, the invited member misses the added event or calls subscription.list before broadcast-worker processes room_created, and user-service coalesces nil with nil, returning hasUnread=false and no activity timestamp. A later system event does not repair that client state because the reducer deliberately ignores system events for unread activity; delayed or lost canonical processing also leaves the database incorrect. Initialize rooms.lastUserMsgAt to rooms.createdAt in the durable room-creation write before subscriptions become visible. This matches the existing system-only freeze while making creation and its unread floor atomic.

```

</details>

docs/client-api.md
```diff
@@ -5145,7 +5146,18 @@ reorder fans a few-KB event no matter how many chats a section holds.
 favorite, not a bot). All built-ins default to `sortMode: "mostRecent"`.
 
 **sortMode**: `"custom"` (sort the section's chats by `subscription.sectionOrder`) or
-`"mostRecent"` (sort by last message — the default and fallback).
+`"mostRecent"` (sort by last **user** activity — the default and fallback; same key as
+[`subscription.list`](#subscriptionlist): `room.lastMsgAt` descending, falling back to
```
julianshen: _📝 documentation_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`tgd-review`_

**The sidebar sorting description contradicts the wire contract**

This paragraph says `room.lastMsgAt` falls back to `room.lastMsgAt`, then warns that sorting on `lastMsgAt` alone permits system-event reordering. Under the new documented wire contract, the server has already coalesced its internal timestamps into the single serialized `room.lastMsgAt`, so clients should sort directly on that field. Rewrite this paragraph to separate the server-internal fallback from the client-visible sorting rule.

<details>

```
In `docs/client-api.md` around line 5150: This paragraph says `room.lastMsgAt` falls back to `room.lastMsgAt`, then warns that sorting on `lastMsgAt` alone permits system-event reordering. Under the new documented wire contract, the server has already coalesced its internal timestamps into the single serialized `room.lastMsgAt`, so clients should sort directly on that field. Rewrite this paragraph to separate the server-internal fallback from the client-visible sorting rule.

```

</details>

julianshen (commented):
tGD inline review

broadcast-worker/store_mongo.go
```diff
@@ -145,7 +153,11 @@ func (m *mongoStore) lastMessageUpdate(u *roomLastMsgUpdate) any {
...
+	if !systemOnly {
```
julianshen, line 157: _🎯 concurrency_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`mongodb`, `performance-and-concurrency`, `tgd-review` (3 rules)_

**Concurrent flushes can regress lastUserMsgAt.**

Each replica orders messages only inside its own coalescing window. Replica A can buffer user message T1, replica B can buffer later T2, and B can flush first; A then unconditionally overwrites `lastUserMsgAt` with T1. A member whose `lastSeenAt` lies between T1 and T2 is incorrectly reported as caught up, while ordering and activity-window filtering remain stale until another user message repairs the room. Make this field monotonic atomically in MongoDB: use `$max` in the plain update and an equivalent max expression in pipeline updates, with a test applying T2 before T1.

<details>

- **`mongodb`**:184: The system-only pipeline interprets any room with missing/null `lastMsgAt` as newly created and writes `createdAt` into `lastUserMsgAt`. A legacy empty room can therefore receive an unrelated rename or membership event and become unread for every existing member with no `lastSeenAt`, although no user activity occurred. Persist the floor explicitly during true room creation and have generic system-only updates only preserve an existing value. Supporting a newly added member in a legacy empty room requires per-subscription state rather than a room-wide inference.
- **`tgd-review`**:184: For any room with `createdAt` but no existing `lastMsgAt`, this system-only path pins `lastUserMsgAt` to `createdAt`, regardless of whether the event is room creation or an unrelated later rename/member event. A pre-field empty room whose existing members have no `lastSeenAt` therefore becomes unread despite having no user activity, contradicting the documented legacy behavior. Persist the floor during actual room creation and have arbitrary system-only writes preserve an existing value without inferring one from `createdAt`.

</details>

<details>

```
In `broadcast-worker/store_mongo.go` around line 157: Each replica orders messages only inside its own coalescing window. Replica A can buffer user message T1, replica B can buffer later T2, and B can flush first; A then unconditionally overwrites `lastUserMsgAt` with T1. A member whose `lastSeenAt` lies between T1 and T2 is incorrectly reported as caught up, while ordering and activity-window filtering remain stale until another user message repairs the room. Make this field monotonic atomically in MongoDB: use `$max` in the plain update and an equivalent max expression in pipeline updates, with a test applying T2 before T1.

```

</details>

user-service/service/subscriptions.go
```diff
@@ -594,7 +594,7 @@ func buildLocalRoom(sub *model.EnrichedSubscription) *model.SubscriptionRoom {
...
 		UserCount:         sub.UserCount,
 		AppCount:          sub.AppCount,
-		LastMsgAt:         sub.LastMsgAt,
+		LastMsgAt:         timeutil.Coalesce(sub.LastUserMsgAt, sub.LastMsgAt),
```
julianshen, line 597: _🏷️ nats-request-reply_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`nats`_

**Existing history requests can use a truncated ceiling**

The subscription.list reply now serializes user activity under the existing room.lastMsgAt field. Clients following the previous contract can echo that value into a Load History/Load Next NATS request as meta.lastMsgAt. After user message T1 and system message T2, the request carries T1; history accepts the valid hint, skips its room lookup, and returns a successful but truncated timeline that omits T2. Documentation cannot update deployed clients. Preserve full activity in lastMsgAt and expose user activity separately, or version/disable the history hint before changing this wire field's semantics.

<details>

```
In `user-service/service/subscriptions.go` around line 597: The subscription.list reply now serializes user activity under the existing room.lastMsgAt field. Clients following the previous contract can echo that value into a Load History/Load Next NATS request as meta.lastMsgAt. After user message T1 and system message T2, the request carries T1; history accepts the valid hint, skips its room lookup, and returns a successful but truncated timeline that omits T2. Documentation cannot update deployed clients. Preserve full activity in lastMsgAt and expose user activity separately, or version/disable the history hint before changing this wire field's semantics.

```

</details>

julianshen, line 597: _🏷️ silent-failure_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`silent-failure`_

**Legacy history hints silently omit newer system messages**

This now serializes user activity under the existing `SubscriptionRoom.lastMsgAt` field. After a user message at T1 and a system message at T2, previously deployed clients can echo T1 as the documented history `meta.lastMsgAt` hint. History accepts that valid hint, skips its Mongo lookup, and returns an apparently successful timeline bounded at T1, silently omitting T2. Documentation cannot protect existing clients. Preserve full activity in `lastMsgAt` and expose user activity separately, or version/disable the history hint before changing this field's semantics.

<details>

```
In `user-service/service/subscriptions.go` around line 597: This now serializes user activity under the existing `SubscriptionRoom.lastMsgAt` field. After a user message at T1 and a system message at T2, previously deployed clients can echo T1 as the documented history `meta.lastMsgAt` hint. History accepts that valid hint, skips its Mongo lookup, and returns an apparently successful timeline bounded at T1, silently omitting T2. Documentation cannot protect existing clients. Preserve full activity in `lastMsgAt` and expose user activity separately, or version/disable the history hint before changing this field's semantics.

```

</details>

julianshen (commented):
tGD inline review

mliu33 (approved):
Excellent! Thanks!

PR #388, opened 2026-08-27
fix(upload-service): size server and Drive-upload timeouts for large file uploads

nalydadad:
## Problem

Uploading a large file fails with a misleading `request must be multipart/form-data` 400.

Root cause: the server's `ReadTimeout` was 30s and covers reading the **entire request body**. Against the production 2 GiB ceiling (`FILE_UPLOAD_MAX_FILE_SIZE=2147483648`), any client slower than ~68 MiB/s hits the read deadline mid-body, `c.MultipartForm()` fails, and the handler collapses every parse error into the generic multipart message. Small files pass because they finish within 30s — which is why only large (or concurrent) uploads failed.

## Changes

**upload-service/main.go**
- New `HTTP_TIMEOUT` env config (default **15m**) for the server's `ReadTimeout` and `WriteTimeout`. Sizing basis (documented on the field): `FILE_UPLOAD_MAX_FILE_SIZE / slowest supported client bandwidth` — 15m serves a 2 GiB upload from clients down to ~2.3 MiB/s. One value covers both because `WriteTimeout` starts at end-of-header, so it must span the body read as well as the streamed downloads it already bounded.
- `ReadHeaderTimeout: 5s` keeps the header phase on its own short bound (matching search-service / user-service / pkg/health), so the longer body budget does not become a slowloris window.

**upload-service/handler.go**
- Both `c.MultipartForm()` failure sites now attach the real parse error via `errcode.WithCause`. The cause is logged server-side only (never serialized to the client), so a read deadline, a client disconnect, and a genuinely non-multipart request stop looking identical in the logs — the diagnosability gap that made this bug hard to trace.

**pkg/drive**
- The upload client silently inherited restyutil's 30s default timeout (only the download client had an explicit 5m). Both clients now share `Config.Timeout`, env-tunable as `DRIVE_TIMEOUT` / `LEGACY_DRIVE_TIMEOUT` (default **5m**, zero-guarded in `NewClient`). Sized for the internal service→Drive leg, where 2 GiB moves in tens of seconds; it deliberately does not track the public-leg `HTTP_TIMEOUT`.

## Testing

- `make build SERVICE=upload-service`
- `make test SERVICE=upload-service` and `go test -race ./pkg/drive/` — green
- `gofmt -l upload-service/ pkg/drive/` — clean

### DIFF
--- a/upload-service/main.go
+++ b/upload-service/main.go
@@ -31,6 +31,14 @@ import (
 // dir, so ephemeral storage must fit the concurrent-upload total.
 const maxMultipartMemory = 1 << 20
 
+// httpTimeout bounds both directions of a transfer:
+// FILE_UPLOAD_MAX_FILE_SIZE (2 GiB in production) / slowest supported client
+// bandwidth, so 15m serves clients down to ~2.3 MiB/s. It is also the ceiling on
+// how long a stalled connection holds its spooled temp file. Read and Write are
+// equal because WriteTimeout starts at end-of-header and so must cover the body
+// read as well; the post-read Drive leg is internal and takes seconds.
+const httpTimeout = 15 * time.Minute
+
 type config struct {
 	Port    string `env:"PORT"      envDefault:"8080"`
 	DevMode bool   `env:"DEV_MODE"  envDefault:"false"`
@@ -47,8 +55,8 @@ type config struct {
 
 	Pool mongoutil.PoolConfig
 	// No blanket request timeout here: upload-service streams potentially-large
-	// file downloads bounded by MinioDownloadTimeout and the server WriteTimeout
-	// (both 5m); a short per-request context deadline would cancel those streams.
+	// file downloads bounded by MinioDownloadTimeout and the server WriteTimeout;
+	// a short per-request context deadline would cancel those streams.
 
 	// MaxImages caps the number of images per image-upload request.
 	MaxImages int `env:"MAX_IMAGES" envDefault:"10"`
@@ -180,10 +188,13 @@ func run() error {
 
 	addr := fmt.Sprintf(":%s", cfg.Port)
 	srv := &http.Server{
-		Addr:         addr,
-		Handler:      r,
-		ReadTimeout:  30 * time.Second,
-		WriteTimeout: 5 * time.Minute, // downloads stream potentially-large bodies
+		Addr:    addr,
+		Handler: r,
+		// The header phase keeps a short bound of its own: httpTimeout is sized for
+		// bodies, and inheriting it would let a slowloris hold a connection for 15m.
+		ReadHeaderTimeout: 5 * time.Second,
+		ReadTimeout:       httpTimeout,
+		WriteTimeout:      httpTimeout,
 	}
 
 	srvErr := make(chan error, 1)
--- a/pkg/drive/uploader.go
+++ b/pkg/drive/uploader.go
@@ -16,6 +16,12 @@ import (
 	"github.com/hmchangw/chat/pkg/restyutil"
 )
 
+// driveTimeout is the whole-request ceiling on a Drive transfer. Sized for the
+// internal leg only — 2 GiB moves in tens of seconds over the private network —
+// so it does not track the public client leg's server timeouts. Overrides
+// restyutil's 30s default, which large streamed bodies blow through.
+const driveTimeout = 5 * time.Minute
+
 // MultipartFile is an opened multipart file plus its name, ready to upload.
 type MultipartFile struct {
 	File     multipart.File
@@ -32,8 +38,7 @@ type Client struct {
 }
 
 // NewClient builds a Drive client. Both underlying clients skip TLS verification
-// (the Drive is reached over a private network); the download client uses a
-// 5-minute timeout to allow large streamed bodies.
+// (the Drive is reached over a private network) and share driveTimeout.
 //
 // The upload side keeps only the *http.Client — a deliberate exception to the
 // Resty-for-outbound-HTTP guideline: the bulk upload streams its body, and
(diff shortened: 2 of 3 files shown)


julianshen:
**5 findings · 2 unique issues · 2 inline comments posted** — 🔴 5 blocking

<details>

* `pkg/drive/uploader.go`
* `upload-service/handler.go`
* `upload-service/main.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

upload-service/main.go
```diff
@@ -180,10 +188,13 @@ func run() error {
...
-		Handler:      r,
-		ReadTimeout:  30 * time.Second,
-		WriteTimeout: 5 * time.Minute, // downloads stream potentially-large bodies
+		Addr:    addr,
+		Handler: r,
+		// The header phase keeps a short bound of its own: httpTimeout is sized for
+		// bodies, and inheriting it would let a slowloris hold a connection for 15m.
+		ReadHeaderTimeout: 5 * time.Second,
+		ReadTimeout:       httpTimeout,
+		WriteTimeout:      httpTimeout,
```
julianshen, line 197: _🏷️ reliability_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`distributed-system`, `performance-and-concurrency`, `silent-failure`, `tgd-review` (4 rules)_

**The write deadline can hide a successfully committed upload.**

WriteTimeout starts after the headers and therefore consumes the same 15-minute budget while MultipartForm reads the body. A 2 GiB upload at the documented ~2.3 MiB/s uses about 14m50s, leaving roughly 10 seconds for the Drive transfer even though pkg/drive/uploader.go says that transfer can take tens of seconds. The write deadline does not itself cancel the handler, so Drive can commit the file after the deadline and the subsequent success response write can fail; the client sees a connection timeout and may retry an already-completed upload. Give writes separate budget for the entire body read plus the five-minute Drive timeout and a response margin, or reset the write deadline after the body is parsed.

<details>

- **`performance-and-concurrency`**:197: WriteTimeout is an absolute deadline installed after request headers, so it includes both MultipartForm's body read and all subsequent handler work. A 2 GiB client uploading near the documented minimum bandwidth can consume almost the full 15 minutes, after which the handler may spend up to driveTimeout (5 minutes) transferring the file internally. The response write then occurs after its deadline and fails; the Drive upload may already have succeeded, causing a visible failure and potentially duplicate work when the client retries. Set the write deadline to the maximum body-read duration plus the Drive timeout and a response margin, or apply separate phase/route-specific deadlines.
- **`distributed-system`**:198: WriteTimeout starts after headers and therefore shares the same 15-minute budget as the public body read. A supported 2 GiB upload taking about 14m45s leaves roughly 15 seconds for the documented post-read Drive leg, which can take tens of seconds. Drive can successfully store the object, but the subsequent response write then fails because its deadline has expired, so the caller observes a timeout despite success and may leave an orphan or retry the upload. Set the write deadline to at least the inbound-body budget plus the Drive request ceiling and a response margin (or otherwise give the post-read operation a separate deadline).
- **`tgd-review`**:198: WriteTimeout starts after the request headers and therefore covers both reading the upload body and the subsequent Drive upload. A 2 GiB body at the claimed ~2.3 MiB/s takes about 14m50s, leaving roughly 10 seconds even though the Drive leg is documented as taking tens of seconds and may run for up to five minutes. The Drive upload can complete after the write deadline, leaving the server unable to return a response and causing clients to observe failure or retry a successful upload. Keep ReadTimeout as the body-read ceiling, but give WriteTimeout an additional explicit allowance for the Drive leg and response.

</details>

<details>

```
In `upload-service/main.go` around line 197: WriteTimeout starts after the headers and therefore consumes the same 15-minute budget while MultipartForm reads the body. A 2 GiB upload at the documented ~2.3 MiB/s uses about 14m50s, leaving roughly 10 seconds for the Drive transfer even though pkg/drive/uploader.go says that transfer can take tens of seconds. The write deadline does not itself cancel the handler, so Drive can commit the file after the deadline and the subsequent success response write can fail; the client sees a connection timeout and may retry an already-completed upload. Give writes separate budget for the entire body read plus the five-minute Drive timeout and a response margin, or reset the write deadline after the body is parsed.

```

</details>

upload-service/main.go
```diff
@@ -180,10 +188,13 @@ func run() error {
...
-		Addr:         addr,
-		Handler:      r,
-		ReadTimeout:  30 * time.Second,
-		WriteTimeout: 5 * time.Minute, // downloads stream potentially-large bodies
+		Addr:    addr,
+		Handler: r,
+		// The header phase keeps a short bound of its own: httpTimeout is sized for
+		// bodies, and inheriting it would let a slowloris hold a connection for 15m.
+		ReadHeaderTimeout: 5 * time.Second,
+		ReadTimeout:       httpTimeout,
```
julianshen, line 196: _⚡ memory-exhaustion_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`oom`_

**The 15-minute body window enables multipart memory pinning**

A client can send multipart headers plus roughly 10 MiB of non-file fields and then stall before the terminating boundary. Go's ParseMultipartForm retains those values while waiting for the rest of the body (its non-file allowance is maxMultipartMemory plus 10 MiB), so every in-progress request can pin about 11 MiB for the new 15-minute ReadTimeout. Because http.Server has no in-flight request or connection bound here, a modest sustained stream of such requests can accumulate enough retained forms and goroutines to cause GC thrashing, container eviction, or process OOM; ReadHeaderTimeout does not help after headers are complete. Add a bounded, nonblocking in-flight upload gate before MultipartForm parsing, rejecting excess requests, or enforce a body-rate/connection limit at the listener or trusted proxy while preserving the legitimate large-upload timeout.

<details>

```
In `upload-service/main.go` around line 196: A client can send multipart headers plus roughly 10 MiB of non-file fields and then stall before the terminating boundary. Go's ParseMultipartForm retains those values while waiting for the rest of the body (its non-file allowance is maxMultipartMemory plus 10 MiB), so every in-progress request can pin about 11 MiB for the new 15-minute ReadTimeout. Because http.Server has no in-flight request or connection bound here, a modest sustained stream of such requests can accumulate enough retained forms and goroutines to cause GC thrashing, container eviction, or process OOM; ReadHeaderTimeout does not help after headers are complete. Add a bounded, nonblocking in-flight upload gate before MultipartForm parsing, rejecting excess requests, or enforce a body-rate/connection limit at the listener or trusted proxy while preserving the legitimate large-upload timeout.

```

</details>

julianshen (commented):
tGD inline review

mliu33 (approved):

PR #337, opened 2026-08-21
perf(o11y): cut the NATS metrics down to what is read, and gate the next ones

flywindy:
Follow-up to the review of #286, which asked two things: **do these metrics cost us on the hot path**, and **do we actually need them**. This answers both by deleting rather than explaining, then adds gates so the next metric has to answer them before it lands.

## At a glance

| | |
|---|---|
| **Instruments** | **7 removed, 3 added** — net −4 |
| **Hot path** | disabled metrics 89 ns/2 allocs → **3.27 ns/0 allocs**; `Publisher` construction 8,094 → **8** allocations |
| **Cardinality** | one unbounded label fixed (`roomId`, one series per room) |
| **Standards** | request/reply moved to the OTel RPC semantic conventions |
| **New gates** | 2 blocking — a semgrep ruleset and a registry test, both themselves tested |
| **Size** | 67 files, +4,575 / −912, 63 commits |

## Suggested review order

It is a big diff, but most of it is docs and gates. The order below is roughly decreasing importance.

| # | What | Where | Size |
|---|---|---|---|
| 1 | **The core change** — deletions, lazy attribute sets, RPC semconv | `pkg/natsmetrics/` | 14 files, +1,160 / −231 |
| 2 | **The gates** — what stops this recurring | `.semgrep/metrics.yml` + `.go` fixtures, `pkg/obs/instrument_registry_test.go` | 3 files, +1,296 |
| 3 | **The contract** — every instrument next to what reads it | `docs/specs/o11y/nats-metrics-contract.md` | new file |
| 4 | **Service wiring** — thin, repetitive, skimmable | 32 service files | +614 / −161 |
| 5 | Everything else | `pkg/roomkeymetrics`, `pkg/natsutil`, `pkg/natsrouter`, `Makefile` | +510 |

If you only read one thing, read **`docs/specs/o11y/nats-metrics-contract.md` §13** — the instrument registry with its "Read by" column. It is the artifact this PR exists to create.

## What was removed, and why

| Instrument | Why it is gone |
|---|---|
| `chat.nats.publish.retries` | **No producer anywhere in the repo.** No path loops around its own publish, so it could never leave zero. |
| `chat.nats.requests`, `chat.nats.request.handled` | Each shadowed a histogram. `rpc_*_call_duration_seconds_count` is the same number on one fewer series. |
| `chat.nats.request.duration`, `chat.nats.request.handler.duration` | Renamed to the OTel convention (below), not deleted. |
[...]

PR #390, opened 2026-08-27
Support HTTP Basic auth for authenticating proxies

yenta:
## Summary

Add support for HTTP Basic authentication to proxies via separate `ProxyUsername` and `ProxyPassword` configuration fields. This allows deployments using authenticating proxies to avoid embedding credentials in the proxy URL, where URL metacharacters in passwords would require percent-encoding and where the secret would be mixed with non-secret configuration.

## Key Changes

- **Config struct**: Add `ProxyUsername` and `ProxyPassword` fields to `msgraph.Config` (separate from `ProxyURL`)
- **Proxy application**: Refactor `applyProxyURL` → `applyProxy` to accept the full `Config` and apply credentials as Basic auth on the transport
- **Constructor signatures**: Update `NewDirectoryClient` and `NewGroupReaderClient` to return `(T, error)` instead of `T`, enabling validation at construction time
- **Validation**: Fail fast at client construction if credentials are set without a proxy URL, or if a password is set without a username
- **Precedence**: Explicit `ProxyUsername`/`ProxyPassword` override any userinfo embedded in `ProxyURL`
- **Backward compatibility**: Deployments already using embedded userinfo in `ProxyURL` continue to work unchanged
- **Coverage**: All error-returning constructors (`NewMeetingsClient`, `NewChatsClient`, `NewChatMembersClient`, `NewUserListerClient`, `NewMeetingsDirectoryClient`, `NewDirectoryClient`, `NewGroupReaderClient`, `NewPresenceClient`) now honor proxy credentials
- **Environment variables**: Add `GRAPH_PROXY_USERNAME` and `GRAPH_PROXY_PASSWORD` to all services' config structs and docker-compose files
- **Tests**: Comprehensive test suite covering Basic auth transmission, special characters in passwords, credential precedence, backward compatibility, validation errors, and password redaction in error messages

## Implementation Details

- Credentials are applied to the HTTP transport's proxy function, so they are sent on both the token request and Graph request (including through CONNECT tunnels)
- Password is never logged — validation errors redact it
- The separate fields allow passwords with `@ : / ? # %` to be used verbatim without percent-encoding
- Rotating a password now requires changing one secret rather than a full connection string

https://claude.ai/code/session_01K2wpgBxSSUf55WQbo1aevP

## Summary by CodeRabbit

* **New Features**
  * Added optional Microsoft Graph proxy configuration with URL, username, and password environment variables.
  * Added Basic authentication for HTTP(S) proxies and authenticated SOCKS5 proxy support.
  * Explicit Graph proxy settings take precedence over standard proxy variables and embedded URL credentials.
* **Bug Fixes**
  * Invalid configurations now fail fast with clear, password-safe errors.
  * Added warnings when proxy credentials are sent over unencrypted connections.
* **Documentation**
  * Documented authentication, precedence, fallback behavior, supported schemes, and configuration requirements.


yenta:
All six review findings verified against the code and addressed in 2813be8. One was a real leak.

**The leak.** `url.Parse` returns a `*url.Error` whose `Error()` embeds the raw input, so a malformed `GRAPH_PROXY_URL` carrying userinfo put the proxy password into the returned error — and from there into the server log via `errcode.Classify`. I reproduced it before fixing:

```
parse "://user:sup3rs3cr3t@proxy.corp:8080": missing protocol scheme
```

`Redacted()` cannot cover this path; it needs a URL that parsed. The fix keeps the underlying reason (`uerr.Err` — "missing protocol scheme", no URL) and drops the value. Three malformed-URL cases now assert the password is absent; they failed before the fix.

Worth naming why this went unnoticed: the existing leak test only covered credentials passed as separate fields, which is exactly the case the new settings encourage. The embedded-userinfo path is the legacy one, and it was the unguarded one.

**The rest:**

| Finding | Action |
|---|---|
| Unsupported proxy schemes accepted | Rejected at construction. Confirmed against `transport.go:119-121` that `net/http` proxies only `http`, `https`, `socks5`, `socks5h`. Positive and negative tests added. |
| `applyProxy` error not wrapped | Wrapped in all **eight** constructors, not just `NewChatsClient` — fixing one would have left the package inconsistent, and every call site is already in this diff. |
| Test helpers `panic` | Take `*testing.T`, use `t.Helper()` + `require.NoError`. |
| Test env not isolated | `os.Unsetenv` never restored; now `t.Setenv(k, "")`, which restores on cleanup and reads back as unset (verified: `env` falls back to the empty `envDefault`). Each `setRequiredEnv` also clears the three vars so default assertions don't depend on the host env. |
| Fallback docs imprecise | Correct, and the imprecise sentence was mine. "No Graph traffic depends on `HTTPS_PROXY`" holds only where `GRAPH_PROXY_URL` is set; empty still falls back. Fixed in the docs, the config comment, and all six compose files. Also noted that a `GRAPH_PROXY_URL` embedding userinfo is itself a secret. |

**Not doing: Docstring Coverage (54.41% < 80%).** That check counts test functions, which here are named descriptively and carry comments explaining what each one guards. Adding boilerplate docstrings to reach the threshold would add noise without adding information. Happy to revisit if a maintainer wants it.

Local verification on this head: `make test` exit 0 (116 packages, 0 failures), `make lint` 0 issues, `make sast-gosec` pass, repo `.semgrep/` rules clean on the changed packages, `go vet -tags integration ./...` clean.

---

yenta:
CI is green on `2813be8` (11/11 checks), and all six CodeRabbit findings are addressed and their threads resolved.

One of them was a real leak worth calling out for reviewers: `url.Parse` puts the raw input in its `*url.Error`, so a **malformed** `GRAPH_PROXY_URL` carrying embedded userinfo would have put the proxy password into the returned error — and from there into the server log via `errcode.Classify`. `Redacted()` cannot cover that path, since it needs a URL that parsed. `applyProxy` now returns only the underlying `*url.Error.Err` ("missing protocol scheme" and friends), which keeps the diagnostic without the value. Three malformed-URL cases are pinned by `TestProxyCredentials_MalformedURLNeverLeaksEmbeddedPassword`, each confirmed failing before the fix.

**Not doing: the Docstring Coverage pre-merge check (54.41% vs 80%).** Every exported function in the 13 non-test files this PR touches already has a doc comment, and `golangci-lint` is clean. The gap is test functions, which Go convention does not require doc comments for — of the functions in this diff, 131 are test functions against 62 non-test ones, and the check counts both. Adding boilerplate docstrings to table-driven test cases to clear the threshold would add noise without adding information, so I left it. Happy to revisit if the repo wants that threshold enforced on tests.

---

julianshen:
**No actionable comments.** ✅

<details>

* `docs/msgraph-client.md`
* `pkg/msgraph/chats.go`
* `pkg/msgraph/groups.go`
* `pkg/msgraph/groups_test.go`
* `pkg/msgraph/members.go`
* `pkg/msgraph/msgraph.go`
* `pkg/msgraph/msgraph_test.go`
* `pkg/msgraph/presence.go`
* `room-service/deploy/docker-compose.yml`
* `room-service/main.go`
* `teams-chat-member-sync/deploy/docker-compose.yml`
* `teams-chat-member-sync/main.go`
* `teams-chat-member-sync/main_test.go`
* `teams-chat-sync/deploy/docker-compose.yml`
* `teams-chat-sync/main.go`
* `teams-chat-sync/main_test.go`
* `teams-hr-sync/config.go`
* `teams-hr-sync/config_test.go`
* `teams-hr-sync/deploy/docker-compose.yml`
* `teams-hr-sync/integration_test.go`
* `teams-hr-sync/main.go`
* `teams-user-sync/config.go`
* `teams-user-sync/config_test.go`
* `teams-user-sync/deploy/docker-compose.yml`
* `teams-user-sync/main.go`
* `tools/graphmock/server_test.go`
* `user-presence-service/sync/config_test.go`
* `user-presence-service/sync/deploy/docker-compose.yml`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

julianshen (commented):
tGD inline review

pkg/msgraph/msgraph.go
```diff
@@ -331,23 +346,54 @@ func mutableTransport(hc *http.Client) *http.Transport {
...
 	}
 	if proxyURL.Scheme == "" || proxyURL.Host == "" {
 		// Redacted() masks any embedded proxy credentials before it reaches logs.
 		return fmt.Errorf("invalid graph proxy url %q: scheme and host are required", proxyURL.Redacted())
 	}
+	if !supportedProxySchemes[proxyURL.Scheme] {
+		// net/http would accept this here and fail on the first request instead.
+		return fmt.Errorf("unsupported graph proxy scheme %q: want http, https, socks5 or socks5h", proxyURL.Scheme)
+	}
+	if cfg.ProxyUsername != "" {
```
julianshen, line 425: _🏷️ failure-handling_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`_

**Colon-containing Basic usernames fail only on the first request**

For an HTTP(S) proxy, Go constructs Basic credentials as `username:password`. If the configured or embedded username contains `:`, the proxy splits at the first colon and receives a different username/password pair. The client still constructs successfully; the first token request then receives 407, and every Graph operation fails. After resolving the effective userinfo, reject `:` in usernames for `http`/`https` schemes (while retaining it for SOCKS authentication), and add coverage for both explicit and embedded usernames.

<details>

```
In `pkg/msgraph/msgraph.go` around line 394: For an HTTP(S) proxy, Go constructs Basic credentials as `username:password`. If the configured or embedded username contains `:`, the proxy splits at the first colon and receives a different username/password pair. The client still constructs successfully; the first token request then receives 407, and every Graph operation fails. After resolving the effective userinfo, reject `:` in usernames for `http`/`https` schemes (while retaining it for SOCKS authentication), and add coverage for both explicit and embedded usernames.

```

</details>

pkg/msgraph/msgraph.go
```diff
@@ -331,23 +346,54 @@ func mutableTransport(hc *http.Client) *http.Transport {
...
 	if err != nil {
-		return fmt.Errorf("parse graph proxy url: %w", err)
+		// *url.Error carries the raw input, so returning it verbatim would leak a
+		// password embedded as userinfo. Redacted() can't help here — it needs a
+		// URL that parsed. Keep the underlying reason, drop the value.
+		var uerr *url.Error
+		if errors.As(err, &uerr) {
+			return fmt.Errorf("parse graph proxy url: %w", uerr.Err)
+		}
+		return errors.New("parse graph proxy url: malformed value")
```
julianshen: _🔒 security_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**Malformed escape errors can reveal proxy password bytes**

Unwrapping `url.Error.Err` does not guarantee that user input is removed. For example, parsing `http://user:%zz@proxy.corp` produces an underlying `invalid URL escape "%zz"` error, exposing password bytes in the returned construction error and consequently in logs. The existing tests miss malformed escapes. Return the generic malformed-value error for every parse failure.

<details>

> ‼️ **IMPORTANT**

```suggestion
		return errors.New("parse graph proxy url: malformed value")
```

</details>

<details>

```
In `pkg/msgraph/msgraph.go` around line 377: Unwrapping `url.Error.Err` does not guarantee that user input is removed. For example, parsing `http://user:%zz@proxy.corp` produces an underlying `invalid URL escape "%zz"` error, exposing password bytes in the returned construction error and consequently in logs. The existing tests miss malformed escapes. Return the generic malformed-value error for every parse failure.

```

</details>

docs/msgraph-client.md
```diff
@@ -61,6 +63,44 @@ room-service constructs this client via `NewMeetingsClient(cfg)`, which honors
 `Config.ProxyURL` (from `GRAPH_PROXY_URL`) and fails fast on a malformed proxy
 value at startup.
 
+### Authenticating proxies
+
+Set `GRAPH_PROXY_USERNAME`/`GRAPH_PROXY_PASSWORD` alongside `GRAPH_PROXY_URL`
+and the client sends `Proxy-Authorization: Basic` on every hop — the token
+request and the Graph request alike, through the CONNECT tunnel. Credentials
```
julianshen: _📝 documentation_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**The Basic-auth claim is incorrect for SOCKS proxies**

The documentation says these credentials always produce `Proxy-Authorization: Basic`, while the same section and `supportedProxySchemes` explicitly allow `socks5` and `socks5h`. Go uses URL userinfo for the SOCKS5 username/password handshake for those schemes; it does not send an HTTP proxy-authorization header. Clarify that Basic applies only to `http`/`https` proxies and describe SOCKS5 authentication separately, or reject authenticated SOCKS URLs if Basic is the intended contract.

<details>

```
In `docs/msgraph-client.md` around line 70: The documentation says these credentials always produce `Proxy-Authorization: Basic`, while the same section and `supportedProxySchemes` explicitly allow `socks5` and `socks5h`. Go uses URL userinfo for the SOCKS5 username/password handshake for those schemes; it does not send an HTTP proxy-authorization header. Clarify that Basic applies only to `http`/`https` proxies and describe SOCKS5 authentication separately, or reject authenticated SOCKS URLs if Basic is the intended contract.

```

</details>

pkg/msgraph/msgraph.go
yenta: Correct, and this one undercut my earlier fix. I had assumed `*url.Error.Err` never quotes the input; it does. Reproduced before fixing:

```
url.Parse("http://user:%zzsecret@proxy.corp")
  full err : parse "http://user:%zzsecret@proxy.corp": invalid URL escape "%zz"
  uerr.Err : invalid URL escape "%zz"
```

So the unwrap still leaked the start of the password. Taking the suggestion: every parse failure now returns `malformed value in GRAPH_PROXY_URL` with nothing from the input attached. Three malformed-escape cases (password start, password interior, username) now assert the fragment is absent — all three failed before the change. Fixed in ab47f10.

---

pkg/msgraph/msgraph.go
yenta: Confirmed against the transport: `proxyAuth()` builds `base64(username + ":" + password)`, so `username="corp:svc"` becomes the payload `corp:svc:pw` and the proxy reads user=`corp`, pass=`svc:pw` — 407 on the first token request, with construction having succeeded. RFC 7617 forbids the colon in the user-id for exactly this.

Rejected at construction now for `http`/`https`, covering both routes you named: the explicit `GRAPH_PROXY_USERNAME`, and a percent-encoded colon smuggled through URL userinfo (`http://corp%3Asvc:pw@proxy.corp` — `url.Parse` decodes it back into `Username()`). The rejection message carries the scheme, never the password.

Your SOCKS caveat is right and I kept it: RFC 1929 uses length-prefixed fields, so a colon is ordinary data there. A regression test pins that `socks5`/`socks5h` still accept `corp:svc`. Fixed in ab47f10.

---

docs/msgraph-client.md
yenta: Right, and the inconsistency was mine — the "Basic on every hop" sentence predated `socks5`/`socks5h` being added to `supportedProxySchemes` in the previous commit, and I didn't revisit it. Confirmed in `transport.go:1803`: the SOCKS branch hands the userinfo to `socksUsernamePassword.Authenticate` (RFC 1929 sub-negotiation) and sends no `Proxy-Authorization` header.

Took the "clarify" option rather than rejecting authenticated SOCKS URLs, since the schemes genuinely work — the docs and the `Config` comment now split the mechanism by scheme instead of claiming Basic everywhere. Fixed in ab47f10.

---

teams-user-sync/config_test.go
```diff
@@ -16,6 +16,11 @@ func setRequiredEnv(t *testing.T) {
 	t.Setenv("GRAPH_CLIENT_SECRET", "secret")
 	t.Setenv("MONGO_READ_URI", "mongodb://read:27017")
 	t.Setenv("MONGO_WRITE_URI", "mongodb://write:27017")
+	// Cleared so the default-value assertions don't depend on the host env; the
+	// proxy tests set them explicitly.
+	t.Setenv("GRAPH_PROXY_URL", "")
+	t.Setenv("GRAPH_PROXY_USERNAME", "")
+	t.Setenv("GRAPH_PROXY_PASSWORD", "")
```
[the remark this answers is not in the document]
yenta: Agreed — the tests are named for the default, so they should exercise the absent variable, not an explicitly empty one. They read back the same here only because these fields carry an empty `envDefault`, which makes the coverage accidental rather than intended.

Fixed in 30db6fb with a small `unsetProxyEnv` helper: `os.Unsetenv` for the duration of the test, prior value restored on cleanup (`t.Setenv` has no unset counterpart). Applied to all five services that read the vars, not just the three flagged, so the pattern doesn't diverge.

Verified the helper actually isolates rather than passing by accident — re-ran each package with the vars set in the environment:

```
GRAPH_PROXY_URL=http://host.proxy:9999 GRAPH_PROXY_USERNAME=hostuser \
GRAPH_PROXY_PASSWORD=hostpw go test -race -count=1 ./<pkg>/...
```

All five still pass; without the helper the default assertions would read the host values.

---

pkg/msgraph/msgraph.go
```diff
@@ -331,23 +348,65 @@ func mutableTransport(hc *http.Client) *http.Transport {
...
-		return fmt.Errorf("parse graph proxy url: %w", err)
+		// Nothing from a failed parse may reach the log: *url.Error quotes the
+		// whole input, and its underlying error quotes fragments of it too
+		// (`invalid URL escape "%zz"` would expose the start of a password).
+		// Redacted() can't help — it needs a URL that parsed.
+		return errors.New("parse graph proxy url: malformed value in GRAPH_PROXY_URL")
 	}
 	if proxyURL.Scheme == "" || proxyURL.Host == "" {
 		// Redacted() masks any embedded proxy credentials before it reaches logs.
 		return fmt.Errorf("invalid graph proxy url %q: scheme and host are required", proxyURL.Redacted())
```
julianshen: _🔒 security_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`distributed-system`, `tgd-review` (2 rules)_

**Malformed proxy URLs can still leak embedded passwords.**

For `GRAPH_PROXY_URL=user:supersecret@proxy.corp:8080`, `url.Parse` succeeds by treating `user` as the scheme and the remainder as opaque data. `proxyURL.User` is therefore nil, so `Redacted()` does not mask `supersecret`; this error includes the password and is subsequently wrapped for service logs. Return a static error from the scheme/host validation instead of interpolating any parsed URL.

<details>

- **`distributed-system`**:402: With `ProxyURL` set to `http://:secret@proxy:8080` and the explicit credential fields empty, parsing succeeds, but the earlier password-without-username check examines only `cfg.ProxyPassword`. This check then accepts the empty username, so construction succeeds and net/http sends Basic credentials for `:secret`; an authenticating proxy rejects the first token or Graph request with 407. Validate parsed userinfo after applying explicit credentials so this configuration fails during startup as documented.

<details>

```text
	if proxyURL.User != nil {
		username := proxyURL.User.Username()
		if password, ok := proxyURL.User.Password(); ok && password != "" && username == "" {
			return errors.New("graph proxy password set without a username")
		}
		// Basic sends "user:password", so an HTTP(S) proxy splits a colon-bearing
		// username at the wrong place and answers 407 on the first request — RFC 7617
		// forbids the colon for that reason. SOCKS5 negotiates length-prefixed fields
		// (RFC 1929), where a colon is ordinary data.
		if basicAuthProxySchemes[proxyURL.Scheme] && strings.Contains(username, ":") {
			return fmt.Errorf("invalid graph proxy username for a %s proxy: must not contain ':'", proxyURL.Scheme)
		}
	}
```

</details>


</details>

<details>

```
In `pkg/msgraph/msgraph.go` around line 393: For `GRAPH_PROXY_URL=user:supersecret@proxy.corp:8080`, `url.Parse` succeeds by treating `user` as the scheme and the remainder as opaque data. `proxyURL.User` is therefore nil, so `Redacted()` does not mask `supersecret`; this error includes the password and is subsequently wrapped for service logs. Return a static error from the scheme/host validation instead of interpolating any parsed URL.

```

</details>

pkg/msgraph/msgraph.go
```diff
@@ -331,23 +348,65 @@ func mutableTransport(hc *http.Client) *http.Transport {
...
+	proxyURL, err := url.Parse(cfg.ProxyURL)
 	if err != nil {
-		return fmt.Errorf("parse graph proxy url: %w", err)
+		// Nothing from a failed parse may reach the log: *url.Error quotes the
+		// whole input, and its underlying error quotes fragments of it too
+		// (`invalid URL escape "%zz"` would expose the start of a password).
+		// Redacted() can't help — it needs a URL that parsed.
+		return errors.New("parse graph proxy url: malformed value in GRAPH_PROXY_URL")
 	}
 	if proxyURL.Scheme == "" || proxyURL.Host == "" {
```
julianshen: _🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**A proxy port without a hostname passes validation.**

`url.Parse("http://:8080")` produces a non-empty `Host` (`:8080`) but an empty hostname, so this check passes and the constructor succeeds despite the documented host requirement. Requests then attempt to dial an empty/local host instead of failing at startup. Validate `proxyURL.Hostname() == ""` rather than only `proxyURL.Host == ""`.

<details>

```
In `pkg/msgraph/msgraph.go` around line 391: `url.Parse("http://:8080")` produces a non-empty `Host` (`:8080`) but an empty hostname, so this check passes and the constructor succeeds despite the documented host requirement. Requests then attempt to dial an empty/local host instead of failing at startup. Validate `proxyURL.Hostname() == ""` rather than only `proxyURL.Host == ""`.

```

</details>

julianshen (commented):
tGD inline review

pkg/msgraph/msgraph.go
yenta: Confirmed, and this is the third shape of this leak — the one that shows my assumption about `Redacted()` was wrong from the start. The first two were parse *failures*; this one parses cleanly:

```
url.Parse("user:supersecret@proxy.corp:8080")
  Scheme="user"  Host=""  Opaque="supersecret@proxy.corp:8080"  User=nil
  Redacted() = "user:supersecret@proxy.corp:8080"
```

`Redacted()` masks userinfo only when `url.Parse` populated `User`. Without a scheme the whole value lands in `Opaque` and `User` is nil, so it masks nothing and handed the password to the scheme/host rejection, which `errcode.Classify` then logs.

Conclusion I should have reached two rounds ago: there is no safe way to echo a bad proxy URL. Every rejection in `applyProxy` is now static, the sole exception being the scheme name in the unsupported-scheme error, which carries no credential.

Your `distributed-system` note is also correct and fixed in the same commit: the password-without-username check read only `cfg.ProxyPassword`, so `http://:secret@proxy:8080` passed and Basic went on to send `:secret`. Userinfo is now validated *after* the explicit credentials are applied, so it covers both sources, and the colon check moved into that same block rather than testing `proxyURL.User` separately — which is what your proposed fix does. Fixed in f9ef3ae.

---

pkg/msgraph/msgraph.go
yenta: Right — verified:

```
url.Parse("http://:8080")  ->  Host=":8080" (non-empty)   Hostname()="" (empty)
```

So the `Host == ""` test passed it through and the failure moved to the first dial, which is exactly what the fail-fast validation exists to prevent. Now checks `Hostname()`. Covered for `http`, `https` and `socks5` in `TestApplyProxy_RejectsPortWithoutHostname`, all three confirmed failing before the change. Fixed in f9ef3ae.

---

pkg/msgraph/msgraph.go
```diff
@@ -331,22 +348,76 @@ func mutableTransport(hc *http.Client) *http.Transport {
...
+			// Basic would send ":password"; the proxy answers 407 on the first request.
+			return errors.New("graph proxy password set without a username")
+		}
+		// Basic sends "user:password", so an HTTP(S) proxy splits a colon-bearing
+		// username at the wrong place and answers 407 — RFC 7617 forbids the colon
+		// for that reason. SOCKS5 negotiates length-prefixed fields (RFC 1929),
+		// where a colon is ordinary data.
+		if basicAuthProxySchemes[proxyURL.Scheme] && strings.Contains(username, ":") {
+			return fmt.Errorf("invalid graph proxy username for a %s proxy: must not contain ':'", proxyURL.Scheme)
+		}
```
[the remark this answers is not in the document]
yenta: The finding is correct and I have not taken the proposed fix. Flagging that explicitly rather than resolving, because the call belongs to a human.

The mechanism is exactly as described: Basic is base64 and RFC 1929 sends the password in the clear, and both cross to the proxy before the CONNECT tunnel exists. The TLS that protects Graph traffic runs *inside* that tunnel, so it does not cover the credential hop. Only an `https://` proxy encrypts it.

What I did not do is reject credentialed `http`/`socks5`, or gate them behind an insecure opt-in. That would break the deployment this setting was added for: the requirement here is an authenticating corporate proxy, and those are overwhelmingly plain `http`. Whether an `https://` proxy endpoint exists is a decision for whoever operates the proxy — the client cannot impose it, and failing construction would leave them with no working configuration at all.

Instead, following this repo's existing handling of `TEAMS_TLS_INSECURE`: warn once at construction, and document the exposure so the choice is informed. `docs/msgraph-client.md` now states the CWE plainly and names the two ways out — an `https://` proxy URL, or a trusted network segment between service and proxy. The warning carries the scheme and proxy host only; tests assert neither username nor password appears in it, that `https` stays silent, and that an unauthenticated proxy warns about nothing.

I have put the same decision to the repo owner. If they want encrypted proxy transport required — or an explicit `GRAPH_PROXY_ALLOW_INSECURE`-style opt-in — it is a small change and I will make it. Leaving this thread open until they weigh in.

On the second point in the walkthrough, external callers of the bare `New`: there are none. Every caller outside `pkg/msgraph` uses an error-returning constructor, and all eight of those apply the proxy. `New` is only reached internally by those constructors, and the `Config` doc says so.

Fixed in 6f96f8a.

---

pkg/msgraph/msgraph.go
```diff
@@ -331,22 +351,89 @@ func mutableTransport(hc *http.Client) *http.Transport {
...
+		// (`invalid URL escape "%zz"` would expose the start of a password).
+		// Redacted() can't help — it needs a URL that parsed.
+		return errors.New("parse graph proxy url: malformed value in GRAPH_PROXY_URL")
+	}
+	// Hostname(), not Host: "http://:8080" has a non-empty Host but nothing to
+	// dial. The message interpolates nothing — Redacted() only masks userinfo
+	// that url.Parse actually populated, and a scheme-less "user:pw@host" parses
+	// with User nil and the whole value in Opaque, so Redacted() would hand the
+	// password straight to the log.
+	if proxyURL.Scheme == "" || proxyURL.Hostname() == "" {
```
julianshen, line 408: _🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**Out-of-range proxy ports fail only on the first request**

For `http://proxy.corp:99999`, `url.Parse` succeeds and the scheme/hostname checks pass, so the constructor reports success. The first token or Graph request then fails while dialing the invalid port, contradicting the documented fail-fast behavior. Validate the numeric port range before installing the proxy and cover this case in the invalid-URL tests.

<details>

> ‼️ **IMPORTANT**

```suggestion
	if proxyURL.Scheme == "" || proxyURL.Hostname() == "" {
		return errors.New("invalid graph proxy url: GRAPH_PROXY_URL needs a scheme and host")
	}
	if port := proxyURL.Port(); port != "" {
		n, err := strconv.Atoi(port)
		if err != nil || n < 1 || n > 65535 {
			return errors.New("invalid graph proxy url: GRAPH_PROXY_URL has an invalid port")
		}
	}
```

</details>

<details>

```
In `pkg/msgraph/msgraph.go` around line 399: For `http://proxy.corp:99999`, `url.Parse` succeeds and the scheme/hostname checks pass, so the constructor reports success. The first token or Graph request then fails while dialing the invalid port, contradicting the documented fail-fast behavior. Validate the numeric port range before installing the proxy and cover this case in the invalid-URL tests.

```

</details>

---

pkg/msgraph/msgraph_test.go
```diff
@@ -842,3 +850,490 @@ func TestCreateOnlineMeeting_UsesObjectIDs(t *testing.T) {
...
+	}))
+	defer proxy.Close()
+
+	lister, err := NewUserListerClient(
+		Config{
+			TenantID: "t", ClientID: "c", ClientSecret: "s",
+			ProxyURL:      proxy.URL,
+			ProxyUsername: "proxyuser",
+			ProxyPassword: "proxypass",
+		},
```
julianshen, line 905: _🧪 tests_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`tgd-review`_

**The proxy-auth test never exercises HTTPS CONNECT**

Both test endpoints are explicitly HTTP, so this only verifies `Proxy-Authorization` on ordinary proxied requests. Production token and Graph endpoints use HTTPS, where authentication is carried on CONNECT instead. Add a CONNECT-capable test proxy with TLS targets and assert that both CONNECT requests receive the expected Basic credential.

<details>

```
In `pkg/msgraph/msgraph_test.go` around line 900: Both test endpoints are explicitly HTTP, so this only verifies `Proxy-Authorization` on ordinary proxied requests. Production token and Graph endpoints use HTTPS, where authentication is carried on CONNECT instead. Add a CONNECT-capable test proxy with TLS targets and assert that both CONNECT requests receive the expected Basic credential.

```

</details>

---

julianshen (commented):
### tGDBot review of `db17ecf` — 2 findings in 30 files

1 🟠 warning · 1 🔵 suggestion
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/390#issuecomment-5435331372), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

pkg/msgraph/msgraph.go
yenta: Confirmed — `url.Parse` only checks that the port is digits, never its range:

```
http://proxy.corp:99999   ok  scheme="http" hostname="proxy.corp" port="99999"
http://proxy.corp:65536   ok  scheme="http" hostname="proxy.corp" port="65536"
http://proxy.corp:0       ok  scheme="http" hostname="proxy.corp" port="0"
```

So the scheme/hostname checks passed it through and the failure moved to the first dial — the thing eager validation exists to prevent.

Took the suggestion, with `0` rejected too (a valid port is 1-65535, and `:0` would otherwise ask the kernel for an ephemeral port). Covered by `TestApplyProxy_RejectsOutOfRangePort` across `http` and `socks5`; the four out-of-range cases all failed before the change and the boundary cases (`:1`, `:65535`, no port) pass. The message stays static like the other rejections.

Fixed in 29a795b.

---

pkg/msgraph/msgraph_test.go
yenta: Right, and this was the weakest part of the test suite: the PR body claims the credentials are sent "including through CONNECT tunnels" and nothing proved it. Both endpoints were `WithTokenURL("http://login.example.test/token")` / `WithBaseURL("http://graph.example.test")`, so only the plain-proxied path — where the header rides the request itself — was exercised. Production talks to `login.microsoftonline.com` and `graph.microsoft.com` over https, where `Proxy-Authorization` is emitted on CONNECT instead, a separate branch in `net/http`.

Added `TestProxyCredentials_SentOnConnectTunnel`: two `httptest.NewTLSServer` targets behind a CONNECT-only proxy that records `Proxy-Authorization` per tunnelled host, then splices the connections. It answers `405` to anything that is not CONNECT, so the assertion cannot pass unless a real tunnel was opened — the existing http test stays as the counterpart for the non-tunnelled path.

Two details worth noting: the proxy dials only the targets the test declares rather than `r.Host` (gosec flags the latter as G704 SSRF, and a test double has no reason to be an open relay), and idle tunnels are dropped via `CloseIdleConnections` before `httptest`'s cleanup waits on the handler, so the test can't hang.

Fixed in 29a795b.

---

pkg/msgraph/msgraph.go
```diff
@@ -331,22 +351,99 @@ func mutableTransport(hc *http.Client) *http.Transport {
...
-		// Redacted() masks any embedded proxy credentials before it reaches logs.
-		return fmt.Errorf("invalid graph proxy url %q: scheme and host are required", proxyURL.Redacted())
+	// Validate the effective userinfo, whether it came from the explicit fields
+	// or from the URL — the checks above only saw the explicit ones.
+	if proxyURL.User != nil {
+		username := proxyURL.User.Username()
+		if password, ok := proxyURL.User.Password(); ok && password != "" && username == "" {
+			// Basic would send ":password"; the proxy answers 407 on the first request.
+			return errors.New("graph proxy password set without a username")
+		}
```
julianshen, line 443: _🏷️ failure-handling_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`, `tgd-review` (2 rules)_

**Oversized SOCKS5 credentials fail only on the first request**

With a `socks5`/`socks5h` proxy and an effective username or password longer than 255 bytes, `applyProxy` returns success and the service finishes startup. The first token request then reaches Go's SOCKS authentication path, which cannot encode RFC 1929's one-byte username/password lengths, so it fails before contacting the proxy; consequently no token or Graph synchronization can succeed. Validate both decoded URL userinfo and explicit credentials at construction, preserving the intended fail-fast behavior, and cover both credential sources in tests.

<details>

> ‼️ **IMPORTANT**

```suggestion
	if proxyURL.User != nil {
		username := proxyURL.User.Username()
		password, hasPassword := proxyURL.User.Password()
		if hasPassword && password != "" && username == "" {
			// Basic would send ":password"; the proxy answers 407 on the first request.
			return errors.New("graph proxy password set without a username")
		}
		if (proxyURL.Scheme == "socks5" || proxyURL.Scheme == "socks5h") &&
			(len(username) > 255 || len(password) > 255) {
			return errors.New("invalid graph proxy credentials: SOCKS5 username and password must each be at most 255 bytes")
		}
```

</details>

<details>

- **`tgd-review`**:423: The effective-userinfo validation does not enforce SOCKS5's one-byte username and password lengths. With a `socks5`/`socks5h` proxy and either credential longer than 255 bytes, construction succeeds, but Go's SOCKS dialer rejects the credential only when the first Graph request is made. This breaks the documented fail-fast behavior. Validate the effective SOCKS username and password lengths before installing the proxy, return a static error that does not expose either value, and add boundary tests for 255 and 256 bytes.

</details>

<details>

```
In `pkg/msgraph/msgraph.go` around line 423: With a `socks5`/`socks5h` proxy and an effective username or password longer than 255 bytes, `applyProxy` returns success and the service finishes startup. The first token request then reaches Go's SOCKS authentication path, which cannot encode RFC 1929's one-byte username/password lengths, so it fails before contacting the proxy; consequently no token or Graph synchronization can succeed. Validate both decoded URL userinfo and explicit credentials at construction, preserving the intended fail-fast behavior, and cover both credential sources in tests.

```

</details>

---

julianshen (commented):
### tGDBot review of `3f267f7` — 2 findings in 30 files

2 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/390#issuecomment-5435331372), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

pkg/msgraph/msgraph.go
yenta: Confirmed against the bundled dialer — `net/http/socks_bundle.go:448`:

```go
case socksAuthMethodUsernamePassword:
	if len(up.Username) == 0 || len(up.Username) > 255 || len(up.Password) > 255 {
		return errors.New("invalid username/password")
	}
```

It runs while authenticating, i.e. on the first Graph request, so construction reported the config sound and every token and sync call then failed — the same shape as the port finding.

Took the suggestion, plus the empty-username case from the same guard: `len(Username) == 0` fails identically, and `socks5://:@proxy` reaches it because the existing password-without-username check only fires when the password is non-empty. So for `socks5`/`socks5h` the username must be 1-255 bytes and the password at most 255.

`len()` is bytes, not runes, which matches the wire — a 200-rune CJK password is over the limit and is now rejected here rather than at the proxy.

Scoped to the SOCKS schemes: Basic imposes no length limit, and `TestApplyProxy_RejectsUnencodableSocksCredentials` pins that with a 256-byte credential on `http` still being accepted. It covers both sources (explicit vars and embedded userinfo), the 255/256 boundary, and asserts the rejection never echoes the credential.

Fixed in 7b43bee.

---

yenta:
## `sast` is red on this PR, and it is not this PR's failure

`sast` fails on `7b43bee` at the **semgrep** stage (`gosec=PASS govulncheck=PASS semgrep=FAIL rule-tests=PASS`), with one blocking finding in a file this PR does not touch:

```
pkg/circuitbreaker/metric.go
 ❯❯❱ semgrep.metrics-no-per-call-attribute-set
    53┆ s.gauge.Record(ctx, int64(to), attrs)
```

### It reproduces on `main` alone

Checked out `main` at `b6ea40b` in a clean worktree and ran the repo-local rules with CI's own invocation:

```
semgrep scan --error --severity=WARNING --severity=ERROR --metrics=off \
  --exclude=chat-frontend --exclude=testdata --exclude=docs --exclude=.semgrep --config=.semgrep/ .
→ Ran 18 rules on 1798 files: 1 finding.   (identical finding, same file, same line)
```

Two commits landed on `main` today, and `pkg/circuitbreaker/metric.go` (`e94e185`, #188) precedes the rule (`bf0ea62`, #337), so the rule is what turned `main` red. CI runs on the PR merge ref, so every open PR will hit this until `main` is fixed — the second time this week a `main`-side SAST change has done that (#429's gosec `-confidence low` was the first).

### The finding is a false positive

```go
func (s StateGauge) Track(ctx context.Context, name string) Option {
	// Built once at wiring time rather than per transition.
	attrs := metric.WithAttributes(attribute.String("breaker", name))
	return WithOnTransition(func(from, to State) {
		...
		s.gauge.Record(ctx, int64(to), attrs)
	})
}
```

The set is built **once per `Track` call, at wiring time**, and captured by the closure; the closure is what runs per transition. That is precisely the pattern the rule wants. The rule's own comment draws the line it is enforcing — *"constructing the option in the same function that records is per-call; building it in a constructor and looking it up here is the pattern this rule wants"* — but its intraprocedural taint sees the closure body as part of the enclosing function, so a captured-once set reads as source-and-sink within one function. Closure capture is the gap: the two sides run at different frequencies while sharing one lexical scope.

### No fix exists yet

None of the 30 open PRs touches `pkg/circuitbreaker/metric.go` or this rule, so there is nothing to port into this PR, and I have not widened this PR with an unrelated fix. Proposed patch for whoever owns it, matching the repo's existing suppression style (155 `// nosemgrep:` sites):

```go
	return WithOnTransition(func(from, to State) {
		slog.WarnContext(ctx, "circuit breaker transition",
			"breaker", name, "from", from.String(), "to", to.String())
		// nosemgrep: metrics-no-per-call-attribute-set -- attrs is built once per
		// Track call at wiring time and captured here; the rule's intraprocedural
		// taint cannot see that the closure runs at a different frequency.
		s.gauge.Record(ctx, int64(to), attrs)
	})
```

Teaching the rule to ignore a source captured by a returned func literal would fix the class rather than this instance, but that is a change to `.semgrep/metrics.yml` and belongs with its author.

I have not spent a CI re-run on this: the finding is deterministic static analysis, so a re-run cannot change it, and reproducing it on `main`'s own tree is stronger evidence than a second identical run. Everything else on this PR is green, and I will keep it watched.

---

yenta:
## `test` is red too, same root cause, also not this PR's

Following the `sast` comment above: `test` also fails on `7b43bee`, on the same file and from the same `main` merge.

```
--- FAIL: TestInstrumentNamesAreLiterals
    Should be empty, but was [pkg/circuitbreaker/metric.go: Int64Gauge(StateMetricName]
FAIL	github.com/hmchangw/chat/pkg/obs
```

Reproduced on `main` at `b6ea40b` in a clean worktree — `go test -run TestInstrumentNamesAreLiterals ./pkg/obs/` fails identically there, with nothing from this branch present.

So `#337` landed two guards, and `pkg/circuitbreaker/metric.go` (`#188`, merged earlier) trips both:

| Check | Guard | Verdict |
|---|---|---|
| `sast` | semgrep `metrics-no-per-call-attribute-set` | false positive — the set is built once and captured by the closure |
| `test` | `pkg/obs.TestInstrumentNamesAreLiterals` | **true positive** — `Int64Gauge(StateMetricName)` passes a const, not a literal |

The second one is a real violation of the new contract, and the test's own doc comment states the intended fix: the name must be a literal at the construction site so `docs/specs/o11y/nats-metrics-contract.md` can be checked against it, *"and the only way past the registry is to add the row"*. So for whoever owns it:

```go
	g, err := meter.Int64Gauge("circuit_breaker_state",
		metric.WithDescription("Circuit breaker state (0=closed, 1=open, 2=half-open)"))
```

keeping `StateMetricName` for the dashboard/consumer side, plus the matching row in the contract doc. (`skipInstrument` only exempts `data-migration/`, so there is no allowlist route here.)

Both failures are `main`'s, both predate this branch, and neither is something this PR should carry — I have not touched `pkg/circuitbreaker`. `main` is currently red on `sast` **and** `test`, so every open PR is blocked until it is fixed. Everything else on this PR passes, and I will keep it watched and re-run once `main` is green.

---

pkg/msgraph/msgraph.go
```diff
@@ -331,22 +351,119 @@ func mutableTransport(hc *http.Client) *http.Transport {
...
+		return fmt.Errorf("unsupported graph proxy scheme %q: want http, https, socks5 or socks5h", proxyURL.Scheme)
+	}
+	if cfg.ProxyUsername != "" {
+		proxyURL.User = url.UserPassword(cfg.ProxyUsername, cfg.ProxyPassword)
+	}
+	// Validate the effective userinfo, whether it came from the explicit fields
+	// or from the URL — the checks above only saw the explicit ones.
+	if proxyURL.User != nil {
+		username := proxyURL.User.Username()
+		password, _ := proxyURL.User.Password()
```
julianshen, line 432: _🏷️ failure-behavior_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`_

**Empty HTTP proxy userinfo bypasses fail-fast validation.**

For `GRAPH_PROXY_URL=http://:@proxy:8080` (or `http://@proxy:8080`), `proxyURL.User` is non-nil but both decoded fields are empty, so the `password != ""` condition does not reject it. Construction succeeds, then `net/http` treats the non-nil userinfo as credentials and sends `Proxy-Authorization: Basic Og==`; an authenticating proxy returns 407 on the token request, leaving the service running while every Graph operation fails. Reject effective userinfo with an empty username, or clear entirely empty userinfo so no authentication header is emitted, and add an HTTP-proxy test covering this URL shape.

<details>

```
In `pkg/msgraph/msgraph.go` around line 432: For `GRAPH_PROXY_URL=http://:@proxy:8080` (or `http://@proxy:8080`), `proxyURL.User` is non-nil but both decoded fields are empty, so the `password != ""` condition does not reject it. Construction succeeds, then `net/http` treats the non-nil userinfo as credentials and sends `Proxy-Authorization: Basic Og==`; an authenticating proxy returns 407 on the token request, leaving the service running while every Graph operation fails. Reject effective userinfo with an empty username, or clear entirely empty userinfo so no authentication header is emitted, and add an HTTP-proxy test covering this URL shape.

```

</details>

---

julianshen (commented):
### tGDBot review of `8b1c078` — 1 finding in 30 files

1 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/390#issuecomment-5435331372), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

pkg/msgraph/msgraph.go
yenta: Confirmed on both halves. `url.Parse` keeps the userinfo:

```
http://:@proxy.corp:8080     User!=nil username="" password="" -> "Basic Og=="
http://@proxy.corp:8080      User!=nil username="" password="" -> "Basic Og=="
http://proxy.corp:8080       User=nil  -> no Proxy-Authorization
```

and `net/http` keys the header off nothing but that non-nil (`transport.go:986`):

```go
func (cm *connectMethod) proxyAuth() string {
	if cm.proxyURL == nil { return "" }
	if u := cm.proxyURL.User; u != nil {
		username := u.Username()
		password, _ := u.Password()
		return "Basic " + basicAuth(username, password)
	}
	return ""
}
```

There is no emptiness test anywhere on that path, so `Basic Og==` goes out and an authenticating proxy answers 407 — the pod starts and every Graph call fails, exactly as you describe.

Took the "reject" branch rather than clearing the userinfo: silently dropping it would make `GRAPH_PROXY_URL=http://@proxy:8080` mean *unauthenticated egress*, which is the failure mode the credentials-without-a-URL check already exists to prevent. Better to stop at construction and have the operator fix the value.

The root cause was that the password-without-username check only fired when the password was non-empty, so both-empty slipped through. A username is now required whenever userinfo is present at all, for every scheme — which also subsumes the SOCKS empty-username case from the previous thread, so that branch keeps only its length limit.

`TestApplyProxy_RejectsEmptyUserinfo` covers `http`, `https` and `socks5` for both spellings, and pins that `http://user@proxy:8080` (username, no password) and a plain no-userinfo URL still reach the transport unchanged. The three http/https cases failed before the change; socks5 already passed, from the earlier fix.

Fixed in 9d4cf3e.

---

pkg/msgraph/msgraph_test.go
```diff
@@ -842,3 +852,724 @@ func TestCreateOnlineMeeting_UsesObjectIDs(t *testing.T) {
...
+func proxyTargetOf(t *testing.T, c any) *url.URL {
+	t.Helper()
+	g, ok := c.(*graphClient)
+	require.True(t, ok, "client must be backed by *graphClient")
+	tr, ok := g.httpClient.Transport.(*http.Transport)
+	require.True(t, ok, "transport must be *http.Transport")
+	require.NotNil(t, tr.Proxy, "proxy must be configured on the transport")
+	req, err := http.NewRequest(http.MethodGet, "https://graph.microsoft.com/v1.0/me", nil)
+	require.NoError(t, err)
+	u, err := tr.Proxy(req)
```
julianshen: _🏷️ dead-code_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`dead-code`_

**proxyTargetOf builds a request that the fixed proxy ignores.**

applyProxy installs http.ProxyURL(proxyURL), whose callback returns the fixed URL without inspecting its request. Therefore the hardcoded request in proxyTargetOf—and its impossible construction-error assertion—cannot affect any caller. Invoke the callback with nil instead.

<details>

> ‼️ **IMPORTANT**

```suggestion
	u, err := tr.Proxy(nil)
```

</details>

<details>

```
In `pkg/msgraph/msgraph_test.go` around line 866: applyProxy installs http.ProxyURL(proxyURL), whose callback returns the fixed URL without inspecting its request. Therefore the hardcoded request in proxyTargetOf—and its impossible construction-error assertion—cannot affect any caller. Invoke the callback with nil instead.

```

</details>

---

julianshen (commented):
### tGDBot review of `9d4cf3e` — 1 finding in 30 files

1 🔵 suggestion
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/390#issuecomment-5435331372), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

pkg/msgraph/msgraph_test.go
yenta: Correct — `ProxyURL`'s closure does not even name its parameter (`transport.go:505`):

```go
func ProxyURL(fixedURL *url.URL) func(*Request) (*url.URL, error) {
	return func(*Request) (*url.URL, error) {
		return fixedURL, nil
	}
}
```

So the hand-built Graph request and the `require.NoError` on a construction that cannot fail were both inert. Taken as suggested.

One note I added rather than leaving implicit: `nil` is safe *because* `applyProxy` installs `http.ProxyURL`. If the proxy func ever becomes request-dependent — NO_PROXY handling would do it — this panics instead of failing quietly, which is the right way for that change to announce itself. The comment says so, so the coupling is deliberate rather than accidental.

Fixed in a51bbdd.

---

mliu33 (approved):
Excellent, thanks!

PR #391, opened 2026-08-27
feat(loadgen): make a soak failure diagnosable from the metric that reports it

flywindy:
A spike in `loadgen_soak_errors{action="subscription_list"}` named the action and nothing else. Finding out which account could not load its sidebar meant guessing from timestamps: every lane callback logged `"error", err`, and the error had been built as `fmt.Errorf("issue %s request", action)` with the subject and account it had in hand thrown away.

Chasing one such spike took several days of Mongo queries and arithmetic that a single log line would have answered.

### The failing request now says who it was

`soakRequestError` carries the identity past the point where it is known. `soakRPCClient.Call` is the one chokepoint every lane's RPC passes through and it already holds the action, subject, error class, reason and retry count, so the carrier is stamped there rather than threaded through nine call paths. Account and room join it from the request, which each lane already had — `SubscriptionList` took an account from `pickRoom` and discarded it.

**42 of the 44 request sites** now carry an account, a room, or both. The two that carry neither are `presence_query` and the batch `rooms_info` read, which are addressed to no single account and no single room.

`soakErrorAttrs` unpacks the carrier at the nine lane callbacks and falls back to the bare error for anything that never reached the RPC client. The ~200 startup and shutdown log sites are untouched: those are one-shot, unique, and have nothing to attach.

```
msg="run Cassandra soak room read"
error="room read lane: subscription_list request failed: response payload exceeds maximum size"
action=subscription_list  account=alice
subject=chat.user.alice.request.user.tw01.subscription.list
error_class=response_too_large  error_reason=response_too_large  attempts=1
```

`account` is what joins a loadgen error to the `subject` on the server's own log line for the same call.

### Two measurements the run was not taking

**`loadgen_soak_reply_bytes{action}`** sizes every successful paged read against the broker's 128 KiB `max_payload`. This is the metric whose absence cost those days — a distribution climbing into the top bucket is the warning that the next reply comes back as `response_too_large` instead of data. Failures leave it zero on purpose: a transport error has no reply to size and an oversize one carries only the compact envelope, so recording either would report a healthy small page where the run in fact got nothing.

**`loadgen_soak_rows{action}`** exposes `sample.Messages`, which the lanes have always computed and the recorder has always dropped. Latency alone cannot separate a slow page from a large one.

Both label only the bounded `action`. Accounts, rooms and subjects stay in the log, never in a metric label.

### An interrupted run no longer reports itself as a healthy one

`Call`'s two `ctx.Err()` guards returned a bare error and a zero-valued result. `soakReadCollectorRecorder` derives the outcome from the error class alone, so an empty class is not "unclassified" — it is `soakOutcomeSucceeded`. Every read still in flight when a run was torn down padded the success count and folded its latency into the histogram, and the log line that went with it named neither the action nor the account. Pre-existing, and squarely in the machinery this PR is about.

Both guards now carry the identity and classify. The class is `canceled` rather than `internal`: the run going away is not the site failing, and `internal` is in `transientSoakError`, so the old mapping would also have bought a retry against a context that is already dead.

Cancellation is the reason **only when nothing else happened.** If an attempt reached the wire and timed out, that failure is why the operation failed and the teardown merely ended the retrying — overwriting it would erase a timeout whose effect on the server is unknown, which is exactly what the mutation accounting must not lose. Both exits therefore keep the earlier class *and* name both causes in the message, joined with a second `%w` so `errors.Is` matches the cancellation and the transport error alike. A test pins that `soakRoomMutationNeverSent` still refuses to call an interrupted mutation proven-unsent.

### The startup summary reports what a run actually loaded

Not what its knobs asked for. The two diverge whenever the site has fewer eligible users than `SOAK_MAX_USERS`, and nothing reported the difference — a run could measure a workload nobody chose and look configured doing it.

```
msg="Cassandra soak topology loaded" runId=... borrowed_users=11883 active_users=2000
rooms=10000 subscriptions=768800 subs_per_user=64.7
messages_per_active_user_per_day=2376
```

That last number is the point: 55 sends/s over 2,000 accounts is 2,376 messages per person per day, which no human sustains. Printed on the first page of the log it is impossible to miss; derived by hand from three separate knobs it went unnoticed for the life of a run.

### The room distribution is configurable

`SOAK_ROOM_ZIPF_S` and `SOAK_ROOM_ZIPF_V` replace hardcoded constants, defaulting to the same `1.2` and `1.0` so no percentile recorded before this change becomes incomparable with one after it.

`S` must exceed 1 and `V` must be at least 1, and both must be finite. `math/rand` returns a **nil generator** rather than an error outside that range — and worse for `NaN`, which passes its `s <= 1` guard and yields a non-nil generator whose `Uint64` never terminates. `SOAK_ROOM_ZIPF_S=NaN` is a reachable typo (`strconv.ParseFloat` accepts it), so the run would hang on its first send with no error and no log. Refused at startup and again in `newSoakRoomPicker`.

`V` is the flattening knob, and the only one — `S` cannot go below 1:

| `V` | hottest room's share of all sends |
|---|---|
| 1 (default) | 20.8% |
| 5 | 5.0% |
| 10 | 2.6% |

## Defects found after the first commit

Three in self-review, four more raised in review and verified against the code before acting:

- **`loadgen_soak_rows` was observed for actions that have no rows.** Around thirty series stood as a solid bar at `rows=0`, reading as "this action returns no rows" when the action has no concept of one. Gating on a non-zero count would have been wrong — an empty page is a real answer — so `RowsCounted` marks the samples that describe a page.
- **The action was printed up to three times.** The carrier prefixed it, `Call` names it, and four lane adapters wrapped that with `"issue %s request"`. The carrier now passes its cause through and the lane wrapper names the *lane* (`"room read lane: %w"`), which is context `Call` genuinely does not carry.
- **Point lookups polluted the row histogram.** `RoomInfoFor` asks for exactly one room and shares the `room_state_read` label with `RoomState`'s member list, so its constant 1 stood in the same distribution as a real count. `SubscriptionFor` and `SubscriptionByRoom` answer 0-or-1 by construction. All three keep assigning `Messages` and leave `RowsCounted` false.
- **Two room-scoped reads dropped the room.** The index probe and the by-room subscription read send a room in the body, which is gone by the time the failure is logged. A sweep of every request literal confirmed these were the only two with a room in scope and no `RoomID`.
- **The eight `hugeParam` suppressions were mangled** by a botched multi-site edit: the rationale ran backwards with its continuation printed above the directive, and three sites carried a stale block and a duplicate directive. Collapsed to one single-line directive each; confirmed still load-bearing (removing one reports `request is heavy (104 bytes)`).

## Verification

| Gate | Result |
|---|---|
| `make test SERVICE=tools/loadgen` (`-race`, whole package) | exit 0 |
| `golangci-lint run ./tools/loadgen/...` | 0 issues |
| `sast` (gosec + govulncheck + semgrep) | passing in CI. `gosec` is configured with `-exclude-dir=tools`; `semgrep` could not run in the authoring sandbox (proxy returns 403 for `semgrep.dev`) and CI covers it |

Every new test was watched failing first. Seven behaviours were re-checked against a deliberate mutation, each failing for the right reason: the carrier's omit-empty rule, its unwrapping through a `%w` chain, the `RowsCounted` gate, the action-once message, the identity through a dead context, the class not being overwritten by a teardown, and both causes surviving an interrupted retry.

## Known, deliberately out of scope

- **`ReplyBytes` is not recorded for mutations, verifications or sends** — only the read and presence lanes copy it from the RPC result. The metric help says "paged reads" so it does not overstate its coverage.
- **A decode failure discards a real reply's size**, which is precisely when it would be interesting.
- **`nolintlint` is not enabled** in `.golangci.yml`, which is why the duplicate directives above survived review. Worth enabling; not something to slip into this PR.
- `soakTopologySummary`'s `json` tags are unused; `round2` and `secondsPerDay` are generic identifiers in a large `package main`.
- No `values.yaml` **default** changed. The two Zipf knobs are added to the chart as empty strings, following the `subscriptionListLimit` pattern, so an unset value renders no env var and the Go defaults stand.

## Not in this PR

`pkg/natsrouter`'s `"reply failed"` line still carries neither the reply size nor a request ID, so an oversize failure cannot be joined to a loadgen error by anything but the account. That is a one-line change to shared production code and was deliberately kept separate.

## For the reviewer to decide

Failure logs now carry `account`, `room_id` and `subject`. Metrics do not — those stay bounded to `action`. Whether the existing log access and retention policy is the right home for those identities is an owner call, not something this PR assumes.

---

## Summary by CodeRabbit

* **New Features**
  * Added configurable room-popularity distributions for soak load tests.
  * Added Prometheus metrics for successful paged-read reply sizes and row counts.
  * Load topology summaries now appear in structured run logs.

* **Bug Fixes**
  * Improved cancellation and retry reporting, including clearer error classifications and context.
  * Row-count metrics now exclude operations without real row counts while including empty pages.
  * Configuration validation now rejects invalid or non-finite distribution values.

* **Documentation**
  * Documented new configuration options, metrics, and detailed lane-failure logging.


tools/loadgen/soak_config.go
```diff
@@ -286,6 +292,12 @@ func validateSoakConfig(cfg *soakConfig, cassandraKeyspace string) error {
 	// MAX_SUBSCRIPTION_LIMIT, but that is the service's own configuration and
 	// loadgen cannot read it, whereas a negative page size is wrong under every
 	// configuration.
+	if cfg.RoomZipfS <= 1 {
+		return fmt.Errorf("SOAK_ROOM_ZIPF_S must be greater than 1, got %v", cfg.RoomZipfS)
+	}
+	if cfg.RoomZipfV < 1 {
+		return fmt.Errorf("SOAK_ROOM_ZIPF_V must be at least 1, got %v", cfg.RoomZipfV)
+	}
```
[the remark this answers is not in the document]
flywindy: Confirmed, and the consequence is worse than "outside the usable numeric domain" — fixed in 5f6f892.

I checked it empirically rather than by reading. `strconv.ParseFloat("NaN")` returns NaN with no error, so `SOAK_ROOM_ZIPF_S=NaN` is a reachable typo. It then passes three guards in a row: `NaN <= 1` is false, so `validateSoakConfig` admits it; math/rand's own guard is also `s <= 1`, so `NewZipf` returns a **non-nil** generator; and `newSoakRoomPicker`'s nil check therefore passes too. The generator's `Uint64` never terminates — a test that called it timed out instead of returning.

So the run hangs on its first send with no error and no log, which is strictly worse than the nil dereference these bounds were added to prevent. Rejected at both levels now (`validateSoakConfig` and `newSoakRoomPicker`), with table tests covering NaN and +Inf for each of S and V at both.

Used `math.IsNaN(x) || math.IsInf(x, 0)` rather than adding an `isFinite` helper — one call site each, and the stdlib predicates say what they check.

---

tools/loadgen/soak_error.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Error passes the cause through unchanged. The action reaches the reader
+// twice over already — once from the lane's own wrap, once from the action
+// attr — and a third copy here only lengthens the line.
+func (e *soakRequestError) Error() string {
+	if e.err == nil {
+		return string(e.Action) + " request failed"
+	}
+	return e.err.Error()
+}
```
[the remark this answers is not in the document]
flywindy: Right, and this was already listed as out of scope in the PR body — your comment is what changed my mind, so it's fixed in 5f6f892.

I'd removed the carrier's own copy in the previous commit, taking it from three to two. What I left was the pre-existing pair: `soakRPCClient.Call` names the action in every error it returns, and four lane adapters wrapped that with `"issue %s request"`. My reason for stopping was `CLAUDE.md`'s "never return bare err", which rules out simply deleting the lane wrap.

Your framing — remove the action from *one* textual wrapper — is the way through that I'd missed. The lane wrapper now names the lane (`"room read lane: %w"`, `"search read lane: %w"`, `"user read lane: %w"`, `"room mutation lane: %w"`), which is context `Call` genuinely does not carry, instead of repeating what it does. `Call` keeps naming the action once.

The regression test drives the real lane through a failing transport rather than composing the wrappers itself, since the duplication lives in how the layers compose — a test that builds the string by hand would keep passing after a regression. Verified by mutation: restoring `"issue %s request"` fails it with the two-copy message quoted in the failure output.

---

flywindy:
Two review items had no inline thread to reply on.

## Outside-diff finding: `soak_main.go:155-160`, "do not mark every read sample as a page"

Correct, and fixed in 5f6f892 — but the two examples it names split.

`get_message_by_id` is a real hit: it records a hardcoded `Messages: 1`, which is not a count of anything. Marking it put a constant series next to the real pages.

`presence_query` is not. It records `len(response.States)` (`soak_presence.go:208`) — a genuine batch size, and one of the more useful numbers the lane produces. Left alone.

A third site the finding did not reach is worse than either: `soak_userread.go:343` records `sample.Messages = response.Count` for `user_subscription_count`. That is the user's *entire* subscription total, not rows in the reply — the reply carries one number. It would have reported values like 97 into a histogram whose help says "rows returned per paged read".

Rather than propagate a page flag per producer, counting is now explicit at the source: the sample carries `RowsCounted`, set only through a `countRows` helper, which the 22 sites that count a real length call. The two exceptions keep assigning `Messages` directly and are visibly not marked. The recorder forwards the marker instead of asserting one. Verified by mutation — hardcoding the recorder's marker back to `true` fails the uncounted-read case.

Also renamed the collector-side field so one concept does not carry two names across the hop (`HasPage` → `RowsCounted`).

## Pre-merge check: Docstring Coverage 60% vs 80%

Declining this one, and it is a disagreement with the check rather than an oversight.

This repository deliberately keeps comments sparse and reason-giving. It ships a `remove_comments` maintenance skill whose stated job is to "delete WHAT-comments, shorten what remains to max 2 lines (target 65% 1-line)". Adding a docstring to every touched function to reach 80% would restate signatures the code already states, and would be reverted the next time that skill runs.

The functions in this diff that carry a non-obvious contract are documented: `soakRequestError` and `soakErrorAttrs` say what a metric label cannot hold; `countRows` says which counts qualify and why; `newSoakRoomPicker` says why the parameter bounds exist and what happens without them. The ones without a docstring are `Error`/`Unwrap` implementations and test helpers whose names are their whole contract.

Happy to be overruled by the repo owner if the threshold is meant to bind.

---

julianshen:

<details>

* `tools/loadgen/metrics.go`
* `tools/loadgen/soak_distribution.go`
* `tools/loadgen/soak_distribution_test.go`
* `tools/loadgen/soak_error.go`
* `tools/loadgen/soak_error_test.go`
* `tools/loadgen/soak_mutation.go`
* `tools/loadgen/soak_payload_metrics_test.go`
* `tools/loadgen/soak_presence.go`
* `tools/loadgen/soak_read.go`
* `tools/loadgen/soak_roomops.go`
* `tools/loadgen/soak_rpc.go`
* `tools/loadgen/soak_search.go`
* `tools/loadgen/soak_topology_summary.go`
* `tools/loadgen/soak_topology_summary_test.go`
* `tools/loadgen/soak_zipf_test.go`

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

tools/loadgen/soak_rpc.go
```diff
@@ -376,14 +388,28 @@ func (c *soakRPCClient) Call(
 	if err := ctx.Err(); err != nil {
 		return result, err
 	}
+	// carry stamps the request identity onto every failure leaving this
+	// function, so the lane logger above does not have to thread it through.
```
julianshen, line 396: _🏷️ error-handling_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`, `silent-failure`, `tgd-review` (3 rules)_

**Pre-canceled RPCs are recorded with an empty error class**

The new carrier is created after the early `ctx.Err()` return. If an action context is already canceled when `Call` starts, the method returns a zero-valued result without request identity or an error classification. Read callers copy that empty `ErrorClass` into the sample, so the recorder can classify the operation as successful; the top-level callbacks then log the error and return nil. This undercounts canceled operations and leaves only a generic log without action/account context. Define `carry` before the cancellation check, classify the context error into `result.ErrorClass`/`ErrorReason`, and return `carry(err)`.

<details>

- **`distributed-system`**:392: An action context can expire after lane preparation but before Call begins. The initial ctx.Err() check then returns before carry is defined, so the lane receives a bare context error; soakErrorAttrs cannot find a soakRequestError and logs none of the action, subject, account, room, class, or attempt fields. During a timeout or shutdown this makes the canceled broker request unidentifiable. Construct carry before the context check and return carry(err), with a test using an already-canceled context.
- **`tgd-review`**:396: Call returns directly from the ctx.Err guard before carry is defined. A lane invoked with an already-canceled context therefore logs only its generic lane wrapper and loses the action, subject, account, and room that this change promises on every failure. Define the carrier before the guard and return carry(err) there; add a pre-canceled-context test to cover this path.

</details>

<details>

```
In `tools/loadgen/soak_rpc.go` around line 392: The new carrier is created after the early `ctx.Err()` return. If an action context is already canceled when `Call` starts, the method returns a zero-valued result without request identity or an error classification. Read callers copy that empty `ErrorClass` into the sample, so the recorder can classify the operation as successful; the top-level callbacks then log the error and return nil. This undercounts canceled operations and leaves only a generic log without action/account context. Define `carry` before the cancellation check, classify the context error into `result.ErrorClass`/`ErrorReason`, and return `carry(err)`.

```

</details>

tools/loadgen/soak_search.go
```diff
@@ -171,14 +173,15 @@ func (r *soakSearchReader) IndexedAt(
 	err := r.call(ctx, soakRPCRequest{
 		Action:  soakRPCSearchIndexProbe,
 		Subject: subject.SearchMessages(account, r.cfg.SiteID),
+		Account: account,
```
julianshen: _🏷️ observability_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`, `elasticsearch` (2 rules)_

**Room-scoped reads omit the room from failure identity.**

IndexedAt knows roomID and sends it in the body, but only Account reaches the failure carrier. SubscriptionByRoom has the same omission at tools/loadgen/soak_userread.go:358. If either request times out, the body is unavailable to soakErrorAttrs and the logged account and reusable subject cannot distinguish concurrent requests for different rooms. This prevents a failed verification/read from being tied to the affected room. Set RoomID: roomID on both soakRPCRequest values.

<details>

- **`elasticsearch`**:176: The `search_index_probe` request filters Elasticsearch results to `roomID`, but only `Account` is copied into the error carrier. If this probe times out or Elasticsearch returns an error, `soakErrorAttrs` omits `room_id`, so the failure log cannot identify which room's indexing observation failed. Include `RoomID` as is done for other room-scoped RPCs.

<details>

```text
		Account: account, RoomID: roomID,
```

</details>


</details>

<details>

```
In `tools/loadgen/soak_search.go` around line 176: IndexedAt knows roomID and sends it in the body, but only Account reaches the failure carrier. SubscriptionByRoom has the same omission at tools/loadgen/soak_userread.go:358. If either request times out, the body is unavailable to soakErrorAttrs and the logged account and reusable subject cannot distinguish concurrent requests for different rooms. This prevents a failed verification/read from being tied to the affected room. Set RoomID: roomID on both soakRPCRequest values.

```

</details>

tools/loadgen/soak_roomread.go
```diff
@@ -269,10 +273,11 @@ func (r *soakRoomReader) RoomInfoFor(ctx context.Context, roomID string) (soakRo
 	err := r.call(ctx, soakRPCRequest{
 		Action:  soakRPCRoomStateRead,
 		Subject: subject.RoomsInfoBatchSubscribe(r.cfg.SiteID),
+		RoomID:  roomID,
 		Body:    soakRoomsInfoRequest{RoomIDs: []string{roomID}},
 		Timeout: r.cfg.RequestTimeout, RetryMode: soakRetrySafe,
 	}, &response, func(sample *soakReadSample) {
-		sample.Messages = len(response.Rooms)
+		sample.countRows(len(response.Rooms))
 	})
```
julianshen, line 283: _🏷️ observability_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**Point lookups pollute the paged-row histogram.**

RoomInfoFor requests exactly one room and treats any successful response count as a row sample, although the new metric explicitly excludes constant-count point reads such as get-message. SubscriptionFor and soakUserReader.SubscriptionByRoom do the same for one-room subscription lookups. These 0/1 samples, including their reply sizes, pollute the distributions and in this file are combined under room_state_read with the genuinely variable member-list count. Preserve Messages if existing accounting needs it, but leave RowsCounted false for these point lookups and add coverage analogous to the get-message test.

<details>

```
In `tools/loadgen/soak_roomread.go` around line 281: RoomInfoFor requests exactly one room and treats any successful response count as a row sample, although the new metric explicitly excludes constant-count point reads such as get-message. SubscriptionFor and soakUserReader.SubscriptionByRoom do the same for one-room subscription lookups. These 0/1 samples, including their reply sizes, pollute the distributions and in this file are combined under room_state_read with the genuinely variable member-list count. Preserve Messages if existing accounting needs it, but leave RowsCounted false for these point lookups and add coverage analogous to the get-message test.

```

</details>

tools/loadgen/soak_mutation.go
```diff
@@ -602,6 +608,14 @@ func (m *soakMutator) readReactionState(
 	return false, nil
 }
 
+// gained the account and room that identify a failure. One copy per RPC sits
```
julianshen: _🧹 readability_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`tgd-review`_

**Duplicated nolint blocks obscure the RPC helpers.**

The new callMutation comment contains two overlapping rationales and two consecutive gocritic directives. The same patch artifact appears above soakReader.call and soakRoomMutator.call. Consolidate each helper to one concise rationale and one nolint directive; duplicate directives may also be reported as unused when nolintlint enforcement is enabled.

<details>

```
In `tools/loadgen/soak_mutation.go` around line 611: The new callMutation comment contains two overlapping rationales and two consecutive gocritic directives. The same patch artifact appears above soakReader.call and soakRoomMutator.call. Consolidate each helper to one concise rationale and one nolint directive; duplicate directives may also be reported as unused when nolintlint enforcement is enabled.

```

</details>

julianshen (commented):
tGD inline review

tools/loadgen/soak_rpc.go
flywindy: Confirmed on every point, including the consequence — fixed in dff60a0e.

I traced the chain rather than taking it on the description. `Call` returns a zero-valued `result`; `soakRoomReader.call` copies `result.ErrorClass` into the sample (`soak_roomread.go:388`); `soakReadCollectorRecorder.Record` derives the outcome from that class alone (`soak_main.go:152`). An empty class is therefore not "unclassified" — it is `soakOutcomeSucceeded`. So a read interrupted at teardown was counted as a completed read, with its latency folded into the histogram.

Two corrections to the scope:

The in-loop guard at the top of each attempt (`soak_rpc.go:426`) had the same defect, and it is the worse of the two because it can fire after attempts were already spent. Both guards now classify and carry.

This is pre-existing, not introduced here — `git show e7e94c90:tools/loadgen/soak_rpc.go` has both bare returns. It is in scope anyway: this PR's premise is that every failure carries its identity, and a silent success miscount sits in the machinery it touches.

On classification, I did not reuse the existing mapping. `classifySoakRPCError(context.Canceled)` fell through to `soakErrorInternal`, which would spike a server-fault class at the end of every run and — since `internal` is in `transientSoakError` — buy a retry against a context that is already dead. Added `soakErrorCanceled` instead: bounded label, non-transient, and distinct from `timeout`, which `DeadlineExceeded` keeps mapping to because the two say different things about a site under load.

Covered by a pre-canceled call, an expired deadline, a cancellation between attempts, and the whole chain driven through a real lane asserting the recorded class is not empty. Verified by mutation: restoring the bare return fails the first of those.

---

tools/loadgen/soak_search.go
flywindy: Correct at both sites — fixed in dff60a0e.

`IndexedAt` scopes the query with `RoomIDs: []string{roomID}` and `SubscriptionByRoom` sends `soakUserRoomRequest{RoomID: roomID}`, and in both cases the body is gone by the time the lane logs the failure. Both now set `RoomID: roomID` on the request.

I swept the rest rather than fixing only the two you named: every `soakRPCRequest` literal in the package that omits `RoomID`, cross-checked against what is in scope at the call site. Nineteen omit it and seventeen are right to — `presence_query` and `rooms_info` are batches, `room_create` has no room yet, `search_messages`/`search_rooms` carry no room filter, and the user-profile family is account-scoped. These two were the only genuine misses.

The regression test for `SubscriptionByRoom` compares the carrier's `RoomID` against the `roomId` decoded from the recorded request body, so it cannot pass by coincidence on a fixture with one room.

---

tools/loadgen/soak_roomread.go
flywindy: Right on all three, and the `room_state_read` observation is the sharpest part of it — fixed in dff60a0e.

That label is written by two lanes. `RoomState` records `len(response.Members)`, which is the variable count the histogram exists to show; `RoomInfoFor` asks for exactly one room and gets one back. Counting the second stood a constant 1 in the same distribution as the first — the same defect `get_message_by_id` was excluded for one commit earlier, which I missed because I was looking for hardcoded literals rather than for calls whose shape guarantees the count.

`SubscriptionFor` and `SubscriptionByRoom` are the 0-or-1 case, and `SubscriptionFor`'s own doc comment already says so ("This answers 0-or-1 for the room asked about"). All three now assign `Messages` directly and leave `RowsCounted` false, matching how the two existing exceptions are written.

I checked the neighbours rather than assuming: `RoomsInfo` batches through `pickRoomBatch` and stays counted, and the `search_index_probe` reply is a real page (`Size: r.cfg.PageSize`) despite being room-filtered, so it stays counted too.

Coverage is a table over both room-reader point lookups plus one for `SubscriptionByRoom`, and — because excluding them must not empty the distribution they were polluting — a test that `RoomState` is still counted under the same label. Verified by mutation: restoring `countRows` on `RoomInfoFor` fails the table.

---

tools/loadgen/soak_mutation.go
flywindy: Confirmed, and it is worse than three sites — fixed in dff60a0e.

All eight `hugeParam` suppressions were mangled by a botched multi-site edit of mine, not just the three with a duplicate directive. Every one of them read backwards:

```go
// failure identity; one copy per RPC is nothing beside the marshal and round trip.
//
//nolint:gocritic // hugeParam: soakRPCRequest crossed 80 bytes when it gained the
```

The directive's rationale ends mid-sentence on "gained the", and its continuation is the line printed *above* it. So even the five sites that looked clean were unreadable. The three you found carried a stale four-line block on top of that plus the second directive.

Each is now one single-line directive, matching the twenty pre-existing `//nolint:gocritic` lines in this package. I did not take the suppression itself on faith — removing one reports `hugeParam: request is heavy (104 bytes)`, so it is load-bearing and the rationale is accurate.

On `nolintlint`: it is not enabled in `.golangci.yml` today, so the duplicates were not being reported — which is exactly why they survived review. Worth enabling separately; not something I would slip into this PR.

---

flywindy:
@coderabbitai your merge-risk note on `dff60a0e` found a real inconsistency, but the direction is inverted — and chasing it turned up a defect I had just introduced. Fixed in 75a5086f.

The note says retry interruptions "can still be reported under the preceding request error category instead of as cancellation or deadline failures". That describes the backoff-sleep path, which keeps the previous attempt's class. I checked what that class is load-bearing for before touching it, and it should stay:

An attempt that reached the wire and timed out has an **unknown effect on the server**. For a mutation that is the whole ballgame — the accounting has to keep treating it as ambiguous and reconcilable. Reporting `canceled` there would say the run simply stopped, which understates the uncertainty, and `soakRoomMutationNeverSent`'s own doc comment spells out why: "a timeout, a dropped connection, or a reply that failed to decode may all have been delivered, so calling those not_sent would erase a real effect from the accounting."

So the rule is: **cancellation is the reason only when nothing else happened.**

Under that rule the sleep path was already right and my new guard was wrong. `dff60a0e` classified the context error unconditionally at the top of each attempt, so a teardown after a failed attempt overwrote `timeout` with `canceled`. The inconsistency you detected was between my new guard and the existing path — not within the old code. The guard now classifies only while the class is still empty.

I checked whether the overwrite had actually misled anything rather than assuming it had: the mutation lanes compare `ErrorClass` against `not_found`, `soakRoomMutationNeverSent` keys on `request_encode` or an empty class, and `soak_failure.go:484` reads `soakSendReplyResult` — a different type fed by the JetStream send lane, not by `Call`. So no consumer was reachable. That is luck, not design, so the not_sent invariant is now pinned by a test case instead of by inspection.

Verified by mutation: dropping the emptiness check reports the teardown in place of the timeout and fails the between-attempts test.

Docstring Coverage (62.12%) — my earlier comment on this stands; the position hasn't changed and I won't repeat the reasoning.

---

tools/loadgen/soak_rpc.go
```diff
@@ -367,28 +385,53 @@ func newSoakRPCClient(
...
 	}
 
 	for attempt := 1; attempt <= c.retry.MaxAttempts; attempt++ {
 		if err := ctx.Err(); err != nil {
-			return result, err
+			// Only when no attempt reached the wire. Once one has failed, that
+			// failure is why the operation failed and the cancellation merely
+			// ended the retrying — overwriting it would erase a timeout whose
+			// effect on the server is unknown.
+			if result.ErrorClass == "" {
```
julianshen, line 435: _🏷️ error-handling_ | _🟠 Warning_ | _⚡ Quick fix_ | _`silent-failure`_

**Retry cancellation discards the request failure that triggered retrying**

After an RPC attempt fails transiently, cancellation can be observed either during `Sleep` or immediately before the next attempt. Both paths return an error wrapping only the cancellation while retaining the previous attempt's error class. Consequently, `soakErrorAttrs` can log `error="context canceled"` with (for example) `error_class="internal"`, hiding the actual transport or service error that caused the retry. Preserve the last `requestErr` and join or wrap it together with the cancellation at both exits so the failure log contains both causes.

<details>

```
In `tools/loadgen/soak_rpc.go` around line 431: After an RPC attempt fails transiently, cancellation can be observed either during `Sleep` or immediately before the next attempt. Both paths return an error wrapping only the cancellation while retaining the previous attempt's error class. Consequently, `soakErrorAttrs` can log `error="context canceled"` with (for example) `error_class="internal"`, hiding the actual transport or service error that caused the retry. Preserve the last `requestErr` and join or wrap it together with the cancellation at both exits so the failure log contains both causes.

```

</details>

julianshen (commented):
tGD inline review

tools/loadgen/soak_rpc.go
flywindy: Correct, and it is the exact complement of the fix in `75a5086f` — fixed in cb9d2a1c.

That commit made the cancellation exits *keep* the failed attempt's class, on the grounds that an attempt which reached the wire is why the operation failed. What it did not do was carry the same information into the message, so the two disagreed on the line:

```
error="context canceled"   error_class=timeout
```

A reader has to decide which of those is the truth, and the transport error that caused the retrying is not on the line at all. You are right that both belong there.

`lastRequestErr` now carries the attempt's failure forward and `soakInterruptedError` joins the two with a second `%w`, so the message names why the operation failed and why it stopped being retried, and `errors.Is` matches both the cancellation and the underlying transport error rather than only the first.

Two things I checked rather than assumed:

The guard before the first attempt has no second cause, so it names the action instead of returning the bare context error. That is `Call`'s own convention on every other return, and a test pins that the action still appears exactly once in the chain — the duplication removed in `5f6f892` does not creep back in through this path.

The join does not weaken the not_sent invariant. `soakRoomMutationNeverSent` refuses to call a mutation proven-unsent when `errors.Is(err, context.Canceled)` matches, and it still matches through a multi-`%w` chain. A test case covers that, because an interrupted mutation may well have already reached the server.

Verified by mutation: dropping the second cause fails both interrupted-retry cases (backoff and next-attempt).

---

tools/loadgen/soak_rpc.go
```diff
@@ -367,28 +385,59 @@ func newSoakRPCClient(
...
 	var result soakRPCResult
+	// carry stamps the request identity onto every failure leaving this
+	// function, so the lane logger above does not have to thread it through.
+	// Declared before the first guard: a context that died before the wire is
+	// still a failure the operator has to be able to place.
+	carry := func(err error) error {
+		if err == nil {
+			return nil
+		}
+		return &soakRequestError{
```
julianshen, line 403: _🏷️ observability_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`_

**A canceled retry can be reported even though it never reached transport.**

After a transient first failure, `Sleep` can finish and the existing `result.Retries++` runs, then the context can be canceled before the next loop's context guard. That guard returns a carrier with `Attempts=1` and `Retries=1`, although only the initial transport request occurred; the same value also reaches the retry metric through the read sample. This misreports shutdown races as executed retries. Increment `Retries` only after the context guard when an attempt after the first is actually about to call the transport, preserving `Retries == Attempts-1`.

<details>

```
In `tools/loadgen/soak_rpc.go` around line 403: After a transient first failure, `Sleep` can finish and the existing `result.Retries++` runs, then the context can be canceled before the next loop's context guard. That guard returns a carrier with `Attempts=1` and `Retries=1`, although only the initial transport request occurred; the same value also reaches the retry metric through the read sample. This misreports shutdown races as executed retries. Increment `Retries` only after the context guard when an attempt after the first is actually about to call the transport, preserving `Retries == Attempts-1`.

```

</details>

tools/loadgen/deploy/k8s/values.schema.json
```diff
@@ -310,6 +312,8 @@
         "userReadRate": {"type": "string", "minLength": 1},
         "subscriptionListLimit": {"type": "string", "pattern": "^([0-9]+)?$"},
         "subscriptionListIncludeLastMessage": {"type": "string", "enum": ["", "true", "false"]},
+        "roomZipfS": {"type": "string", "pattern": "^([0-9]+(\\.[0-9]+)?)?$"},
```
julianshen: _🏷️ configuration_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**Invalid Zipf values pass Helm schema validation**

These patterns accept values the runtime explicitly rejects, such as `roomZipfS: "1"` and `roomZipfV: "0.5"`. Helm therefore accepts and deploys the chart, but `validateSoakConfig` then refuses to start the load generator. Align the schema with the documented constraints (`S > 1`, `V >= 1`) while retaining the intentional empty-string sentinel.

<details>

```
In `tools/loadgen/deploy/k8s/values.schema.json` around line 315: These patterns accept values the runtime explicitly rejects, such as `roomZipfS: "1"` and `roomZipfV: "0.5"`. Helm therefore accepts and deploys the chart, but `validateSoakConfig` then refuses to start the load generator. Align the schema with the documented constraints (`S > 1`, `V >= 1`) while retaining the intentional empty-string sentinel.

```

</details>

julianshen (commented):
tGD inline review

tools/loadgen/soak_rpc.go
flywindy: Confirmed — fixed in ef75f658.

Traced the exact window: attempt 1 passes the guard, `Attempts++` → 1, the transport fails transiently, `lastRequestErr` is set, `Sleep` returns nil, `Retries++` → 1. Attempt 2's guard then finds the context dead and returns `Attempts=1, Retries=1` with one request ever on the wire. That count is not confined to the carrier either — the lane copies it into `soakReadSample.Retries` and `SoakCollector` adds it to `loadgen_soak_retries_total{action,phase}`, so a shutdown race inflated a real counter.

The increment moves to the top of the loop, guarded on `attempt > 1`, where the transport call is about to happen. `Retries == Attempts-1` now holds on every exit.

One thing I checked before moving it, because it would have been a silent regression: `delay := c.backoff(result.Retries)` reads the counter as its exponent. Both arrangements present the same value at that point — 0 on the first backoff, 1 on the second — so the sequence is unchanged, and the existing delay assertions (`{100ms, 150ms}`) still pass untouched.

I also added the opposite test rather than only the one you described, so the fix cannot quietly trade an over-count for an under-count: a run that actually retries still reports `Attempts=2, Retries=1`.

Verified by mutation: restoring the old increment site fails the shutdown-race test and leaves the executed-retry test passing.

---

tools/loadgen/deploy/k8s/values.schema.json
flywindy: Correct — fixed in ef75f658.

I copied the `subscriptionListLimit` pattern for shape and never went back to make it say what the runtime actually requires. `^([0-9]+(\.[0-9]+)?)?$` admits `"1"`, `"0.5"` and `"0"`, all of which `validateSoakConfig` refuses, so Helm renders and deploys and the operator learns about it from a crash-looping pod.

The patterns now mirror the bounds exactly:

- `roomZipfV` — `^[1-9][0-9]*(\.[0-9]+)?$`, i.e. `V >= 1`.
- `roomZipfS` — `S > 1`, which needs two branches because `1.2` is valid and `1.0` is not: `^([2-9][0-9]*|[1-9][0-9]+)(\.[0-9]+)?$` for an integer part of 2 or more, and `^1\.[0-9]*[1-9][0-9]*$` for `1.` followed by a non-zero fraction.

Both keep the empty-string sentinel as a separate `{"const": ""}` branch, so an unset value still renders no env var and the Go defaults stand.

`helm` is not installed in the authoring sandbox, so I could not run `make validate-loadgen-k8s`. Rather than eyeball the regexes I validated them with `jsonschema` over a 24-value table covering both directions — accepted: `""`, `1.2`, `1.05`, `2`, `10`, `3.5` for S and `1`, `1.0`, `5`, `10.5` for V; rejected: `1`, `1.0`, `1.00`, `0.8`, `0`, `-2`, `abc`, `1.2.3` for S and `0.5`, `0`, `0.999`, `-1` for V. All 24 behave as the runtime does.

A bonus worth noting: `NaN`, `Inf` and `Infinity` are now rejected at Helm time as well, so the silent hang fixed in `5f6f892` can no longer be introduced through `values.yaml` either — previously the old pattern rejected them only by accident of not matching digits, and nothing tied that to the runtime rule.

---

flywindy:
@coderabbitai your merge-risk note on `ef75f658` — "a canceled room mutation may still be recorded as rejected even when the service outcome is unknown" — is correct. Fixed in cd4aecd8.

`observeAdmission`'s switch ends in a default branch that calls `AbandonWithReason(failureResultBad, failureReasonAdmissionRejected)` and settles the operation, which hands the candidate back to the pool as untouched. A cancellation fell through to it, because `transientSoakError(soakErrorCanceled)` is false — deliberately, since a dead context must not buy a retry.

So a run torn down mid-mutation asserted the service had refused a request that may well have been delivered and applied, and closed the operation so nothing would reconcile it. The branch directly above already spells out why that is the wrong direction, for the unknown-status case:

> calling it a rejection would close the operation, hand the candidate back as untouched, and leave a real effect with nothing left to reconcile it

One correction to the framing: this is not something the cancellation work introduced. Before `dff60a0e` the same path carried an empty error class, and `transientSoakError("")` is false too, so it landed in the same default branch. Naming the class made the path legible; it was already wrong.

A cancellation now joins the ambiguous case — never resent, never called `not_sent`, left active for reconciliation. When nothing was sent the reconciler finds nothing and closes it, which is the safe direction; when something was sent it is the only correct one.

Verified by mutation rather than by reading: dropping the cancellation from the guard closes the operation as `bad` again and fails the new test, while `TestSoakRoomLanes_ExplicitRejectionClosesTheOperation` keeps passing — a real `forbidden` envelope is still a refusal, which is the half that must not regress.

Worth noting for anyone reading the thread: this one only surfaced in your risk summary, not as an inline comment, and the inline pass reported no actionable findings on the same commit. It was the more valuable of the two outputs here.

---

vjauhari-work (approved):
LGTM!

PR #395, opened 2026-08-27
perf(room-service): project the five unprojected Mongo reads

hmchangw:
## What

Five `room-service` store reads fetched whole documents, against the "always project precisely" rule in CLAUDE.md §6. Each now projects only the fields its callers actually read, following the existing `roomReadProjection` / `subscriptionReadProjection` pattern already in `store_mongo.go`.

| Method | Now projects | Was |
|---|---|---|
| `GetUser` | `_id, account, engName, chineseName, roles` | all ~24 `model.User` fields |
| `GetApp` | `_id, assistant` | full `App` incl. `Sponsors`, `AppViewURL` |
| `FindDMSubscription` | `_id, roomId` | full ~35-field `Subscription` |
| `GetThreadSubscriptionByParent` | `_id, threadRoomId` | full `ThreadSubscription` |
| `getRoomSubscriptions` | `_id, u._id, u.account, roles, joinedAt` | full `Subscription`, **per member** |

`GetUser` is the one worth a second look beyond throughput: it was pulling `users.services` — the bcrypt credential block — on every create-room and every `IsPlatformAdmin` gate. That no longer leaves Mongo on this path.

`getRoomSubscriptions` is the one that scales worst: it decoded a ~35-field, pointer-heavy `Subscription` for every member of a room to build a `RoomMember` from five fields.

## How the field sets were derived

Each projection is the union of every field read by an actual call site, traced transitively — `GetUser`'s result is passed on to `handleCreateSelfDM` / `handleCreateRoomDMOrBotDM` / `handleCreateRoomChannel` / `publishCreateRoom`, and those only ever read `ID`, `Account`, `EngName`, `ChineseName`, plus `Roles` via `model.IsPlatformAdmin`. No call site consumes the whole struct or calls `IsActive()`.

Note for reviewers: `SubscriptionUser.ID` maps to `bson:"_id"`, so the nested path in `roomMemberSubProjection` is `u._id`. `u.id` projects nothing and would silently zero `Member.ID` in every member list — there's a comment on the var to that effect.

## Tests

Adds five `*_ProjectionFields_Integration` tests matching the existing guards, each pinning both what the projection **must return** (with the reason it's needed) and what it **must exclude**, so a future call site that starts reading a dropped field fails loudly rather than silently seeing a zero value.

## Verification

- `make test SERVICE=room-service` — passes (`-race`)
- `go vet ./room-service/...` and `go vet -tags=integration ./room-service/...` — clean
- `make lint` — 0 issues
- `gosec` and the repo-local semgrep rules — clean

**Integration tests were not executed.** Docker and MongoDB are both unavailable in the environment this was written in, so the new tests were verified to compile under `-tags=integration` but never run. They should fail against the pre-change code (with no projection the full document decodes, so every "must exclude" assertion fails), but that is reasoning, not an observed red. **Please run `make test-integration SERVICE=room-service` before merging** — that run is also what would automatically catch a nested-path mistake like the `u.id` / `u._id` one noted above.

## Not included

No wire-visible behavior changes, so `docs/client-api.md` needs no update — every field any handler reads is still returned.

`GetTeamsMeeting` was left alone despite being technically unprojected: it's a flat 5-field struct where two fields are the query keys, so a projection is churn with drift risk for no measurable gain.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GZFrWJvqo8jwe68736JmvB)_


julianshen:
**No actionable comments.** ✅

<details>

</details>
<details>

* `cassandra`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

PR #404, opened 2026-08-27
feat(history-service): resolve legacy members_removed display names

evelynchang5525:
## Summary

Implements server-side display-name resolution for legacy `members_removed` system messages in the history service. When a page contains these rows, the raw account identifier baked into the message text is replaced with the user's display name via a single batched Mongo query, with results cached to avoid repeated lookups during scroll-back.

## Changes

- **New matcher & batch resolver** (`sysmsgname.go`): Pure string logic to identify legacy `members_removed` rows (type + exact suffix match), extract the account, and rewrite the message text with the resolved display name. Batches all accounts on a page into one `FindUsersByAccounts` call; pages with no qualifying rows issue zero store calls.

- **Extended UserStore interface** (`service.go`): Added `FindUsersByAccounts(ctx context.Context, accounts []string) ([]model.User, error)` to the consumer-side interface, enabling the batch lookup.

- **Wired into all message read paths** (`messages.go`): Integrated the resolution pass into six return paths (LoadHistory, LoadNextMessages, loadSurroundingByMessageID, assembleSurrounding, GetMessageByID, GetMessagesByIDs) after attachment decoding and before page fitting, so the rewritten bytes are weighed in the trim budget.

- **User profile cache** (`config.go`, `main.go`): Added `UserCacheSize` and `UserCacheTTL` config fields (defaults: 50k entries, 10m TTL) and wired an LRU+TTL cache in front of the Mongo store to avoid re-resolving the same accounts across multiple pages.

- **Documentation** (`docs/client-api.md`): Noted the server-side rewrite on the `msg` field for legacy rows.

## Implementation Details

- **Degradation**: Store errors or missing user documents leave rows untouched (read as raw account, exactly as today). A name lookup failure never fails a history load.

- **Scope**: Changes are confined to `messages.go` only; `pin.go` and `threads.go` are out of scope (a `members_removed` row can never be pinned or become a thread parent).

- **Ordering**: The pass runs after `redactUnavailableQuotes` / `setDecodedAttachments` and before `fitPage` / `fitWindow`, so the budget trim accounts for the rewritten text.

- **Test coverage**: 90%+ coverage on new logic; all six handler paths tested end-to-end; batching and degradation assertions included.

https://claude.ai/code/session_01HV2ybDmGevhTNb6MpDmQDZ

## Summary by CodeRabbit

* **Documentation**
  * Added an implementation plan for displaying removed members’ names in legacy system messages.
  * Documented the planned display-name substitution and best-effort fallback behavior when user details are unavailable.
  * Described planned updates to message retrieval and client API documentation.
  * Included configurable caching considerations to support efficient name resolution.

### DIFF
--- a/docs/client-api.md
+++ b/docs/client-api.md
@@ -2909,7 +2909,7 @@ Used by every history-service method that returns messages. Mirrors the Cassandr
 | `createdAt` | string | RFC 3339 timestamp. |
 | `messageId` | string | 17- or 20-char base62. |
 | `sender` | [MessageParticipant](#messageparticipant) | The message author. |
-| `msg` | string | The message body. |
+| `msg` | string | The message body. For legacy `members_removed` system messages, history-service substitutes the removed user's display name for the stored account identifier on read; an account with no matching user is returned unchanged. |
 | `mentions` | [MessageParticipant](#messageparticipant)[] | Optional. |
 | `attachments` | [Attachment](#attachment)[] | Optional. Decoded attachment objects (history-service decodes the stored blobs on read). |
 | `card` | [MessageCard](#messagecard) | Optional. |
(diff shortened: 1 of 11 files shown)


mliu33 (approved):
Thanks

PR #409, opened 2026-08-28
refactor(notification-worker): extract store.go with mockgen directives

hmchangw:
## Problem

notification-worker had **no `store.go`, no `//go:generate mockgen` directive, and no generated mocks** — the only service of its kind in the repo missing all three. Its Mongo data access was spread across three files:

| Where | What |
|---|---|
| `main.go:77-178` | `mongoMemberLoader` + `fillHomeSites` (~100 lines of bson/projection/cursor code) |
| `threads.go` | `mongoThreadFollowers` |
| `usersettings.go` | `mongoUserSettings` |

Consequences: every test double was hand-rolled, and the adapters were reachable only from the integration suite — production store code was being exercised out of `main.go`.

## Change

Restores the mandated per-service layout:

- **`store.go`** — the three store interfaces the service consumes (`MemberStore`, `ThreadFollowerLister`, `UserSettingsSnapshotter`) plus `ThreadRoomInfo`, with the missing `//go:generate mockgen` directive.
- **`store_mongo.go`** — the Mongo implementations, moved **verbatim**, with compile-time interface assertions.
- **`mock_store_test.go`** — generated; regeneration verified idempotent.
- **`threads.go`** — removed, fully relocated. **`usersettings.go`** keeps only the `notifSettings` domain value type and its resolution, which is not store code.

`mongoMemberLoader` → `mongoMemberStore` and `Load` → `ListMembers`, so the type no longer collides in spirit with the `memberLoader` func type that `cachedMemberLookup` takes as its seam. `main.go` drops **521 → 415 lines**.

Interfaces stay segregated rather than collapsing into one `Store` god-interface — the handler consumes each independently, which is what the consumer-defined-interface rule asks for.

## The gap this closes

`store_test.go` drives the generated mocks at the boundaries where a store failure changes the ack decision. Most importantly it closes the audit's `high` finding: **the member-fetch error branch had no test at all**, because `stubMembers.GetMembers` returned `nil` unconditionally and no fixture could make it fail. That branch decides redeliver-vs-drop for the hottest dependency in the service.

Also pinned: thread-follower store failure NAKs, settings store failure fails **open** (it gates delivery, so silencing the site would be worse), and a store error propagates rather than degrading into an empty audience — an empty list silently means "nobody to notify".

Verified the new NAK test is not vacuous. Swallowing the error makes it fail:

```
--- FAIL: TestHandle_MemberFetchError_NAKs
    Error: An error is expected but got nil.
```

## Testing

| Check | Result |
|---|---|
| `make lint` | 0 issues |
| `make test SERVICE=notification-worker` (`-race`) | pass |
| `make generate` | no diff (idempotent) |
| `gosec` | clean |
| `go vet -tags=integration` | clean |
| `go build ./...` | clean |

`govulncheck` and `semgrep` could not run in this environment (`vuln.go.dev` blocked by the proxy; `pipx`/`uv` version conflict) — unverified here rather than passed; CI covers them.

## On coverage — setting expectations honestly

The readiness report predicted this extraction would be the single largest lever on the 55.6% coverage figure. **Measured here it moves only 55.6% → 56.0%**, and the report's framing was optimistic.

The reason: extraction *relocates* the untested Mongo code rather than testing it. That code is inherently integration-tested, and Docker is unavailable in this environment so the `//go:build integration` suite cannot contribute to the number. The real gain should appear in CI where Docker exists, and the mocks now make the *consumers* of these interfaces unit-testable, which is what the new tests use.

The larger remaining lever is the rest of `main()` — the semaphore worker loop and shutdown ordering — which this PR deliberately does not touch.

## Notes for the reviewer

- Independent of the three other open branches. All touch `main.go`, but in different regions, so conflicts should be minimal — this PR only removes lines 77-178 and adjusts the wiring near line 249.
- Existing hand-written stubs in `handler_test.go` are left in place. They back ~120 behavioural tests and swapping them wholesale for gomock would be large, risky churn with no behavioural gain. The generated mocks are used where they earn their place: store-failure paths.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RCuJG55DPZkkoihJJXumxh)_


julianshen:
**No actionable comments.** ✅

<details>

* `notification-worker/mock_store_test.go`
* `notification-worker/store.go`
* `notification-worker/store_mongo.go`
* `notification-worker/store_test.go`
* `notification-worker/threads.go`
* `notification-worker/usersettings.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

hmchangw:
Not high priority right now.

PR #413, opened 2026-08-28
fix(history-service): strip the quotes around the legacy members_removed account

evelynchang5525:
## Summary

Follow-up to #404, which shipped the legacy `members_removed` display-name resolution against the wrong message format. The stored sentence **quotes** the account:

```
"bob" has been removed from the channel.
```

The merged matcher cut only the trailing suffix and took the whole remaining prefix as the account, so it looked up `"bob"` **with the quote characters included**. That value matches no `users.account` document, so every row fell through to the unresolved-account path and kept its raw text. The feature was a silent no-op on real data — safe, but with no effect.

## The fix

`extractRemovedAccount` now matches the quoted form specifically, returns the account from between the quotes, and the rewrite re-renders the resolved name back inside them:

| | |
|---|---|
| Stored | `"bob" has been removed from the channel.` |
| Returned | `"Bob 鮑勃" has been removed from the channel.` |

Keeping the quotes also matches how `room-worker/sysmsg.go` quotes names on modern `member_removed` messages, so migrated and current rows render alike.

An **unquoted** prefix no longer matches at all. Combined with the `type == "members_removed"` gate, that tightens the guard against rewriting an ordinary user message that happens to end in the same sentence.

## Changes

- **`sysmsgname.go`** — `extractRemovedAccount` requires `"…"` around a non-empty account and returns the inner text; new `quoteRemoved` helper renders the sentence back with the quotes intact.
- **`sysmsgname_test.go`** — matcher table rewritten to the real format, now 14 cases.
- **`messages_test.go`** — the `LoadHistory` end-to-end test uses the quoted shape.
- **`docs/client-api.md`** — the `msg` field now states the actual stored format and that the quotes are preserved.

## Edge cases added

`""` (empty quotes), a lone `"`, opening-quote-only, closing-quote-only, unquoted prefix, and an account containing its own quote (`"bo"b"` → `bo"b`, since only the outermost pair are delimiters).

## Behavior unchanged

Everything else from #404 holds: one batched `$in` per response regardless of how many rows qualify, zero store calls on a page with no qualifying rows, and best-effort degradation — a lookup failure or an account with no user document leaves the row exactly as it reads today.

## Verification

- `make test SERVICE=history-service` — pass, with `-race`
- `make lint` — **0 issues**
- Coverage — **100%** on all four functions in `sysmsgname.go`
- Red phase confirmed the defect before the fix: `expected: "bob"` / `actual: "\"bob\""`

## Note on the branch

#404 was squash-merged, so this branch was brought up to date by merging `main` in rather than stacking on merged history. That merge resolved two conflicts against work landed since: the `closeOnCleanupIn(t, …)` test helper and the preview warm-back config (#406) — both taken from `main`.

https://claude.ai/code/session_01HV2ybDmGevhTNb6MpDmQDZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01HV2ybDmGevhTNb6MpDmQDZ)_

### DIFF
--- a/history-service/internal/service/sysmsgname.go
+++ b/history-service/internal/service/sysmsgname.go
@@ -21,16 +21,32 @@ const (
 )
 
 // extractRemovedAccount returns the account baked into a legacy members_removed
-// row's text. Reports false for every other row, leaving it untouched.
+// row's text, which quotes it: `"bob" has been removed from the channel.`
+// Reports false for every other row, leaving it untouched.
+//
+// The quotes are part of the stored sentence, not delimiters we may keep: an
+// account carrying them matches no user document, so the row would silently go
+// unresolved.
 func extractRemovedAccount(m *models.Message) (string, bool) {
 	if m == nil || m.Type != legacyMembersRemovedType {
 		return "", false
 	}
-	account, found := strings.CutSuffix(m.Msg, removedFromChannelSuffix)
-	if !found || account == "" {
+	quoted, found := strings.CutSuffix(m.Msg, removedFromChannelSuffix)
+	if !found {
 		return "", false
 	}
-	return account, true
+	// Needs both quotes plus at least one character between them; `""` and a
+	// lone `"` name nobody. An inner quote belongs to the account itself.
+	if len(quoted) < 3 || !strings.HasPrefix(quoted, `"`) || !strings.HasSuffix(quoted, `"`) {
+		return "", false
+	}
+	return quoted[1 : len(quoted)-1], true
+}
+
+// quoteRemoved renders the sentence back with name where the account stood,
+// keeping the quotes the stored form uses.
+func quoteRemoved(name string) string {
+	return `"` + name + `"` + removedFromChannelSuffix
 }
 
 // resolveRemovedMemberNames rewrites every legacy members_removed row in msgs,
@@ -83,7 +99,7 @@ func (s *HistoryService) resolveRemovedMemberNames(ctx context.Context, msgs []m
 			continue
 		}
 		if name, found := names[account]; found {
-			msgs[i].Msg = name + removedFromChannelSuffix
+			msgs[i].Msg = quoteRemoved(name)
 		}
 	}
 }
--- a/history-service/internal/service/sysmsgname.go
+++ b/history-service/internal/service/sysmsgname.go
@@ -1,9 +1,12 @@
 package service
 
 import (
+	"context"
+	"log/slog"
 	"strings"
 
 	"github.com/hmchangw/chat/history-service/internal/models"
+	"github.com/hmchangw/chat/pkg/displayfmt"
 )
 
 const (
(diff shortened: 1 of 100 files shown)


julianshen:
**No actionable comments.** ✅

### Related work

- [PR #404](https://github.com/hmchangw/newchat/pull/404) — feat\(history-service\): resolve legacy members\_removed display names (merged)
- [PR #406](https://github.com/hmchangw/newchat/pull/406) — Queue preview warm-backs off the request path (merged)

<details>

* `docs/client-api.md`
* `history-service/internal/service/sysmsgname.go`
* `history-service/internal/service/sysmsgname_test.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

history-service/internal/service/sysmsgname.go
```diff
@@ -21,16 +21,32 @@ const (
...
 	}
-	account, found := strings.CutSuffix(m.Msg, removedFromChannelSuffix)
-	if !found || account == "" {
+	quoted, found := strings.CutSuffix(m.Msg, removedFromChannelSuffix)
+	if !found {
+		return "", false
+	}
+	// Needs both quotes plus at least one character between them; `""` and a
+	// lone `"` name nobody. An inner quote belongs to the account itself.
+	if len(quoted) < 3 || !strings.HasPrefix(quoted, `"`) || !strings.HasSuffix(quoted, `"`) {
```
julianshen, line 40: _🏷️ protocol compatibility_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`_

**Unquoted legacy rows are no longer resolved**

The base implementation and tests treated `bob has been removed from the channel.` as a valid legacy row, but the new parser rejects it unless the account is quoted. After deployment, replaying or reading any previously supported unquoted row makes `extractRemovedAccount` return false, so the user lookup is skipped and clients see the stored account identifier instead of the display name. Accept both stored formats during the compatibility window and preserve each row's original quoting style when substituting the name; retain tests for both quoted and unquoted historical rows.

<details>

```
In `history-service/internal/service/sysmsgname.go` around line 40: The base implementation and tests treated `bob has been removed from the channel.` as a valid legacy row, but the new parser rejects it unless the account is quoted. After deployment, replaying or reading any previously supported unquoted row makes `extractRemovedAccount` return false, so the user lookup is skipped and clients see the stored account identifier instead of the display name. Accept both stored formats during the compatibility window and preserve each row's original quoting style when substituting the name; retain tests for both quoted and unquoted historical rows.

```

</details>

---

julianshen (commented):
### tGDBot review of `893bae3` — 1 finding in 4 files

1 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/413#issuecomment-5455425158), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

mliu33 (approved):

history-service/internal/service/sysmsgname.go
evelynchang5525: Verified the mechanics — accurate: an unquoted row that #404's matcher accepted is rejected here. But the "previously supported" framing doesn't hold against what that code actually did in production, so I'm leaving the behaviour as-is and flagging it to the author rather than adding the compatibility branch.

Two things:

**#404 resolved nothing in production.** Its matcher cut only the trailing suffix and took the whole prefix as the account, so on the real stored form it extracted `"bob"` *including the quote characters*. That value matches no `users.account` document, so every row fell through to the unresolved-account path and kept its raw text. The unquoted branch was reachable only for a format that, per the repo owner, is not what these rows store — the sample the original implementation was written against was later corrected to the quoted form. So nothing that worked before stops working: before #404 there was no resolution, under #404 there was no resolution, and here quoted rows resolve.

**The strictness is load-bearing, not incidental.** `members_removed` rows are migrated data; the `type` gate is the primary guard, and requiring the quotes is the second one that keeps an ordinary user message ending in the same sentence from being rewritten. Accepting the bare form would re-open that and add a branch for data we have no evidence exists.

If mixed formats *do* exist in the migrated corpus, this is worth doing and I'll add it — dual-form matching plus preserving each row's original quoting on substitution, with tests for both. That's a question about the data rather than the diff, so I've put it to the author and am leaving this thread open for them to decide.

---

PR #415, opened 2026-08-28
fix(history-service): evict subscription access cache on membership change

### LINKED ISSUE #414: [Fix] Re-add with "Don't include history" still shows old messages (stale history-service subscription cache)
> **Verified against:** `main`@c604d0cf — code read after a fresh `git pull` (code is source of truth over docs).

## Summary
A member re-added with **"Don't include chat history"** can still read the old messages for a short window. Expected: the invitee sees no message history from before the re-add. Actual: the invitee sees the full prior history for up to the history-service subscription-cache TTL (default 2 min). The stored `historySharedSince` boundary is correct — the leak is a stale read-cache that is never invalidated on a membership change.

## Reproduce steps
- **Preconditions:** a channel with history; history-service subscription cache enabled (default: `HISTORY_SUB_CACHE_SIZE=100000`, `HISTORY_SUB_CACHE_TTL=2m`). Reading history first is required (it seeds the stale entry).
- **Steps:**
  1. User B is a full member of channel A; messages exist (call them OLD).
  2. **B reads history** (`msg.history`) as a full member — sees all OLD. (This caches B's access window as full-access.)
  3. Another member **removes** B.
  4. Another member **re-adds** B with **history mode = none** ("Don't include chat history").
  5. B **reads history again immediately** (within the cache TTL, same history-service instance).
- **Expected vs actual:** Expected — B sees no OLD messages (only messages from the re-add onward). Actual — B sees **all** OLD messages. After the cache TTL elapses, B correctly sees none of the OLD messages.
- **Evidence:** reproduced on our dev stack with the exact sequence above; right after the restricted re-add B saw all pre-existing messages, and after the TTL window B saw only the post-re-add message. The re-added subscription's stored `historySharedSince` equalled the re-add instant (i.e. the boundary is written correctly) — confirming the leak is the cache, not the write path.

## Root cause  (verified, file:line)
- The history read gate does an **unbounded** read when the access window is `nil` (full access): `history-service/internal/service/messages.go:75-84` (`accessSince == nil` → read all).
- The access window comes from a process-local LRU+TTL cache with **no active invalidation**: `history-service/internal/readcache/readcache.go:5` (doc comment: "Freshness is TTL-bounded; there is no active invalidation") and `:132` `SubscriptionCache.GetHistorySharedSince` (caches only `subscribed=true` results). TTL default `HISTORY_SUB_CACHE_TTL=2m` (`history-service/internal/config/config.go:80`).
- Sequence: B's full-member read caches `(B, room) → sharedSince=nil, subscribed=true`. The remove deletes B's subscription in Mongo but **does not** bust the cache. The restricted re-add writes the correct boundary (`room-worker/handler.go:146` `historySharedSincePtr` → the add timestamp), but the read gate keeps serving the cached `nil` until the entry expires → unbounded history read.
- **Sibling issue:** because a `subscribed=true` entry survives removal, a **just-removed** member can also keep reading history for up to the TTL.

## Message-path trace
1. Client request struct + docs/client-api.md — no change (existing `msg.history` request).
2. Gatekeeper mapping (request → canonical) — N/A (read path).
3. Canonical model / event structs (pkg/model) — no change.
4. Workers (message-worker / broadcast-worker / notification-worker) — no change to the write path; the fix adds a cache-eviction consumer in history-service (below).
5. Storage tables (Cassandra / Mongo) — no change; the Mongo subscription (with the correct `historySharedSince`) is already the source of truth.
6. Read / history path (history-service) — **change here**: invalidate the subscription cache on membership change so the gate reloads the fresh boundary.
7. Client-facing events (Leg-2 fan-out) — no change.

## Fix  (Primary + alternatives)
- **Primary (recommended): event-driven eviction.** Add an `Evict(account, roomID)` method to the history-service subscription cache and have history-service consume the membership-change signal, evicting `(account, roomID)` on **remove, re-add, and history-sharing change**. The implementer must confirm which subject history-service can subscribe to that carries `(account, roomID)` on all three transitions (candidates: the per-account `subscription.update` core-NATS subject; the site-internal `member_added` / `member_removed` events) — pick the one history-service's NATS credentials permit and that fires on all three. This is the same "bust on change" shape room-worker already uses for room-meta.
- **Alternative A (rejected): shorten/parameterize the TTL.** Reduces the window but never closes it — an access-control grant should not be served stale at all.
- **Alternative B (heavier): version-token validation.** Stamp a subscription/room version and revalidate cached entries against it — more moving parts than an eviction hook for the same effect.

## Contracts & docs
- docs/client-api.md updated same PR? N — no wire/contract change.
- Schema change? N — no Cassandra/Mongo schema or index change.

## Tests
- TDD red→green: a failing test that seeds a full-access cache entry, simulates a membership change, and asserts the next access read reflects the **new** (restricted / not-subscribed) window.
- **e2e (dev stack):** required — spans room-worker (membership event) + history-service (cache eviction). Drive remove → restricted re-add → immediate history read and assert OLD messages are **not** returned (no stale window). (Harness already exists: `readd-dontinclude-history-leak`.)

## Acceptance criteria
- After a remove / restricted re-add / history-sharing change, the very next `msg.history` read reflects the member's current boundary — no stale full-access window.
- A removed member cannot read history after removal (no residual `subscribed=true`).

- Which subject history-service subscribes to for the eviction signal (implementer to confirm on all three transitions + credential scope).
- Whether to also evict on a plain unsubscribe (soft toggle) — decide during implementation.

## PR / merge
- pr-self-audit is run as an internal pre-review gate; its result is NOT in the PR body.
- Source repo `hmchangw/newchat`: external review + external merge — open/rebase/CI-green/request-review; do not solo-merge.

chenjr0719:
## Summary

A member re-added with **"Don't include chat history"** — and a **just-removed** member — could still read pre-boundary history for up to the history-service subscription-cache TTL (`HISTORY_SUB_CACHE_TTL`, default `2m`). The stored `historySharedSince` boundary is written correctly; the leak is the read-side access cache. This PR keeps that cache and adds **active invalidation** plus the correctness hardening needed to make it safe under primary outage, NATS reconnect, and membership churn.

Fixes #414

## What's included

- Active eviction of the subscription access-check cache on membership change.
- Correctness hardening so the cache cannot re-authorize a revoked grant during a primary outage, a NATS reconnect, or under eviction churn.

## Changes

**Active invalidation (`readcache` + `history-service/cmd`)**
- `SubscriptionCache.Evict(account, roomID)` drops the cached access-check entry (same `account + "\x00" + roomID` key `GetHistorySharedSince` uses) so the next read reloads the current boundary. Local-only, mirroring `PreviewCache.Invalidate`.
- When the sub-cache is enabled, subscribe to `chat.user.*.event.subscription.update` and evict `(account, roomID)` on the `added` / `removed` actions (boundary-changing only; `read`/`mute`/`favorite`/`role`/`section` are skipped). Non-queue core NATS so every instance evicts its own local cache; best-effort (an unparseable subject/payload logs and skips).

**Strict `primary` read preference for access reads**
- `SUBSCRIPTION_READ_PREFERENCE` now defaults to `primary` (was `primaryPreferred`). `primaryPreferred` falls back to a lagging secondary during a primary outage, which can re-cache a revoked grant; strict `primary` is read-your-writes and fails closed instead. Env-overridable.

**Fail-closed returns an error, never a fabricated zero value**
- The cache is a generic `ttlCache[V]` backing access grants, room times, and `minUserLastSeenAt`. When repeated evictions race every load attempt, the load path now returns an error (`errLoadUnstable`) instead of `(zero, nil)` — a zero value would hand room-times/min-seen callers fabricated data as success. `getAccessSince` already treats a lookup error as fail-closed.

**In-flight load cancellation + concurrency cap**
- On evict/purge, `singleflight.Forget` only detaches the superseded load; it kept running until the fetch timeout, so under churn + slow Mongo those piled up (a goroutine + context each). Each shared load now carries a cancelable context that `remove`/`purge` cancels immediately.
- A semaphore (`HISTORY_SUB_CACHE_MAX_INFLIGHT`, default `256`) bounds concurrent source loads so a burst of distinct-key misses cannot spawn unbounded loads.

**Reconnect purge delivery barrier**
- The reconnect callback is asynchronous, so a read could reach a restored subscription before the purge ran and serve a stale grant. The cache now suspends on NATS disconnect (stops serving/storing and purges) and resumes on reconnect (purges, then re-arms). Reads during the barrier fall through to a fresh source lookup (fail-closed) rather than serving the pre-reconnect cache.

No wire/contract change (read path only); no schema or index change; `docs/client-api.md` unaffected.

## Verification

- `go build ./history-service/...` — clean.
- `go test -race ./history-service/internal/readcache/...` — pass, including new tests: superseded-load cancellation (no goroutine leak), the `maxInflight` concurrency bound, the suspend/resume barrier, and a read racing a reconnect flap never serving a stale grant.
- `go test ./history-service/internal/config/... ./history-service/internal/service/...` — pass.
- Lint (pinned golangci-lint): clean.
- e2e against the dev stack (remove → restricted re-add → immediate history read) to follow.

### DIFF
--- a/history-service/cmd/main.go
+++ b/history-service/cmd/main.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/nats-io/nats.go"
 	"go.mongodb.org/mongo-driver/v2/mongo/options"
 	"go.mongodb.org/mongo-driver/v2/mongo/readpref"
 
@@ -158,6 +159,13 @@ func main() {
 	}
 	slog.Info("mongo key read preference configured", "readPreference", keyReadPref.Mode().String())
 
+	subReadPref, err := mongoutil.ParseReadPreference(cfg.Mongo.SubscriptionReadPreference)
+	if err != nil {
+		slog.Error("invalid mongo subscription read preference", "value", cfg.Mongo.SubscriptionReadPreference, "error", err)
+		os.Exit(1)
+	}
+	slog.Info("mongo subscription read preference configured", "readPreference", subReadPref.Mode().String())
+
 	cassSession, err := cassutil.Connect(cassutil.Config{
 		Hosts:    cfg.Cassandra.Hosts,
 		Keyspace: cfg.Cassandra.Keyspace,
@@ -224,7 +232,7 @@ func main() {
 
 	cassRepo := cassrepo.NewRepository(cassSession, bucketSizer, cfg.MessageReadMaxBuckets, cipher)
 	db := mongoClient.Database(cfg.Mongo.DB)
-	subRepo := mongorepo.NewSubscriptionRepo(db)
+	subRepo := mongorepo.NewSubscriptionRepo(db, subReadPref)
 	roomRepo := mongorepo.NewRoomRepo(db, previewCipher, preview.Key{SiteID: cfg.SiteID, Epoch: cfg.PreviewKeyEpoch})
 	threadRoomRepo := mongorepo.NewThreadRoomRepo(db)
 	threadSubRepo := mongorepo.NewThreadSubscriptionRepo(db)
@@ -274,12 +282,45 @@ func main() {
 	base := subL2Source{l2: subL2, inner: subRepo}
 	var subSource service.SubscriptionRepository = base
 	if cfg.SubCacheSize > 0 && cfg.SubCacheTTL > 0 {
-		sc, err := readcache.NewSubscriptionCache(base, cfg.SubCacheSize, cfg.SubCacheTTL)
+		sc, err := readcache.NewSubscriptionCache(base, cfg.SubCacheSize, cfg.SubCacheTTL,
+			readcache.WithMaxInflight(cfg.SubCacheMaxInflight))
 		if err != nil {
 			slog.Error("init subscription cache failed", "error", err)
 			os.Exit(1)
 		}
 		subSource = sc
+		// Active invalidation: evict a member's cached access window the moment a
+		// membership change fans out, so the next msg.history read reflects the
+		// current boundary instead of the stale full-access entry (#414).
+		if err := startSubCacheInvalidation(ctx, nc, sc); err != nil {
+			slog.Error("subscribe subscription.update for cache invalidation failed", "error", err)
+			os.Exit(1)
+		}
+		// Core NATS gives no replay: an eviction fanned out while this instance was
+		// disconnected is lost, and the stale full-access entry would keep
+		// authorizing reads until TTL. Suspend the cache on disconnect (stop serving
+		// or storing grants + purge) and Resume it on reconnect (purge, then re-arm).
+		// The suspend barrier holds across the whole outage AND until the reconnect
+		// purge finishes, so no read is served from the pre-reconnect cache before
+		// the reconnect callback runs — it falls through to the primary instead.
+		// Chain the existing handlers (connection metrics) rather than replace them.
+		raw := nc.NatsConn()
+		prevReconnect := raw.Opts.ReconnectedCB
+		raw.SetReconnectHandler(func(c *nats.Conn) {
+			if prevReconnect != nil {
+				prevReconnect(c)
+			}
+			sc.Resume()
+			slog.Warn("nats reconnected: purged and re-armed subscription access cache to drop possibly-missed evictions")
+		})
+		prevDisconnect := raw.Opts.DisconnectedErrCB
+		raw.SetDisconnectErrHandler(func(c *nats.Conn, err error) {
+			if prevDisconnect != nil {
+				prevDisconnect(c, err)
+			}
+			sc.Suspend()
+			slog.Warn("nats disconnected: suspended subscription access cache; grants can't be invalidated while offline")
+		})
 		slog.Info("subscription cache enabled",
 			"size", cfg.SubCacheSize, "ttl", cfg.SubCacheTTL,
 			"sub_l2_ttl", cfg.SubL2.TTL,
--- a/history-service/cmd/subinvalidate.go
+++ b/history-service/cmd/subinvalidate.go
@@ -1,0 +1,85 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log/slog"
+
+	o11ynats "github.com/flywindy/o11y/nats"
+	"github.com/nats-io/nats.go"
+
+	"github.com/hmchangw/chat/pkg/subject"
+)
+
+// evictor is the subset of *readcache.SubscriptionCache the invalidation
+// subscription drives. Declared at the consumer so the decode step is testable
+// with a spy rather than a live cache.
+type evictor interface {
+	Evict(account, roomID string)
+}
+
+// subUpdateEnvelope is the minimal projection shared by both events published on
+// the subscription.update subject: model.SubscriptionUpdateEvent ("added", …)
+// and model.SubscriptionRemovedEvent ("removed"). Both nest roomId under
+// "subscription", and both carry the action at top level — so this single shape
+// decodes either without depending on which one arrived.
+type subUpdateEnvelope struct {
+	Action       string `json:"action"`
+	Subscription struct {
+		RoomID string `json:"roomId"`
+	} `json:"subscription"`
+}
+
+// isAccessBoundaryAction reports whether a subscription.update action changes a
+// member's history-access window (subscribed / historySharedSince). Only these
+// need an eviction: a re-add — including "Don't include history" — publishes
+// "added"; a remove publishes "removed". mute/favorite/read/role/section leave
+// the boundary untouched, and evicting on the high-frequency "read" would churn
+// the cache for no correctness gain.
+func isAccessBoundaryAction(action string) bool {
+	return action == "added" || action == "removed"
+}
+
+// evictOnSubscriptionUpdate decodes one subscription.update message and evicts
+// the (account, roomID) access-check entry when the action changes the access
+// boundary. account comes from the subject (the encoded token the event was
+// addressed to); roomID from the payload. Best-effort: an unparseable subject or
+// payload logs and returns without evicting — never panics, never crashes the
+// consumer.
+func evictOnSubscriptionUpdate(ev evictor, subj string, payload []byte) {
+	token, ok := subject.ParseSubscriptionUpdateAccount(subj)
+	if !ok {
+		slog.Warn("subscription.update: unparseable subject, skipping eviction", "subject", subj)
+		return
+	}
+	// The subject token is the encoded transport form; decode it so the cache key
+	// matches the decoded account natsrouter hands the read handlers (a ".bot"
+	// account is stored decoded). A no-op for every non-bot account.
+	account := subject.DecodeAccount(token)
+	var env subUpdateEnvelope
+	if err := json.Unmarshal(payload, &env); err != nil {
+		slog.Warn("subscription.update: decode failed, skipping eviction", "error", err, "account", account)
+		return
+	}
+	if !isAccessBoundaryAction(env.Action) || env.Subscription.RoomID == "" {
+		return
+	}
+	ev.Evict(account, env.Subscription.RoomID)
+}
+
+// startSubCacheInvalidation subscribes to every account's subscription.update
+// and evicts history-service's access-check cache on membership changes, closing
+// the stale-full-access window (#414). Non-queue core NATS: every instance must
+// receive every event to evict its own local cache. Best-effort — a dropped
+// event just falls back to the cache TTL. nc.Drain at shutdown tears the
+// subscription down, so the returned handle is discarded.
+func startSubCacheInvalidation(ctx context.Context, nc *o11ynats.Conn, ev evictor) error {
+	_, err := nc.Subscribe(ctx, subject.SubscriptionUpdateWildcard(), func(_ context.Context, msg *nats.Msg) {
+		evictOnSubscriptionUpdate(ev, msg.Subject, msg.Data)
+	})
+	if err != nil {
+		return fmt.Errorf("subscribe to subscription updates: %w", err)
+	}
+	return nil
+}
--- a/history-service/internal/config/config.go
+++ b/history-service/internal/config/config.go
@@ -48,6 +48,13 @@ type MongoConfig struct {
 	// primaryPreferred rather than primary: without it, encrypted history cannot be
 	// read at all when there is no primary.
 	KeyReadPreference string `env:"KEY_READ_PREFERENCE" envDefault:"primaryPreferred"`
+	// SubscriptionReadPreference covers the subscriptions collection, which backs
+	// the access-control reads (GetSubscription / GetHistorySharedSince). It pins to
+	// strict primary — not primaryPreferred — so the read is read-your-writes and
+	// fails closed on a primary outage: primaryPreferred would fall back to a lagging
+	// secondary that could re-cache a stale pre-revocation grant until TTL, silently
+	// defeating the eviction. Env-overridable for operators who accept that trade-off.
+	SubscriptionReadPreference string `env:"SUBSCRIPTION_READ_PREFERENCE" envDefault:"primary"`
 }
 
 // NATSConfig holds NATS connection settings (env prefix: NATS_).
@@ -94,6 +101,10 @@ type Config struct {
 	// ttl to 0 to disable.
 	SubCacheSize int           `env:"HISTORY_SUB_CACHE_SIZE" envDefault:"100000"`
 	SubCacheTTL  time.Duration `env:"HISTORY_SUB_CACHE_TTL"  envDefault:"2m"`
+	// SubCacheMaxInflight caps concurrent source loads the sub-cache can have in
+	// flight, so a burst of distinct-key misses under slow Mongo can't spawn
+	// unbounded goroutines. 0 takes the package default.
+	SubCacheMaxInflight int `env:"HISTORY_SUB_CACHE_MAX_INFLIGHT" envDefault:"256"`
 
 	// Room metadata cache (room times + minUserLastSeenAt). lastMsgAt advances
 	// on every message, so the TTL is short by default; client room hints cover
@@ -192,6 +203,9 @@ func validate(cfg *Config) error {
 	if cfg.SubCacheTTL < 0 {
 		return fmt.Errorf("HISTORY_SUB_CACHE_TTL must be >= 0, got %s", cfg.SubCacheTTL)
 	}
+	if cfg.SubCacheMaxInflight < 0 {
+		return fmt.Errorf("HISTORY_SUB_CACHE_MAX_INFLIGHT must be >= 0, got %d", cfg.SubCacheMaxInflight)
+	}
 	if cfg.RoomCacheSize < 0 {
 		return fmt.Errorf("HISTORY_ROOM_CACHE_SIZE must be >= 0, got %d", cfg.RoomCacheSize)
 	}
@@ -228,6 +242,9 @@ func validate(cfg *Config) error {
 	}
 	if _, err := mongoutil.ParseReadPreference(cfg.Mongo.KeyReadPreference); err != nil {
 		return fmt.Errorf("MONGO_KEY_READ_PREFERENCE: %w", err)
+	}
+	if _, err := mongoutil.ParseReadPreference(cfg.Mongo.SubscriptionReadPreference); err != nil {
+		return fmt.Errorf("MONGO_SUBSCRIPTION_READ_PREFERENCE: %w", err)
 	}
 	if err := cfg.Pool.Validate(); err != nil {
 		return err
--- a/history-service/internal/readcache/readcache.go
+++ b/history-service/internal/readcache/readcache.go
@@ -2,16 +2,21 @@
 // MongoDB lookups on the history-service read hot path: the subscription
 // access check, room times (lastMsgAt/createdAt), and minUserLastSeenAt.
 //
-// Freshness is TTL-bounded; there is no active invalidation. Following the
-// message-gatekeeper precedent, only positive subscriptions are cached —
-// "not subscribed" results and errors are never stored, so revoked access
-// goes stale by at most the subscription TTL and the negative path stays
-// always-fresh. Loads are deduped with singleflight.
+// Freshness is TTL-bounded; the subscription access cache is additionally
+// evicted on membership change (SubscriptionCache.Evict, driven by the
+// subscription.update subscription in cmd) so revoked or narrowed access takes
+// effect at once rather than after the TTL. Following the message-gatekeeper
+// precedent, only positive subscriptions are cached — "not subscribed" results
+// and errors are never stored, so the negative path stays always-fresh. Loads
+// are deduped with singleflight.
 package readcache
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	lru "github.com/hashicorp/golang-lru/v2/expirable"
@@ -26,6 +31,16 @@ import (
 // the singleflight goroutine or pin the in-flight key. See the design spec.
 const fetchTimeout = 10 * time.Second
 
+// defaultMaxInflight bounds concurrent source loads per cache when a caller
+// gives no explicit cap. A burst of distinct-key misses would otherwise spawn a
+// goroutine + fetch context each, so this caps the fan-out; 0 disables it.
+const defaultMaxInflight = 256
+
+// loadHandle carries the cancel of an in-flight source load so a concurrent
+// remove/purge can cancel the now-superseded load instead of letting it run to
+// fetchTimeout. Boxed so a completing load deletes only its own entry.
+type loadHandle struct{ cancel context.CancelFunc }
+
 // Recorder records the outcome of a cache lookup. cachemetrics.Recorder
 // satisfies it; tests substitute a spy.
 type Recorder interface {
@@ -39,18 +54,47 @@ type ttlCache[V any] struct {
 	lru     *lru.LRU[string, V]
 	sf      singleflight.Group
 	metrics Recorder
+	// mu guards gen and the check-then-Add in getOrLoad so an eviction cannot
+	// interleave between the generation check and the store. gen bumps on every
+	// remove; a load that raced an eviction sees the changed gen and skips its
+	// now-stale store. Global, not per-key: an eviction of any key skips any load
+	// in flight at that moment — correct, and at membership-change rates the rare
+	// extra reload is cheaper than tracking a per-key counter.
+	// ponytail: global generation, per-key if evict rate ever dwarfs read misses.
+	mu       sync.Mutex
+	gen      uint64
+	inflight map[string]*loadHandle // key → cancel of the in-flight load, guarded by mu
+
+	// sem bounds concurrent source loads; nil disables the bound.
+	sem chan struct{}
+	// valid gates cache use. Flipped false by suspend (NATS disconnect) so reads
+	// fall through to a fresh source lookup instead of serving a possibly-stale
+	// entry, and true again by resume only after the purge has cleared the cache.
+	valid atomic.Bool
 }
 
 // newTTLCache builds an LRU+TTL cache. rec records hit/miss/error outcomes;
 // pass a cachemetrics.For(name, "l1") so each cache reports its own series.
-func newTTLCache[V any](size int, ttl time.Duration, rec Recorder) (*ttlCache[V], error) {
+// maxInflight caps concurrent source loads (<=0 → defaultMaxInflight; a nil sem
+// disables the cap only via the sentinel checked below).
+func newTTLCache[V any](size int, ttl time.Duration, maxInflight int, rec Recorder) (*ttlCache[V], error) {
 	if size <= 0 {
 		return nil, fmt.Errorf("readcache: size must be positive, got %d", size)
 	}
 	if ttl <= 0 {
 		return nil, fmt.Errorf("readcache: ttl must be positive, got %v", ttl)
 	}
-	return &ttlCache[V]{lru: lru.NewLRU[string, V](size, nil, ttl), metrics: rec}, nil
+	c := &ttlCache[V]{
+		lru:      lru.NewLRU[string, V](size, nil, ttl),
+		metrics:  rec,
+		inflight: make(map[string]*loadHandle),
+	}
+	if maxInflight <= 0 {
+		maxInflight = defaultMaxInflight
+	}
+	c.sem = make(chan struct{}, maxInflight)
+	c.valid.Store(true)
+	return c, nil
 }
 
 // getOrLoad returns the cached value for key, or invokes load on miss. load
@@ -59,30 +103,148 @@ func newTTLCache[V any](size int, ttl time.Duration, rec Recorder) (*ttlCache[V]
 // nothing is cached and the error is returned.
 // If ctx is canceled before the shared load finishes, getOrLoad returns ctx.Err() immediately
 // while the load continues detached in the background, bounded by fetchTimeout.
-// remove drops key if present; absent keys are a no-op.
-func (c *ttlCache[V]) remove(key string) { c.lru.Remove(key) }
+// remove drops key if present; absent keys are a no-op. Forget detaches any
+// in-flight singleflight load for key so a caller arriving after the eviction
+// starts a FRESH source read instead of joining the doomed in-flight one (which
+// read the pre-eviction boundary). The gen bump makes that doomed load skip its
+// store and, in getOrLoad, re-read rather than return its now-stale value.
+func (c *ttlCache[V]) remove(key string) {
+	c.mu.Lock()
+	c.gen++
+	c.lru.Remove(key)
+	c.sf.Forget(key)
+	// Cancel the superseded in-flight load: Forget only detaches it, leaving the
+	// goroutine + fetch context pinned until fetchTimeout. Under churn + slow Mongo
+	// those pile up, so cancel it now — it fails closed rather than storing a value
+	// read before this eviction.
+	if h := c.inflight[key]; h != nil {
+		h.cancel()
+		delete(c.inflight, key)
+	}
+	c.mu.Unlock()
+}
+
+// purge drops every entry and bumps gen so any in-flight loads fail the
+// stability check in getOrLoad and re-read (or fail closed) rather than storing
+// a value read before the purge, and cancels every in-flight load so none pins a
+// goroutine past the purge. Used to cold-start the cache on NATS reconnect, when
+// this instance may have missed evictions while disconnected.
+func (c *ttlCache[V]) purge() {
+	c.mu.Lock()
+	c.gen++
+	c.lru.Purge()
+	for key, h := range c.inflight {
+		h.cancel()
+		delete(c.inflight, key)
+	}
+	c.mu.Unlock()
+}
+
+// suspend stops the cache serving or storing entries and purges what it holds.
+// Called on NATS disconnect: while offline this instance hears no evictions, so
+// every cached grant may already be stale — reads must fall through to the source
+// until resume re-arms the cache.
+func (c *ttlCache[V]) suspend() {
+	c.valid.Store(false)
+	c.purge()
+}
+
+// resume purges any entries re-cached during the outage, then re-arms the cache.
+// Ordering matters: valid flips true only after the purge, so no read is served
+// from the pre-reconnect cache once the connection (and eviction delivery) is back.
+func (c *ttlCache[V]) resume() {
+	c.purge()
+	c.valid.Store(true)
+}
 
 func (c *ttlCache[V]) getOrLoad(ctx context.Context, key string, load func(context.Context) (V, bool, error)) (V, error) {
-	if v, ok := c.lru.Get(key); ok {
-		c.metrics.Hit(ctx)
-		return v, nil
+	// Skip the cache entirely while suspended: a possibly-stale entry must not be
+	// served after a disconnect until resume re-arms the cache.
+	if c.valid.Load() {
+		if v, ok := c.lru.Get(key); ok {
+			c.metrics.Hit(ctx)
+			return v, nil
+		}
 	}
 
 	resCh := c.sf.DoChan(key, func() (any, error) {
 		// Re-check under singleflight in case a sibling populated the entry.
-		if cached, ok := c.lru.Get(key); ok {
-			return cached, nil
+		if c.valid.Load() {
+			if cached, ok := c.lru.Get(key); ok {
+				return cached, nil
+			}
 		}
-		fetchCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), fetchTimeout)
-		defer cancel()
-		val, store, err := load(fetchCtx)
-		if err != nil {
-			return val, err
+		// Bound concurrent source loads so a burst of distinct-key misses can't
+		// spawn unbounded goroutines. Detached from the caller ctx (like the fetch
+		// below) so a leader's cancel never poisons its coalesced waiters; bounded
+		// by fetchTimeout so a saturated cache fails closed rather than hanging.
+		if c.sem != nil {
+			select {
+			case c.sem <- struct{}{}: // fast path: a slot is free, no timer allocated
+			default:
+				// Contended: wait for a slot, bounded so a saturated cache fails
+				// closed rather than hanging.
+				waitCtx, waitCancel := context.WithTimeout(context.WithoutCancel(ctx), fetchTimeout)
+				select {
+				case c.sem <- struct{}{}:
+					waitCancel()
+				case <-waitCtx.Done():
+					waitCancel()
+					var zero V
+					return zero, fmt.Errorf("readcache: load slots exhausted: %w", waitCtx.Err())
+				}
+			}
+			defer func() { <-c.sem }()
 		}
-		if store {
-			c.lru.Add(key, val)
+		// Load, then confirm no eviction (a gen bump) raced this load. If one did,
+		// the value we just read may predate the membership change that triggered
+		// the eviction, so it must NOT be returned to ANY singleflight waiter — a
+		// stale positive grant here lets a revoked member read history. Re-read
+		// against the new generation and re-check, looping until the generation is
+		// stable across a full load. Bounded: if evictions keep racing every
+		// attempt, fail closed with an error (never a fabricated zero value that a
+		// generic-V caller would read as real data).
+		const maxLoadAttempts = 5
+		for attempt := 0; attempt < maxLoadAttempts; attempt++ {
+			fetchCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), fetchTimeout)
+			// Register the cancel so a concurrent remove/purge for this key cancels
+			// this now-superseded load rather than letting it run to fetchTimeout.
+			h := &loadHandle{cancel: cancel}
+			// One critical section: snapshot gen before loading (so a remove during
+			// the load is detected) and publish the cancel handle together.
+			c.mu.Lock()
+			gen := c.gen
+			c.inflight[key] = h
+			c.mu.Unlock()
+
+			val, store, err := load(fetchCtx)
+
+			c.mu.Lock()
+			if c.inflight[key] == h {
+				delete(c.inflight, key)
+			}
+			c.mu.Unlock()
+			cancel()
+			if err != nil {
+				return val, fmt.Errorf("load cache entry: %w", err)
+			}
+
+			// Check-and-store under the same lock as remove so an eviction can't
+			// slip between the generation check and the Add. Never store while
+			// suspended — the entry would outlive the barrier it is meant to bypass.
+			c.mu.Lock()
+			stable := c.gen == gen
+			if stable && store && c.valid.Load() {
+				c.lru.Add(key, val)
+			}
+			c.mu.Unlock()
+			if stable {
+				return val, nil
+			}
+			// An eviction landed during this load: loop and re-read the source.
 		}
-		return val, nil
+		var zero V
+		return zero, errors.New("readcache: evictions raced every load attempt, failing closed")
 	})
 	select {
 	case res := <-resCh:
@@ -120,10 +282,24 @@ type SubscriptionCache struct {
 	cache *ttlCache[subEntry]
 }
 
+// CacheOption tunes a cache constructor.
+type CacheOption func(*cacheOptions)
+
+type cacheOptions struct{ maxInflight int }
+
+// WithMaxInflight caps concurrent source loads; <=0 takes defaultMaxInflight.
+func WithMaxInflight(n int) CacheOption {
+	return func(o *cacheOptions) { o.maxInflight = n }
+}
+
 // NewSubscriptionCache wraps inner with an LRU+TTL cache of size entries and
 // the given TTL. size and ttl must be positive.
-func NewSubscriptionCache(inner SubscriptionSource, size int, ttl time.Duration) (*SubscriptionCache, error) {
-	cache, err := newTTLCache[subEntry](size, ttl, cachemetrics.For("history_sub", "l1"))
+func NewSubscriptionCache(inner SubscriptionSource, size int, ttl time.Duration, opts ...CacheOption) (*SubscriptionCache, error) {
+	var o cacheOptions
+	for _, opt := range opts {
+		opt(&o)
+	}
+	cache, err := newTTLCache[subEntry](size, ttl, o.maxInflight, cachemetrics.For("history_sub", "l1"))
 	if err != nil {
 		return nil, err
 	}
@@ -151,6 +327,40 @@ func (c *SubscriptionCache) GetHistorySharedSince(ctx context.Context, account, 
 // projects (mongorepo.subscriptionReadProjection) — read a new field, widen that first.
 func (c *SubscriptionCache) GetSubscription(ctx context.Context, account, roomID string) (*pkgmodel.Subscription, error) {
 	return c.inner.GetSubscription(ctx, account, roomID)
+}
+
+// Evict drops the cached access-check entry for (account, roomID) so the next
+// GetHistorySharedSince reloads the current subscription boundary from the
+// source. A membership change (remove, restricted re-add) alters that boundary
+// while the LRU keeps serving the stale full-access entry until its TTL — the
+// #414 leak. The key must match GetHistorySharedSince's exactly. Local only,
+// like PreviewCache.Invalidate: a sibling replica's copy lives out its own TTL.
+func (c *SubscriptionCache) Evict(account, roomID string) {
+	c.cache.remove(account + "\x00" + roomID)
+}
+
+// Purge drops every cached access-check entry. Core NATS delivers no replay, so
+// an instance that was disconnected while a membership change fanned out never
+// received the eviction and would keep serving the stale full-access entry until
+// TTL (#414). Call this from the NATS reconnect handler to cold-start the cache:
+// the next read of each key reloads the current boundary from the source.
+func (c *SubscriptionCache) Purge() {
+	c.cache.purge()
+}
+
+// Suspend stops the cache serving or storing grants and purges it. Call it from
+// the NATS disconnect handler: while offline this instance receives no evictions,
+// so a cached grant may already be stale — reads must fall through to the source
+// (fail-closed against the primary) until Resume re-arms the cache.
+func (c *SubscriptionCache) Suspend() {
+	c.cache.suspend()
+}
+
+// Resume purges any grant re-cached during the outage and re-arms the cache.
+// Call it from the reconnect handler: the purge runs before the cache is re-armed
+// so no read races in on the pre-reconnect cache once eviction delivery is back.
+func (c *SubscriptionCache) Resume() {
+	c.cache.resume()
 }
 
 // RoomSource is the room metadata reads the cache fronts.
@@ -182,11 +392,11 @@ type RoomCache struct {
 // NewRoomCache wraps inner with LRU+TTL caches for room times and
 // minUserLastSeenAt. size and ttl must be positive.
 func NewRoomCache(inner RoomSource, size int, ttl time.Duration) (*RoomCache, error) {
-	times, err := newTTLCache[roomTimes](size, ttl, cachemetrics.For("history_room_times", "l1"))
+	times, err := newTTLCache[roomTimes](size, ttl, defaultMaxInflight, cachemetrics.For("history_room_times", "l1"))
 	if err != nil {
 		return nil, err
 	}
-	minSeen, err := newTTLCache[*time.Time](size, ttl, cachemetrics.For("history_room_min_seen", "l1"))
+	minSeen, err := newTTLCache[*time.Time](size, ttl, defaultMaxInflight, cachemetrics.For("history_room_min_seen", "l1"))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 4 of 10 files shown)


chenjr0719:
Thanks for the review — holding this for a follow-up commit before it should merge. Two valid points to address first:

1. Wrap the `nc.Subscribe` error with operation context on return.
2. Close the in-flight-load vs `Evict` race: a cache-miss load that started before an eviction can `Add` its result *after* the evict, re-seeding a stale access entry until the TTL. I'll add a generation guard so a load whose entry was evicted mid-flight does not store, plus a concurrent regression test.

Will push the fixes and re-request review.

julianshen:
**16 findings · 11 unique issues · 11 inline comments posted** — 🔴 5 blocking · 🟠 8 warning · 🔵 3 suggestion

> [!NOTE]

### Related work

- [Issue #414](https://github.com/hmchangw/newchat/issues/414) — \[Fix\] Re-add with "Don't include history" still shows old messages \(stale history-service subscription cache\) (open)

<details>
<summary>📒 Files reviewed (10)</summary>

* `history-service/cmd/subinvalidate.go`
* `history-service/cmd/subinvalidate_test.go`
* `history-service/internal/mongorepo/subscription.go`
* `history-service/internal/mongorepo/subscription_test.go`
* `history-service/internal/readcache/metrics_test.go`
* `history-service/internal/readcache/readcache.go`
* `history-service/internal/readcache/readcache_test.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

history-service/internal/readcache/readcache.go
```diff
@@ -80,7 +102,13 @@ func (c *ttlCache[V]) getOrLoad(ctx context.Context, key string, load func(conte
 			return val, err
 		}
 		if store {
-			c.lru.Add(key, val)
+			// Check-and-store under the same lock as remove so an eviction can't
+			// slip between the generation check and the Add.
+			c.mu.Lock()
```
julianshen, line 235: _🎯 correctness_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`distributed-system`_

**Invalidated in-flight loads still authorize waiting readers**

The generation check only prevents caching the stale value; the closure still returns that value to every singleflight waiter. Timeline: a miss reads the old full-access subscription and pauses, the membership is revoked and Evict increments gen, then a new history request joins the existing singleflight call. The Add is skipped, but both requests receive the stale positive result and can read revoked history. On a generation mismatch, discard the loaded result and retry the source read, or fail closed rather than returning it.

<details>

```
In `history-service/internal/readcache/readcache.go` around line 107: The generation check only prevents caching the stale value; the closure still returns that value to every singleflight waiter. Timeline: a miss reads the old full-access subscription and pauses, the membership is revoked and Evict increments gen, then a new history request joins the existing singleflight call. The Add is skipped, but both requests receive the stale positive result and can read revoked history. On a generation mismatch, discard the loaded result and retry the source read, or fail closed rather than returning it.

```

</details>

---

history-service/cmd/subinvalidate.go
```diff
@@ -0,0 +1,85 @@
...
+
+// startSubCacheInvalidation subscribes to every account's subscription.update
+// and evicts history-service's access-check cache on membership changes, closing
+// the stale-full-access window (#414). Non-queue core NATS: every instance must
+// receive every event to evict its own local cache. Best-effort — a dropped
+// event just falls back to the cache TTL. nc.Drain at shutdown tears the
+// subscription down, so the returned handle is discarded.
+func startSubCacheInvalidation(ctx context.Context, nc *o11ynats.Conn, ev evictor) error {
+	_, err := nc.Subscribe(ctx, subject.SubscriptionUpdateWildcard(), func(_ context.Context, msg *nats.Msg) {
+		evictOnSubscriptionUpdate(ev, msg.Subject, msg.Data)
```
julianshen, line 79: _🏷️ nats-reconnect_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`, `mongodb`, `nats` (3 rules)_

**Core NATS reconnects can preserve revoked history access**

`SubscriptionUpdateWildcard()` is consumed through a Core NATS subscription, so an instance receives no replay for updates published while its connection is disconnected or its subscription is unavailable. Concrete path: the instance caches a member's full-access entry, disconnects, the member is removed or re-added with restricted history while disconnected, then the instance reconnects and continues serving the stale cached entry until TTL expiry because no later event evicts it. This is a reachable authorization failure. Clear the entire subscription cache on disconnect/reconnect before reads resume, or use a replayable per-instance JetStream consumer and gate readiness until it catches up.

<details>

- **`distributed-system`**:79: This non-durable core NATS subscription has no replay or recovery for missed invalidations, as the comment explicitly acknowledges. Timeline: a replica disconnects, another service commits a membership removal and publishes the update, the replica reconnects without receiving that event, and its existing positive cache entry continues authorizing history reads until TTL expiry. Clear or bypass the cache across disconnect/reconnect boundaries, or consume invalidations through a replayable per-replica subscription so processing resumes from the last acknowledged event.
- **`mongodb`**:77: This non-durable subscription cannot replay events missed during a disconnect. A replica can cache full access, disconnect while an `added` restriction or `removed` event is published, then reconnect and serve the stale MongoDB-derived access entry until its TTL expires, exposing history after access was narrowed or revoked. Use a durable/replayable invalidation mechanism tied to the membership write, or stop caching positive access grants. At minimum, clear the entire subscription cache on reconnect and subscriber loss so missed events cannot preserve stale grants.

</details>

<details>

```
In `history-service/cmd/subinvalidate.go` around line 79: `SubscriptionUpdateWildcard()` is consumed through a Core NATS subscription, so an instance receives no replay for updates published while its connection is disconnected or its subscription is unavailable. Concrete path: the instance caches a member's full-access entry, disconnects, the member is removed or re-added with restricted history while disconnected, then the instance reconnects and continues serving the stale cached entry until TTL expiry because no later event evicts it. This is a reachable authorization failure. Clear the entire subscription cache on disconnect/reconnect before reads resume, or use a replayable per-instance JetStream consumer and gate readiness until it catches up.

```

</details>

---

history-service/internal/readcache/readcache.go
```diff
@@ -60,7 +72,12 @@ func newTTLCache[V any](size int, ttl time.Duration, rec Recorder) (*ttlCache[V]
 // If ctx is canceled before the shared load finishes, getOrLoad returns ctx.Err() immediately
 // while the load continues detached in the background, bounded by fetchTimeout.
 // remove drops key if present; absent keys are a no-op.
-func (c *ttlCache[V]) remove(key string) { c.lru.Remove(key) }
+func (c *ttlCache[V]) remove(key string) {
+	c.mu.Lock()
+	c.gen++
+	c.lru.Remove(key)
+	c.mu.Unlock()
+}
```
julianshen, line 141: _🔒 security_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`tgd-review`_

**Eviction still exposes stale in-flight results to later readers**

Incrementing the generation prevents an old load from being stored, but it does not detach that load from the singleflight group. If a full-access load begins before removal, then Evict runs, a history request arriving afterward still joins the old flight and directly receives its stale unrestricted result. Call Forget during eviction so post-eviction readers start a fresh source lookup; the existing concurrency test should also issue a second read while the old load remains blocked to cover this path.

<details>

> ‼️ **IMPORTANT**

```suggestion
func (c *ttlCache[V]) remove(key string) {
	c.mu.Lock()
	c.gen++
	c.lru.Remove(key)
	c.sf.Forget(key)
	c.mu.Unlock()
}
```

</details>

<details>

```
In `history-service/internal/readcache/readcache.go` around line 75: Incrementing the generation prevents an old load from being stored, but it does not detach that load from the singleflight group. If a full-access load begins before removal, then Evict runs, a history request arriving afterward still joins the old flight and directly receives its stale unrestricted result. Call Forget during eviction so post-eviction readers start a fresh source lookup; the existing concurrency test should also issue a second read while the old load remains blocked to cover this path.

```

</details>

---

history-service/cmd/subinvalidate.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// startSubCacheInvalidation subscribes to every account's subscription.update
+// and evicts history-service's access-check cache on membership changes, closing
+// the stale-full-access window (#414). Non-queue core NATS: every instance must
+// receive every event to evict its own local cache. Best-effort — a dropped
+// event just falls back to the cache TTL. nc.Drain at shutdown tears the
+// subscription down, so the returned handle is discarded.
+func startSubCacheInvalidation(ctx context.Context, nc *o11ynats.Conn, ev evictor) error {
+	_, err := nc.Subscribe(ctx, subject.SubscriptionUpdateWildcard(), func(_ context.Context, msg *nats.Msg) {
```
julianshen, line 78: _⚡ performance_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`performance-and-concurrency`_

**Global wildcard broadcasts high-frequency read updates to every replica**

This non-queue wildcard delivers every account's subscription update to every history-service replica, although only "added" and "removed" are useful. The same subject also carries the explicitly high-frequency "read" action, so normal read traffic is amplified across all replicas and each delivery incurs subject parsing and JSON decoding. Under sufficient traffic, the async subscription can become a slow consumer and drop or delay the membership events needed for invalidation. Publish boundary changes on a dedicated subject and subscribe here only to that subject.

<details>

```
In `history-service/cmd/subinvalidate.go` around line 78: This non-queue wildcard delivers every account's subscription update to every history-service replica, although only "added" and "removed" are useful. The same subject also carries the explicitly high-frequency "read" action, so normal read traffic is amplified across all replicas and each delivery incurs subject parsing and JSON decoding. Under sufficient traffic, the async subscription can become a slow consumer and drop or delay the membership events needed for invalidation. Publish boundary changes on a dedicated subject and subscribe here only to that subject.

```

</details>

---

history-service/internal/readcache/readcache.go
```diff
@@ -39,6 +42,15 @@ type ttlCache[V any] struct {
 	lru     *lru.LRU[string, V]
 	sf      singleflight.Group
 	metrics Recorder
+	// mu guards gen and the check-then-Add in getOrLoad so an eviction cannot
+	// interleave between the generation check and the store. gen bumps on every
+	// remove; a load that raced an eviction sees the changed gen and skips its
+	// now-stale store. Global, not per-key: an eviction of any key skips any load
+	// in flight at that moment — correct, and at membership-change rates the rare
+	// extra reload is cheaper than tracking a per-key counter.
+	// ponytail: global generation, per-key if evict rate ever dwarfs read misses.
```
julianshen, line 63: _⚡ performance_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`performance-and-concurrency`_

**Global generation can prevent unrelated cache entries from filling**

Every add/remove event for any account increments this single generation on every replica. Consequently, an eviction for key B while a source load for key A is in flight causes A to skip its cache store. When aggregate membership-change frequency approaches or exceeds source-load latency, nearly every miss overlaps an unrelated eviction, so the cache repeatedly reloads MongoDB; singleflight only coalesces callers within each individual load wave. Track invalidation versions per key with bounded lifecycle, or use keyed/striped synchronization so only eviction of the loaded key suppresses its store.

<details>

```
In `history-service/internal/readcache/readcache.go` around line 51: Every add/remove event for any account increments this single generation on every replica. Consequently, an eviction for key B while a source load for key A is in flight causes A to skip its cache store. When aggregate membership-change frequency approaches or exceeds source-load latency, nearly every miss overlaps an unrelated eviction, so the cache repeatedly reloads MongoDB; singleflight only coalesces callers within each individual load wave. Track invalidation versions per key with bounded lifecycle, or use keyed/striped synchronization so only eviction of the loaded key suppresses its store.

```

</details>

---

julianshen (commented):
### tGDBot review of `fa6fc34` — 7 findings in 5 files

5 🔴 blocking · 2 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/415#issuecomment-5455391199), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

chenjr0719:
Thanks for the thorough review — these are valid and I'm holding this before it merges. Two of the blockers share a root cause I under-covered: the generation guard only prevents *storing* the stale value, it doesn't protect the readers of the in-flight load.

1. **In-flight singleflight waiters still receive the stale result** (readcache.go correctness/security): `getOrLoad` returns the loaded value to every waiter even when the gen check skips the `Add`, so a request that joins a full-access load in flight during an `Evict` still gets the unrestricted result. Needs `sf.Forget(key)` on eviction + fail-closed on a gen mismatch (discard + re-read the source), not just skip-store.
2. **Core NATS has no replay across a disconnect** (subinvalidate.go): a replica that misses an `added`/`removed` while disconnected keeps serving the stale grant until TTL. Needs the cache cleared on reconnect/subscriber-loss, or a replayable per-replica consumer gated on catch-up.

I'll also address the two warnings (a dedicated subject for boundary changes instead of the high-frequency wildcard, and the global-generation over-invalidation). Reworking now — will re-request review. Holding merge until then.

chenjr0719:
Addressed the blocking findings on `fa6fc34` (now rebased onto `main`, head `2b166106`):

- **In-flight-load return hole** (readcache.go): the generation guard previously only skipped the *store*, so a load that raced an eviction still *returned* its pre-eviction positive grant to every singleflight waiter, and a request joining the in-flight load after the eviction got it too. Closed in two parts: `remove()` now calls `sf.Forget(key)` under the same lock as the gen bump (a caller arriving after the eviction starts a fresh load instead of joining the doomed one), and `getOrLoad()` re-reads the source when the generation changed during a load, looping until the generation is stable across a full load — bounded, and failing **closed** (deny) if it cannot stabilize rather than returning a possibly-stale grant.
- **Reconnect gap** (subinvalidate.go / main.go): core NATS gives no replay, so an instance that missed an eviction while disconnected kept serving the stale entry until TTL. Added `SubscriptionCache.Purge()` (drops all entries + bumps the generation) and wired it into a NATS reconnect handler (chained onto the existing one) so a reconnected replica starts cold.

New `-race` tests cover both: a second concurrent reader plus the in-flight loader must both observe the narrowed boundary after an evict, and a purge must force a fresh source read.

On the two warnings:

- **Global generation** — kept as-is; it is an intentional, documented tradeoff (at membership-change rates the rare extra reload is cheaper than per-key version bookkeeping). Noted inline.
- **Dedicated subject for boundary changes** — this one is a **follow-up**, not addressed here, because it cannot be done inside history-service alone: it requires the producer to publish `added`/`removed` onto a new dedicated subject instead of the shared `subscription.update` wildcard (which also carries high-frequency `read`). That is a cross-service change and is out of scope for this PR; tracking it separately.

history-service/internal/config/config.go
```diff
@@ -37,6 +37,12 @@ type MongoConfig struct {
 	// primaryPreferred rather than primary: without it, encrypted history cannot be
 	// read at all when there is no primary.
 	KeyReadPreference string `env:"KEY_READ_PREFERENCE" envDefault:"primaryPreferred"`
+	// SubscriptionReadPreference covers the subscriptions collection, which backs
+	// the access-control reads (GetSubscription / GetHistorySharedSince). It must
+	// be read-your-writes: the service default is secondaryPreferred, and a lagging
+	// secondary would let the sub-cache's evict-then-reload re-cache a stale
+	// pre-revocation grant until TTL — silently defeating the eviction.
+	SubscriptionReadPreference string `env:"SUBSCRIPTION_READ_PREFERENCE" envDefault:"primaryPreferred"`
```
julianshen: _🏷️ distributed-consistency_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`distributed-system`, `mongodb`, `silent-failure`, `tgd-review` (4 rules)_

**Primary fallback can re-cache revoked access from a lagging secondary**

`primaryPreferred` does not provide the read-your-writes contract claimed above it. Concrete timeline: a revocation is acknowledged by the primary and its NATS event evicts this cache; the primary then becomes unavailable before a secondary has applied that write; the reload falls back to that secondary, reads the old full-access subscription, and caches it until TTL. The revoked member can consequently read history. Default this access-control collection to `primary` and reject configurable modes such as `primaryPreferred` or `nearest` that may read secondaries; a primary outage should fail the lookup closed rather than return stale authorization state.

<details>

- **`mongodb`**:45: The subscriptions collection backs authorization, but `primaryPreferred` falls back to a potentially lagging secondary when no primary is available. Concrete path: a removal is committed and its NATS eviction is processed, the MongoDB primary then becomes unavailable, and the next history request reloads from a secondary that has not applied the removal; the old positive grant is returned and cached until TTL, exposing revoked history. Default this collection to `primary` so loss of the primary fails closed rather than serving stale authorization data; consider rejecting secondary-capable overrides if this guarantee is mandatory.

<details>

```text
	SubscriptionReadPreference string `env:"SUBSCRIPTION_READ_PREFERENCE" envDefault:"primary"`
```

</details>

- **`tgd-review`**:46: The comment requires read-your-writes behavior, but `primaryPreferred` falls back to a secondary when no primary is available. Reachable authorization path: a revocation is committed and its invalidation event evicts the cache, the primary becomes unavailable before every secondary has applied that write, and the reload selects a lagging secondary and caches the old full-access subscription until TTL. Use `primary` for this security-sensitive collection, or explicitly resolve the availability/security tradeoff and prevent configurations with secondary fallback if immediate revocation is required.
- **`silent-failure`**:46: `primaryPreferred` does not satisfy the stated read-your-writes requirement. Reachable path: a revocation is acknowledged on the old primary, that primary becomes unavailable before the invalidation-triggered reload, and a lagging secondary still contains the pre-revocation full-access subscription. The reload then succeeds and caches that stale grant until TTL, allowing unauthorized history reads. Require `primary` for this access-control collection and reject weaker configured modes so loss of an authoritative read propagates as an error instead of fallback success.

</details>

<details>

```
In `history-service/internal/config/config.go` around line 45: `primaryPreferred` does not provide the read-your-writes contract claimed above it. Concrete timeline: a revocation is acknowledged by the primary and its NATS event evicts this cache; the primary then becomes unavailable before a secondary has applied that write; the reload falls back to that secondary, reads the old full-access subscription, and caches it until TTL. The revoked member can consequently read history. Default this access-control collection to `primary` and reject configurable modes such as `primaryPreferred` or `nearest` that may read secondaries; a primary outage should fail the lookup closed rather than return stale authorization state.

```

</details>

---

history-service/cmd/main.go
```diff
@@ -190,6 +198,38 @@ func main() {
...
+			os.Exit(1)
+		}
+		// Core NATS gives no replay: an eviction fanned out while this instance was
+		// disconnected is lost, and the stale full-access entry would keep
+		// authorizing reads until TTL. Purge the cache on reconnect so we resume
+		// cold and reload every boundary from the source. Chain the existing
+		// reconnect handler (connection metrics) rather than replace it.
+		raw := nc.NatsConn()
+		prevReconnect := raw.Opts.ReconnectedCB
+		raw.SetReconnectHandler(func(c *nats.Conn) {
```
julianshen, line 309: _🏷️ nats-lifecycle_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`nats`_

**Reconnect callbacks do not gate delivery before the cache purge**

NATS reconnect callbacks are asynchronous and run after subscriptions have been restored, so `sc.Purge()` is not a barrier before new history requests can be dispatched. Concrete path: the disconnect callback purges; an already-running history handler subsequently reloads and caches an old positive grant; the member is revoked while disconnected and that update is missed; after reconnection, a history request reaches the restored subscription before `ReconnectedCB` executes and reads the stale grant. Calling the previous callback before purging widens this window. Use a replayable per-instance consumer and gate request readiness until it catches up, or stop caching positive grants across connection transitions with a delivery barrier that cannot admit requests before purge completes.

<details>

```
In `history-service/cmd/main.go` around line 215: NATS reconnect callbacks are asynchronous and run after subscriptions have been restored, so `sc.Purge()` is not a barrier before new history requests can be dispatched. Concrete path: the disconnect callback purges; an already-running history handler subsequently reloads and caches an old positive grant; the member is revoked while disconnected and that update is missed; after reconnection, a history request reaches the restored subscription before `ReconnectedCB` executes and reads the stale grant. Calling the previous callback before purging widens this window. Use a replayable per-instance consumer and gate request readiness until it catches up, or stop caching positive grants across connection transitions with a delivery barrier that cannot admit requests before purge completes.

```

</details>

---

history-service/internal/readcache/readcache.go
```diff
@@ -73,16 +106,46 @@ func (c *ttlCache[V]) getOrLoad(ctx context.Context, key string, load func(conte
...
+		// Load, then confirm no eviction (a gen bump) raced this load. If one did,
+		// the value we just read may predate the membership change that triggered
+		// the eviction, so it must NOT be returned to ANY singleflight waiter — a
+		// stale positive grant here lets a revoked member read history. Re-read
+		// against the new generation and re-check, looping until the generation is
+		// stable across a full load. Bounded: if evictions keep racing every
+		// attempt, fail closed — return the zero value, never the stale grant. For
+		// SubscriptionCache the zero subEntry is subscribed=false (access denied);
+		// for the other caches it is simply an un-stored miss that re-resolves.
+		const maxLoadAttempts = 5
```
julianshen, line 207: _🏷️ availability_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`, `tgd-review` (2 rules)_

**Unrelated evictions can falsely deny valid history access**

The generation is global to the whole subscription cache, while this loop allows only five attempts. Concrete interleaving: account A misses and starts a MongoDB load; an update for unrelated account B increments `gen`; A retries, and four more unrelated membership updates overlap the following loads. The loop then returns the zero `subEntry` with a nil error, which `GetHistorySharedSince` interprets as `subscribed=false`, denying a valid member and amplifying the original request into five database reads. Since every account's updates are broadcast to every replica, this can occur during ordinary membership bursts. Track generations per cache key, with bounded cleanup, so only eviction of A's entry invalidates A's load.

<details>

- **`tgd-review`**:55: The generation is global to the entire subscription cache. Therefore an eviction for any account invalidates every in-flight load, including unrelated keys. If five membership changes overlap the five attempts for one lookup, `getOrLoad` returns the zero `subEntry` with no error, which `GetHistorySharedSince` interprets as not subscribed. At sufficient global membership-change volume, valid members receive incorrect access denials and each miss can issue five MongoDB reads. Track generations per key, or use keyed/striped synchronization so only eviction of the loaded key forces a retry.

</details>

<details>

```
In `history-service/internal/readcache/readcache.go` around line 118: The generation is global to the whole subscription cache, while this loop allows only five attempts. Concrete interleaving: account A misses and starts a MongoDB load; an update for unrelated account B increments `gen`; A retries, and four more unrelated membership updates overlap the following loads. The loop then returns the zero `subEntry` with a nil error, which `GetHistorySharedSince` interprets as `subscribed=false`, denying a valid member and amplifying the original request into five database reads. Since every account's updates are broadcast to every replica, this can occur during ordinary membership bursts. Track generations per cache key, with bounded cleanup, so only eviction of A's entry invalidates A's load.

```

</details>

---

history-service/internal/readcache/readcache.go
```diff
@@ -59,8 +71,29 @@ func newTTLCache[V any](size int, ttl time.Duration, rec Recorder) (*ttlCache[V]
...
-func (c *ttlCache[V]) remove(key string) { c.lru.Remove(key) }
+// remove drops key if present; absent keys are a no-op. Forget detaches any
+// in-flight singleflight load for key so a caller arriving after the eviction
+// starts a FRESH source read instead of joining the doomed in-flight one (which
+// read the pre-eviction boundary). The gen bump makes that doomed load skip its
+// store and, in getOrLoad, re-read rather than return its now-stale value.
+func (c *ttlCache[V]) remove(key string) {
+	c.mu.Lock()
+	c.gen++
+	c.lru.Remove(key)
```
julianshen, line 114: _⚡ memory-exhaustion_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`oom`_

**Forgetting active loads removes the per-key memory bound**

Each Evict calls Forget without canceling the existing singleflight operation. A reachable sequence is: a slow history cache miss starts a MongoDB load, an added/removed event evicts that key, another history request starts a new load, and this repeats while earlier loads remain active. Forgotten calls retain their goroutines, result channels, closures, and detached request contexts; the new retry loop can keep each alive for up to five separate fetchTimeout periods. Sustained membership churn plus reads during MongoDB latency can therefore accumulate unbounded concurrent loads, causing GC thrashing, process OOM/container eviction, and cascading database pressure. Preserve the required post-eviction freshness while adding cancellation for superseded per-key loads and/or a hard semaphore/backpressure limit on concurrent source loads.

<details>

```
In `history-service/internal/readcache/readcache.go` around line 82: Each Evict calls Forget without canceling the existing singleflight operation. A reachable sequence is: a slow history cache miss starts a MongoDB load, an added/removed event evicts that key, another history request starts a new load, and this repeats while earlier loads remain active. Forgotten calls retain their goroutines, result channels, closures, and detached request contexts; the new retry loop can keep each alive for up to five separate fetchTimeout periods. Sustained membership churn plus reads during MongoDB latency can therefore accumulate unbounded concurrent loads, causing GC thrashing, process OOM/container eviction, and cascading database pressure. Preserve the required post-eviction freshness while adding cancellation for superseded per-key loads and/or a hard semaphore/backpressure limit on concurrent source loads.

```

</details>

---

history-service/internal/readcache/readcache.go
```diff
@@ -73,16 +106,46 @@ func (c *ttlCache[V]) getOrLoad(ctx context.Context, key string, load func(conte
...
+				c.lru.Add(key, val)
+			}
+			c.mu.Unlock()
+			if stable {
+				return val, nil
+			}
+			// An eviction landed during this load: loop and re-read the source.
 		}
-		return val, nil
+		// Fail closed: could not complete a load without a concurrent eviction.
```
julianshen: _🏷️ error-handling_ | _🟠 Warning_ | _⚡ Quick fix_ | _`silent-failure`_

**Exhausted cache retries are returned as a successful zero value**

If five generation changes overlap a load, the retry loop exhausts and returns `zero, nil`. Because this generic cache also backs room times and minimum-last-seen metadata, the current caller receives fabricated zero metadata as a successful result; it does not re-resolve as the comment claims. Subscription callers similarly cannot distinguish an operational failure from a genuine denial. Return a non-nil error after exhausting retries so handlers do not consume the fallback as valid data.

<details>

> ‼️ **IMPORTANT**

```suggestion
		return zero, fmt.Errorf("load cache entry: invalidated during all %d attempts", maxLoadAttempts)
```

</details>

<details>

```
In `history-service/internal/readcache/readcache.go` around line 145: If five generation changes overlap a load, the retry loop exhausts and returns `zero, nil`. Because this generic cache also backs room times and minimum-last-seen metadata, the current caller receives fabricated zero metadata as a successful result; it does not re-resolve as the comment claims. Subscription callers similarly cannot distinguish an operational failure from a genuine denial. Return a non-nil error after exhausting retries so handlers do not consume the fallback as valid data.

```

</details>

---

julianshen (commented):
### tGDBot review of `df7f877` — 9 findings in 9 files

5 🔴 blocking · 4 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/415#issuecomment-5455391199), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

chenjr0719:
Holding before merge — the latest automated review posted findings I need to work through. Will re-request once addressed.

history-service/cmd/subinvalidate.go
```diff
@@ -0,0 +1,85 @@
...
+		evictOnSubscriptionUpdate(ev, msg.Subject, msg.Data)
+	})
```
julianshen, line 80: _🔒 security_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`, `silent-failure` (2 rules)_

**Dropped invalidations leave revoked access cached**

This non-queue Core NATS wildcard receives every subscription update, including the explicitly high-frequency `read` action, and filters only after subject parsing and JSON decoding. If the asynchronous subscription becomes a slow consumer, NATS can drop a removal or restricted re-add while the connection remains established. The new disconnect handler therefore never suspends the cache, and the callback cannot observe the missing message; an existing `subscribed=true` entry continues authorizing revoked history access until the two-minute default TTL. Retain the subscription and treat its slow-consumer/asynchronous errors as cache-integrity failures by suspending and purging while the subscription is re-established, or move boundary changes to a dedicated replayable consumer and gate cache use until catch-up completes.

<details>

- **`distributed-system`**:81: This Core NATS wildcard receives every subscription update, including the explicitly high-frequency `read` action, on every replica. Concrete path: a replica has a full-access grant cached; read updates fill the asynchronous subscription's pending queue; Core NATS drops messages as a slow-consumer condition without disconnecting; a subsequent `removed` or restrictive `added` update is dropped; because no disconnect callback suspends the cache, the next history request hits the stale grant and exposes revoked history until TTL expiry. Move access-boundary changes to a dedicated low-volume subject and consume them through a durable per-replica mechanism, or suspend/purge the cache whenever delivery loss is reported.

</details>

<details>

```
In `history-service/cmd/subinvalidate.go` around line 80: This non-queue Core NATS wildcard receives every subscription update, including the explicitly high-frequency `read` action, and filters only after subject parsing and JSON decoding. If the asynchronous subscription becomes a slow consumer, NATS can drop a removal or restricted re-add while the connection remains established. The new disconnect handler therefore never suspends the cache, and the callback cannot observe the missing message; an existing `subscribed=true` entry continues authorizing revoked history access until the two-minute default TTL. Retain the subscription and treat its slow-consumer/asynchronous errors as cache-integrity failures by suspending and purging while the subscription is re-established, or move boundary changes to a dedicated replayable consumer and gate cache use until catch-up completes.

```

</details>

---

julianshen (commented):
### tGDBot review of `94759ea` — 16 findings in 10 files

5 🔴 blocking · 8 🟠 warning · 3 🔵 suggestion
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/415#issuecomment-5455391199), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

history-service/internal/config/config.go
```diff
@@ -48,6 +48,13 @@ type MongoConfig struct {
 	// primaryPreferred rather than primary: without it, encrypted history cannot be
 	// read at all when there is no primary.
 	KeyReadPreference string `env:"KEY_READ_PREFERENCE" envDefault:"primaryPreferred"`
+	// SubscriptionReadPreference covers the subscriptions collection, which backs
+	// the access-control reads (GetSubscription / GetHistorySharedSince). It pins to
+	// strict primary — not primaryPreferred — so the read is read-your-writes and
+	// fails closed on a primary outage: primaryPreferred would fall back to a lagging
+	// secondary that could re-cache a stale pre-revocation grant until TTL, silently
+	// defeating the eviction. Env-overridable for operators who accept that trade-off.
+	SubscriptionReadPreference string `env:"SUBSCRIPTION_READ_PREFERENCE" envDefault:"primary"`
```
julianshen, line 57: _🏷️ mongodb-consistency_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`mongodb`_

**Strict-primary authorization reads can be configured back to stale reads.**

Validation accepts every MongoDB read preference, and the new test explicitly accepts `nearest`. If an operator selects `nearest`, `secondaryPreferred`, or `primaryPreferred`, a removal can be committed on the primary and trigger eviction, but the subsequent `subscriptions` lookup can read the old full-access document from a lagging secondary. That positive grant is then cached for the subscription TTL, allowing the removed member to read history. Reject every mode except `primary`, or hard-code `readpref.Primary()` for this authorization repository.

<details>

```
In `history-service/internal/config/config.go` around line 57: Validation accepts every MongoDB read preference, and the new test explicitly accepts `nearest`. If an operator selects `nearest`, `secondaryPreferred`, or `primaryPreferred`, a removal can be committed on the primary and trigger eviction, but the subsequent `subscriptions` lookup can read the old full-access document from a lagging secondary. That positive grant is then cached for the subscription TTL, allowing the removed member to read history. Reject every mode except `primary`, or hard-code `readpref.Primary()` for this authorization repository.

```

</details>

---

history-service/cmd/main.go
```diff
@@ -274,12 +282,45 @@ func main() {
...
+			if prevReconnect != nil {
+				prevReconnect(c)
+			}
+			sc.Resume()
+			slog.Warn("nats reconnected: purged and re-armed subscription access cache to drop possibly-missed evictions")
+		})
+		prevDisconnect := raw.Opts.DisconnectedErrCB
```
julianshen, line 316: _🏷️ nats-lifecycle_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`nats`_

**Async disconnect callbacks leave a stale-cache delivery window**

The cache is suspended only when NATS executes DisconnectedErrCB, but lifecycle callbacks are asynchronous and are not a delivery barrier for subscription handlers. A concrete interleaving is: this replica caches full access, disconnects and queues this callback, a revocation is published and missed, the connection quickly restores its subscriptions, and a history request handler reads the still-valid stale entry before the queued disconnect callback runs. The later Suspend/Resume calls cannot retract the unauthorized result. Do not rely on connection callbacks as the barrier: either disable positive-grant caching, use a replayable per-instance consumer and gate history request delivery until catch-up, or restructure subscription lifecycle so request subscriptions cannot resume until cache suspension/purge is synchronously established.

<details>

```
In `history-service/cmd/main.go` around line 316: The cache is suspended only when NATS executes DisconnectedErrCB, but lifecycle callbacks are asynchronous and are not a delivery barrier for subscription handlers. A concrete interleaving is: this replica caches full access, disconnects and queues this callback, a revocation is published and missed, the connection quickly restores its subscriptions, and a history request handler reads the still-valid stale entry before the queued disconnect callback runs. The later Suspend/Resume calls cannot retract the unauthorized result. Do not rely on connection callbacks as the barrier: either disable positive-grant caching, use a replayable per-instance consumer and gate history request delivery until catch-up, or restructure subscription lifecycle so request subscriptions cannot resume until cache suspension/purge is synchronously established.

```

</details>

---

history-service/cmd/subinvalidate.go
```diff
@@ -0,0 +1,85 @@
...
+	if !isAccessBoundaryAction(env.Action) || env.Subscription.RoomID == "" {
+		return
+	}
+	ev.Evict(account, env.Subscription.RoomID)
+}
+
+// startSubCacheInvalidation subscribes to every account's subscription.update
+// and evicts history-service's access-check cache on membership changes, closing
+// the stale-full-access window (#414). Non-queue core NATS: every instance must
+// receive every event to evict its own local cache. Best-effort — a dropped
```
julianshen, line 74: _🔒 security_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`tgd-review`_

**Best-effort Core NATS still leaves revoked grants cached**

The comment explicitly accepts dropped invalidations falling back to the cache TTL. A reachable path remains: a full-access grant is cached, membership is revoked in MongoDB, but the Core NATS event is dropped or its publish fails while this history replica remains connected. No disconnect barrier or later replay removes the grant, so history remains accessible until the two-minute default TTL expires. Use a durable, replayable invalidation tied to the membership write, or stop caching positive grants when reliable invalidation cannot be guaranteed.

<details>

```
In `history-service/cmd/subinvalidate.go` around line 74: The comment explicitly accepts dropped invalidations falling back to the cache TTL. A reachable path remains: a full-access grant is cached, membership is revoked in MongoDB, but the Core NATS event is dropped or its publish fails while this history replica remains connected. No disconnect barrier or later replay removes the grant, so history remains accessible until the two-minute default TTL expires. Use a durable, replayable invalidation tied to the membership write, or stop caching positive grants when reliable invalidation cannot be guaranteed.

```

</details>

---

history-service/internal/readcache/readcache.go
```diff
@@ -59,30 +103,148 @@ func newTTLCache[V any](size int, ttl time.Duration, rec Recorder) (*ttlCache[V]
...
+				waitCtx, waitCancel := context.WithTimeout(context.WithoutCancel(ctx), fetchTimeout)
+				select {
+				case c.sem <- struct{}{}:
+					waitCancel()
+				case <-waitCtx.Done():
+					waitCancel()
+					var zero V
+					return zero, fmt.Errorf("readcache: load slots exhausted: %w", waitCtx.Err())
+				}
+			}
```
julianshen, line 196: _⚡ memory-exhaustion_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`, `oom`, `performance-and-concurrency`, `tgd-review` (4 rules)_

**The semaphore leaves unbounded miss goroutines queued for 10 seconds**

Each distinct `(account, roomID)` miss creates a new `singleflight.DoChan` goroutine before this semaphore is acquired. Once 256 loads are active, every additional distinct-room history request enters the `default` branch and remains detached for up to `fetchTimeout`; caller cancellation cannot release it because `context.WithoutCancel(ctx)` is used. Singleflight retains the goroutine, result channel, closure, context, and timer meanwhile. Sustained high-cardinality misses therefore accumulate roughly request-rate × 10 seconds of waiting objects, causing GC thrashing and potentially process OOM or container eviction despite the source-call cap. Reject immediately when no slot is available, or move admission ahead of goroutine creation with a genuinely bounded queue.

<details>

> ‼️ **IMPORTANT**

```suggestion
		// Bound concurrent source loads so a burst of distinct-key misses can't
		// keep detached load goroutines queued behind slow Mongo. Fail closed
		// immediately when saturated rather than retaining each miss until timeout.
		if c.sem != nil {
			select {
			case c.sem <- struct{}{}:
				defer func() { <-c.sem }()
			default:
				var zero V
				return zero, errors.New("readcache: load slots exhausted")
			}
		}
```

</details>

<details>

- **`performance-and-concurrency`**:177: `singleflight.DoChan` starts one goroutine for every distinct-key miss before that goroutine acquires `c.sem`. A burst exceeding 256 distinct keys therefore leaves every excess flight, timer, result channel, and detached context waiting for up to 10 seconds; caller cancellation cannot remove them because `WithoutCancel` is used. Repeated eviction of a waiting key can also `Forget` the uncancellable flight and create another. At sustained high miss rates this can cause substantial memory growth even though Mongo concurrency remains capped. Apply bounded admission before spawning detached flights, or use a bounded flight scheduler that preserves same-key coalescing and rejects/cancels excess work.
- **`distributed-system`**:185: The semaphore is acquired inside the function passed to `singleflight.DoChan`, after a detached goroutine has already been created for each distinct key. If MongoDB stalls and 256 loads occupy the slots, every subsequent distinct-key miss creates another goroutine and waits up to 10 seconds using `context.WithoutCancel`; caller cancellation does not release that work. At sustained request volume, retained goroutines, channels, and contexts grow with the ten-second arrival window and can exhaust the process even though source calls are capped. Fail fast when no slot is available, or admit work through a genuinely bounded queue before starting detached singleflight work.
- **`tgd-review`**:185: DoChan starts a detached goroutine before this semaphore is acquired. Once 256 loads are active, every distinct-key miss still creates another singleflight goroutine that retains its key, closure, result channel, and timeout context for up to ten seconds while waiting. A sufficiently large burst can therefore retain an unbounded number of goroutines despite the stated cap. Bound admission before spawning detached work, use a bounded waiting queue, or fail immediately when no slot is available.

</details>

<details>

```
In `history-service/internal/readcache/readcache.go` around line 175: Each distinct `(account, roomID)` miss creates a new `singleflight.DoChan` goroutine before this semaphore is acquired. Once 256 loads are active, every additional distinct-room history request enters the `default` branch and remains detached for up to `fetchTimeout`; caller cancellation cannot release it because `context.WithoutCancel(ctx)` is used. Singleflight retains the goroutine, result channel, closure, context, and timer meanwhile. Sustained high-cardinality misses therefore accumulate roughly request-rate × 10 seconds of waiting objects, causing GC thrashing and potentially process OOM or container eviction despite the source-call cap. Reject immediately when no slot is available, or move admission ahead of goroutine creation with a genuinely bounded queue.

```

</details>

---

history-service/internal/readcache/readcache.go
```diff
@@ -39,18 +54,47 @@ type ttlCache[V any] struct {
...
 	metrics Recorder
+	// mu guards gen and the check-then-Add in getOrLoad so an eviction cannot
+	// interleave between the generation check and the store. gen bumps on every
+	// remove; a load that raced an eviction sees the changed gen and skips its
+	// now-stale store. Global, not per-key: an eviction of any key skips any load
+	// in flight at that moment — correct, and at membership-change rates the rare
+	// extra reload is cheaper than tracking a per-key counter.
+	// ponytail: global generation, per-key if evict rate ever dwarfs read misses.
+	mu       sync.Mutex
+	gen      uint64
```
julianshen, line 65: _🏷️ mongodb-operational_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`mongodb`_

**Unrelated evictions repeatedly invalidate MongoDB subscription loads.**

`gen` is global to the entire cache, while every account's membership event calls `remove` and increments it. A cold lookup for account A can complete its MongoDB read, observe an eviction for unrelated account B, and issue the read again. If five unrelated updates overlap its five attempts, the valid lookup returns an error after consuming five primary reads. Track invalidation state per cache key, while retaining a separate global epoch only for purge/suspend operations.

<details>

```
In `history-service/internal/readcache/readcache.go` around line 65: `gen` is global to the entire cache, while every account's membership event calls `remove` and increments it. A cold lookup for account A can complete its MongoDB read, observe an eviction for unrelated account B, and issue the read again. If five unrelated updates overlap its five attempts, the valid lookup returns an error after consuming five primary reads. Track invalidation state per cache key, while retaining a separate global epoch only for purge/suspend operations.

```

</details>

---

history-service/internal/readcache/readcache.go
```diff
@@ -59,30 +103,148 @@ func newTTLCache[V any](size int, ttl time.Duration, rec Recorder) (*ttlCache[V]
...
+			}
+			defer func() { <-c.sem }()
 		}
-		if store {
-			c.lru.Add(key, val)
+		// Load, then confirm no eviction (a gen bump) raced this load. If one did,
+		// the value we just read may predate the membership change that triggered
+		// the eviction, so it must NOT be returned to ANY singleflight waiter — a
+		// stale positive grant here lets a revoked member read history. Re-read
+		// against the new generation and re-check, looping until the generation is
```
julianshen, line 203: _🏷️ availability_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`nats`_

**Unrelated invalidations can exhaust another key's retries**

The generation is global, so every processed added/removed event invalidates every in-flight cache load. If a valid lookup for key A overlaps five updates for unrelated keys, each retry observes a new generation and the lookup ends with an error after five MongoDB reads even though A never changed. Because this consumer receives all accounts' boundary events on every replica, ordinary aggregate membership churn can cause avoidable history-request failures and database amplification. Track invalidation generations per key, with bounded cleanup, so only eviction of the loaded key forces a retry.

<details>

```
In `history-service/internal/readcache/readcache.go` around line 203: The generation is global, so every processed added/removed event invalidates every in-flight cache load. If a valid lookup for key A overlaps five updates for unrelated keys, each retry observes a new generation and the lookup ends with an error after five MongoDB reads even though A never changed. Because this consumer receives all accounts' boundary events on every replica, ordinary aggregate membership churn can cause avoidable history-request failures and database amplification. Track invalidation generations per key, with bounded cleanup, so only eviction of the loaded key forces a retry.

```

</details>

---

history-service/internal/readcache/readcache.go
```diff
@@ -59,30 +103,148 @@ func newTTLCache[V any](size int, ttl time.Duration, rec Recorder) (*ttlCache[V]
...
+		// generic-V caller would read as real data).
+		const maxLoadAttempts = 5
+		for attempt := 0; attempt < maxLoadAttempts; attempt++ {
+			fetchCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), fetchTimeout)
+			// Register the cancel so a concurrent remove/purge for this key cancels
+			// this now-superseded load rather than letting it run to fetchTimeout.
+			h := &loadHandle{cancel: cancel}
+			// One critical section: snapshot gen before loading (so a remove during
+			// the load is detected) and publish the cancel handle together.
+			c.mu.Lock()
```
julianshen, line 215: _⚡ performance_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`performance-and-concurrency`_

**Unrelated evictions still multiply source loads.**

The generation remains cache-wide, so an eviction for key B invalidates an in-flight load for key A. During an aggregate membership-change burst, each active miss can perform up to five Mongo reads while retaining its semaphore slot; five overlapping unrelated evictions then fail the lookup, while queued misses wait or time out. This amplifies database load and can starve all load slots. Track invalidation versions per key, retaining a separate global epoch for purge/suspend, so only eviction of the loaded key forces a retry.

<details>

```
In `history-service/internal/readcache/readcache.go` around line 215: The generation remains cache-wide, so an eviction for key B invalidates an in-flight load for key A. During an aggregate membership-change burst, each active miss can perform up to five Mongo reads while retaining its semaphore slot; five overlapping unrelated evictions then fail the lookup, while queued misses wait or time out. This amplifies database load and can starve all load slots. Track invalidation versions per key, retaining a separate global epoch for purge/suspend, so only eviction of the loaded key forces a retry.

```

</details>

---

history-service/internal/readcache/readcache.go
```diff
@@ -39,18 +54,47 @@ type ttlCache[V any] struct {
...
+	// mu guards gen and the check-then-Add in getOrLoad so an eviction cannot
+	// interleave between the generation check and the store. gen bumps on every
+	// remove; a load that raced an eviction sees the changed gen and skips its
+	// now-stale store. Global, not per-key: an eviction of any key skips any load
+	// in flight at that moment — correct, and at membership-change rates the rare
+	// extra reload is cheaper than tracking a per-key counter.
+	// ponytail: global generation, per-key if evict rate ever dwarfs read misses.
+	mu       sync.Mutex
+	gen      uint64
+	inflight map[string]*loadHandle // key → cancel of the in-flight load, guarded by mu
```
julianshen, line 66: _🏷️ availability_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`tgd-review`_

**One global generation invalidates unrelated cache loads**

Every removal increments the cache-wide generation, so an eviction for key B invalidates a source load for unrelated key A. Each collision forces another database read, and five unrelated invalidations make A return an error. This also affects other users of ttlCache, including preview invalidations. Track versions per key for remove operations, while retaining a separate global epoch for purge/suspend.

<details>

```
In `history-service/internal/readcache/readcache.go` around line 66: Every removal increments the cache-wide generation, so an eviction for key B invalidates a source load for unrelated key A. Each collision forces another database read, and five unrelated invalidations make A return an error. This also affects other users of ttlCache, including preview invalidations. Track versions per key for remove operations, while retaining a separate global epoch for purge/suspend.

```

</details>

---

history-service/internal/readcache/readcache.go
```diff
@@ -59,30 +103,148 @@ func newTTLCache[V any](size int, ttl time.Duration, rec Recorder) (*ttlCache[V]
...
+			if cached, ok := c.lru.Get(key); ok {
+				return cached, nil
+			}
 		}
-		fetchCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), fetchTimeout)
-		defer cancel()
-		val, store, err := load(fetchCtx)
-		if err != nil {
-			return val, err
+		// Bound concurrent source loads so a burst of distinct-key misses can't
```
julianshen, line 177: _🏷️ dead-code_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`dead-code`_

**The nil-semaphore branch can never execute.**

`newTTLCache` unconditionally initializes `c.sem` with `make(chan struct{}, maxInflight)` after replacing every non-positive value with the default. Consequently, every usable cache reaches this site with a non-nil semaphore, making the false arm of `if c.sem != nil` unreachable. Remove the wrapper and update the nearby comments that claim nil can disable the bound.

<details>

```
In `history-service/internal/readcache/readcache.go` around line 177: `newTTLCache` unconditionally initializes `c.sem` with `make(chan struct{}, maxInflight)` after replacing every non-positive value with the default. Consequently, every usable cache reaches this site with a non-nil semaphore, making the false arm of `if c.sem != nil` unreachable. Remove the wrapper and update the nearby comments that claim nil can disable the bound.

```

</details>

---

history-service/internal/readcache/readcache_test.go
```diff
@@ -140,6 +140,421 @@ func TestSubscriptionCache_SingleflightDedupesConcurrentMisses(t *testing.T) {
...
+
+	close(src.release) // let all loads drain
+	for i := 0; i < n; i++ {
+		<-done
+	}
+	assert.LessOrEqual(t, src.peak.Load(), int32(2), "no more than maxInflight loads may run at once")
+}
+
+// raceSafeSubSource is a concurrency-safe source whose access boundary can be
+// flipped from another goroutine, for the barrier race test.
```
julianshen, line 442: _🏷️ dead-code_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`dead-code`_

**The race-safe source's call counter is never observed.**

`raceSafeSubSource.calls` is only incremented by `GetHistorySharedSince`; the barrier race test never reads or asserts it. This leaves an unnecessary atomic operation on every iteration of the stress test. Remove the field and the corresponding `s.calls.Add(1)` call.

<details>

- **`dead-code`**:215: `gatedSubSource.calls` is incremented in `GetHistorySharedSince`, but the added race test never loads or asserts it. The atomic write has no synchronization consumer and cannot affect the test result. Remove both the field and `g.calls.Add(1)`.

</details>

<details>

```
In `history-service/internal/readcache/readcache_test.go` around line 442: `raceSafeSubSource.calls` is only incremented by `GetHistorySharedSince`; the barrier race test never reads or asserts it. This leaves an unnecessary atomic operation on every iteration of the stress test. Remove the field and the corresponding `s.calls.Add(1)` call.

```

</details>

---

chenjr0719:
Holding this — the automated review flagged blocking findings on the latest head; I'm reviewing them before this should merge. Will update once addressed.

chenjr0719:
Thanks for the thorough pass. Disposition of the findings, so nothing is silently dropped:

**Corrected — false positive:** the `config.go:45` "primary fallback can re-cache from a lagging secondary" points at the general `READ_PREFERENCE` (`secondaryPreferred`), which covers message/history reads. The **subscription-access** path does not use it — the subscription repo is built with `SUBSCRIPTION_READ_PREFERENCE` (default `primary`, read-your-writes, no secondary fallback); see `config.go` `SubscriptionReadPreference` and where `NewSubscriptionRepo(db, subReadPref)` is wired in `cmd/main.go`. So the access cache is not sourced from a secondary.

**Accepted tradeoff — documented, not fixed here:** the remaining blocking findings (an in-flight singleflight load returning a stale grant to a late joiner; a core-NATS reconnect/dropped-invalidation preserving a revoked grant to TTL; the async reconnect callback not gating delivery before the purge) all reduce to the same root: this is the maintainer's shipped design of **caching a positive access grant and invalidating it over core NATS**, which has no replay and therefore an inherent, bounded staleness window. This PR *tightens* that window materially versus the prior 2-minute gap — strict `primary` reads, fail-closed on cache error, cancel + concurrency-cap on superseded loads, and a suspend/resume purge barrier on reconnect — but it cannot make a positive-grant cache provably leak-free without changing that design (a replayable per-replica consumer, or not caching positive grants at all). We're consciously accepting the residual bounded window rather than re-architecting the cache in this PR; happy to file a follow-up for a replayable-consumer redesign if the team wants to close it fully.

Net: the fixes are in and correct for the design; the residual is an accepted, documented, bounded tradeoff. Surfacing for human review on that basis.

history-service/cmd/subinvalidate.go
mliu33: Yes, this is my major concern too. Don't subscribe to the subscription.update events which has high amount of traffic.

history-service/cmd/subinvalidate.go
mliu33: No need. This is minor and acceptable error path. If nats disconnected, we accept the stale state until TTL expired, which is not long.

history-service/internal/readcache/readcache.go
mliu33: This needs to be addressed

mliu33 (commented):
Let's make this change on hold to discuss alternative solution, thanks!

history-service/cmd/subinvalidate.go
mliu33: Yes, I do have concern for history service to listen to every sub update event just for cache invalidation. We should think about a better and more efficient approach.

Also, I think this issue is minor to me...

c.c @chenjr0719

PR #432, opened 2026-08-30
Add pagination hasMore flag to ListRoomMembers

yenta:
## Summary
Extends `ListRoomMembers` to return a `hasMore` boolean flag alongside the member list, enabling clients to determine whether additional pages exist without a separate count query. The store over-fetches one row beyond the requested limit to decide the flag, then trims it before returning to the caller.

## Key Changes

- **Store interface & implementation**: `ListRoomMembers` now returns `([]model.RoomMember, bool, error)` instead of `([]model.RoomMember, error)`. The boolean indicates whether at least one row follows the current page.

- **Over-fetch strategy**: Added `pagedLimit()` helper that returns `limit + 1` when a limit is set (enabling the probe), and `0` (unbounded) when limit is nil, zero, negative, or `math.MaxInt` (to avoid overflow). Added `trimOverFetch()` generic helper that drops the probe row and reports whether it existed.

- **Both code paths updated**:
  - `getRoomMembers()` (room_members collection): applies over-fetch to the aggregation pipeline `$limit` stage, trims before enrichment to avoid unnecessary lookups
  - `getRoomSubscriptions()` (subscriptions fallback): applies over-fetch to `SetLimit()`, trims before enrichment

- **Unlimited requests**: When no limit is provided, the query returns all rows and `hasMore` is always `false`.

- **Handler integration**: `listMembers()` now unpacks the hasMore flag and includes it in `ListRoomMembersResponse`.

- **Response model**: `ListRoomMembersResponse` gains a `HasMore` field (always present in JSON, `false` when omitted).

- **Test coverage**: 
  - Added comprehensive unit tests for `pagedLimit()` and `trimOverFetch()` helpers
  - Added integration tests covering: hasMore on first/middle/last pages, unlimited requests, subscriptions fallback path, and enrichment with over-fetch
  - Updated all existing test call sites to unpack the new boolean return value

- **Documentation**: Updated `docs/client-api.md` and `docs/client-api/request-reply.md` to explain the hasMore flag and clarify pagination semantics (advance offset by the limit sent).

## Implementation Details

- The over-fetch is transparent to callers — the probe row is always trimmed before the response leaves the store layer
- Enrichment (user/org display resolution) happens after trimming, so the probe row incurs no lookup cost
- The approach is database-agnostic: both MongoDB aggregation and driver Find() support the pattern
- All mocks and handler tests updated to reflect the new signature

https://claude.ai/code/session_019vQFe3G3j2twBMvpzmktvA

## Summary by CodeRabbit

* **New Features**
  * Added offset-based pagination indicators to member-list responses.
  * Responses now include `hasMore` to show whether additional members are available.
  * Added support for paginated member retrieval while preserving unlimited requests.

* **Documentation**
  * Updated member-list API documentation with pagination rules, offset usage, and response examples.

* **Tests**
  * Added coverage for pagination behavior, response serialization, offsets, limits, and fallback member sources.


mliu33 (approved):
Thanks!

PR #441, opened 2026-08-31
fix(room-worker,bot-room-service): self-heal absent room key on member add

chenjr0719:
## Summary

A channel room with no stored encryption key would permanently fail the member-add key fan-out (`room-worker`) or silently skip it (`bot-room-service`), so a newly-added member never received a room key and could not decrypt — with no retry. Both consuming paths now **self-heal**: when the store holds no key, they mint a fresh one (mirroring the create path) and fan it out, instead of hard-failing or dropping.

The mint is **going-forward-only** and targets a **never-keyed** channel (legacy / migrated, pre-encryption): it unblocks new members so new messages are readable. To keep a lost-key event visible to ops, the existing key-absent metric still fires on the minted-on-absent path. `SetIfAbsent` (not a plain `Set`) is used so concurrent minters — a JetStream redelivery or a racing worker — converge on a single v0 key rather than fanning out rival bytes.

## What's included

- A `keyPairOrHeal` helper in each of the two membership services: `Get` the key, return it if present, otherwise record the key-absent metric, generate a fresh key, and `SetIfAbsent` it.
- The member-add fan-out in both services rewired onto that helper.
- In `bot-room-service`, the key is now resolved **before** any subscription is committed (mirroring room-worker), so a key failure can't strand a just-committed member keyless on retry.
- Unit tests for the healed member-add flow, the helper directly, the key-before-commit ordering, and the present-key/concurrent cases.

## Changes

- `room-worker/handler.go` — `processAddMembers` uses `keyPairOrHeal`; new helper.
- `bot-room-service/handler.go` — `handleAdd` resolves/heals the key up front (before the commit loop) via `keyPairOrHeal`; removes a latent nil-pointer path on the absent-key case.

Left unchanged (already self-heal): both services' remove/rotate paths mint a fresh key via `CommitRotation`'s `SetIfAbsent` fallback. The room-create path (`existingRoomKey`) already provisions a fallback key. The server `key.ensure` RPC already mints if absent.

## Verification

- `go build` / `go test -count=1` on `room-worker`, `bot-room-service` — pass, including the new self-heal, helper, ordering, and concurrent tests.
- `go vet -tags integration` — clean (integration tests compile).
- `golangci-lint` — 0 issues.
- Verified on our test environment: member-add to a keyless channel succeeds and mints the key; the baseline (without this change) permanent-fails.

## Summary by CodeRabbit

- **New Features**
  - Room member additions now automatically recover when a room encryption key is missing.
  - Newly generated keys are securely persisted and delivered to added members.
  - Existing room keys are reused without being replaced.

- **Bug Fixes**
  - Member additions no longer fail solely because a room key is absent.
  - Key retrieval or storage failures stop the operation before partial membership changes occur.
  - Improved consistency for key distribution across room services.

### DIFF
--- a/bot-room-service/handler.go
+++ b/bot-room-service/handler.go
@@ -311,6 +311,16 @@ func (h *handler) handleAdd(c *natsrouter.Context, req BotMembersBatchRequest) (
 	created := h.now()
 	added := []string{}
 	newAccounts := []string{}
+
+	// Resolve (and heal) the room key BEFORE committing any subscription: a key
+	// failure after a member is committed would strand them keyless, since a retry
+	// sees them as a dup (newlyAdded=false) and the fan-out below skips them.
+	// Mirrors room-worker's key-before-commit ordering. See keyPairOrHeal.
+	pair, err := h.keyPairOrHeal(c, roomID)
+	if err != nil {
+		return nil, err
+	}
+
 	for _, userID := range req.UserIDs {
 		u, err := h.store.FindUser(c, userID)
 		if err != nil {
@@ -344,15 +354,9 @@ func (h *handler) handleAdd(c *natsrouter.Context, req BotMembersBatchRequest) (
 		}
 	}
 
-	// Fan out the room's current key only to newly-subscribed accounts — duplicate adds already
+	// Fan out the pre-resolved key only to newly-subscribed accounts — duplicate adds already
 	// have it from their original add; the key isn't re-rotated for adds (mirrors room-worker.buildAndFanOutRoomKey).
 	if len(newAccounts) > 0 {
-		// Self-heal a key-absent room instead of silently skipping the fan-out
-		// (new member couldn't decrypt, no retry). See keyPairOrHeal.
-		pair, err := h.keyPairOrHeal(c, roomID)
-		if err != nil {
-			return nil, err
-		}
 		h.fanOutKey(c, roomID, newAccounts, model.RoomKeyEvent{
 			RoomID:     roomID,
 			Version:    pair.Version,
--- a/bot-room-service/handler.go
+++ b/bot-room-service/handler.go
@@ -15,6 +15,7 @@ import (
 	"github.com/hmchangw/chat/pkg/model"
 	"github.com/hmchangw/chat/pkg/natsrouter"
 	"github.com/hmchangw/chat/pkg/outbox"
+	"github.com/hmchangw/chat/pkg/roomkeymetrics"
 	"github.com/hmchangw/chat/pkg/roomkeysender"
 	"github.com/hmchangw/chat/pkg/roomkeystore"
 	"github.com/hmchangw/chat/pkg/subauthcache"
@@ -346,21 +347,18 @@ func (h *handler) handleAdd(c *natsrouter.Context, req BotMembersBatchRequest) (
 	// Fan out the room's current key only to newly-subscribed accounts — duplicate adds already
 	// have it from their original add; the key isn't re-rotated for adds (mirrors room-worker.buildAndFanOutRoomKey).
 	if len(newAccounts) > 0 {
-		pair, err := h.keyStore.Get(c, roomID)
+		// Self-heal a key-absent room instead of silently skipping the fan-out
+		// (which left the new member unable to decrypt with no retry). Mirrors the
+		// create path's mint; going-forward-only — see keyPairOrHeal.
+		pair, err := h.keyPairOrHeal(c, roomID)
 		if err != nil {
-			if errors.Is(err, roomkeystore.ErrNoCurrentKey) {
-				// Legacy/broken room with no key: nothing to fan out, not fatal.
-				slog.WarnContext(c, "no current key on add-member; skip fan-out", "roomID", roomID)
-			} else {
-				return nil, fmt.Errorf("get room key: %w", err)
-			}
-		} else {
-			h.fanOutKey(c, roomID, newAccounts, model.RoomKeyEvent{
-				RoomID:     roomID,
-				Version:    pair.Version,
-				PrivateKey: pair.KeyPair.PrivateKey,
-			}, "fan out room key on add failed")
+			return nil, fmt.Errorf("get room key: %w", err)
 		}
+		h.fanOutKey(c, roomID, newAccounts, model.RoomKeyEvent{
+			RoomID:     roomID,
+			Version:    pair.Version,
+			PrivateKey: pair.KeyPair.PrivateKey,
+		}, "fan out room key on add failed")
 	}
 
 	// Skip sysmsg on all-dup batches (true no-op: no message, no OUTBOX event).
@@ -556,6 +554,31 @@ func (h *handler) fanOutKey(ctx context.Context, roomID string, accounts []strin
 			slog.WarnContext(ctx, warnMsg, "account", acct, "roomID", roomID, "error", err)
 		}
 	}
+}
+
+// keyPairOrHeal returns the room's current key, minting and persisting a fresh
+// key when the room has none (legacy / never-keyed). Going-forward-only: it
+// unblocks new members but does NOT recover history a lost key had encrypted.
+// RecordKeyAbsent fires on the mint so ops can spot a lost-key event. SetIfAbsent
+// converges concurrent minters on a single v0 key. Mirrors the create path's mint.
+func (h *handler) keyPairOrHeal(ctx context.Context, roomID string) (*roomkeystore.VersionedKeyPair, error) {
+	pair, err := h.keyStore.Get(ctx, roomID)
+	if err != nil && !errors.Is(err, roomkeystore.ErrNoCurrentKey) {
+		return nil, fmt.Errorf("get room key: %w", err)
+	}
+	if pair != nil {
+		return pair, nil
+	}
+	roomkeymetrics.RecordKeyAbsent(ctx, "")
+	fresh, err := roomkeystore.GenerateKeyPair()
+	if err != nil {
+		return nil, fmt.Errorf("generate room key: %w", err)
+	}
+	committed, err := h.keyStore.SetIfAbsent(ctx, roomID, *fresh)
+	if err != nil {
+		return nil, fmt.Errorf("store room key: %w", err)
+	}
+	return committed, nil
 }
 
 // rotateAndFanOut commits before fan-out; a predicted current+1 mislabels keys when removals race.
--- a/bot-room-service/handler.go
+++ b/bot-room-service/handler.go
@@ -348,11 +348,10 @@ func (h *handler) handleAdd(c *natsrouter.Context, req BotMembersBatchRequest) (
 	// have it from their original add; the key isn't re-rotated for adds (mirrors room-worker.buildAndFanOutRoomKey).
 	if len(newAccounts) > 0 {
 		// Self-heal a key-absent room instead of silently skipping the fan-out
-		// (which left the new member unable to decrypt with no retry). Mirrors the
-		// create path's mint; going-forward-only — see keyPairOrHeal.
+		// (new member couldn't decrypt, no retry). See keyPairOrHeal.
 		pair, err := h.keyPairOrHeal(c, roomID)
 		if err != nil {
-			return nil, fmt.Errorf("get room key: %w", err)
+			return nil, err
 		}
 		h.fanOutKey(c, roomID, newAccounts, model.RoomKeyEvent{
 			RoomID:     roomID,
--- a/room-service/handler.go
+++ b/room-service/handler.go
@@ -488,7 +488,24 @@ func (h *Handler) getRoomKey(c *natsrouter.Context) (*model.RoomKeyGetResponse, 
 			return nil, fmt.Errorf("get room key: %w", err)
 		}
 		if existing == nil {
-			return nil, errRoomKeyAbsent
+			// Migrated/legacy channel with no key: mint on demand (roll-out).
+			// Gate to channels (DMs aren't encrypted); SetIfAbsent so concurrent
+			// pulls converge on one key.
+			room, rErr := h.store.GetRoomAppRead(ctx, roomID)
+			if rErr != nil {
+				return nil, fmt.Errorf("get room key: room lookup: %w", rErr)
+			}
+			if room.Type != model.RoomTypeChannel {
+				return nil, errRoomKeyAbsent
+			}
+			roomkeymetrics.RecordKeyAbsent(ctx, "GetHeal")
+			pair, gErr := roomkeystore.GenerateKeyPair()
+			if gErr != nil {
+				return nil, fmt.Errorf("get room key: generate: %w", gErr)
+			}
+			if existing, err = h.keyStore.SetIfAbsent(ctx, roomID, *pair); err != nil {
+				return nil, fmt.Errorf("get room key: heal: %w", err)
+			}
 		}
 		// #nosec G117 -- RoomKeyGetResponse.PrivateKey is the intended payload: on-demand key delivery to the authorized room member over an auth-callout-gated per-user NATS subject, not a leak
 		return &model.RoomKeyGetResponse{
--- a/room-service/handler.go
+++ b/room-service/handler.go
@@ -488,24 +488,7 @@ func (h *Handler) getRoomKey(c *natsrouter.Context) (*model.RoomKeyGetResponse, 
 			return nil, fmt.Errorf("get room key: %w", err)
 		}
 		if existing == nil {
-			// Migrated/legacy channel with no key: mint on demand (roll-out).
-			// Gate to channels (DMs aren't encrypted); SetIfAbsent so concurrent
-			// pulls converge on one key.
-			room, rErr := h.store.GetRoomAppRead(ctx, roomID)
-			if rErr != nil {
-				return nil, fmt.Errorf("get room key: room lookup: %w", rErr)
-			}
-			if room.Type != model.RoomTypeChannel {
-				return nil, errRoomKeyAbsent
-			}
-			roomkeymetrics.RecordKeyAbsent(ctx, "GetHeal")
-			pair, gErr := roomkeystore.GenerateKeyPair()
-			if gErr != nil {
-				return nil, fmt.Errorf("get room key: generate: %w", gErr)
-			}
-			if existing, err = h.keyStore.SetIfAbsent(ctx, roomID, *pair); err != nil {
-				return nil, fmt.Errorf("get room key: heal: %w", err)
-			}
+			return nil, errRoomKeyAbsent
 		}
 		// #nosec G117 -- RoomKeyGetResponse.PrivateKey is the intended payload: on-demand key delivery to the authorized room member over an auth-callout-gated per-user NATS subject, not a leak
 		return &model.RoomKeyGetResponse{
--- a/room-worker/handler.go
+++ b/room-worker/handler.go
@@ -1006,14 +1006,13 @@ func (h *Handler) processAddMembers(ctx context.Context, data []byte) (err error
 	// pair cannot go stale between here and the fan-out below.
 	var pair *roomkeystore.VersionedKeyPair
 	if len(subs) > 0 {
+		// A key-absent room here would otherwise permanently fail the "added"
+		// fan-out, leaving the new member unable to decrypt with no retry. Instead
+		// self-heal: mint a fresh key for the (never-keyed) room. Going-forward-only
+		// — see keyPairOrHeal.
 		var err error
-		pair, err = h.keyStore.Get(ctx, req.RoomID)
+		pair, err = h.keyPairOrHeal(ctx, req.RoomID)
 		if err != nil {
-			roomkeymetrics.RecordStoreError(ctx, "Get")
-			return fmt.Errorf("get room key for subscription fan-out: %w", err)
-		}
-		// A keyless "added" event would leave the new member unable to decrypt with no retry.
-		if err := requireKeyPair(ctx, pair); err != nil {
 			return err
 		}
 	}
@@ -1693,6 +1692,35 @@ func (h *Handler) existingRoomKey(ctx context.Context, roomID string, fallbackPa
 		return nil, fmt.Errorf("store room key: %w", err)
 	}
 	return &roomkeystore.VersionedKeyPair{Version: ver, KeyPair: *fallbackPair}, nil
+}
+
+// keyPairOrHeal returns the room's current key, minting and persisting a fresh
+// key when the store holds none. A key-absent room is a never-keyed channel
+// (legacy / pre-encryption); the mint unblocks going-forward messages but does
+// NOT recover history that a lost key had encrypted — it is going-forward-only.
+// RecordKeyAbsent still fires on the mint so ops can spot a lost-key event.
+// SetIfAbsent (not Set) converges concurrent minters — JetStream redelivery or a
+// racing worker — on a single v0 key instead of fanning out rival bytes.
+func (h *Handler) keyPairOrHeal(ctx context.Context, roomID string) (*roomkeystore.VersionedKeyPair, error) {
+	pair, err := h.keyStore.Get(ctx, roomID)
+	if err != nil {
+		roomkeymetrics.RecordStoreError(ctx, "Get")
+		return nil, fmt.Errorf("get room key: %w", err)
+	}
+	if pair != nil {
+		return pair, nil
+	}
+	roomkeymetrics.RecordKeyAbsent(ctx, "")
+	fallback, err := roomkeystore.GenerateKeyPair()
+	if err != nil {
+		return nil, fmt.Errorf("generate fallback room key: %w", err)
+	}
+	committed, err := h.keyStore.SetIfAbsent(ctx, roomID, *fallback)
+	if err != nil {
+		roomkeymetrics.RecordStoreError(ctx, "SetIfAbsent")
+		return nil, fmt.Errorf("store fallback room key: %w", err)
+	}
+	return committed, nil
 }
 
 func (h *Handler) processCreateRoomChannel(ctx context.Context, req *model.CreateRoomRequest, room *model.Room, requester *model.User, pair *roomkeystore.VersionedKeyPair, requestID string, acceptedAt, now time.Time) error {
--- a/room-worker/handler.go
+++ b/room-worker/handler.go
@@ -1006,10 +1006,8 @@ func (h *Handler) processAddMembers(ctx context.Context, data []byte) (err error
 	// pair cannot go stale between here and the fan-out below.
 	var pair *roomkeystore.VersionedKeyPair
 	if len(subs) > 0 {
-		// A key-absent room here would otherwise permanently fail the "added"
-		// fan-out, leaving the new member unable to decrypt with no retry. Instead
-		// self-heal: mint a fresh key for the (never-keyed) room. Going-forward-only
-		// — see keyPairOrHeal.
+		// Self-heal a key-absent room instead of permanently failing the "added"
+		// fan-out (new member couldn't decrypt, no retry). See keyPairOrHeal.
 		var err error
 		pair, err = h.keyPairOrHeal(ctx, req.RoomID)
 		if err != nil {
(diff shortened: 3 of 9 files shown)


julianshen:
**No actionable comments.** ✅

### Disputed

- `bot-room-service/handler.go:580` (`distributed-system`) — The discussion confirms this helper cannot distinguish a never-keyed room from one whose current-key record was lost, yet SetIfAbsent installs the replacement as version 0. Concrete timeline: a room uses v0, rotates, then loses its current-key record while clients retain the original v0; a subsequent add heals the room with different bytes under v0. New ciphertext and delayed historical ciphertext now share `(roomID, version=0)` but require different keys, causing decryption failures regardless of which key a client caches. Observability and eventual refetch do not remove that identity collision. The analogous room-worker helper has the same behavior. Only mint v0 when durable metadata proves the room was never keyed; otherwise allocate above a durable version high-water mark or fail closed.
- `bot-room-service/handler.go:320` (`distributed-system`) — A reachable interleaving remains: add reads version N; a concurrent removal snapshots survivors before the prospective member exists and rotates to N+1; add then commits the subscription and fans out only N. Because the rotation omitted that account, the newly added member lacks the current key and initially cannot decrypt N+1 messages. Rotation CAS prevents rival rotations but does not fence membership changes. On-demand refetch may repair the client after a decrypt miss, but the add path still delivers a stale key. Serialize additions with survivor snapshot/rotation using the same per-room fencing mechanism, or durably associate membership with key delivery and retry until the member has the current version.
- `room-worker/handler.go:1716` (`mongodb`) — When an encrypted room loses its current room-key document, `Get` returns no pair and this `SetIfAbsent` path stores fresh bytes as version 0, as confirmed by the new tests. A room that previously used version 0 can therefore have two different keys identified by `(roomID, 0)`: clients retaining the original key cannot decrypt new ciphertext, while resolving historical version-0 ciphertext with the replacement key also fails. The same helper in `bot-room-service/handler.go` has this behavior. Preserve a durable per-room version high-water mark and atomically allocate the next version, or only create version 0 when durable metadata proves the room was never keyed; otherwise fail closed.
- `bot-room-service/handler.go:319` (`performance-and-concurrency`) — The pre-resolved key is reused after the subscription writes without fencing against removal/rotation. A reachable interleaving is: add reads version V; removal snapshots survivors before the new member is committed; removal commits V+1 and fans it out without that member; add then commits the member and sends only cached V. The member lacks the current key, causing a decrypt failure and an extra on-demand key fetch. That fallback limits the duration but does not prevent the concurrency regression: with the previous post-commit lookup, this interleaving would have returned V+1. Coordinate bot membership commits with the removal snapshot/rotation using the same per-room synchronization, or atomically associate membership with a key version and retry delivery when rotation overlaps.
- `bot-room-service/handler.go:580` (`tgd-review`) — When a previously keyed room loses its current-key record, this helper cannot distinguish that state from a never-keyed room and installs fresh bytes as version 0. If a retired version-0 key or ciphertext still exists, old and new data now share the same `(roomID, version)` identity but require different keys, so one cannot be decrypted reliably. `RecordKeyAbsent` only observes the collision. The discussion acknowledges this as an intentional tradeoff, but the data-integrity risk remains. Preserve a durable version high-water mark or epoch, prove the room was never keyed before minting v0, or fail closed when history is unknown. The equivalent helper in `room-worker/handler.go` needs the same treatment.

### Local review memory

- `bot-room-service/handler.go:320` — @chenjr0719: The premise that room-worker holds a per-room lock this path lacks isn't the case — neither path takes a room lock; both rely on CommitRotation's version compare-and-swap for rotation safety. The add/rotation interleaving is pre-existing (… ([thread](https://github.com/hmchangw/newchat/pull/441#discussion_r3892767184))
- `bot-room-service/handler.go:357` — @chenjr0719: The best-effort fan-out here is the existing DM-path convention (the fan-out is fire-and-forget); a member that misses the event refetches the key on demand. Not introduced by this change. ([thread](https://github.com/hmchangw/newchat/pull/441#discussion_r3892767194))
- `room-worker/handler.go:1012` — @chenjr0719: requireKeyPair is still referenced (buildAndFanOutRoomKey), so it isn't dead — the unused-code linter passes. ([thread](https://github.com/hmchangw/newchat/pull/441#discussion_r3892767199))
- `bot-room-service/handler.go:580` — @chenjr0719: Scoped down: the remaining heal mints only for a never-keyed channel on member-add — no prior key, so no version-0 identity reuse. A genuinely lost key of a previously-rotated room is surfaced via the key-absent metric and treated as going… ([thread](https://github.com/hmchangw/newchat/pull/441#discussion_r3892767181))
- `bot-room-service/handler.go:580` — @chenjr0719: To be precise (correcting my earlier framing): keyPairOrHeal mints v0 whenever the current key is absent — it cannot distinguish never-keyed from lost-key, so this is a deliberate going-forward-only policy, not a never-keyed-only guarantee… ([thread](https://github.com/hmchangw/newchat/pull/441#discussion_r3892767181))
- `room-worker/handler.go:1716` — @chenjr0719: Same as the bot-room-service thread: this mint fires only for a never-keyed channel on add (no prior version 0 to collide with). A lost key of a rotated room is metric-surfaced and going-forward-only by design. The key.get read-path mint h… ([thread](https://github.com/hmchangw/newchat/pull/441#discussion_r3892767187))
- `room-worker/handler.go:1716` — @chenjr0719: To be precise (correcting my earlier framing): keyPairOrHeal mints v0 whenever the current key is absent — it cannot distinguish never-keyed from lost-key, so this is a deliberate going-forward-only policy, not a never-keyed-only guarantee… ([thread](https://github.com/hmchangw/newchat/pull/441#discussion_r3892767187))
- `bot-room-service/handler.go:319` — @chenjr0719: Same as the tGD thread on this line: the remaining heal mints only for a never-keyed channel on member-add — no prior version 0 to collide with. A lost key of a previously-rotated room is surfaced via the key-absent metric and treated as g… ([thread](https://github.com/hmchangw/newchat/pull/441#discussion_r3893452902))
- `bot-room-service/handler.go:319` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟠 Major_ **Do not reuse version 0 after key loss.** keyPairOrHeal can create and store a new key at version 0 when the current record is absent. If the room previously used version 0 and only the cu… ([thread](https://github.com/hmchangw/newchat/pull/441#discussion_r3893452902))
- `bot-room-service/handler.go:319` — @coderabbitai: @chenjr0719 , the read-path revert narrows the scope, but it does not resolve this member-add path. keyPairOrHeal cannot distinguish a never-keyed room from a room that lost its current-key record. Both cases produce ErrNoCurrentKey ,… ([thread](https://github.com/hmchangw/newchat/pull/441#discussion_r3893452902))

<details>

* `bot-room-service/handler_test.go`
* `room-worker/handler.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

bot-room-service/handler.go
```diff
@@ -558,6 +559,31 @@ func (h *handler) fanOutKey(ctx context.Context, roomID string, accounts []strin
...
+	}
+	if pair != nil {
+		return pair, nil
+	}
+	roomkeymetrics.RecordKeyAbsent(ctx, "")
+	fresh, err := roomkeystore.GenerateKeyPair()
+	if err != nil {
+		return nil, fmt.Errorf("generate room key: %w", err)
+	}
+	committed, err := h.keyStore.SetIfAbsent(ctx, roomID, *fresh)
```
julianshen, line 580: _🎯 distributed-correctness_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`, `tgd-review` (2 rules)_

**Healing can reuse version 0 for different key bytes.**

The comments explicitly allow this path to handle a lost key, while SetIfAbsent initializes the healed key at version 0 (also asserted by the new tests). A room can therefore have an old version-0 key cached by clients or retained in the archive, lose its current-key record, and then mint different bytes under version 0. New messages tagged with version 0 will be resolved with the old cached or archived key and fail decryption. SetIfAbsent prevents concurrent minters from diverging, but does not prevent identity reuse. Confirm through durable metadata that the room was never keyed before minting version 0; otherwise atomically allocate a version or epoch greater than every retained key, or fail safely when history cannot be established. The same issue affects the new healing paths in room-service and room-worker.

<details>

- **`tgd-review`**:580: This helper handles an acknowledged lost-key state as well as never-keyed rooms, but `SetIfAbsent` installs the fresh bytes as version 0. A previously encrypted room has already associated version 0 with different key bytes. Future messages can therefore reuse `(roomID, version=0)` for a new key: clients retaining the original v0 key may fail to decrypt new messages, while replacing that cache entry makes historical v0 messages undecryptable. The new room-service and room-worker healing paths have the same issue. Preserve a monotonic per-room version/high-water mark when healing, or only mint v0 after proving the room has never held a key; a genuinely lost-key state should fail rather than reuse a historical version.

</details>

<details>

```
In `bot-room-service/handler.go` around line 580: The comments explicitly allow this path to handle a lost key, while SetIfAbsent initializes the healed key at version 0 (also asserted by the new tests). A room can therefore have an old version-0 key cached by clients or retained in the archive, lose its current-key record, and then mint different bytes under version 0. New messages tagged with version 0 will be resolved with the old cached or archived key and fail decryption. SetIfAbsent prevents concurrent minters from diverging, but does not prevent identity reuse. Confirm through durable metadata that the room was never keyed before minting version 0; otherwise atomically allocate a version or epoch greater than every retained key, or fail safely when history cannot be established. The same issue affects the new healing paths in room-service and room-worker.

```

</details>

---

julianshen (commented):
### tGDBot review of `1a37bc7` — 10 findings in 9 files

6 🔴 blocking · 3 🟠 warning · 1 🔵 suggestion
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/441#issuecomment-5475646868), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

bot-room-service/handler.go
```diff
@@ -310,6 +311,16 @@ func (h *handler) handleAdd(c *natsrouter.Context, req BotMembersBatchRequest) (
 	created := h.now()
 	added := []string{}
 	newAccounts := []string{}
+
+	// Resolve (and heal) the room key BEFORE committing any subscription: a key
+	// failure after a member is committed would strand them keyless, since a retry
+	// sees them as a dup (newlyAdded=false) and the fan-out below skips them.
+	// Mirrors room-worker's key-before-commit ordering. See keyPairOrHeal.
+	pair, err := h.keyPairOrHeal(c, roomID)
+	if err != nil {
```
julianshen, line 320: _🏷️ ordering_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`, `performance-and-concurrency` (2 rules)_

**A concurrent rotation can make the pre-resolved key stale.**

The key is read before any subscription is committed and then reused after the membership writes. A concrete interleaving is: add reads version N; a removal snapshots survivors without the prospective member and rotates to N+1; add commits the member and fans out only version N. The rotation omitted that member, so they never receive the current key and cannot decrypt messages using N+1. Unlike room-worker's documented room lock, this path has no version validation or fencing around the read, membership mutation, and fan-out. Serialize bot additions and removals with the same per-room distributed lock and fencing token, or atomically associate the membership mutation with a key version and retry delivery if the version changes.

<details>

- **`performance-and-concurrency`**:319: A concurrent removal can rotate the room key between this lookup and the subscription commit. For example: add reads version V; removal snapshots survivors before the new account exists, rotates to V+1, and fans V+1 out; add then commits the account and sends it only V. The new member consequently lacks the current key and cannot decrypt subsequent messages. Coordinate add commits with removal/rotation using the same per-room synchronization, or redesign the commit/fan-out flow so a concurrent rotation cannot omit the newly committed member and retries repair duplicate subscriptions.

</details>

<details>

```
In `bot-room-service/handler.go` around line 320: The key is read before any subscription is committed and then reused after the membership writes. A concrete interleaving is: add reads version N; a removal snapshots survivors without the prospective member and rotates to N+1; add commits the member and fans out only version N. The rotation omitted that member, so they never receive the current key and cannot decrypt messages using N+1. Unlike room-worker's documented room lock, this path has no version validation or fencing around the read, membership mutation, and fan-out. Serialize bot additions and removals with the same per-room distributed lock and fencing token, or atomically associate the membership mutation with a key version and retry delivery if the version changes.

```

</details>

---

room-worker/handler.go
```diff
@@ -1695,6 +1692,35 @@ func (h *Handler) existingRoomKey(ctx context.Context, roomID string, fallbackPa
...
+	}
+	if pair != nil {
+		return pair, nil
+	}
+	roomkeymetrics.RecordKeyAbsent(ctx, "")
+	fallback, err := roomkeystore.GenerateKeyPair()
+	if err != nil {
+		return nil, fmt.Errorf("generate fallback room key: %w", err)
+	}
+	committed, err := h.keyStore.SetIfAbsent(ctx, roomID, *fallback)
```
julianshen, line 1716: _🏷️ data-modeling_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`mongodb`_

**Healing a lost key reuses version 0 with different key bytes**

The room-key document write treats every missing current key as a never-keyed room and commits a fresh version-0 key; the helper comments explicitly include a lost-key event, and the new tests expect version 0. If an encrypted room previously used version 0 and its MongoDB current-key state is lost while messages or clients retain the old key, this path creates different bytes under the same `(roomID, version)` identity. Cached clients then cannot decrypt new version-0 messages, while `GetByVersion(0)` cannot distinguish historical ciphertext from the replacement key. Only auto-heal when durable room metadata proves the room was never keyed; otherwise return the absent-key error, or allocate from a monotonic version counter that survives key-document loss. The same correction is needed for the bot-room-service and room-service healing paths.

<details>

```
In `room-worker/handler.go` around line 1716: The room-key document write treats every missing current key as a never-keyed room and commits a fresh version-0 key; the helper comments explicitly include a lost-key event, and the new tests expect version 0. If an encrypted room previously used version 0 and its MongoDB current-key state is lost while messages or clients retain the old key, this path creates different bytes under the same `(roomID, version)` identity. Cached clients then cannot decrypt new version-0 messages, while `GetByVersion(0)` cannot distinguish historical ciphertext from the replacement key. Only auto-heal when durable room metadata proves the room was never keyed; otherwise return the absent-key error, or allocate from a monotonic version counter that survives key-document loss. The same correction is needed for the bot-room-service and room-service healing paths.

```

</details>

---

room-service/handler.go
```diff
@@ -488,7 +488,24 @@ func (h *Handler) getRoomKey(c *natsrouter.Context) (*model.RoomKeyGetResponse,
...
-			return nil, errRoomKeyAbsent
+			// Migrated/legacy channel with no key: mint on demand (roll-out).
+			// Gate to channels (DMs aren't encrypted); SetIfAbsent so concurrent
+			// pulls converge on one key.
+			room, rErr := h.store.GetRoomAppRead(ctx, roomID)
+			if rErr != nil {
+				return nil, fmt.Errorf("get room key: room lookup: %w", rErr)
+			}
+			if room.Type != model.RoomTypeChannel {
+				return nil, errRoomKeyAbsent
```
julianshen: _🎯 correctness_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`silent-failure`_

**Channel type alone cannot distinguish a rollout from a lost key.**

If an encrypted channel's current key disappears because of corruption or an operational failure, `Get` returns no key and this branch treats it exactly like a never-keyed legacy channel. It persists a fresh key and returns success, although that key cannot decrypt existing ciphertext; clients then continue encrypting under the replacement, masking the key-loss incident and complicating recovery. The new helpers in `bot-room-service/handler.go` and `room-worker/handler.go` make the same unconditional inference. Only self-heal when durable room metadata positively identifies a never-keyed rollout candidate; for a room known to have been keyed, surface the missing-key failure without calling `SetIfAbsent`.

<details>

```
In `room-service/handler.go` around line 499: If an encrypted channel's current key disappears because of corruption or an operational failure, `Get` returns no key and this branch treats it exactly like a never-keyed legacy channel. It persists a fresh key and returns success, although that key cannot decrypt existing ciphertext; clients then continue encrypting under the replacement, masking the key-loss incident and complicating recovery. The new helpers in `bot-room-service/handler.go` and `room-worker/handler.go` make the same unconditional inference. Only self-heal when durable room metadata positively identifies a never-keyed rollout candidate; for a room known to have been keyed, surface the missing-key failure without calling `SetIfAbsent`.

```

</details>

---

bot-room-service/handler.go
```diff
@@ -343,24 +354,14 @@ func (h *handler) handleAdd(c *natsrouter.Context, req BotMembersBatchRequest) (
 		}
 	}
 
-	// Fan out the room's current key only to newly-subscribed accounts — duplicate adds already
+	// Fan out the pre-resolved key only to newly-subscribed accounts — duplicate adds already
```
julianshen, line 357: _🏷️ delivery_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`, `nats` (2 rules)_

**A failed key fan-out is acknowledged as a successful add.**

The new ordering prevents key-store failure after a subscription commit, but it does not close the broker failure gap. The accompanying TestHandleAdd_KeyFanOutSendFailureDoesNotFailOp confirms that a per-account send failure is only logged. The subscription can commit, the key publish can fail during a broker or network outage, and the caller still receives success. A retry then sees newlyAdded=false and skips fan-out, leaving the member without the key until an explicit recovery request. Persist a key-delivery event atomically with the subscription mutation, then publish it through a retrying transactional outbox with an idempotent consumer; do not acknowledge the add based only on a best-effort send.

<details>

- **`nats`**:357: In the key-heal path, the subscription is committed before `fanOutKey` publishes the key. As confirmed by `TestHandleAdd_KeyFanOutSendFailureDoesNotFailOp`, a NATS send failure is only logged, so `handleAdd` still returns success. The member is then recorded as added without receiving the key. Even if the caller retries, `UpsertSubscription` reports a duplicate, `newAccounts` remains empty, and this fan-out is skipped permanently. Persist a pending key-delivery/outbox record with the subscription and retry publication until it succeeds, or otherwise track delivery so duplicate add requests retry the key fan-out instead of reporting success.

</details>

<details>

```
In `bot-room-service/handler.go` around line 357: The new ordering prevents key-store failure after a subscription commit, but it does not close the broker failure gap. The accompanying TestHandleAdd_KeyFanOutSendFailureDoesNotFailOp confirms that a per-account send failure is only logged. The subscription can commit, the key publish can fail during a broker or network outage, and the caller still receives success. A retry then sees newlyAdded=false and skips fan-out, leaving the member without the key until an explicit recovery request. Persist a key-delivery event atomically with the subscription mutation, then publish it through a retrying transactional outbox with an idempotent consumer; do not acknowledge the add based only on a best-effort send.

```

</details>

---

room-worker/handler.go
```diff
@@ -1006,14 +1006,11 @@ func (h *Handler) processAddMembers(ctx context.Context, data []byte) (err error
 	// pair cannot go stale between here and the fan-out below.
 	var pair *roomkeystore.VersionedKeyPair
 	if len(subs) > 0 {
+		// Self-heal a key-absent room instead of permanently failing the "added"
+		// fan-out (new member couldn't decrypt, no retry). See keyPairOrHeal.
 		var err error
-		pair, err = h.keyStore.Get(ctx, req.RoomID)
+		pair, err = h.keyPairOrHeal(ctx, req.RoomID)
```
julianshen, line 1012: _🏷️ dead-code_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`dead-code`_

**Replacing the last call leaves requireKeyPair unused.**

This path now calls keyPairOrHeal instead of requireKeyPair, leaving the old helper with no callers. Its key-absence handling can therefore never execute. Delete requireKeyPair and remove any sentinel or metric scaffolding referenced exclusively by that helper.

<details>

```
In `room-worker/handler.go` around line 1012: This path now calls keyPairOrHeal instead of requireKeyPair, leaving the old helper with no callers. Its key-absence handling can therefore never execute. Delete requireKeyPair and remove any sentinel or metric scaffolding referenced exclusively by that helper.

```

</details>

---

chenjr0719:
Thanks for the thorough pass — appreciate the depth here. The central finding (healing mints version 0, which collides with a retained/archived version 0 if the room was previously keyed and only its current-key record was lost) is a real distributed-identity concern; it intersects a deliberate going-forward-only design choice for never-keyed (migrated) rooms, so I'm re-evaluating whether to distinguish never-keyed from lost-key durably (e.g. mint above the retired-key high-water-mark) rather than always v0. I'm also looking at the bot add/rotation ordering relative to the room-worker path's fencing.

Holding this out of merge until that's resolved — please don't merge pending the follow-up. Will update here.

room-service/handler.go
```diff
@@ -488,7 +488,24 @@ func (h *Handler) getRoomKey(c *natsrouter.Context) (*model.RoomKeyGetResponse,
...
+			}
+			roomkeymetrics.RecordKeyAbsent(ctx, "GetHeal")
+			pair, gErr := roomkeystore.GenerateKeyPair()
```
mliu33: don't fallabck to generate on this endpoint. This function is used only for GET

chenjr0719: Done — reverted; `key.get` is a strict read again (no generate). The member-add self-heal path is unchanged.

room-service/handler.go
chenjr0719: Resolved — the on-demand mint on this read endpoint has been reverted per maintainer feedback; `key.get` no longer generates a key.

bot-room-service/handler.go
chenjr0719: Scoped down: the remaining heal mints only for a never-keyed channel on member-add — no prior key, so no version-0 identity reuse. A genuinely lost key of a previously-rotated room is surfaced via the key-absent metric and treated as going-forward-only by design, not silently reconstructed here. (The read-path mint that raised the broader identity concern has been reverted.)

room-worker/handler.go
chenjr0719: Same as the bot-room-service thread: this mint fires only for a never-keyed channel on add (no prior version 0 to collide with). A lost key of a rotated room is metric-surfaced and going-forward-only by design. The key.get read-path mint has been reverted.

bot-room-service/handler.go
chenjr0719: The premise that room-worker holds a per-room lock this path lacks isn't the case — neither path takes a room lock; both rely on CommitRotation's version compare-and-swap for rotation safety. The add/rotation interleaving is pre-existing (not introduced by this change) and self-corrects: a member holding a stale version refetches on a decrypt miss.

bot-room-service/handler.go
chenjr0719: The best-effort fan-out here is the existing DM-path convention (the fan-out is fire-and-forget); a member that misses the event refetches the key on demand. Not introduced by this change.

room-worker/handler.go
chenjr0719: requireKeyPair is still referenced (buildAndFanOutRoomKey), so it isn't dead — the unused-code linter passes.

chenjr0719:
Follow-up to my hold note: addressed. The `key.get` on-demand mint is reverted (it's a read-only GET again, per @mliu33), and the PR is now scoped to just the member-add self-heal. I've replied on each review thread — the read-path version-identity concern is gone with the revert, and the remaining add-heal only mints for a never-keyed channel (no prior key to collide with). Ready for another look.

bot-room-service/handler.go
```diff
@@ -310,6 +311,16 @@ func (h *handler) handleAdd(c *natsrouter.Context, req BotMembersBatchRequest) (
 	created := h.now()
 	added := []string{}
 	newAccounts := []string{}
+
+	// Resolve (and heal) the room key BEFORE committing any subscription: a key
+	// failure after a member is committed would strand them keyless, since a retry
+	// sees them as a dup (newlyAdded=false) and the fan-out below skips them.
+	// Mirrors room-worker's key-before-commit ordering. See keyPairOrHeal.
+	pair, err := h.keyPairOrHeal(c, roomID)
```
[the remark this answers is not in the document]
chenjr0719: Same as the tGD thread on this line: the remaining heal mints only for a never-keyed channel on member-add — no prior version 0 to collide with. A lost key of a previously-rotated room is surfaced via the key-absent metric and treated as going-forward-only by design, not silently reconstructed. The read-path (key.get) mint that raised the broader identity concern has been reverted from this PR.

bot-room-service/handler.go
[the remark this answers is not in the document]
chenjr0719: The key is resolved and persisted *before* any subscription is committed, so a member that misses the (best-effort, fire-and-forget by the existing DM-fan-out convention) key event can still fetch the already-minted key via the read RPC — recoverable, not a permanent strand. The commit-then-fan-out structure is pre-existing; making bot key delivery fully durable (an outbox for the key event) is a worthwhile separate change but beyond this PR's add-heal scope.

bot-room-service/handler.go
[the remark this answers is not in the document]
chenjr0719: To be precise (correcting my earlier framing): keyPairOrHeal mints v0 whenever the current key is absent — it cannot distinguish never-keyed from lost-key, so this is a deliberate going-forward-only policy, not a never-keyed-only guarantee. We are keeping the simpler v0 heal, for these reasons:

1. The intended trigger is a never-keyed (migrated/legacy) channel, where v0 is the correct first version with nothing to collide with.
2. A lost current-key record is an abnormal partial-data-loss state (no normal flow deletes the room's current key — rotation replaces it). In that state the lost key's own messages are already unrecoverable regardless; the only additional effect is shadowing a retired v0 on a rotated-then-lost room, and RecordKeyAbsent surfaces that event for ops.
3. Clients converge on the current key via the read path.

We are accepting lost-key as a going-forward-only, observable event rather than gating the common migration path on durable never-keyed metadata. If you would prefer identity preservation over simplicity, the clean alternative is minting above the retired-key high-water-mark (never-keyed -> v0; rotated-then-lost -> max(retired)+1, preserving the retired history); we can adopt that if you'd like — flagging it as a conscious tradeoff, not an oversight.

bot-room-service/handler.go
chenjr0719: To be precise (correcting my earlier framing): keyPairOrHeal mints v0 whenever the current key is absent — it cannot distinguish never-keyed from lost-key, so this is a deliberate going-forward-only policy, not a never-keyed-only guarantee. We are keeping the simpler v0 heal, for these reasons:

1. The intended trigger is a never-keyed (migrated/legacy) channel, where v0 is the correct first version with nothing to collide with.
2. A lost current-key record is an abnormal partial-data-loss state (no normal flow deletes the room's current key — rotation replaces it). In that state the lost key's own messages are already unrecoverable regardless; the only additional effect is shadowing a retired v0 on a rotated-then-lost room, and RecordKeyAbsent surfaces that event for ops.
3. Clients converge on the current key via the read path.

We are accepting lost-key as a going-forward-only, observable event rather than gating the common migration path on durable never-keyed metadata. If you would prefer identity preservation over simplicity, the clean alternative is minting above the retired-key high-water-mark (never-keyed -> v0; rotated-then-lost -> max(retired)+1, preserving the retired history); we can adopt that if you'd like — flagging it as a conscious tradeoff, not an oversight.

room-worker/handler.go
chenjr0719: To be precise (correcting my earlier framing): keyPairOrHeal mints v0 whenever the current key is absent — it cannot distinguish never-keyed from lost-key, so this is a deliberate going-forward-only policy, not a never-keyed-only guarantee. We are keeping the simpler v0 heal, for these reasons:

1. The intended trigger is a never-keyed (migrated/legacy) channel, where v0 is the correct first version with nothing to collide with.
2. A lost current-key record is an abnormal partial-data-loss state (no normal flow deletes the room's current key — rotation replaces it). In that state the lost key's own messages are already unrecoverable regardless; the only additional effect is shadowing a retired v0 on a rotated-then-lost room, and RecordKeyAbsent surfaces that event for ops.
3. Clients converge on the current key via the read path.

We are accepting lost-key as a going-forward-only, observable event rather than gating the common migration path on durable never-keyed metadata. If you would prefer identity preservation over simplicity, the clean alternative is minting above the retired-key high-water-mark (never-keyed -> v0; rotated-then-lost -> max(retired)+1, preserving the retired history); we can adopt that if you'd like — flagging it as a conscious tradeoff, not an oversight.

mliu33 (approved):

PR #442, opened 2026-08-31
fix(bot-message-handler): resolve bot sender display name from the app name

chenjr0719:
## Summary

A bot message's stored `UserDisplayName` came from `CombineWithFallback(EngName, ChineseName, Account)`, but a bot identity — parsed from the `X-Bot-Identity` header — carries **none** of the name fields (only `id`/`account`/`siteId`), so it fell back to the raw bot account (e.g. `myapp.bot`). Every other surface (room-list preview, history, member.list) resolves the friendly app name via an `AppNameByAccount` lookup, so a bot's message row showed the raw account while those surfaces showed the app name — a visible mismatch.

This resolves the app name at write time through the shared `preview.BotAwareDisplayName` helper (the same one preview/history use), falling back to the composed name when no app matches.

## What's included

- `bot-message-handler` gains a cached `AppNameByAccount` lookup (`preview.CachedAppNameLookup`), wired once at startup.
- Both bot send paths (`handleSendDM`, `handleSendRoom`) compose the sender display name via `preview.BotAwareDisplayName`.
- Regression test: a bare bot identity (no `EngName`, as the live header carries) resolves the display name from the app name, not the raw account.

## Changes

- `bot-message-handler/store.go`, `store_mongo.go` — add `AppNameByAccount(ctx, botAccount)` (queries `apps` by `assistant.name`, precise projection).
- `bot-message-handler/handler.go` — `handler` holds a `preview.AppNameLookup`; the two `UserDisplayName` compositions use `preview.BotAwareDisplayName`.
- `bot-message-handler/main.go` — build the cached lookup and inject it.

No wire-struct change: `UserDisplayName` already exists on the message and its meaning is unchanged — only its value source is corrected — so no `docs/client-api.md` change is needed.

## Verification

`go build`, unit tests (incl. the new resolution test), `go vet -tags integration`, and golangci-lint are all green for `bot-message-handler`.

### DIFF
--- a/bot-message-handler/handler.go
+++ b/bot-message-handler/handler.go
@@ -16,6 +16,7 @@ import (
 	"github.com/hmchangw/chat/pkg/idgen"
 	"github.com/hmchangw/chat/pkg/model"
 	"github.com/hmchangw/chat/pkg/natsrouter"
+	"github.com/hmchangw/chat/pkg/preview"
 	"github.com/hmchangw/chat/pkg/subject"
 )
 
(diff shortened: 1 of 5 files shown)


chenjr0719:
Closing this — the bot sender display-name resolution will be handled elsewhere in the pipeline, so this change isn't needed here.

PR #447, opened 2026-08-31
fix(search-sync-worker): bind bot collection to its own stream's subject

hmchangw:
## Problem

The bot message collection binds `BOT-MESSAGES-CANONICAL-{siteID}` (`messages.go:104-107`), whose only subject is `chat.bot.canonical.{siteID}.>` (`pkg/stream/stream.go:97`). But `FilterSubjects` returned the **user** stream's `chat.msg.canonical.{siteID}.*` for every non-teams collection:

```go
func (c *messageCollection) FilterSubjects(siteID string) []string {
	if c.teamsOnly {
		return []string{subject.MsgTeamsCanonicalBatch(siteID)}
	}
	return []string{subject.MsgCanonicalMessageWildcard(siteID)} // ← also hit by the bot collection
}
```

A consumer filter that is not a subset of its stream's interest is rejected by `js.CreateOrUpdateConsumer`, and `main.go:348-355` treats that as fatal:

- Where the bot stream exists, the pod **exits at startup in `default` mode** — crashloop.
- Where it does not, bot messages are **silently never indexed**.

`messages_test.go:622` asserted the wrong string, so the test suite agreed with the bug.

## Fix

Filter subjects are now parameterized per collection alongside `streamCfg`, mirroring how this file already parameterizes the stream config and consumer name. The stream and its filter are set in the same place, so they cannot drift apart:

| Collection | Stream | Filter |
|---|---|---|
| user | `MESSAGES-CANONICAL-{site}` | `chat.msg.canonical.{site}.*` |
| bot | `BOT-MESSAGES-CANONICAL-{site}` | `chat.bot.canonical.{site}.created` |
| teams | `MESSAGES-TEAMS-{site}` | `chat.teams.msg.canonical.{site}.batch` |

`chat.bot.canonical.{siteID}.created` is the only subject published onto that stream (`bot-message-handler/handler.go:199`, `bot-room-service/sysmsg.go:65`) and the same leaf the sibling consumer binds (`bot-message-worker/main.go:208`).

`FilterSubjects` now just delegates, which also drops the `teamsOnly` branch from it. `teamsOnly` still gates payload decoding and the template/mapping pushes, unchanged.

## Tests

Written first, both confirmed failing against the old code:

1. `TestMessageCollection_FilterSubjects` — corrected the bot assertion. Failed with `expected: ["chat.bot.canonical.site-a.created"] / actual: ["chat.msg.canonical.site-a.*"]`.
2. `TestMessageCollection_FilterSubjectsAreOnTheirOwnStream` — new table-driven guard asserting each collection's filter subjects are matched by its own stream's subjects, using NATS token rules (`*` = one token, `>` = one or more trailing). Failed only on the `bot` subtest (`filter "chat.msg.canonical.site-a.*" is not carried by stream "BOT-MESSAGES-CANONICAL-site-a"`), with `user` and `teams` passing — which isolates the defect and confirms the matcher itself is sound.

This second test is the real regression guard: any future stream/filter mismatch in this service now fails in CI rather than at rollout.

## Verification

| Gate | Result |
|---|---|
| `make test SERVICE=search-sync-worker` | ok, 2.097s (`-race`) |
| `make test` (full repo) | 124 packages ok, 0 failures |
| `make lint` | 0 issues |
| `make sast-gosec` | exit 0, no findings |

`govulncheck` and `semgrep` could not run in this sandbox (`vuln.go.dev` blocked by the agent proxy; `semgrep` not installed) — they need to pass in CI.

## Scope

Production change is confined to `search-sync-worker/messages.go`. No client-facing handler or `pkg/model` struct changed, so `docs/client-api.md` needs no update. No new dependencies.

Found by the production-readiness audit of this service, where it was the sole `critical` finding.

---
_Generated by [Claude Code](https://claude.ai/code/session_018kfXTCHTVMSrBxwJvHHoNK)_

### DIFF
--- a/search-sync-worker/messages_test.go
+++ b/search-sync-worker/messages_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
+	"strings"
 	"testing"
 	"time"
 
@@ -619,7 +620,8 @@ func TestMessageCollection_FilterSubjects(t *testing.T) {
 	assert.Equal(t, []string{"chat.msg.canonical.site-a.*"}, user.FilterSubjects("site-a"))
 
 	bot := newBotMessageCollection("msgs-v1", false)
-	assert.Equal(t, []string{"chat.msg.canonical.site-a.*"}, bot.FilterSubjects("site-a"))
+	assert.Equal(t, []string{"chat.bot.canonical.site-a.created"}, bot.FilterSubjects("site-a"),
+		"the bot collection binds BOT-MESSAGES-CANONICAL, which carries only chat.bot.canonical.{siteID}.>")
 
 	teams := newTeamsMessageCollection("msgs-v1", "site-a", false)
 	assert.Equal(t, []string{"chat.teams.msg.canonical.site-a.batch"}, teams.FilterSubjects("site-a"))
@@ -628,6 +630,59 @@ func TestMessageCollection_FilterSubjects(t *testing.T) {
 	pattern, body := teams.MappingUpdate()
 	assert.Empty(t, pattern)
 	assert.Nil(t, body)
+}
+
+// subjectMatches reports whether a NATS subject matches a filter pattern, per the
+// server's token rules: "*" matches exactly one token, ">" matches one or more
+// trailing tokens.
+func subjectMatches(pattern, subj string) bool {
+	pt := strings.Split(pattern, ".")
+	st := strings.Split(subj, ".")
+	for i, tok := range pt {
+		if tok == ">" {
+			return i < len(st)
+		}
+		if i >= len(st) {
+			return false
+		}
+		if tok != "*" && tok != st[i] {
+			return false
+		}
+	}
+	return len(pt) == len(st)
+}
+
+// A consumer's filter subject must be a subset of its own stream's interest, or
+// CreateOrUpdateConsumer fails at startup and the service exits. This guards the whole
+// class: any collection whose stream and filter drift apart fails here, not at rollout.
+func TestMessageCollection_FilterSubjectsAreOnTheirOwnStream(t *testing.T) {
+	const siteID = "site-a"
+	tests := []struct {
+		name string
+		coll *messageCollection
+	}{
+		{"user", newMessageCollection("msgs-v1", siteID, time.Time{}, false)},
+		{"bot", newBotMessageCollection("msgs-v1", false)},
+		{"teams", newTeamsMessageCollection("msgs-v1", siteID, false)},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			streamSubjects := tt.coll.StreamConfig(siteID).Subjects
+			require.NotEmpty(t, streamSubjects)
+			for _, filter := range tt.coll.FilterSubjects(siteID) {
+				matched := false
+				for _, streamSubj := range streamSubjects {
+					if subjectMatches(streamSubj, filter) {
+						matched = true
+						break
+					}
+				}
+				assert.True(t, matched,
+					"filter %q is not carried by stream %q (subjects %v)",
+					filter, tt.coll.StreamConfig(siteID).Name, streamSubjects)
+			}
+		})
+	}
 }
 
 // fakeTeamsUserResolver returns a fixed id→identity map (nil error).
--- a/search-sync-worker/messages.go
+++ b/search-sync-worker/messages.go
@@ -36,20 +36,23 @@ type teamsUserResolver interface {
 	ResolveIdentities(ctx context.Context, teamsUserIDs []string) (map[string]teamsIdentity, error)
 }
 
-// messageCollection implements Collection for message search sync; streamCfg + consumerName are
-// parameterized so one type consumes user, bot, or Teams-migration canonical streams. syncFrom is
-// the legacy-replay cutoff (zero disables it).
+// messageCollection implements Collection for message search sync; streamCfg, filterSubjs +
+// consumerName are parameterized so one type consumes user, bot, or Teams-migration canonical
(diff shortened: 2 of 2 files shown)


julianshen:

<details>

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

julianshen (commented):
### tGDBot review of `1e511fe` — 1 finding in 2 files

1 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/447#issuecomment-5487293243), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

search-sync-worker/messages_test.go
```diff
@@ -630,6 +632,59 @@ func TestMessageCollection_FilterSubjects(t *testing.T) {
...
+		}
+		if i >= len(st) {
+			return false
+		}
+		if tok != "*" && tok != st[i] {
+			return false
+		}
+	}
+	return len(pt) == len(st)
+}
```
julianshen, line 653: _🧪 tests_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**Wildcard filters can incorrectly pass the stream-subset test**

The helper matches the filter string as though it were a concrete subject rather than checking whether the filter's entire subject set is contained by the stream. For example, a stream pattern `chat.msg.*` and filter `chat.msg.>` return true because `*` accepts the literal `>` token and the token counts match, even though the filter also accepts multi-token subjects absent from the stream. This can let the test pass for a configuration that JetStream rejects at startup. Compare the two wildcard languages directly and cover this case with a regression test.

<details>

> ‼️ **IMPORTANT**

```suggestion
// subjectMatches reports whether every subject accepted by filterPattern is also
// accepted by streamPattern, using NATS wildcard rules.
func subjectMatches(streamPattern, filterPattern string) bool {
	streamTokens := strings.Split(streamPattern, ".")
	filterTokens := strings.Split(filterPattern, ".")

	for i := 0; ; i++ {
		if i >= len(filterTokens) {
			return i >= len(streamTokens)
		}
		if i >= len(streamTokens) {
			return false
		}

		streamToken := streamTokens[i]
		filterToken := filterTokens[i]
		if streamToken == ">" {
			return true
		}
		if filterToken == ">" {
			return false
		}
		if filterToken == "*" && streamToken != "*" {
			return false
		}
		if filterToken != "*" && streamToken != "*" && filterToken != streamToken {
			return false
		}
	}
}
```

</details>

<details>

```
In `search-sync-worker/messages_test.go` around line 635: The helper matches the filter string as though it were a concrete subject rather than checking whether the filter's entire subject set is contained by the stream. For example, a stream pattern `chat.msg.*` and filter `chat.msg.>` return true because `*` accepts the literal `>` token and the token counts match, even though the filter also accepts multi-token subjects absent from the stream. This can let the test pass for a configuration that JetStream rejects at startup. Compare the two wildcard languages directly and cover this case with a regression test.

```

</details>

---

hmchangw:
Closing in favour of #462, which fixes the same bug more completely.

## Correction to this PR's description

The "Problem" section above is **wrong about the failure mode**. It claims:

> A consumer filter that is not a subset of its stream's interest is rejected by `js.CreateOrUpdateConsumer` … the pod **exits at startup in `default` mode** — crashloop.

That does not happen on the pinned server version. In `nats-server v2.12.6`:

- `JSConsumerFilterNotSubsetErr` (10093, *"consumer filter subject is not a valid subset of the interest subjects"*) is **defined but never raised** — no references outside `server/jetstream_errors_generated.go` and `server/errors.json`.
- `checkConsumerCfg` (`server/consumer.go:678`) validates filter syntax, empty filters, `FilterSubject` and `FilterSubjects` both being set, and filters overlapping *each other* — but performs **no check that a filter is a subset of the stream's subjects**.

So the consumer is created successfully, matches nothing, and reports healthy with zero pending and zero redeliveries. #462's description has this right: it is a silent indexing gap, not a crashloop. Nothing was alerting and nothing was restarting, which is precisely why a passing test could keep it hidden.

Verified by reading the vendored server source; no Docker in the environment where this was checked, so it was not confirmed against a live NATS.

## Why #462 supersedes this

Same root cause, same core approach (couple the filter to the stream so they cannot be chosen independently). #462 additionally:

1. Adds a `subject.BotCanonicalMessageWildcard` builder, so the filter comes from a builder rather than a narrower literal.
2. Adds a startup invariant (`checkFilterSubjects`) that refuses to start a consumer whose filter shares no subject with its own stream — turning the next recurrence from a silent correctness loss into a container that will not start.
3. Handles `bot-room-service`'s bare-`model.Message` sysmsgs, which the corrected filter newly routes to this consumer. They decode to an empty event, fail the missing-id check, and are Ack-dropped as poison with an error log per membership change. **This PR would have shipped that noise.** An independent review lens flagged the same hazard here at `medium`, arriving at it separately — `bot-room-service/sysmsg.go:16` documents the differing wire shape, and `BuildAction` (`search-sync-worker/messages.go:197`) unmarshals into `model.MessageEvent`.
4. Documents that deploying does not backfill — the existing `bot-message-sync` durable keeps its cursor, so the corrected filter indexes from now forward, not from the start of retention.

This PR covers none of items 1–4.

One point from the review here worth carrying over to #462 if it is not already addressed: the consumer-creation failure log at `search-sync-worker/main.go:350-354` records `stream`, `consumer` and `error`, but not the filter subjects it asked for or the subjects the stream carries. Both are in scope at the call site. For any future filter/stream mismatch that *does* produce an error, those two keys turn a source-reading investigation into a single log line.

---

PR #449, opened 2026-09-01
feat(search-service): cache HR and app enrichment lookups behind MongoStore

evelynchang5525:
## What

`search.messages` enrichment (`search-service/enrich.go`) hit MongoDB on every request for the same HR users and bot apps. This puts pod-local LRU+TTL caches in front of the two account-keyed lookups, behind a decorator implementing the existing `MongoStore` interface — so `enrich.go`, `handler.go` and the interface itself are untouched.

| `MongoStore` method | Policy | Why |
|---|---|---|
| `UsersByAccounts` | per-account `HRUser`, LRU+TTL | account-keyed, highly repeatable; HR names change rarely |
| `AppsByAssistantNames` | per-bot-account `AppRef`, LRU+TTL | same shape; app names change only on rename |
| `SubscriptionsByRoomIDs` | pass through | per-caller **and** per-room, and `isSubscribed` is volatile |
| `SearchAppsByName` | pass through | a paged text query, not a keyed lookup |

Pass-through is achieved by embedding `MongoStore`, so any method added to the interface later is uncached by default — the safe direction.

## How it behaves

- **Batch-aware.** Each call splits requested keys into cached and uncached, queries Mongo for the remainder only, then merges. All keys cached ⇒ zero round-trips.
- **Negative caching.** A key with no row is cached as a tombstone at the same TTL, so one departed user or misconfigured bot can't force a query on every search it appears in. Same policy as `pkg/preview/appname.go`.
- **Errors are never cached** and propagate as before; `enrich.go` logs and degrades.
- **Disable switch.** A non-positive size or TTL disables that tier; with both off the constructor returns the inner store unwrapped.
- **No singleflight** — deliberately out of scope. These loads are already batched per request against indexed `$in` lookups, so concurrent misses cost one extra cheap query each, not a stampede on one hot key.

## Config

New knobs on `SearchConfig` (`envPrefix:"SEARCH_"`), also added to the service's local `docker-compose.yml` with matching defaults:

| Env var | Default |
|---|---|
| `SEARCH_HR_CACHE_SIZE` | `130000` |
| `SEARCH_HR_CACHE_TTL` | `24h` |
| `SEARCH_APP_CACHE_SIZE` | `1000` |
| `SEARCH_APP_CACHE_TTL` | `24h` |

The HR tier is sized to hold the full employee directory rather than a hot subset, and both TTLs sit at a day because HR and app names change on the order of days.

Declared in the service rather than a `pkg/` `TTLConfig`: these are pod-local caches, not a shared Valkey key, so no two services can disagree about an entry only one process can see. None of the names collide with `ESConfig`, which shares the `SEARCH_` prefix.

Metrics: `cachemetrics.For("search_hr", "l1")` / `("search_app", "l1")`, recorded **per key**, so `cache_hits_total` reads as "keys served without touching Mongo".

## Trade-offs — worth a reviewer's attention

**Memory.** At full occupancy the HR tier holds ~130k entries at roughly 300–450 B each, so **~40–60 MB resident per pod**, plus ~0.4 MB for the app tier. Worth checking against the pod memory limit before this ships.

**Staleness is bounded by the TTL, now a day.** Cached values feed rendered labels only, never a decision — `AppInfo.IsSubscribed` is stamped from the **uncached** `SubscriptionsByRoomIDs`, so subscription state is never stale. Two consequences of the 24h TTL:

- A renamed HR user or app shows the old name for up to 24h, and pods can disagree with each other for that window.
- A *newly created* user or app stays negatively cached — rendering as a bare account name — for up to 24h after anything searched for it. If new-account visibility matters, `SEARCH_HR_CACHE_TTL` is the knob to shorten; the size increase is independent and carries no staleness cost.

## Testing

Written test-first throughout. 22 new test cases covering cold miss, warm hit, partial-batch narrowing, tombstones, dedup, TTL expiry, LRU eviction, error-not-cached, disabled-tier pass-through, nil-inner, per-key metric counts, and concurrent access under `-race`. The fakes record the exact key slice of every call, so the tests prove the cache narrowed the query rather than merely returning the right answer.

- `make lint` — 0 issues
- `make test` (repo-wide, `-race`) — pass
- `store_cache.go` — 100% statement coverage on every function
- **CI green on the current head**, including `sast` and `test-integration (search-service)`.

`govulncheck` and the semgrep registry ruleset could not run in the authoring environment (egress-policy 403 on `vuln.go.dev` / `semgrep.dev`), so the CI `sast` job is their first real execution against this branch — it passed. `go.mod` is untouched (`hashicorp/golang-lru/v2` was already a direct require).

## Docs

No `docs/client-api.md` change: no handler registration moved, and no request/response or event struct in `pkg/model` gained, lost, renamed or retyped a field. The caches are internal to how the service sources data it already returns.

## Summary by CodeRabbit

* **New Features**
  * Added configurable caching for HR and application search data.
  * Improved repeated searches with local caching, batch lookups, and support for caching unavailable results.
  * Added options to disable either cache independently.

* **Configuration**
  * Updated default cache sizes and extended cache duration to 24 hours.
  * Added environment-based controls for cache size and expiration.
  * Added validation for cache expiration settings, including support for disabled caches.

* **Bug Fixes**
  * Prevented unavailable results from overwriting newly cached data during concurrent searches.

* **Tests**
  * Added coverage for cache hits, misses, expiration, eviction, errors, concurrency, and disabled-cache behavior.


evelynchang5525:
Answering the two open items from the automated review, plus a status update on the SAST caveat in the description.

**SAST is now closed.** The `sast` job passed in CI. `govulncheck` and the semgrep registry ruleset could not run in the authoring environment (egress-policy 403 on `vuln.go.dev` / `semgrep.dev`), so this CI run is their first real execution against the branch. `lint`, `test-integration (search-service)` and `prewarm` are also green; `test` is still running.

**Docstring coverage (30% vs 80%) — not acting on this.** The metric analysed 30 functions and found 9 documented. Those 9 are *every* declaration in `search-service/store_cache.go` — the file is at 100%. The remaining 21 are test functions in `store_cache_test.go` and `config_test.go`. Go convention doesn't put doc comments on `TestXxx` functions, and this repo actively trims comment noise rather than adding it. Adding 21 docstrings that restate test names would make the diff worse, so I'm leaving the check red and flagging the reason rather than satisfying it mechanically.

**Assistant-name scoping — the premise is right, but the cache doesn't change the contract.** I checked: `assistant_name_idx` (`user-service/mongorepo/apps.go:67-70`) is a plain, *non-unique* index on `assistant.name`, so duplicate names are indeed possible.

What matters is that the cache key is exactly the key the uncached query already collapses on. `AppsByAssistantNames` builds `out[r.Assistant.Name] = AppRef{…}` (`search-service/store_mongo.go:130-133`), so with duplicate assistant names the *uncached* path already returns one arbitrary winner per query, and — with no sort on the find — potentially a different winner each time. The cache stores whichever one that query returned, which makes the result more stable within a TTL, not less. No new ambiguity is introduced.

The genuinely new behaviour is the one already documented in the description: if an assistant name is reassigned from one app to another, search results can show the old app's identity for up to `SEARCH_APP_CACHE_TTL` (default 5m), per pod. That's a rendered label — `AppInfo.IsSubscribed` is stamped from the **uncached** `SubscriptionsByRoomIDs`, so subscription state is never stale. Operators can set `SEARCH_APP_CACHE_TTL=0` to disable the tier outright.

If the intended contract is that `assistant.name` *is* unique, the right fix is a unique index in `user-service` (which owns it), not a change here — happy to open a separate issue for that if you want it.

---

julianshen:

### Disputed

- `search-service/store_cache.go:84` (`performance-and-concurrency`) — If multiple requests miss the same hot keys before the first load completes, every caller reaches `load(ctx, missing)` independently; the concurrent test explicitly allows 32 callers to produce 32 loader calls. Cold starts and synchronized TTL expiry can therefore create a burst of overlapping `$in` queries per pod, multiplied across replicas, consuming Mongo connections and increasing search latency or timeouts. The prior response notes this does not exceed the uncached steady-state load, but it still leaves the new cache unable to coalesce duplicate fills at the exact miss boundary. Track in-flight work per key, let waiters recheck the cache and narrow their remaining batch, and ensure waiter cancellation does not cancel a shared load needed by other callers.

### Local review memory

- `search-service/store_cache.go:89` — @evelynchang5525: Not taking this one, but it's a fair thing to raise and I want to give the reasoning rather than just cite the YAGNI note in the description. **This change cannot increase Mongo load relative to today.** Without the cache, *every* search.… ([thread](https://github.com/hmchangw/newchat/pull/449#discussion_r3901991947))

<details>

* `docs/superpowers/plans/2026-09-01-search-service-hr-app-lru-caches.md`
* `search-service/deploy/docker-compose.yml`
* `search-service/main.go`
* `search-service/store_cache.go`
* `search-service/store_cache_test.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

search-service/store_cache.go
```diff
@@ -0,0 +1,167 @@
...
+		missing = append(missing, k)
+	}
+	if len(missing) == 0 {
+		return out, nil
+	}
+
+	loaded, err := load(ctx, missing)
+	if err != nil {
+		return nil, err
+	}
```
julianshen, line 100: _🏷️ observability_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`_

**Backend load failures never increment the cache error metric.**

When Mongo fails for a cold or expired batch, each key is recorded as a miss, `load` returns an error, and the function exits without calling the recorder's `Error` method. Repeated failures therefore appear as ordinary cache misses while the error counter remains zero, masking whether enrichment keys were successfully loaded or failed at the network/database boundary. Record an error for every unresolved key, matching the documented per-key metric semantics, and assert `errCount` in the existing error test.

<details>

> ‼️ **IMPORTANT**

```suggestion
	loaded, err := load(ctx, missing)
	if err != nil {
		for range missing {
			rec.Error(ctx)
		}
		return nil, err
	}
```

</details>

<details>

```
In `search-service/store_cache.go` around line 79: When Mongo fails for a cold or expired batch, each key is recorded as a miss, `load` returns an error, and the function exits without calling the recorder's `Error` method. Repeated failures therefore appear as ordinary cache misses while the error counter remains zero, masking whether enrichment keys were successfully loaded or failed at the network/database boundary. Record an error for every unresolved key, matching the documented per-key metric semantics, and assert `errCount` in the existing error test.

```

</details>

---

julianshen, line 100: _🎯 concurrency_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`performance-and-concurrency`_

**Concurrent cache misses fan out into duplicate Mongo queries.**

After multiple request goroutines observe the same absent key, each reaches `load` before any result is cached. The concurrent test explicitly permits 32 callers to produce 32 loader calls. This occurs on cold starts and again when hot entries expire, causing every in-flight search to issue overlapping Mongo `$in` queries; replicas starting or expiring together multiply the spike and can increase Mongo load and search latency. Coalesce in-flight fills while preserving batching, then recheck the cache and query only keys still missing after waiting; cancellation must stop only the waiter, not the shared load.

<details>

```
In `search-service/store_cache.go` around line 82: After multiple request goroutines observe the same absent key, each reaches `load` before any result is cached. The concurrent test explicitly permits 32 callers to produce 32 loader calls. This occurs on cold starts and again when hot entries expire, causing every in-flight search to issue overlapping Mongo `$in` queries; replicas starting or expiring together multiply the spike and can increase Mongo load and search latency. Coalesce in-flight fills while preserving batching, then recheck the cache and query only keys still missing after waiting; cancellation must stop only the waiter, not the shared load.

```

</details>

---

docs/superpowers/plans/2026-09-01-search-service-hr-app-lru-caches.md
```diff
@@ -0,0 +1,1001 @@
...
+	t.Run("defaults", func(t *testing.T) {
+		setRequiredSearchEnv(t)
+		for _, k := range []string{"SEARCH_HR_CACHE_SIZE", "SEARCH_HR_CACHE_TTL", "SEARCH_APP_CACHE_SIZE", "SEARCH_APP_CACHE_TTL"} {
+			require.NoError(t, os.Unsetenv(k))
+		}
+
+		cfg, err := env.ParseAs[Config]()
+
+		require.NoError(t, err)
+		assert.Equal(t, 8192, cfg.Search.HRCacheSize)
```
julianshen: _📝 documentation_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**Task snippets prescribe obsolete cache defaults**

The Design section, SearchConfig, config tests, and Docker Compose use 130000/24h/1000/24h, but this Task 1 test and later implementation and Compose snippets prescribe 8192/5m/1024/5m. Following this implementation plan would undo the shipped defaults. Update every task snippet to match the actual defaults.

<details>

```
In `docs/superpowers/plans/2026-09-01-search-service-hr-app-lru-caches.md` around line 99: The Design section, SearchConfig, config tests, and Docker Compose use 130000/24h/1000/24h, but this Task 1 test and later implementation and Compose snippets prescribe 8192/5m/1024/5m. Following this implementation plan would undo the shipped defaults. Update every task snippet to match the actual defaults.

```

</details>

---

search-service/store_cache_test.go
```diff
@@ -0,0 +1,487 @@
...
+
+// countingRecorder counts cache outcomes; it satisfies cacheRecorder.
+type countingRecorder struct {
+	mu                     sync.Mutex
+	hits, misses, errCount int
+}
+
+func (c *countingRecorder) Hit(context.Context)   { c.mu.Lock(); c.hits++; c.mu.Unlock() }
+func (c *countingRecorder) Miss(context.Context)  { c.mu.Lock(); c.misses++; c.mu.Unlock() }
+func (c *countingRecorder) Error(context.Context) { c.mu.Lock(); c.errCount++; c.mu.Unlock() }
```
julianshen, line 56: _🏷️ dead-code_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`dead-code`_

**The test recorder's error counter is written but never read.**

`Error` increments `errCount`, but no test reads that field and `lookupCached` never invokes `rec.Error`; the method only needs to satisfy `cacheRecorder`. Remove the dead counter while retaining a no-op interface method.

<details>

> ‼️ **IMPORTANT**

```suggestion
	hits, misses int
}

func (c *countingRecorder) Hit(context.Context)  { c.mu.Lock(); c.hits++; c.mu.Unlock() }
func (c *countingRecorder) Miss(context.Context) { c.mu.Lock(); c.misses++; c.mu.Unlock() }
func (*countingRecorder) Error(context.Context)  {}
```

</details>

<details>

```
In `search-service/store_cache_test.go` around line 51: `Error` increments `errCount`, but no test reads that field and `lookupCached` never invokes `rec.Error`; the method only needs to satisfy `cacheRecorder`. Remove the dead counter while retaining a no-op interface method.

```

</details>

---

julianshen (commented):
### tGDBot review of `49885e0` — 4 findings in 6 files

3 🟠 warning · 1 🔵 suggestion
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/449#issuecomment-5490809319), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

search-service/store_cache.go
evelynchang5525: Confirmed and fixed in `66edddf` — you're right, and my earlier reasoning here was wrong.

I'd previously concluded (and an earlier reviewer agreed) that a pod-local LRU has no read-failure mode, so `Error` was unreachable by design. That's true of the *cache layer* in isolation, but it's not what this repo means by the counter. `pkg/userstore/cache.go:93` and `pkg/roommetacache/roommetacache.go:133` are both L1 LRUs over a backing store, and both record `metrics.Error(ctx)` when the backing load fails — and record `Miss` *only* on a successful load. So the established convention is:

- `Hit` — served from cache
- `Miss` — fell through **and** the backing store answered
- `Error` — fell through **and** the backing load failed

My version recorded `Miss` at lookup time, before the load, so a Mongo outage made this tier report a 0% hit rate with zero errors while every other tier reported errors under the same outage. That breaks the single-panel-across-all-caches property `pkg/cachemetrics`' package doc describes.

I applied the fix slightly differently from the suggested patch: recording `Error` *in addition to* the already-recorded `Miss` would double-count each key. Instead, `Miss` moved out of the scan loop to after a successful load, and the failure path records `Error` per unresolved key:

```go
loaded, err := load(ctx, missing)
if err != nil {
    for range missing {
        rec.Error(ctx)
    }
    return nil, err
}
for range missing {
    rec.Miss(ctx)
}
```

`TestLookupCached_ErrorPropagatesAndIsNotCached` now asserts `errCount == 1, misses == 0` on the failing call, then `misses == 1, errCount == 1` after the retry succeeds — pinning both that a failure is not a miss and that a recovery doesn't add a second error. `lookupCached` stays at 100% coverage.

---

docs/superpowers/plans/2026-09-01-search-service-hr-app-lru-caches.md
evelynchang5525: Fixed in `66edddf`. This was my own inconsistency: when I retuned the defaults I updated the plan's Design table but missed the Task 1 test snippet, the Task 1 `SearchConfig` snippet, and the Task 4 Compose snippet, so the document contradicted itself.

All snippets now read `130000` / `24h` / `1000` / `24h`, matching the shipped code. The only remaining `5m` in the file is `RESTRICTED_ROOMS_CACHE_TTL`, which is a pre-existing unrelated knob and correct as-is.

---

search-service/store_cache_test.go
evelynchang5525: Not applying this one — it's been overtaken by the fix for the observability finding on `store_cache.go`.

The premise was correct at the time: `lookupCached` never called `rec.Error`, so `errCount` was genuinely dead. But that was itself the defect the other thread identified — a failed backing load should record `Error`, matching `pkg/userstore` and `pkg/roommetacache`. As of `66edddf` `lookupCached` does call `rec.Error`, so the counter is live and `TestLookupCached_ErrorPropagatesAndIsNotCached` now asserts on it (`errCount == 1, misses == 0` on failure; `misses == 1, errCount == 1` after the retry succeeds).

Replacing it with a no-op `Error` method would now delete the only assertion proving outages are counted as errors rather than misses, so the counter stays.

---

search-service/store_cache.go
evelynchang5525: Not taking this one, but it's a fair thing to raise and I want to give the reasoning rather than just cite the YAGNI note in the description.

**This change cannot increase Mongo load relative to today.** Without the cache, *every* `search.messages` request issues these two `$in` queries. With the cache and no coalescing, a cold pod issues at most what it already issues in production, and strictly fewer as entries fill. The "spike" on cold start is the current steady-state behaviour, not a new one the cache introduces. A duplicate fill is wasted work, but it is bounded by concurrent in-flight requests, not by the number of keys, and each one is a batched `$in` on an indexed field (`account`, and `assistant_name_idx` on `assistant.name`).

The retune actually cuts the expiry-driven case you describe: at a 24h TTL rather than 5m, entries expire ~288× less often, so synchronised expiry herds are correspondingly rarer.

You're right that there's repo precedent for coalescing — `pkg/userstore` and `pkg/roommetacache` both use singleflight. The difference is shape: those are single-key caches where one hot key (a popular room, a busy user) is a genuine stampede target. This one is batch-keyed, so a singleflight key would have to be the batch, and two overlapping-but-unequal batches don't share a key. Doing it properly means per-key in-flight tracking with a recheck-and-narrow after waiting — which is why you've labelled it a heavy lift, and I agree. That's a real design change, not a fix, and it belongs behind a measurement rather than in front of one.

The metric to decide it now exists and is correct as of `66edddf`: `cache_misses_total{cache="search_hr",tier="l1"}` against request volume shows redundant fills directly. If that ratio turns out to be material in production, I'm happy to add per-key coalescing as a follow-up — say the word and I'll open an issue.

---

search-service/store_cache.go
```diff
@@ -0,0 +1,177 @@
...
+	for range missing {
+		rec.Miss(ctx)
+	}
+	for _, k := range missing {
+		v, found := loaded[k]
+		c.Add(k, cacheEntry[T]{val: v, found: found})
+		if found {
+			out[k] = v
+		}
```
julianshen, line 120: _🏷️ distributed-consistency_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`_

**A delayed cache fill can overwrite newer account state.**

Concurrent misses for the same key are filled without ordering or fencing. For example, request A reads an absent app, but its response is delayed; the app is then created, and request B reads and caches the new AppRef; when A resumes, this unconditional Add replaces the positive entry with a tombstone. Subsequent requests render the bare account for the full 24-hour TTL even though this pod already observed the app. Renames can similarly regress from the new name to the old one. Coalesce in-flight loads per key, or attach a generation/version to fills and reject an older completion before updating the cache.

<details>

```
In `search-service/store_cache.go` around line 98: Concurrent misses for the same key are filled without ordering or fencing. For example, request A reads an absent app, but its response is delayed; the app is then created, and request B reads and caches the new AppRef; when A resumes, this unconditional Add replaces the positive entry with a tombstone. Subsequent requests render the bare account for the full 24-hour TTL even though this pod already observed the app. Renames can similarly regress from the new name to the old one. Coalesce in-flight loads per key, or attach a generation/version to fills and reject an older completion before updating the cache.

```

</details>

---

search-service/store_cache.go
```diff
@@ -0,0 +1,177 @@
...
+// entirely in lookupCached and records nothing, so disabling a tier costs
+// performance and nothing else.
+//
+// expirable.LRU has no Close in v2.0.7 and its reaper goroutine runs for the
+// process lifetime, so build these once at startup — never per request.
+func newEntryLRU[T any](size int, ttl time.Duration) *lru.LRU[string, cacheEntry[T]] {
+	if size <= 0 || ttl <= 0 {
+		return nil
+	}
+	return lru.NewLRU[string, cacheEntry[T]](size, nil, ttl)
```
julianshen, line 50: _🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**A tiny positive cache TTL can crash the service.**

The TTL comes directly from environment configuration, but only non-positive values are rejected. In the pinned golang-lru v2.0.7 implementation, NewLRU starts its cleanup ticker with ttl/100; a value such as SEARCH_HR_CACHE_TTL=1ns therefore calls time.NewTicker(0) and panics in the reaper goroutine, terminating the process. Very small values also create an excessively busy cleanup loop. Validate both cache TTLs against a practical positive minimum before constructing the LRU, and add a boundary test for rejected or disabled undersized TTLs.

<details>

```
In `search-service/store_cache.go` around line 39: The TTL comes directly from environment configuration, but only non-positive values are rejected. In the pinned golang-lru v2.0.7 implementation, NewLRU starts its cleanup ticker with ttl/100; a value such as SEARCH_HR_CACHE_TTL=1ns therefore calls time.NewTicker(0) and panics in the reaper goroutine, terminating the process. Very small values also create an excessively busy cleanup loop. Validate both cache TTLs against a practical positive minimum before constructing the LRU, and add a boundary test for rejected or disabled undersized TTLs.

```

</details>

---

julianshen (commented):
### tGDBot review of `66edddf` — 2 findings in 6 files

2 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/449#issuecomment-5490809319), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

search-service/store_cache.go
evelynchang5525: Confirmed and fixed in `75f9a6b`. I reproduced it before changing anything rather than taking the reading on trust — `newEntryLRU[HRUser](8, 50*time.Nanosecond)` in a throwaway test:

```
panic: non-positive interval for NewTicker
goroutine 35 [running]:
time.NewTicker(0x0?)
github.com/hashicorp/golang-lru/v2/expirable.NewLRU[...].func1()
	golang-lru/v2@v2.0.7/expirable/expirable_lru.go:83
created by ...expirable.NewLRU[...] in goroutine 34
```

`expirable_lru.go:82-83` is `go func(...) { ticker := time.NewTicker(res.ttl / numBuckets) ... }` with `numBuckets = 100`, so any TTL under 100ns yields `NewTicker(0)`, and the panic is in a bare goroutine with no recover — the process dies.

Fixed by validation rather than clamping, so a misconfiguration can't be silently absorbed. `SearchConfig.Validate` rejects a positive TTL below `minCacheTTL` (1s) for both tiers, and `main.go` calls it alongside the existing `cfg.Valkey/Pool/Guard.Validate()` checks — failing startup with a named env var and a non-zero exit, per CLAUDE.md's fail-fast-on-config rule. Zero stays legal: it's the documented disable switch.

The floor is 1s rather than 100ns because the boundary isn't the only problem — a 10µs TTL wouldn't panic but would spin the reaper at 100ns intervals. Boundary tests cover 50ns, `minCacheTTL - 1ns`, exactly `minCacheTTL`, zero (disabled), and the shipped 24h defaults; `Validate` is at 100% coverage.

---

search-service/store_cache.go
evelynchang5525: Real, and materially worse because of the 24h retune — a 5m TTL made this self-heal in minutes. Mitigated in `75f9a6b`.

I took the targeted half rather than the full fencing, because the damaging case is specifically **tombstone over positive**: a bare account name pinned for a whole day. A tombstone is now not written when a live positive entry is already present:

```go
v, found := loaded[k]
if !found {
    if e, live := c.Get(k); live && e.found {
        continue
    }
}
c.Add(k, cacheEntry[T]{val: v, found: found})
```

This is precisely scoped to the race and no-ops otherwise. A key only reaches the fill loop if it *missed* at scan time, so a positive entry being present at `Add` time means someone filled it concurrently — which is exactly your scenario. In the normal path there is nothing to block: a live positive entry would have been a cache hit and never queried, and an expired one is gone by the time the tombstone lands. So it cannot cause a deleted row to linger.

The caller's own returned map still reflects what its query saw; only the cache is protected.

Test `TestLookupCached_TombstoneDoesNotClobberConcurrentPositiveFill` seeds the positive entry from *inside* the loader, which is the actual interleaving — my first attempt seeded it beforehand and was wrong, because the scan then saw a hit and never reached the load. I mutation-checked it: with the guard removed the test fails (the entry is clobbered by a zero-value `HRUser`), so it isn't a tautology. A companion test pins that a tombstone is still written when no concurrent fill happened.

**What this does not fix**, to be explicit: the positive→positive rename regression you also mention. Two concurrent loads both returning a row can still land older-over-newer, because without a generation/version there's nothing to compare. Closing that needs the per-key coalescing or version fencing tracked on the other thread — still my recommendation to decide with the (now correct) `cache_misses_total` signal rather than up front, but I'd rather name the remaining gap than imply this closes it.

---

search-service/store_cache.go
```diff
@@ -0,0 +1,199 @@
...
+	out := make(map[string]T, len(keys))
+	seen := make(map[string]struct{}, len(keys))
+	var missing []string
+	for _, k := range keys {
+		if _, dup := seen[k]; dup {
+			continue
+		}
+		seen[k] = struct{}{}
+		if e, ok := c.Get(k); ok {
+			rec.Hit(ctx)
```
mliu33, line 77: Can we remove metrics recorder as it does not provide much use for us?

search-service/store_cache.go
```diff
@@ -0,0 +1,199 @@
...
+		v, found := loaded[k]
+		if !found {
+			// Our load started before a concurrent one that has since cached a
+			// positive value — the row was created while our query was in
+			// flight. Writing this tombstone over it would render a bare
+			// account name for a whole TTL (a day at the default) even though
+			// this pod has already seen the row. Outside that race the key is
+			// absent or a tombstone, so this check no-ops.
+			if e, live := c.Get(k); live && e.found {
+				continue
```
mliu33, line 114: Should we put in the output before "continue"?

mliu33 (approved):
Awesome work! Just some minor comments, thanks!

PR #450, opened 2026-09-01
docs(load-testing): SLO calibration programme — execution plan, first-run runbook, and the P2 task brief

flywindy:
Docs-only. Turns "we drafted SLI/SLOs by gut feel and never validated them" into a scheduled programme with an executable first run.

## What's here

| File | What it is |
|---|---|
| `execution-priority-plan.md` | The scheduling document — four tracks, gate backlog G1–G10, execution rules |
| `first-slo-run-runbook.md` | The operator page for the first measuring run: preconditions, marks, queries, validity gate, run protocol |
| `first-slo-run-report.md` | The output contract, as a filled-in worked example (invented but internally consistent numbers) |
| `slo-measurement-map.md` | Per-journey, hop-by-hop: which instrument covers which segment, and where the gaps are |
| `p2-implementation-task.md` | The missing J1 instruments, written to be handed to an implementer |
| `p2-instrumentation-spec.md` | Rationale for P2 — superseded on every implementation detail by the task brief |
| `extreme-scenarios.md` | X1–X10 worst-case shapes derived from the code, with amplification arithmetic |

Plus corrections to three existing docs (below).

## What kind of review this needs

**This is a planning artefact. CI cannot validate it** — docs-only paths trigger no workflow, so there are zero check runs. Its correctness rests entirely on the code readings cited inline, and those citations have been wrong repeatedly: across review rounds this branch corrected label names, histogram suffixes, `rate()`'s reset semantics, the `withBroadcastMetricLabels` call-site count, the shipped `MaxDeliver`, and a recorder that would have counted a panic as a successful enqueue. None of that is visible in a diff.

**The useful review is someone checking the code citations**, particularly in `p2-implementation-task.md` §3–§5 and `first-slo-run-runbook.md` §5/§7. Line numbers will drift; the task brief says to re-verify with `grep` before editing, but the other documents cite line numbers too and will rot.

## Corrections to existing docs

- **`failure/nats-jetstream.md`, `failure/cassandra.md`, `extreme-scenarios.md` — the retry model was wrong everywhere.** All three assumed the repo-default `MaxDeliver`. `message-worker` (`main.go:360`) and `broadcast-worker` (`main.go:558`) wrap their settings in `stream.WithOutageRetryBudget`, which raises the cap until the schedule spans `stream.OutageRetryWindow` = 1 hour — **17 and 18 deliveries, roughly two hours of redelivery**, not twelve minutes. `message-gatekeeper` and `room-worker` take the plain default and are 6, not the 5 the docs claimed. This changes how long a failure test must run before a terminal drop is even possible.
- `failure/nats-jetstream.md` — also: `jsretry.DefaultBackoff`'s schedule was wrong, and the "gatekeeper Naks immediately" claim was wrong. It uses `jsretry.Nak` with a backoff, so the budget spans minutes, not milliseconds.
- `soak/cassandra-soak-plan.md` — documented the thread-count path as bounded to 99 rows. `pkg/threadcount/count.go` scans the whole partition; cost grows with thread length. Also reconciles a `compaction_window_size` discrepancy.

## What this does *not* claim

Worth stating up front, because two rounds of review caught the docs overclaiming it: **P2 does not make any J1 SLO hard-gateable.** Every numerator is consumer-side and counts delivery attempts, so compensating errors cancel — one logical message lost and another processed twice under a failed Ack reads as `good=2 / valid=2 = 100%`, with the backlog drained and the lag signal clean. G4 is now written as *computable and calibration-ready*; a real gate needs P7's logical-outcome dedup ledger, or a per-window redelivery-bias bound. Where neither exists the verdict is `INCONCLUSIVE`, not a number.

## Two things that are deliberately open

**1. `docs/reviews/` was not cleaned.** `CLAUDE.md` says to delete everything under it before opening a PR. The two files there came from `main`, not from this branch — deleting them would widen this PR into unrelated cleanup and remove another author's working notes. Suggest a separate PR; say the word if you'd rather it happen here.

**2. Federation is an unanswered product question.** Track 4.6 asks whether production launches multi-site. If it does, federation is a launch blocker rather than a smoke test: SLO-9 is entirely unvalidated, no loadgen traffic drives it, `outbox-worker`/`inbox-worker` carry no consumer or domain metrics, and X8's arithmetic suggests the 30 s bound may be unreachable for a bulk membership change at realistic inter-site RTT. The row is written as an open decision on purpose — it needs an answer, not a review.

## Not affected

`docs/client-api.md` — no client-facing handler, request/response struct, or event struct changes. No Go changes at all, so no lint/test/coverage gates apply.


julianshen:
**29 findings · 24 unique issues · 8 inline comments posted** — 🔴 6 blocking · 🟠 22 warning · 🔵 1 suggestion

> [!NOTE]

### 💬 Additional comments (16)


**`docs/load-testing/slo/first-run-runbook.md`**

_🎯 NATS correctness_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`distributed-system`, `nats` (2 rules)_

**The SLO-2 query excludes every failed enqueue.**

Query 6 in §9a divides the age bucket by `broadcast_channel_enqueue_age_seconds_count`. Section 5.1 records that histogram only when the room-subject enqueue is accepted. If `publishRoomEvent` fails, or broadcast-worker never consumes the canonical message, neither the bucket nor `_count` changes, so the ratio can remain 100% despite an SLO-2 failure. Use the run-window delta of the upstream `messages_canonical_published_total{broadcast_path="room_subject"}` denominator, applying the documented `unknown` worst-case rule, while retaining the age bucket as the numerator.

<details>

- **`distributed-system`**: The §9a query uses the age histogram `_count` as denominator, although §1 defines the same upstream denominator as SLO-1b and the histogram is recorded only when enqueue succeeds. With 100 room-subject canonicals, 50 fast successful enqueues and 50 failed enqueues, bucket/count reports 50/50 = 100%; the actual ratio is 50/100. Divide each age bucket by `messages_canonical_published_total{broadcast_path="room_subject"}` using the t2−t0 delta and the required worst-case handling for `unknown`; retain histogram count only as a diagnostic.

</details>

---

**`docs/load-testing/slo/p2-implementation-task.md`**

_🏷️ time-consistency_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`, `nats` (2 rules)_

**The age histogram subtracts clocks from different hosts.**

`msg.Metadata().Timestamp` is assigned using the NATS server's wall clock, while `time.Now()` uses the broadcast-worker host's clock. Being served by that broker does not put the worker on the same clock. If the broker is 1.2 seconds behind, immediate enqueues fail the 1-second SLO; if it is ahead, valid samples become `negative_age` and invalidate the window. Require and monitor a clock-offset bound tight enough for this SLO, propagate that uncertainty into verdicts, or move the age calculation to a single-clock measurement boundary.

<details>

- **`nats`**: Section 5.1 subtracts the NATS server-assigned `msg.Metadata().Timestamp` from broadcast-worker's local `time.Now()` and claims this avoids clock skew. These timestamps still come from different hosts. A worker clock 500 ms ahead inflates every measured age by 500 ms, while only negative skew is rejected by `_age_invalid_total`; valid messages can therefore be scored late against the 1 s bound. Add a precondition and validity check that measures and bounds NATS-server-to-worker clock offset, or redesign the measurement so both endpoints use one clock domain.

</details>

---

**`docs/load-testing/slo/first-run-runbook.md`**

_🏷️ measurement-consistency_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`distributed-system`_

**Raw snapshots produce invalid deltas after a service restart.**

The method subtracts instant counter sums, but the validity gate does not reject service restarts or replica changes. If a history-service replica contributes 1,000,000 calls at t0, restarts, and its old series becomes stale before t1, t1 contains only the new process's counter; the delta loses pre-restart observations or becomes negative. Ordinary rescheduling or a rolling deployment can therefore publish nonsensical SLO curves. Add a validity check that invalidates any window with target/restart changes, or use reset-aware calculations over explicit t0–t1 and t0–t2 intervals.

---

**`docs/load-testing/slo/first-run-runbook.md`**

_🏷️ NATS lifecycle_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`nats`_

**The t2 barrier ignores relevant independent consumers.**

The §5 table and §7 step 6 mark `t2` when only message-worker reaches its floor, while §9a says this protocol remains unchanged for SLO-1b/2. The canonical consumers are independent: broadcast-worker can remain backlogged after message-worker drains, causing its later enqueues to be excluded from the snapshot and producing a false failure. Likewise, gatekeeper can still have delayed or ack-pending admissions when SLO-1a is snapped. Require the MESSAGES gatekeeper durable and every numerator-producing MESSAGES-CANONICAL durable relevant to the reported SLOs to reach their pre-run floors and remain stable before both async barriers.

---

**`docs/load-testing/extreme-scenarios.md`**

_🏷️ cassandra-data-modeling_ | _🟠 Warning_ | _⚡ Quick fix_ | _`cassandra`_

**The hot-partition threshold is understated.**

For a 360-hour bucket, the stated 100,000-row threshold is reached at `100000 / (360 × 60) = 4.63` messages/minute, not roughly 7; at 7 messages/minute the partition already has 151,200 rows. The lowest proposed test rate, 1 message/second, produces 1.296 million rows per full bucket and skips the threshold by about 13×, so it cannot locate the safe-rate transition. Correct the arithmetic and add seeded cases around 0.05, 0.08, and 0.15 messages/second equivalent.

---

**`docs/load-testing/slo/p2-implementation-task.md`**

_🏷️ delivery-semantics_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`, `silent-failure` (2 rules)_

**Lost PubAcks can remove published messages from the denominator.**

If the broker commits a canonical publish but its PubAck is lost, the first call returns an error and records nothing. The retry receives a duplicate PubAck, which this design records only in the unlabelled duplicate counter, so the published message is absent from the SLO denominator while broadcast-worker can increment the numerator. This can bias SLO-1b green and cancel a separate failed enqueue. Treat any duplicate as INCONCLUSIVE, or include all duplicates in a conservative worst-case denominator; merely documenting the approximation does not bound it.

<details>

- **`silent-failure`**: §4.1 excludes duplicate PubAcks from the denominator. If the original publish commits but its ack is lost, the first attempt is not counted and the retry is classified as duplicate and also excluded. If that canonical event then fails persistence or enqueue, its numerator is absent too, so the reported ratio remains green. Although `_duplicate_total` exposes the ambiguity, the runbook never makes it a validity gate. Mark windows with duplicate deltas INCONCLUSIVE unless an exact broker/ledger denominator or a conservative bound is used.

</details>

---

**`docs/load-testing/slo/p2-implementation-task.md`**

_🏷️ observability_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`_

**The acceptance checklist reintroduces the unsafe named-return pattern.**

Section 4.2 mandates a failed-by-default outcome sentinel and explicitly forbids deriving the outcome from a named return because a panic leaves `err` nil. The P2a checklist later requires recording the failure paths “via a named return + defer.” Following that criterion can report `ok` when a panic unwinds and `jobguard` subsequently Acks and drops the canonical message. Replace the checklist wording with “failed-default outcome sentinel + defer” and explicitly retain the panic test.

---

**`docs/load-testing/execution-priority-plan.md`**

_🏷️ elasticsearch-operations_ | _🟠 Warning_ | _⚡ Quick fix_ | _`elasticsearch`_

**The search ceiling is scheduled without its Elasticsearch gate.**

The readiness snapshot says the run-scoped index and ES telemetry contract are both absent and therefore block ES capacity, but Track 2.6 permits the staging search ramp with only G1. Following that row can drive search to breach against an unscoped Elasticsearch deployment without shard, rejection, breaker, merge, or watermark evidence. Split login and search into separate rows and require G7 for the staging search run; verify both search-sync writes and search-service reads use the same run-owned index or alias before starting.

---

**`docs/load-testing/slo/measurement-map.md`**

_🏷️ elasticsearch-search-behavior_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`elasticsearch`_

**No-hit queries cannot calibrate representative search availability.**

The soak search lane is documented as producing queries that match nothing, yet its `status="ok"` ratio is used as SLO-7 calibration evidence. A no-hit Elasticsearch request avoids the fetch, source decoding, and any hit-processing path, so failures that occur only when returning real results cannot affect this ratio. The resulting availability can be materially greener than user searches. Seed searchable documents and require a representative nonzero hit distribution before using this lane for calibration, or explicitly report the result as no-hit query-path availability and exclude it from target selection.

---

**`docs/load-testing/slo/p2-implementation-task.md`**

_🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`nats`, `silent-failure`, `tgd-review` (3 rules)_

**The acceptance checklist contradicts the panic-safe recorder design.**

Section 4.2 correctly requires an outcome sentinel initialized to `failed` and explicitly rejects inspecting a named return in the defer, because a panic leaves that return nil and would record a successful enqueue before `jobguard` Acks the dropped message. The P2a acceptance checklist instead requires recording all failure paths “via a named return + defer.” An implementer following the checklist can ship the exact green-bias bug the design warns against. Change the checklist to require the default-failed sentinel plus defer.

<details>

- **`silent-failure`**: §4.2 requires a default-`failed` outcome sentinel and explicitly rejects deriving the outcome from a named return: during a panic, the named error can remain nil while `jobguard` recovers and Acks the dropped message. The §6 checklist contradicts this by requiring recording “via a named return + defer” and enumerating only the three ordinary error returns. Replace that checklist item with the sentinel design and require panic as the fourth failed exit path, so implementers cannot satisfy the checklist with the unsafe recorder.
- **`tgd-review`**: Section 4.2 explicitly forbids a named return because a panic can leave `err == nil` and record a successful enqueue before `jobguard` Acks the dropped message. The P2a checklist later requires the recorder to fire “via a named return + defer,” contradicting that design and its panic test. Change the checklist to require the failure-default outcome sentinel plus `defer`; otherwise the implementer can satisfy the final checklist while introducing the exact false-green metric the brief warns against.

</details>

---

**`docs/load-testing/slo/first-run-runbook.md`**

_🎯 concurrency_ | _🟠 Warning_ | _⚡ Quick fix_ | _`performance-and-concurrency`_

**The P2 settle barrier ignores the independent broadcast consumer.**

Section 9a snapshots the new enqueue metrics at the existing `t2`, but sections 5 and 7 define `t2` only by message-worker's pending and ack-pending counts. The canonical stream feeds message-worker and broadcast-worker independently, so under broadcast-worker slowdown message-worker can reach its floor first. The snapshot then omits later enqueue successes and age observations, materially understating SLO-1b/2. Require broadcast-worker's durable to remain at its pre-run floor for a scrape interval too, using the slower consumer's cap, or define a separate broadcast settle mark.

---

**`docs/load-testing/slo/first-run-runbook.md`**

_⚡ performance_ | _🟠 Warning_ | _⚡ Quick fix_ | _`performance-and-concurrency`_

**Growing backlog is incorrectly classified as an invalid measurement.**

Section 6 makes monotonically growing consumer backlog invalidate the entire window. When dispatch ratio is achieved and the observer remains healthy, that growth is the system failing to sustain offered load—the exact backpressure or capacity result the programme seeks. Discarding such runs creates survivorship bias and can calibrate targets only from runs where queues stayed flat. Distinguish apparatus failure from system failure: missing/unreliable backlog telemetry is INCONCLUSIVE, while observed backlog growth is a valid unsustainable-load result; only SLIs that require a completed settle window should remain unavailable.

---

**`docs/load-testing/slo/first-run-runbook.md`**

_🏷️ reliability_ | _🟠 Warning_ | _⚡ Quick fix_ | _`performance-and-concurrency`_

**Raw counter subtraction does not survive service restarts.**

The scored queries subtract instant counter snapshots, but the validity gate does not detect restarts of gatekeeper, message-worker, history-service, or search-service. If load triggers an OOM or a pod rolls between marks, its counters reset; after the old Prometheus series becomes stale, `@t1/@t2 - @t0` can become negative or omit pre-restart work, producing materially wrong ratios. Add a restart/process-start-time validity check for every contributing service and suppress affected ratios while retaining the restart as a system failure, or calculate reset-aware increases over each exact marked interval.

---

**`docs/load-testing/slo/first-run-runbook.md`**

_🏷️ failure-masking_ | _🟠 Warning_ | _⚡ Quick fix_ | _`silent-failure`_

**Rolling validity queries can forget earlier invalid samples**

§9a checks `broadcast_path="unknown"` and `_age_invalid_total` with `increase(...[$window])`, despite §5 requiring fixed snapshots because rolling windows slide. If either invalid event occurs early in the measurement and the query is evaluated or refreshed after it leaves `$window`, the gate reads zero; the omitted denominator or age sample can then produce an apparently healthy ratio. Record these counters at `t0-async` and `t2` and gate on their exact snapshot deltas, just like the associated SLO ratios.

---

**`docs/load-testing/slo/p2-implementation-task.md`**

_🏷️ planning_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**P2a acceptance incorrectly requires a P2b metric.**

The P2a checklist requires every unsmokeable family, including `_age_invalid_total`, to be pinned by tests. Section 5.2 defines that counter as P2b work, while §1 requires P2a to be independently mergeable without P2b. As written, P2a cannot satisfy its own acceptance criteria without widening scope. Remove `_age_invalid_total` from the P2a item and add the exact-name/label assertion to the P2b checklist.

---

**`docs/load-testing/slo/first-run-runbook.md`**

_🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**The broadcast-worker retry cap does not match the stated formula.**

Using the document's own inputs for `MaxDeliver=18`, LowLatencyBackoff `0.2/1/5/30s` with the last value repeated, and consumer delays `30/60/120/240/480s` repeated, the 17 per-delivery maxima sum to 6,690s, not 8,130s. Adding the stated 510s execution margin yields about two hours plus one scrape, rather than 2h25m. Recalculate the broadcast-worker row and the corresponding X9 scheduling guidance, or document the additional delays currently omitted from the formula.

<details>
<summary>📒 Files reviewed (11)</summary>

* `docs/load-testing/README.md`
* `docs/load-testing/execution-priority-plan.md`
* `docs/load-testing/extreme-scenarios.md`
* `docs/load-testing/failure/cassandra.md`
* `docs/load-testing/failure/nats-jetstream.md`
* `docs/load-testing/slo/first-run-report.md`
* `docs/load-testing/slo/first-run-runbook.md`
* `docs/load-testing/slo/measurement-map.md`
* `docs/load-testing/slo/p2-implementation-task.md`
* `docs/load-testing/slo/p2-instrumentation-spec.md`
* `docs/load-testing/soak/cassandra-soak-plan.md`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

docs/load-testing/failure/nats-jetstream.md
```diff
@@ -15,8 +15,8 @@
...
 | Request/reply | One timeout, no application retry | Failover shows up as a timeout or as no-responders. A side-effecting RPC whose reply was lost may still have run |
 | Readiness | `RECONNECTING` counts as ready; only `DISCONNECTED`/`CLOSED` do not | A pod stays ready for as long as it keeps reconnecting, while none of its NATS-backed functionality works |
-| Consumer defaults | AckWait 30s, `MaxDeliver=5`, `MaxAckPending=1000` | Retries are finite. An outage longer than the delivery budget ends in a terminal drop, not in a pending backlog |
-| Retry backoff | `jsretry.DefaultBackoff` 1s/5s/30s/2m; low-latency 200ms/1s/5s/30s | Against `MaxDeliver=5` the whole budget is spent in minutes. Max-delivery advisories are the only enumeration of what was dropped |
+| Consumer defaults | AckWait 30s, `MaxDeliver=6` (`pkg/stream/consumer.go:18`), `MaxAckPending=1000` | Retries are finite. An outage longer than the delivery budget ends in a terminal drop, not in a pending backlog |
```
julianshen, line 18: _🏷️ retry-behavior_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`distributed-system`_

**Broadcast handler failures exhaust retries in minutes, not two hours.**

A delayed NAK schedules the next delivery itself; consumer `BackOff` applies when a message is left unacknowledged. The runbook identifies broadcast-worker's application schedule as 200 ms/1 s/5 s/30 s with the final value repeated and `MaxDeliver=18`. Seventeen handler-error delays therefore total about 426 seconds, not two hours. A roughly ten-minute downstream outage can exhaust and drop broadcast work while this table says budget remains. Document and test handler-error exhaustion separately from crash/hang/AckWait exhaustion throughout the affected plans.

<details>

```
In `docs/load-testing/failure/nats-jetstream.md` around line 18: A delayed NAK schedules the next delivery itself; consumer `BackOff` applies when a message is left unacknowledged. The runbook identifies broadcast-worker's application schedule as 200 ms/1 s/5 s/30 s with the final value repeated and `MaxDeliver=18`. Seventeen handler-error delays therefore total about 426 seconds, not two hours. A roughly ten-minute downstream outage can exhaust and drop broadcast work while this table says budget remains. Document and test handler-error exhaustion separately from crash/hang/AckWait exhaustion throughout the affected plans.

```

</details>

---

docs/load-testing/extreme-scenarios.md
```diff
@@ -0,0 +1,474 @@
...
+| 50 (I7 p99) | 50 | 1 |
+| 500 (I7 max) | 500 | 1 |
+| 5 000 | 5 000 | 1 |
+| 50 000 | 50 000 | 10 |
+| 200 000 | 200 000 | 40 → likely hits the 15 s ceiling |
+
+**Verdict: not acceptable as written.** The 15 s backstop converts an unbounded
+partition into a *timeout* rather than into back-pressure, and the timeout fires
+inside a JetStream handler with `AckWait=30s`. The path runs in `message-worker`
+(`message-worker/store_cassandra.go`), whose `MaxDeliver` is **17**, not the repo
```
julianshen, line 80: _🎯 cassandra-correctness_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`cassandra`_

**X1's oracle misses partially persisted thread replies.**

The documented mutation sequence inserts the reply before `CountAndLatest`, then updates two parent rows. A 15 s scan timeout on the final delivery therefore need not drop the reply; it can leave child rows committed while the parent's count/latest metadata remains stale. Concurrent scans can likewise write an older summary last, since `IF EXISTS` checks existence rather than monotonicity. The proposed read-back checks only reply IDs, so it passes this inconsistency. Reconcile parent count/latest against `thread_messages_by_thread` after concurrent sends and timeout exhaustion, and correct the expected failure mode.

<details>

```
In `docs/load-testing/extreme-scenarios.md` around line 80: The documented mutation sequence inserts the reply before `CountAndLatest`, then updates two parent rows. A 15 s scan timeout on the final delivery therefore need not drop the reply; it can leave child rows committed while the parent's count/latest metadata remains stale. Concurrent scans can likewise write an older summary last, since `IF EXISTS` checks existence rather than monotonicity. The proposed read-back checks only reply IDs, so it passes this inconsistency. Reconcile parent count/latest against `thread_messages_by_thread` after concurrent sends and timeout exhaustion, and correct the expected failure mode.

```

</details>

---

docs/load-testing/soak/cassandra-soak-plan.md
```diff
@@ -143,12 +143,19 @@ is non-destructive. It therefore does **not** cover:
...
 ## D. Config Facts (verified in repo)
 
-- **`compaction_window_size = 72 HOURS`** — pinned in
-  `docker-local/cassandra/init/10-table-messages_by_room.cql:41`,
-  `docker-local/cassandra/migrations/2026-05-twcs-message-tables.cql:40`,
-  `docs/cassandra_message_model.md:160`. Coupled to `MESSAGE_BUCKET_HOURS`
-  (`message-worker/main.go:39`, `history-service/internal/config/config.go:46`,
-  envDefault `72`) — the two must change together.
+- **`compaction_window_size` — the repo disagrees with itself, CONFIRM BEFORE ANY RUN.**
+  The init DDL sets **360**
```
julianshen, line 147: _🏷️ cassandra-migration_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`cassandra`_

**The TWCS mismatch is detected but never remediated.**

The plan identifies init DDL at 360 hours and the applied migration at 72 hours, but schedules only querying the deployed value. An environment built through that migration remains mismatched with 360-hour application buckets, blocking runs under execution rule 7 and amplifying SSTables per partition read. Add an owned gate for a new forward `ALTER TABLE` migration rather than editing the already-applied migration, update every schema mirror, roll it through existing environments, and verify the live schema before declaring a run ready. The rollout must also account for existing SSTables.

<details>

```
In `docs/load-testing/soak/cassandra-soak-plan.md` around line 147: The plan identifies init DDL at 360 hours and the applied migration at 72 hours, but schedules only querying the deployed value. An environment built through that migration remains mismatched with 360-hour application buckets, blocking runs under execution rule 7 and amplifying SSTables per partition read. Add an owned gate for a new forward `ALTER TABLE` migration rather than editing the already-applied migration, update every schema mirror, roll it through existing environments, and verify the live schema before declaring a run ready. The rollout must also account for existing SSTables.

```

</details>

---

docs/load-testing/extreme-scenarios.md
```diff
@@ -0,0 +1,474 @@
...
+message = up to N subscription-document writes. An edit repeats it. This is the
+`end-to-end-plan.md` T1-5 item, still marked unasserted.
+
+| Room size | Sub docs written per `@all` | 1 `@all`/s |
+|---|---:|---:|
+| 100 | 100 | 100 writes/s |
+| 2 000 | 2 000 | 2 000 writes/s |
+| 5 000 | 5 000 | 5 000 writes/s |
+
+**Verdict: acceptable shape, unvalidated magnitude.** `UpdateMany` is the right
```
julianshen, line 163: _🏷️ mongodb-atomicity_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`mongodb`_

**The @all test does not verify partial UpdateMany recovery.**

A primary stepdown can interrupt this multi-document `UpdateMany` after only some subscription documents are updated; MongoDB guarantees atomicity per document, not for the whole matched set. The proposed test measures latency and index use but could call the shape acceptable without checking whether normal retries update the remaining members or duplicate effects on already-updated members. Add failover injection and a full target-set read-back before accepting the result.

<details>

> ‼️ **IMPORTANT**

```suggestion
**Test.** Mention-storm: `@all` at 0.2 / 1 / 5 per second into rooms of 500 /
2 000 / 5 000, with an edit-follow-up share matching I4. Measure Mongo write p99,
`db.client.operation.duration`, oplog throughput, and whether the filter uses the
`(roomId, u.account)` index. Force a primary failover while an `UpdateMany` is in
flight, allow the production retry path to run, then read back every targeted
subscription and verify convergence without omissions or duplicate mention effects.
Mongo may be driven to capacity (environments §3).
```

</details>

<details>

```
In `docs/load-testing/extreme-scenarios.md` around line 160: A primary stepdown can interrupt this multi-document `UpdateMany` after only some subscription documents are updated; MongoDB guarantees atomicity per document, not for the whole matched set. The proposed test measures latency and index use but could call the shape acceptable without checking whether normal retries update the remaining members or duplicate effects on already-updated members. Add failover injection and a full target-set read-back before accepting the result.

```

</details>

---

docs/load-testing/extreme-scenarios.md
```diff
@@ -0,0 +1,474 @@
...
+1. **`CLIENTSIM_RAMP_RATE` defaults to 50/s** — a paced ramp, which is what a
+   connection-plane soak wants and the opposite of what this scenario is. Raise
+   it: 5 000 clients at 2 000/s arrive in 2.5 s.
+2. **The *reconnect* variant is better driven by a fault than by a ramp.** Hold
+   the fleet at target, then drop the WebSocket listener or the LB and let every
+   client reconnect and re-walk at once. That is the actual production mechanism
+   (a gateway blip), and it is already clientsim's own path — post-reconnect
+   resync with bounded backoff. Pair it with the failure programme rather than
+   running it as a fresh ramp.
+
```
julianshen, line 263: _🏷️ mongodb-observability_ | _🟠 Warning_ | _⚡ Quick fix_ | _`mongodb`_

**The login-storm test omits its quantified MongoDB workload.**

The scenario calculates 65,000 back-to-back `subscription.list` calls for 5,000 users with 500 rooms and states that each page runs a Mongo aggregation, but the test records only RPC, Cassandra, and NATS signals. If Mongo query latency, pool contention, or pagination plan degradation causes the RPC knee, the run cannot attribute it. Record user-service Mongo operation rate/latency and pool pressure, and inspect representative page query plans.

<details>

> ‼️ **IMPORTANT**

```suggestion
**Test.** Cold-start storm, `clientsim`-driven. Step U through 500 / 1k / 2k /
5k at a burst ramp rate, with loadgen seeding and running background traffic.
Measure `subscription.list` and `channel_history` p99 (both are
`rpc_server_call_duration_seconds` methods since #337), user-service Mongo command
rate and `db.client.operation.duration`, Mongo pool pressure, Cassandra read latency,
and NATS request/reply timeouts. Capture `explain("executionStats")` for representative
early and late `subscription.list` pages, and retain the `includeLastMessage` A/B.
Then repeat the reconnect variant under a listener fault. This is also the only
realistic way to exercise SLO-3's declared journey (auth → connect → initial data)
end to end — which loadgen cannot do, since it never touches auth-service's HTTP leg.
```

</details>

<details>

```
In `docs/load-testing/extreme-scenarios.md` around line 256: The scenario calculates 65,000 back-to-back `subscription.list` calls for 5,000 users with 500 rooms and states that each page runs a Mongo aggregation, but the test records only RPC, Cassandra, and NATS signals. If Mongo query latency, pool contention, or pagination plan degradation causes the RPC knee, the run cannot attribute it. Record user-service Mongo operation rate/latency and pool pressure, and inspect representative page query plans.

```

</details>

---

docs/load-testing/slo/first-run-report.md
```diff
@@ -0,0 +1,300 @@
...
+
+---
+
+## Part 3 — The validity gate
+
+Read this **before** the ratios (§6). All six must pass or the run is
+INCONCLUSIVE regardless of how the numbers look.
+
+| Check | M1 | M2 | M4 |
+|---|---|---|---|
```
julianshen, line 68: _🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**The report template omits required validity checks.**

The template says all six checks must pass, but its checklist differs from the runbook §6 gate. It omits `loadgen_failure_invalidations_total`, Mongo probe/heartbeat freshness, and the observer `unverified` threshold, replacing them with `t2` timing and neighbour activity. An operator can therefore complete this output contract and report a run that the SOP declares inconclusive. Align the template with the runbook, adding run-specific checks separately rather than substituting them.

<details>

```
In `docs/load-testing/slo/first-run-report.md` around line 68: The template says all six checks must pass, but its checklist differs from the runbook §6 gate. It omits `loadgen_failure_invalidations_total`, Mongo probe/heartbeat freshness, and the observer `unverified` threshold, replacing them with `t2` timing and neighbour activity. An operator can therefore complete this output contract and report a run that the SOP declares inconclusive. Align the template with the runbook, adding run-specific checks separately rather than substituting them.

```

</details>

---

docs/load-testing/slo/first-run-runbook.md
```diff
@@ -0,0 +1,994 @@
...
+> Why it works and which SLOs it covers:
+> [`measurement-map.md`](measurement-map.md) §7.
+
+| | |
+|---|---|
+| **Produces** | Six artefacts, listed in **§7a** — achieved good-ratio *curves* for **SLO-4/5**, a ratio for **SLO-7**, an approximate indicator for **SLO-1a**, one-sided bounds for **1b/2**, the loadgen-vs-production latency gap, and the verified measurement apparatus itself. Evidence a target gets chosen *from*, not a verdict against one |
+| **How many runs** | **3–6 shakedown runs** (instrument only, nothing reported) then **3 identical measurement runs**, median + spread. Which knobs may move between them, and which may not: **§7b** |
+| **Stance** | **This run is calibration-only and gates nothing.** `common/sli-slo.md` remains the acceptance contract for any gating run until Track 1.3 amends it — §4a |
+| **loadgen code changes** | **None** — but the warm-up→measurement boundary needs a run protocol, not a code path; see §7 |
+| **Dashboard changes** | **Yes** — the existing dashboard has 22 panels and not one SLO ratio |
```
julianshen, line 17: _🏷️ intent-match_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**The scheduled first run cannot produce the promised SLO-3 evidence.**

The execution plan's Track 1.0b promises SLO-3/4/5/7 evidence and an output covering seven of nine SLOs, but this operator runbook explicitly excludes SLO-3. Its protocol only runs the soak; §8 says SLO-3 instead requires a separate `max-rps --workload=login` run, which is absent from the sequence and artefacts. Either schedule that companion run here or revise Track 1.0b to promise six SLOs.

<details>

```
In `docs/load-testing/slo/first-run-runbook.md` around line 17: The execution plan's Track 1.0b promises SLO-3/4/5/7 evidence and an output covering seven of nine SLOs, but this operator runbook explicitly excludes SLO-3. Its protocol only runs the soak; §8 says SLO-3 instead requires a separate `max-rps --workload=login` run, which is absent from the sequence and artefacts. Either schedule that companion run here or revise Track 1.0b to promise six SLOs.

```

</details>

---

docs/load-testing/slo/measurement-map.md
```diff
@@ -0,0 +1,580 @@
+# SLO Measurement Map
+
+> For each journey: the real path hop by hop, which instrument sits on which hop
+> today, where the SLO's declared boundary actually falls, and which segments are
+> dark. Read this before arguing about whether an SLO "passed" — most
+> disagreements are really about which hop the number came from.
+>
+> Verified against the code on 2026-08-27. Rows marked **#337** land when that PR
```
julianshen, line 8: _📝 documentation_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`tgd-review`_

**The map still describes merged metrics as pending.**

This header says #337-marked rows will land when the PR merges, while the execution plan and runbook state that #337 merged on 2026-08-30 and is already on `main`. Updating the header avoids making current instrumentation look unavailable.

<details>

> ‼️ **IMPORTANT**

```suggestion
> Verified against the code on 2026-08-27. Rows marked **#337** are on `main`
> since #337 merged on 2026-08-30 (`bf0ea62`).
```

</details>

<details>

```
In `docs/load-testing/slo/measurement-map.md` around line 7: This header says #337-marked rows will land when the PR merges, while the execution plan and runbook state that #337 merged on 2026-08-30 and is already on `main`. Updating the header avoids making current instrumentation look unavailable.

```

</details>

---

julianshen (commented):
### tGDBot review of `c04ce29` — 29 findings in 11 files

6 🔴 blocking · 22 🟠 warning · 1 🔵 suggestion
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/450#issuecomment-5581153442), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

PR #446, opened 2026-08-31
fix(room-worker): render bot app name in member add/remove system messages

nalydadad:
## Problem

Member add/remove/self-leave system messages compose the member's display name from the `users` collection fields (`EngName`/`ChineseName`/`Account`) with no bot override. A bot's user record has no human names, so the composer falls back to the raw `.bot` account string — e.g. `"helper.bot" left the chatroom`.

Every other member-name-rendering path (room-list preview, reactions, live bot messages) goes through the shared `preview.BotAwareDisplayName`, which substitutes the bot's registered app name from the `apps` collection. This sys-message write path in `room-worker` was the only one missing it.

## Fix

- Route the member side of all four sys-message composition sites (self-leave, forced remove, add members, room-create members_added) through a new `Handler.memberDisplayName`, which delegates to the shared `preview.BotAwareDisplayName`.
- Adapt the existing `store.GetApp` to the `preview.AppNameLookup` contract (`ErrAppNotFound` → `("", nil)`), wrapped in the existing `preview.CachedAppNameLookup` (TTL LRU + singleflight) at `NewHandler`, matching how broadcast-worker and history-service consume the same lookup.
- `sysmsg.go` format helpers now take a pre-resolved member name string; the sentence templates are unchanged.

Degradation: app not found or Mongo error falls back to today's composed name (raw account for bots) and the message still publishes — a display name is never worth failing a membership flow over.

## Scope (deliberate)

- Requester names are unchanged (bot-initiated flows go through bot-room-service, which emits no Content).
- `SysMsgData` wire shape unchanged; no client-facing schema change, so no `docs/client-api.md` update needed.
- No backfill of historical messages already persisted in Cassandra.

## Testing

- TDD: new handler-level tests for bot removed / bot self-leave (table-driven: found / not-found / lookup-error) / single bot added, all red before the fix.
- Human paths keep their exact expected strings; gomock's no-expectation rule enforces `GetApp` is never called for non-bot members.
- `make test SERVICE=room-worker` (race) green; `golangci-lint run ./room-worker/...` 0 issues.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Bot-related member-added and member-removed messages now display the registered app name.
  * Improved fallback behavior when an app name is unavailable or lookup fails.
  * Self-leave messages now consistently use the account name when no app name can be resolved.

Issue #366, opened 2026-08-24
broadcast-worker: bot app-name lookup is an uncached Mongo read per bot message on the fan-out path


hmchangw:
Raised by @julianshen in review on #222.

## What

`appNameRepo.lookup` (`broadcast-worker/preview.go:128`) issues a Mongo `FindOne` against `apps` for **every eligible message sent by a bot account**, with no TTL cache and no singleflight:

```go
func (r *appNameRepo) lookup(ctx context.Context, botAccount string) (string, error) {
	app, err := r.apps.FindOne(ctx, bson.M{"assistant.name": botAccount},
		mongoutil.WithProjection(bson.M{"name": 1, "_id": 0}))
	...
}
```

It is reached from `preview.BotAwareDisplayName`, which gates on `model.IsBot(account)` — so ordinary user traffic never pays it. Bot traffic pays it once per message.

## Why it matters here specifically

The read pattern is not new: `history-service/internal/mongorepo/app.go:27` has had the identical uncached `FindOne` on `main` for reaction Actor rendering and the lazy preview walk. What #222 changes is **which path it sits on**.

- On `main`, it fires per *read* (rooms.get walk) and per *reaction*.
- With #222, it also fires per *insert*, inside `broadcast-worker`'s message fan-out — the hottest path in the system, and the one whose throughput bounds delivery latency for every room member.

A bot posting at a steady rate produces one `apps` read per message, all for a value that changes approximately never.

## Blast radius

Bounded, but real:

- Correctness is unaffected. The lookup runs inside the seal's own 2s budget (`previewSealTimeout`), and a failure degrades to the composed display name with a `Warn` — it never fails the fan-out.
- Throughput is the exposure. During `apps`-collection latency, every bot message spends up to the seal budget waiting on a read whose answer is static, holding a worker slot that the required fan-out needs.

## Suggested fix

A bounded TTL cache keyed by `botAccount`, plus singleflight so a cold key under concurrent bot traffic collapses to one read. Both `broadcast-worker` and `history-service` should share it — the two call sites already share `preview.BotAwareDisplayName`, so the cache belongs behind the same `preview.AppNameLookup` seam rather than being duplicated.

Negative caching matters too: `("", nil)` for a non-app bot account is a legitimate result and should be cached, or a misconfigured bot generates the same per-message read forever.

TTL can be generous (minutes) — an app rename showing a stale name for a few minutes is not a correctness concern.

## Not in #222

[...]

PR #451, opened 2026-09-01
fix(history-service): resolve bot app names and normalize legacy member sys-msg types

evelynchang5525:
Two read-path follow-ups to #404 / #413 on legacy migrated system messages, plus the type normalization the frontend needs.

## Why

#404 / #413 swapped the raw account baked into a legacy `members_removed` sentence for a display name, but resolved only through the `users` collection. An account is a **user or a bot**, and a `.bot` account frequently has no user document at all — so those rows still rendered the raw account.

Separately, the migrated Cassandra rows carry plural types (`members_removed`, `members_left`) that no client knows.

## What changed

**Bot names.** A `.bot` account now resolves to its registered app name (`apps.assistant.name == account` → `apps.name`), degrading to the composed user name and then the raw account. Same precedence `room-worker` adopted for the live path in #446, so a legacy row and a modern one render a bot identically. Bot accounts stay in the existing `FindUsersByAccounts` batch too — a bot may carry a user document, and that batch is issued either way.

**Type normalization**, on the wire only — nothing is written back to Cassandra:

| stored | returned |
|---|---|
| `members_removed` | `member_removed` |
| `members_left` | `member_left` (text preserved verbatim) |

Unconditional and independent of the text rewrite: a `members_removed` row whose sentence never matched the suffix still gets the type its client expects.

**Ordering is load-bearing.** `extractRemovedAccount` gates on the *legacy plural* type, so normalizing first would leave every legacy sentence stuck with its raw account. Both passes live behind one entry point (`normalizeLegacySysMsgs`) so no call site can drift them apart, and a test pins the order. Position at each of the six `messages.go` call sites is unchanged — after `setDecodedAttachments`, before the budget trim; normalization only ever shortens the encoded row, so trimming after it stays conservative.

All resolution is at read time — nothing migrated — so a renamed bot or user shows its current name on the next history load.

## Read cost

New `preview.AppNameCache` serves the single-account and batch lookups over **one** bounded TTL cache (`AppNamesByAccounts` is a single `$in` on the indexed `apps.assistant.name`, mirroring `room-service`'s `findAppsForDisplay`). The batch peeks the cache per account and queries only the misses, coalescing concurrent cold sets through singleflight on a canonicalised miss-set key.

| page | apps reads |
|---|---|
| no bots | 0 |
| all bots warm | 0 |
| N bots, some cold | 1, covering only the cold ones |
| N concurrent pages, same cold bot | 1 total |

Because both halves share one cache, a bot resolved for a reaction actor is free to the legacy sys-msg page and the reverse. `CachedAppNameLookup` keeps its signature and behaviour on the same struct, so `broadcast-worker` and `room-worker` are untouched.

A failed batch returns the cache hits it already gathered rather than discarding them, and caches nothing from the failure so the next call retries. A `users` failure is logged and fallen through rather than returned — it says nothing about `apps`, so a bot on that page can still resolve, while non-bot rows end up untouched exactly as before.

## Scope

`pin.go` / `threads.go` stay out, per the #404 decision: a legacy member row can never be pinned or become a thread parent. No new dependencies.

**Not done here: classifying the legacy types as system messages.** This PR normalizes the type on the *wire*; it does not add `members_removed` / `members_left` to `pkg/model.systemMessageTypes`, so `IsSystemMessageType` still returns false for a *stored* legacy row. That leaves two real bugs, both on paths that read migrated Cassandra rows directly:

- `history-service/internal/service/rooms.go:255` — the lazy preview walk can select a legacy member row as a room-list preview.
- `data-migration/es-index-migrator/runner.go:63` — the backfill indexes legacy member rows as searchable content, though its own comment says it must match the live path's system-message drop.

Deliberately out of scope here because it is classification rather than rendering, and it touches `pkg/model` plus a migration tool. **Fixed separately in #458**, which is cut from `main` and independent of this branch.

*(An earlier revision of this description claimed the preview gap was sealed by `broadcast-worker` / `roomlist-worker` and could not be fixed in history-service. That was wrong on both counts — `broadcast-worker`'s insert-side predicate only ever sees newly-created messages, which carry modern types, and the vulnerable walk is in history-service. Corrected above.)*

Also not done here: bounding this best-effort enrichment with a child timeout. Fair idea, but the sibling `FindUsersByAccounts` on the same path has the same property, so it belongs in its own PR rather than being applied to one of the two reads.

## Testing

TDD throughout. Covers bot with app row, bot with a user doc (app name wins), bot with no app row, bot with neither, app lookup error, `users` error while a bot still resolves, nil user store, nil app store, `members_left` type-only rewrite, modern types untouched, and the ordering guard. Read-cost behaviour is asserted directly: one users batch + one apps batch per page, a second page served entirely from cache, a mixed page querying only the newly-cold bot, cross-path reuse between the reaction and sys-msg halves, and 8 concurrent cold batches collapsing to a single read (in any account order).

Plus a mongorepo integration test for the `$in` filter, the nested `assistant.name` projection, absent accounts, and an app with no assistant.


julianshen:
**6 findings · 5 unique issues · 5 inline comments posted** — 🟠 6 warning

### Related work

- [PR #404](https://github.com/hmchangw/newchat/pull/404) — feat\(history-service\): resolve legacy members\_removed display names (merged)
- [PR #413](https://github.com/hmchangw/newchat/pull/413) — fix\(history-service\): strip the quotes around the legacy members\_removed account (merged)
- [PR #446](https://github.com/hmchangw/newchat/pull/446) — fix\(room-worker\): render bot app name in member add/remove system messages (merged)
- [PR #458](https://github.com/hmchangw/newchat/pull/458) — fix\(model\): classify legacy migrated member types as system messages (open)

<details>
<summary>📒 Files reviewed (13)</summary>

* `docs/client-api.md`
* `docs/superpowers/plans/2026-09-01-history-service-legacy-sysmsg-bot-name-and-type.md`
* `history-service/internal/mongorepo/app.go`
* `history-service/internal/mongorepo/app_test.go`
* `history-service/internal/service/sysmsgname.go`
* `history-service/internal/service/sysmsgname_test.go`
* `pkg/preview/appname.go`
* `pkg/preview/appname_test.go`
* `pkg/preview/preview.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

history-service/internal/service/sysmsgname.go
```diff
@@ -49,17 +74,63 @@ func quoteRemoved(name string) string {
...
+	for _, account := range accounts {
+		if model.IsBot(account) {
+			bots = append(bots, account)
+		}
+	}
+	if len(bots) == 0 {
+		return nil
+	}
+	names, err := s.apps.AppNamesByAccounts(ctx, bots)
+	if err != nil {
```
julianshen, line 114: _🏷️ reliability_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`, `mongodb`, `performance-and-concurrency`, `tgd-review` (4 rules)_

**Bot-name lookups bypass the shared cache and singleflight.**

Every history request containing a legacy bot calls the raw `AppNamesByAccounts` store, while `AppNameByAccount` is explicitly documented as the cached path. Only duplicates within one page are deduplicated. If many clients or retries load the same legacy page, they issue one identical Mongo query each; when Mongo slows, those handlers wait on the same downstream and retries add more load, potentially consuming the request deadline after Cassandra history already succeeded so callers receive a timeout. The PR description asserts one read per TTL via the existing cache, but this implementation does not provide that. Route distinct bot accounts through the shared cached/singleflight lookup, or add an equivalent shared batch cache; optional enrichment should also have a shorter timeout than the overall history request.

<details>

- **`performance-and-concurrency`**:112: `appNamesFor` calls `s.apps.AppNamesByAccounts` directly on every qualifying read. Because all six history read paths invoke this normalization, concurrent clients loading a page containing the same legacy bot message each issue an identical Mongo query; there is no TTL reuse or singleflight coalescing, creating a cache stampede and additional database load. This also contradicts the PR description's assertion that the existing cached lookup is reused. Route these lookups through the service-wide cached/singleflight app-name lookup, or add equivalent bounded caching and request coalescing for the batch path.
- **`mongodb`**:112: `appNamesFor` calls `apps.AppNamesByAccounts` directly, so every history request containing a legacy removed-bot row issues a new indexed query against `apps`, even when the same bot was resolved moments earlier. This contradicts the stated one-read-per-TTL behavior and creates a concrete extra MongoDB read per qualifying page under repeated history loads. Route deduplicated bot accounts through the existing cached `s.appName`/bot-aware lookup, or add a shared bounded TTL cache for the batch lookup rather than querying `AppStore` directly on every request.
- **`tgd-review`**:111: `appNamesFor` calls `AppNamesByAccounts` directly whenever a response contains a bot, and that repository method executes `FindMany` on every invocation. Repeated history loads for the same bot therefore never use the existing cached `s.appName` lookup, contradicting the stated five-minute TTL/singleflight behavior and adding an apps query to every qualifying page. Route deduplicated bot accounts through the cached lookup, or cache the batch lookup, and test that consecutive normalizations reuse results.

</details>

<details>

```
In `history-service/internal/service/sysmsgname.go` around line 112: Every history request containing a legacy bot calls the raw `AppNamesByAccounts` store, while `AppNameByAccount` is explicitly documented as the cached path. Only duplicates within one page are deduplicated. If many clients or retries load the same legacy page, they issue one identical Mongo query each; when Mongo slows, those handlers wait on the same downstream and retries add more load, potentially consuming the request deadline after Cassandra history already succeeded so callers receive a timeout. The PR description asserts one read per TTL via the existing cache, but this implementation does not provide that. Route distinct bot accounts through the shared cached/singleflight lookup, or add an equivalent shared batch cache; optional enrichment should also have a shorter timeout than the overall history request.

```

</details>

---

history-service/internal/service/sysmsgname.go
```diff
@@ -49,17 +74,63 @@ func quoteRemoved(name string) string {
...
+	bots := make([]string, 0, len(accounts))
+	for _, account := range accounts {
+		if model.IsBot(account) {
+			bots = append(bots, account)
+		}
+	}
+	if len(bots) == 0 {
+		return nil
+	}
+	names, err := s.apps.AppNamesByAccounts(ctx, bots)
```
julianshen: _🏷️ nats-backpressure_ | _🟠 Warning_ | _⚡ Quick fix_ | _`nats`_

**Uncached app lookups add a Mongo read to every matching NATS request.**

Each history request containing a legacy bot-removal row reaches this direct `AppNamesByAccounts` query before its request handler can return a reply. The results are not cached across requests, so repeated loads of the same room and bot generate one Mongo query per NATS request; the tests only verify batching within a single invocation. Under frequent history loads or Mongo latency, this increases database pressure and delays replies enough to consume callers' request timeouts. Cache these results for the service lifetime with bounded TTL/singleflight semantics, or route the deduplicated bot accounts through the existing cached per-account app-name lookup.

<details>

```
In `history-service/internal/service/sysmsgname.go` around line 111: Each history request containing a legacy bot-removal row reaches this direct `AppNamesByAccounts` query before its request handler can return a reply. The results are not cached across requests, so repeated loads of the same room and bot generate one Mongo query per NATS request; the tests only verify batching within a single invocation. Under frequent history loads or Mongo latency, this increases database pressure and delays replies enough to consume callers' request timeouts. Cache these results for the service lifetime with bounded TTL/singleflight semantics, or route the deduplicated bot accounts through the existing cached per-account app-name lookup.

```

</details>

---

julianshen (commented):
### tGDBot review of `e1c7f90` — 5 findings in 10 files

5 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/451#issuecomment-5491273544), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

pkg/preview/appname.go
```diff
@@ -31,22 +31,95 @@ func CachedAppNameLookup(inner AppNameLookup) AppNameLookup {
...
+
+// Names resolves many bot accounts, reading ONLY the ones the cache cannot answer —
+// so a page of warm bots costs nothing and a page of cold ones costs one read.
+//
+// On a failed fetch it returns the cache hits it already gathered alongside the
+// error: a partial answer renders more names than none, and the caller degrades the
+// rest to their composed names either way. Nothing from a failure is cached, so the
+// next call retries those accounts.
+//
+// Deliberately not singleflighted: the key is a set rather than an account, and this
```
julianshen: _🏷️ reliability_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`, `nats`, `performance-and-concurrency`, `tgd-review` (4 rules)_

**Concurrent cold batch lookups still stampede Mongo.**

The shared TTL cache fixes sequential rereads, but the batch path is explicitly not singleflighted. After a cache expiry, concurrent history requests containing the same legacy bot can all observe the miss before any request populates the cache, so every request invokes `AppNamesByAccounts` with the same account. During an apps slowdown or partition, these optional reads amplify Mongo load and can consume each request's parent deadline after the Cassandra page has already loaded; caller retries then repeat the herd. Coalesce concurrent batch misses (at minimum by a canonicalized miss-set key, or with per-account in-flight reservations for overlapping batches) and bound this best-effort enrichment with a shorter child timeout so history can return composed/raw names before the overall deadline.

<details>

- **`performance-and-concurrency`**:85: The TTL cache fixes repeated sequential reads, but the batch path explicitly remains outside singleflight. After startup or a TTL expiry, N concurrent history requests containing the same legacy bot can all observe a miss before any request populates the cache, causing N identical `AppNamesByAccounts` Mongo queries. During an apps-store failure, misses are not cached, so every incoming request continues hitting the failing dependency. This preserves the concurrent-load portion of the prior concern and can amplify Mongo load and history latency. Coalesce batch misses using a separate singleflight group keyed by a canonicalized account set, with a cache recheck inside the leader, or coordinate in-flight ownership per account while retaining one batch query.
- **`nats`**:85: The shared TTL cache fixes sequential rereads, but this code explicitly leaves batch misses without singleflight. After startup or TTL expiry, concurrent NATS history handlers requesting the same legacy bot all observe a miss and independently call `c.many`; each handler then waits for Mongo before sending its otherwise-ready history reply. Slow reads and requester retries can therefore amplify Mongo load and response time. Coalesce identical cold batches with singleflight and recheck the cache inside the leader, or track per-account in-flight loads while preserving batched fetches. The prior cache-bypass concern is only partially addressed.
- **`tgd-review`**:85: The shared TTL cache now prevents repeated sequential reads, but concurrent pages containing the same cold bot can all observe a miss and call `c.many` independently because the batch path explicitly avoids singleflight. This leaves the concurrent retry/load-amplification path from the prior finding unresolved. Coalesce concurrent batch misses, including overlapping account sets, and add a concurrency test proving one downstream read for the same cold bot.

</details>

<details>

```
In `pkg/preview/appname.go` around line 85: The shared TTL cache fixes sequential rereads, but the batch path is explicitly not singleflighted. After a cache expiry, concurrent history requests containing the same legacy bot can all observe the miss before any request populates the cache, so every request invokes `AppNamesByAccounts` with the same account. During an apps slowdown or partition, these optional reads amplify Mongo load and can consume each request's parent deadline after the Cassandra page has already loaded; caller retries then repeat the herd. Coalesce concurrent batch misses (at minimum by a canonicalized miss-set key, or with per-account in-flight reservations for overlapping batches) and bound this best-effort enrichment with a shorter child timeout so history can return composed/raw names before the overall deadline.

```

</details>

---

pkg/preview/appname.go
```diff
@@ -31,22 +31,95 @@ func CachedAppNameLookup(inner AppNameLookup) AppNameLookup {
...
+		if name != "" {
+			names[account] = name
+		}
+	}
+	if len(misses) == 0 {
+		return names, nil
+	}
+
+	fetched, err := c.many(ctx, misses)
+	if err != nil {
```
julianshen, line 118: _🏷️ mongodb_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`mongodb`_

**Concurrent cold pages still stampede the apps collection.**

The new LRU prevents sequential repeat reads, but `Names` deliberately lacks singleflight. When concurrent history requests contain the same uncached bot, each reaches `c.many`, causing a separate `apps` `$in` query before any request warms the cache. This preserves the operational path from the prior finding: retries or simultaneous page loads amplify MongoDB traffic and latency, contrary to the asserted one-read-per-TTL behavior. Coalesce batch misses—ideally by canonicalized account set or per-account in-flight state—and recheck the cache after acquiring leadership before querying MongoDB.

<details>

```
In `pkg/preview/appname.go` around line 110: The new LRU prevents sequential repeat reads, but `Names` deliberately lacks singleflight. When concurrent history requests contain the same uncached bot, each reaches `c.many`, causing a separate `apps` `$in` query before any request warms the cache. This preserves the operational path from the prior finding: retries or simultaneous page loads amplify MongoDB traffic and latency, contrary to the asserted one-read-per-TTL behavior. Coalesce batch misses—ideally by canonicalized account set or per-account in-flight state—and recheck the cache after acquiring leadership before querying MongoDB.

```

</details>

---

history-service/internal/service/sysmsgname.go
```diff
@@ -49,17 +74,64 @@ func quoteRemoved(name string) string {
...
 }
 
+// removedMemberName renders one account's display name, preferring a bot's
+// registered app name over the composed one — the precedence room-worker applies
+// on the live path, so a legacy row and a modern one show a bot alike.
+//
+// An empty return means "nothing resolved": leave the row exactly as stored.
+func removedMemberName(account string, u *model.User, appNames map[string]string) string {
+	if name, ok := appNames[account]; ok && name != "" {
+		return name
```
julianshen, line 84: _🏷️ dead-code_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`dead-code`_

**The map-presence check is logically redundant.**

A missing `map[string]string` key already yields the empty string, so `name != ""` can only be true when the key exists. The `ok` value can never affect this condition and is vestigial; use the value-only lookup.

<details>

> ‼️ **IMPORTANT**

```suggestion
	if name := appNames[account]; name != "" {
```

</details>

<details>

```
In `history-service/internal/service/sysmsgname.go` around line 84: A missing `map[string]string` key already yields the empty string, so `name != ""` can only be true when the key exists. The `ok` value can never affect this condition and is vestigial; use the value-only lookup.

```

</details>

---

julianshen (commented):
### tGDBot review of `c93a666` — 6 findings in 13 files

5 🟠 warning · 1 🔵 suggestion
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/451#issuecomment-5491273544), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

pkg/preview/appname.go
evelynchang5525: Valid — fixed in c217d84. My "deliberately not singleflighted" note was the wrong call: I reasoned that page reads don't need the coalescing the per-message fan-out does, but a TTL expiry lets every in-flight request miss at once regardless of how the path is described, and history-service serves enough traffic for that herd to be real.

`Names` now coalesces on a canonicalised miss-set key (sorted, NUL-joined), so the same accounts arriving in any order share one leader rather than opening two. The leader is `loadNames`, which rechecks the cache before reading — the same recheck `loadAppName` already does on the single path — so a leader arriving after an in-flight load completed and released the key serves the warm answer instead of re-querying. A partly-warm set reads only its cold half.

The group is separate from the per-account one (`sfMany` vs `sf`): a single-account batch would produce the same key as the single-account path while carrying a `map[string]string` instead of a `string`, so sharing one group risks a bad type assertion.

Three tests cover it: 8 concurrent cold batches collapse to one read; the same accounts in mixed order still collapse to one; and `loadNames` re-entered on a warm set reads nothing while a partly-warm set asks only for the cold account.

Not doing the child-timeout half of the suggestion. It's a fair point that best-effort enrichment shouldn't be able to spend the whole request budget, but the sibling `FindUsersByAccounts` on this same path has exactly the same property, so bounding only the apps read would be inconsistent without fixing the real issue. That's a broader change to how history-service budgets optional enrichment, and it belongs in its own PR rather than widening this one.

---

pkg/preview/appname.go
evelynchang5525: Same finding as the thread on line 85 — fixed in c217d84. `Names` now singleflights on a canonicalised miss-set key, and the leader (`loadNames`) rechecks the cache after acquiring leadership before touching Mongo, exactly as suggested.

---

history-service/internal/service/sysmsgname.go
evelynchang5525: Correct — applied in c217d84. A missing key yields the zero value, so `name != ""` already implies presence and the `ok` could never change the outcome. Now `if name := appNames[account]; name != ""`.

---

history-service/internal/service/sysmsgname.go
evelynchang5525: This was against `e1c7f90` and is already addressed by `c93a666`, which landed after the review ran. `appNamesFor` no longer calls the store directly — it goes through `s.appNames`, one half of a shared `preview.AppNameCache` built once in `New`, so the batch and the per-account reaction lookup sit over the same bounded TTL cache and a name either one resolves is free to the other.

You were right that the PR description contradicted the code at that commit; I've since rewritten it to match.

The concurrency half of your point survived into `c93a666` and you re-reported it on the new head — that one is now fixed too, in c217d84: see the thread on `pkg/preview/appname.go`.

---

history-service/internal/service/sysmsgname.go
evelynchang5525: Addressed by `c93a666` (shared TTL cache across requests) and `c217d84` (singleflight on the batch), both landed after this review ran. Repeated loads of the same room and bot now cost zero apps reads while the entry is warm, and concurrent cold ones collapse to a single read.

---

pkg/preview/appname.go
```diff
@@ -31,22 +33,145 @@ func CachedAppNameLookup(inner AppNameLookup) AppNameLookup {
...
-			return "", err
+		if name != "" {
+			names[account] = name
+		}
+	}
+	if len(misses) == 0 {
+		return names, nil
+	}
+
+	v, err, _ := c.sfMany.Do(missSetKey(misses), func() (any, error) {
```
julianshen, line 115: _🏷️ distributed-systems_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`, `tgd-review` (2 rules)_

**Exact-set singleflight does not coalesce overlapping bot lookups**

After TTL expiry, concurrent pages requesting `[a.bot, b.bot]` and `[a.bot, c.bot]` compute different singleflight keys, so both issue Mongo queries containing `a.bot`; concurrent `Name` calls also use a separate group and never join the batch. A popular bot can therefore be queried once per concurrent page during downstream slowness, rather than once total. These independent reads can also race with a rename: a newer result may be cached first, then a delayed older batch unconditionally overwrites it in `loadNames`, serving the stale name for the TTL. Coordinate in-flight work per account or merge overlapping batches, and fence cache writes so older loads cannot replace newer results.

<details>

- **`tgd-review`**:115: The key covers the entire miss set, so concurrent pages missing `[a.bot, b.bot]` and `[a.bot, c.bot]` use different keys and both query `a.bot`. A concurrent `Name(a.bot)` also uses the separate `sf` group. This contradicts the claimed one-read behavior for concurrent pages sharing a cold bot and can amplify Mongo reads under load. Coordinate in-flight work per account or through a shared batcher, and add tests for overlapping sets and concurrent single/batch lookups; the current tests cover only identical sets and sequential cross-path reuse.

</details>

<details>

```
In `pkg/preview/appname.go` around line 115: After TTL expiry, concurrent pages requesting `[a.bot, b.bot]` and `[a.bot, c.bot]` compute different singleflight keys, so both issue Mongo queries containing `a.bot`; concurrent `Name` calls also use a separate group and never join the batch. A popular bot can therefore be queried once per concurrent page during downstream slowness, rather than once total. These independent reads can also race with a rename: a newer result may be cached first, then a delayed older batch unconditionally overwrites it in `loadNames`, serving the stale name for the TTL. Coordinate in-flight work per account or merge overlapping batches, and fence cache writes so older loads cannot replace newer results.

```

</details>

---

julianshen, line 115: _🏷️ failure-handling_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`_

**Shared app-name loads inherit one caller's cancellation**

`singleflight.Do` runs the shared Mongo fetch with the leader's request context and gives waiters no cancellation path. If a nearly expired request becomes leader, its deadline aborts the fetch and healthy followers receive the same error and render fallback names. Conversely, a follower whose request is canceled remains blocked until the leader's database call finishes, retaining handler resources during an outage. Use `DoChan` with a per-caller context select, and execute shared work under an independently bounded context rather than an arbitrary leader's request context.

<details>

```
In `pkg/preview/appname.go` around line 115: `singleflight.Do` runs the shared Mongo fetch with the leader's request context and gives waiters no cancellation path. If a nearly expired request becomes leader, its deadline aborts the fetch and healthy followers receive the same error and render fallback names. Conversely, a follower whose request is canceled remains blocked until the leader's database call finishes, retaining handler resources during an outage. Use `DoChan` with a per-caller context select, and execute shared work under an independently bounded context rather than an arbitrary leader's request context.

```

</details>

---

pkg/preview/appname.go
```diff
@@ -31,22 +33,145 @@ func CachedAppNameLookup(inner AppNameLookup) AppNameLookup {
...
+		return c.loadNames(ctx, misses)
+	})
```
julianshen, line 117: _🏷️ mongodb-lifecycle_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`mongodb`_

**A canceled singleflight leader aborts lookups for healthy requests.**

The shared `apps.assistant.name` MongoDB query uses the context of whichever request becomes the `sfMany` leader. If request A starts a cold lookup, request B joins the same miss-set, and A is canceled, MongoDB aborts the query and B receives the same error despite its valid context, returning a raw/composed bot name. Conversely, canceling waiting request B cannot unblock it because `singleflight.Do` does not observe B's context. Use a bounded cache-owned context for the shared Mongo operation and `DoChan` plus a select on each caller's context so one request cannot cancel or retain another.

<details>

```
In `pkg/preview/appname.go` around line 117: The shared `apps.assistant.name` MongoDB query uses the context of whichever request becomes the `sfMany` leader. If request A starts a cold lookup, request B joins the same miss-set, and A is canceled, MongoDB aborts the query and B receives the same error despite its valid context, returning a raw/composed bot name. Conversely, canceling waiting request B cannot unblock it because `singleflight.Do` does not observe B's context. Use a bounded cache-owned context for the shared Mongo operation and `DoChan` plus a select on each caller's context so one request cannot cancel or retain another.

```

</details>

---

pkg/preview/appname.go
```diff
@@ -31,22 +33,145 @@ func CachedAppNameLookup(inner AppNameLookup) AppNameLookup {
...
+		if name != "" {
+			names[account] = name
+		}
+	}
+	if len(misses) == 0 {
+		return names, nil
+	}
+
+	v, err, _ := c.sfMany.Do(missSetKey(misses), func() (any, error) {
+		return c.loadNames(ctx, misses)
```
julianshen, line 116: _⚡ performance_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`performance-and-concurrency`_

**Overlapping cold batches bypass stampede protection**

The singleflight key covers the entire miss set, so it only coalesces identical batches. After TTL expiry, concurrent pages missing [popular.bot, a.bot] and [popular.bot, b.bot] use different keys; both recheck before either query completes and both read popular.bot. A burst of heterogeneous pages can therefore issue one Mongo query per request for the same popular account, amplifying load precisely during a cold-cache or slow-database event. Coordinate in-flight work per account while retaining batch fetches—for example, reserve unresolved accounts, batch only newly reserved misses, and make callers with overlapping reservations wait and recheck the cache. Add a concurrency test with overlapping, non-identical sets.

<details>

```
In `pkg/preview/appname.go` around line 116: The singleflight key covers the entire miss set, so it only coalesces identical batches. After TTL expiry, concurrent pages missing [popular.bot, a.bot] and [popular.bot, b.bot] use different keys; both recheck before either query completes and both read popular.bot. A burst of heterogeneous pages can therefore issue one Mongo query per request for the same popular account, amplifying load precisely during a cold-cache or slow-database event. Coordinate in-flight work per account while retaining batch fetches—for example, reserve unresolved accounts, batch only newly reserved misses, and make callers with overlapping reservations wait and recheck the cache. Add a concurrency test with overlapping, non-identical sets.

```

</details>

---

julianshen, line 116: _🎯 concurrency_ | _🟠 Warning_ | _⚡ Quick fix_ | _`performance-and-concurrency`_

**Singleflight followers ignore their own cancellation**

singleflight.Group.Do blocks every follower until the leader's database call finishes and provides no way to observe the follower's context. If one app lookup is slow, later history requests for the same miss set remain blocked even after their clients cancel or their deadlines expire. Under a cancellation burst this retains handler goroutines and request state for the leader's full query duration. Use DoChan and select on ctx.Done() so each follower can return promptly; keep the shared load bounded independently if it must outlive an individual caller, and test that a canceled follower exits while the leader remains blocked.

<details>

```
In `pkg/preview/appname.go` around line 116: singleflight.Group.Do blocks every follower until the leader's database call finishes and provides no way to observe the follower's context. If one app lookup is slow, later history requests for the same miss set remain blocked even after their clients cancel or their deadlines expire. Under a cancellation burst this retains handler goroutines and request state for the leader's full query duration. Use DoChan and select on ctx.Done() so each follower can return promptly; keep the shared load bounded independently if it must outlive an individual caller, and test that a canceled follower exits while the leader remains blocked.

```

</details>

---

julianshen (commented):
### tGDBot review of `c217d84` — 6 findings in 13 files

6 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/451#issuecomment-5491273544), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

pkg/preview/appname.go
evelynchang5525: Factually correct, and I'm not fixing it in this PR. Reasoning, since this is the third round on this subsystem.

**The overlap claim is right.** The key is the whole miss set, so `[a.bot, b.bot]` and `[a.bot, c.bot]` are different keys and both read `a.bot`. What the current code buys is coalescing of *identical* cold sets — strictly better than the no-coalescing state you flagged last round, but short of per-account optimality.

**The rename race is also right, and lands inside staleness this package already accepts.** Two overlapping batches can have the slower-but-older one `cache.Add` over the newer result, serving a stale name for up to the TTL. But `appname.go`'s own header sets that tolerance explicitly: *"An app's display name changes about as often as the app is renamed, so the TTL is generous: a few minutes of a stale name in a room-list row is not a correctness concern."* A rename losing a race is bounded by the same 5 minutes an un-raced rename already waits. It is not a new class of staleness.

**Why not fix it here.** The suggested remedy — per-account reservations, or merging overlapping batches, plus fencing cache writes by load order — is a genuine concurrency primitive with its own deadlock and goroutine-leak surface, in a package `broadcast-worker`, `room-worker` and `history-service` all depend on. Building that inside a PR about rendering names in legacy system messages is exactly the widening I should not do on my own, and the window it closes is bounded by a 5-minute TTL after which every caller is warm regardless.

Worth doing as its own change, where it can be designed and tested as a primitive rather than as a side effect. Happy to open that issue — say the word and I will.

---

evelynchang5525: The mechanics are right, but this is a **pre-existing, documented and deliberate** property of this package rather than something this PR introduces.

`CachedAppNameLookup` — untouched by this PR, and the path `broadcast-worker` and `room-worker` have used since #366 — carries exactly this contract in its own doc comment:

> Errors are not cached, and the load runs on the caller's context so it stays inside the seal's budget. **A cancelled leader fails its followers too; they are all under that same budget**, and the caller degrades to the composed display name for the message either way.

So the repo already weighed leader-cancellation-poisons-followers and accepted it, for a stated reason that holds here too: this is best-effort enrichment. A follower inheriting the leader's cancellation renders one row with a composed name instead of an app name. That is the designed degradation, not a failure.

My batch half inherits those semantics deliberately, so both halves of one cache behave alike. Switching only `Names` to `DoChan` + per-caller select + a cache-owned bounded context would make the two halves of the same struct disagree about cancellation; switching both changes runtime behaviour for `broadcast-worker` and `room-worker`, which is well outside a PR about legacy system-message names.

The follower-blocked-past-its-own-cancellation half (your other thread) is the more substantive of the two and I'm not dismissing it — it is also identical to what the single-account path has always done, and bounded by one Mongo query. Both belong in a change that revisits this primitive deliberately, with `CachedAppNameLookup`'s contract updated in the same breath. Happy to open that issue.

---

pkg/preview/appname.go
evelynchang5525: Same finding as the `failure-handling` thread on line 115 — see the full reply there.

Short version: leader-cancellation-poisons-followers is the documented, pre-existing contract of `CachedAppNameLookup` ("A cancelled leader fails its followers too; they are all under that same budget"), which this PR does not touch and which `broadcast-worker` and `room-worker` already run on. The batch half inherits it deliberately so both halves of one cache agree. Revisiting it means changing that shared contract, which belongs in its own change rather than here.

---

pkg/preview/appname.go
evelynchang5525: Same finding as the `distributed-systems` thread on line 115 — see the full reply there.

Short version: correct that the miss-set key only coalesces identical batches, so a popular bot in heterogeneous cold pages can still be read once per page. The remedy you describe (per-account reservations feeding a batch of newly-reserved misses) is a concurrency primitive in a package three services depend on, and the window it closes is bounded by the 5-minute TTL. Worth its own change, not a side effect of a PR about legacy system-message names.

---

evelynchang5525: This is the sharpest of the five, and I'm still not fixing it here — see the full reply on the `failure-handling` thread (line 115).

You're right that `singleflight.Do` gives a follower no way to observe its own context, so a cancelled follower holds its goroutine until the leader's query returns. Two things bound it: the wait is one Mongo query against an indexed `$in`, and it is character-for-character what the single-account path has done since #366 — `CachedAppNameLookup` uses `sf.Do` the same way, and `broadcast-worker`'s per-message fan-out sits on it, which is a far hotter path than history pages.

So this is a property of the primitive, not of the batch half. Fixing it properly means moving both halves to `DoChan` with a per-caller select and a cache-owned bounded context, and rewriting `CachedAppNameLookup`'s stated contract — a change to shared behaviour for three services. It should be reviewed as that, not smuggled in under a PR about legacy system-message names.

---

chenjr0719:
Reviewing this one — could you hold the merge briefly? It touches the member / system-message wire shape (bot app-name resolution + legacy sys-msg type normalization). Clients switch on those sys-msg type values and read those name fields, so I want to confirm old values still resolve and no existing field changes shape. Will follow up shortly.

evelynchang5525:
@chenjr0719 Holding — nothing will be merged from my side (I don't merge or approve). Both of your concerns are checkable against the diff, so here's the evidence up front to save you the dig.

## 1. No existing field changes shape

**This PR touches no `pkg/model` file at all.** Full file list against the merge base (`dcf77b5`):

```
docs/client-api.md
docs/superpowers/plans/2026-09-01-...md
history-service/internal/mongorepo/app.go              (+ its test)
history-service/internal/service/messages.go           (+ its test)
history-service/internal/service/mocks/mock_repository.go
history-service/internal/service/service.go
history-service/internal/service/sysmsgname.go         (+ its test)
pkg/preview/appname.go                                 (+ its test)
pkg/preview/preview.go
```

No wire struct is added to, removed from, or retyped. The only two fields whose **values** change are `msg` and `type`, both still plain strings:

- `msg` — the quoted account inside a legacy `members_removed` sentence is replaced by a display name. Same field, same type, same sentence shape (quotes and trailing period preserved).
- `type` — normalized, see below.

`pkg/preview/preview.go` gains only a new `AppNamesLookup` func type; no serialized struct.

## 2. Old values still resolve — and the direction matters

The normalization runs **legacy → modern**, never the reverse:

| stored in Cassandra | returned to clients |
|---|---|
| `members_removed` | `member_removed` |
| `members_left` | `member_left` |

So a client switching on `member_removed` / `member_left` keeps matching everything it matched before, **and now additionally catches the migrated rows it previously couldn't classify**.

The compatibility risk would be a client switching on the *legacy plural* strings. There isn't one. `chat-frontend/src/lib/messageType.js` is the authoritative client-side set, and it holds exactly the seven modern types:

```js
export const SYSTEM_MESSAGE_TYPES = new Set([
  'room_created', 'members_added', 'member_removed', 'member_left',
  'room_renamed', 'room_restricted', 'teams_meet_started',
])
```

`members_removed` / `members_left` appear nowhere in it, and a repo-wide grep finds them in no `.js/.jsx/.ts/.tsx/.vue` file at all. (Note `members_added` **is** current and plural — the plural shape is not the legacy tell.)

## The one behaviour change worth your eye

This is the intended fix rather than a regression, but since you're checking client switching it's the thing to sanity-check against your own client:

**Before**, a legacy row arrived with `type: "members_removed"`, failed `isSystemMessageType()`, and rendered as an ordinary user message — a normal bubble with a sender. **After**, it arrives as `member_removed` and renders as centered system chrome, like every other membership notice.

That is the point of the change, but if any client you own does something other than "unknown type ⇒ render as normal message", that's where it would show.

Also relevant: this PR normalizes the type **on the wire only** — nothing is written back to Cassandra, and `IsSystemMessageType` still returns false for a *stored* legacy row. #458 is the sibling that fixes the server-side classification (room-list preview selection and the search backfill); it's independent and cut from `main`.

Happy to answer anything else, or to adjust if your client does read the legacy values somewhere I can't see.

---

chenjr0719:
Thanks — that's the confirmation I needed, and it matches what I found. Bot-name resolution rewrites the msg text only (no field reshaped), and the type normalization maps the legacy plurals to the existing canonical `member_removed`/`member_left` at read time, so anything switching on the modern types keeps matching and additionally picks up the migrated rows. Confirmed on our side too — nothing reads the legacy plural strings.

No wire-shape blocker from me — please don't let my earlier note hold you up. This isn't my full review yet (a couple of non-blocking quality observations may follow), but none of it should gate the merge.

mliu33 (approved):
Excellent work, thanks

PR #454, opened 2026-09-01
fix(broadcast-worker): stop retrying a shedding downstream at 200ms

hmchangw:
## The bug

`broadcast-worker` settles every failure on `LowLatencyBackoff`:

```go
// broadcast-worker/main.go:539 (before)
jsretry.Settle(handlerCtx, msg, jsretry.LowLatencyBackoff, handler.HandleMessage(handlerCtx, msg.Data()))
```

The 200ms first rung is deliberate and correct for fan-out — a sub-second blip must not be visible to a client waiting on delivery. It is the wrong curve for one case: **a downstream that is shedding rather than failing.**

The thread-parent path issues a synchronous request to `history-service`, which replies `Unavailable` ("service busy") once its admission cap saturates. Retrying that in 200ms aims more load at the service that is already failing. And `MaxDeliver` is sized by `WithOutageRetryBudget` to cover a one-hour window — against `LowLatencyBackoff`'s jittered minimums that is **~18 deliveries, the first five inside 78 seconds.**

So one degraded `history-service` turns every thread reply into up to 18 requests against it. **Offered load rises as capacity falls**, and the brownout does not self-limit.

### Why it compounds

There is a two-hop correlated path. When `message-gatekeeper` soft-fails parent resolution it publishes the canonical event *without* `ThreadParentMessageCreatedAt`/`ThreadParentSenderAccount` — which is precisely the condition that makes this fetch fire. So a `history-service` failure produces *more* `history-service` requests, from a second service.

The existing escape hatch does not catch it: `parentResolveExhausted` is gated on `errcode.Terminal(err)`, and `Terminal` returns `false` for `CodeUnavailable`. The Ack-drop added for exactly this ack-pending reason never fires on a shedding downstream.

## The fix

Pick the schedule from the error, mirroring `search-sync-worker.itemBackoff` for ES 429s:

```go
err := handler.HandleMessage(handlerCtx, msg.Data())
jsretry.Settle(handlerCtx, msg, settleBackoff(err), err)
```

`jsretry.BackpressureBackoff` already exists for this and its own doc says why:

> a backend that just rejected the write needs time to drain, and a one-second retry only feeds the overload that caused the rejection

Ordinary transient failures keep the 200ms rung, so the fan-out latency property that motivated `LowLatencyBackoff` is unchanged.

## The delivery budget stays correct

Worth stating because `pkg/stream/consumer.go` documents a past incident of exactly this shape — a budget sized against one schedule applied to a consumer settling with another, giving ~4 minutes instead of the intended hour.

`MaxDeliver` is derived once from `LowLatencyBackoff` — the *faster* schedule, so the *larger* count — and both schedules share a 10m repeating tail. A message settling on the slower curve therefore still has enough deliveries to cover the outage window; it simply uses fewer of them per unit time.

`TestSettleBackoff_TailMatchesDeliveryBudget` pins that coupling, so a future change to either tail fails loudly rather than silently shortening the real retry window.

## Tests

- `TestSettleBackoff` — table-driven over shedding / rate-limited / wrapped / ordinary / internal / not-found / nil.
- `TestSettleBackoff_SheddingDoesNotRetryFast` — the regression guard: asserts a shed error's first rung is ≥30s **and** that an ordinary failure still retries in under a second, so a future "simplification" to one schedule fails either way.
- `TestSettleBackoff_TailMatchesDeliveryBudget` — the budget coupling above.

## Verification

- `go test ./broadcast-worker/ -race` — pass
- `make lint` — `0 issues`
- No store interface, client-facing handler, or `pkg/model` wire struct touched

## Deliberately not in this PR

The audit found the other half of this interaction: **`history-service` defaults to `REQUEST_TIMEOUT=10s` while all three of its hot-path callers request with a hardcoded 2s** (`broadcast-worker/parent_fetcher.go:19`, `message-gatekeeper/fetcher_history.go:19`, `room-service/reader_history.go:18`). NATS request/reply carries no cancellation, so the responder keeps an admission slot and a DB connection for 8 seconds past the caller's departure — effective capacity `256/2s → 256/10s ≈ 25 req/s`.

Fixing that properly means propagating the caller's deadline over the request so the responder can bound itself to `min(configured, caller)`. That touches `pkg/natsrouter`, `pkg/natsutil` and every RPC in the fleet — a design change that deserves its own review, not a rider on a hot-path bug fix. Lowering `history-service`'s guard to 2s instead would truncate `user-service`, which legitimately calls it with a 5s budget.

**This PR breaks the amplification loop, which is the half that turns a brownout into an outage.** The wasted-capacity half remains and is worth doing next.

---

### DIFF
--- a/broadcast-worker/backoff.go
+++ b/broadcast-worker/backoff.go
@@ -1,0 +1,48 @@
+package main
+
+import (
+	"errors"
+	"time"
+
+	"github.com/hmchangw/chat/pkg/errcode"
+	"github.com/hmchangw/chat/pkg/jsretry"
+)
+
+// settleBackoff picks the retry schedule a failed message calls for.
+//
+// broadcast-worker normally settles on LowLatencyBackoff, whose 200ms first
+// rung is the point: a sub-second blip must not be visible to a client waiting
+// on fan-out. That is the wrong curve for one case — a downstream that is
+// *shedding*.
+//
+// The thread-parent path issues a synchronous request to history-service, which
+// replies Unavailable ("service busy") once its admission cap is saturated.
+// Retrying that in 200ms aims more load at the service that is already failing,
+// so offered load rises as capacity falls. BackpressureBackoff exists for
+// exactly this — its doc notes "a one-second retry only feeds the overload that
+// caused the rejection" — and search-sync-worker already routes ES 429s to it.
+//
+// The delivery budget stays correct across both: MaxDeliver is derived once
+// from LowLatencyBackoff (the faster schedule, so the larger count), and both
+// schedules share a 10m repeating tail, so a message settling on the slower
+// curve still has enough deliveries to cover the outage window. A future change
+// that breaks that shared tail must re-derive the budget from the slower
+// schedule — TestSettleBackoff_TailMatchesDeliveryBudget guards it.
+func settleBackoff(err error) []time.Duration {
+	if isDownstreamShedding(err) {
+		return jsretry.BackpressureBackoff
+	}
+	return jsretry.LowLatencyBackoff
+}
+
+// isDownstreamShedding reports whether err is a dependency telling us it is over
+// capacity, as opposed to failing. historyParentFetcher propagates the typed
+// remote error precisely so it can be classified here, and natsutil.RequestFailure
+// maps a request timeout or a missing responder onto Unavailable too.
+func isDownstreamShedding(err error) bool {
+	var ee *errcode.Error
+	if !errors.As(err, &ee) {
+		return false
+	}
+	return ee.Code == errcode.CodeUnavailable || ee.Code == errcode.CodeTooManyRequests
+}
--- a/broadcast-worker/backoff_test.go
+++ b/broadcast-worker/backoff_test.go
@@ -1,0 +1,96 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/hmchangw/chat/pkg/errcode"
+	"github.com/hmchangw/chat/pkg/jsretry"
+)
+
+func TestSettleBackoff(t *testing.T) {
+	tests := []struct {
+		name string
+		err  error
+		want []time.Duration
+	}{
+		{
+			name: "a shedding downstream gets the slow curve",
+			err:  errcode.Unavailable("service busy"),
+			want: jsretry.BackpressureBackoff,
+		},
+		{
+			name: "an explicitly rate-limited downstream gets the slow curve",
+			err:  errcode.TooManyRequests("slow down"),
+			want: jsretry.BackpressureBackoff,
+		},
+		{
+			name: "the shedding signal survives wrapping",
+			err:  fmt.Errorf("fetch thread parent abc: %w", errcode.Unavailable("service busy")),
+			want: jsretry.BackpressureBackoff,
+		},
+		{
+			name: "an ordinary infra failure keeps the fan-out curve",
+			err:  errors.New("mongo: connection reset"),
+			want: jsretry.LowLatencyBackoff,
+		},
+		{
+			name: "an internal error is not backpressure",
+			err:  errcode.Internal("boom"),
+			want: jsretry.LowLatencyBackoff,
+		},
+		{
+			name: "a not-found is not backpressure",
+			err:  errcode.NotFound("no such message"),
+			want: jsretry.LowLatencyBackoff,
+		},
+		{
+			name: "success keeps the default curve",
+			err:  nil,
+			want: jsretry.LowLatencyBackoff,
+		},
+	}
+	for _, tt := range tests {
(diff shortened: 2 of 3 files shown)


julianshen:
**4 findings · 1 unique issue · 1 inline comment posted** — 🟠 4 warning

<details>

* `broadcast-worker/backoff.go`
* `broadcast-worker/backoff_test.go`
* `broadcast-worker/main.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

broadcast-worker/backoff.go
```diff
@@ -0,0 +1,48 @@
...
+// isDownstreamShedding reports whether err is a dependency telling us it is over
+// capacity, as opposed to failing. historyParentFetcher propagates the typed
+// remote error precisely so it can be classified here, and natsutil.RequestFailure
+// maps a request timeout or a missing responder onto Unavailable too.
+func isDownstreamShedding(err error) bool {
+	var ee *errcode.Error
+	if !errors.As(err, &ee) {
+		return false
+	}
+	return ee.Code == errcode.CodeUnavailable || ee.Code == errcode.CodeTooManyRequests
```
julianshen, line 47: _🏷️ retry behavior_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`, `nats`, `performance-and-concurrency`, `tgd-review` (4 rules)_

**Transport outages are misclassified as downstream shedding**

The function's own contract states that request timeouts and missing responders are mapped to CodeUnavailable, but this condition sends every CodeUnavailable to the ≥30s backpressure schedule. Failure timeline: history-service briefly loses connectivity or has no responder, natsutil returns Unavailable, the worker selects BackpressureBackoff, and the service recovers immediately; the message still cannot be retried for at least 30 seconds instead of LowLatencyBackoff's sub-second first rung. This makes an ordinary transient network/service-discovery failure user-visible despite the change intending to slow only explicit load shedding. Preserve a distinct overload signal across the boundary—for example, make admission rejection return CodeTooManyRequests and reserve CodeUnavailable for transport/availability failures—then apply BackpressureBackoff only to that overload signal.

<details>

- **`nats`**:44: The classifier sends every `CodeUnavailable` error to the slow backpressure schedule, although the added comment states that NATS request timeouts and missing responders are also mapped to `Unavailable`. A transient no-responder or timed-out history request therefore causes the JetStream message's first retry to be delayed by at least 30 seconds rather than the prior 200ms, making an ordinary connection/responder transition visible as delayed fan-out. Preserve distinct provenance for an explicit remote overload response (for example, a dedicated code or wrapped marker) and select `BackpressureBackoff` only for that marker or `CodeTooManyRequests`; transport timeout/no-responder failures should retain `LowLatencyBackoff`. Add tests for both NATS failure mappings.
- **`performance-and-concurrency`**:47: This routes every CodeUnavailable error to the 30-second backpressure schedule, although the preceding comment states that NATS request timeouts and missing responders are also mapped to Unavailable. During a brief deployment gap, lost response, or transient timeout, affected broadcasts therefore wait at least 30 seconds despite the dependency potentially recovering immediately, materially increasing fan-out latency. Preserve a distinct admission-rejection signal (for example, TooManyRequests or a typed shedding cause) and use the slow schedule only for that signal; ordinary Unavailable transport failures should retain LowLatencyBackoff.
- **`tgd-review`**:47: This treats every `CodeUnavailable` as backpressure, although the preceding comment states that `natsutil.RequestFailure` also maps request timeouts and missing responders to `Unavailable`. Those reachable transport failures will now use the ≥30s first retry instead of the intended 200ms transient-failure retry, delaying fan-out even when no downstream explicitly shed load. Preserve a distinct overload/rate-limit signal at the history request boundary and select `BackpressureBackoff` only for that signal; add timeout and no-responder tests to ensure they retain `LowLatencyBackoff`.

</details>

<details>

```
In `broadcast-worker/backoff.go` around line 47: The function's own contract states that request timeouts and missing responders are mapped to CodeUnavailable, but this condition sends every CodeUnavailable to the ≥30s backpressure schedule. Failure timeline: history-service briefly loses connectivity or has no responder, natsutil returns Unavailable, the worker selects BackpressureBackoff, and the service recovers immediately; the message still cannot be retried for at least 30 seconds instead of LowLatencyBackoff's sub-second first rung. This makes an ordinary transient network/service-discovery failure user-visible despite the change intending to slow only explicit load shedding. Preserve a distinct overload signal across the boundary—for example, make admission rejection return CodeTooManyRequests and reserve CodeUnavailable for transport/availability failures—then apply BackpressureBackoff only to that overload signal.

```

</details>

---

julianshen (commented):
### tGDBot review of `f937ce8` — 4 findings in 3 files

4 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/454#issuecomment-5491375028), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

PR #423, opened 2026-08-29
Mark marshal failures as permanent to prevent futile retries

hmchangw:
## Summary

Introduces `errcode.MarshalFailed()` to mark serialization failures as permanent, non-retryable errors. When a value cannot be marshaled (e.g., NaN float, erroring custom marshaler), it fails identically on every redelivery, so JetStream workers should Ack-drop it instead of exhausting their `MaxDeliver` budget on doomed retries.

## Key Changes

- **New API**: `errcode.MarshalFailed(what string, cause error)` wraps a marshal error as permanent with code `CodeInternal`. The cause is logged server-side by `Classify` but never reaches the client; the message names only the value being marshaled (e.g., "marshal message event").

- **Message Gatekeeper**: 
  - Detects permanent errors via `errcode.IsPermanent()` before checking for typed `*errcode.Error`
  - Records permanent faults as `resultFailed/reasonInternal` (not `resultRejected`) so the rejection series stays a pure client-error signal
  - Ack-drops permanent faults instead of Nak-ing them for redelivery
  - Replaces bare `fmt.Errorf` marshal errors with `MarshalFailed` calls

- **Broadcast Worker**: Replaces all `fmt.Errorf("marshal …: %w", err)` patterns with `errcode.MarshalFailed()` calls across thread events, mutations, encryption, and channel events

- **Room Worker**: Replaces marshal error wrapping in federation, room creation, and rename handlers with `MarshalFailed` calls

- **Message Worker**: Replaces marshal error wrapping in thread unread, subscription, and reply event handlers with `MarshalFailed` calls

- **Notification Worker**: Replaces marshal error wrapping in push batch emission with `MarshalFailed` calls

- **Metrics**: Adds `reasonInternal` to gatekeeper reason codes to distinguish permanent server faults from client rejections

## Implementation Details

- `MarshalFailed` uses `Permanent(Internal(…))` to compose the permanent marker with an internal error code
- The cause is attached via `WithCause()` so it's available for server-side logging but excluded from client replies
- Tests verify: permanent classification, client message sanitization (no cause leakage), cause reachability for logs, and nil-cause safety
- Integration test confirms broadcast-worker's `publishMutation` correctly marks marshal failures as permanent

https://claude.ai/code/session_017nKyc2rJv8pLJnpLUosk6C

## Summary by CodeRabbit

* **Bug Fixes**
  * Serialization failures are now consistently classified as permanent internal errors.
  * Messages that cannot succeed are acknowledged instead of repeatedly retried.
[...]

PR #455, opened 2026-09-01
fix(message-gatekeeper): tell the sender when a send is dropped, and contain panics

hmchangw:
Two failure modes on the one hop every user message crosses. **`MESSAGES` has exactly one consumer** — nothing else in the repo binds `stream.Messages` — so this service's 1000-slot `MaxAckPending` budget *is* the site's entire send capacity. Both of these are site-wide.

## 1. Sends are dropped silently

The infra branch NAKs and returns without replying. It already computes `final` (the configured last delivery) for its metrics and does nothing else with it:

```go
final := natsmetrics.IsFinalDeliveryFromContext(ctx)
result := resultRetry
if final { result = resultFailed }          // used for metrics only
...
jsretry.Nak(ctx, msg, jsretry.DefaultBackoff, "process message failed (infra)")
```

So when the budget is spent, JetStream drops the message and **the sender is never told.** It published onto MESSAGES *successfully* and then waits forever on a reply subject nothing will ever publish to.

The sequence at ~100 failing sends/sec:

| t | state |
|---|---|
| 0s | Mongo breaker opens; sends begin NAKing with delay |
| ~10s | the 1000-slot budget fills — the server delivers nothing further on this consumer |
| 10s+ | clients keep publishing successfully and never receive a reply |
| 1h01–2h03 | each parked send is dropped permanently, still silently |

> **Correction, twice over.** The first version of this description gave that last row as "~6.3 min", then "6.3–12.6 min". Both are now wrong, for different reasons.
>
> The first was the jittered *floor* quoted as the figure: `jsretry.DefaultBackoff` is `{1s, 5s, 30s, 2m, 10m}`, and `jsretry` applies AWS equal jitter, so each delay lands in `[d/2, d]` — at `MaxDeliver=6` that is 6.3 min best case, 12.6 min worst.
>
> The second went stale when **#484** landed on `main`: `message-gatekeeper` now builds its consumer with `stream.WithOutageRetryBudget(s, jsretry.DefaultBackoff)`, which takes `MaxDeliver` from 6 to **17** and the window to **1h01 floor / 2h03 nominal** (measured against the merged tree, not derived by hand). The reply this PR adds still fires on the final delivery, so the sender now waits that much longer to learn a send is over.
>
> That is the right pairing rather than a collision. #484's own comment on `buildConsumerConfig` says the budget matters because the message "would be dropped after ~2.6 min **with no reply ever sent**" — it lengthened the retry; this PR supplies the reply at the end of it.

**Fix:** reply once, on the final delivery, with `unavailable`. Earlier deliveries stay silent — the send may still succeed, and reporting failure while a retry is pending would be wrong.

`MarshalQuiet`, not `Marshal`: this branch already logs the cause, and `Marshal` runs `Classify`, which would log it a second time (CLAUDE.md: never log AND return).

**What this buys.** No message is saved and no capacity is added — the drop still happens. What changes is that a dropped send acquires a terminal state. `msg.send` is a plain publish whose result arrives asynchronously on `chat.user.{account}.response.{requestID}`, a subject clients already hold a long-lived subscription to (it also carries every room operation's `AsyncJobResult`), so a late reply does reach them. Today that subscription simply never hears anything: the publish succeeded, so the client has positive evidence of acceptance followed by silence indistinguishable from slowness. Afterwards the two states are distinguishable — *no reply* means still in flight and may yet succeed; *`unavailable`* means it is over — which is what makes a retry decidable. What a retry is safe *from* is narrower than this PR first claimed; see Docs below.

## 2. No panic containment

The handler was passed to `natsmetrics.Start` as a bare closure, and `Consume` spawns a goroutine per message with **no `recover()`**. A panic kills the process with the message un-acked, so JetStream redelivers it on `{30s,1m,2m,4m,8m}` — **one malformed payload crash-loops the pod repeatedly, and no message on the site is accepted for that whole window.**

Every other MESSAGES-CANONICAL consumer already wraps in `jobguard` (`broadcast-worker`, `message-worker`, `notification-worker`, `room-worker`). `guardedProcessor` here mirrors `broadcast-worker`'s exactly.

Note the trade this makes, deliberately: `jobguard.Run` Acks on panic, so a poison message is dropped rather than retried, and the sender gets no reply (the new `unavailable` covers only the infra branch). For a deterministic poison payload that is strictly better than crash-looping the site. For a *non-deterministic* panic it converts a retry that might have succeeded on another pod into a permanent drop of one message. Closing that asymmetry — replying before the Ack — is a separate change.

The drop is not silent, which is what makes the trade acceptable: `jobguard.Run` calls `MarkHandlerPanic`, and `natsmetrics.Message` implements it as `MarkTerminal(TerminalInternal)` (`pkg/natsmetrics/metrics.go:500`), so each Ack-drop increments `chat.nats.terminal.failures{reason=internal}` alongside jobguard's `dropped poison message after panic (Ack)` WARN. It is alertable.

## Tests

- **`TestHandleJetStreamMsg_FinalDeliveryRepliesToSender`** — drives a real tracked final delivery with a failing canonical publish, asserts the sender gets an `unavailable` envelope. Confirmed RED before the fix:
  ```
  Error: "[]" should have 1 item(s), but has 0
  Messages: the last delivery must tell the sender the send failed
  ```
  It drives finality through `natsmetrics.Track(..., maxDeliver)` rather than a hardcoded 6, so #484's change to the budget does not touch it.
- **`TestHandleJetStreamMsg_NonFinalDeliveryStaysSilent`** — pins the other half, so a future change can't start reporting failure on every retry.
- **`TestGuardedProcessor_RecoversPanic`** — asserts the panic is contained *and* the poison message is Acked rather than left to redeliver.

## Docs

The new reply is a client-visible error on `msg.send`, so per CLAUDE.md §5 both move in this PR:

- `docs/client-api.md` §4 gains the `unavailable` row, and its **delivery guarantee** paragraph is corrected — it claimed infrastructure failures "produce no immediate reply", which is no longer true of the last one. The new text draws the distinction a client needs: *no reply* means the send may still succeed; *`unavailable`* means it is over.
- `docs/client-api/request-reply.md` (derived view) gains the matching bullet.

`unavailable` is an established wire code rather than a new one — auth-service, upload-service and the read-receipts path already return it — so no client needs to learn a new code to handle this.

> **Correction (`ab9d598`).** The first version of that row told clients they "may safely retry with the same `id` (the canonical publish is deduplicated on it)". The dedup id is real — `handler.go:525` publishes `WithMsgID(natsutil.CanonicalDedupID(&evt))`, which for a create is the message id itself — but the window is shut by the time the advice is read. `pkg/stream.MessagesCanonical` sets `Name` and `Subjects` only, so the server applies `StreamDefaultDuplicatesWindow` = **2 minutes**, while this reply is now sent an hour or more into the ladder. It matters only when the original publish landed and its *acknowledgement* was lost — the shape a JetStream wobble actually takes — and then a retry is a genuine second copy, fanned out and notified twice. Cassandra is keyed on `message_id`, so stored history still converges; that is the part the row can honestly promise, and now the only part it does.
>
> Related, and **not** fixed here: the retry ladder's own long rungs already sit far beyond that 2-minute window, so a lost-PubAck message can be double-published by the internal retries alone, with no client involved. #484's longer budget widens that gap rather than narrowing it. Setting `Duplicates` explicitly on MESSAGES-CANONICAL (≥ the full ladder) belongs in `pkg/stream` plus ops/IaC provisioning, not in this PR.

## Verification

- `go test ./message-gatekeeper/ -race` — pass
- `make lint` — `0 issues`
- No store interface changed, so no `make generate` needed
- **Rebased onto `main` at `ae81189`** — replayed with no conflicts. `go build ./...`, `go vet ./...`, the full `go test ./... -race` and `golangci-lint run` (0 issues) all pass on the result. An earlier rebase onto `3ec5822` had one conflict in `message-gatekeeper/main.go`, where both sides added an import on the same line (`jobguard` here, `jsretry` on main) and both were kept; that resolution is carried in the current history. The branch's original grpc bump for GO-2026-6348 dropped out as an already-applied cherry-pick, because `main` landed the identical `v1.83.1` bump via #497.

## Known limits of the fix

- **Finality is this pod's belief, not the server's.** `IsFinalDelivery` compares `numDelivered` against the `MaxDeliver` this pod computed. `CreateOrUpdateConsumer` normally keeps the two in sync, but a rollout that *changes* the effective cap leaves old pods holding the old number — one that believes the cap is lower will reply `unavailable` while the server still has deliveries left.

  > **Correction.** An earlier version said "This deploy is exactly such a rollout, since #484 moves the cap from 6 to 17." That is not this deploy. #484 merged separately on 2026-09-15 (`0e1a4de`), and `WithOutageRetryBudget` is already on `main` outside this diff — this PR does not touch `buildConsumerConfig` at all. **#484 is confirmed to deploy ahead of this change**, so old and new pods both hold `MaxDeliver=17` across this rollout: the effective cap does not move, and there is no window for a premature `unavailable`. The hazard stays documented because it is real for any *future* change to the cap — whoever makes that change has to sequence it, since this reply path trusts the pod's own number.

- **Unlimited redelivery silently disables it.** `IsFinalDelivery` requires `maxDeliver > 0`, so under `CONSUMER_MAX_DELIVER=-1` there is no final delivery, no reply is ever sent, and the silent drop returns with no signal. It fails in the safe direction — never a false `unavailable` — but it is invisible.

## Scope

The audit found more on this service — the consumer binds all of `msg.>` with no `FilterSubjects`, the parent-resolution path has no overall deadline (6.15s worst case), and `main.go` sits at 2.4% coverage. Each is a separate change; this PR is the two site-wide failure modes.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Clients now receive a final `unavailable` reply when message delivery retries are exhausted, instead of waiting indefinitely.
  * Earlier delivery failures continue retrying silently, and clients may retry using the same message ID.
  * Message-processing panics are contained to prevent repeated redelivery loops and service disruption.

* **Documentation**
  * Updated client API and request-reply documentation to clarify retry behavior and terminal delivery failures.


julianshen:
**6 findings · 4 unique issues · 4 inline comments posted** — 🔴 4 blocking · 🟠 2 warning

- **A lost publish acknowledgement can produce a false failure reply**
  - `message-gatekeeper/handler.go:251` (`distributed-system`) — A lost publish acknowledgement can produce a false failure reply
  - `message-gatekeeper/main.go:289` (`silent-failure`) — Recovered handler panics are acknowledged as successful sends
  - `message-gatekeeper/main.go:292` (`nats`) — Panic recovery ACKs messages that have not completed processing.
  - `message-gatekeeper/main.go:289` (`distributed-system`) — Acking a recovered panic permanently drops the send without a reply

<details>

* `docs/client-api.md`
* `message-gatekeeper/final_delivery_test.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

message-gatekeeper/handler.go
```diff
@@ -242,6 +242,15 @@ func (h *Handler) HandleJetStreamMsg(ctx context.Context, msg jetstream.Msg) {
 			// flow terminal for the infra path; the Error line below carries the cause.
 			slog.Log(ctx, logctx.LevelFlow, "gatekeeper nak", "phase", "nak", "request_id", req.RequestID)
 			slog.ErrorContext(ctx, "process message failed (infra)", "error", err, "account", account, "room_id", roomID)
+			if final {
+				// The delivery budget is spent: JetStream drops the message after
+				// this NAK, so nothing will ever publish to the sender's reply
+				// subject. The client published onto MESSAGES successfully and
+				// would otherwise wait forever for a reply that cannot come.
+				// MarshalQuiet because this branch already logged the cause.
+				h.sendReply(ctx, account, &req, errnats.MarshalQuiet(
```
julianshen, line 251: _🎯 distributed-correctness_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`_

**A lost publish acknowledgement can produce a false failure reply**

A canonical JetStream publish can commit successfully and still return an error if its PubAck is lost or times out. If this happens on the final delivery, this branch sends `unavailable` even though the canonical message exists and may reach recipients. Deduplication only resolves ambiguous outcomes on a subsequent attempt; there is no subsequent gatekeeper attempt after this final NAK. The sender therefore observes failure while the message is delivered, contradicting the new documented claim that the send did not happen. Distinguish pre-publish store failures from ambiguous publish errors and report the latter as outcome-unknown with same-ID reconciliation/retry semantics, or persist an authoritative acceptance status before returning a definitive result.

<details>

```
In `message-gatekeeper/handler.go` around line 251: A canonical JetStream publish can commit successfully and still return an error if its PubAck is lost or times out. If this happens on the final delivery, this branch sends `unavailable` even though the canonical message exists and may reach recipients. Deduplication only resolves ambiguous outcomes on a subsequent attempt; there is no subsequent gatekeeper attempt after this final NAK. The sender therefore observes failure while the message is delivered, contradicting the new documented claim that the send did not happen. Distinguish pre-publish store failures from ambiguous publish errors and report the latter as outcome-unknown with same-ID reconciliation/retry semantics, or persist an authoritative acceptance status before returning a definitive result.

```

</details>

---

message-gatekeeper/main.go
```diff
@@ -273,3 +274,18 @@ func buildConsumerConfig(s stream.ConsumerSettings) jetstream.ConsumerConfig {
...
+// natsmetrics.Consume spawns a goroutine per message with no recover(), so an
+// unrecovered panic in HandleJetStreamMsg takes the process down with the
+// message un-acked — and MESSAGES has exactly one consumer, so JetStream
+// redelivers it into a crash loop that stops the site accepting any message at
+// all for the length of the backoff schedule. On panic jobguard Acks, dropping
+// the poison message instead. Every other MESSAGES-CANONICAL consumer already
+// composes this way; this one did not.
+func guardedProcessor(process natsmetrics.ProcessMessage) natsmetrics.ProcessMessage {
+	return func(msgCtx context.Context, msg *natsmetrics.Message) {
+		jobguard.Run(msg, func() { process(msgCtx, msg) })
```
julianshen, line 289: _🏷️ silent-failure_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`, `nats`, `silent-failure` (3 rules)_

**Recovered handler panics are acknowledged as successful sends**

`jobguard.Run` acknowledges the message after `HandleJetStreamMsg` panics, as explicitly asserted by `TestGuardedProcessor_RecoversPanic`. If the panic occurs before the canonical publish or sender reply, JetStream permanently removes an unprocessed send and the sender waits indefinitely, converting the task failure into apparent success. Recover the panic without ACKing: NAK non-final deliveries for retry, and on the final delivery send an `unavailable` reply before terminally handling the message. The existing message ID deduplication makes redelivery preferable to silent loss.

<details>

- **`nats`**:292: The new test explicitly verifies that `jobguard.Run` ACKs when the processor panics. If `HandleJetStreamMsg` panics before publishing the canonical message or replying, that ACK marks the source delivery successful and JetStream permanently removes it; the sender receives no reply. Because the panic is now recovered, retrying no longer crash-loops the process. Recover by NAKing with bounded backoff instead, and on the final delivery send `unavailable` when the reply identity is recoverable before terminating the message.
- **`distributed-system`**:289: The added test establishes that `jobguard.Run` Acks when `process` panics. If `HandleJetStreamMsg` panics before publishing the canonical message or a response, the guard recovers and Acks, JetStream removes the message immediately, and the sender receives neither success nor the new terminal `unavailable` reply. Thus panic containment reintroduces the silent permanent-drop behavior this PR is intended to remove. Recover through an application-aware path that NAKs with bounded backoff while deliveries remain and emits an appropriate terminal/indeterminate response before Ack/Term on the final attempt, when the account and request ID are recoverable.

</details>

<details>

```
In `message-gatekeeper/main.go` around line 289: `jobguard.Run` acknowledges the message after `HandleJetStreamMsg` panics, as explicitly asserted by `TestGuardedProcessor_RecoversPanic`. If the panic occurs before the canonical publish or sender reply, JetStream permanently removes an unprocessed send and the sender waits indefinitely, converting the task failure into apparent success. Recover the panic without ACKing: NAK non-final deliveries for retry, and on the final delivery send an `unavailable` reply before terminally handling the message. The existing message ID deduplication makes redelivery preferable to silent loss.

```

</details>

---

message-gatekeeper/handler.go
```diff
@@ -242,6 +242,15 @@ func (h *Handler) HandleJetStreamMsg(ctx context.Context, msg jetstream.Msg) {
...
+					errcode.Unavailable("message could not be accepted, please retry")))
```
julianshen, line 252: _🏷️ request-reply_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`nats`_

**A publish timeout can be reported as a definite dropped send.**

This final-delivery branch also handles canonical JetStream publish errors. A publish can time out after the server appended the canonical message but before its acknowledgement reached the publisher. On that concrete path, canonical consumers can deliver the message while this branch tells the sender it "could not be accepted" and the documentation says the send is over. Preserve the error provenance and only claim a definite failure for errors known to occur before publication, or describe canonical publish-ack failures as an unknown outcome and instruct clients to retry the same deduplicated ID.

<details>

```
In `message-gatekeeper/handler.go` around line 252: This final-delivery branch also handles canonical JetStream publish errors. A publish can time out after the server appended the canonical message but before its acknowledgement reached the publisher. On that concrete path, canonical consumers can deliver the message while this branch tells the sender it "could not be accepted" and the documentation says the send is over. Preserve the error provenance and only claim a definite failure for errors known to occur before publication, or describe canonical publish-ack failures as an unknown outcome and instruct clients to retry the same deduplicated ID.

```

</details>

---

docs/client-api.md
```diff
@@ -6458,8 +6458,9 @@ Delivered on `chat.user.{account}.response.{requestId}`. See [Error envelope](#6
...
 | `quoted parent {id} not found` | `not_found` | — | The quoted message lookup failed (deleted, cross-room, …). |
 | `quoted parent {id} thread context mismatch: …` | `bad_request` | — | A quoted message must be in the same thread context (main-room or the same thread) as the new message — except a `tshow: true` thread reply, which may also be quoted from its parent channel room. |
+| `message could not be accepted, please retry` | `unavailable` | — | An infrastructure failure (store or canonical-publish error) that was still failing on the **last** JetStream delivery. Sent once, at the end of the retry budget, because the message is dropped after it — see the delivery guarantee below. Earlier deliveries are retried silently. The send did not happen; the client may safely retry with the same `id` (the canonical publish is deduplicated on it). |
```
julianshen, line 6461: _🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**A publish error does not prove that the send did not happen.**

Canonical-publish errors can be indeterminate: the server may accept the publish while its acknowledgement is lost or times out. On the final attempt, the handler would then return `unavailable` even though the canonical message exists. This row says “The send did not happen,” and the delivery paragraph plus `docs/client-api/request-reply.md` say the reply means it will not succeed, giving clients a false definitive result. Describe `unavailable` as an unconfirmed outcome and retain the safe same-ID retry guidance, or only make the definitive claim for errors known to occur before publication. The response text in `message-gatekeeper/handler.go` should be made similarly non-definitive.

<details>

```
In `docs/client-api.md` around line 6461: Canonical-publish errors can be indeterminate: the server may accept the publish while its acknowledgement is lost or times out. On the final attempt, the handler would then return `unavailable` even though the canonical message exists. This row says “The send did not happen,” and the delivery paragraph plus `docs/client-api/request-reply.md` say the reply means it will not succeed, giving clients a false definitive result. Describe `unavailable` as an unconfirmed outcome and retain the safe same-ID retry guidance, or only make the definitive claim for errors known to occur before publication. The response text in `message-gatekeeper/handler.go` should be made similarly non-definitive.

```

</details>

---

julianshen (commented):
### tGDBot review of `49ab9b8` — 6 findings in 5 files

4 🔴 blocking · 2 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/455#issuecomment-5491842979), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

hmchangw:
## `lint` is red on a finding I cannot attribute to this PR

Not pushing a fix for it, and not re-running again. Here is exactly what I know, including the part that does not fit.

**The failure** (both attempts of run `35078299136`, head `20e4e29`):

```
pkg/errcode/permanent_test.go:102:7:  SA5011(related information): this check suggests that the pointer can be nil (staticcheck)
pkg/errcode/permanent_test.go:105:10: SA5011: possible nil pointer dereference (staticcheck)
```

**Why it does not look like this PR's:**

- This PR's diff is 5 files — `docs/client-api.md`, `docs/client-api/request-reply.md`, and three under `message-gatekeeper/`. It touches **no file under `pkg/`**, and `pkg/errcode/permanent_test.go` is byte-identical to `main` (unchanged since #423).
- A full `golangci-lint run` reproduces clean on this exact head: **0 issues**, both warm-cache and after `golangci-lint cache clean`. Same linter version CI installs (v2.11.4), same Go (1.25.13), same core count (4).
- The same full run on `main@3ec5822` is also 0 issues.
- #453, rebased onto the identical base, passed `lint` in CI.

**The part that does not fit, stated plainly:** `main` is green in CI, so the usual "red on the base too" attribution does not apply here. #453's green lint job restored the *same* cache key as both of this PR's red runs (`golangci-lint.cache-Linux-2958-199ce439…`), which rules out cache poisoning as the differentiator. Same base, same cache, same toolchain — green there, red here, and the only delta is five files that cannot influence analysis of `pkg/errcode` (facts flow from dependencies to dependents, and `pkg/errcode` does not depend on `message-gatekeeper`). The one measurable difference is wall-clock: 71s on the green run vs 107s on both red ones, which is consistent with staticcheck losing the `testing`-package no-return facts that SA5011 needs to see that `t.Fatal` ends the test. I could not confirm that, so I am not asserting it.

I re-ran the failed job once to check for nondeterminism; it reproduced identically, so the failure is deterministic in CI and that re-run is spent.

**Proposed patch** (not applied here — it is in a package this PR does not own, and widening the diff to chase a finding I cannot attribute seems worse than leaving it visible):

```go
 		if ee == nil {
 			t.Fatal("terminal must return the typed *Error")
+			return
 		}
 		if ee.Code != tc.wantCode {
```

The `return` is unreachable at runtime, but it makes the dereference provably unreachable when `ee` is nil *by control flow alone*, so SA5011 cannot fire whether or not staticcheck infers `t.Fatal` as no-return. It removes the analyzer's dependence on that inference rather than suppressing the warning, so it is not a `//nolint` and it does not weaken the assertion. I have not been able to verify it empirically, because I cannot reproduce the failure it is meant to fix.

Happy to push it here or as a standalone PR against `main` — say which.

---

PR #458, opened 2026-09-01
fix(model): classify legacy migrated member types as system messages

evelynchang5525:
Follow-up to #451, which flagged this as out of scope. Independent of it — this branch is cut from `main` and touches neither `sysmsgname.go` nor `pkg/preview`.

## The bug

The migration wrote member removals and self-leaves into Cassandra as `members_removed` / `members_left`. Nothing in this repo produces those types any more, but `IsSystemMessageType` doesn't recognise them — so the two paths that classify by the **stored** type treat a membership notice as ordinary user content:

- **`history-service/internal/service/rooms.go:255`** — the lazy room-list preview walk can select a legacy member row as a room's preview. The walk reads Cassandra directly, so the wire-side type rewrite in #451 never runs on it.
- **`data-migration/es-index-migrator/runner.go:63`** — the backfill indexes legacy member rows as searchable content. Its own comment says "Sys-messages are dropped on the live path […]; the backfill must match or it re-introduces them" — it currently doesn't.

## The fix

Two named constants, added to `systemMessageTypes`.

## Why this is safe to put in the shared map

`IsSystemMessageType` has 8 call sites. Five read the type off MESSAGES-CANONICAL, where these values **can never appear** — no producer emits them, and the migration wrote them straight into Cassandra without republishing. So this is a no-op on all of them:

| consumer | what it gates | reachable by a legacy type? |
|---|---|---|
| `message-worker` (nil-sender fallback, metric label) | sender resolution, metrics | no — live only |
| `notification-worker` `isNotifiable` | push notifications | no — live only |
| `search-sync-worker` | live index filter | no — live only |
| `roomlist-worker` | `SystemMsg` wire flag | no — live only |
| `broadcast-worker/preview.go` | insert-side preview eligibility | no — live only |
| **`history-service` preview walk** | preview eligibility | **yes — fixed** |
| **`es-index-migrator`** | backfill index filter | **yes — fixed** |

Keeping one membership list also matters: `notification-worker`'s comment states new system types are "safe-by-default as long as they join `IsSystemMessageType` (the single membership list)". A preview-only legacy set would have fixed the walk, left the backfill broken, and created the second concept that comment warns against. It would also have split the predicate `rooms.go:254` deliberately shares with the insert side ("so the two cannot disagree").

**The caveat, stated in the code:** "legacy types never reach live paths" is a claim about data, not code. It holds today because no producer emits them — but a replay tool that republished historical rows onto MESSAGES-CANONICAL would put them on the live paths, where they'd classify as system messages. Correct, but worth knowing before building one. The constants carry that note.

**Also worth knowing:** the plural shape is not the tell — `members_added` is a *current* type, not a legacy one. Asserted in the test so nobody "fixes" it by pattern.

## Why making rows ineligible can't wrongly clear a preview

`previewEmpty` is destructive (`previewAfterMutation` clears the stored preview on it), so excluding more rows deserves a second look. The walk only returns `previewEmpty` on `!page.HasNext` — genuine end of history. An ineligible tail that exhausts the scan budget returns `previewDegraded` instead: "a survivor may exist, so unknown, never a clear." So a room whose recent tail is all legacy member rows degrades to unknown rather than losing its preview.

## Operational note: this stops recurrence, it does not clean up

If a backfill has **already completed** against a persistent index, the legacy member documents are in it and this PR leaves them there. The job only ever upserts — `buildMessageAction` emits `ActionIndex` with `DocID: msg.MessageID`, externally versioned — so a skipped row is simply not written, never retracted. On a fresh site, or a run into a new versioned index, this PR is sufficient by itself.

They also cannot be removed by a `delete_by_query` on type: `searchindex.MessageFields` carries no `type` field, so in ES a legacy member document is indistinguishable from an ordinary message except by its content text.

Two cleanup routes, both deliberately out of scope here:

- **Reindex into a fresh index** — the shape the job was designed for ("every write is versioned/idempotent … so a re-run or overlap with live traffic converges"). With this change the new index never receives them; the alias cuts over. No delete logic needed.
- **A targeted delete pass** re-deriving the affected `message_id`s from Cassandra `messages_by_room` and issuing bulk deletes against the existing index. Only needed if reindexing isn't an option.

A data-cleanup job wants its own operational review, so it isn't bundled here.

## Testing

TDD — the red phase was verified by removing the map entries and confirming the two new preview-walk subtests fail on `IsSystemMessageType`, then restoring.

- `pkg/model`: both types classify as system; constant values pinned; `members_added` asserted still current.
- `history-service`: both added to the existing `TestPreviewWalk_SkipsEverySystemType` table, so the walk provably steps over a legacy row to the next eligible message.
- `es-index-migrator`: both added to `TestRunMessages_SkipsSystemMessages`, asserting they're filtered rather than counted as build failures.

`make test` (full repo, `-race`) pass · `make lint` 0 issues · `gosec` pass · semgrep repo rules 0 findings. govulncheck and semgrep's registry ruleset are network-blocked in the dev environment and run in CI.

https://claude.ai/code/session_01JKU3KzpKzseUknJKgfL3Xv

### DIFF
--- a/data-migration/es-index-migrator/runner_test.go
+++ b/data-migration/es-index-migrator/runner_test.go
@@ -84,6 +84,12 @@ func TestRunMessages_SkipsSystemMessages(t *testing.T) {
 				{MessageID: "m1", RoomID: "room1", CreatedAt: now},
 				{MessageID: "sys1", RoomID: "room1", CreatedAt: now, Type: model.MessageTypeMembersAdded},
 				{MessageID: "sys2", RoomID: "room1", CreatedAt: now, Type: model.MessageTypeRoomRenamed},
+				// Migrated rows. The backfill streams Cassandra directly, so these
+				// arrive with their stored plural type and must be filtered like any
+				// other membership notice — the live path never indexes their modern
+				// equivalents, and the backfill has to match.
+				{MessageID: "sys3", RoomID: "room1", CreatedAt: now, Type: model.MessageTypeLegacyMembersRemoved},
+				{MessageID: "sys4", RoomID: "room1", CreatedAt: now, Type: model.MessageTypeLegacyMembersLeft},
 				{MessageID: "m2", RoomID: "room1", CreatedAt: now, Type: model.MessageTypeImportant},
 			} {
 				if err := fn(m); err != nil {
--- a/history-service/internal/service/preview_walk_test.go
+++ b/history-service/internal/service/preview_walk_test.go
@@ -280,6 +280,10 @@ func TestPreviewWalk_SkipsEverySystemType(t *testing.T) {
 		model.MessageTypeRoomRenamed,
 		model.MessageTypeRoomRestricted,
 		model.MessageTypeTeamsMeetStarted,
+		// Migrated rows. The walk reads Cassandra directly, so it classifies by the
+		// STORED type — the wire-side rewrite in normalizeLegacySysMsgs never runs here.
+		model.MessageTypeLegacyMembersRemoved,
+		model.MessageTypeLegacyMembersLeft,
 	}
 	for _, st := range systemTypes {
 		t.Run(st, func(t *testing.T) {
--- a/pkg/model/event.go
+++ b/pkg/model/event.go
@@ -726,6 +726,21 @@ const (
 	MessageTypeTeamsMeetStarted = "teams_meet_started"
 )
 
+// Legacy member types exist only on migrated Cassandra rows — nothing in this repo
+// produces them, and the plural shape is not the tell (MessageTypeMembersAdded is
+// current). history-service rewrites them to their modern equivalents on the wire, so
+// clients never see them; these constants are for classifying the STORED type.
+//
+// A replay tool that ever republished historical rows onto MESSAGES-CANONICAL would
+// put them on the live paths too, where they would classify as system messages —
+// correct, but worth knowing before building one.
+const (
+	// MessageTypeLegacyMembersRemoved is the migrated form of MessageTypeMemberRemoved.
+	MessageTypeLegacyMembersRemoved = "members_removed"
+	// MessageTypeLegacyMembersLeft is the migrated form of MessageTypeMemberLeft.
+	MessageTypeLegacyMembersLeft = "members_left"
+)
+
 // systemMessageTypes is the set of system/event Message.Type values (not user content),
 // for fast membership checks. Keep in sync with the MessageType* constants above.
 var systemMessageTypes = map[string]struct{}{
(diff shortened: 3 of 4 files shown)


julianshen:
## Needs clarification

Has this backfill already populated any persistent index, and if so what removes the existing legacy documents?
Reply in this thread, or post: `answer clar_uf5xjorebn5ioqtl3i64z5enej: <your answer>`

---

_🤖 Posted by [tGDBot](https://github.com/julianshen/tGDBot)_

julianshen:
**No actionable comments.** ✅

### Needs clarification

Has this backfill already populated any persistent index, and if so what removes the existing legacy documents?

`answer clar_uf5xjorebn5ioqtl3i64z5enej: <your answer>`

[Open the question](https://github.com/hmchangw/newchat/pull/458#issuecomment-5497696897)

### Related work

- [PR #451](https://github.com/hmchangw/newchat/pull/451) — fix\(history-service\): resolve bot app names and normalize legacy member sys-msg types (open)

<details>

* `data-migration/es-index-migrator/runner_test.go`
* `history-service/internal/service/preview_walk_test.go`
* `pkg/model/event.go`
* `pkg/model/model_test.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

evelynchang5525:
answer clar_uf5xjorebn5ioqtl3i64z5enej: Good catch — the honest answer is that **this PR prevents recurrence but removes nothing**, and the cleanup is harder than a delete-by-query. Three facts from the code:

**1. The backfill only ever upserts.** `buildMessageAction` returns `searchengine.ActionIndex` with `DocID: msg.MessageID` into `searchindex.MessageIndexName(prefix, msg.CreatedAt)`, externally versioned on `CreatedAt`. There is no delete path in the job at all. So after this change a skipped legacy row is simply *not written* — re-running does not retract a document an earlier run already indexed.

**2. So yes: if a backfill has already completed against a persistent index, those documents are in it, and this PR leaves them there.** On a fresh site, or a run into a new versioned index, this PR is sufficient on its own.

**3. And they can't be deleted by type, because the indexed document has no type field.** `searchindex.MessageFields` carries MessageID, RoomID, SiteID, UserID, UserAccount, Content, CreatedAt, EditedAt, UpdatedAt, ThreadParentID, ThreadParentCreatedAt, TShow, Attachments, Card — no `type`. A `delete_by_query` therefore has nothing to match on; the legacy member documents are indistinguishable in ES from ordinary messages except by their content text.

That leaves two workable cleanups, both outside this PR:

- **Reindex into a fresh index** — the shape the job was built for ("every write is versioned/idempotent … so a re-run or overlap with live traffic converges"). With this change in place the new index simply never receives them, and the alias cuts over. Cleanest, and no delete logic needed.
- **A targeted delete pass** re-deriving the offending `message_id`s from Cassandra (`messages_by_room` where `type` in `members_removed`/`members_left`) and issuing bulk deletes against the existing index. Needed only if reindexing is not on the table.

I did not add either here: the fix belongs where it is (stop indexing them), and a data-cleanup job is a separate change with its own operational review. Happy to open an issue for whichever route you prefer — say the word.

I have added this as an operational note to the PR description so an operator does not have to find it in a thread.

---

mliu33 (approved):

PR #460, opened 2026-09-01
refactor(loadgen): extract soak foundations and user-read lane

flywindy:
## Summary

- Add characterization contracts for soak metric descriptors and the deployment manifest.
- Extract distribution, topology, wire DTOs, and RPC transport into focused internal soak packages.
- Move the user-read lane as the first end-to-end package consumer and adapt its samples back to the shared root recorder.
- Keep moved-package coverage visible through repeatable coverage include paths, and make the gate reject any include that matched nothing.

## Scope

This is PR 1 of the three-PR soak refactor roadmap. It establishes the package foundations and proves them with one complete lane.

No behavior, CLI/env contract, wire JSON schema, or Prometheus metric name/label change is intended. Compatibility aliases remain for the message, room, member, presence, and search lanes that still live in the root loadgen package.

## Review by commit

1. bede39d0 — lock metric and manifest compatibility contracts
2. abf3b56f — extract deterministic distribution logic
3. 9ebe3eed — extract topology construction and membership invariants
4. e4d91ac1 — extract wire DTO contracts
5. 0669d6f7 — extract RPC actions, retries, transport, and error classification
6. 52baf7d7 — extract the user-read lane and preserve moved-package coverage
7. 54ed18d7 — wire userread directly, add the recorder adapter, and remove userread-only shims
8. 01b55ab2 — convert rather than hand-copy the user-read sample, so the compiler pins the two shapes together
9. 495d2c47 — mark the adapter test's `math/rand` as load-generator randomness (SAST gate)
10. 435d47f4 — reject any coverage include pattern that matched no statements (review follow-up)

Each commit builds independently. Commits 1 and 7 passed the complete loadgen race suite; extraction commits 2-6 passed their focused package race tests. Commits 8-10 passed the full loadgen + coveragecheck race suite.

## Verification

- make fmt
- make build SERVICE=tools/loadgen
- make test SERVICE=tools/coveragecheck
- make test SERVICE=tools/loadgen
- make coverage-loadgen-soak
  - root soak aggregate: 81.25% (required 80%)
  - core union: 90.13% (required 90%)
  - internal soak packages: 95.26% (required 90%)
  - userread: 98.2%

The coveragecheck change adds a failure condition but moves none of these numbers: all three `coverage-loadgen-soak` invocations were re-run against a real profile and every include still matches (`total=` 4881 / 4153 / 548). Misspelling an include now exits 1 instead of passing on a sibling's matches.

Two claims in this PR are backed by a deliberate negative test rather than assertion:

- Reverting the sample conversion in 01b55ab2 and adding a field to one side fails the build at the adapter (`cannot convert sample ... to type *soakReadSample`); the previous hand-copy compiled and dropped the field.
- Reverting 435d47f4 makes its regression test fail with `An error is expected but got nil`.

## Local validation limitations

- make lint reaches an unchanged origin/main Windows type-check failure at user-service/main.go:395: undefined: syscall.Kill and reports 0 lint issues.
- The combined local SAST run passed gosec and Semgrep rule tests. govulncheck is blocked by the same Windows type-check failure; the Semgrep scan is blocked by a local semgrep-core RPC validation failure. Remote CI remains required for those gates. **This gap is what let the `math-random-used` finding reach CI** — the suppression convention itself is sound, and the same CI scan proves it: `internal/soak/userread/userread_test.go` imports `math/rand` with the comment and was not flagged, while the adapter test without it was.
- Docker integration tests and a manual soak smoke run were not run; this PR does not move integration-test ownership.

## Follow-ups

- PR 2: move the remaining non-failure soak lanes plus engine/store/collector/workload boundaries and retire their compatibility aliases.
- PR 3: extract the failure ledger, WAL, observer contracts, and soak runtime adapter without introducing a soak/failure package cycle.
- Naming: `distribution` and `topology` dropped the `soak` prefix once the package name carried it; `rpc` and `wire` kept their old identifiers behind exported alias shims (138 and 65 lines), and `userread` keeps local import aliases so its body did not have to change. Converging these on the `distribution` form — and deleting the shims — is a rename-only follow-up, kept out of this PR so the extraction stays reviewable.

## Summary by CodeRabbit

* **New Features**
  * Improved soak-test coverage reporting with support for multiple include patterns.
  * Added reusable components for soak distribution, topology, RPC, user-read operations, and wire data.
  * Preserved compatibility with existing load-generation workflows.

* **Tests**
  * Expanded validation for retry behavior, user reads, payload distributions, topology, wire formats, metrics, manifests, and error handling.
  * Added contract checks for coverage thresholds and serialized data compatibility.


julianshen:

### Local review memory

- `tools/loadgen/internal/soak/topology/topology.go:107` — @coderabbitai: _🩺 Stability & Availability_ _🟡 Minor_ _⚡ Quick win_ **Reject nil configuration pointers before use.** A nil configuration causes a panic in both topology entry points. Return a configuration error before reading configuration fields… ([thread](https://github.com/hmchangw/newchat/pull/460#discussion_r3906259373))

<details>
<summary>📒 Files reviewed (35)</summary>

* `Makefile`
* `tools/coveragecheck/main.go`
* `tools/coveragecheck/main_test.go`
* `tools/loadgen/internal/soak/distribution/distribution.go`
* `tools/loadgen/internal/soak/distribution/distribution_test.go`
* `tools/loadgen/internal/soak/rpc/error.go`
* `tools/loadgen/internal/soak/rpc/error_test.go`
* `tools/loadgen/internal/soak/rpc/rpc_behavior_test.go`
* `tools/loadgen/internal/soak/topology/topology.go`
* `tools/loadgen/internal/soak/topology/topology_test.go`
* `tools/loadgen/internal/soak/userread/userread.go`
* `tools/loadgen/internal/soak/userread/userread_test.go`
* `tools/loadgen/internal/soak/wire/wire.go`
* `tools/loadgen/internal/soak/wire/wire_contract_test.go`
* `tools/loadgen/soak_distribution_test.go`
* `tools/loadgen/soak_manifest_contract_test.go`
* `tools/loadgen/soak_metrics_contract_test.go`
* `tools/loadgen/soak_rpc_bridge.go`
* `tools/loadgen/soak_rpc_bridge_test.go`
* `tools/loadgen/soak_search_test.go`
* `tools/loadgen/soak_topology_bridge.go`
* `tools/loadgen/soak_topology_bridge_test.go`
* `tools/loadgen/soak_userread_adapter.go`
* `tools/loadgen/soak_userread_adapter_test.go`
* `tools/loadgen/soak_wire_bridge.go`
* `tools/loadgen/soak_wire_test.go`
* `tools/loadgen/soak_zipf_test.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

tools/coveragecheck/main.go
```diff
@@ -97,12 +97,35 @@ func readCoverageProfile(
 	if stats.total == 0 {
 		return coverageStats{}, fmt.Errorf(
 			"coverage include pattern %q matched no statements",
-			include,
+			strings.Join(includes, ","),
```
julianshen, line 126: _🏷️ silent-failure_ | _🟠 Warning_ | _⚡ Quick fix_ | _`silent-failure`_

**Unmatched include patterns are silently accepted**

The final check only rejects the scope when every include misses. In the new Makefile invocation, `tools/loadgen/soak_` can contribute statements even if the user-read include is misspelled or absent, so the command can report passing coverage without measuring the moved lane; the broader internal-package check can similarly pass from other packages. Track matches per include and return an error naming each pattern that matched no statements, with a test covering one valid and one unmatched include.

<details>

```
In `tools/coveragecheck/main.go` around line 100: The final check only rejects the scope when every include misses. In the new Makefile invocation, `tools/loadgen/soak_` can contribute statements even if the user-read include is misspelled or absent, so the command can report passing coverage without measuring the moved lane; the broader internal-package check can similarly pass from other packages. Track matches per include and return an error naming each pattern that matched no statements, with a test covering one valid and one unmatched include.

```

</details>

---

julianshen (commented):
### tGDBot review of `54ed18d` — 1 finding in 35 files

1 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/460#issuecomment-5497459953), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

tools/loadgen/internal/soak/topology/topology.go
```diff
@@ -83,19 +94,19 @@ func selectSoakUsers(
...
-) (soakTopology, error) {
-	if ids == nil || ids.channelRoomID == nil || ids.subscriptionID == nil {
-		return soakTopology{}, fmt.Errorf("soak identity generator is required")
+	ids *IdentitySource,
+) (Topology, error) {
+	if ids == nil || ids.NewChannelRoomID == nil || ids.NewSubscriptionID == nil {
+		return Topology{}, fmt.Errorf("soak identity generator is required")
 	}
-	borrowed, active, err := selectSoakUsers(users, siteID, cfg.MaxUsers, cfg.ActiveUsers, seed)
+	borrowed, active, err := selectUsers(users, siteID, cfg.MaxUsers, cfg.ActiveUsers, seed)
```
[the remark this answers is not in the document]
flywindy: Checked this against the call graph and it is not reachable — declining.

`Build` has exactly one production caller, `soak_topology_bridge.go:25`, which passes the address of a composite literal (`&soaktopology.BuildConfig{...}`) and so can never pass nil. The bridge's own `cfg *soakConfig` comes from `soak_seed.go:89` via `input.Config`, and `seedSoakTopology` already rejects that at `soak_seed.go:68-70`:

```go
if input.Config == nil {
    return soakTopology{}, fmt.Errorf("soak configuration is required")
}
```

Its only caller, `soak_main.go:384`, passes `Config: &cfg.Soak`. The remaining callers of `Build` are tests, all of which pass `&cfg`. So the suggested check would duplicate an existing guard on the only path that reaches it.

One fair point underneath the finding: `Build` does nil-check `ids` at line 105 but not `cfg`, so the two pointer parameters are treated inconsistently. That is a style question rather than a defect, and adding validation is outside what this PR is allowed to change — it is a behaviour-preserving extraction, so that a red contract test can only mean the move broke something. Noting it for the follow-up PR that cleans up this package's naming and API surface.

---

tools/coveragecheck/main.go
flywindy: Confirmed and fixed in 435d47f.

The guard was global (`stats.total == 0`), so with `-include tools/loadgen/soak_ -include tools/loadgen/internal/soak/userread/` a stale or misspelled second pattern left the first matching the root soak files and the gate passed while never measuring the moved lane. `readCoverageProfile` now tracks matches per include and names every pattern that matched nothing.

The original empty-scope check is kept rather than replaced — it is still reachable when every matched file is `-exclude`d, and `percent()` would otherwise surface that as a `0.00%` score instead of the scoping mistake it is. It now reports `matched only excluded files`.

Both cases have a regression test, as asked. The first fails without the fix with `An error is expected but got nil`. Verified against a real profile that the three `coverage-loadgen-soak` invocations still match (`total=` 4881 / 4153 / 548) and that misspelling the second pattern exits 1.

Worth flagging that this gap matters beyond this PR: later phases of the refactor split these prefixes the same way, so a silently unmeasured scope would have been the default failure mode. Thanks for catching it.

---

vjauhari-work (approved):
Done!

PR #394, opened 2026-08-27
feat(mongo): configurable read preferences for primary-loss availability

hmchangw:
Audits every MongoDB read-preference site in the repo and decides, per service
and per collection, which should move off the `primary` default so reads survive
a primary-down incident.

Key points:
- `primaryPreferred` and `secondaryPreferred` are equivalent for availability;
  the former is the correct default since it is a steady-state no-op.
- Classification turns on read-after-write exposure and on side effects outside
  Mongo (Cassandra writes, NATS publishes, MS Graph calls), not on whether a
  service writes to Mongo.
- The Graph-calling teams-* workers stay on `primary` deliberately: failing
  closed beats acting on a stale read with non-idempotent external effects.
- admin-service needs an explicit transaction read preference in the same commit
  as its client flip, or password reset breaks.

Includes a feature-by-feature availability matrix for a primary-down incident.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_015S8fgZU8TpsfWrHxxwF4eS

## Summary by CodeRabbit

* **New Features**
  * Added configurable MongoDB read preferences across services, including separate settings for client and room-key reads.
  * Improved read availability during primary database outages while preserving primary routing for transactions and authoritative session operations.

* **Bug Fixes**
  * Increased retired room-key retention from 20 to 30 minutes.
  * Startup now rejects invalid read-preference settings.
  * Room-key index creation failures no longer prevent service startup.
  * Improved upload handling with longer transfer timeouts and faster header protection.

* **Documentation**
  * Updated MongoDB availability and rollout documentation.

PR #461, opened 2026-09-01
perf(user-service): serve subscription.list sort keys from a secondary

hmchangw:
AggregateSubscriptions runs four reads and all four inherited the client
preference. One of them does not need it.

resolveSortKeys' $in over the account's cache misses is the only rooms read
sized by the whole subscription set rather than by the page, and its value is
already allowed to lag by SUBS_SORTKEY_CACHE_TTL — a hit serves a 15s-old key
without comment, so insisting a miss be fresher was self-contradictory. It now
goes through roomsSecondary, built from the repo's existing MONGO_READ_PREFERENCE
option like every other *Secondary handle.

The page baseline read keeps the plain handle. It is the freshening read: it
seeds the sort-key cache and produces the row's user-visible room fields, so a
lagging value there would be written back into the cache for every later reader
of that room. It is also page-sized, so offloading it buys little.

Both subscription reads are untouched. They are what decide whether a row appears
at all, so visibility cannot lag; only ordering can.

What a replica adds beyond the TTL, recorded on resolveSortKeys: its absences are
not authoritative. A room created inside the replica's lag reads as absent and the
existing Missing caching then holds that verdict for the TTL, so a just-created
room can sort last — or, under a withinDays window, drop out — until the entry
expires. The page baseline read writes the real key back for any such room that
reaches the page.

Corrects two comments the change invalidates, and one it does not: the reads were
never pinned to primary. MONGO_CLIENT_READ_PREFERENCE has defaulted to
primaryPreferred since #394, so "All primary" described an invariant the code did
not have.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01VPWJds4zc6jD8vpPXY8W6y

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved subscription sorting reads by routing sort-key lookups according to the configured database read preference.
  * Kept page baseline reads on the primary database for more consistent results.
  * Clarified behavior when no secondary read preference is configured.

### DIFF
--- a/user-service/mongorepo/readpref_test.go
+++ b/user-service/mongorepo/readpref_test.go
@@ -39,8 +39,7 @@ func TestSubscriptionRepo_ReadPreferenceRouting(t *testing.T) {
 		assert.NotSame(t, r.subscriptions.Raw(), r.subscriptionsSecondary.Raw())
 		// GetAppSubscription (dedup guard) must keep using the primary handle.
 		assert.NotSame(t, r.subscriptions.Raw(), r.subscriptionsSecondary.Raw())
-		// The sort-key read is already allowed to lag by SUBS_SORTKEY_CACHE_TTL, so it
-		// takes the secondary; the page baseline read stays on the primary handle.
+		// The sort-key read takes the secondary; the page baseline read does not.
 		assert.NotSame(t, r.rooms, r.roomsSecondary)
 	})
 }
--- a/user-service/mongorepo/readpref_test.go
+++ b/user-service/mongorepo/readpref_test.go
@@ -30,6 +30,7 @@ func TestSubscriptionRepo_ReadPreferenceRouting(t *testing.T) {
 		r := NewSubscriptionRepo(db, 0, 0)
 		assert.Same(t, r.enriched, r.enrichedSecondary)
 		assert.Same(t, r.subscriptions, r.subscriptionsSecondary)
+		assert.Same(t, r.rooms, r.roomsSecondary)
 	})
 
 	t.Run("secondary preference clones read handles but not the primary ones", func(t *testing.T) {
@@ -38,6 +39,9 @@ func TestSubscriptionRepo_ReadPreferenceRouting(t *testing.T) {
 		assert.NotSame(t, r.subscriptions.Raw(), r.subscriptionsSecondary.Raw())
 		// GetAppSubscription (dedup guard) must keep using the primary handle.
 		assert.NotSame(t, r.subscriptions.Raw(), r.subscriptionsSecondary.Raw())
+		// The sort-key read is already allowed to lag by SUBS_SORTKEY_CACHE_TTL, so it
+		// takes the secondary; the page baseline read stays on the primary handle.
+		assert.NotSame(t, r.rooms, r.roomsSecondary)
 	})
 }
 
--- a/user-service/mongorepo/subscriptions.go
+++ b/user-service/mongorepo/subscriptions.go
@@ -41,10 +41,16 @@ type SubscriptionRepo struct {
 	// showTeamsAccounts (SHOW_TEAMS_ROOM_ACCOUNTS) allowlists accounts that see
 	// Teams rooms even when showTeamsRoom is false.
 	showTeamsAccounts map[string]bool
-	// rooms backs the list path's batched reads: sort keys and page data. The
-	// collection belongs to room-service; we only name it.
-	rooms    *mongo.Collection
-	sortKeys *sortKeyCache
+	// rooms backs the page baseline read, which both seeds the sort-key cache and
+	// produces the row's user-visible room fields. The collection belongs to
+	// room-service; we only name it.
+	rooms *mongo.Collection
+	// roomsSecondary serves the sort-key read alone — the one rooms read whose $in
+	// is sized by the account's whole subscription set rather than by the page, and
+	// one whose value the cache already lets lag by its TTL. resolveSortKeys
+	// documents what a replica adds on top of that.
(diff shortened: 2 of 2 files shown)


julianshen:

### Related work

- [PR #394](https://github.com/hmchangw/newchat/pull/394) — feat\(mongo\): configurable read preferences for primary-loss availability (merged)

<details>

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

user-service/mongorepo/subscriptions.go
```diff
@@ -511,7 +520,7 @@ func (r *SubscriptionRepo) resolveSortKeys(ctx context.Context, subs []subLite,
 	if len(misses) == 0 {
```
julianshen, line 520: _🏷️ consistency_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`_

**Replica-stale keys can hide active rooms from windowed lists**

When an old room receives a new message on the primary, a `withinDays` request treats its cached out-of-window key as a miss. If `roomsSecondary` still has the old timestamp, this query returns that stale key, the room is excluded before pagination, and the page-baseline read cannot repair it because it only reads selected rows. The stale verdict then remains cached for the TTL and can be renewed while the replica remains behind, so a currently active room disappears from the user's activity-window list. For windowed requests, use the plain/authoritative rooms handle, or recheck secondary results that are missing or outside the cutoff before caching and excluding them.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 520: When an old room receives a new message on the primary, a `withinDays` request treats its cached out-of-window key as a miss. If `roomsSecondary` still has the old timestamp, this query returns that stale key, the room is excluded before pagination, and the page-baseline read cannot repair it because it only reads selected rows. The stale verdict then remains cached for the TTL and can be renewed while the replica remains behind, so a currently active room disappears from the user's activity-window list. For windowed requests, use the plain/authoritative rooms handle, or recheck secondary results that are missing or outside the cutoff before caching and excluding them.

```

</details>

---

julianshen, line 520: _🎯 concurrency_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`performance-and-concurrency`_

**A stale secondary response can overwrite a fresh primary cache entry**

The repository-wide sort-key cache is populated unconditionally by both resolveSortKeys and the primary-backed enrichListRows baseline read. With this change, request A can begin a lagging secondary lookup, request B can then read the current room baseline from the primary and cache it, and request A can finally overwrite that entry with an older key or Missing. If A's page does not contain that room, no primary read repairs it; later requests can misorder or exclude the room under withinDays for a newly reset cache TTL. Under sustained concurrent misses, this can recur after each expiry. Make secondary cache publication conditional on the entry remaining unchanged since the miss (for example, a versioned compare-and-set), while keeping primary baseline writes authoritative, and add an interleaving test.

<details>

```
In `user-service/mongorepo/subscriptions.go` around line 520: The repository-wide sort-key cache is populated unconditionally by both resolveSortKeys and the primary-backed enrichListRows baseline read. With this change, request A can begin a lagging secondary lookup, request B can then read the current room baseline from the primary and cache it, and request A can finally overwrite that entry with an older key or Missing. If A's page does not contain that room, no primary read repairs it; later requests can misorder or exclude the room under withinDays for a newly reset cache TTL. Under sustained concurrent misses, this can recur after each expiry. Make secondary cache publication conditional on the entry remaining unchanged since the miss (for example, a versioned compare-and-set), while keeping primary baseline writes authoritative, and add an interleaving test.

```

</details>

---

user-service/mongorepo/readpref_test.go
```diff
@@ -38,6 +39,8 @@ func TestSubscriptionRepo_ReadPreferenceRouting(t *testing.T) {
 		assert.NotSame(t, r.subscriptions.Raw(), r.subscriptionsSecondary.Raw())
 		// GetAppSubscription (dedup guard) must keep using the primary handle.
 		assert.NotSame(t, r.subscriptions.Raw(), r.subscriptionsSecondary.Raw())
+		// The sort-key read takes the secondary; the page baseline read does not.
+		assert.NotSame(t, r.rooms, r.roomsSecondary)
```
julianshen, line 43: _🧪 tests_ | _🟠 Warning_ | _`tgd-review`_

**The test does not exercise the sort-key read path.**

This assertion only proves that the constructor creates distinct collection handles. If resolveSortKeys were changed back to query r.rooms, this test would still pass, so the routing behavior introduced by the PR is unprotected. Add a test that invokes resolveSortKeys and observes which collection/read preference receives the query, for example via Mongo command monitoring or an injectable collection abstraction.

<details>

```
In `user-service/mongorepo/readpref_test.go` around line 43: This assertion only proves that the constructor creates distinct collection handles. If resolveSortKeys were changed back to query r.rooms, this test would still pass, so the routing behavior introduced by the PR is unprotected. Add a test that invokes resolveSortKeys and observes which collection/read preference receives the query, for example via Mongo command monitoring or an injectable collection abstraction.

```

</details>

---

julianshen (commented):
### tGDBot review of `e60bde9` — 3 findings in 2 files

3 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/461#issuecomment-5502819618), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

PR #462, opened 2026-09-01
fix(search-sync-worker): bot messages were never indexed

hmchangw:
## The bug

`newBotMessageCollection` binds `BOT-MESSAGES-CANONICAL-{site}`, whose subjects are `chat.bot.canonical.{site}.>`. But `FilterSubjects` branched only on `teamsOnly`, so the bot collection fell through to the default:

```go
return []string{subject.MsgCanonicalMessageWildcard(siteID)} // chat.msg.canonical.{site}.*
```

The two subject spaces are disjoint at the third token — `bot` vs `msg` — so the consumer's filter selects nothing the stream carries. **No bot message has ever reached the bot search index.**

## Why it was silent

A filter that matches nothing is not an error state in JetStream. The consumer is created, reports healthy, has zero pending and zero redeliveries. There is no failure to alert on.

It also does not degrade anything visible: `bot-message-worker` consumes the same stream with a correct filter, so bot messages are still persisted and delivered normally. The only symptom is search returning an incomplete answer — a wrong result, not an error — and only for someone who knows the message exists and goes looking for it.

A test locked it in: `messages_test.go` asserted the bot collection's filter *was* `chat.msg.canonical.site-a.*`, so the bug had a passing test describing it as intended behaviour.

> **Correction.** An earlier version of this description claimed the unindexed content included the system messages `bot-room-service/sysmsg.go` publishes on the same subject. That was wrong on two counts, both surfaced in review: `BuildAction` deliberately skips `model.IsSystemMessageType` as *"UI chrome, not searchable content"*, and those payloads are a bare `model.Message` rather than the `MessageEvent` envelope, so they never decode into an indexable event at all. What this PR recovers is bot-authored messages. The sysmsg shape is handled in the second commit — see below.

## The fix

Three parts, smallest to largest scope:

1. **Couple the filter to the stream.** `messages.go` gains a `filterSubjects func(siteID) []string` field beside the existing `streamCfg`, set by each constructor. The two can no longer be chosen independently in different places. `teamsOnly` still drives its four other behaviours — it is only removed from the filter decision, which was never what it meant.
2. **`subject.BotCanonicalMessageWildcard`** in `pkg/subject/bot.go`, so the bot filter is built by a subject builder like every other one, not a literal.
3. **A startup invariant.** `checkFilterSubjects` (`subjectoverlap.go`) refuses to start a consumer whose filter shares no subject with its own stream.

Part 3 is the part that matters beyond this one bug: it turns "the next collection repeats this" from a silent correctness loss into a container that will not start.

## Review fixes (319829a)

Two real defects found in review, both verified by reproducing them first:

- **The invariant used containment where it needed intersection.** A stream binding the literal `chat.events.created` and a consumer filtering `chat.events.*` do share that subject and work fine, but containment returned `false` — so the check meant to catch a filter that selects nothing would instead have refused to start a working config, reporting the opposite of the truth. Now a symmetric intersection test, renamed to match.
- **The corrected filter newly routes `bot-room-service`'s bare-`model.Message` sysmsgs to this consumer**, where they decode to an empty event, fail the missing-id check and are Ack-dropped as poison with an error log per membership change. That noise would have been introduced by this PR. They are system messages, which are not indexed in any shape, so they are now accounted for as *filtered*. A bare message of any other type stays poison — that would be a publisher skipping the envelope, and it should be loud.

Not fixed here, deliberately: the same shape mismatch also affects `bot-message-worker`, which decodes `MessageEvent` from this subject and passes `evt.Message` to `SaveMessage` unconditionally, so these sysmsgs reach Cassandra as empty rows today. Correcting the envelope would start persisting membership sysmsgs into message history for the first time — a product decision about what belongs in history, not a side effect a search-indexing PR should ship.

## The guard no longer disables itself on an unreadable stream (27df464)

Found in review on `09c242c`. `preflightFilters` substituted the local `StreamConfig` when the deployed-subject lookup failed or came back empty, and could then return **success**.

That check could never fail, which is what made it a hole rather than a rough edge: the filters are built from the same `pkg/subject` constants as `streamCfg.Subjects`, so checking one against the other is a compile-time tautology. A transient lookup failure against a stream whose deployed subjects **had** drifted would pass, create the inert consumer, and log the filter as verified — reintroducing the exact failure this guard exists to catch, through its own error path. The sharpest version, raised in review: credentials that permit consumer creation but deny `STREAM.INFO` would make the guard *permanently* inert for INBOX and HR — the two externally-owned streams the live read was added for.

The fallback's stated justification also did not hold. It read *"the stream may not exist yet, and consumer creation reports that far better than this check can"* — but `main.go` calls `CreateOrUpdateConsumer` immediately after and exits on error, so an absent stream aborted startup either way.

Now both cases return an error, with the cause wrapped so the operator can tell an absent stream from a denied permission from a blip (asserted via `errors.Is`). The `declared` parameter is gone rather than left unused, and the log message moved from `"consumer filter does not match its stream"` to `"consumer filter preflight failed"`, which is true of both outcomes.

**Operational note:** `search-sync-worker` now refuses to start if it cannot read a stream's deployed subjects. For service-owned streams `bootstrap.go` creates them first, so this only bites when NATS is genuinely unreachable or the credentials cannot read stream info.

## Tests

TDD, RED confirmed at each stage:

- Corrected the assertion that encoded the bug.
- `TestCollections_FilterSubjectsAreCapturedByTheirStream` — all 6 collections, the invariant itself.
- `TestCheckFilterSubjects` — 4 cases including the exact shape of this bug.
- `TestMessageCollection_BuildAction_BareSystemMessageIsFiltered` — 5 cases separating the sysmsg shape from genuine poison.
- `TestPreflightFilters` — live-vs-declared drift, plus the two refusal paths above and an `errors.Is` check that the lookup cause survives.

### The subject matcher is checked against nats-server, not by hand (09c242c)

`subjectsIntersect` was originally verified by fifteen hand-written expectations. That is the same mechanism that shipped the bug this PR fixes — a table stating what the author believed, which a reviewer then has to re-derive NATS wildcard semantics to check.

It is now checked against an oracle: every valid subject pattern over `{a, b, *, >}` up to four tokens, asserted to agree with **nats-server's own `SubjectsCollide`** — the function the broker uses to answer this exact question. 160 patterns, 25,600 pairs, and no expected value stated anywhere, so the test cannot enshrine a misunderstanding.

- Four hand-written cases remain as documentation of *intent*: the bug, the containment finding above, the shape every shipped collection uses, and a cross-site filter.
- `TestSubjectsIntersect_InvalidSubjects` stays and earns its place — the oracle's corpus is filtered by `server.IsValidSubject`, so a non-final `>`, an empty subject and a trailing empty token never reach it.
- Mutation-checked in both directions: making `*` match only another `*` fails the oracle on 6544 of 25,600 pairs while the invalid-subject table stays green; letting `>` absorb from any position fails the invalid-subject table while the oracle stays green. Neither test is redundant.
- `nats-server` is a **test-only** import (nineteen other `_test.go` files already use it). `go tool nm` reports zero `nats-server` symbols in the service binary — linking it into production would have cost ~4.6 MB and put the whole server in this worker's `govulncheck` surface.

## Verification

- `make test SERVICE=search-sync-worker` (`-race`) — ok
- full repo `make test` — 0 failures (`pkg/subject` is shared)
- `go vet -tags integration ./...` — clean (`//go:build integration` files are not compiled by `make test`)
- `make lint` — 0 issues
- `gosec` — clean
- Every function in `subjectoverlap.go` at 100% statement coverage

## Deploying this does not backfill

`pkg/stream/consumer.go:45-47`: *"DeliverPolicy is honored only at consumer creation. Updating an existing durable via `js.CreateOrUpdateConsumer` does not reset its cursor position."*

The `bot-message-sync` durable already exists with a live cursor, so shipping the corrected filter starts indexing **from now**. Recovering history means recreating or versioning the durable so it starts at `DeliverAll` — which replays all retained `BOT-MESSAGES-CANONICAL` on deploy, and only reaches as far back as retention; anything older needs a reindex from the authoritative store. Review also raised a second reason to version it: with a shared name, an old pod restarting mid-rollout flips the filter back.

I have not changed the durable name here, because that would attach a stream replay of a size I can't state to a correctness fix, and CLAUDE.md assigns stream lifecycle to ops/IaC. It's a two-line change and I'll add it on request — see [the thread](https://github.com/hmchangw/newchat/pull/462#discussion_r3909915361).

https://claude.ai/code/session_01UxZYPse6BHHNiL53bcP6py

## Summary by CodeRabbit

* **New Features**
  * Added support for matching canonical bot message events by site.
  * Added validation to ensure message consumers use filters covered by their streams.

* **Bug Fixes**
  * Corrected bot message filtering to use the appropriate event subjects.
  * Bare system messages are now safely ignored instead of being treated as invalid events.
  * Workers now refuse to start when a consumer filter cannot match its deployed stream, preventing silent indexing gaps.


julianshen:
## Needs clarification

Will deployment recreate or version the bot-message-sync durable to backfill retained events?
Reply in this thread, or post: `answer clar_rdgubefm7us3br4pmm6zzgehic: <your answer>`

---

_🤖 Posted by [tGDBot](https://github.com/julianshen/tGDBot)_

julianshen:

### Disputed

- `search-sync-worker/messages.go:89` (`nats`) — The corrected filter retains the existing `bot-message-sync` durable. Its cursor has already advanced while the disjoint filter selected no bot events, so updating the filter does not replay retained history and previously published bot messages remain absent from search. During a rolling deployment, a restarted old pod can also update the shared durable back to the old filter; bot events published before a new pod restores it are skipped past permanently. Use a versioned durable with an explicitly planned replay, or coordinate stopping all old workers and resetting/reindexing the consumer from the authoritative store.
- `search-sync-worker/messages.go:89` (`tgd-review`) — The corrected collection retains the existing `bot-message-sync` durable. During a rolling deployment, an old pod starting after a new one can update that same durable back to the disjoint `chat.msg...` filter. Bot events published while it is active are skipped, and restoring the corrected filter does not rewind the consumer cursor, leaving those messages permanently absent from search. The unchanged cursor also provides no historical backfill. Version the durable with coordinated replay and cleanup, or require all old instances to stop before any new instance starts and perform an explicit reindex.

### Local review memory

- `search-sync-worker/messages.go:89` — @hmchangw: Agreed on the mechanism — it's the "Deploying this does not backfill" section of the PR description, from pkg/stream/consumer.go:45-47 . Not fixed in code, and I want to be explicit about why rather than quietly leave it. **The rolling-de… ([thread](https://github.com/hmchangw/newchat/pull/462#discussion_r3909778545))
- `search-sync-worker/messages.go:89` — @hmchangw: Closing the loop on this one: the decision is **keep the existing bot-message-sync durable name**, made by the repo owner after the trade-off was laid out. So it ships unversioned and this is not pending action. Recording the accepted co… ([thread](https://github.com/hmchangw/newchat/pull/462#discussion_r3909778545))

<details>

* `pkg/subject/bot.go`
* `pkg/subject/bot_test.go`
* `search-sync-worker/subjectoverlap.go`
* `search-sync-worker/subjectoverlap_test.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

julianshen (commented):
### tGDBot review of `afc6e7b` — 5 findings in 5 files

2 🔴 blocking · 3 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/462#issuecomment-5502693970), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

search-sync-worker/messages.go
```diff
@@ -61,22 +67,24 @@ type messageCollection struct {
...
-		devMode:      devMode,
-		streamCfg:    botMessagesStreamCfg,
-		consumerName: "bot-message-sync",
+		indexPrefix:    indexPrefix,
+		devMode:        devMode,
+		streamCfg:      botMessagesStreamCfg,
+		filterSubjects: botMessagesFilter,
+		consumerName:   "bot-message-sync",
 	}
 }
```
julianshen, line 89: _🏷️ delivery/recovery_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`, `elasticsearch`, `nats` (3 rules)_

**Reusing the durable leaves all historical bot messages unindexed**

The corrected collection still uses the existing `bot-message-sync` durable. That durable has already scanned the stream while carrying the disjoint `chat.msg...` filter, so updating its filter does not reset its cursor or replay earlier bot/system events. After deployment, only newly published events are indexed; every retained event before the cursor remains absent and search continues returning materially incomplete results. Use a versioned durable configured with `DeliverAll`, or stop old workers and delete/recreate the durable as a coordinated deployment step. A versioned durable also prevents an old binary during a rolling deployment from updating the shared durable back to the incorrect filter.

<details>

- **`elasticsearch`**:85: Existing deployments retain the unchanged `bot-message-sync` durable and its current stream cursor when this filter is updated. Consequently, all bot and system-message events published before rollout remain absent from the Elasticsearch bot-message index, so searches for those retained messages still return materially incomplete results. Coordinate a migration that recreates or versions the durable with `DeliverAll` before rollout, and reindex from the authoritative store for events no longer covered by stream retention.
- **`nats`**:87: The corrected filter is applied while retaining the existing `bot-message-sync` consumer name. On an existing deployment, updating that durable's filter does not rewind its JetStream cursor; after the old nonmatching filter advanced past bot events, only newly published events will be indexed. Thus searches remain incomplete for all retained historical bot and system messages. Coordinate deletion/reset of this durable before rollout, or introduce a versioned durable that starts with `DeliverAll` and clean up the old durable afterward.

</details>

<details>

```
In `search-sync-worker/messages.go` around line 89: The corrected collection still uses the existing `bot-message-sync` durable. That durable has already scanned the stream while carrying the disjoint `chat.msg...` filter, so updating its filter does not reset its cursor or replay earlier bot/system events. After deployment, only newly published events are indexed; every retained event before the cursor remains absent and search continues returning materially incomplete results. Use a versioned durable configured with `DeliverAll`, or stop old workers and delete/recreate the durable as a coordinated deployment step. A versioned durable also prevents an old binary during a rolling deployment from updating the shared durable back to the incorrect filter.

```

</details>

---

search-sync-worker/messages.go
```diff
@@ -120,15 +129,29 @@ func (c *messageCollection) ConsumerName() string {
...
+// userMessagesFilter selects the single-token message events on
+// MESSAGES-CANONICAL (created/updated/deleted/...).
+func userMessagesFilter(siteID string) []string {
 	return []string{subject.MsgCanonicalMessageWildcard(siteID)}
 }
 
+// botMessagesFilter is the same shape on BOT-MESSAGES-CANONICAL.
+func botMessagesFilter(siteID string) []string {
+	return []string{subject.BotCanonicalMessageWildcard(siteID)}
+}
```
julianshen, line 148: _🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**The bot filter consumes an incompatible system-message payload shape.**

This filter now receives `bot-room-service/sysmsg.go` publications on `chat.bot.canonical.{site}.created`, but that publisher sends a raw `model.Message` while `messageCollection.BuildAction` expects `model.MessageEvent`. Unmarshalling succeeds with an empty nested message, then fails the missing-ID validation; `Handler.AddWithContext` logs and permanently ack-drops it as poison. Thus the system messages cited in the PR remain unindexed and produce false poison telemetry. Normalize the sysmsg publisher to the shared `MessageEvent` envelope, or explicitly handle this raw shape, and add a test for it.

<details>

```
In `search-sync-worker/messages.go` around line 148: This filter now receives `bot-room-service/sysmsg.go` publications on `chat.bot.canonical.{site}.created`, but that publisher sends a raw `model.Message` while `messageCollection.BuildAction` expects `model.MessageEvent`. Unmarshalling succeeds with an empty nested message, then fails the missing-ID validation; `Handler.AddWithContext` logs and permanently ack-drops it as poison. Thus the system messages cited in the PR remain unindexed and produce false poison telemetry. Normalize the sysmsg publisher to the shared `MessageEvent` envelope, or explicitly handle this raw shape, and add a test for it.

```

</details>

---

search-sync-worker/subjectcover.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+		if pt == "*" {
+			// One token against one token. A filter token of `>` would select a
+			// whole subtree, which a single-token wildcard does not cover.
+			if f[i] == ">" {
+				return false
+			}
+			continue
+		}
+		if pt != f[i] {
```
julianshen: _🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**Containment is stricter than the no-match startup invariant.**

The check rejects valid filters that overlap a stream without being wholly contained by one stream pattern. For example, stream subject `chat.events.created` and consumer filter `chat.events.*` do consume a message, but this literal-versus-wildcard comparison returns false and startup reports that the consumer "would consume nothing." If the invariant is meant to detect disjoint filters as documented, test pattern intersection/collision rather than containment. If full containment is intentional, rename the invariant and error so operators are not given a false diagnosis.

<details>

```
In `search-sync-worker/subjectcover.go` around line 46: The check rejects valid filters that overlap a stream without being wholly contained by one stream pattern. For example, stream subject `chat.events.created` and consumer filter `chat.events.*` do consume a message, but this literal-versus-wildcard comparison returns false and startup reports that the consumer "would consume nothing." If the invariant is meant to detect disjoint filters as documented, test pattern intersection/collision rather than containment. If full containment is intentional, rename the invariant and error so operators are not given a false diagnosis.

```

</details>

---

hmchangw: Confirmed and fixed in 319829a — thank you, the counter-example is exact.

`subjectCovers("chat.events.created", "chat.events.*")` returned `false`, and I verified it does before changing anything. That config consumes fine, so the invariant would have refused to start a working service while reporting the opposite of the truth — worse than the silent bug it was added to catch.

The predicate is now a symmetric intersection (`subjectsIntersect`), and the file and error message were renamed to say what they decide. Two cases flipped to `true` as a result and are now in the table: a literal stream leaf against a `*` filter, and a `*` stream subject against a `>` filter. The disjoint cases — including the `chat.bot.…` vs `chat.msg.…` case this PR exists for — still fail, and every case is asserted in both directions to hold symmetry.

---

search-sync-worker/messages.go
hmchangw: Confirmed, and it corrects a claim in my PR description. Fixed in 319829a.

Verified the chain rather than taking it on faith: `MessageEvent.Message` is `json:"message"`, a bare `model.Message` has no such key, so it decodes to a zero event; `BuildAction` then returns `missing message id`; `handler.go:126-131` treats every `BuildAction` error as poison — `slog.Error` + Ack + `dispAckedPoison`. I reproduced exactly that error in a test before touching the code.

One correction to your remedy, though. Normalizing the publisher would **not** get these indexed: `BuildAction` skips `model.IsSystemMessageType` deliberately — *"UI chrome, not searchable content."* So the only real cost here is false poison telemetry, which matters because that metric otherwise means data loss. My PR description claimed the missing indexed content included system messages; that was wrong and I've corrected it.

The fix is therefore to account for them as **filtered**, not to make them indexable. A bare message of any *other* type stays poison — that would be a publisher that skipped the envelope, and it should stay loud.

I did **not** normalize `bot-room-service/sysmsg.go`, deliberately. The same mismatch hits `bot-message-worker`, which decodes `MessageEvent` from this subject and passes `evt.Message` to `SaveMessage` unconditionally — so today these sysmsgs reach Cassandra as empty rows. Fixing the envelope there would start persisting membership sysmsgs into message history for the first time, which is a product decision about what belongs in history, not a side effect a search-indexing PR should ship. It's a real pre-existing bug and worth its own change.

---

search-sync-worker/messages.go
hmchangw: Agreed on the mechanism — it's the "Deploying this does not backfill" section of the PR description, from `pkg/stream/consumer.go:45-47`. Not fixed in code, and I want to be explicit about why rather than quietly leave it.

**The rolling-deployment half is a genuine addition and I had missed it.** With a shared durable name, an old pod that restarts mid-rollout calls `CreateOrUpdateConsumer` with the old filter and flips it back. Narrow, but real, and it argues for versioning independently of backfill.

**What I'm not willing to do unilaterally is make the deploy replay the stream.** A new durable name starts at `DeliverAll`, so the first pod to come up re-reads all retained `BOT-MESSAGES-CANONICAL` and re-indexes it. I don't know this deployment's retention or stream volume, so I can't size that burst, and CLAUDE.md puts stream lifecycle with ops/IaC. Shipping a rename means a correctness fix silently carries a replay whose cost I can't state — that decision belongs to whoever owns the deploy, not to this PR.

The two halves also don't need the same answer. If the rollout ordering is the concern, versioning fixes it cheaply. If backfill is the goal, a versioned durable only reaches as far back as retention, so for anything older the answer is a reindex from the authoritative store regardless — which is a migration, not a consumer config.

So: happy to add a versioned durable name plus removal of the old one as a follow-up commit here if that's the call, and it's a two-line change. I'd rather have it be a decision than an assumption.

---

hmchangw:
answer clar_rdgubefm7us3br4pmm6zzgehic: As written, no — the PR keeps the existing `bot-message-sync` durable, so deploying it indexes from the current cursor forward and backfills nothing. That is stated in the "Deploying this does not backfill" section rather than left implicit, but it is a deliberate non-decision, not a settled one: recreating or versioning the durable makes the deploy replay all retained `BOT-MESSAGES-CANONICAL`, and I can't size that burst without knowing this deployment's retention and volume — and CLAUDE.md assigns stream lifecycle to ops/IaC. Whoever owns the deploy should decide between (a) ship as-is and accept a forward-only fix, (b) version the durable so it starts at `DeliverAll` and backfills to the retention horizon, or (c) (b) plus a reindex from the authoritative store for anything older than retention. I've offered the two-line change for (b) on the review thread and will add it on request. Full reasoning: https://github.com/hmchangw/newchat/pull/462#discussion_r3909915361

---

hmchangw:
@julianshen check latest

---

julianshen (commented):
### tGDBot review of `319829a` — 1 finding in 5 files

1 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/462#issuecomment-5502693970), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

search-sync-worker/messages.go
```diff
@@ -208,6 +241,17 @@ func (c *messageCollection) BuildAction(data []byte) ([]searchengine.BulkAction,
...
+// isBareSystemMessage reports whether data is a bare model.Message carrying a
+// system message type, rather than the MessageEvent envelope. A bare message of
+// any OTHER type is still a publisher bug worth surfacing, so it stays poison.
+func isBareSystemMessage(data []byte) bool {
+	var m model.Message
+	if err := json.Unmarshal(data, &m); err != nil {
+		return false
+	}
+	return m.ID != "" && model.IsSystemMessageType(m.Type)
+}
```
julianshen, line 253: _🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**The bare-message check also hides malformed envelopes**

`json.Unmarshal` into `model.Message` ignores the envelope's unknown `message` field. Consequently, `{"id":"m","type":"members_added","message":{}}` first produces an empty `evt.Message`, then passes this helper and is silently treated as filtered, although the stated behavior keeps non-bare empty envelopes as poison. Confirm that the `message` key is absent before accepting the payload as bare, and add this malformed-envelope case to the test.

<details>

> ‼️ **IMPORTANT**

```suggestion
func isBareSystemMessage(data []byte) bool {
	var shape struct {
		Message json.RawMessage `json:"message"`
	}
	if err := json.Unmarshal(data, &shape); err != nil || shape.Message != nil {
		return false
	}

	var m model.Message
	if err := json.Unmarshal(data, &m); err != nil {
		return false
	}
	return m.ID != "" && model.IsSystemMessageType(m.Type)
}
```

</details>

<details>

```
In `search-sync-worker/messages.go` around line 247: `json.Unmarshal` into `model.Message` ignores the envelope's unknown `message` field. Consequently, `{"id":"m","type":"members_added","message":{}}` first produces an empty `evt.Message`, then passes this helper and is silently treated as filtered, although the stated behavior keeps non-bare empty envelopes as poison. Confirm that the `message` key is absent before accepting the payload as bare, and add this malformed-envelope case to the test.

```

</details>

---

hmchangw:
`sast` is failing on `671f37a`, and it is not this PR's.

**What is failing.** `govulncheck` only — the job summary reads `gosec=PASS govulncheck=FAIL semgrep=PASS rule-tests=PASS`. One advisory:

```
Vulnerability #1: GO-2026-6348
    Heap Memory Exhaustion (OOM) via HTTP/2 DATA Frame Fragmentation in google.golang.org/grpc
  Module: google.golang.org/grpc
    Found in: google.golang.org/grpc@v1.82.1
    Fixed in: google.golang.org/grpc@v1.83.1
    Example traces: pkg/natsmetrics/metrics.go:478:19 → sync.Once.Do → transport.NewHTTP2Client
```

**Why it is not this PR's.** The branch touches `pkg/subject/{bot,bot_test}.go` and five files under `search-sync-worker/` — no `go.mod`, no `pkg/natsmetrics`. `google.golang.org/grpc v1.82.1` is an indirect dependency already pinned on `main`, and the reported call path runs through `pkg/natsmetrics`, which this branch does not modify. It is a newly published advisory against an existing dependency, so it fails every branch equally.

Confirmed on the base branch rather than inferred: the latest `main` push run ([35046729798](https://github.com/hmchangw/newchat/actions/runs/35046729798), commit `3e8eda2`) has **`sast` = failure with `test` and `lint` = success**, and the preceding `main` runs fail the same way.

**No fix exists to port.** No open PR touches `go.mod`. The patch is a one-line dependency bump plus `go mod tidy`:

```
go get google.golang.org/grpc@v1.83.1
go mod tidy
```

That is repo-wide — it unblocks `main` and every open PR, not just this one — so it belongs in its own change rather than inside a search-sync-worker fix, where it would be an unrelated dependency upgrade riding a correctness PR.

**No re-run.** The rule of thumb here is one re-run to confirm a failure is not the PR's, but `govulncheck` against a fixed `go.mod` is deterministic and the base-branch evidence above is stronger than a repeat would be. Re-running would only consume CI time.

This PR stays watched until it is green.

---

hmchangw:
## CI: `sast` was failing on `main`, not on this branch

The red `sast` check was **not introduced by this PR**. Evidence:

- `main` @ `3e8eda2` ([run 35046729798](https://github.com/hmchangw/newchat/actions/runs/35046729798)) fails the **same `sast` job at the same step** ("SAST (gosec + govulncheck + semgrep, fail on medium+)"). The last several `main` runs are all red.
- On this PR, `test`, `lint` and `test-integration (search-sync-worker)` all pass. Only `sast` was red, and its summary line reads `gosec=PASS govulncheck=FAIL semgrep=PASS rule-tests=PASS`.
- The finding is in `pkg/natsmetrics/metrics.go:478`, which this branch does not touch.

The single blocking finding:

```
GO-2026-6348 — Heap Memory Exhaustion (OOM) via HTTP/2 DATA Frame Fragmentation
Module: google.golang.org/grpc
  Found in: google.golang.org/grpc@v1.82.1
  Fixed in: google.golang.org/grpc@v1.83.1
```

No open PR bumps it, so rather than leave this PR parked behind a base-branch failure I've carried the fix here in `870c771`. `grpc` is an **indirect** dependency (via the OpenTelemetry exporters) and nothing in the repo imports it directly, so the change is one `go.mod` line plus its `go.sum` entries — no call sites. It no-ops once `main` takes the same bump.

Happy to split that commit into its own PR against `main` if you'd rather keep this one scoped to the bot-filter fix — say the word and I'll move it.

### Verification

| Gate | Result |
|---|---|
| `go build ./...` | clean |
| `make test` (full repo) | 124 packages, 0 failures |
| `make lint` | 0 issues |
| `make sast-gosec` | exit 0 |
| `make sast-vuln` | **could not run here** — the environment's gateway denies `vuln.go.dev` with a 403 (policy denial, confirmed via the proxy status endpoint) |

So govulncheck itself is unverified locally and **CI is the check** for this commit. The version was not guessed — v1.83.1 is what govulncheck's own output names as fixed.

---

search-sync-worker/messages.go
hmchangw: Confirmed and fixed in `c135ac2`. The finding is exactly right, and I reproduced it before changing anything.

`model.Message` has no `message` field, so `encoding/json` silently ignores that key. Your payload therefore decoded as a bare message and was Acked as filtered:

```
{"id":"m","type":"members_added","message":{}}
```

`{"message":null}` behaved the same way. Both are now failing-first test cases (`messages_test.go`), asserted to stay `missing message id`.

The sniff now checks for the key explicitly rather than inferring its absence from an empty decode:

```go
var envelope struct {
	Message json.RawMessage `json:"message"`
}
if err := json.Unmarshal(data, &envelope); err != nil || envelope.Message != nil {
	return false
}
```

Which is your suggested shape — thank you, I took it essentially as written.

Worth naming why this one mattered more than its severity suggests: this sniff is the only thing separating "silently dropped" from "poison alarm", and that alarm is what makes real data loss visible. Loosening it is the same class of bug as the disjoint filter this PR exists to fix — something going quietly missing with no signal. A bare message of any non-system type still stays poison, unchanged.

---

search-sync-worker/messages.go
hmchangw: Closing the loop on this one: the decision is **keep the existing `bot-message-sync` durable name**, made by the repo owner after the trade-off was laid out. So it ships unversioned and this is not pending action.

Recording the accepted consequences plainly, since they are real and the thread should not imply otherwise:

1. **No backfill.** The cursor has already advanced past every retained bot message, so indexing begins at deploy time. Recovering history needs a reindex from the authoritative store, which is a migration rather than a consumer config — and, as noted above, a versioned durable would only have reached back as far as retention anyway.
2. **The rolling-deploy window stays open.** An old pod restarting mid-rollout calls `CreateOrUpdateConsumer` with the disjoint filter and flips it back; messages published in that window are skipped permanently, because the cursor moves past them and the new pods do not rewind. Narrow, but real — worth a line in the deploy notes, and worth deploying this in a way that minimises old/new overlap.

Both remain a two-line change if that call is revisited later. Leaving this thread unresolved deliberately: it is a legitimate open concern that was accepted, not one that was addressed, and resolving it would misrepresent that.

---

hmchangw:
## CI update: the MinIO integration jobs are an infra failure, also present on `main`

On head `c135ac2`, `test-integration (pkg/minioutil)` and `test-integration (pkg/poolartifact)` fail. **Neither is this PR's.**

Every test in the package dies at *container creation*, before any test body runs:

```
testutil.MinIO: start minio: run minio: generic container: create container:
Error response from daemon: pull access denied for minio/minio,
repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

That is a Docker Hub image-pull denial for `minio/minio` on the runner — almost certainly anonymous-pull rate limiting, since the image plainly exists.

Evidence it is not this branch:

- **`main` @ `3e8eda2` fails identically.** [`test-integration (media-service)`](https://github.com/hmchangw/newchat/actions/runs/35046729798/job/104638135129) on the base branch, ~4 hours before my push, reports the same error verbatim: `prewarm shared containers: start minio: run minio: ... pull access denied for minio/minio`.
- The failure is at image pull, so no Go change can cause or cure it. The only repo-wide change on this branch is the `grpc` bump in `870c771`, and a module version cannot affect a `docker pull`.
- **`test-integration (search-sync-worker)` — the job that covers this PR's code — passes** on `c135ac2`, as does `roomlist-worker`, `portal-service` and `pkg/roommetacache`.

**No fix exists to port.** This needs registry credentials on the runner (a `docker login` step / pull-through cache), or pinning these images to a mirror — CI infrastructure, outside this PR's diff and outside what a service fix should carry.

I'll re-run the failed jobs once (GitHub only offers that after the run completes) in case it is transient rate limiting, and report back. I am not touching the tests: skipping or quarantining them to get green would hide a real capability gap.

Still pending on this head: `sast`, which should now pass with the `grpc` v1.83.1 bump — I'll confirm when it reports.

---

hmchangw:
## Re-run result: MinIO failure is persistent, not transient — and `sast` is now green

Reporting back on both things I said I'd confirm.

### `sast` — **passing** ✅

[Job 104684772258](https://github.com/hmchangw/newchat/actions/runs/35061991168/job/104684772258) on `c135ac2`. The `google.golang.org/grpc` v1.83.1 bump in `870c771` cleared GO-2026-6348. That was the only gate I could not verify locally (the authoring environment's gateway denies `vuln.go.dev`), so CI was the check — and it passed.

### MinIO — re-ran once, failed identically

I re-ran the failed jobs on run 35061991168. `pkg/minioutil` failed again with the same error on every test:

```
minio_integration_test.go:121: testutil.MinIO: start minio: ... pull access denied for minio/minio
```

Two failures on the same commit rules out transient Docker Hub rate limiting. **This is a persistent pull denial and needs a CI-infrastructure fix** — registry credentials on the runner (a `docker login` step), a pull-through cache, or mirroring the image. It is outside this PR's diff and outside what a service fix should carry, and it affects `main` and every open PR equally, not just this branch.

I have now spent the one re-run and will not re-run again — a third attempt would just be noise. I am also not skipping, quarantining or otherwise weakening these tests to force green; they are catching a genuine capability gap in CI.

### Where the PR stands on `c135ac2`

| | |
|---|---|
| `sast`, `test`, `lint`, `prewarm` | ✅ pass |
| `test-integration (search-sync-worker)` | ✅ pass — the job covering this PR's code |
| ~50 other integration jobs | ✅ pass |
| 5 MinIO jobs | ❌ infra (`pkg/minioutil`, `pkg/poolartifact`, `media-service`, `upload-service`, `client-update-service`) |
| Merge state | `blocked` — no conflict with `main` |

Everything within this PR's control is green. I'll keep watching until it merges or closes.

---

hmchangw:
## Status after the rebase: 4 of 5 checks green, `lint` red on a file this PR does not touch

Rebased onto `main` (`3ec5822`) and force-pushed — head is now `b0759ed`. The rebase **dropped the grpc bump commit as already-applied**, since main landed the identical bump in `24f10b3` (#497). The PR is now 4 commits with no `go.mod`/`go.sum` in its diff, scoped purely to `search-sync-worker` + `pkg/subject`.

A useful side effect: because `go.sum` is no longer modified, CI's affected-target detection now schedules only `test-integration (search-sync-worker)` instead of all 61 integration jobs. **The MinIO `pull access denied` problem is therefore no longer in play on this PR** — and it has in any case been fixed upstream (main's run [35072194523](https://github.com/hmchangw/newchat/actions/runs/35072194523) is fully green, `pkg/minioutil` included). My earlier comments calling it a persistent blocker are superseded.

### Current head `b0759ed`

| Check | Result |
|---|---|
| `prewarm` | ✅ |
| `test` | ✅ |
| `sast` | ✅ |
| `test-integration (search-sync-worker)` | ✅ |
| `lint` | ❌ |

### The `lint` failure is not from this PR's code

```
pkg/errcode/permanent_test.go:102:7  SA5011(related information): this check suggests that the pointer can be nil
pkg/errcode/permanent_test.go:105:10 SA5011: possible nil pointer dereference
```

Evidence:

- **Not in this PR's diff.** The diff is `pkg/subject/{bot.go,bot_test.go}` and `search-sync-worker/*`. `pkg/errcode` imports nothing this PR changes, so nothing here can affect its analysis.
- **Byte-identical to main.** `git diff 18d8784 origin/main -- pkg/errcode/permanent_test.go` is empty — the rebase did not touch this file.
- **`lint` passes on main** at the very commit this branch is rebased onto, and it passed on the pre-rebase head `c135ac2` with the same file content.
- **Not reproducible locally under any condition I could construct**: golangci-lint v2.11.4 (CI's pinned version), Go 1.25.13, golangci-lint cache cleared, a cold `GOCACHE`, and CI's exact invocation (`golangci-lint run`, no args). 0 issues every time.
- **Reproduces deterministically in CI** — I re-ran the failed jobs once and it failed identically. Both attempts logged a hit on the same restored cache: `golangci-lint.cache-Linux-2958-199ce439…`, and a re-run cannot invalidate it.

The finding itself is the classic SA5011 shape — `if ee == nil { t.Fatal(...) }` followed by `ee.Code` — which staticcheck only accepts when it models `(*testing.T).Fatal` as non-returning. A stale analysis-fact cache would explain both the CI failure and the local pass.

I have not edited `pkg/errcode` (outside this PR's scope) and have not added a `nolint` or weakened the lint gate to force green. Flagging it for a decision rather than papering over it.

---

hmchangw:
## CI is green on `8d483b5`

Rebased onto `main` `ae81189`, which carries #500 — the one-line SA5011 fix to `pkg/errcode/permanent_test.go`. That cleared the lint failure.

| Check | Result |
|---|---|
| `prewarm` | ✅ |
| `lint` | ✅ |
| `sast` | ✅ |
| `test` | ✅ |
| `test-integration (search-sync-worker)` | ✅ |

[Run 35184026348](https://github.com/hmchangw/newchat/actions/runs/35184026348).

This also settles the open question from my previous comment. I said there that #500's own CI could not demonstrate the fix, because `main` lints clean on that file either way, and that the real test would be this PR going green after a rebase. It did — so the cause was the nil-check-then-dereference pattern being fragile to the analyzer's environment, not the stale-Actions-cache theory I also floated. Recording that explicitly since I raised both possibilities.

Nothing outstanding on my side. The PR is 4 commits scoped to `search-sync-worker` + `pkg/subject`, no merge conflict, and the remaining gate is human review.

One thread is deliberately left unresolved — [the durable-rename discussion](https://github.com/hmchangw/newchat/pull/462#discussion_r3909778545). The decision was to keep the `bot-message-sync` name, and its two accepted consequences are recorded there: no backfill of pre-deploy bot messages, and a narrow mixed-version rollout window where an old pod can restore the disjoint filter. Both are worth a line in the deploy notes.

---

search-sync-worker/subjectoverlap.go
```diff
@@ -0,0 +1,139 @@
...
+	subjects, err := lookup(ctx, streamName)
+	if err != nil || len(subjects) == 0 {
+		slog.WarnContext(ctx, "could not read deployed stream subjects; checking filters against the local declaration",
+			"stream", streamName,
+			"consumer", consumerName,
+			"declared", declared,
+			"error", err,
+		)
+		subjects = declared
+	}
```
julianshen: _🏷️ error-handling_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`silent-failure`_

**A failed stream lookup disables the startup invariant**

When the deployed-subject lookup fails or returns no subjects, this substitutes the local declaration and can return success. If the deployed stream has drifted to disjoint subjects, a transient lookup failure therefore lets JetStream create the inert consumer, after which the worker appears healthy while indexing nothing. Propagate lookup and empty-result failures so startup cannot claim the filter was verified.

<details>

> ‼️ **IMPORTANT**

```suggestion
	if err != nil || len(subjects) == 0 {
		slog.ErrorContext(ctx, "could not read deployed stream subjects",
			"stream", streamName,
			"consumer", consumerName,
			"declared", declared,
			"error", err,
		)
		if err != nil {
			return fmt.Errorf("read deployed subjects for stream %q: %w", streamName, err)
		}
		return fmt.Errorf("stream %q has no deployed subjects", streamName)
	}
```

</details>

<details>

```
In `search-sync-worker/subjectoverlap.go` around line 127: When the deployed-subject lookup fails or returns no subjects, this substitutes the local declaration and can return success. If the deployed stream has drifted to disjoint subjects, a transient lookup failure therefore lets JetStream create the inert consumer, after which the worker appears healthy while indexing nothing. Propagate lookup and empty-result failures so startup cannot claim the filter was verified.

```

</details>

---

search-sync-worker/subjectoverlap.go
```diff
@@ -0,0 +1,139 @@
...
+// against filters built from the same constants can only restate a compile-time
+// fact. Ops narrowing the real subjects is the one drift that happens in
+// production, and only a live read can see it.
+//
+// Falls back to the declared subjects when the stream cannot be read — it may not
+// exist yet, and consumer creation reports that far better than this check can.
+// Refusing to start there would turn a silent-indexing bug into an outage.
+func preflightFilters(ctx context.Context, lookup streamSubjectsFunc, streamName string, declared, filters []string, consumerName string) error {
+	subjects, err := lookup(ctx, streamName)
+	if err != nil || len(subjects) == 0 {
```
julianshen: _🎯 operational-correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`elasticsearch`_

**Lookup failures bypass the deployed-subject invariant**

When deployed stream lookup fails or returns no subjects, this falls back to the local declaration. For an externally owned stream whose deployed subjects have drifted, credentials may permit consumer creation but deny stream-info lookup; the stale declaration then passes and an inert consumer is created, leaving that collection's Elasticsearch index silently incomplete. Fail startup when deployed subjects cannot be read, or restrict fallback to service-owned streams that were successfully reconciled in this startup.

<details>

```
In `search-sync-worker/subjectoverlap.go` around line 129: When deployed stream lookup fails or returns no subjects, this falls back to the local declaration. For an externally owned stream whose deployed subjects have drifted, credentials may permit consumer creation but deny stream-info lookup; the stale declaration then passes and an inert consumer is created, leaving that collection's Elasticsearch index silently incomplete. Fail startup when deployed subjects cannot be read, or restrict fallback to service-owned streams that were successfully reconciled in this startup.

```

</details>

---

julianshen (commented):
### tGDBot review of `09c242c` — 2 findings in 7 files

1 🔴 blocking · 1 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/462#issuecomment-5502693970), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

search-sync-worker/subjectoverlap.go
hmchangw: Confirmed and fixed in `27df464`. The finding is right, and the interesting part is that my own justification for the fallback does not survive checking.

The comment on that function read:

> Falls back to the declared subjects when the stream cannot be read — it may not exist yet, and consumer creation reports that far better than this check can. Refusing to start there would turn a silent-indexing bug into an outage.

That reasoning is wrong on its own terms. `main.go` calls `CreateOrUpdateConsumer` immediately after the preflight and `os.Exit(1)`s on error, so an absent stream aborted startup **either way** — the fallback never prevented the outage it was defending against. What it actually did was cover the *other* case: the stream exists but its info is unreadable (denied credentials, a blip), which is exactly where the local declaration is evidence of nothing.

And the substitution could never fail, which is the part that makes this a real hole rather than a rough edge. The filters are built from the same `pkg/subject` constants as `streamCfg.Subjects`, so checking one against the other is a compile-time tautology. A transient lookup failure against a stream whose deployed subjects **had** drifted would therefore pass, create the inert consumer, and log the filter as verified — the precise failure this guard exists to catch, reintroduced by its own error path.

The fix propagates instead:

```go
subjects, err := lookup(ctx, streamName)
if err != nil {
    return fmt.Errorf("consumer %q: read deployed subjects for stream %q: %w", consumerName, streamName, err)
}
if len(subjects) == 0 {
    return fmt.Errorf("consumer %q: stream %q reports no deployed subjects", consumerName, streamName)
}
```

Close to your suggested shape. Two deliberate differences: the cause is wrapped rather than logged-and-replaced, because the operator's next move depends on *why* the read failed — absent stream, denied permission, or a blip — and there is now a test asserting `errors.Is` holds through it; and the `declared` parameter is removed rather than left unused, so the signature says that the function no longer consults the declaration at all. `main.go` still logs `declaredSubjects` as operator context, and its log message moved from `"consumer filter does not match its stream"` to `"consumer filter preflight failed"`, which is true of both outcomes now.

TDD as usual: both expectations were written first, failed to compile against the old signature, then failed on behaviour once it changed.

Thanks — this was the better catch of the two, because it was the guard silently disabling itself.

---

search-sync-worker/subjectoverlap.go
hmchangw: Same root cause as [the blocking finding](https://github.com/hmchangw/newchat/pull/462#discussion_r4040455946) and fixed by the same change in `27df464` — the lookup failure and the empty-subject case now both return an error instead of substituting the local declaration. Full reasoning is on that thread; the short version is that the substituted check could never fail, because the filters are built from the same constants as the declaration.

Your framing added the sharpest version of the scenario, so it is worth recording: **credentials that permit consumer creation but deny `STREAM.INFO`**. That one is not hypothetical the way a transient blip is — it is a stable configuration that would have made the guard permanently inert for INBOX and HR, the two externally-owned streams the live read was added for in the first place, while every pod reported healthy. The blip case self-heals on restart; yours does not.

On your second option — *restrict fallback to service-owned streams that were successfully reconciled in this startup* — I considered it and did not take it. It is sound, but it buys nothing here: `bootstrap.go` runs `CreateOrUpdateStream` **before** the preflight, so for a service-owned stream with `BOOTSTRAP_STREAMS=true` the stream provably exists by the time we look it up. If the read-back then fails, NATS is broken enough that failing is the right answer rather than a case to special-case. The dev ergonomics that fallback would have protected are already covered by bootstrap creating the stream first.

So it is the unconditional version, and the conditional stays unwritten rather than written and unexercised.

---

hmchangw:
Not needed. Implemented elsewhere.

PR #465, opened 2026-09-02
feat(roomsubcache): carry IsSubscribed on Member, gate botDM fan-out on it

hmchangw:
## What

Adds `IsSubscribed` to `roomsubcache.Member`, gates `broadcast-worker`'s botDM fan-out on it, and busts the shared cache entry from `user-service` when the flag is written.

## Why

`broadcast-worker`'s DM fan-out needs that flag. The alternative — moving `publishDMEvents` off `ListRoomMembers` and onto a direct `subscriptions` query — would give up the resilience #188 introduced: a warm `roomsubcache` entry keeps DM fan-out serving a room while Mongo is down. Widening the projection keeps that property and is a far smaller change than swapping the read.

`isSubscribed` is safe to serve from a TTL'd entry because it is **stable** — an explicit subscribe/unsubscribe flips it. The per-message read state (`lastSeenAt` / `hasMention` / `threadUnread`) is not, and deliberately stays off `Member`: `roomlist-worker` rewrites those on every message, so a cached copy would be wrong within milliseconds.

## Changes

**Cache layer**

- `Member` gains `IsSubscribed bool` (`json:"isSubscribed,omitempty"`), projected and decoded by `NewMongoLoader`.
- `cacheKeySchemaVersion` **v4 → v5**. A v4 entry decodes cleanly under the v5 shape, which is precisely the hazard — every member would read as unsubscribed for a full TTL with nothing to signal it. The version segment makes those entries miss and refill from Mongo.
- `legacyCacheKeySchemaVersion` (a single string) becomes `legacyCacheKeySchemaVersions` (`{"v4", "v3"}`), and `Invalidate` busts every generation via a new `allGenerationKeys` helper.

That last one is the non-obvious part. A rolling v4→v5 deploy still has v4 binaries serving from their own key. If a bust cleared only v5 and v3, a membership change would leave those v4 binaries serving a stale member list until the TTL expired. `valkeyutil.BustKeys` is already variadic, so this costs one more pipelined `DEL`.

**Consumer**

- `broadcast-worker/publishDMEvents` skips a member whose botDM row is unsubscribed. This mirrors `user-service`'s `activeSubscriptionFilter` (`roomType == "botDM"` requires `isSubscribed`), which hides the room from `subscription.list` — delivering a live event would resurrect a room the user removed.
- The gate keys off the **row's own** `RoomType`, never the room's. Only the human side of a botDM carries the flag (`buildDMPairSubs` stores the type each counterpart implies), and plain DM rows never set `isSubscribed` — it is `bson:",omitempty"` and only app subscriptions write it. A gate that ignored `RoomType` would therefore silence every human DM; a test case pins that.

**Invalidation**

`ROOMSUBCACHE_TTL` defaults to **90m**, and nothing invalidated this tier on an app subscribe/unsubscribe — only mute changes and member add/remove do. The gate alone would mean a re-subscribed user receives no live botDM events for up to 90 minutes. So `user-service.SetAppSubscription` now busts the room's entry on both write paths.

- Invalidation-only, so `user-service` needs no `ROOMSUBCACHE_TTL` of its own — it never fills this tier, and adding the knob would only create a way for it to drift from its peers.
- Wired as an optional `service.Option`; absent when `VALKEY_ADDRS` is empty, matching how the badge cache degrades.

## Docs

`docs/client-api.md` and its derived `docs/client-api/events.md` view both state the botDM fan-out rule ("published once per non-bot member"). That rule is now narrower, so both record the unsubscribed-member skip. No request/response schema changed and `pkg/model` is untouched.

## Blast radius

The schema bump invalidates cached entries for every reader of the shared key — `broadcast-worker` and `notification-worker`. That is a one-time cold refill from Mongo per room after deploy, not a correctness issue.

`legacyCacheKeySchemaVersions` grows by one generation per bump and is pruned by hand. `v3` should be dropped once no v3 binary can be running; until then every invalidation pays three `DEL`s.

## Verification

- `make test` — full suite, race detector, 0 failures
- `make lint` — 0 issues
- `gosec` — PASS
- `make sast-semgrep-test` — repo-owned rules pass against their fixtures (2/2)
- `semgrep --config=p/golang` — **could not run**; the sandbox proxy blocks `semgrep.dev` (403)
- `govulncheck` — **could not run**; the sandbox proxy blocks `vuln.go.dev` (403)
- Loader integration test extended (`TestMongoLoader_HistorySharedSince`) to assert `isSubscribed` is projected and that an absent field decodes `false`. **Not executed** — integration tests need Docker, unavailable here. Worth a run before merge.

Both changes were written test-first. The fan-out gate's RED run failed on exactly one of three subtests (the unsubscribed case delivering when it should not), with the subscribed and plain-DM guard cases already passing. The invalidation test was staged: the option was added as plumbing only so the failure was the assertion (`Invalidate` never called), not a compile error.

https://claude.ai/code/session_013UohPLkZS1BXUoKP9mzch3

### DIFF
--- a/pkg/roomsubcache/loader_mongo.go
+++ b/pkg/roomsubcache/loader_mongo.go
@@ -21,6 +21,7 @@ var memberFindOpts = options.Find().SetProjection(bson.M{
 	"roomType":           1,
 	"muted":              1,
 	"historySharedSince": 1,
+	"isSubscribed":       1,
 })
 
 // homeSiteFindOpts projects the users lookup down to the two fields the
@@ -62,6 +63,7 @@ func NewMongoLoader(subscriptions, users *mongo.Collection) Loader {
 				RoomType           model.RoomType `bson:"roomType"`
 				Muted              bool           `bson:"muted"`
 				HistorySharedSince *time.Time     `bson:"historySharedSince"`
+				IsSubscribed       bool           `bson:"isSubscribed"`
 			}
 			if err := cur.Decode(&doc); err != nil {
 				return nil, fmt.Errorf("decode subscription: %w", err)
@@ -78,6 +80,7 @@ func NewMongoLoader(subscriptions, users *mongo.Collection) Loader {
 				IsBot:              doc.User.IsBot,
 				Muted:              doc.Muted,
 				HistorySharedSince: hssMs,
+				IsSubscribed:       doc.IsSubscribed,
 			})
 		}
 		if err := cur.Err(); err != nil {
--- a/pkg/roomsubcache/roomsubcache.go
+++ b/pkg/roomsubcache/roomsubcache.go
@@ -65,6 +65,12 @@ type Member struct {
 	// pre-fix entries (whose siteId carried the room's site) miss instead of
 	// misrouting the RPC forever.
 	HomeSiteID string `json:"homeSiteId,omitempty"`
+	// IsSubscribed is the subscription's own flag, denormalised for the fan-out
+	// path. Safe to serve from a TTL'd entry because it is stable: an explicit
+	// subscribe/unsubscribe flips it, unlike lastSeenAt/hasMention/threadUnread,
+	// which roomlist-worker rewrites on every message and which therefore must
+	// never be cached here.
+	IsSubscribed bool `json:"isSubscribed,omitempty"`
 }
 
 // Cache stores and retrieves a room's member list.
@@ -152,20 +158,32 @@ func NewValkeyCache(client valkeyutil.Client, opts ...Option) Cache {
 // the Entry{Members, CachedAt} envelope. That is a change of JSON kind, array
 // to object, so reusing v3 would have each binary in a rolling deploy handed
 // the other's shape — an unmarshal error, not a silent zero value.
-const cacheKeySchemaVersion = "v4"
+//
+// Bumped to v5 when IsSubscribed was added. A v4 entry decodes cleanly under
+// the v5 shape, which is precisely the danger: every member would read as
+// unsubscribed for the full TTL with nothing to signal it.
+const cacheKeySchemaVersion = "v5"
 
-// legacyCacheKeySchemaVersion is the pre-envelope generation. Only invalidation
-// touches it: while a rolling deploy can still have a v3 binary live, a bust
-// must clear both or that binary keeps serving a member list for a room whose
-// membership just changed. Drop this once no v3 binary can be running.
-const legacyCacheKeySchemaVersion = "v3"
+// legacyCacheKeySchemaVersions are the superseded generations, newest first.
+// Only invalidation touches them: a rolling deploy can still have a binary of
+// an older generation live, and a bust must clear that generation's key too or
+// that binary keeps serving a member list for a room whose membership just
+// changed. Drop an entry once no binary of that generation can be running.
+var legacyCacheKeySchemaVersions = []string{"v4", "v3"}
 
 func cacheKey(roomID string) string {
 	return "room:" + cacheKeySchemaVersion + ":" + roomID + ":subs"
 }
 
-func legacyCacheKey(roomID string) string {
-	return "room:" + legacyCacheKeySchemaVersion + ":" + roomID + ":subs"
+// allGenerationKeys returns the current key followed by every superseded one,
+// in the order Invalidate deletes them.
+func allGenerationKeys(roomID string) []string {
+	keys := make([]string, 0, 1+len(legacyCacheKeySchemaVersions))
+	keys = append(keys, cacheKey(roomID))
+	for _, v := range legacyCacheKeySchemaVersions {
+		keys = append(keys, "room:"+v+":"+roomID+":subs")
+	}
+	return keys
 }
 
 // Get returns the cached member list for roomID. On absence it returns
@@ -240,10 +258,10 @@ func (c *valkeyCache) Set(ctx context.Context, roomID string, members []Member, 
 	return nil
 }
 
-// Invalidate removes the cached entry for roomID, in both key generations —
-// see legacyCacheKeySchemaVersion. Those keys carry no hash tag and so land in
-// different cluster slots; the client pipelines one DEL per key, so neither
-// delete can take the other down with a CROSSSLOT rejection.
+// Invalidate removes the cached entry for roomID in every key generation — see
+// legacyCacheKeySchemaVersions. Those keys carry no hash tag and so land in
+// different cluster slots; the client pipelines one DEL per key, so no delete
+// can take another down with a CROSSSLOT rejection.
 //
 // Best-effort, like every other tier's invalidation: a bust runs after the
 // authoritative write has committed, and BustKeys is what strips the caller's
@@ -253,5 +271,5 @@ func (c *valkeyCache) Invalidate(ctx context.Context, roomID string) {
 	if roomID == "" {
 		return
 	}
-	valkeyutil.BustKeys(ctx, c.client, "roomsubcache", cacheKey(roomID), legacyCacheKey(roomID))
+	valkeyutil.BustKeys(ctx, c.client, "roomsubcache", allGenerationKeys(roomID)...)
 }
--- a/pkg/roomsubcache/loader_mongo_integration_test.go
+++ b/pkg/roomsubcache/loader_mongo_integration_test.go
@@ -19,8 +19,9 @@ import (
 // reader of the shared cache key depends on: the loader maps the nested
 // subscription shape into the flat Member, carries Muted through, and converts
 // historySharedSince (a Date) to epoch-ms — leaving it nil for full-access
-// members. A loader that dropped either field would silently unmute users and
-// widen their history windows for whichever service reads the entry next.
+// members, and carries IsSubscribed. A loader that dropped any of these would
+// silently unmute users, widen their history windows, or mark every member
+// unsubscribed for whichever service reads the entry next.
 func TestMongoLoader_HistorySharedSince(t *testing.T) {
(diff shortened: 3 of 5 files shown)


julianshen:
**3 findings · 2 unique issues · 2 inline comments posted** — 🔴 2 blocking · 🟠 1 warning

> [!NOTE]

### Related work

- [PR #188](https://github.com/hmchangw/newchat/pull/188) — MongoDB outage survival for send \+ history load, and extract broadcast-worker's room-level writes into roomlist-worker (merged)

<details>

* `pkg/roomsubcache/loader_mongo.go`
* `pkg/roomsubcache/loader_mongo_integration_test.go`
* `pkg/roomsubcache/metrics_test.go`
* `pkg/roomsubcache/roomsubcache_test.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

pkg/roomsubcache/roomsubcache.go
```diff
@@ -253,5 +271,5 @@ func (c *valkeyCache) Invalidate(ctx context.Context, roomID string) {
 	if roomID == "" {
```
julianshen, line 271: _🏷️ distributed-systems_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`, `tgd-review` (2 rules)_

**Old binaries cannot invalidate the new v5 cache key**

Invalidation compatibility is only one-way. During a rolling deployment, a v5 process can populate `room:v5:r1:subs`; then an authoritative subscription change can commit on a still-running v4 process, whose compiled `Invalidate` deletes only v4 and v3. The v5 entry remains and v5 readers serve stale membership or `IsSubscribed` until its TTL expires—for example, an unsubscribe can still be included in later DM fan-out. Having v5 delete legacy keys does not address invalidations originating from legacy binaries. Use a two-phase rollout (first deploy code that still uses v4 but also invalidates v5, fully drain older binaries, then enable v5), or adopt a generation-independent invalidation mechanism/value-compatibility check. Add a mixed-version test that applies the old invalidation key set after seeding a v5 entry.

<details>

- **`tgd-review`**:165: During a rolling v4→v5 deployment, a v5 binary can populate `room:v5:<room>:subs`, then a membership change can be handled by a still-running v4 binary. That binary's existing `Invalidate` only deletes v4 and v3, so the v5 entry survives until TTL expiry. A v5 reader can consequently serve removed members or omit newly added members; for fan-out, that can send messages to a removed recipient. Adding v4 to the new binary's deletion list only handles the opposite direction. Avoid changing the read key in a single rolling release: use a two-phase rollout that first teaches every binary to invalidate v5, or retain the shared key and make new readers detect/refill entries lacking the field.

</details>

<details>

```
In `pkg/roomsubcache/roomsubcache.go` around line 271: Invalidation compatibility is only one-way. During a rolling deployment, a v5 process can populate `room:v5:r1:subs`; then an authoritative subscription change can commit on a still-running v4 process, whose compiled `Invalidate` deletes only v4 and v3. The v5 entry remains and v5 readers serve stale membership or `IsSubscribed` until its TTL expires—for example, an unsubscribe can still be included in later DM fan-out. Having v5 delete legacy keys does not address invalidations originating from legacy binaries. Use a two-phase rollout (first deploy code that still uses v4 but also invalidates v5, fully drain older binaries, then enable v5), or adopt a generation-independent invalidation mechanism/value-compatibility check. Add a mixed-version test that applies the old invalidation key set after seeding a v5 entry.

```

</details>

---

pkg/roomsubcache/loader_mongo.go
```diff
@@ -78,6 +80,7 @@ func NewMongoLoader(subscriptions, users *mongo.Collection) Loader {
 				IsBot:              doc.User.IsBot,
 				Muted:              doc.Muted,
 				HistorySharedSince: hssMs,
+				IsSubscribed:       doc.IsSubscribed,
```
julianshen, line 83: _🏷️ dead-code_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`dead-code`_

**IsSubscribed is populated but has no production consumer**

The loader projects and stores IsSubscribed, but the changed production code never reads the resulting Member field; only serialization and tests observe it. The PR description also says its broadcast-worker consumer is deferred to a separate change. This makes the projection, stored value, and accompanying v5 cache churn vestigial until that consumer lands. Add the production consumer in this change, or defer the field and schema bump until it is ready.

<details>

```
In `pkg/roomsubcache/loader_mongo.go` around line 83: The loader projects and stores IsSubscribed, but the changed production code never reads the resulting Member field; only serialization and tests observe it. The PR description also says its broadcast-worker consumer is deferred to a separate change. This makes the projection, stored value, and accompanying v5 cache churn vestigial until that consumer lands. Add the production consumer in this change, or defer the field and schema bump until it is ready.

```

</details>

---

julianshen (commented):
### tGDBot review of `6911d19` — 3 findings in 5 files

2 🔴 blocking · 1 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/465#issuecomment-5506144340), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

pkg/roomsubcache/roomsubcache.go
hmchangw: Confirmed, and the mechanism is exactly as stated. An old binary's `Invalidate` is compiled with `legacyCacheKeySchemaVersions = {"v3"}` and so deletes `room:v4:…` and `room:v3:…`, never `room:v5:…`. What this PR adds closes only the v5→v4 direction, which is what the code comment claims and all it claims.

Worth noting the hazard is not new to v5 — every bump in this file has had it. `v3→v4` was benign by accident: that bump changed the JSON *kind* (bare `[]Member` array → `Box` object), so a cross-generation read fails to decode and lands on the miss branch. `v4→v5` is the first bump where the old value decodes cleanly, which is exactly why the one-way invalidation starts to bite here rather than earlier.

Blast radius, to size it: both generations live during a rolling deploy of a binary that touches this key (`notification-worker` invalidates on membership/mute changes; `broadcast-worker` also reads and writes it), a membership change landing on an old binary, staleness bounded by `ROOMSUBCACHE_TTL`.

Rather than the two-phase rollout, the contained fix is to stop changing the read key and make the new field detectable in the *value*:

- `IsSubscribed *bool` (`json:"isSubscribed,omitempty"`); the key stays at `v4`.
- A reader treats an entry whose members lack the field as unusable and refills from Mongo. `NewMongoLoader` fills every member, so testing one member is sufficient; an empty-member entry carries no shape risk.
- Every generation then reads, writes and invalidates the *same* key, so invalidation is complete in both directions and no generation can be stranded.
- Old binaries are unaffected when they decode a new entry: their `Member` has no such field and `encoding/json` ignores it.

That costs one pointer per member and removes the v5 cold refill, and it retires the whole bump-the-key-on-every-field class rather than patching this instance of it.

It does change this PR's approach, so it is the author's call — flagging with the patch rather than pushing it.

---

pkg/roomsubcache/loader_mongo.go
hmchangw: Accurate as a description: nothing in production reads `Member.IsSubscribed` yet, and the `broadcast-worker` consumer is deliberately a separate change.

The part of this that has teeth is the cache churn, not the unread field. An unconsumed field on a projection costs a Mongo field and a few bytes per member; the v5 key bump costs every reader of the shared key (`broadcast-worker`, `notification-worker`) a cold refill from Mongo, for a field none of them reads yet. That is the vestigial cost worth avoiding.

The `*bool` alternative proposed on the other thread removes it — no key bump, so no refill — which leaves the field landing ahead of its consumer at roughly zero cost and makes the split scope cheap rather than wasteful. The other option is to land the `broadcast-worker` consumer in this PR and pay the refill once for something that is actually read.

Both are the author's call, so holding here rather than changing scope unilaterally.

---

hmchangw: Addressed on the current head. `e794668` lands the consumer in this PR: `publishDMEvents` now skips a `botDM` row whose `IsSubscribed` is false, and `user-service` busts the room's cache entry on both the unsubscribe and the re-subscribe path. The field is read in production, so it is no longer vestigial and the v5 refill buys something.

Resolving.

---

pkg/roomsubcache/roomsubcache.go
hmchangw: Still open, and `e794668` makes it sharper rather than resolving it. The mechanism is untouched — that commit only edited a doc comment in this file — but the finding now has a live consumer and a measured window:

- `publishDMEvents` gates `botDM` fan-out on `IsSubscribed`, so a stale v5 entry no longer costs an abstract "wrong member list": it delivers into a botDM the user removed — the exact outcome the new gate exists to prevent.
- `ROOMSUBCACHE_TTL` defaults to **90m** (stated in the new `user-service` comments), so the stale window this finding describes is up to an hour and a half, not a blip.

One adjacent point the same commit introduces, independent of key generations and worth deciding at the same time: `user-service`'s invalidation is itself new here, so during *its* rollout an old instance writes `isSubscribed` and busts nothing at all. Same 90m exposure, reached without any v4/v5 disagreement.

Both are covered by the same deploy ordering — land `user-service` (and any other invalidator) before `broadcast-worker`'s gate starts reading the flag — which is the two-phase shape you proposed, arrived at from the other direction. The `*bool` alternative from my earlier reply removes the generation half of it permanently; it does not remove the new-invalidator half, which only ordering fixes.

Leaving this for the author to decide between the two; no code change pushed for it.

---

PR #467, opened 2026-09-02
feat(history-service): add PREVIEW_WARMBACK_ENABLED kill switch

hmchangw:
`rooms.get` falls back to a Cassandra bucket walk for any room without a stored list preview, then warms the result back onto the room doc through a bounded background writer so the next read is cheap. That write had no off switch: `newPreviewWarmer` always started its workers.

## What this adds

`PREVIEW_WARMBACK_ENABLED` (default `true`) lets an operator withhold the optional write without touching the read. The walk still runs and the client still gets its preview; only the self-healing write back is skipped, so the room re-walks on a later read — once per preview-cache TTL, or every read where that cache is off or the walk resolves nothing.

Scope: the mutation path's preview repair (`persistMutatedPreview`, which withdraws a stale preview after an edit or delete) is correctness rather than optimization, and is deliberately **not** covered by the toggle.

## Design

Disabled is a `nopPreviewWarmer` behind a new `previewWriter` interface, mirroring the `RoomTimesCache` / `nopRoomTimesCache` idiom the package already uses for an optional tier. An earlier revision in this branch encoded "off" by constructing `previewWarmer` with `closed: true`; that was replaced because it overloaded a lifecycle field with a configuration meaning — `previewWarmer.Close` is nil-channel-safe only via its `closed` guard, so a warmer born closed put a `close(nil)` panic one idiomatic refactor away, and it left shutdown indistinguishable from off for any future log, metric or fast path. `newPreviewWarmer` keeps its original signature and can no longer build a warmer that cannot write.

`main.go` warns at startup when the switch is off, grouped with the `PAGE_TRIMMING_ENABLED` warning; `docker-compose.yml` surfaces the knob.

## Testing

TDD throughout — tests written first and confirmed failing. New coverage: config default/on/off parsing, and service-level assertions that a disabled warm-back still serves the walked preview (gomock fails the test if the write is attempted) and that `Close` stays fast and idempotent with nothing to drain. The disabled test was mutation-checked: forcing the wiring to always build a live warmer makes it fail.

CI is green on `602c84e` — `lint`, `test`, `test-integration (history-service)`, `sast` and `prewarm` all pass. Package coverage: service 93.3%, config 96.4%.

No client-facing behavior changes, so `docs/client-api.md` is untouched — the preview is served identically either way.

## Reviewer notes

- The two operator-facing messages (the startup warning and the Compose comment) word the preview-cache caveat differently; there is an open review thread on `cmd/main.go` about aligning them.
- Test fixtures elsewhere in `internal/service` build `config.Config` literals directly, so they now get the zero value (`false`) for this field and wire the no-op writer. Nothing fails today — only `rooms_test.go` exercises `RoomsGet` — but a future `rooms.get` test added through those fixtures would silently run against a disabled writer.

https://claude.ai/code/session_01J9xm99qoxHG2FYFBM8k4p9

### DIFF
--- a/history-service/cmd/main.go
+++ b/history-service/cmd/main.go
@@ -362,15 +362,14 @@ func main() {
 	} else {
 		slog.Warn("page trimming DISABLED — oversize replies fail with response_too_large")
 	}
+	if !cfg.PreviewWarmBackEnabled {
+		slog.Warn("preview warm-back DISABLED — rooms without a stored preview re-walk Cassandra once per preview-cache TTL")
+	}
 	opts = append(opts, service.WithPageBudget(pageBudget))
 
 	// The service reads the tier on the degraded path; the seeder above writes it.
 	if roomTimes != nil {
 		opts = append(opts, service.WithRoomTimesCache(roomTimes))
-	}
-
-	if !cfg.PreviewWarmBackEnabled {
-		slog.Warn("preview warm-back DISABLED — rooms without a stored preview re-walk Cassandra on every rooms.get")
 	}
 
 	svc := service.New(cassRepo, subSource, roomSource, pub, threadRoomRepo, threadSubRepo, userSource, appRepo, &cfg, opts...)
--- a/history-service/cmd/main.go
+++ b/history-service/cmd/main.go
@@ -369,6 +369,10 @@ func main() {
 		opts = append(opts, service.WithRoomTimesCache(roomTimes))
 	}
 
+	if !cfg.PreviewWarmBackEnabled {
+		slog.Warn("preview warm-back DISABLED — rooms without a stored preview re-walk Cassandra on every rooms.get")
+	}
+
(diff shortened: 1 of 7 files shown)


julianshen:
**No actionable comments.** ✅

> [!NOTE]

### Local review memory

- `history-service/cmd/main.go:366` — @coderabbitai: _🎯 Functional Correctness_ _🟡 Minor_ _⚡ Quick win_ **Use the same preview-cache caveat in both operator-facing messages.** The preview cache limits repeat Cassandra walks only when it is enabled and the walk returns a positive previe… ([thread](https://github.com/hmchangw/newchat/pull/467#discussion_r3915148715))

<details>

* `history-service/internal/service/rooms_test.go`
* `history-service/internal/service/warmback.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

mliu33 (approved):

PR #469, opened 2026-09-03
Add subscription-list max-rps workload to loadgen

hmchangw:
Adds a ninth `max-rps` workload driving `subscription.list` over NATS request/reply, so the sidebar cold-open path gets the capacity ramp the soak and daily programs already assume but never measured.

```bash
loadgen seed     --workload=subscription-list --preset=realistic
loadgen max-rps  --workload=subscription-list --preset=realistic
loadgen teardown --workload=subscription-list --preset=realistic
```

Scope is NATS only. An HTTP arm against `GET /api/v1/subscriptions` was considered and deliberately deferred — it needs a real Keycloak token per fixture account, and the realm ships only `alice`/`bob`, so it wants its own design.

## Workload shape

Every request is the call a client makes on connect — `type=current`, `limit=200`, `includeLastMessage=true`, `offset=0` — because the reconnect burst, not steady browsing, is what this endpoint has to survive.

Accounts are picked **uniformly**, unlike `room-read`'s Zipf skew: a client cold-opens its sidebar once per connect, so there is no hot-account concentration to model. The skew that matters is across rooms *within* one account's page, which the `realistic` fixtures carry.

## Use `--preset=realistic`

Not incidental. The uniform presets put every account in **exactly one room**:

| preset | accounts | max subs/account | accounts with 2+ |
|---|---|---|---|
| `small` / `medium` / `large` | 10 / 1000 / 10000 | **1** | **0** |
| `realistic` | 955 | 9 | 794 |

A ramp on `medium` measures a one-row reply — fast, and silent about the endpoint under real sidebars. The workload warns at startup when the preset averages under two subscriptions per account, and logs mean rows per page at every step, so a degenerate run is visible in the output instead of just looking quick.

## Why it seeds its own fixtures

`seed --workload=subscription-list` is not interchangeable with the plain `seed`. `BuildFixtures` leaves three fields at their zero value that `subscription.list`'s match filter reads, and a miss on any of them returns an **empty page rather than an error** — so the stock fixtures would have produced a ramp measuring the cost of finding nothing:

| Field | Why the zero value fails |
|---|---|
| `open` | no bson `omitempty`, so `false` persists as `open:false` and `match["open"]={$ne:false}` drops the row |
| `roomType` | the `current` match needs `dm`/`channel` (or a subscribed `botDM`); `""` matches no branch |
| `name` | `subLite` projects it for the self-DM pin and the name tiebreak |

The seeder also spreads each room's `lastMsgAt` and each member's `lastSeenAt`. An all-equal sort key would collapse the list comparator onto its name tiebreak and hide the real sort cost; a uniform `lastSeenAt` would make every row resolve `hasUnread` the same way.

## Reply classification

Four outcomes are kept apart. An errcode envelope and a `{}` body both decode to zero rows, so folding either into "empty page" would report a service outage or a contract break as a fast, healthy, empty sidebar. The `{}` check is on presence of the collection, not its length — a genuinely empty sidebar replies with an empty array.

Empty pages **count as failures**. Every seeded account owns subscriptions, so a zero-row reply means the fixtures are wrong — and since an empty page is also the fastest possible reply, scoring it as a success would let a misconfigured run report a record-breaking ramp.

## Note for whoever runs this first

`--list-limit=200` can exceed the NATS payload ceiling. `docs/client-api.md` recommends 200, but that advice is for the HTTP form, which has no ceiling; the NATS reply caps at 128 KB and the handler does not truncate. When a page overflows, the service cannot publish the reply and the request times out. **A ramp reporting timeouts at every step, including the lowest, is likely hitting the cap rather than a capacity limit** — lower `--list-limit` to confirm. The default is left at 200 rather than quietly lowered, since surfacing this is arguably the more useful result. Documented in the README.

`--list-type` accepts `current` and `rooms`. `apps` is rejected with a specific reason: user-service serves it, but it matches subscribed `botDM` rows alone and these fixtures seed none, so every page would come back empty.

## Files

| File | |
|---|---|
| `subscriptionlist_seed.go` | fixtures builder, degenerate-preset detection, `seed`/`teardown` wiring |
| `subscriptionlist_collector.go` | samples, error classes, row stats |
| `subscriptionlist_generator.go` | account pick, request build, reply classification |
| `maxrps_subscriptionlist.go` | `rpsWorkload` adapter |
| `maxrps.go` / `main.go` / `deploy/Makefile` / `README.md` | wiring and docs |

Each with a table-driven test file, written first. The request body is marshalled once at construction — re-encoding per call would put the load box's CPU on the critical path and surface as emit underrun.

No `docs/client-api.md` change: this only reads existing endpoints, it does not touch a handler.

## Verification

- All five CI checks green, **including `test-integration (tools)`** — the integration test drives user-service's real `SubscriptionRepo` against real Mongo to prove the seeded fixtures survive its match filter.
- `make lint`: 0 issues. `go test ./tools/loadgen/ -race`: pass. `gosec` and the repo-owned semgrep rules: 0 findings.
- CLI validation confirmed against the built binary — bad `--list-type`, `--list-type=apps`, zero `--list-limit`, unknown preset and zero `--request-timeout` all exit 2 before the NATS dial.
- Coverage on new code is 87–100%, except `newSubscriptionListWorkload` at 0%: it dials NATS, the same untested boundary as `newRoomReadWorkload` and every sibling constructor.

### Review history

Two defects were found after the first push and fixed in place, both worth knowing about when reading the diff:

1. **CI caught the degenerate-preset problem** (7302123). The paging integration test failed a precondition on `small`; investigating it surfaced that three of four presets give one subscription per account, and that the docs recommended one of them. Fixed with the startup warning, per-step row logging, and the preset guidance above.
2. **CodeRabbit caught `--list-type=apps`** (fa39b7c) accepting a value the fixtures cannot serve — every page empty, every request a failure. Now rejected with a reason.

https://claude.ai/code/session_01A8S2JxAqtUKQ5pd2FosCkw

## Summary by CodeRabbit

* **New Features**
  * Added a `subscription-list` workload for benchmarking subscription-list requests.
  * Added options for list type, page size, and last-message inclusion.
  * Added commands to seed and tear down subscription-list fixtures.
  * Added reporting for latency, returned rows, paging, errors, empty pages, and pacing metrics.

* **Documentation**
  * Documented setup, usage, configuration, realistic fixture requirements, and workload limitations, including unsupported app listings.

* **Tests**
  * Added coverage for workload execution, fixture generation, request handling, metrics, paging, and integration behavior.

### DIFF
--- a/tools/loadgen/README.md
+++ b/tools/loadgen/README.md
@@ -537,6 +537,83 @@ make -C tools/loadgen/deploy teardown-roomread PRESET=medium
 - Presets are the messages presets (`small`/`medium`/`large`/`realistic`); room
   size distribution drives floor-write contention.
 
+## Subscription-list workload (sidebar cold-open benchmark)
+
+Finds the maximum sustainable RPS for `subscription.list`
+(`user-service.ListSubscriptions`, the
+`chat.user.{account}.request.user.{siteID}.subscription.list` request/reply
+RPC). Every request is the call a client makes on connect — `type=current`,
+`limit=200`, `includeLastMessage=true`, `offset=0` — because the reconnect
+burst, not steady browsing, is what this endpoint has to survive.
+
+Accounts are picked **uniformly**, unlike the room-read workload's Zipf skew: a
+client cold-opens its sidebar once per connect, so there is no hot-account
+concentration to model. The skew that matters is across rooms *within* one
+account's page, and the preset fixtures already carry it.
+
+### Quick start
+
+```
+make -C tools/loadgen/deploy up
+make -C tools/loadgen/deploy seed-sublist PRESET=medium
+make -C tools/loadgen/deploy run-max-rps WORKLOAD=subscription-list PRESET=medium
+```
+
+Override the ramp with `STEPS` (default `200,500,1000,2000,5000`):
+
+```
+make -C tools/loadgen/deploy run-max-rps WORKLOAD=subscription-list PRESET=medium STEPS=500,1k,2k,5k
+```
+
+Tear down the fixtures:
+
+```
+make -C tools/loadgen/deploy teardown-sublist PRESET=medium
+```
+
+### Why this workload seeds its own fixtures
+
+`seed --workload=subscription-list` is **not** interchangeable with the plain
+`seed`. `BuildFixtures` leaves three fields at their zero value that
+`subscription.list`'s match filter reads, and a miss on any of them returns an
+empty page rather than an error — so the stock fixtures would produce a ramp
+measuring the cost of finding nothing:
+
+| Field | Why the zero value fails |
+|---|---|
+| `open` | no bson `omitempty`, so `false` persists as `open:false` and `match["open"]={$ne:false}` drops the row |
+| `roomType` | the `current` match needs `dm`/`channel` (or a subscribed `botDM`); `""` matches no branch |
+| `name` | `subLite` projects it for the self-DM pin and the name tiebreak |
+
+The seeder also spreads each room's `lastMsgAt` and each member's `lastSeenAt`.
+Both matter: an all-equal sort key would collapse the list comparator onto its
+name tiebreak and hide the real sort cost, and a uniform `lastSeenAt` would make
+every row resolve `hasUnread` the same way.
+
+### Notes
+
+- Synchronous request/reply: gated on p95/p99 latency and error rate only (no
+  consumer-pending signal). Defaults: `--slo-p95=100ms`, `--slo-p99=250ms`,
+  `--slo-error-rate=0.001`.
+- **Empty pages count as failures.** Every seeded account owns subscriptions, so
+  a zero-row reply means the fixtures or `--list-type` are wrong. An empty page
+  is also the fastest possible reply, so scoring it as a success would let a
+  misconfigured run report a record-breaking ramp. A service error envelope and
+  a `{}` body are kept distinct from it — both decode to zero rows, and folding
+  them together would report an outage as a healthy empty sidebar.
+- **`--list-limit=200` can exceed the NATS payload ceiling.** `docs/client-api.md`
+  recommends 200 for the HTTP form, which has no ceiling; the NATS reply is
+  capped at 128 KB and the handler does not truncate. When a page overflows, the
+  service cannot publish the reply and the request times out — so a ramp that
+  reports timeouts at *every* step, including the lowest, is likely hitting the
+  cap rather than a capacity limit. Lower `--list-limit` to confirm.
+- The NATS page is additionally capped server-side by `MAX_SUBSCRIPTION_LIMIT`
+  (default 1000), so `--list-limit` above that is silently clamped.
+- Single-site only: all seeded users are local.
+- Presets are the messages presets (`small`/`medium`/`large`/`realistic`).
+  `realistic` is the most representative — its mixed room sizes and DM rows
+  exercise the per-row app/HR enrichment branches that a uniform preset skips.
+
 ## History workload (LoadHistory / GetThreadMessages benchmark)
 
 Benchmarks the synchronous read path:
@@ -746,7 +823,7 @@ list of steps, holds at each step for a measurement window, evaluates SLO
 signals, and reports the largest step at which every signal passed.
 
 ```bash
-loadgen max-rps --workload=messages|history|read-receipt|room-read|thread-read|login|search --preset=<name> [flags]
+loadgen max-rps --workload=messages|history|read-receipt|room-read|thread-read|subscription-list|login|search --preset=<name> [flags]
 ```
 
 ### Quick start
@@ -768,6 +845,10 @@ loadgen max-rps --workload=login --preset=medium
 
 # search: drive search-service request/reply (reads the existing index)
 loadgen max-rps --workload=search --preset=medium
+
+# subscription-list: seed the list-shaped fixtures first, then ramp
+loadgen seed --workload=subscription-list --preset=medium
+loadgen max-rps --workload=subscription-list --preset=medium --steps=200,500,1k,2k
 ```
 
 Via the deploy Makefile:
@@ -781,10 +862,13 @@ make -C tools/loadgen/deploy run-max-rps WORKLOAD=history PRESET=history-medium 
 
 | Flag | Default | Notes |
 |------|---------|-------|
-| `--workload` | `messages` | `messages`, `history`, `read-receipt`, `room-read`, `thread-read`, `login`, or `search` |
+| `--workload` | `messages` | `messages`, `history`, `read-receipt`, `room-read`, `thread-read`, `subscription-list`, `login`, or `search` |
 | `--preset` | (required) | an existing preset for the chosen workload (`read-receipt` reuses the history presets; `login` and `search` reuse the message presets for their account set) |
-| `--steps` | messages `500,1k,2k,5k,10k` / history+read-receipt `200,500,1k,2k,5k` / login `50,100,200,500,1k` / search `100,200,500,1k,2k` | explicit ordered RPS list; `k` suffix = ×1000 |
-| `--request-timeout` | `5s` | **history / read-receipt / room-read / thread-read / login / search**: per-request reply timeout |
+| `--steps` | messages `500,1k,2k,5k,10k` / history+read-receipt+subscription-list `200,500,1k,2k,5k` / login `50,100,200,500,1k` / search `100,200,500,1k,2k` | explicit ordered RPS list; `k` suffix = ×1000 |
+| `--request-timeout` | `5s` | **history / read-receipt / room-read / thread-read / subscription-list / login / search**: per-request reply timeout |
+| `--list-type` | `current` | **subscription-list only**: `current`, `rooms`, or `apps` |
+| `--list-limit` | `200` | **subscription-list only**: page size; see the NATS payload-ceiling note above |
+| `--include-last-message` | `true` | **subscription-list only**: request the per-room `previewMessage` enrichment; `=false` isolates the query from the enrichment cost |
 | `--auth-url` | `$AUTH_URL` | **login only**: auth-service base URL |
 | `--login-key-pool` | `256` | **login only**: pre-generated NKey pool size |
 | `--search-mix` | `messages:60,rooms:30,users:10` | **search only**: endpoint mix |
--- a/tools/loadgen/README.md
+++ b/tools/loadgen/README.md
@@ -542,9 +542,11 @@ make -C tools/loadgen/deploy teardown-roomread PRESET=medium
 Finds the maximum sustainable RPS for `subscription.list`
 (`user-service.ListSubscriptions`, the
 `chat.user.{account}.request.user.{siteID}.subscription.list` request/reply
-RPC). Every request is the call a client makes on connect — `type=current`,
-`limit=200`, `includeLastMessage=true`, `offset=0` — because the reconnect
-burst, not steady browsing, is what this endpoint has to survive.
+RPC). By default every request is the call a client makes on connect —
+`type=current`, `limit=200`, `includeLastMessage=true`, `offset=0` — because the
+reconnect burst, not steady browsing, is what this endpoint has to survive.
+`--list-type`, `--list-limit` and `--include-last-message` change that shape;
+`offset` is always 0, since a cold open asks for the first page.
 
 Accounts are picked **uniformly**, unlike the room-read workload's Zipf skew: a
 client cold-opens its sidebar once per connect, so there is no hot-account
@@ -553,7 +555,7 @@ account's page, and the preset fixtures already carry it.
 
 ### Quick start
 
-```
+```sh
 make -C tools/loadgen/deploy up
 make -C tools/loadgen/deploy seed-sublist PRESET=realistic
 make -C tools/loadgen/deploy run-max-rps WORKLOAD=subscription-list PRESET=realistic
@@ -563,13 +565,13 @@ make -C tools/loadgen/deploy run-max-rps WORKLOAD=subscription-list PRESET=reali
 
 Override the ramp with `STEPS` (default `200,500,1000,2000,5000`):
 
-```
+```sh
 make -C tools/loadgen/deploy run-max-rps WORKLOAD=subscription-list PRESET=realistic STEPS=500,1k,2k,5k
 ```
 
 Tear down the fixtures:
 
-```
+```sh
 make -C tools/loadgen/deploy teardown-sublist PRESET=realistic
 ```
 
--- a/tools/loadgen/README.md
+++ b/tools/loadgen/README.md
@@ -555,20 +555,22 @@ account's page, and the preset fixtures already carry it.
 
 ```
 make -C tools/loadgen/deploy up
-make -C tools/loadgen/deploy seed-sublist PRESET=medium
-make -C tools/loadgen/deploy run-max-rps WORKLOAD=subscription-list PRESET=medium
+make -C tools/loadgen/deploy seed-sublist PRESET=realistic
+make -C tools/loadgen/deploy run-max-rps WORKLOAD=subscription-list PRESET=realistic
 ```
+
+`realistic` is not incidental here — see the preset note below.
 
 Override the ramp with `STEPS` (default `200,500,1000,2000,5000`):
 
 ```
-make -C tools/loadgen/deploy run-max-rps WORKLOAD=subscription-list PRESET=medium STEPS=500,1k,2k,5k
+make -C tools/loadgen/deploy run-max-rps WORKLOAD=subscription-list PRESET=realistic STEPS=500,1k,2k,5k
 ```
 
 Tear down the fixtures:
 
 ```
-make -C tools/loadgen/deploy teardown-sublist PRESET=medium
+make -C tools/loadgen/deploy teardown-sublist PRESET=realistic
 ```
 
 ### Why this workload seeds its own fixtures
@@ -610,9 +612,15 @@ every row resolve `hasUnread` the same way.
 - The NATS page is additionally capped server-side by `MAX_SUBSCRIPTION_LIMIT`
   (default 1000), so `--list-limit` above that is silently clamped.
 - Single-site only: all seeded users are local.
-- Presets are the messages presets (`small`/`medium`/`large`/`realistic`).
-  `realistic` is the most representative — its mixed room sizes and DM rows
-  exercise the per-row app/HR enrichment branches that a uniform preset skips.
+- **Use `--preset=realistic`.** The uniform presets (`small`/`medium`/`large`)
+  put every account in exactly one room, so each reply carries a single row — a
+  ramp over them measures the latency of a one-row page, which is fast and says
+  nothing about the endpoint under real sidebars. Only `realistic` has the mixed
+  room sizes that give an account several rooms (mean ~3, max ~9), and its DM
+  rows are also what exercise the per-row HR-enrichment branch. The workload logs
+  a startup warning when the chosen preset averages under two subscriptions per
+  account, and logs the mean rows per page at every step, so a degenerate run is
+  visible in the output rather than silently fast.
 
 ## History workload (LoadHistory / GetThreadMessages benchmark)
 
@@ -847,8 +855,9 @@ loadgen max-rps --workload=login --preset=medium
 loadgen max-rps --workload=search --preset=medium
 
 # subscription-list: seed the list-shaped fixtures first, then ramp
-loadgen seed --workload=subscription-list --preset=medium
-loadgen max-rps --workload=subscription-list --preset=medium --steps=200,500,1k,2k
+# (realistic is the only preset whose accounts hold more than one room)
+loadgen seed --workload=subscription-list --preset=realistic
+loadgen max-rps --workload=subscription-list --preset=realistic --steps=200,500,1k,2k
 ```
 
 Via the deploy Makefile:
--- a/tools/loadgen/README.md
+++ b/tools/loadgen/README.md
@@ -598,7 +598,7 @@ every row resolve `hasUnread` the same way.
   consumer-pending signal). Defaults: `--slo-p95=100ms`, `--slo-p99=250ms`,
   `--slo-error-rate=0.001`.
 - **Empty pages count as failures.** Every seeded account owns subscriptions, so
-  a zero-row reply means the fixtures or `--list-type` are wrong. An empty page
+  a zero-row reply means the fixtures are wrong. An empty page
   is also the fastest possible reply, so scoring it as a success would let a
   misconfigured run report a record-breaking ramp. A service error envelope and
   a `{}` body are kept distinct from it — both decode to zero rows, and folding
@@ -875,7 +875,7 @@ make -C tools/loadgen/deploy run-max-rps WORKLOAD=history PRESET=history-medium 
 | `--preset` | (required) | an existing preset for the chosen workload (`read-receipt` reuses the history presets; `login` and `search` reuse the message presets for their account set) |
 | `--steps` | messages `500,1k,2k,5k,10k` / history+read-receipt+subscription-list `200,500,1k,2k,5k` / login `50,100,200,500,1k` / search `100,200,500,1k,2k` | explicit ordered RPS list; `k` suffix = ×1000 |
 | `--request-timeout` | `5s` | **history / read-receipt / room-read / thread-read / subscription-list / login / search**: per-request reply timeout |
-| `--list-type` | `current` | **subscription-list only**: `current`, `rooms`, or `apps` |
+| `--list-type` | `current` | **subscription-list only**: `current` or `rooms`. `apps` is rejected — user-service serves it, but it matches subscribed `botDM` rows alone and these fixtures seed none, so every page would come back empty |
 | `--list-limit` | `200` | **subscription-list only**: page size; see the NATS payload-ceiling note above |
 | `--include-last-message` | `true` | **subscription-list only**: request the per-room `previewMessage` enrichment; `=false` isolates the query from the enrichment cost |
 | `--auth-url` | `$AUTH_URL` | **login only**: auth-service base URL |
--- a/tools/loadgen/main.go
+++ b/tools/loadgen/main.go
@@ -162,7 +162,7 @@ func dispatch(ctx context.Context, cfg *config) int {
 
 func runSeed(ctx context.Context, cfg *config, args []string) int {
 	fs := flag.NewFlagSet("seed", flag.ExitOnError)
-	workload := fs.String("workload", "messages", "messages|thread|members|history|read-receipt|room-read|thread-read|botroom|soak")
+	workload := fs.String("workload", "messages", "messages|thread|members|history|read-receipt|room-read|thread-read|subscription-list|botroom|soak")
 	preset := fs.String("preset", "", "preset name")
 	seed := fs.Int64("seed", 42, "RNG seed")
 	readRatio := fs.Float64("read-ratio", 0.7, "read-receipt only: fraction of each room's subscribers to mark as readers")
@@ -202,6 +202,8 @@ func runSeed(ctx context.Context, cfg *config, args []string) int {
 		return runSeedReadReceipt(ctx, cfg, *preset, *seed, *readRatio)
 	case "room-read":
 		return runSeedRoomRead(ctx, cfg, *preset, *seed)
+	case "subscription-list":
+		return runSeedSubscriptionList(ctx, cfg, *preset, *seed)
 	case "thread-read":
 		return runSeedHistory(ctx, cfg, *preset, *seed)
 	case "botroom":
@@ -340,7 +342,7 @@ func runTeardownBotRoom(ctx context.Context, cfg *config, preset string, seed in
 
 func runTeardown(ctx context.Context, cfg *config, args []string) int {
 	fs := flag.NewFlagSet("teardown", flag.ExitOnError)
-	workload := fs.String("workload", "messages", "messages|thread|members|history|room-read|thread-read|botroom|soak")
+	workload := fs.String("workload", "messages", "messages|thread|members|history|room-read|thread-read|subscription-list|botroom|soak")
 	preset := fs.String("preset", "", "preset name (required to identify which room keys to delete)")
 	seed := fs.Int64("seed", 42, "RNG seed (must match the seed used at seed time)")
 	_ = fs.Parse(args)
@@ -362,6 +364,8 @@ func runTeardown(ctx context.Context, cfg *config, args []string) int {
 		return runTeardownHistory(ctx, cfg, *preset, *seed)
 	case "room-read":
 		return runTeardownRoomRead(ctx, cfg, *preset, *seed)
+	case "subscription-list":
+		return runTeardownSubscriptionList(ctx, cfg, *preset, *seed)
 	case "thread-read":
 		return runTeardownHistory(ctx, cfg, *preset, *seed)
 	case "botroom":
@@ -529,6 +533,50 @@ func runTeardownRoomRead(ctx context.Context, cfg *config, preset string, seed i
 		return 1
 	}
 	slog.Info("teardown complete (room-read)", "preset", preset)
+	return 0
+}
+
+func runSeedSubscriptionList(ctx context.Context, cfg *config, preset string, seed int64) int {
+	p, ok := BuiltinPreset(preset)
+	if !ok {
+		fmt.Fprintf(os.Stderr, "unknown preset: %s\n", preset)
+		return 2
+	}
+	db, _, cleanup, err := connectStores(ctx, cfg)
+	if err != nil {
+		return 1
+	}
+	defer cleanup()
+	fixtures := BuildSubscriptionListFixtures(&p, seed, cfg.SiteID, time.Now().UTC())
+	// No SeedRoomKeys: the list path never decrypts, so no room keys are written.
+	if err := Seed(ctx, db, &fixtures); err != nil {
+		slog.Error("seed", "error", err)
+		return 1
+	}
+	slog.Info("seed complete (subscription-list)",
+		"preset", p.Name,
+		"users", len(fixtures.Users),
+		"rooms", len(fixtures.Rooms),
+		"subs", len(fixtures.Subscriptions))
+	return 0
+}
+
+func runTeardownSubscriptionList(ctx context.Context, cfg *config, preset string, seed int64) int {
+	if _, ok := BuiltinPreset(preset); !ok {
+		fmt.Fprintf(os.Stderr, "unknown preset: %s\n", preset)
+		return 2
+	}
+	db, _, cleanup, err := connectStores(ctx, cfg)
+	if err != nil {
+		return 1
+	}
+	defer cleanup()
+	// No TeardownRoomKeys: runSeedSubscriptionList writes no room keys.
+	if err := Teardown(ctx, db); err != nil {
+		slog.Error("teardown", "error", err)
+		return 1
+	}
+	slog.Info("teardown complete (subscription-list)", "preset", preset)
 	return 0
 }
 
--- a/tools/loadgen/main.go
+++ b/tools/loadgen/main.go
@@ -365,7 +365,7 @@ func runTeardown(ctx context.Context, cfg *config, args []string) int {
 	case "room-read":
 		return runTeardownRoomRead(ctx, cfg, *preset, *seed)
 	case "subscription-list":
-		return runTeardownSubscriptionList(ctx, cfg, *preset, *seed)
+		return runTeardownSubscriptionList(ctx, cfg, *preset)
 	case "thread-read":
 		return runTeardownHistory(ctx, cfg, *preset, *seed)
 	case "botroom":
@@ -561,7 +561,9 @@ func runSeedSubscriptionList(ctx context.Context, cfg *config, preset string, se
 	return 0
 }
 
-func runTeardownSubscriptionList(ctx context.Context, cfg *config, preset string, seed int64) int {
+// No seed parameter: teardown drops the seeded collections wholesale, so unlike
+// the seeding side it does not depend on which fixtures were generated.
+func runTeardownSubscriptionList(ctx context.Context, cfg *config, preset string) int {
 	if _, ok := BuiltinPreset(preset); !ok {
 		fmt.Fprintf(os.Stderr, "unknown preset: %s\n", preset)
 		return 2
--- a/tools/loadgen/maxrps.go
+++ b/tools/loadgen/maxrps.go
@@ -114,7 +114,7 @@ func runMaxRPS(ctx context.Context, cfg *config, args []string) int {
 	// search-only tunable:
 	searchMixFlag := fs.String("search-mix", "messages:60,rooms:30,users:10", "search only: endpoint mix")
 	// subscription-list-only tunables:
-	listType := fs.String("list-type", "current", "subscription-list only: current|rooms|apps")
+	listType := fs.String("list-type", "current", "subscription-list only: current|rooms (apps needs botDM fixtures this workload does not seed)")
 	listLimit := fs.Int("list-limit", 200, "subscription-list only: page size (docs/client-api.md recommends 200)")
 	includeLastMessage := fs.Bool("include-last-message", true, "subscription-list only: request the per-room previewMessage enrichment")
 	csvPath := fs.String("csv", "", "optional CSV output path")
@@ -305,8 +305,12 @@ func runMaxRPS(ctx context.Context, cfg *config, args []string) int {
 			fmt.Fprintf(os.Stderr, "unknown preset: %s\n", *preset)
 			return 2
 		}
-		if !validSubscriptionListTypes[*listType] {
-			fmt.Fprintln(os.Stderr, "--list-type must be one of current|rooms|apps")
+		if !workloadSupportedListTypes[*listType] {
+			if validSubscriptionListTypes[*listType] {
+				fmt.Fprintf(os.Stderr, "--list-type=%s is not supported by this workload: its fixtures seed no botDM rows, so every page would come back empty\n", *listType)
+				return 2
+			}
+			fmt.Fprintln(os.Stderr, "--list-type must be one of current|rooms")
 			return 2
 		}
 		if *listLimit <= 0 {
--- a/tools/loadgen/maxrps.go
+++ b/tools/loadgen/maxrps.go
@@ -11,7 +11,7 @@ import (
 
 func defaultSteps(workload string) string {
 	switch workload {
-	case "history", "read-receipt", "room-read", "thread-read":
+	case "history", "read-receipt", "room-read", "thread-read", "subscription-list":
 		return "200,500,1000,2000,5000"
 	case "login":
 		// Login is a low-rate journey — one per session, not per message — and
@@ -85,7 +85,7 @@ func validateSLORatio(name string, r sloRatio, wantBound bool) bool {
 // the report. Returns the process exit code.
 func runMaxRPS(ctx context.Context, cfg *config, args []string) int {
 	fs := flag.NewFlagSet("max-rps", flag.ExitOnError)
-	workload := fs.String("workload", "messages", "messages|thread|history|read-receipt|room-read|thread-read|login|search")
+	workload := fs.String("workload", "messages", "messages|thread|history|read-receipt|room-read|thread-read|subscription-list|login|search")
 	preset := fs.String("preset", "", "preset name")
 	seed := fs.Int64("seed", 42, "RNG seed")
 	stepsFlag := fs.String("steps", "", "ascending RPS list, e.g. 500,1k,2k,5k,10k (default depends on workload)")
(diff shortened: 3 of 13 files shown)


julianshen:
**6 findings · 5 unique issues · 5 inline comments posted** — 🟠 5 warning · 🔵 1 suggestion

<details>
<summary>📒 Files reviewed (13)</summary>

* `tools/loadgen/README.md`
* `tools/loadgen/deploy/Makefile`
* `tools/loadgen/main.go`
* `tools/loadgen/maxrps.go`
* `tools/loadgen/maxrps_subscriptionlist.go`
* `tools/loadgen/maxrps_subscriptionlist_test.go`
* `tools/loadgen/subscriptionlist_collector.go`
* `tools/loadgen/subscriptionlist_collector_test.go`
* `tools/loadgen/subscriptionlist_generator.go`
* `tools/loadgen/subscriptionlist_generator_test.go`
* `tools/loadgen/subscriptionlist_integration_test.go`
* `tools/loadgen/subscriptionlist_seed.go`
* `tools/loadgen/subscriptionlist_seed_test.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

tools/loadgen/subscriptionlist_collector.go
```diff
@@ -0,0 +1,154 @@
...
+// type are wrong, not that the service is fast — counting it as a success would
+// report a healthy ramp measuring nothing.
+const errClassEmptyPage errClass = "empty_page"
+
+// SubscriptionListSample captures one completed subscription.list round-trip.
+// Rows and HasMore are kept so the report can show the page size actually
+// measured: a ramp over 3-row pages is not a ramp over sidebar-sized ones.
+type SubscriptionListSample struct {
+	Latency time.Duration
+	Rows    int
```
julianshen, line 19: _🏷️ dead-code_ | _🟠 Warning_ | _⚡ Quick fix_ | _`dead-code`_

**Row statistics are collected but never reach the report.**

`classifyReply` populates `Rows` and `HasMore`, and `RecordSample` aggregates them into `totalRows` and `hasMore`. However, `RunStep` only passes the collector to `buildSubscriptionListInputs`, which extracts latency and never calls `TotalRows`, `MeanRows`, or `HasMoreCount`; the collector is then discarded. Consequently, the promised page-size telemetry has no production consumer. Either add these statistics to the report handoff or remove the fields, counters, accessors, assignments, and tests.

<details>

```
In `tools/loadgen/subscriptionlist_collector.go` around line 19: `classifyReply` populates `Rows` and `HasMore`, and `RecordSample` aggregates them into `totalRows` and `hasMore`. However, `RunStep` only passes the collector to `buildSubscriptionListInputs`, which extracts latency and never calls `TotalRows`, `MeanRows`, or `HasMoreCount`; the collector is then discarded. Consequently, the promised page-size telemetry has no production consumer. Either add these statistics to the report handoff or remove the fields, counters, accessors, assignments, and tests.

```

</details>

---

tools/loadgen/subscriptionlist_generator.go
```diff
@@ -0,0 +1,183 @@
...
+	ctx = natsutil.WithRequestID(ctx, idgen.GenerateRequestID())
+
+	start := time.Now()
+	reply, err := g.cfg.Requester.Request(ctx, subj, g.body, g.cfg.RequestTimeout)
+	latency := time.Since(start)
+
+	if err != nil {
+		// Run-level cancellation isn't a real failure — the run is draining.
+		if ctx.Err() != nil {
+			return
```
julianshen, line 168: _🏷️ failure-handling_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`_

**Step shutdown censors in-flight failures from the SLO sample.**

When a hold ends, runSubscriptionListFor cancels genCtx before waiting for admitted requests. Those requests use the same context, so a request still awaiting NATS or the service can return cancellation—or later return its timeout after cancellation—and this branch drops it because ctx.Err() is set. The request is then absent from both AttemptedOps and FailedOps. Slow calls disproportionately cross the boundary, biasing latency and error-rate measurements downward and potentially overstating sustainable RPS. Stop dispatch at the boundary while allowing already-admitted requests to complete under their independent request timeout, then wait for and record their outcomes.

<details>

```
In `tools/loadgen/subscriptionlist_generator.go` around line 140: When a hold ends, runSubscriptionListFor cancels genCtx before waiting for admitted requests. Those requests use the same context, so a request still awaiting NATS or the service can return cancellation—or later return its timeout after cancellation—and this branch drops it because ctx.Err() is set. The request is then absent from both AttemptedOps and FailedOps. Slow calls disproportionately cross the boundary, biasing latency and error-rate measurements downward and potentially overstating sustainable RPS. Stop dispatch at the boundary while allowing already-admitted requests to complete under their independent request timeout, then wait for and record their outcomes.

```

</details>

---

tools/loadgen/maxrps_subscriptionlist.go
```diff
@@ -0,0 +1,158 @@
...
+// (hold-only) collector. subscription.list is synchronous request/reply, so
+// there is no consumer queue and Pending stays empty; the single
+// "subscription-list" latency series gates.
+//
+// Empty pages join the failure count. Every seeded account owns subscriptions,
+// so a zero-row reply means the fixtures or the list type are wrong — and since
+// an empty page is also the fastest possible reply, scoring it as a success
+// would let a misconfigured run report a record-breaking ramp.
+func buildSubscriptionListInputs(targetRPS int, hold time.Duration, c *SubscriptionListCollector) rpsStepInputs {
+	samples := c.Samples()
```
julianshen, line 32: _🏷️ observability_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`_

**The report collapses all four failure classes into one count.**

The collector distinguishes transport timeouts, service error envelopes, malformed replies, and empty pages, but this conversion immediately sums them into FailedOps; RunStep then discards the collector. Thus a broker/service timeout, a protocol incompatibility, and bad fixtures produce the same report even though they require different action. Expose the per-class counts in the step report, or at minimum emit a per-step diagnostic summary before the collector is discarded.

<details>

```
In `tools/loadgen/maxrps_subscriptionlist.go` around line 32: The collector distinguishes transport timeouts, service error envelopes, malformed replies, and empty pages, but this conversion immediately sums them into FailedOps; RunStep then discards the collector. Thus a broker/service timeout, a protocol incompatibility, and bad fixtures produce the same report even though they require different action. Expose the per-class counts in the step report, or at minimum emit a per-step diagnostic summary before the collector is discarded.

```

</details>

---

tools/loadgen/maxrps.go
```diff
@@ -295,6 +299,33 @@ func runMaxRPS(ctx context.Context, cfg *config, args []string) int {
 			return 1
 		}
 		w, cleanup, presetID = rw, clean, p.Name
+	case "subscription-list":
+		p, ok := BuiltinPreset(*preset)
+		if !ok {
+			fmt.Fprintf(os.Stderr, "unknown preset: %s\n", *preset)
+			return 2
+		}
+		if !validSubscriptionListTypes[*listType] {
```
julianshen: _🎯 MongoDB fixture correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`mongodb`_

**--list-type=apps always benchmarks an empty result**

The option is accepted, but `BuildSubscriptionListFixtures` creates only channel and DM subscriptions, while the changed integration test confirms the `apps` aggregation returns no rows because there are no subscribed botDM fixtures. Running with `--list-type=apps` therefore makes every MongoDB query return an empty page, which the collector counts as failure and excludes from latency samples, so this advertised mode cannot produce a capacity result. Either seed subscribed botDM rows for selectable accounts or remove `apps` from validation, help text, documentation, and tests.

<details>

```
In `tools/loadgen/maxrps.go` around line 308: The option is accepted, but `BuildSubscriptionListFixtures` creates only channel and DM subscriptions, while the changed integration test confirms the `apps` aggregation returns no rows because there are no subscribed botDM fixtures. Running with `--list-type=apps` therefore makes every MongoDB query return an empty page, which the collector counts as failure and excludes from latency samples, so this advertised mode cannot produce a capacity result. Either seed subscribed botDM rows for selectable accounts or remove `apps` from validation, help text, documentation, and tests.

```

</details>

---

tools/loadgen/subscriptionlist_seed.go
```diff
@@ -0,0 +1,92 @@
...
+		if !ok {
+			continue
+		}
+		s.Open = true
+		s.RoomType = room.Type
+		s.Name = subListRowName(room, membersByRoom[s.RoomID], s.User.Account)
+
+		behind := time.Duration(r.Int63n(int64(subListLastSeenSpread)))
+		lastSeenAt := room.LastMsgAt.Add(-behind).UTC()
+		s.LastSeenAt = &lastSeenAt
```
julianshen: _🎯 MongoDB fixture correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`mongodb`_

**Last-seen timestamps make nearly every subscription unread**

Every subscription's `lastSeenAt` is generated at or before its room's activity timestamp. Only an `Int63n` result of exactly zero avoids placing it earlier, which is negligible over a nanosecond-sized 14-day range. Consequently, the aggregation's `hasUnread` comparison receives effectively no read rows, contrary to this workload's stated goal of exercising a mix of read and unread results. Generate some timestamps equal to the latest activity and place the remainder behind it.

<details>

> ‼️ **IMPORTANT**

```suggestion
		lastSeenAt := room.LastMsgAt.UTC()
		if r.Intn(2) == 0 {
			behind := time.Duration(r.Int63n(int64(subListLastSeenSpread)))
			lastSeenAt = room.LastMsgAt.Add(-behind).UTC()
		}
		s.LastSeenAt = &lastSeenAt
```

</details>

<details>

```
In `tools/loadgen/subscriptionlist_seed.go` around line 71: Every subscription's `lastSeenAt` is generated at or before its room's activity timestamp. Only an `Int63n` result of exactly zero avoids placing it earlier, which is negligible over a nanosecond-sized 14-day range. Consequently, the aggregation's `hasUnread` comparison receives effectively no read rows, contrary to this workload's stated goal of exercising a mix of read and unread results. Generate some timestamps equal to the latest activity and place the remainder behind it.

```

</details>

---

julianshen (commented):
### tGDBot review of `cda6e4e` — 13 findings in 13 files

10 🟠 warning · 3 🔵 suggestion
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/469#issuecomment-5529060655), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

tools/loadgen/maxrps_subscriptionlist.go
```diff
@@ -0,0 +1,158 @@
...
+
+// runSubscriptionListFor runs gen.Run for d (or until ctx cancels), then stops
+// it and waits for all in-flight requests to drain. Mirrors
+// maxrps_roomread.go's runRoomReadFor for this generator type.
+func runSubscriptionListFor(ctx context.Context, gen *subscriptionListGenerator, d time.Duration) error {
+	genCtx, cancel := context.WithCancel(ctx)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
```
julianshen, line 139: _🏷️ nats-lifecycle_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`nats`, `performance-and-concurrency` (2 rules)_

**Window shutdown cancels and erases the slowest NATS requests**

At every warmup or hold boundary, `cancel()` cancels the same `genCtx` passed into all in-flight `Requester.Request` calls before `wg.Wait()` runs. `doList` then deliberately discards any resulting error when `ctx.Err() != nil`. Consequently, delayed or missing responders at the end of a measurement window are aborted and omitted from both `AttemptedOps` and `FailedOps`, censoring exactly the slow tail during overload and potentially overstating sustainable RPS. This is cancellation, not the documented drain. Stop new dispatches separately while allowing already-launched RPCs to finish under `RequestTimeout` and record their outcomes, then wait for them before building the step inputs.

<details>

- **`performance-and-concurrency`**:136: When the warmup or hold timer expires, `cancel()` cancels both the dispatcher and every in-flight `Requester.Request`, because those calls use the same `genCtx`. `doList` then suppresses their cancellation errors, so up to `MaxInFlight` already-dispatched requests disappear from `AttemptedOps` and the latency/error samples. Near saturation this preferentially drops the slow tail and can make the SLO result look better than the actual hold. Stop new dispatches with a separate signal while allowing existing requests to finish under the outer context and their request timeout; only outer-context cancellation should abort them.

</details>

<details>

```
In `tools/loadgen/maxrps_subscriptionlist.go` around line 133: At every warmup or hold boundary, `cancel()` cancels the same `genCtx` passed into all in-flight `Requester.Request` calls before `wg.Wait()` runs. `doList` then deliberately discards any resulting error when `ctx.Err() != nil`. Consequently, delayed or missing responders at the end of a measurement window are aborted and omitted from both `AttemptedOps` and `FailedOps`, censoring exactly the slow tail during overload and potentially overstating sustainable RPS. This is cancellation, not the documented drain. Stop new dispatches separately while allowing already-launched RPCs to finish under `RequestTimeout` and record their outcomes, then wait for them before building the step inputs.

```

</details>

---

tools/loadgen/subscriptionlist_generator.go
```diff
@@ -0,0 +1,183 @@
...
+	if len(parsed.Subscriptions) == 0 {
+		g.cfg.Collector.RecordEmptyPage(latency)
+		return
+	}
+	g.cfg.Collector.RecordSample(SubscriptionListSample{
+		Latency: latency,
+		Rows:    len(parsed.Subscriptions),
+		HasMore: parsed.HasMore,
+	})
+}
```
julianshen, line 220: _⚡ performance_ | _🟠 Warning_ | _⚡ Quick fix_ | _`performance-and-concurrency`_

**Successful subscription pages are parsed twice.**

Every successful response is inspected by `errcode.Parse` and then decoded again by `json.Unmarshal`. With the default 200-row, include-last-message response potentially approaching 128 KB and a 5k RPS step, this duplicate parsing can consume substantial load-generator CPU. Parsing also occurs before `requestOne` releases its in-flight slot, so it can cause saturation or emit underruns below the service's actual capacity. Decode the success response first and invoke `errcode.Parse` only when the expected collection is absent.

<details>

> ‼️ **IMPORTANT**

```suggestion
func (g *subscriptionListGenerator) classifyReply(reply []byte, latency time.Duration) {
	var parsed soakSubscriptionListResponse
	if err := json.Unmarshal(reply, &parsed); err != nil {
		g.cfg.Collector.RecordBadReply(latency)
		return
	}
	if parsed.Subscriptions != nil {
		if len(parsed.Subscriptions) == 0 {
			g.cfg.Collector.RecordEmptyPage(latency)
			return
		}
		g.cfg.Collector.RecordSample(SubscriptionListSample{
			Latency: latency,
			Rows:    len(parsed.Subscriptions),
			HasMore: parsed.HasMore,
		})
		return
	}
	if _, ok := errcode.Parse(reply); ok {
		g.cfg.Collector.RecordError(errClassReply, latency)
		return
	}
	g.cfg.Collector.RecordBadReply(latency)
}
```

</details>

<details>

```
In `tools/loadgen/subscriptionlist_generator.go` around line 152: Every successful response is inspected by `errcode.Parse` and then decoded again by `json.Unmarshal`. With the default 200-row, include-last-message response potentially approaching 128 KB and a 5k RPS step, this duplicate parsing can consume substantial load-generator CPU. Parsing also occurs before `requestOne` releases its in-flight slot, so it can cause saturation or emit underruns below the service's actual capacity. Decode the success response first and invoke `errcode.Parse` only when the expected collection is absent.

```

</details>

---

tools/loadgen/subscriptionlist_generator.go
```diff
@@ -0,0 +1,183 @@
...
+		g.cfg.Collector.RecordBadReply(latency)
+		return
+	}
+	if len(parsed.Subscriptions) == 0 {
+		g.cfg.Collector.RecordEmptyPage(latency)
+		return
+	}
+	g.cfg.Collector.RecordSample(SubscriptionListSample{
+		Latency: latency,
+		Rows:    len(parsed.Subscriptions),
```
julianshen, line 217: _🏷️ silent-failure_ | _🟠 Warning_ | _⚡ Quick fix_ | _`silent-failure`_

**Malformed subscription rows are recorded as successful replies**

After checking only that the collection is present and nonempty, this records every row as a success without validating its required fields. For example, `{"subscriptions":[{}],"hasMore":false}` decodes and contributes a successful low-latency sample with zero FailedOps even though a client cannot use the row. This masks a response-contract failure much like the explicitly rejected top-level `{}` case. Validate required fields such as the room identifier for every returned subscription and call RecordBadReply before RecordSample when any row is malformed.

<details>

```
In `tools/loadgen/subscriptionlist_generator.go` around line 171: After checking only that the collection is present and nonempty, this records every row as a success without validating its required fields. For example, `{"subscriptions":[{}],"hasMore":false}` decodes and contributes a successful low-latency sample with zero FailedOps even though a client cannot use the row. This masks a response-contract failure much like the explicitly rejected top-level `{}` case. Validate required fields such as the room identifier for every returned subscription and call RecordBadReply before RecordSample when any row is malformed.

```

</details>

---

tools/loadgen/subscriptionlist_seed.go
```diff
@@ -0,0 +1,92 @@
...
+		s := &f.Subscriptions[i]
+		room, ok := roomByID[s.RoomID]
+		if !ok {
+			continue
+		}
+		s.Open = true
+		s.RoomType = room.Type
+		s.Name = subListRowName(room, membersByRoom[s.RoomID], s.User.Account)
+
+		behind := time.Duration(r.Int63n(int64(subListLastSeenSpread)))
```
julianshen: _🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**The timestamp spread makes virtually every row unread.**

`LastSeenAt` is always set to `LastMsgAt` minus a nonnegative random duration. Consequently, nearly every subscription has activity strictly after its last-seen timestamp, contrary to the stated goal of producing a realistic mix of read and unread rows. Generate both read timestamps (equal to or later than the room activity) and unread timestamps (earlier), and add a test asserting that a sufficiently large fixture contains both states.

<details>

```
In `tools/loadgen/subscriptionlist_seed.go` around line 71: `LastSeenAt` is always set to `LastMsgAt` minus a nonnegative random duration. Consequently, nearly every subscription has activity strictly after its last-seen timestamp, contrary to the stated goal of producing a realistic mix of read and unread rows. Generate both read timestamps (equal to or later than the room activity) and unread timestamps (earlier), and add a test asserting that a sufficiently large fixture contains both states.

```

</details>

---

tools/loadgen/maxrps_subscriptionlist.go
```diff
@@ -0,0 +1,158 @@
...
+		},
+	}
+}
+
+// subscriptionListWorkload drives subscription.list requests at a given RPS. As
+// with the other workloads the natsutil connection and metrics server are
+// captured by the cleanup closure, not stored on the struct.
+type subscriptionListWorkload struct {
+	cfg                *config
+	preset             *Preset
```
julianshen: _🏷️ dead-code_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`dead-code`_

**The workload retains two fields that it never reads.**

`preset` and `metrics` are stored on `subscriptionListWorkload`, but none of its methods reads either field. The preset is used directly to construct fixtures, while the HTTP server retains the metrics handler independently. Remove these fields and their keyed initializer entries; keep the constructor-local `metrics` variable for creating the handler.

<details>

```
In `tools/loadgen/maxrps_subscriptionlist.go` around line 52: `preset` and `metrics` are stored on `subscriptionListWorkload`, but none of its methods reads either field. The preset is used directly to construct fixtures, while the HTTP server retains the metrics handler independently. Remove these fields and their keyed initializer entries; keep the constructor-local `metrics` variable for creating the handler.

```

</details>

---

tools/loadgen/main.go
```diff
@@ -505,6 +509,50 @@ func runTeardownRoomRead(ctx context.Context, cfg *config, preset string, seed i
...
+	}
+	slog.Info("seed complete (subscription-list)",
+		"preset", p.Name,
+		"users", len(fixtures.Users),
+		"rooms", len(fixtures.Rooms),
+		"subs", len(fixtures.Subscriptions))
+	return 0
+}
+
+func runTeardownSubscriptionList(ctx context.Context, cfg *config, preset string, seed int64) int {
```
julianshen: _🏷️ dead-code_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`dead-code`_

**The subscription-list teardown seed parameter is ignored.**

`runTeardown` passes the parsed seed into `runTeardownSubscriptionList`, but that function never reads it: teardown validates only the preset and then calls the global `Teardown`. Remove the `seed` parameter and the corresponding argument at the call site so the signature does not imply seed-dependent cleanup.

<details>

```
In `tools/loadgen/main.go` around line 537: `runTeardown` passes the parsed seed into `runTeardownSubscriptionList`, but that function never reads it: teardown validates only the preset and then calls the global `Teardown`. Remove the `seed` parameter and the corresponding argument at the call site so the signature does not imply seed-dependent cleanup.

```

</details>

---

tools/loadgen/maxrps_subscriptionlist.go
```diff
@@ -0,0 +1,158 @@
...
+// there is no consumer queue and Pending stays empty; the single
+// "subscription-list" latency series gates.
+//
+// Empty pages join the failure count. Every seeded account owns subscriptions,
+// so a zero-row reply means the fixtures or the list type are wrong — and since
+// an empty page is also the fastest possible reply, scoring it as a success
+// would let a misconfigured run report a record-breaking ramp.
+func buildSubscriptionListInputs(targetRPS int, hold time.Duration, c *SubscriptionListCollector) rpsStepInputs {
+	samples := c.Samples()
+	failed := c.TimeoutErrors() + c.ReplyErrors() + c.BadReplyCount() + c.EmptyPageCount()
```
julianshen, line 33: _🏷️ observability_ | _🔵 Suggestion_ | _🏗️ Heavy lift_ | _`tgd-review`_

**Collected page-size statistics never reach the report.**

`SubscriptionListSample` records `Rows` and `HasMore`, but this conversion retains only latency before the local collector becomes unreachable. As a result, `TotalRows`, `MeanRows`, and `HasMoreCount` cannot substantiate the comment that the report shows the page size actually measured; a nonempty but unexpectedly tiny page can still yield an apparently valid ramp. Propagate these statistics into the step/report output or emit a workload summary, with a test covering the reported values.

<details>

```
In `tools/loadgen/maxrps_subscriptionlist.go` around line 33: `SubscriptionListSample` records `Rows` and `HasMore`, but this conversion retains only latency before the local collector becomes unreachable. As a result, `TotalRows`, `MeanRows`, and `HasMoreCount` cannot substantiate the comment that the report shows the page size actually measured; a nonempty but unexpectedly tiny page can still yield an apparently valid ramp. Propagate these statistics into the step/report output or emit a workload summary, with a test covering the reported values.

```

</details>

---

tools/loadgen/maxrps.go
```diff
@@ -295,6 +299,33 @@ func runMaxRPS(ctx context.Context, cfg *config, args []string) int {
...
+		}
+		slw, clean, err := newSubscriptionListWorkload(ctx, cfg, &p, *seed, subscriptionListParams{
+			ListType: *listType, Limit: *listLimit,
+			IncludeLastMessage: includeLastMessage, RequestTimeout: *requestTimeout,
+		})
+		if err != nil {
+			slog.Error("init subscription-list workload", "error", err)
+			return 1
+		}
+		w, cleanup, presetID = slw, clean, p.Name
```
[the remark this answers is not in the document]
hmchangw: Confirmed and fixed in fa39b7c.

You're right, and the gap was worse than an unused option: `subscriptionlist_integration_test.go` already pinned "apps is empty because the seeder writes no botDM rows" — so the behaviour was documented and the flag was left accepting the value anyway. Every `apps` request would return an empty page, which the collector counts as a failure with no latency sample, so a ramp would have reported a 100% failure rate against a healthy service.

I took the second of your two options. Seeding subscribed `botDM` rows needs bot users and app records to make the `lookupApps` enrichment meaningful — a different fixture family, and more than this PR should grow to carry.

Rather than deleting `apps` from the accepted set, the two facts are now pinned separately:

- `validSubscriptionListTypes` still mirrors user-service's `validListTypes` (`current`/`rooms`/`apps`) — the service contract.
- `workloadSupportedListTypes` (`current`/`rooms`) is what these fixtures can serve.

Validation keys on the second, and the error names the actual reason so nobody reads it as "the service doesn't support apps":

```
--list-type=apps is not supported by this workload: its fixtures seed no botDM rows, so every page would come back empty
```

An unknown value still gets the plain `must be one of current|rooms`. A unit test asserts the fixtures contain no botDM rows across all three presets, so the guard can be lifted the day that stops being true.

---

tools/loadgen/subscriptionlist_generator.go
hmchangw: Confirmed and fixed in f505850. This was the most consequential of the batch.

I traced it before acting, because `pacedDispatch`'s doc comment claims it "drains in-flight work up to drainGracePeriod" and I wanted to know whether that drain was real. It isn't. `pacedDispatchRateWithTicks` does wait up to 5s on `wg.Wait()` — but the in-flight `do(ctx)` calls hold the *same* cancelled context, so their NATS requests abort immediately and `doList` discards the result. The wait is nominal: it waits on goroutines that already bailed.

So at every warmup and hold boundary, up to `MaxInFlight` (default 200) admitted requests vanished from both `AttemptedOps` and `FailedOps`. You're right that the bias is directional rather than random — the fast ones already completed, so what gets erased is the tail, and in-flight count is highest exactly at the steps that decide PASS/TRIP.

Fixed as you described: the config now carries a `RunCtx`, requests execute under it, and the per-window context stops dispatch alone. Genuine run cancellation still aborts and discards, which is what the `ctx.Err()` check remains for. Two tests pin the halves apart — a dispatch-cancelled request is still measured, an outer-cancelled one is not — and they replaced an earlier test of mine that had encoded the buggy behaviour as correct.

Worth flagging for the maintainers rather than fixing here: `runRoomReadFor` and `runThreadReadFor` have the identical structure, so the same bias applies to those workloads. I've left them alone as out of scope for this PR.

---

tools/loadgen/subscriptionlist_seed.go
hmchangw: Confirmed and fixed in f505850.

I measured it before changing anything: on the `realistic` preset the split was **0 caught-up / 2823 unread**. Exactly as you say — `Int63n` is non-negative, so `lastMsgAt - offset` is at or before the activity for every row, and only an exact-zero draw over a nanosecond-range would land equal. The `hasUnread` comparison resolved the same way every time and the caught-up branch never executed. My own comment claimed "a realistic mix of read and unread rows", which was simply false.

Roughly 40% are now seeded at or after the room's activity via a `subListCaughtUpShare` constant, the rest behind it. I capped the ahead-offset at an hour rather than letting it use the full 14-day spread, so a caught-up member cannot be dragged past `now` and out of the activity window.

I took your and the parallel `tgd-review` comment's suggestion to add a test: it asserts both states are present *and* that neither is under 20% of rows, so a future change that leaves one branch technically-nonzero-but-vestigial still fails.

---

tools/loadgen/subscriptionlist_generator.go
hmchangw: Agreed and applied in f505850, essentially your suggested restructure.

The ordering point is the one I'd missed: an errcode envelope is valid JSON that decodes into the response type leaving `Subscriptions` nil, so decoding first loses nothing — the nil collection is exactly the signal that says "now go ask `errcode.Parse`". Success replies, which are the overwhelming majority and by far the largest, are now parsed once.

Your note about *where* the cost lands is the part that made this worth fixing rather than filing: the parse happens inside `do()`, before the semaphore slot is released, so wasted CPU there converts directly into saturation and emit-underrun counts and would have been read as the service hitting its ceiling. That is the same class of problem as the per-request body marshalling already hoisted out of this path.

One deviation from your snippet: I kept `json.Unmarshal` failure as an immediate bad reply before the nil-collection branch, since a body that is not JSON at all cannot be an errcode envelope either.

---

tools/loadgen/subscriptionlist_generator.go
hmchangw: Fair, and fixed in f505850.

The inconsistency you're pointing at is real: I explicitly rejected top-level `{}` as a contract violation on the grounds that a reply a client cannot consume must not read as a served request, then accepted `{"subscriptions":[{}]}` — which fails the same test, and is likewise among the fastest replies possible, so it would have contributed a fast successful sample.

`classifyReply` now rejects the page as a bad reply if any row is missing `roomId`, checked after the empty-page branch so a legitimately empty sidebar is still distinguished. Table test covers a row without a room, a mixed good/bad page, and a well-formed page.

I stopped at `roomId` rather than validating the full row. It is the one field with no legitimate empty value and the one the client keys everything else off; going further would start asserting a payload contract this workload has no expectation for, which is the line `soakSubscriptionRow` already draws deliberately.

---

tools/loadgen/maxrps_subscriptionlist.go
hmchangw: Agreed and addressed in f505850.

`rpsStepInputs` has no field for per-class counts and adding one would change a struct shared by nine workloads, so I took your "at minimum" option: the per-step diagnostic now emits all four classes separately —

```
subscription-list step shape  rps=2000 pages=59812 mean_rows=8.4 has_more_pages=0
  timeout_errors=188 service_errors=0 bad_replies=0 empty_pages=0
```

Your framing is the right one and I've put it in the comment: a broker timeout, a protocol incompatibility and bad fixtures each need a different response, and summing them into `FailedOps` renders all three as the same number. That is especially bad here given the NATS payload-ceiling failure mode documented in the README — an oversized page manifests purely as timeouts, and the timeout count is now the thing that distinguishes it from a genuine capacity limit.

The line also emits when a step produced only failures, so a fully-failing step is not as quiet as one that never ran.

---

tools/loadgen/subscriptionlist_collector.go
hmchangw: Correct as posted — you reviewed `cda6e4e`, where the row statistics genuinely had no consumer and the collector was discarded at the end of every step.

That was fixed in 7302123, before this comment landed, and for the reason your sibling finding gives: a non-empty but unexpectedly tiny page could still yield an apparently valid ramp. `RunStep` now calls `logStepPageShape` before the collector goes out of scope, and f505850 widened that line to carry the four failure classes as well.

`TotalRows` had remained genuinely unused, so it is now exercised only by the collector's own tests rather than by production code. I've kept it: it is the accumulator `MeanRows` divides, so removing the accessor would not remove the field, and it is two lines. Say the word if you'd rather it went.

---

tools/loadgen/maxrps_subscriptionlist.go
hmchangw: Correct — both removed in f505850, along with their initializer entries. The constructor-local `metrics` variable stays, since the HTTP handler needs it.

Worth noting for anyone comparing files: `roomReadWorkload` stores the same two fields and does not read them either. I copied that struct when building this workload. I've only cleaned up the new file — changing the sibling is outside this PR — so the duplication you're seeing across the two is pre-existing rather than something this change introduced.

---

tools/loadgen/main.go
hmchangw: Correct — removed in f505850, along with the argument at the call site, plus a comment saying why teardown takes no seed: it drops the seeded collections wholesale, so unlike the seeding side it does not depend on which fixtures were generated.

Same caveat as the dead-fields thread: `runTeardownRoomRead` and `runTeardownMembers` also take a `seed` they never read, so my new function now differs slightly from its siblings. I judged a dead parameter in new code worth removing over matching a pattern that is itself the defect; happy to revert for consistency if a maintainer prefers the uniform signatures.

---

tools/loadgen/maxrps_subscriptionlist_test.go
```diff
@@ -0,0 +1,161 @@
...
+func TestLogStepPageShape_ReportsMeasuredPages(t *testing.T) {
+	c := NewSubscriptionListCollector()
+	c.RecordSample(SubscriptionListSample{Latency: time.Millisecond, Rows: 4, HasMore: true})
+	c.RecordSample(SubscriptionListSample{Latency: time.Millisecond, Rows: 2})
+	c.RecordEmptyPage(time.Millisecond)
+
+	assert.NotPanics(t, func() { logStepPageShape(500, c) })
+	assert.InDelta(t, 3.0, c.MeanRows(), 0.001)
+	assert.Equal(t, 1, c.HasMoreCount())
+	assert.Equal(t, 1, c.EmptyPageCount())
```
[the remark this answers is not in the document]
hmchangw: Right, and fixed in f505850. The test asserted the collector I had just populated by hand — it would have passed if `logStepPageShape` were an empty function body, which makes it worthless for the one thing that function does.

It now swaps in a JSON `slog` handler, runs the call, and asserts the emitted record's attributes: `rps`, `pages`, `mean_rows`, `has_more_pages`, and the four failure-class counts f505850 added. Two more tests use the same helper to pin the emit/don't-emit boundary — silent when the step produced nothing at all, but still reporting when a step produced only failures, so a fully-failing step is not as quiet as one that never ran.

This is the second test of mine in this PR that asserted around the behaviour instead of on it (the earlier one claimed to check account selection and only checked the subject prefix). Both are now assertions on the actual output.

---

tools/loadgen/README.md
```diff
@@ -537,6 +537,91 @@ make -C tools/loadgen/deploy teardown-roomread PRESET=medium
...
+make -C tools/loadgen/deploy up
+make -C tools/loadgen/deploy seed-sublist PRESET=realistic
+make -C tools/loadgen/deploy run-max-rps WORKLOAD=subscription-list PRESET=realistic
+```
+
+`realistic` is not incidental here — see the preset note below.
+
+Override the ramp with `STEPS` (default `200,500,1000,2000,5000`):
+
+```
```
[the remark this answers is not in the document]
hmchangw: Fixed in f505850 — all three opening fences in the subscription-list section now carry `sh`.

---

tools/loadgen/README.md
```diff
@@ -781,10 +871,13 @@ make -C tools/loadgen/deploy run-max-rps WORKLOAD=history PRESET=history-medium
...
-| `--request-timeout` | `5s` | **history / read-receipt / room-read / thread-read / login / search**: per-request reply timeout |
+| `--steps` | messages `500,1k,2k,5k,10k` / history+read-receipt+subscription-list `200,500,1k,2k,5k` / login `50,100,200,500,1k` / search `100,200,500,1k,2k` | explicit ordered RPS list; `k` suffix = ×1000 |
+| `--request-timeout` | `5s` | **history / read-receipt / room-read / thread-read / subscription-list / login / search**: per-request reply timeout |
+| `--list-type` | `current` | **subscription-list only**: `current` or `rooms`. `apps` is rejected — user-service serves it, but it matches subscribed `botDM` rows alone and these fixtures seed none, so every page would come back empty |
+| `--list-limit` | `200` | **subscription-list only**: page size; see the NATS payload-ceiling note above |
+| `--include-last-message` | `true` | **subscription-list only**: request the per-room `previewMessage` enrichment; `=false` isolates the query from the enrichment cost |
```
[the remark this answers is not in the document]
hmchangw: Fair — fixed in f505850. The overview now reads "by default every request is…" and names the three flags that change the shape.

One clarification kept in the text: `offset` is genuinely fixed at 0, not a default. There is no flag for it, because the workload models a cold open, which asks for the first page — so listing it beside the three configurable values would have been the mirror-image inaccuracy.

---

tools/loadgen/subscriptionlist_seed.go
hmchangw: Confirmed and fixed in f505850 — same finding as the `mongodb`-tagged thread on this file, answered there too.

Measured before changing anything: **0 caught-up / 2823 unread** on the `realistic` preset. Exactly your diagnosis — `Int63n` is non-negative, so every row landed at or before its room's activity and the caught-up branch never executed. My own comment claimed a "realistic mix", which was false.

About 40% are now seeded at or after the room's activity, the rest behind it. I took your suggestion on the test specifically: it asserts both states are present **and** that neither falls under 20% of rows, so a change that leaves one branch technically-nonzero-but-vestigial still fails rather than passing on a single stray row.

---

tools/loadgen/main.go
```diff
@@ -505,6 +509,52 @@ func runTeardownRoomRead(ctx context.Context, cfg *config, preset string, seed i
...
+		fmt.Fprintf(os.Stderr, "unknown preset: %s\n", preset)
+		return 2
+	}
+	db, _, cleanup, err := connectStores(ctx, cfg)
+	if err != nil {
+		return 1
+	}
+	defer cleanup()
+	// No TeardownRoomKeys: runSeedSubscriptionList writes no room keys.
+	if err := Teardown(ctx, db); err != nil {
```
julianshen, line 577: _🏷️ mongodb-lifecycle_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`mongodb`_

**Subscription-list teardown deletes unrelated fixture data**

The workload-specific teardown validates and logs `preset` but then calls the global `Teardown(ctx, db)`, which the preceding comment says drops seeded collections wholesale. If another workload has fixtures in the same database, running `teardown --workload=subscription-list` removes those users, rooms, subscriptions, and related data too, breaking subsequent or concurrent runs. Scope deletion to IDs generated from this preset and seed, or tag this workload's documents and delete by that tag; thread the seed into teardown if deterministic IDs are used.

<details>

```
In `tools/loadgen/main.go` around line 550: The workload-specific teardown validates and logs `preset` but then calls the global `Teardown(ctx, db)`, which the preceding comment says drops seeded collections wholesale. If another workload has fixtures in the same database, running `teardown --workload=subscription-list` removes those users, rooms, subscriptions, and related data too, breaking subsequent or concurrent runs. Scope deletion to IDs generated from this preset and seed, or tag this workload's documents and delete by that tag; thread the seed into teardown if deterministic IDs are used.

```

</details>

---

tools/loadgen/subscriptionlist_collector.go
```diff
@@ -0,0 +1,154 @@
...
+	return &SubscriptionListCollector{errors: map[errClass]int{}}
+}
+
+// RecordSample stores one completed-call sample and folds in its row stats.
+func (c *SubscriptionListCollector) RecordSample(s SubscriptionListSample) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.samples = append(c.samples, s)
+	c.totalRows += s.Rows
+	if s.HasMore {
```
julianshen, line 47: _⚡ memory-exhaustion_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`oom`_

**Successful requests accumulate an unbounded sample tape**

Every successful RPC appends a 24-byte sample for the entire hold window. A 5,000-RPS one-hour step retains roughly 432 MB here; `Samples()` then copies that tape and `subscriptionListLatencies` allocates another slice, pushing peak memory beyond 1 GB before reporting. Because step rates and hold durations are configurable, a reachable long/high-rate benchmark can OOM or evict the loadgen container and invalidate the ramp. Use an online bounded quantile structure, or enforce a hard sample/window limit and mark overflowed steps inconclusive.

<details>

```
In `tools/loadgen/subscriptionlist_collector.go` around line 47: Every successful RPC appends a 24-byte sample for the entire hold window. A 5,000-RPS one-hour step retains roughly 432 MB here; `Samples()` then copies that tape and `subscriptionListLatencies` allocates another slice, pushing peak memory beyond 1 GB before reporting. Because step rates and hold durations are configurable, a reachable long/high-rate benchmark can OOM or evict the loadgen container and invalidate the ramp. Use an online bounded quantile structure, or enforce a hard sample/window limit and mark overflowed steps inconclusive.

```

</details>

---

tools/loadgen/maxrps_subscriptionlist.go
```diff
@@ -0,0 +1,192 @@
...
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		_ = gen.Run(genCtx)
+	}()
+	err := waitOrCancel(ctx, d)
+	cancel()
+	wg.Wait()
+	return err
```
julianshen, line 145: _🏷️ error-handling_ | _🟠 Warning_ | _⚡ Quick fix_ | _`silent-failure`, `tgd-review` (2 rules)_

**Generator failures are discarded and reported as completed steps.**

If `gen.Run` returns an error—for example from its explicit nonpositive-rate or body-marshalling checks—line 139 discards it. `waitOrCancel` then returns nil when the window expires, so `RunStep` receives an empty measurement instead of the generator failure. Capture the goroutine result and return it after cancellation and draining; optionally select on it to fail immediately.

<details>

> ‼️ **IMPORTANT**

```suggestion
func runSubscriptionListFor(ctx context.Context, gen *subscriptionListGenerator, d time.Duration) error {
	genCtx, cancel := context.WithCancel(ctx)
	var wg sync.WaitGroup
	runErrCh := make(chan error, 1)
	wg.Add(1)
	go func() {
		defer wg.Done()
		runErrCh <- gen.Run(genCtx)
	}()
	waitErr := waitOrCancel(ctx, d)
	cancel()
	wg.Wait()
	if waitErr != nil {
		return waitErr
	}
	return <-runErrCh
}
```

</details>

<details>

- **`tgd-review`**:136: `subscriptionListGenerator.Run` can return validation or body-construction errors, but the goroutine drops that result. On that path, `runSubscriptionListFor` waits for the window and returns only `waitOrCancel`'s nil result, so `RunStep` reports an empty completed measurement instead of failing. Capture the generator error and return it after the goroutine finishes.

<details>

```text
	var wg sync.WaitGroup
	var runErr error
	wg.Add(1)
	go func() {
		defer wg.Done()
		runErr = gen.Run(genCtx)
	}()
	err := waitOrCancel(ctx, d)
	cancel()
	wg.Wait()
	if runErr != nil {
		return runErr
	}
	return err
```

</details>


</details>

<details>

```
In `tools/loadgen/maxrps_subscriptionlist.go` around line 133: If `gen.Run` returns an error—for example from its explicit nonpositive-rate or body-marshalling checks—line 139 discards it. `waitOrCancel` then returns nil when the window expires, so `RunStep` receives an empty measurement instead of the generator failure. Capture the goroutine result and return it after cancellation and draining; optionally select on it to fail immediately.

```

</details>

---

julianshen (commented):
### tGDBot review of `f505850` — 6 findings in 13 files

5 🟠 warning · 1 🔵 suggestion
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/469#issuecomment-5529060655), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

tools/loadgen/subscriptionlist_seed.go
```diff
@@ -0,0 +1,135 @@
...
+// seededLastSeenAt places a member either at/after the room's activity (read) or
+// behind it (unread), so a page carries both. Ahead is capped well short of the
+// activity spread so a caught-up member cannot be dragged past `now`.
+func seededLastSeenAt(r *rand.Rand, lastMsgAt time.Time) *time.Time {
+	offset := time.Duration(r.Int63n(int64(subListLastSeenSpread)))
+	at := lastMsgAt.Add(-offset).UTC()
+	if r.Float64() < subListCaughtUpShare {
+		// Equal counts as read, so a zero offset is a legitimate draw here.
+		at = lastMsgAt.Add(offset % time.Hour).UTC()
+	}
```
julianshen, line 89: _🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**Caught-up fixtures can receive future read timestamps.**

`lastMsgAt` can be arbitrarily close to `now`, while caught-up rows add `offset % time.Hour`. Whenever that offset exceeds the room's age, `LastSeenAt` lands in the future despite the preceding claim that it cannot pass `now`. Equality already counts as read, so it can exercise the caught-up branch without creating future timestamps.

<details>

> ‼️ **IMPORTANT**

```suggestion
		at = lastMsgAt.UTC()
```

</details>

<details>

```
In `tools/loadgen/subscriptionlist_seed.go` around line 89: `lastMsgAt` can be arbitrarily close to `now`, while caught-up rows add `offset % time.Hour`. Whenever that offset exceeds the room's age, `LastSeenAt` lands in the future despite the preceding claim that it cannot pass `now`. Equality already counts as read, so it can exercise the caught-up branch without creating future timestamps.

```

</details>

---

tools/loadgen/subscriptionlist_collector.go
```diff
@@ -0,0 +1,154 @@
...
+		c.hasMore++
+	}
+}
+
+// RecordError tallies a per-class transport/reply error.
+func (c *SubscriptionListCollector) RecordError(class errClass, _ time.Duration) {
```
julianshen, line 53: _🏷️ dead-code_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`dead-code`_

**Error-recording methods discard every supplied latency.**

`RecordError`, `RecordBadReply`, and `RecordEmptyPage` explicitly ignore their `time.Duration` arguments. The generator nevertheless measures and forwards `latency` on each failure path, but the collector never stores or reports it. Remove these parameters and update callers, avoiding `time.Since` on transport-error paths when no latency is needed, or retain the values in collector metrics if failure latency is intended to be consumed.

<details>

```
In `tools/loadgen/subscriptionlist_collector.go` around line 53: `RecordError`, `RecordBadReply`, and `RecordEmptyPage` explicitly ignore their `time.Duration` arguments. The generator nevertheless measures and forwards `latency` on each failure path, but the collector never stores or reports it. Remove these parameters and update callers, avoiding `time.Since` on transport-error paths when no latency is needed, or retain the values in collector metrics if failure latency is intended to be consumed.

```

</details>

---

PR #472, opened 2026-09-04
chore(deps): upgrade o11y SDK to v0.12.0 and wire the Elasticsearch MeterProvider

flywindy:
## What

Upgrades `github.com/flywindy/o11y` from v0.11.0 to v0.12.0 and adapts the one call site the release breaks.

v0.12.0 gives the Elasticsearch integration an SDK-owned `db.client.operation.duration` histogram (o11y ADR 0027), so `elasticsearch.NewClient` now takes a `MeterProvider` after the `TracerProvider` and rejects a nil one — the facade never falls back to the global provider. That is the **only** breaking change in the release.

## Blast radius

Smaller than the import count suggests:

- 107 files import o11y, but the `gin`, `nats`, `mongo`, `cassandra` and `minio` sub-packages are **byte-identical** between the two versions, so the 43 NATS and 10 Gin call sites are untouched.
- The SDK's own `go.mod` is unchanged between versions. After `go mod tidy`, this repo's entire dependency delta is one line in `go.mod` and two in `go.sum`.
- `go build ./...` and `go vet ./...` across the whole repo produced exactly **one** compile error: `pkg/searchengine/factory.go`.

Changes:

- `searchengine.Observability` grows `MeterProvider()`, converging on the shape `mongoutil` and `cassutil` already use. `*o11y.SDK` satisfies it as-is, so the two call sites (`search-service`, `search-sync-worker`) are **unchanged**.
- `newESClient` passes the meter provider through.
- Two test doubles gain the method.

## The one thing reviewers should look at

**The new metric counts some routine 404s as failures, so an error-ratio panel has to exclude them by operation — not by status alone.**

o11y decides "failure" by each client API's own contract. `searchengine` builds the **low-level** client, whose contract is `esapi.Response.IsError` — status > 299. The typed client's exemption (a 404 from `Get`/`Delete`/`Exists` is a normal result) does not apply here. Reading the adapter, ES 404s fall into three groups:

| Group | Operations | Behaviour |
|---|---|---|
| **Routine** — excluded from the ratio | `get` (`GetDoc` → `nil, false, nil`), `indices.get_mapping` (`GetIndexMapping` → `nil, nil`) | `search-service`'s `loadRestricted` reads the per-account user-room doc on every Valkey miss, and a user with no restricted rooms legitimately has none |
| **Real failure** — stays counted | `update_by_query` | Its path carries no `ignore_unavailable`, so a missing index answers 404 and the adapter fails the caller on any non-200 |
| **Cannot 404 on a missing index** | `search`, `indices.put_mapping` | Both paths set `ignore_unavailable=true&allow_no_indices=true` |

This is **not a defect** — nothing broke, and `GetDoc`'s contract to its caller is unchanged. The same 404 has always set the span status to `Error`; v0.11.0 did that too. What is new is that the fact is now aggregatable, so it can mislead a dashboard.

`storage-dependency-metrics.md` §4 sets this out, and §7 gains an Elasticsearch error-ratio query that excludes 404 for the two routine operations only — as a union of two selectors, since a label matcher cannot express "not (op=get and status=404)" — seeded from the denominator so a service with traffic and no errors reads 0% rather than no data. Because that query depends on the exact operation strings, all three groups are pinned by `TestAdapter_RoutineMissIsCountedAsAnError` (table-driven, asserts `db.operation.name`) and `TestAdapter_UpdateByQueryMissIsARealFailure`.

## Cardinality decision

The index label (`db.collection.name`) is left **on**, which is o11y's default. Message indices roll monthly (`searchindex.MessageIndexName`, `{prefix}-{YYYY-MM}`), which is the case ADR 0027 warns about, but exposure is bounded: the label is emitted only for a request addressing exactly one index — a bulk sets its index per action line, and search reads go through a wildcard — and `o11y.WithMaxUniqueCollections` collapses overflow to `other` at the export boundary.

A test pins the label so opting out later (`elasticsearch.WithCollectionMetricLabel(false)`) is a visible decision. **Post-deploy check:** if `db_collection_name="other"` appears on these series, that is the signal to opt out.

## Fixes this release carries that the repo was exposed to

- `log`: `traceId`/`spanId` were nested inside the group on loggers derived with `WithGroup`, breaking top-level log-to-trace correlation. `pkg/logctx/handler.go` delegates `WithGroup`.
- `log`: the SDK handlers mutated a `slog.Record` they did not own, violating the `slog.Handler` contract. Services wrap the SDK handler via `obs.InitWithLoggerHandler`.
- `metrics`: a failed `Init` on the OTLP push path leaked the MeterProvider's export goroutine for the life of the process.
- `profiling`: a Pyroscope endpoint carrying `user:password@host` was written to stdout and the OTLP log pipeline.
- `redis`: `db.client.connection.create_time` records seconds but matched no view, so every dial fell into the first millisecond-scale bucket. `pkg/valkeyutil` benefits directly.

## Verification

**CI is green on the current head `9f605b4` — all 60 checks:** `lint`, `test`, `sast`, and 56 `test-integration` targets, including the three this change actually touches (`pkg/searchengine`, `search-service`, `search-sync-worker`). `b39339a` was green on the same 60 before it.

Run locally before each push:

| Check | Result |
|---|---|
| `go vet ./...` | exit 0 |
| `go vet -tags integration ./...` | exit 0 |
| `make lint` | 0 issues |
| `make test` (`-race`) | 131 packages ok, 0 failures, 0 data races |
| `make sast-gosec` | PASS |

Three things could **not** run in the authoring environment and were called out as open gaps when this PR was opened. CI has since closed all three, so this section is kept only to say how they were covered:

- **Integration tests** — no Docker locally, so they were compiled but not executed. CI ran them; all 56 targets pass.
- **`govulncheck`** — the environment's network policy blocks `vuln.go.dev`. CI's `sast` job passes.
- **`semgrep`** — not installed locally. Same `sast` job, passing.

`pkg/searchengine` unit coverage moves **71.0% → 76.6%**. Still under the 80% floor, but that is a pre-existing gap covered by the integration tests; closing it is out of scope for a dependency upgrade.

## Note on `docs/reviews/`

CLAUDE.md asks that `docs/reviews/` be emptied before opening a PR. This branch adds nothing there — the 39 files are already on `main` — so deleting them would add unrelated churn to a dependency PR. Left alone deliberately; say the word if you want them cleared in a separate change.

## Note on the CodeRabbit docstring warning

CodeRabbit's review found no actionable comments and rates merge risk minimal on `9f605b4`. Its one pre-merge warning is docstring coverage (60% vs an 80% threshold), and I have deliberately not chased it: every function it counts is an **unexported test helper** in a `_test.go` file, the same package's existing helpers (`fakeTransport`, `jsonResponse` in `adapter_test.go`) carry no doc comments either, and this repo ships a `remove_comments` tool whose stated job is deleting exactly the kind of WHAT-comment that would be added. Happy to revisit if a reviewer disagrees.

https://claude.ai/code/session_019pidETKALQRKh8pv4rXmpq


julianshen:
**No actionable comments.** ✅

<details>
<summary>📒 Files reviewed (10)</summary>

* `docs/specs/o11y/o11y-integration.md`
* `go.mod`
* `go.sum`
* `pkg/searchengine/factory.go`
* `pkg/searchengine/integration_test.go`
* `pkg/searchengine/observability.go`
* `pkg/searchengine/observability_test.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

julianshen (commented):
### tGDBot review of `b39339a` — 4 findings in 10 files

4 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/472#issuecomment-5549069776), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

docs/specs/o11y/storage-dependency-metrics.md
```diff
@@ -177,6 +199,20 @@ sum by (db_system_name, service_name) (
   rate(db_client_operation_duration_seconds_count[5m])
 )
 
+# Elasticsearch error ratio. The low-level client counts a routine document
+# miss as a failure, so 404 must be excluded (§4). A label matcher cannot
+# express "not (system=es and status=404)", so Elasticsearch gets its own query
+# rather than a carve-out inside the one above.
+sum by (service_name) (
+  rate(db_client_operation_duration_seconds_count{
+    db_system_name="elasticsearch", error_type!="", error_type!="404"
```
julianshen: _🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`, `elasticsearch`, `silent-failure`, `tgd-review` (4 rules)_

**The error-ratio query hides every Elasticsearch 404**

The matcher `error_type!="404"` excludes 404s from every Elasticsearch operation, not only routine `GetDoc` misses. For example, a search against a missing index returns 404 and fails the caller, but its `db_operation_name="search"` sample is omitted from the numerator, understating real failures. Use `db_operation_name` to exclude 404 only for the Get endpoint, while retaining 404s from other operations; add an assertion for the Get operation label before relying on its exact value.

<details>

- **`silent-failure`**:209: The matcher removes every Elasticsearch 404, although only a missing document in `GetDoc` is established as routine. For example, searching a deleted or misconfigured index also returns 404; those failed requests increase the denominator but not the numerator, potentially showing a 0% error ratio during an outage. Suppress only recognized document-miss outcomes in instrumentation, or retain a separate alert/query covering all 404 responses so genuine failures remain visible.
- **`distributed-system`**:209: The numerator excludes every `error_type="404"`, although only `GetDoc` document misses are established as routine. If a search targets a missing index or alias during rollover or migration, Elasticsearch returns 404, the SDK records a failed `search` operation, and this query removes it; operators can see a zero error ratio while user searches fail. Exclude only 404 samples carrying the `GetDoc` operation label, while retaining 404s from search and other endpoints.
- **`elasticsearch`**:207: When a service has traffic but has emitted no non-404 error since startup, the filtered numerator has no series. PromQL division drops denominator-only labels, so the panel returns no Elasticsearch sample instead of a 0% error ratio. This is reachable because successful samples lack `error_type` and routine 404 samples are explicitly excluded. Seed the numerator from the denominator with zero before dividing.

<details>

```text
(
  sum by (service_name) (
    rate(db_client_operation_duration_seconds_count{
      db_system_name="elasticsearch", error_type!="", error_type!="404"
    }[5m])
  )
  or on (service_name)
  0 * sum by (service_name) (
    rate(db_client_operation_duration_seconds_count{db_system_name="elasticsearch"}[5m])
  )
)
/
sum by (service_name) (
  rate(db_client_operation_duration_seconds_count{db_system_name="elasticsearch"}[5m])
)
```

</details>


</details>

<details>

```
In `docs/specs/o11y/storage-dependency-metrics.md` around line 208: The matcher `error_type!="404"` excludes 404s from every Elasticsearch operation, not only routine `GetDoc` misses. For example, a search against a missing index returns 404 and fails the caller, but its `db_operation_name="search"` sample is omitted from the numerator, understating real failures. Use `db_operation_name` to exclude 404 only for the Get endpoint, while retaining 404s from other operations; add an assertion for the Get operation label before relying on its exact value.

```

</details>

---

flywindy: Confirmed and fixed in 9f605b4 — but the fix is wider than the proposal, because reading the adapter turned up three groups rather than two.

**Routine 404 (excluded):** `get` (`GetDoc` → `nil, false, nil`) and **`indices.get_mapping`** (`GetIndexMapping` → `nil, nil`). The second was missing from §4 entirely, so "exclude 404 only for the Get endpoint" would have started counting a normal missing-mapping check as a failure.

**Real failure (still counted):** `update_by_query`. Its path (`adapter.go:251`) carries no `ignore_unavailable`, so a missing index answers 404 and the adapter fails the caller on any non-200 — exactly the outage the old matcher would have hidden. This is the concrete case the finding is right about.

**Cannot 404 on a missing index:** `search` and `indices.put_mapping` both set `ignore_unavailable=true&allow_no_indices=true` (`adapter.go:358`, `:221`). So the specific example given — a search against a missing index returning 404 — does not occur in this repo. The finding stands on `update_by_query` instead.

The numerator is now the union of two selectors (a label matcher cannot express "not (op=get and status=404)"), excluding 404 for those two operations only. I also took the `elasticsearch` rule's seeding point: it is wrapped in `or on (service_name) 0 * <denominator>`, so a service with traffic and no errors reads 0% instead of no data.

Since the query now depends on the exact operation strings, all three behaviours are pinned by tests rather than left to drift: `TestAdapter_RoutineMissIsCountedAsAnError` is table-driven over both document-miss endpoints and asserts `db.operation.name`, and `TestAdapter_UpdateByQueryMissIsARealFailure` holds the counterexample.

---

hmchangw (approved):
Thanks!

PR #474, opened 2026-09-06
refactor(loadgen): extract soak runtime packages

flywindy:
## Summary

- normalize the extracted soak package APIs so package names carry the domain context and exported identifiers stay concise
- extract the catalog, send, read verification, mutation, search, presence, collector, workload, topology, and run lifecycle responsibilities into focused `internal/soak` packages
- compose production seed, workload, and teardown paths directly from the extracted package APIs while retaining only the root adapters still required by the failure runtime and Prometheus wiring
- land the refactor design and the findings ledger the deferred review threads point at

This is a behavior-preserving refactor. It does not change the client API, persistence schema, or soak configuration contract.

## Review order

The commits are intentionally structured for by-commit review:

1. `3c21ede` — normalize internal soak package APIs
2. `0d97a67` — extract soak message catalog
3. `aa742ae` — extract soak send lane
4. `5e1cad7` — extract soak read verification
5. `3ee8cb5` — extract soak mutation lane
6. `3f69ca9` — extract soak search and presence lanes
7. `5d1358b` — extract soak collector and workload
8. `2f7b4de` — extract soak run lifecycle
9. `82323cf` — integrate extracted soak runtime

The ninth commit moves durable manifest transitions into `internal/soak/run`, rewires production entry points to package APIs, and adds composition contracts that prevent the entry points from returning to the legacy wrappers.

Five commits were added in review and are independently reviewable:

10. `82d7ff5` — suppress G601 on the catalog test's address-taken range (SAST gate)
11. `3f5f3d1` — point the lane-registry guards at the live registry
12. `7ea063f` — suppress math/rand on the four extracted soak bridges (SAST gate)
13. `6f95b77` — drop the presence fixture's inert recorder and candidate count
14. `656db43` — add the refactor design and the findings ledger (docs only)

`3f5f3d1` is the one that changes more than a comment. `soak_workload.go` kept its own copy of the 14-lane registry after the workload package took it over, and both lane assertions were pointed at that copy — which production never reads, since `soak_main.go` builds through `workload.New` and `newSoakWorkload` has no production callers. Adding a lane to the live registry therefore needed no test change, and the package had no lane test of its own. The root copy is deleted, the workload package gained a `Lanes(cfg)` view of its own registry, and the assertions moved onto it. `soak_edge_test.go` keeps the one check only the root can make: the four lanes the failure ledger routes observers by still exist under those exact names — a coupling that had been covered only as a side effect of both copies living in the root.

## Scope boundary

- failure-aware root adapters remain in place because the failure runtime still consumes them
- extracting that runtime and deleting its remaining root compatibility aliases belongs to the follow-up refactor PR
- root integration tests remain as a small cross-package contract layer; package-local tests cover the detailed behavior

## Open review threads are deferred by design, not overlooked

Six review comments are answered but deliberately left unresolved. Each was verified against the source and each describes behavior that predates this PR — `main` has the same code before the extraction moved it — so fixing any of them here would cost the only signal an 8 800-line move has: that a red test means the move broke something rather than that a fix changed something. That rule is the first thing `docs/load-testing/loadgen-findings.md` states, and commit 14 lands the ledger here so the threads point at something real.

All six are recorded with file:line evidence as F11-F15, and each thread names its ledger entry:

| Ledger | Finding | Note |
|---|---|---|
| F11 | Teardown reads the manifest and proceeds with no atomic claim, while `Prepare` writes `StateRunning` unconditionally | raised independently by both bots; one finding |
| F12 | `run.Teardown` does not defend its own `BatchRooms` precondition | not reachable via the CLI, which validates it — listed because the extraction turned a local invariant into an unguarded public API precondition |
| F13 | A capped search page reports an indexed message as missing | |
| F14 | `SearchTerm` retains the whole message body, against its own doc comment | shares one root cause with F13: generated bodies are a single unbroken token |
| F15 | Concurrent reaction toggles can leave the catalog disagreeing with the server | |

Two further comments were fixed here and resolved: the presence fixture's unread recorder and its inert candidate count (`6f95b77`).

## Validation

- make fmt
- `GOOS=linux CGO_ENABLED=0 make lint` — 0 issues
- make test SERVICE=tools/loadgen
- make build SERVICE=tools/loadgen
- make coverage-loadgen-soak
  - aggregate soak coverage: 84.58%
  - core soak coverage: 90.77%
  - `internal/soak` coverage: 91.30%
- make sast
  - gosec: pass
  - repository Semgrep rule tests: pass
  - govulncheck: locally blocked while loading the existing Windows-incompatible `user-service/main.go` `syscall.Kill` path, with a Go 1.25/1.26 tool version mismatch
  - Semgrep scan: locally blocked by the Windows `semgrep-core` `socketpair` failure

The coverage figures predate commits 10-14. Of those, 10, 12 and 13 are comment-only or test-only, 14 is documentation, and 11 moved assertions between packages while adding `Lanes`/`LaneNames` with three consumers.

The four guards touched by `3f5f3d1` were each verified to fail when their defect is introduced, rather than only to pass as written: a removed lane (`extra elements in list A: "read_receipt"`), a lane wired to no action (`lane=presence has no action`), a renamed failure lane, and a rate the root forgets to map (`lane=user_read has no configured rate`).

## Local validation limitations

- `make lint` reaches an unchanged origin/main Windows type-check failure at `user-service/main.go:395`: `undefined: syscall.Kill`, and reports 0 lint issues.
- The combined local SAST run passed gosec and the Semgrep rule tests. govulncheck is blocked by the same Windows type-check failure; the Semgrep scan is blocked by a local `semgrep-core` RPC validation failure. **This gap is what let five blocking Semgrep findings reach CI across two rounds** — one G601 and four `math/rand`, all of them files that crossed into the new packages without the suppression their siblings already carried. Remote CI remains the only gate that runs these.
- Docker integration tests and a manual soak smoke run were not run locally; this PR does not move integration-test ownership. CI's `test-integration (tools)` job passes.

## Follow-ups

- PR 3: extract the failure ledger, WAL, observer contracts, and soak runtime adapter without introducing a soak/failure package cycle. The design document's sequencing argument for that PR has already been overtaken by this one — it assumed the engine's 27 references into failure would need a temporary bridge, and consumer-side interfaces such as `send.Lifecycle` dissolve the coupling instead. Commit 14 records that correction rather than leaving the advice standing.
- `userread` still declares its own `Sample` and `Recorder`, field-identical to `read.Sample` and `read.SampleRecorder`, which is why `soak_userread_adapter.go` exists. `presence`, `search` and `collector` all import `read` instead. `userread` importing `read` compiles with no cycle, so the duplicate and the adapter can both go.
- Around twenty root constructors now have no production callers and are reached only from the retained tests. All but `newSoakWorkload` are single-line delegations, so those tests still exercise the real package code through a forwarder; `newSoakWorkload`'s one divergent piece was the lane registry, removed in `3f5f3d1`.
- The Makefile coverage gates select files by path prefix, so every phase that moves code must update them in the same commit; the design document records this and the `-include` repeatability that PR #460 added for it.

https://claude.ai/code/session_01PxWejgZWgXFehtiFP9YSmZ


julianshen:
**6 findings · 6 inline comments posted** — 🔴 2 blocking · 🟠 2 warning · 🔵 2 suggestion

### Local review memory

- `tools/loadgen/internal/soak/run/teardown.go:101` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟠 Major_ _🏗️ Heavy lift_ **Fence lifecycle transitions before destructive teardown.** If teardown reads StateSeeded , another process can persist StateRunning before teardown deletes the run dat… ([thread](https://github.com/hmchangw/newchat/pull/474#discussion_r3944635244))
- `tools/loadgen/internal/soak/run/teardown.go:121` — @coderabbitai: _🗄️ Data Integrity & Integration_ _🟠 Major_ _⚡ Quick win_ **Reject a non-positive batch size before cleanup.** If cfg.BatchRooms is zero or negative, ChunkRoomIDs returns no batches. Teardown then deletes ownership records and ma… ([thread](https://github.com/hmchangw/newchat/pull/474#discussion_r3944635245))

<details>
<summary>📒 Files reviewed (86)</summary>

* `Makefile`
* `tools/loadgen/internal/soak/catalog/catalog.go`
* `tools/loadgen/internal/soak/catalog/catalog_test.go`
* `tools/loadgen/internal/soak/catalog/content_test.go`
* `tools/loadgen/internal/soak/collector/collector.go`
* `tools/loadgen/internal/soak/collector/collector_test.go`
* `tools/loadgen/internal/soak/mutation/mutation.go`
* `tools/loadgen/internal/soak/mutation/mutation_test.go`
* `tools/loadgen/internal/soak/mutation/test_support_test.go`
* `tools/loadgen/internal/soak/presence/presence.go`
* `tools/loadgen/internal/soak/presence/presence_test.go`
* `tools/loadgen/internal/soak/presence/test_support_test.go`
* `tools/loadgen/internal/soak/read/read.go`
* `tools/loadgen/internal/soak/read/read_test.go`
* `tools/loadgen/internal/soak/read/test_support_test.go`
* `tools/loadgen/internal/soak/read/verify.go`
* `tools/loadgen/internal/soak/read/verify_test.go`
* `tools/loadgen/internal/soak/rpc/error.go`
* `tools/loadgen/internal/soak/rpc/error_test.go`
* `tools/loadgen/internal/soak/rpc/rpc_behavior_test.go`
* `tools/loadgen/internal/soak/run/integration_test.go`
* `tools/loadgen/internal/soak/run/lifecycle.go`
* `tools/loadgen/internal/soak/run/lifecycle_test.go`
* `tools/loadgen/internal/soak/run/run_test.go`
* `tools/loadgen/internal/soak/run/seed.go`
* `tools/loadgen/internal/soak/run/store.go`
* `tools/loadgen/internal/soak/run/teardown.go`
* `tools/loadgen/internal/soak/search/search.go`
* `tools/loadgen/internal/soak/search/search_test.go`
* `tools/loadgen/internal/soak/search/test_support_test.go`
* `tools/loadgen/internal/soak/send/mock_lifecycle_test.go`
* `tools/loadgen/internal/soak/send/send.go`
* `tools/loadgen/internal/soak/send/send_test.go`
* `tools/loadgen/internal/soak/userread/userread.go`
* `tools/loadgen/internal/soak/userread/userread_test.go`
* `tools/loadgen/internal/soak/wire/wire.go`
* `tools/loadgen/internal/soak/workload/workload.go`
* `tools/loadgen/internal/soak/workload/workload_test.go`
* `tools/loadgen/main.go`
* `tools/loadgen/mock_soak_store_test.go`
* `tools/loadgen/mongo_outage_recovery_integration_test.go`
* `tools/loadgen/soak_catalog_bridge.go`
* `tools/loadgen/soak_catalog_verifier_test.go`
* `tools/loadgen/soak_integration_test.go`
* `tools/loadgen/soak_manifest_contract_test.go`
* `tools/loadgen/soak_mutation_bridge.go`
* `tools/loadgen/soak_mutation_bridge_test.go`
* `tools/loadgen/soak_package_naming_contract_test.go`
* `tools/loadgen/soak_payload_metrics_test.go`
* `tools/loadgen/soak_presence_bridge.go`
* `tools/loadgen/soak_presence_integration_test.go`
* `tools/loadgen/soak_read_bridge.go`
* `tools/loadgen/soak_read_bridge_test.go`
* `tools/loadgen/soak_roommember_integration_test.go`
* `tools/loadgen/soak_rpc_bridge.go`
* `tools/loadgen/soak_run_bridge.go`
* `tools/loadgen/soak_run_integration_helpers_test.go`
* `tools/loadgen/soak_runtime_composition.go`
* `tools/loadgen/soak_runtime_composition_test.go`
* `tools/loadgen/soak_search_bridge.go`
* `tools/loadgen/soak_seed.go`
* `tools/loadgen/soak_seed_test.go`
* `tools/loadgen/soak_send_bridge.go`
* `tools/loadgen/soak_send_failure_integration_test.go`
* `tools/loadgen/soak_store_test.go`
* `tools/loadgen/soak_teardown.go`
* `tools/loadgen/soak_teardown_integration_test.go`
* `tools/loadgen/soak_teardown_test.go`
* `tools/loadgen/soak_test_clock_test.go`
* `tools/loadgen/soak_topology_bridge.go`
* `tools/loadgen/soak_topology_bridge_test.go`
* `tools/loadgen/soak_userread_adapter.go`
* `tools/loadgen/soak_wire_bridge.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

tools/loadgen/internal/soak/run/teardown.go
```diff
@@ -0,0 +1,163 @@
...
+		if cfg.ConfirmKeyspace == "" || cfg.ConfirmKeyspace != configuredKeyspace ||
+			cassandra == nil || cassandra.Keyspace() != configuredKeyspace {
+			return false, fmt.Errorf(
+				"refuse Cassandra truncate: SOAK_CONFIRM_KEYSPACE, CASSANDRA_KEYSPACE, and connected keyspace must match")
+		}
+	} else if cfg.CassandraCleanup != "none" {
+		return false, fmt.Errorf("refuse unknown Cassandra cleanup mode %q", cfg.CassandraCleanup)
+	}
+
+	manifest, err := store.FindManifest(ctx, cfg.RunID)
```
julianshen, line 88: _🎯 distributed-correctness_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`_

**Run lifecycle has no atomic ownership fence.**

After teardown reads an inactive manifest, `Prepare` can write `running` and begin dispatching before deletion starts. Teardown then removes that run's rooms and ownership, may truncate Cassandra, and finally overwrites the manifest as `cleaned`. `Prepare` also accepts an already-running manifest, allowing two workers whose terminal transitions can hide the other's activity; `Seed` has the same check-before-`ResetOwned` gap. Introduce an atomic owner/epoch claim shared by seed, prepare, and teardown; condition heartbeats and terminal transitions on that epoch, and require the teardown claim before destructive work.

<details>

```
In `tools/loadgen/internal/soak/run/teardown.go` around line 88: After teardown reads an inactive manifest, `Prepare` can write `running` and begin dispatching before deletion starts. Teardown then removes that run's rooms and ownership, may truncate Cassandra, and finally overwrites the manifest as `cleaned`. `Prepare` also accepts an already-running manifest, allowing two workers whose terminal transitions can hide the other's activity; `Seed` has the same check-before-`ResetOwned` gap. Introduce an atomic owner/epoch claim shared by seed, prepare, and teardown; condition heartbeats and terminal transitions on that epoch, and require the teardown claim before destructive work.

```

</details>

---

tools/loadgen/internal/soak/search/search.go
```diff
@@ -154,83 +156,68 @@ func (r *soakSearchReader) SearchRooms(ctx context.Context) error {
...
 		// longer than the deadline into a data-loss claim.
-		return soakSearchIndexUnknown, true, err
+		return IndexUnknown, true, err
 	}
 	for i := range response.Messages {
 		if response.Messages[i].MessageID == messageID {
-			return soakSearchIndexFound, true, nil
+			return IndexFound, true, nil
 		}
 	}
```
julianshen, line 198: _🏷️ observability_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`_

**A single capped search page can falsely report an indexed message missing.**

`IndexedAt` queries only `PageSize` hits and returns `IndexMissing` whenever the target is absent from that page. `catalog.SearchTerm` selects the first qualifying token, and the catalog documents that generated bodies are repeated-character filler, so many messages in one hot room can share a term. Once matches exceed `PageSize`, an older target can be fully indexed but omitted from the returned page, producing false loss evidence. Use a unique message-specific searchable token, or paginate until the target is found or results are exhausted.

<details>

```
In `tools/loadgen/internal/soak/search/search.go` around line 198: `IndexedAt` queries only `PageSize` hits and returns `IndexMissing` whenever the target is absent from that page. `catalog.SearchTerm` selects the first qualifying token, and the catalog documents that generated bodies are repeated-character filler, so many messages in one hot room can share a term. Once matches exceed `PageSize`, an older target can be fully indexed but omitted from the returned page, producing false loss evidence. Use a unique message-specific searchable token, or paginate until the target is found or results are exhausted.

```

</details>

---

tools/loadgen/internal/soak/catalog/catalog.go
```diff
@@ -65,35 +67,46 @@ type soakCatalogCandidate struct {
...
 	return hex.EncodeToString(digest[:])
 }
 
-type soakCatalogMessage struct {
-	soakCatalogCandidate
+// SearchTerm reduces a payload to a term the analyzer will match. The returned
+// term is cloned so retaining it does not retain the complete message body.
+func SearchTerm(content string) string {
+	for field := range strings.FieldsSeq(content) {
+		if len(field) >= 3 {
```
julianshen, line 101: _⚡ memory_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`oom`_

**Search-term retention keeps every full generated message body.**

When search observation enables RetainSearchTerms, generated bodies are one unbroken field, as the Catalog comment notes. SearchTerm therefore clones the entire payload, and each accepted catalog entry retains that clone up to RecentTotal. The memory test instead prefixes a short whitespace-delimited token, so it misses the production shape; its own comment documents 351 MB for a full default catalog. Larger payload or catalog settings can cause GC thrashing or container OOM. Generate a bounded searchable token in every payload and retain only that token, enforcing a maximum term length.

<details>

```
In `tools/loadgen/internal/soak/catalog/catalog.go` around line 101: When search observation enables RetainSearchTerms, generated bodies are one unbroken field, as the Catalog comment notes. SearchTerm therefore clones the entire payload, and each accepted catalog entry retains that clone up to RecentTotal. The memory test instead prefixes a short whitespace-delimited token, so it misses the production shape; its own comment documents 351 MB for a full default catalog. Larger payload or catalog settings can cause GC thrashing or container OOM. Generate a bounded searchable token in every payload and retain only that token, enforcing a maximum term length.

```

</details>

---

tools/loadgen/internal/soak/mutation/mutation.go
```diff
@@ -493,17 +497,17 @@ func (m *soakMutator) React(
 	m.catalog.SetReaction(
```
julianshen, line 497: _🎯 concurrency_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`performance-and-concurrency`_

**Concurrent reaction toggles can overwrite newer catalog state**

Reaction slots can concurrently select the same hot message and actor from identical catalog snapshots. If the service processes one request as add and the next as remove, but the goroutines apply their responses in reverse order here, the catalog ends with the reaction present while the server has removed it. Hot-only scope makes this likely whenever reaction latency exceeds the dispatch interval, skewing later actor selection and reconciliation evidence. Serialize selection through catalog update per target, or reserve/version the target and reject stale updates.

<details>

```
In `tools/loadgen/internal/soak/mutation/mutation.go` around line 497: Reaction slots can concurrently select the same hot message and actor from identical catalog snapshots. If the service processes one request as add and the next as remove, but the goroutines apply their responses in reverse order here, the catalog ends with the reaction present while the server has removed it. Hot-only scope makes this likely whenever reaction latency exceeds the dispatch interval, skewing later actor selection and reconciliation evidence. Serialize selection through catalog update per target, or reserve/version the target and reject stale updates.

```

</details>

---

julianshen (commented):
### tGDBot review of `82323cf` — 6 findings in 86 files

2 🔴 blocking · 2 🟠 warning · 2 🔵 suggestion
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/474#issuecomment-5561036917), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

tools/loadgen/internal/soak/presence/test_support_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+type testSleeper struct{}
+
+func (*testSleeper) Sleep(ctx context.Context, _ time.Duration) error {
+	return ctx.Err()
+}
+
+type testReadRecorder struct {
+	mu      sync.Mutex
+	samples []read.Sample
```
julianshen: _🏷️ dead-code_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`dead-code`_

**The presence test recorder stores samples that are never read.**

`newPresenceFixture` passes `testReadRecorder` to the lane, and `Record` appends every sample here, but the added presence tests never inspect this slice and the type provides no accessor. The accumulated samples therefore have no observable effect. Pass a nil recorder to `New` and remove `testReadRecorder`.

<details>

```
In `tools/loadgen/internal/soak/presence/test_support_test.go` around line 40: `newPresenceFixture` passes `testReadRecorder` to the lane, and `Record` appends every sample here, but the added presence tests never inspect this slice and the type provides no accessor. The accumulated samples therefore have no observable effect. Pass a nil recorder to `New` and remove `testReadRecorder`.

```

</details>

---

tools/loadgen/internal/soak/presence/test_support_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+func testTopology(candidates int) *topology.Topology {
+	users := make([]model.User, 0, candidates+2)
+	for i := range candidates + 2 {
+		users = append(users, model.User{
+			ID:      "u" + string(rune('a'+i%26)) + string(rune('0'+i/26)),
+			Account: "user-" + string(rune('a'+i%26)) + string(rune('0'+i/26)),
+		})
+	}
+	return &topology.Topology{ActiveUsers: users[:2]}
```
julianshen: _🏷️ dead-code_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`dead-code`_

**testTopology constructs users that it always discards.**

The helper builds `candidates+2` users but returns only `users[:2]`. For the `testTopology(3)` fixture, three generated `model.User` values are never consumed. Return all requested candidates; `Lane.New` still applies the configured two-connection bound.

<details>

> ‼️ **IMPORTANT**

```suggestion
	users := make([]model.User, 0, candidates)
	for i := range candidates {
		users = append(users, model.User{
			ID:      "u" + string(rune('a'+i%26)) + string(rune('0'+i/26)),
			Account: "user-" + string(rune('a'+i%26)) + string(rune('0'+i/26)),
		})
	}
	return &topology.Topology{ActiveUsers: users}
```

</details>

<details>

```
In `tools/loadgen/internal/soak/presence/test_support_test.go` around line 101: The helper builds `candidates+2` users but returns only `users[:2]`. For the `testTopology(3)` fixture, three generated `model.User` values are never consumed. Return all requested candidates; `Lane.New` still applies the configured two-connection bound.

```

</details>

---

tools/loadgen/internal/soak/presence/test_support_test.go
flywindy: Confirmed and fixed in 6f95b77.

`testReadRecorder` appended every sample under a mutex, `presence_test.go:76` was its only use, and nothing read `.samples` — no accessor existed. `Lane` already guards `l.recorder != nil` before recording (`presence.go:382`), so the fixture now passes nil and the type is gone.

---

tools/loadgen/internal/soak/presence/test_support_test.go
flywindy: Confirmed and fixed in 6f95b77, but not with the suggested edit — that one would change the fixture.

The finding is right: both call sites, `testTopology(3)` and `testTopology(1)`, got the same two users whatever they asked for, so `candidates+2` built three and one user respectively that were then sliced away.

Returning all requested candidates widens the active population, and the presence tests assert against exactly that pair — `user-a0` and `user-b0` appear in the expected payloads at `presence_test.go:164` and `:179` and in the subject assertions at `:319`/`:323`. `testTopology(1)` would drop to a single user, which is not a pair at all.

So the parameter is removed rather than made real, and the ID formula is kept verbatim for the same reason. Same two users, no discarded work, no inert knob.

---

tools/loadgen/internal/soak/run/teardown.go
flywindy: Verified and agreed on the mechanism — deferring the fix, not dismissing it. Tracked as **F11** in `docs/load-testing/loadgen-findings.md`.

What I confirmed against the code: `Teardown` reads the manifest at `teardown.go:88`, refuses only a `StateRunning` manifest with a fresh heartbeat, then falls straight into the deletion loop with nothing claimed. `Lifecycle.Prepare` writes `manifest.State = StateRunning` through an unconditional `PutManifest` (`lifecycle.go:86-89`), and teardown finishes with an equally unconditional `MarkCleaned` (`teardown.go:149`). So a run that starts inside that window dispatches against rooms teardown is deleting, and its state is then overwritten. `Prepare` accepting an already-`StateRunning` manifest is the second half you name.

Why not here: this PR is a behaviour-preserving extraction, and all of this predates it — `soak_teardown.go:154` on `main` calls `MarkCleaned` with no claim either. The extraction moved the code without changing it. Mixing a fix in would cost the only signal an 8,700-line move has: that a red test means the move broke something rather than that the fix changed something.

Exposure today is low — the soak has a single operator — but the failure is silent and destructive, so it is queued ahead of the lighter items. It needs the owner/epoch claim you describe, shared by seed, prepare and teardown, which is a design change rather than a patch.

CodeRabbit raised the same thing at [#discussion_r3944635244](https://github.com/hmchangw/newchat/pull/474#discussion_r3944635244); they are one finding.

---

tools/loadgen/internal/soak/run/teardown.go
```diff
@@ -0,0 +1,163 @@
...
+	if manifest == nil {
+		return false, nil
+	}
+	if manifest.State == StateCleaned {
+		return true, nil
+	}
+	if manifest.State == StateRunning {
+		if manifest.LastHeartbeatAt == nil ||
+			time.Since(*manifest.LastHeartbeatAt) <= cfg.HeartbeatStaleAfter {
+			return true, fmt.Errorf("refuse teardown for active soak run %q", cfg.RunID)
```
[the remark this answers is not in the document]
flywindy: Verified and agreed — deferring, not dismissing. Tracked as **F11** in `docs/load-testing/loadgen-findings.md`.

All three sites check out: the unclaimed read at `teardown.go:88`, `Prepare`'s unconditional `PutManifest` at `lifecycle.go:86-89`, and the unconditional `MarkCleaned` that closes teardown. A run starting between teardown's read and its deletes runs against data teardown is removing, and either writer can overwrite the other's final state.

Not fixed in this PR because it is a behaviour-preserving extraction and this predates it — `main`'s `soak_teardown.go:154` calls `MarkCleaned` with no claim too. Fixing it here would mean a red test could no longer be read as "the move broke something".

tGDBot raised the same finding at [#discussion_r3944767175](https://github.com/hmchangw/newchat/pull/474#discussion_r3944767175) with the fuller `Seed`/`Prepare` framing; they are one item in the ledger.

---

tools/loadgen/internal/soak/run/teardown.go
```diff
@@ -0,0 +1,163 @@
...
+		if page == nil {
+			break
+		}
+		if page.Cursor == "" || len(page.RoomIDs) == 0 {
+			return true, fmt.Errorf("invalid empty ownership page for run %q", cfg.RunID)
+		}
+		if page.Cursor <= after {
+			return true, fmt.Errorf("ownership cursor did not advance for run %q", cfg.RunID)
+		}
+		for _, batch := range ChunkRoomIDs(page.RoomIDs, cfg.BatchRooms) {
```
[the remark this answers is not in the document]
flywindy: The mechanism is real but not currently reachable, so this is tracked as **F12** at Low rather than Major.

Confirmed: `ChunkRoomIDs(page.RoomIDs, cfg.BatchRooms)` at `teardown.go:121` returns no batches for a non-positive size (`run_test.go` pins `ChunkRoomIDs([]string{"room"}, 0)` as nil), and the flow then continues to `DeleteOwnership` and `MarkCleaned` at `teardown.go:137-149` — so the rooms survive and the record that could find them again does not. That part of the finding is exactly right.

What keeps it off the critical list: `validateSoakConfig` already rejects a non-positive `TEARDOWN_BATCH_ROOMS` (`soak_config.go:524-525`), so no CLI invocation can reach it.

It is still worth the guard, for the reason this PR creates: `run.Teardown` is now a public package API with its own `Config`, and it trusts a validation that lives in a different package. That was a local invariant before the extraction and is an unguarded precondition now. Three lines, and it closes before PR 3 or 4 adds a second caller — but it is a behaviour change, so it goes in its own PR rather than into an extraction whose whole claim is that it changed nothing.

---

tools/loadgen/internal/soak/search/search.go
flywindy: Verified, and the collision argument is stronger than stated. Tracked as **F13** in `docs/load-testing/loadgen-findings.md`.

`IndexedAt` scans one `PageSize` response and returns `IndexMissing` when the target is absent (`search.go:194-199`), never paging further — confirmed. On the term: generated bodies are `strings.Repeat("x", size)` (`distribution.go:137`), with no whitespace at all, so `catalog.SearchTerm` returns the whole body as the single field. Every message of the *same generated size* therefore carries an identical term, and sizes come from a bounded log-normal, so exact-size collisions are the norm in a hot room rather than an edge case.

This is the finding I would put ahead of the other deferred ones despite its Blocking-vs-Warning ranking against F11, because of what it costs: the soak's output is evidence, and a false `IndexMissing` sends someone hunting a data-loss bug that never happened. Same class as an existing ledger entry (F1) where the tool reports something other than what occurred.

Not fixed here — `main`'s `soak_search.go:79-80,128` has the same single-page read, so this predates the extraction and this PR is behaviour-preserving.

Worth noting it shares a root cause with your catalog finding at [#discussion_r3944767179](https://github.com/hmchangw/newchat/pull/474#discussion_r3944767179): the body being one unbroken token is what makes the term non-unique *and* what makes retaining it expensive. A bounded, message-unique searchable token in every payload closes both, which is how they are queued.

---

tools/loadgen/internal/soak/catalog/catalog.go
flywindy: Verified, including the part about the test. Tracked as **F14** in `docs/load-testing/loadgen-findings.md`.

`SearchTerm` walks `strings.FieldsSeq(content)` and clones the first field of three or more characters (`catalog.go:99-106`), and its doc comment says the clone means "retaining it does not retain the complete message body". For production payloads that is inverted: `ContentOfSize` generates `strings.Repeat("x", size)` (`distribution.go:137`) with no whitespace, so `FieldsSeq` yields exactly one field — the whole body — and the clone copies all of it. The comment is not just optimistic, it describes the opposite of what happens.

Your point about the test is the sharper half and I confirmed it: `content_test.go:158` builds `CatalogProbeID(i) + " " + strings.Repeat("ab ", …)`, a whitespace-delimited body whose first qualifying field is the short probe ID. It measures a shape production never generates — and still records 351 MB. That is a test certifying a bound it never exercised, which is a pattern the ledger now tracks in its own right alongside F8 and F4.

Not fixed here: `main`'s `soak_catalog.go:74,83-84` has the same `searchProbeTerm` and `retainSearchTerms` behaviour, so it predates the extraction.

Queued together with [#discussion_r3944767176](https://github.com/hmchangw/newchat/pull/474#discussion_r3944767176) — one bounded message-unique token fixes the retention and the false-missing report at once — and the fix should bring `content_test.go` onto the production payload shape, or the new bound goes unmeasured the same way.

---

tools/loadgen/internal/soak/mutation/mutation.go
flywindy: Verified — deferring. Tracked as **F15** in `docs/load-testing/loadgen-findings.md`.

`catalog.SetReaction` is applied unconditionally after the response (`mutation.go:497-504`), with no version or reservation on the target, so two slots that selected the same hot message and actor from identical snapshots can land add-then-remove in either order. The catalog then holds a reaction the server does not, and later actor selection and the reconciliation evidence both drift from it.

Rated Low-Med in the ledger rather than Warning-with-heavy-lift: the window needs reaction latency to exceed the dispatch interval, and the damage is skewed evidence rather than lost or corrupted data. It is queued behind F13/F14, which make the run report things that did not happen.

Not fixed here — the same unconditional apply is on `main` in `soak_mutation.go`, so it predates the extraction and this PR is behaviour-preserving.

---

vjauhari-work (approved):
LGTM!

Issue #473, opened 2026-09-06
o11y: add handler telemetry for RegisterVoid presence routes


flywindy:
## Context

Branch `claude/nats-rpc-operation-vocab-slbows` moves `rpc_method` from subject parsing to route registration. As part of that, `natsrouter.RegisterVoid` takes no method and records **no `rpc.server.call.duration` sample at all**.

Removing them from the RPC family was correct: these routes have no reply subject, so there is no round trip to time, and recording local handler cost as an RPC would misreport what the histogram measures. The repo's own subject taxonomy agrees — these are `.event.` subjects, not `.request.` ones.

But the replacement instrument was never built, and the four affected routes are the fleet's **highest-volume traffic**:

| Handler | Subject (`pkg/subject/subject.go:765-782`) |
| --- | --- |
| `Hello` | `chat.user.{account}.event.presence.{siteID}.hello` |
| `Ping` | `chat.user.{account}.event.presence.{siteID}.ping` |
| `Activity` | `chat.user.{account}.event.presence.{siteID}.activity` |
| `Bye` | `chat.user.{account}.event.presence.{siteID}.bye` |

## What was lost is more than latency

`pkg/natsrouter/register.go:116,122` still compute `c.requestResult`, but the only reader — `recordHandled` — returns early when `!rt.recordRPC`. For these four routes those assignments are now inert.

Before the branch they recorded count **and** `error_type` under `rpc_method="unknown"`. So the regression is a lost **result classification**, not only a lost duration. Concretely: if Valkey rejects the `pingScript` EVALSHA (`presencestore/store.go:242`), every heartbeat fails, presence goes stale fleet-wide, and the only signal is one ERROR log line per heartbeat — no series to alert on, while `rpc.server.call.duration` stays green on the three reply-bearing routes.

A replacement that records only duration would not close this half.

### The duration instrument

- [ ] Add `chat.nats.handler.duration`
- [ ] Attributes: `service_name`, `site`, `operation`, `result`
- [ ] **No `rpc.method`** — these are not RPCs and must not be labelled as though they were
- [ ] `operation` is bounded, exactly: `presence_hello`, `presence_ping`, `presence_activity`, `presence_bye`
- [ ] `result` reuses the existing bounded `RequestResult` vocabulary:
  - successful handler → `success`
  - invalid JSON → `bad_request`
  - typed handler error → its matching classification
  - panic or unknown error → `internal`

### The drop counter

[...]

PR #475, opened 2026-09-07
feat(o11y): declare rpc.method at route registration instead of parsing subjects

flywindy:
> [!IMPORTANT]
> **Merge [#477](https://github.com/hmchangw/newchat/pull/477) first.** Its first commit (`9310b6b`) is also the first commit of this branch, and the version here is **superseded** — two rounds of review found fail-open bugs in it that are fixed in #477, not here. See [Two things in this PR that aren't about labels](#two-things-in-this-pr-that-arent-about-labels). Once #477 lands, this branch rebases on top and drops from 106 files to 92.

## The problem, in one screenshot's worth

We had a latency alarm on `user-service` and could not tell which API was slow.

**Before** — group `rpc_server_call_duration_seconds` by `rpc_method` and you get this:

```
rpc_method="unknown"          ██████████████████████  51 routes collapsed into one bar
rpc_method="room_read"        ████
rpc_method="history_read"     ███
rpc_method="member_read"      ██
...
```

**After** — every route is its own series:

```
rpc_method="list_channel_messages"    ████
rpc_method="get_chatlist"             ███
rpc_method="batch_get_rooms_info"     ██
rpc_method="search_messages"          █
... 92 distinct methods
```

The `unknown` bar was 51 of user-service's 96 routes. The slow one was in there, and nothing in the metric could tell you which.

**Why it happened:** the label was *guessed* at request time by pattern-matching the NATS subject string. The mapper knew 22 operation names, and those only covered room-service and history-service. Any route whose subject shape it didn't recognise fell through to `unknown`.

This PR stops guessing. **The route declares its own method name when it registers.**

---

### 1. How a route gets its label

**Before** — the label is derived from the subject, at runtime, by a mapper that lives somewhere else:

```go
natsrouter.Register(r, subject.SearchMessagesPattern(siteID), h.searchMessages)
//                     ↑ label inferred from this string, at request time
```

**After** — the label is an argument, checked by the compiler:

```go
natsrouter.Register(r, subject.SearchMessagesPattern(siteID),
    natsmetrics.MethodSearchMessages, h.searchMessages)
//  ↑ a constant from a closed vocabulary in pkg/natsmetrics
```

| | Before | After |
|---|---|---|
| Where the name comes from | inferred from the subject string | declared at the call site |
| Forget to handle a route | silently becomes `unknown` | **does not compile** |
| Typo in the name | impossible (nothing to type) — but the whole route silently mislabels | rejected by semgrep |
| Wrong-but-valid name | invisible | **a one-line diff in a golden file** |
| Where you look it up | read the mapper's regexes | read `testdata/routes.golden` |

### 2. What the label values look like

**Before** — 22 coarse buckets, most routes in none of them:

```
room_mutation      room_read       member_read      member_mutation
history_read       history_mutation                 channel_history
thread_open        ...             unknown   ← 51 of 96 user-service routes
```

**After** — one name per route, derived from the *handler*, not the subject. Actual excerpt from `user-service/service/testdata/routes.golden`:

```
add_priority_contact       chat.user.{account}.request.user.site-a.settings.priorityContacts.add
batch_get_badge_counts     chat.server.request.user.site-a.badge.count.batch
count_subscriptions        chat.user.{account}.request.user.site-a.subscription.count
create_chatlist_section    chat.user.{account}.request.user.site-a.chatlist.section.create
get_chatlist               chat.user.{account}.request.user.site-a.chatlist.get
get_current_user           chat.user.{account}.request.user.site-a.me
...
```

Names follow `<verb>_<object>`, and the four read verbs have fixed meanings (Google AIP-131/132/190):

| Prefix | Means |
|---|---|
| `get_` | one resource, or one view of one |
| `list_` | a collection — *whether the response carries a total is irrelevant* |
| `batch_get_` | several specific resources, by keys the caller supplied |
| `search_` | a query: matching, ranking, or scoring |

That last rule is here because three history-service methods returning `[]Message` with limit/cursor had been named `get_` purely because they lacked a total field, which left one golden file carrying two contradictory precedents for the same shape.

---

## ⚠️ What breaks

**Every `rpc_method` value changes.** This is not additive — old values stop existing.

| Old value | Becomes |
|---|---|
| `unknown` | **gone entirely** — split into ~51 named methods, *and* presence heartbeats leave this metric completely |
| `channel_history` | `list_channel_messages` |
| `thread_open` | `list_thread_messages` |
| `room_mutation` | `rename_room`, `create_room`, `move_chat`, … |
| `history_read` | `get_message`, `batch_get_messages`, … |
| `room_read`, `member_read`, … | individual method names |

**What you need to do:** any Grafana panel, alert, or recording rule that filters on `rpc_method` needs a pass. Two traps specifically:

- A panel that *excludes* `rpc_method="unknown"` should have that filter **deleted**, not left dangling — the series is gone, not smaller.
- The presence `hello`/`ping`/`activity`/`bye` handlers now emit **no RPC sample at all**. They were never round-trips (no reply subject), so counting them as RPC latency was wrong — but until #473 lands they have *no* latency metric. That is a deliberate, documented gap, not an oversight.

There is no compatibility shim. A shim would preserve exactly the ambiguity this PR exists to remove.

**Not a breaking change, but worth knowing:** uniqueness is enforced *per router*, not globally, because the identity key is `service_name + rpc_method`. So `mark_all_threads_read` is legitimately registered by both room-service and user-service — those are the two halves of one operation across a hop.

---

## What stops this from rotting

Three gates, cheapest first. Each was verified by deliberately injecting the defect it's meant to catch and confirming it fails.

| Gate | Catches | Misses |
|---|---|---|
| **Compiler** (`type RPCMethod string`) | missing argument, swapped argument order | string literals, casts, variables |
| **Golden files** (`testdata/routes.golden`, one per service) | a valid-but-wrong constant, an unreviewed route change, a new service with no golden, duplicate subjects or methods | a `t.Skip` inside the test body |
| **semgrep** (`.semgrep/rpcmethod.yml`) | string literals, `RPCMethod("typo")` casts, variables — *allowlist*, so unanticipated shapes are rejected by default | — |

To update a golden file, **delete and regenerate — never hand-edit**:

```
rm <package>/testdata/routes.golden && make test SERVICE=<service>
```

It regenerates from the live route table and **fails that run on purpose**, so the new table lands in a diff someone reads rather than being quietly absorbed.

---

## Two things in this PR that aren't about labels

**1. `errcode.FromReply` — review this in [#477](https://github.com/hmchangw/newchat/pull/477), not here.** Instrumenting the client lane meant reading remote error envelopes at 21 call sites. All 21 were hand-rolled, and 3 were wrong the same way: `errcode.Parse` returns `ok` for a code this build doesn't recognise, and the caller treated that as success — the reply fell through to a decoder, JSON ignored the unknown fields, and a remote failure became an empty result recorded as a *successful* call. One site fed the unknown code into `errcode.New`, **which panics**.

```go
// Before — three different hand-rolled shapes, two of them wrong
if ee, ok := errcode.Parse(data); ok && ee.Code.Valid() { ... }   // unknown code = success
if ee, ok := errcode.Parse(data); ok { return ee }                // relays unknown code typed

// After — one helper, semgrep-enforced
if err := errcode.FromReply(data); err != nil { return err }
```

The call-site shape above is final. **The helper's own body on this branch is not** — commit `9310b6b` here is the version before two rounds of review, and both rounds found real bugs in it:

- It asked `Parse` whether a payload was an envelope, but `Parse` answers whether the payload fits *this build's* `Error` struct. `Metadata` is `map[string]string`, so `{"code":"unavailable","error":"upstream down","metadata":{"retryAfter":5}}` failed to decode, `Parse` said "not an envelope", and the reply fell through to the success decoder — the same fail-open the unknown-code branch exists to close.
- The same held for an `"error"` field that isn't a string. Verified end to end: `FromReply` returned nil, the success decoder produced `rooms=[] len 0`, and `GetRoomsInfo` returned `(nil, nil)`.
- The semgrep rule here matches only the two assignment shapes, so `return errcode.Parse(x)`, `f(errcode.Parse(x))` and `parseReply := errcode.Parse` all walk straight through it.

All three are fixed in #477 and none of the fixes are on this branch. **Merge #477 first**; this branch then rebases onto it and these files leave the diff entirely.

**2. Client-side RPC metrics** — 12 outbound call sites now record `rpc.client` duration and outcome. The outcome is captured with a named return + `defer`, so a failure *after* the NATS round-trip (a bad envelope, a decode error) is recorded as a failure rather than a success. `search-service/room_client.go` is the one site still uninstrumented; it's documented rather than silently skipped.

---

## Also not fixed here

`le=10` is still the top histogram bucket, so anything ≥10s reads as exactly 10s. That's why the p99 looked pinned. Whether to extend the bucket range is a separate call with its own cardinality cost — and it's much easier to make once the slow route has a name.

## Follow-up required before merge

hmchangw/newchat#473 — the handler-side counterpart (`chat.nats.handler.duration` and `chat.nats.handler.dropped` on the `.event.` subjects), which is what restores observability to the presence heartbeats. **It has no owner, milestone, or labels yet.**

## Verification

Rebased onto `28b97da`. CI is **21/21 green** on `e47704d`.

Locally on the current head: `make test`, `make lint` (0 issues), `make sast-gosec`, the semgrep rule-test suites, a full-tree `semgrep --config=.semgrep/` sweep (0 findings), `go vet -tags integration ./...`, `git diff --check`, and each of the six commits building and vetting standalone in a throwaway worktree.

**Interaction with #476.** That PR added `hardcoded-credential-literal`, and the new rule fires on this branch's `MethodSetSSOToken` / `MethodRefreshSSOToken` constants — metric label names that happen to end in `Token` and be assigned string literals. They already carried `// #nosec G101`, which is exactly the failure mode #476's own rule message describes: each scanner reads its own directive, so suppressing one leaves the others reporting. Both sites now carry `// nosemgrep: hardcoded-credential-literal` alongside, matching the convention #476 established. Caught by running the sweep locally after rebasing, not by CI.

One pre-existing test flake in `pkg/preview` (a package this PR doesn't touch) surfaced during an earlier CI run; it reproduces on a clean `main`, and a diagnosis plus measured patch are in [this comment](https://github.com/hmchangw/newchat/pull/475#issuecomment-5564465933). It is not this PR's to fix.

## Reading order

| Commit | |
|---|---|
| `9310b6b` | `refactor(errcode)`: one tested helper for reading a remote reply — **superseded by #477, will vanish on rebase** |
| `171a0ee` | `feat(o11y)`: declare `rpc.method` at route registration — **the behaviour change is all here** |
| `872b4bf` | `test(o11y)`: pin every route to its method, fail-closed |
| `80353c0` | `build(sast)`: require a vocabulary constant in every method argument |
| `40a133d` | `feat(o11y)`: instrument the outbound RPC client lane |
| `e47704d` | `docs(o11y)`: the metric contract, the cutover, and what is still open |

106 files, +6225/−697.

---


flywindy:
## CI on `3a1b1ba`: three failures, two mine (fixed), one pre-existing

Pushed `557b91a` with fixes for the two that are this PR's. The third is not, and I'm standing down on it with a proposed patch rather than widening this PR.

### `sast` — mine, fixed

`Findings: 5 (5 blocking)`, all from the repo-owned rules:

- `remote-envelope-must-use-fromreply` at `tools/clientsim/subwalk.go:137` — a site still reading the envelope with `errcode.Parse`, which is exactly what this PR's own new rule forbids. Now uses `errcode.FromReply`.
- `gosec.G304-1` ×4 in `pkg/testutil/routesgolden{,_test}.go` — false positives on a package-level `const`; suppressed with `#nosec` + `nosemgrep` per the repo's existing convention.

Also anchored this rule's exclude pattern to `/pkg/errcode/`, which clears the Semgrepignore-v2 deprecation warning CI emitted for it. (The same warning on `jsretry-marshal-failure-must-be-permanent` is pre-existing and left alone.)

For the record: `gosec=PASS govulncheck=PASS rule-tests=PASS` in that run — only `semgrep` was red, so `govulncheck` is now confirmed green, which I could not verify locally.

### `test-integration (user-service)` — mine, fixed

Three subtests named `unknown-code error envelope — relayed, not masked` asserted:

```go
require.True(t, errors.As(err, &e))
assert.Equal(t, "upstream boom", e.Message)
```

That is precisely the `ok`-alone shape `FromReply` was extracted to eliminate. They were written before the extraction and never updated. `pkg/errcode/remote_test.go` asserts `wantTyped: false` on the *identical* payload — the two suites contradicted each other, and only the unit one runs under `make test`. Now aligned: the error is non-nil, is **not** an `*errcode.Error`, and carries both the unknown code and the remote message as text.

### `test` — not this PR's

`TestAppNameCache_NamesCollapsesRegardlessOfAccountOrder` in `pkg/preview`, a package this PR does not touch (`git diff --name-only origin/main..HEAD -- pkg/preview/` is empty).

Verified pre-existing rather than assumed: checked out `origin/main` clean and ran it there — fails within the first of 50 runs. It is a genuine race in the test, not a real defect in `AppNameCache`, and no fix for it exists in any open PR.

**Why it races.** `close(release)` fires immediately after the spawn loop, so the loader can return before all 8 goroutines reach `sfMany.Do`. `singleflight` deletes the key once the leader's function returns, so a late arrival becomes a *new* leader. The cache is `nil` in this test, so `loadNames`' cache-recheck cannot dedupe it either — second leader, second read, `calls == 2`.

**Proposed patch** — a barrier so the loader cannot return before every goroutine is in flight:

```diff
 	orders := [][]string{{"a.bot", "b.bot"}, {"b.bot", "a.bot"}}
-	var wg sync.WaitGroup
+	var wg, arrived sync.WaitGroup
+	arrived.Add(8)
 	for i := range 8 {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
+			arrived.Done()
 			got, err := c.Names(context.Background(), orders[i%2])
 			assert.NoError(t, err)
 			assert.Equal(t, map[string]string{"a.bot": "Alpha", "b.bot": "Beta"}, got)
 		}()
 	}
+	arrived.Wait()
 	close(release)
 	wg.Wait()
```

Measured on `origin/main`: unpatched fails within 50 runs, patched passes 50/50.

One caveat I'd rather state than paper over: this shrinks the window between "goroutine is running" and "goroutine is inside `sfMany.Do`" but does not close it. A fully deterministic version needs a hook letting the test observe that N callers are parked on the key. `TestAppNameCache_NamesCollapsesConcurrentColdBatches` (line 308) has the same structure and the same latent race; it just hasn't surfaced yet.

Happy to open this as its own PR against `main` if you'd like — it doesn't belong in this one.

---

julianshen:
**5 findings · 5 inline comments posted** — 🔴 1 blocking · 🟠 1 warning · 🔵 3 suggestion

### Related work

- [Issue #473](https://github.com/hmchangw/newchat/issues/473) — o11y: add handler telemetry for RegisterVoid presence routes (open)
- [PR #476](https://github.com/hmchangw/newchat/pull/476) — feat\(sast\): add the hardcoded-credential rule the local gate was missing (merged)

<details>
<summary>📒 Files reviewed (106)</summary>

* `.semgrep/remoteenvelope.go`
* `.semgrep/remoteenvelope.yml`
* `.semgrep/rpcmethod.go`
* `.semgrep/rpcmethod.yml`
* `admin-service/client_update.go`
* `admin-service/room_onduty.go`
* `bot-message-handler/handler.go`
* `bot-message-handler/routes_test.go`
* `bot-message-handler/testdata/routes.golden`
* `bot-room-service/routes_test.go`
* `bot-room-service/testdata/routes.golden`
* `botplatform-service/bot_forwarder.go`
* `botplatform-service/dm_ensurer.go`
* `broadcast-worker/parent_fetcher_test.go`
* `data-migration/oplog-transformer/historyclient.go`
* `docs/load-testing/common/sli-slo.md`
* `docs/superpowers/plans/2026-09-03-rpc-method-registration-vocabulary.md`
* `docs/superpowers/plans/2026-09-04-rpc-method-hardening.md`
* `history-service/internal/publisher/mock_publisher_test.go`
* `history-service/internal/service/routes_test.go`
* `history-service/internal/service/testdata/routes.golden`
* `media-service/emoji_nats.go`
* `media-service/routes_test.go`
* `media-service/testdata/routes.golden`
* `message-gatekeeper/fetcher_history_test.go`
* `notification-worker/badge_client_test.go`
* `notification-worker/parent_fetcher_test.go`
* `notification-worker/presence.go`
* `pkg/errcode/remote.go`
* `pkg/errcode/remote_test.go`
* `pkg/natsmetrics/enums_test.go`
* `pkg/natsmetrics/metrics_test.go`
* `pkg/natsmetrics/opttable.go`
* `pkg/natsmetrics/opttable_test.go`
* `pkg/natsmetrics/prometheus_export_test.go`
* `pkg/natsmetrics/rpcmethod.go`
* `pkg/natsmetrics/rpcmethod_test.go`
* `pkg/natsmetrics/rpcsemconv.go`
* `pkg/natsmetrics/rpcsemconv_test.go`
* `pkg/natsmetrics/toggle_test.go`
* `pkg/natsrouter/README.md`
* `pkg/natsrouter/example_test.go`
* `pkg/natsrouter/integration_test.go`
* `pkg/natsrouter/oversize_integration_test.go`
* `pkg/natsrouter/params.go`
* `pkg/natsrouter/router.go`
* `pkg/natsrouter/routes_golden_coverage_test.go`
* `pkg/natsrouter/shutdown_test.go`
* `pkg/testutil/embeddednats.go`
* `pkg/testutil/routesgolden.go`
* `pkg/testutil/routesgolden_test.go`
* `room-service/handler.go`
* `room-service/routes_test.go`
* `room-service/testdata/routes.golden`
* `room-worker/main.go`
* `room-worker/routes_test.go`
* `room-worker/testdata/routes.golden`
* `search-service/handler.go`
* `search-service/routes_test.go`
* `search-service/testdata/routes.golden`
* `tools/clientsim/subwalk.go`
* `tools/loadgen/maxrps_search.go`
* `translation-service/integration_test.go`
* `translation-service/routes_test.go`
* `translation-service/testdata/routes.golden`
* `user-presence-service/peer_client.go`
* `user-presence-service/routes_test.go`
* `user-presence-service/testdata/routes.golden`
* `user-service/historyclient/client_test.go`
* `user-service/main.go`
* `user-service/presenceclient/client_integration_test.go`
* `user-service/presenceclient/client_test.go`
* `user-service/roomclient/client_test.go`
* `user-service/service/routes_test.go`
* `user-service/service/testdata/routes.golden`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

pkg/natsrouter/register.go
```diff
@@ -14,6 +14,7 @@ import (
 func Register[Req, Resp any](
 	r *Router,
 	pattern string,
+	method natsmetrics.RPCMethod,
```
julianshen, line 17: _🏷️ build_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`tgd-review`_

**The new registrar signature leaves tagged room-service callers broken**

Adding the required method argument breaks the seven existing `natsrouter.Register*` calls in `room-service/integration_test.go`, which this diff does not migrate. Those calls are hidden behind the integration build tag, so ordinary unit tests and lint may remain green, but `make test-integration SERVICE=room-service` will fail to compile with too few arguments. Add the appropriate `natsmetrics.Method*` constant to each integration registration, as was done for the pkg/natsrouter and translation-service integration tests.

<details>

```
In `pkg/natsrouter/register.go` around line 17: Adding the required method argument breaks the seven existing `natsrouter.Register*` calls in `room-service/integration_test.go`, which this diff does not migrate. Those calls are hidden behind the integration build tag, so ordinary unit tests and lint may remain green, but `make test-integration SERVICE=room-service` will fail to compile with too few arguments. Add the appropriate `natsmetrics.Method*` constant to each integration registration, as was done for the pkg/natsrouter and translation-service integration tests.

```

</details>

---

pkg/natsrouter/router.go
```diff
@@ -199,8 +219,82 @@ func (r *Router) Use(mw ...HandlerFunc) {
...
+func (r *Router) addVoidRoute(pattern string, handlers []HandlerFunc) {
+	r.addRoute(pattern, "", false, handlers)
+}
+
+// recordHandled reports one handler result, or nothing at all for a
+// fire-and-forget route. Routed through one helper so the three record sites
+// (stopping gate, admission rejection, handler completion) cannot disagree
+// about whether a void route is recorded.
+func (r *Router) recordHandled(ctx context.Context, rt route, started time.Time, result natsmetrics.RequestResult) {
+	if !rt.recordRPC {
```
julianshen, line 288: _🏷️ observability_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`_

**Void-route failures and drops become invisible**

On a `RegisterVoid` route, stopping/admission rejection and handler completion call `recordHandled`, but this guard discards every result. If a presence ping arrives while `WithMaxConcurrency` is saturated, admission fails, `replyBusy` has no reply subject, the event is dropped, and no metric distinguishes it from successful processing. Handler failures, such as a presence-store error, are likewise invisible. Before this change they produced an unavailable/error sample. Add the planned bounded handler-duration/result and dropped-reason instruments in this cutover, or retain an explicit failure counter until they exist.

<details>

```
In `pkg/natsrouter/router.go` around line 288: On a `RegisterVoid` route, stopping/admission rejection and handler completion call `recordHandled`, but this guard discards every result. If a presence ping arrives while `WithMaxConcurrency` is saturated, admission fails, `replyBusy` has no reply subject, the event is dropped, and no metric distinguishes it from successful processing. Handler failures, such as a presence-store error, are likewise invisible. Before this change they produced an unavailable/error sample. Add the planned bounded handler-duration/result and dropped-reason instruments in this cutover, or retain an explicit failure counter until they exist.

```

</details>

---

user-service/historyclient/client_test.go
```diff
@@ -119,8 +124,194 @@ func TestRoomsGet_Hints(t *testing.T) {
...
+// whether the call failed.
+//
+// Each case asserts the method as well as the class, because asserting only the
+// class would pass if one call were recorded twice under different labels.
+func TestHistoryClient_RecordsClientCall(t *testing.T) {
+	tests := []struct {
+		name      string
+		subject   string
+		method    natsmetrics.RPCMethod
+		call      func(c *Client) error
```
julianshen, line 195: _🏷️ dead-code_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`dead-code`_

**The okReply test-case field is never used.**

No table entry initializes `okReply`, and the test never reads it; the cases use `subject`, `method`, `call`, `subscribe`, `reply`, `wantErr`, and `wantClass`. Remove the vestigial field.

<details>

```
In `user-service/historyclient/client_test.go` around line 195: No table entry initializes `okReply`, and the test never reads it; the cases use `subject`, `method`, `call`, `subscribe`, `reply`, `wantErr`, and `wantClass`. Remove the vestigial field.

```

</details>

---

pkg/natsrouter/routes_golden_coverage_test.go
```diff
@@ -0,0 +1,369 @@
...
+		}
+	}
+	return false, nil
+}
+
+// buildContextFor returns a build.Context that reads through the repo FS, so
+// MatchFile answers exactly as the toolchain would: //go:build and legacy
+// all of it, for the default build with no -tags.
+//
+// This replaced a hand-rolled evaluator that understood only GOOS, GOARCH and
```
julianshen, line 185: _🏷️ dead-code_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`dead-code`_

**buildContextFor ignores its dir argument.**

The caller passes `dir`, but the function only silences it with `_ = dir` at line 218. `MatchFile` receives the directory separately, so this argument cannot affect the returned context. Remove the parameter, its call-site argument, and the blank assignment.

<details>

```
In `pkg/natsrouter/routes_golden_coverage_test.go` around line 185: The caller passes `dir`, but the function only silences it with `_ = dir` at line 218. `MatchFile` receives the directory separately, so this argument cannot affect the returned context. Remove the parameter, its call-site argument, and the blank assignment.

```

</details>

---

.semgrep/rpcmethod.go
```diff
@@ -0,0 +1,157 @@
...
+//
+// The negative cases below therefore cover all three shapes, not just the
+// literal. The literal-only version of this rule passed both the conversion
+// and the variable.
+func registerCallSites(r *natsrouter.Router) {
+	var c clientLane
+	ctx := context.Background()
+	started := time.Now()
+	elapsed := time.Since(started)
+	var err error
```
julianshen, line 46: _🏷️ dead-code_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`dead-code`_

**The blank assignment to err has no effect.**

`err` is already read by the `RecordRPCClientCall` fixture calls below, so assigning it to the blank identifier neither prevents an unused-variable error nor exercises the semgrep rule. Remove this leftover scaffolding.

<details>

```
In `.semgrep/rpcmethod.go` around line 46: `err` is already read by the `RecordRPCClientCall` fixture calls below, so assigning it to the blank identifier neither prevents an unused-variable error nor exercises the semgrep rule. Remove this leftover scaffolding.

```

</details>

---

julianshen (commented):
### tGDBot review of `e47704d` — 5 findings in 106 files

1 🔴 blocking · 1 🟠 warning · 3 🔵 suggestion
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/475#issuecomment-5573272270), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

flywindy:
## Closing. Reopening as a clean branch that does one thing.

The goal here is unchanged and worth doing: `rpc_method` on `rpc_server_call_duration_seconds` is derived by parsing NATS subjects, the mapper only recognises room-service and history-service, and **51 of the fleet's 96 routes record as `rpc_method="unknown"`** across eight services. That is a real, observed, everyday defect on the latency panel — not a hypothetical.

What this branch got wrong is scope. It carries five commits, and only part of them serve that goal.

### Why not just amend it

Its first commit, `9310b6b refactor(errcode): one tested helper for reading a remote reply`, is the pre-review version of #477 — which was closed after verification showed its premise did not hold. Everything else on this branch sits on top of that commit, and it is not separable: reverting it leaves `undefined: errcode.FromReply` in `notification-worker` and two other services, because later commits adopted the helper. Rebuilding is cleaner than unpicking it, and it spares reviewers a history whose first chapter has been retracted.

To be explicit about a claim I made while splitting #477 out: I argued there that the envelope helper was load-bearing for this metric work. It is not. The metric defect in `notification-worker/presence.go` is that `resultErr` was never assigned on the error-envelope branch, so the recorder saw a success. That is fixed by one line — `resultErr = errResp` — with the existing `errcode.Parse`. The helper only extends coverage to malformed envelopes, which nothing in this fleet produces.

### What the new branch will carry

The label change and nothing else:

- split `Operation` into `PublishOperation` and `RPCMethod` (eleven RPC labels were living in the publish enum, multiplying into each other's label space)
- `Register` / `RegisterNoBody` / `RegisterOptionalBody` take an `RPCMethod`, so a route that declares none does not compile
- delete `RequestOperationFromSubject` and its four helpers
- the 96 route registrations, each naming its method
- per-service `testdata/routes.golden` files pinning method→subject

### What it will drop

- **`9310b6b`** — the errcode helper, per above.
- **`80353c0 build(sast): require a vocabulary constant in every method argument`** — the type already restricts the argument; this guards a deliberate `RPCMethod("…")` conversion. Same category as the semgrep rule dropped from #477.
- **`pkg/natsrouter/routes_golden_coverage_test.go`** (369 lines) — an AST scan making a golden file mandatory for any new package that registers a route. It guards "someone adds a service and forgets", which is speculative, and it is the same shape as the contract test #477 was asked to remove.
- **`40a133d feat(o11y): instrument the outbound RPC client lane`** — a separate feature, not part of fixing `unknown` on the server histogram. Worth doing on its own afterwards.

The golden files themselves stay. They are not future-proofing: this branch hand-assigns 96 constants in one pass, and the compiler can only check that *a* method was passed, never that the *right* one was. `r.Register(MethodListMembers, "room.{roomID}.member.add", …)` compiles, is a valid constant, and silently labels add-members as `list_members` — reintroducing the exact defect this work exists to fix. Eleven short golden files are also a far better review artifact than 96 registration calls scattered across handlers.

### On the open review threads

@julianshen — thank you for these, and three of the nits were right and will carry over. One correction on the 🔴 Blocking finding, which claimed `room-service/integration_test.go` calls `natsrouter.Register` with a stale signature: verified four ways, that file contains **zero** `natsrouter.Register` calls. Its nine `Register` hits are all `handler.Register(router)`. `go test -tags integration` compiles it, `go vet -tags integration ./...` exits 0, and CI was green. I should have replied at the time instead of leaving it open — sorry for the silence.

The `RegisterVoid` discards-results point stands and will be addressed in the new branch's description rather than left implicit.

New PR to follow; I will link it here.

---

PR #483, opened 2026-09-10
feat(o11y): declare rpc.method at route registration

flywindy:
## The problem, in one screenshot's worth

There is a latency alarm on `user-service` and no way to tell which API is slow.

**Before** — group `rpc_server_call_duration_seconds` by `rpc_method`:

```
rpc_method="unknown"          ██████████████████████  51 routes collapsed into one bar
rpc_method="room_read"        ████
rpc_method="history_read"     ███
rpc_method="member_read"      ██
...
```

**After** — every route is its own series:

```
rpc_method="list_channel_messages"    ████
rpc_method="get_chatlist"             ███
rpc_method="batch_get_rooms_info"     ██
rpc_method="search_messages"          █
... 92 distinct methods
```

That `unknown` bar was **51 of the fleet's 96 routes** — user-service all 29 of its own, user-presence-service all 7, bot-room-service and search-service 5 each, bot-message-handler and media-service 2 each, translation-service its 1. The slow route was in there, and nothing in the metric could say which.

**Why it happened:** the label was *guessed* at request time by pattern-matching the NATS subject. `natsmetrics.RequestOperationFromSubject`'s suffix rules only ran when the subject's family token was `room` or `orgs`; everything else fell to the default arm and became `unknown`. It also mislabelled: user-service's `chatlist.section.create` ends in `.create`, so it was recorded as `room_mutation`, borrowing room-service's label.

This PR stops guessing. **The route declares its own method name when it registers.**

---

## Before / After

**Before** — the label is derived from the subject, at runtime, by a mapper that lives somewhere else:

```go
natsrouter.Register(r, subject.SearchMessagesPattern(siteID), h.searchMessages)
//                     ↑ label inferred from this string, on every dispatch
```

**After** — the label is an argument, checked by the compiler:

```go
natsrouter.Register(r, subject.SearchMessagesPattern(siteID),
    natsmetrics.MethodSearchMessages, h.searchMessages)
//  ↑ a constant from a closed 92-value vocabulary in pkg/natsmetrics
```

| | Before | After |
|---|---|---|
| Where the name comes from | inferred from the subject string | declared at the call site |
| When it is computed | every dispatch, even with metrics disabled | once, at registration |
| Omit a route's name | silently becomes `unknown` | **does not compile** |
| Supply a meaningless name | n/a | records as `_OTHER` — bounded, never silent |
| Wrong-but-valid constant | invisible | still invisible — see *What does not protect this* |
| Where you look a route's label up | read the mapper's regexes | read the `Register` call, or the table at the end of this description |

**92 method-bearing routes map to 92 distinct label values** — one each, no constant shared between two routes and none declared but unused. That one-to-one is what makes `rpc_method` a fleet-wide key: `sum by (rpc_method)` can never silently merge two different routes, so pinning `service_name` in an SLO expression becomes a redundancy check rather than a correctness requirement.

Names are verb-first `snake_case` per Google AIP-131/132/190, and name what the handler *does* rather than what its subject spells — `mark_room_read` advances a room read position and carries no message id, despite a `.message.` subject.

| Prefix | Means |
|---|---|
| `get_` | one resource, or one view of one |
| `list_` | a collection — whether the response carries a total is irrelevant |
| `batch_get_` | several specific resources, by keys the caller supplied |
| `search_` | a query: matching, ranking, or scoring |

---

## ⚠️ What breaks

**Every `rpc_method` value changes.** This is not additive — old values stop existing. There is no compatibility shim, because a shim would preserve exactly the ambiguity this PR exists to remove.

| Old value | Becomes |
|---|---|
| `unknown` | **gone** — split into 51 named methods, *and* the presence heartbeats leave this metric entirely |
| `channel_history` | `list_channel_messages` |
| `thread_open` | `list_thread_messages` |
| `history_read` | `get_message`, `batch_get_messages`, `list_next_messages`, `list_surrounding_messages`, `list_pinned_messages`, `list_thread_parent_messages` |
| `room_read`, `room_mutation`, `member_read`, `member_mutation`, `history_mutation`, `teams_room` | individual method names |

**What you need to do:** any Grafana panel, alert or recording rule filtering on `rpc_method` needs a pass. Two traps specifically:

- A panel that *excludes* `rpc_method="unknown"` should have that filter **deleted**, not left dangling — the series is gone, not smaller.
- The presence `hello`/`ping`/`activity`/`bye` handlers now emit **no RPC sample at all**. They were never round trips (no reply subject), so counting them as RPC latency was wrong — but until hmchangw/newchat#473 lands they have *no* latency metric and, more importantly, no result classification either. That is a deliberate, documented gap, not an oversight. #473 has no owner or milestone yet, and this PR now carries the test it names as a precondition.

**SLO-4 and SLO-5 change label values**, and neither is measured or alerting yet, so the rename needs no cutover window. `docs/load-testing/common/sli-slo.md` §3 carries the updated expressions. The same change resolves what that section listed as residual contamination: `.msg.next`, `.msg.surrounding` and `.msg.thread.parent` each carry their own method now instead of sharing `history_read` with the routes an SLO describes.

---

## The one design decision worth arguing about

A client's `mark_all_threads_read` on user-service fans out per owning site to room-service's route on `chat.server.request.room.{site}.thread.read.all` — genuinely two halves of one action across a hop. An earlier draft of this work gave both routes the **same** `rpc_method`, on the reasoning that the identity key is `service_name + rpc_method` and the shared name expresses the hop.

This PR names them separately (`mark_all_threads_read` and `mark_site_threads_read`), because:

- A shared name costs `rpc_method` its primary key. `sum by (rpc_method)` — the most natural aggregation — then adds one outer call to N inner ones, so the count is wrong and the latency distribution is bimodal. It also makes forgetting `service_name` in an SLO expression *structurally* wrong for at least one method, rather than hypothetically wrong.
- A call chain belongs in a trace, not a metric label. The repo already runs OTel traces with trace-correlated logs; a hop is a parent/child span, which is where the per-leg breakdown and the fan-out ratio actually live, with a request id tying them.

The two names also record a real difference in scope: `mark_all_threads_read` spans every site the account holds threads on, while `mark_site_threads_read` clears exactly one. The subject's `.room.` token names the owning service, not a room — there is no roomID in either the subject or the request.

The hop is still expressible, in the right place: if the outbound client lane is instrumented later, user-service's call should carry `rpc_method="mark_site_threads_read"` — the method actually being called, per semconv. The same name then appears on one `rpc.client` series and one `rpc.server` series, which is semconv working as intended, and joining them across lanes gives the two-halves view without two *inbound* routes ever sharing a name.

**`Operation` keeps its name.** The eight coarse values the server lane used are deleted, but three the outbound client lane still passes remain (`history_get_message`, `member_read`, `presence_lookup`), so the enum is not purely publish — renaming it to `PublishOperation` would be a false name.

---

## What protects this, and what does not

| Gate | Catches | Misses |
|---|---|---|
| **Compiler** (`method natsmetrics.RPCMethod` is a required parameter) | a route with no method, a swapped argument order | string literals, `RPCMethod("…")` casts, variables |
| **Registration shape** (`addRoute` takes `recordRPC` from the entry point) | a request/reply route silently opting out of recording | — |
| **`rpcmethod_test.go`** | a constant declared but unreachable, a duplicate or malformed value, a vocabulary that stops being closed | which constant a given route passes |
| **`normalizeRPCMethod`** | any out-of-vocabulary value at the record site — bounded to semconv's `_OTHER`, never unbounded cardinality, never a panic, never absent | — |

**What does not protect this:** a *valid but wrong* constant — `MethodGetMessage` on the rename route — compiles, passes every test here, and mislabels silently. Nothing in this PR catches it; the 92 rows below are the review surface. Review found exactly one such defect on this branch (a scope-wrong name, fixed in `3e1c654`), which is a fair sample of how this gap behaves in practice.

The earlier draft added per-service golden files (`testdata/routes.golden`) and a semgrep rule requiring a `natsmetrics.Method*` selector at each call site. **Both are deliberately left out of this PR** to keep it to the vocabulary and the wiring. Note they would not have prevented that defect either — a golden file records whatever name you regenerate it with; it makes a wrong name a diff line someone has to read.

---

## Reading order

| Commit | |
|---|---|
| `cf2178e` | `feat(o11y)`: add the `rpc.method` vocabulary — the 92 constants and their naming rule |
| `56a84bf` | `feat(o11y)`: declare `rpc.method` at route registration — **the behaviour change is all here**, plus the deletion of `RequestOperationFromSubject` and the doc updates |
| `adc723a` | `fix(o11y)`: give the room-scoped thread-read route its own method |
| `180a914` | `fix(sast)`: suppress the repo's own hardcoded-credential rule too |
| `8a5109d` | `test(o11y)`: pin the two invariants this change introduces |
| `d201681` | `fix(o11y)`: decide RPC recording by registration shape, not method value — CodeRabbit finding |
| `3e1c654` | `fix(o11y)`: correct a wrong scope name and three false doc claims — review findings |

## Route coverage

| Service | Routes | Client-facing | Server-to-server |
|---|--:|--:|--:|
| `user-service` | 29 | 28 | 1 |
| `room-service` | 27 | 22 | 5 |
| `history-service` | 17 | 13 | 4 |
| `bot-room-service` | 5 | 0 | 5 |
| `search-service` | 5 | 5 | 0 |
| `user-presence-service` | 3 | 1 | 2 |
| `bot-message-handler` | 2 | 0 | 2 |
| `media-service` | 2 | 2 | 0 |
| `room-worker` | 1 | 0 | 1 |
| `translation-service` | 1 | 1 | 0 |
| **total** | **92** | **72** | **20** |

<details>
<summary><b>All 92 routes and their declared <code>rpc.method</code></b> — the review surface for a valid-but-wrong constant</summary>

**`user-service`** — 29

| `rpc.method` | Subject | Shape |
|---|---|---|
| `add_priority_contact` | `chat.user.{account}.request.user.{site}.settings.priorityContacts.add` | |
| `batch_get_badge_counts` | `chat.server.request.user.{site}.badge.count.batch` | |
| `count_subscriptions` | `chat.user.{account}.request.user.{site}.subscription.count` | |
| `create_chatlist_section` | `chat.user.{account}.request.user.{site}.chatlist.section.create` | |
| `delete_chatlist_section` | `chat.user.{account}.request.user.{site}.chatlist.section.delete` | |
| `get_chatlist` | `chat.user.{account}.request.user.{site}.chatlist.get` | no body |
| `get_current_user` | `chat.user.{account}.request.user.{site}.me` | no body |
| `get_dm_subscription` | `chat.user.{account}.request.user.{site}.subscription.getDM` | |
| `get_settings` | `chat.user.{account}.request.user.{site}.settings.get` | no body |
| `get_subscription_by_room` | `chat.user.{account}.request.user.{site}.subscription.getByRoomID` | |
| `get_thread_unread_summary` | `chat.user.{account}.request.user.{site}.thread.unread.summary` | |
| `get_user_profile` | `chat.user.{account}.request.user.{site}.profile.getByName` | |
| `get_user_status` | `chat.user.{account}.request.user.{site}.status.getByName` | |
| `list_app_categories` | `chat.user.{account}.request.user.{site}.apps.categories` | no body |
| `list_apps` | `chat.user.{account}.request.user.{site}.apps.list` | |
| `list_channel_subscriptions` | `chat.user.{account}.request.user.{site}.subscription.getChannels` | |
| `list_priority_contacts` | `chat.user.{account}.request.user.{site}.settings.priorityContacts.get` | no body |
| `list_subscriptions` | `chat.user.{account}.request.user.{site}.subscription.list` | |
| `list_user_threads` | `chat.user.{account}.request.user.{site}.thread.list` | |
| `mark_all_threads_read` | `chat.user.{account}.request.user.{site}.thread.read.all` | |
| `refresh_sso_token` | `chat.user.{account}.request.user.{site}.sso.refresh` | opt body |
| `remove_priority_contact` | `chat.user.{account}.request.user.{site}.settings.priorityContacts.remove` | |
| `rename_chatlist_section` | `chat.user.{account}.request.user.{site}.chatlist.section.rename` | |
| `reorder_chatlist_sections` | `chat.user.{account}.request.user.{site}.chatlist.section.reorder` | |
| `set_app_subscription` | `chat.user.{account}.request.user.{site}.subscription.setAppSubscription` | |
| `set_chatlist_section_sort_mode` | `chat.user.{account}.request.user.{site}.chatlist.section.setsortmode` | |
| `set_settings` | `chat.user.{account}.request.user.{site}.settings.set` | |
| `set_sso_token` | `chat.user.{account}.request.user.{site}.sso.set` | |
| `set_user_status` | `chat.user.{account}.request.user.{site}.status.set` | |

**`room-service`** — 27

| `rpc.method` | Subject | Shape |
|---|---|---|
| `add_members` | `chat.user.{account}.request.room.{roomID}.{site}.member.add` | |
| `batch_get_rooms_info` | `chat.server.request.room.{site}.info.batch` | |
| `batch_get_thread_rooms_info` | `chat.server.request.room.{site}.thread.info.batch` | |
| `create_room` | `chat.user.{account}.request.room.{site}.create` | |
| `create_teams_meeting` | `chat.user.{account}.request.room.{roomID}.{site}.teams.meeting` | |
| `ensure_room_key` | `chat.server.request.room.{site}.key.ensure` | |
| `get_room_app_command_menu` | `chat.user.{account}.request.room.{roomID}.{site}.app.cmd-menu` | no body |
| `get_room_app_tabs` | `chat.user.{account}.request.room.{roomID}.{site}.app.tabs` | no body |
| `get_room_key` | `chat.user.{account}.request.room.{roomID}.{site}.key.get` | no body |
| `list_member_statuses` | `chat.user.{account}.request.room.{roomID}.{site}.member.statuses` | no body |
| `list_members` | `chat.user.{account}.request.room.{roomID}.{site}.member.list` | no body |
| `list_mentionable_subscriptions` | `chat.user.{account}.request.room.{roomID}.{site}.subscription.mentionable` | no body |
| `list_message_readers` | `chat.user.{account}.request.room.{roomID}.{site}.message.read-receipt` | |
| `list_org_members` | `chat.user.{account}.request.orgs.{orgID}.{site}.members` | no body |
| `mark_room_read` | `chat.user.{account}.request.room.{roomID}.{site}.message.read` | no body |
| `mark_site_threads_read` | `chat.server.request.room.{site}.thread.read.all` | |
| `mark_thread_read` | `chat.user.{account}.request.room.{roomID}.{site}.message.thread.read` | |
| `move_chat` | `chat.user.{account}.request.room.{roomID}.{site}.chat.move` | |
| `open_room` | `chat.user.{account}.request.room.{roomID}.{site}.open` | no body |
| `remove_member` | `chat.user.{account}.request.room.{roomID}.{site}.member.remove` | |
| `rename_room` | `chat.user.{account}.request.room.{roomID}.{site}.room.rename` | |
| `set_room_restricted` | `chat.server.request.room.{site}.restricted` | |
| `start_teams_room_call` | `chat.user.{account}.request.room.{roomID}.{site}.teams.call` | |
| `start_teams_user_call` | `chat.user.{account}.request.teams.{site}.call.user` | |
| `toggle_favorite` | `chat.user.{account}.request.room.{roomID}.{site}.favorite.toggle` | no body |
| `toggle_mute` | `chat.user.{account}.request.room.{roomID}.{site}.mute.toggle` | no body |
| `update_member_role` | `chat.user.{account}.request.room.{roomID}.{site}.member.role-update` | |

**`history-service`** — 17

| `rpc.method` | Subject | Shape |
|---|---|---|
| `batch_get_messages` | `chat.user.{account}.request.room.{roomID}.{site}.msg.get.ids` | |
| `batch_get_room_previews` | `chat.server.request.history.{site}.rooms.get` | |
| `delete_message` | `chat.user.{account}.request.room.{roomID}.{site}.msg.delete` | |
| `edit_message` | `chat.user.{account}.request.room.{roomID}.{site}.msg.edit` | |
| `get_message` | `chat.user.{account}.request.room.{roomID}.{site}.msg.get` | |
| `list_channel_messages` | `chat.user.{account}.request.room.{roomID}.{site}.msg.history` | |
| `list_next_messages` | `chat.user.{account}.request.room.{roomID}.{site}.msg.next` | |
| `list_pinned_messages` | `chat.user.{account}.request.room.{roomID}.{site}.msg.pinned.list` | |
| `list_surrounding_messages` | `chat.user.{account}.request.room.{roomID}.{site}.msg.surrounding` | |
| `list_thread_messages` | `chat.user.{account}.request.room.{roomID}.{site}.msg.thread` | |
| `list_thread_parent_messages` | `chat.user.{account}.request.room.{roomID}.{site}.msg.thread.parent` | |
| `list_thread_subscriptions` | `chat.server.request.thread.{site}.subscription.list` | |
| `migrate_delete_message` | `chat.migration.internal.{site}.msg.delete` | |
| `migrate_edit_message` | `chat.migration.internal.{site}.msg.edit` | |
| `pin_message` | `chat.user.{account}.request.room.{roomID}.{site}.msg.pin` | |
| `toggle_message_reaction` | `chat.user.{account}.request.room.{roomID}.{site}.msg.react` | |
| `unpin_message` | `chat.user.{account}.request.room.{roomID}.{site}.msg.unpin` | |

**`bot-room-service`** — 5

| `rpc.method` | Subject | Shape |
|---|---|---|
| `add_bot_room_members` | `chat.server.bot.request.room.{site}.{roomID}.member.add` | |
| `create_bot_room` | `chat.server.bot.request.room.{site}.create` | |
| `ensure_bot_dm_room` | `chat.server.bot.request.room.{site}.dm.ensure` | |
| `get_bot_room` | `chat.server.bot.request.room.{site}.get` | |
| `remove_bot_room_members` | `chat.server.bot.request.room.{site}.{roomID}.member.remove` | |

**`search-service`** — 5

| `rpc.method` | Subject | Shape |
|---|---|---|
| `search_apps` | `chat.user.{account}.request.search.{site}.apps` | |
| `search_messages` | `chat.user.{account}.request.search.{site}.messages` | |
| `search_orgs` | `chat.user.{account}.request.search.{site}.orgs` | |
| `search_rooms` | `chat.user.{account}.request.search.{site}.rooms` | |
| `search_users` | `chat.user.{account}.request.search.{site}.users` | |

**`user-presence-service`** — 3

| `rpc.method` | Subject | Shape |
|---|---|---|
| `batch_get_peer_presence` | `chat.server.request.presence.{site}.query.batch` | |
| `batch_get_presence` | `chat.user.presence.{site}.query.batch` | |
| `set_manual_presence` | `chat.user.{account}.request.presence.{site}.manual.set` | |

**`bot-message-handler`** — 2

| `rpc.method` | Subject | Shape |
|---|---|---|
| `send_dm` | `chat.server.bot.request.dm.{site}.{userID}.msg.send` | |
| `send_room_message` | `chat.server.bot.request.room.{site}.{roomID}.msg.send` | |

**`media-service`** — 2

| `rpc.method` | Subject | Shape |
|---|---|---|
| `delete_emoji` | `chat.user.{account}.request.emoji.{site}.delete` | |
| `list_emojis` | `chat.user.{account}.request.emoji.{site}.list` | no body |

**`room-worker`** — 1

| `rpc.method` | Subject | Shape |
|---|---|---|
| `create_dm_room` | `chat.server.request.room.{site}.create.dm` | |

**`translation-service`** — 1

| `rpc.method` | Subject | Shape |
|---|---|---|
| `translate_text` | `chat.user.{account}.request.translate.{site}.text` | |

### Records no sample (`RegisterVoid`, declares no method)

| Handler | Subject |
|---|---|
| `Hello` | `chat.user.{account}.event.presence.{site}.hello` |
| `Ping` | `chat.user.{account}.event.presence.{site}.ping` |
| `Activity` | `chat.user.{account}.event.presence.{site}.activity` |
| `Bye` | `chat.user.{account}.event.presence.{site}.bye` |

</details>

Docs updated: `docs/specs/o11y/nats-metrics-contract.md`, `docs/specs/o11y/o11y-metrics-inventory.md`, `docs/load-testing/common/sli-slo.md`, `tools/observability/METRICS.md`, `pkg/natsrouter/README.md`.

`docs/client-api.md` is unchanged on purpose: this touches the registration of client-facing handlers, but no request/response schema, error case or triggered event changes — only a metrics label argument.

## Verification

Locally on the current head: `make test` (140 packages, 0 failures), `make lint` (0 issues), `make sast-gosec` (exit 0), `make sast-semgrep-test`, a full-tree `semgrep --config=.semgrep/` sweep (0 findings), and `go vet -tags=integration ./...` — that last one matters because `make test` does not carry the integration tag, so the integration files this branch touches are otherwise never compiled locally.

`govulncheck` and the semgrep **registry** scan (`p/golang`, `p/security-audit`) cannot run in this environment — the proxy blocks `vuln.go.dev` and `semgrep.dev` — so CI is their first real execution. That gap already cost one red `sast` run: the four SSO-token method constants carried only a gosec `// #nosec G101`, which left the repo's own `hardcoded-credential-literal` rule reporting. Each scanner reads its own directive, so both are now named (`180a914`), and the fix was verified by running the repo-owned config alone.

## Summary by CodeRabbit

- **New Features**
  - RPC metrics now use explicit, standardized method names for accurate route-level monitoring.
  - Invalid or unspecified method labels are reported as `_OTHER`.
  - Routes without reply round trips no longer produce RPC duration samples.

- **Documentation**
  - Updated observability, metrics, and load-testing guidance with the new method labels and recording behavior.

- **Bug Fixes**
  - Corrected channel and thread message metrics labels, replacing outdated classifications.


flywindy:
@coderabbitai review

flywindy:
## On deriving `rpc.method` instead of declaring it

The suggestion came up that 92 explicit method arguments is a lot of typing, and that reflection could derive the name instead. That is a fair objection to the cost, so rather than argue it, I measured every derivation source available against the 92 names this PR declares.

**Summary: nothing that already exists in the code names the operation, because none of it was created to.**

| Source | Stability | What it actually describes | Matches the declared name |
|---|---|---|--:|
| Subject string (the approach this PR removes) | Frozen by deployed clients | routing path | ≤47%, coarse — 51 of 96 routes fell to `unknown` |
| Subject builder name (`subject.MsgHistoryPattern`) | Stable | subject *shape* | **7%** (7/92) |
| Handler func name via reflection (`h.searchMessages`) | **Changes freely on any refactor** | internal implementation | **40%** (37/92) |
| Explicit constant (this PR) | Changes only deliberately | the operation | 100% by construction |

---

### Where the suggestion is right

For a service whose handlers are already named after the operation, reflection is free and exact. search-service is 5 for 5:

```go
h.searchMessages → search_messages     ✅
h.searchRooms    → search_rooms        ✅
h.searchApps     → search_apps         ✅
h.searchUsers    → search_users        ✅
h.searchOrgs     → search_orgs         ✅
```

If the whole fleet looked like this, the argument would be pure toil and I would drop it. It doesn't: 48 of the 85 named handlers produce a different name, and 7 more produce no usable name at all.

### 1. Handler names: 48 of 85 differ, and not by accident

```
bot-room-service   handleCreate      → handle_create        (needs create_bot_room)
bot-room-service   handleAdd         → handle_add           (needs add_bot_room_members)
bot-room-service   handleGet         → handle_get           (needs get_bot_room)
history-service    LoadHistory       → load_history         (needs list_channel_messages)
history-service    RoomsGet          → rooms_get            (needs batch_get_room_previews)
history-service    GetMessagesByIDs  → get_messages_by_i_ds (needs batch_get_messages)
```

`handleCreate` is unambiguous in Go because the receiver scopes it. A Prometheus label has no receiver. And `load_history` is the label **SLO-4 is defined against** — it does not say "channel message list", which is precisely the distinction that made SLO-4 and SLO-5 measurable in the first place.

### 2. Seven routes have no name to reflect on

Seven registrations pass a closure rather than a method value. Go names closures positionally:

```
exported method value  : main.(*Handler).SearchMessages-fm
unexported method value: main.(*Handler).loadHistory-fm
closure                : main.main.func1
second closure         : main.main.func2
```

So those seven all collapse to `func1` — the `unknown` bar coming back, smaller. Worse, the numbering is positional: inserting an unrelated closure earlier in the file renumbers every one after it, silently renaming live metric series.

### 3. Renaming a Go function would silently rename a metric label

This is the part that worries me most. `LoadHistory` → `FetchHistory` is a pure refactor that nothing in this repo prevents. Under reflection it also rewrites the `rpc_method` label that dashboards and SLO recording rules select on, and the diff a reviewer sees is a function rename. Today the coupling runs the other way: changing a label means editing a constant in `pkg/natsmetrics/rpcmethod.go`, where it is visible.

### 4. Subject-derived names measure worse, not better — 7%

Subjects are the most stable thing here: 263 builders in `pkg/subject`, 170 assertions pinning the literal strings, 137 of them documented in `docs/client-api.md` as the client contract. You cannot casually rename one.

But they are frozen by *routing*, not by *meaning*, and the builder names follow the subject shape rather than the operation:

```
MsgHistoryPattern              → msg_history                  (needs list_channel_messages)
ServerBotRoomCreate            → server_bot_room_create        (needs create_bot_room)
ServerBotMsgRoomSendPattern    → server_bot_msg_room_send      (needs send_room_message)
```

That is the approach this PR removes, wearing a different hat, and it measures worse than the version being removed.

The clearest single example is `room-service/handler.go:140`:

```go
RegisterNoBody(r, subject.MessageReadPattern(...), natsmetrics.MethodMarkRoomRead, h.messageRead)
```

The subject says `.message.read`. The operation advances a **room**'s read position and carries no message id at all. The subject is frozen against a naming decision made long ago — frozen is not the same as correct.

---

### The shape of the problem

The stable identifiers (subject, builder name) describe transport, not the operation. The identifier that comes closest to describing the operation (the handler name) is the least stable thing in the system — it is supposed to be free to rename.

So the 92 arguments are not purely redundant typing. They are the only place that records what a route is called *to the outside*, independent of how it is routed or how it is implemented. That is the same reason the label is worth having at all.

### On the cost itself

The 92 one-time arguments are already written. The ongoing cost is one constant plus one argument per new route, which seems proportionate for a value that SLOs select on.

If we want that smaller, two options that do not reintroduce a derivation:

- **A per-service registration table** — collapses scattered calls into one reviewable block, and incidentally covers the one gap this PR admits to (a valid-but-wrong constant). Works where handler signatures are uniform; generic type parameters prevent a single table where they are not.
- **Per-service golden files** (`testdata/routes.golden`) — no less typing, but the 92 route-to-method pairs become one file, so a change is a one-line diff instead of something to spot across 10 services. This existed in an earlier draft and was deliberately left out to keep this PR to the vocabulary and the wiring; it is easy to add back.

Happy to be argued out of this if someone sees a source I have not measured.

---

flywindy:
## Correction: golden files do not prevent a missed name

In my previous comment I said per-service golden files would "cover the one gap this PR admits to". That is wrong and I want to retract it before anyone plans around it.

A golden file works like this: the route table changes → the test fails → you regenerate the file → **the wrong name lands in the golden file**. It converts a silent change into a reviewable diff line. Whether that catches anything depends entirely on a human reading that line. It *surfaces*, it does not *prevent*. Same for the registration-table idea — that is legibility, not enforcement.

The underlying concern is a better question than the typing cost, so here is what "missing one" actually decomposes into.

### What happens today if you get it wrong

| Failure mode | What happens now | Enforced? |
|---|---|---|
| New route, no method argument | **does not compile** | prevented |
| Constant that isn't in the vocabulary | **does not compile** | prevented |
| `RPCMethod("typo")` or `MethodNone` on a request/reply route | records `_OTHER` — bounded and alertable, but not the right name | contained, not correct |
| **Copy the line above, forget to change the method** | **completely silent** — two routes merge into one series | not caught |
| A new service that never uses `natsrouter` | no `rpc.server.call.duration` at all — nothing to notice its absence | not caught |

The fourth row is the realistic one. Copy the neighbouring registration, change the subject and the handler, miss the method.

### That one is mechanically detectable

A copy-paste produces a **duplicate** method within the same router — and duplication can be checked without knowing which name is correct. Measured on the current head: **92 registrations, 92 distinct methods, 0 duplicates.** So an assertion that methods are unique per router passes today and fails the moment someone copy-pastes, with no diff-reading required.

That is prevention rather than surfacing, and it is much cheaper than a golden file. It does not catch a constant copied from a *different* service (fleet-wide uniqueness still holds in that case), but `service_name` separates those at query time.

I have not added it to this PR — flagging it as the concrete answer to the "what stops us missing one" question, to add here or as a follow-up, whichever reviewers prefer.

### The row nothing here can catch

A service that never registers through `natsrouter` gets no label because it gets no metric. That is not a type problem, so no compile-time gate reaches it.

For this repo I checked, and it does not currently apply: every request/reply route goes through `natsrouter`. The only non-router `QueueSubscribe` callers outside tests are `broadcast-worker` and `inbox-worker`, both JetStream event consumers with no reply subject.

### Open question: services outside this repository

This is the part worth deciding before merge, and it is not addressed by anything above.

The vocabulary is an unexported map in `pkg/natsmetrics`, and `Valid()` only recognises what is in it. A service in another repository importing `pkg/natsrouter` therefore has two options, both bad:

1. Open a PR **against this repository** for every new route, to add its constant — a cross-repo bottleneck on ordinary feature work.
2. Pass `RPCMethod("its_method")` — `Valid()` is false, so **every one of its routes records as `_OTHER`**. That is `unknown` under a new name.

If those services do use `pkg/natsrouter`, this needs solving. The property that actually bounds cardinality is "the set is fixed at startup", not "the set is listed in this file" — so an exported declaration hook (a service declaring its own methods in `init`) would preserve the bound without the cross-repo round trip. That is a design change beyond this PR's scope, so I am raising it rather than making it.

If those services run their own router instead, they were never emitting this metric and it is a separate piece of work.

---

julianshen:
**4 findings · 4 inline comments posted** — 🟠 3 warning · 🔵 1 suggestion

### Related work

- [Issue #473](https://github.com/hmchangw/newchat/issues/473) — o11y: add handler telemetry for RegisterVoid presence routes (open)

<details>
<summary>📒 Files reviewed (35)</summary>

* `bot-message-handler/handler.go`
* `docs/load-testing/common/sli-slo.md`
* `media-service/emoji_nats.go`
* `pkg/natsmetrics/enums_test.go`
* `pkg/natsmetrics/opttable_test.go`
* `pkg/natsmetrics/prometheus_export_test.go`
* `pkg/natsmetrics/rpcmethod.go`
* `pkg/natsmetrics/rpcmethod_test.go`
* `pkg/natsmetrics/rpcsemconv.go`
* `pkg/natsmetrics/rpcsemconv_test.go`
* `pkg/natsmetrics/toggle_test.go`
* `pkg/natsrouter/README.md`
* `pkg/natsrouter/example_test.go`
* `pkg/natsrouter/integration_test.go`
* `pkg/natsrouter/oversize_integration_test.go`
* `pkg/natsrouter/router.go`
* `pkg/natsrouter/shutdown_test.go`
* `room-service/handler.go`
* `room-worker/main.go`
* `search-service/handler.go`
* `translation-service/integration_test.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

pkg/natsrouter/register.go
```diff
@@ -104,7 +110,7 @@ func RegisterVoid[Req any](
 		}
 	})
 
-	r.addRoute(pattern, []HandlerFunc{handler})
+	r.addRoute(natsmetrics.MethodNone, false, pattern, []HandlerFunc{handler})
```
julianshen, line 113: _🏷️ observability_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`distributed-system`_

**Void-route drops disappear from telemetry before replacement exists**

When concurrency is saturated or shutdown has begun, `addRoute` rejects a presence event before its handler runs. Passing `false` makes `record` nil, so neither that `RequestUnavailable` outcome nor successful/failed processing produces a metric. Because these routes are fire-and-forget, the publisher receives no acknowledgement and cannot detect the loss; presence may remain stale while metric-based monitoring shows no accepted or dropped work. Keep the existing sample until dedicated void-handler telemetry records admission and terminal outcome, then remove the semantically incorrect RPC-duration sample.

<details>

```
In `pkg/natsrouter/register.go` around line 113: When concurrency is saturated or shutdown has begun, `addRoute` rejects a presence event before its handler runs. Passing `false` makes `record` nil, so neither that `RequestUnavailable` outcome nor successful/failed processing produces a metric. Because these routes are fire-and-forget, the publisher receives no acknowledgement and cannot detect the loss; presence may remain stale while metric-based monitoring shows no accepted or dropped work. Keep the existing sample until dedicated void-handler telemetry records admission and terminal outcome, then remove the semantically incorrect RPC-duration sample.

```

</details>

---

pkg/natsmetrics/rpcmethod_test.go
```diff
@@ -0,0 +1,151 @@
...
+		seen[m] = struct{}{}
+		assert.Regexp(t, snakeCase, string(m), "method %q is not verb-first snake_case", m)
+	}
+}
+
+// TestRPCMethodVocabularyIsComplete catches the list above drifting from the
+// const block. Valid() is driven by its own table, so a constant added to one
+// and not the other shows up as a count mismatch here rather than as a route
+// silently recording under MethodOther in production.
+func TestRPCMethodVocabularyIsComplete(t *testing.T) {
```
julianshen, line 148: _🧪 tests_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**The completeness test does not inspect the constant declarations**

This only compares `allRPCMethods` with `rpcMethodVocabulary`. If a new `MethodFoo` constant is added and used by a route but omitted from both collections, their lengths remain equal, `TestEveryRPCMethodIsValid` never visits it, and production normalizes it to `_OTHER`. Derive the vocabulary from one source or parse the declared `RPCMethod` constants in this test.

<details>

```
In `pkg/natsmetrics/rpcmethod_test.go` around line 148: This only compares `allRPCMethods` with `rpcMethodVocabulary`. If a new `MethodFoo` constant is added and used by a route but omitted from both collections, their lengths remain equal, `TestEveryRPCMethodIsValid` never visits it, and production normalizes it to `_OTHER`. Derive the vocabulary from one source or parse the declared `RPCMethod` constants in this test.

```

</details>

---

pkg/natsmetrics/opttable_test.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 // A key that never occurs must never be built. The publisher's label space is
-// closed but wide — destination x operation x outcome is 1,848 combinations,
+// closed but wide — destination x operation x outcome is 924 combinations,
```
julianshen, line 14: _📝 documentation_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**The documented publisher cross product is miscalculated**

The changed comment says 924 combinations, while `TestLabelSpaceStaysWithinBudget` asserts 1,176 (`12 destinations × 14 operations × 7 outcomes`). The assertion message near line 102 repeats 924. Update both occurrences to 1,176 so the allocation rationale and failure diagnostics match the enforced label budget.

<details>

```
In `pkg/natsmetrics/opttable_test.go` around line 14: The changed comment says 924 combinations, while `TestLabelSpaceStaysWithinBudget` asserts 1,176 (`12 destinations × 14 operations × 7 outcomes`). The assertion message near line 102 repeats 924. Update both occurrences to 1,176 so the allocation rationale and failure diagnostics match the enforced label budget.

```

</details>

---

pkg/natsrouter/example_test.go
```diff
@@ -66,6 +64,27 @@ type Room struct {
...
+// rpc.method names. A rename route taking a greeting payload would be exactly
+// the method-does-not-match-the-handler mismatch this vocabulary exists to
+// prevent, demonstrated in the package's own godoc.
+type RenameRoomRequest struct {
+	Name string `json:"name"`
+}
+
+type OpenRoomRequest struct {
+	IncludeMembers bool `json:"includeMembers"`
+}
```
julianshen, line 77: _🏷️ dead-code_ | _🔵 Suggestion_ | _⚡ Quick fix_ | _`dead-code`_

**The example decodes a request field it never uses.**

`Example_errorHandling` accepts `req OpenRoomRequest`, but its handler never references `req`, so `IncludeMembers` cannot affect the response or error path. Use `RegisterNoBody` and remove `OpenRoomRequest`, or consume the field if the example is intended to demonstrate a request body.

<details>

```
In `pkg/natsrouter/example_test.go` around line 77: `Example_errorHandling` accepts `req OpenRoomRequest`, but its handler never references `req`, so `IncludeMembers` cannot affect the response or error path. Use `RegisterNoBody` and remove `OpenRoomRequest`, or consume the field if the example is intended to demonstrate a request body.

```

</details>

---

julianshen (commented):
### tGDBot review of `3e1c654` — 4 findings in 35 files

3 🟠 warning · 1 🔵 suggestion
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/483#issuecomment-5622373566), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

hmchangw (approved):
thank you so much!

PR #486, opened 2026-09-10
Valkey timeout and startup hardening

hmchangw:
Closes the two mechanical traps that let a Valkey outage escalate into a chat outage. Neither is a logic bug in the cache consumers — both live in how the client is constructed. Both were verified present on `main` before this branch was written.

**The timeout trap.** `dialCluster` passed only `Addrs` and `Password`, so go-redis cluster defaults applied and `ContextTimeoutEnabled` was `false` — a caller's context deadline did not bound socket reads at all. Against a Valkey that *blackholes* packets (the common degraded mode, as opposed to one that refuses connections), every hot-path read stalled for seconds before falling through to Mongo or Elasticsearch. The consumers were fail-open; the network behaviour was fail-slow.

**The startup trap.** The `PING` gated the dial, and ten call sites exit on that error. A shared datastore is the same for every replica, so an outage overlapping a rollout, autoscale or node drain crashloops every pod at once — including the message path — and the crashloop outlives the outage.

## Review order

`pkg/valkeyutil` is 12 of 24 files and everything else depends on its contract, so read it first. Each commit is one self-contained change and builds and vets independently (verified), so commit-by-commit works.

| Commit | Scope |
|---|---|
| `ff321c02` | `pkg/valkeyutil`: timeout profiles + non-fatal startup probe |
| `f6be4b0b` | `pkg/valkeyutil`: `IncrEx` repairs a counter left without an expiry |
| `f8ae897a` | `botplatform-service`: fail-open posture + bypass counter |
| `7ef2c0d9` | `pkg/valkeyutil`: throttle the per-call log when degraded |
| `f40dd2bf` | `docs/health-probes.md` |

The order is load-bearing, not cosmetic: `7ef2c0d9` argues from the timeout profiles (a failure now costs ~150ms, so nothing throttles the log rate any more), which only holds because `ff321c02` sits beneath it.

## Behaviour changes

These are what change in production, and the reason to read the PR rather than the diff:

1. **An unreachable Valkey no longer fails the dial.** `Connect`/`ConnectRaw` return a usable client that self-heals on the next call. The ten `os.Exit`/`return err` branches keep their purpose — they now cover construction and instrumentation failures, not reachability. `WithRequireReachable()` restores fail-fast and only `tools/seed-sample-data` uses it: a one-shot CLI has no fallback and no next call to self-heal into, so aborting the run beats seeding half the data. Default-off so the safe path is what a new caller gets without thinking.
2. **Every Valkey client is now bounded.** `CacheProfile` (150ms reads) for tiers with a fallback, `StoreProfile` (500ms) for `user-presence-service`, which has none. Note the cluster default is `MaxRetries: -1` — no retries at all — so these are *additive*: worst case is `(MaxRetries+1) × ReadTimeout` plus backoff.
3. **`botplatform-service` fails open.** During an outage bot requests are admitted **unthrottled and without duplicate suppression**. Bots are critical, so a lost ceiling beats a dead bot — a conscious trade, not an oversight. `bot_control_bypassed_total{control}` is the only durable signal that protection was off; alert on it.
4. **`IncrEx` changes wire behaviour** from `INCR` + conditional `EXPIRE` to a pipelined `INCR` + `EXPIRE NX`. This fixes a latent bug on `main`: a landed `INCR` whose `EXPIRE` failed left a counter with no expiry, so the window never reset and the caller was throttled *permanently* once past the ceiling — on the success path, long after Valkey recovered.
5. **Valkey fallback logs are throttled.** First occurrence of a message at `WARN`, repeats within 30s at `DEBUG`, window reopens so a sustained outage keeps a heartbeat rather than going silent.

## Design decisions worth a reviewer's attention

**`ContextTimeoutEnabled: true`** is the load-bearing field. Without it the whole profile mechanism is decorative.

**Timeout values are code constants, not environment variables** — a departure from the repo's env-config rule (CLAUDE.md §6). Exposing them adds surface nobody will set correctly under incident pressure, and a wrong value silently reintroduces the trap. Flagged because a reviewer should weigh in.

**Two nearby `Warn`s were deliberately left alone.** `roomsubcache`'s size-cap breach is a possible-abuse signal with no error attached, and `search-service`'s empty-index warning is not a Valkey fallback at all. Throttling either would hide something worth seeing.

**No circuit breaker here.** `pkg/circuitbreaker` exists and is the repo standard, but it is not wired into `valkeyutil` and this PR does not change that — the tightened budget alone takes a blackholed read from seconds to ~150ms. Wiring it is a reasonable follow-up, and should reuse that package rather than hand-roll one.

## Verification

- `go test -race ./...` — **144 packages, 0 failures**.
- `make lint` — 0 issues on every changed package. `gosec` clean.
- Every commit builds and vets independently; the branch is bisectable.
- **The integration tests have not run locally.** Docker is unavailable in the sandbox this was written in, so the two new `IncrEx` TTL tests are compile-checked only (`go vet -tags=integration`). CI is their first real execution. `govulncheck` and `semgrep` are likewise CI-only here.

## Relationship to #238

This supersedes #238, which is being closed. That branch was written against a much older `main` and roughly half of it has since been implemented upstream — `pkg/circuitbreaker` (with the same three-way classifier it had converged on), and `valkeyutil`'s `Config`/`ConnectRaw`/`Instrumented` API, which subsumes its presence work and its raw-client constructor. Rebasing it would have re-introduced a second competing breaker. What survived is what remained genuinely missing on `main`, rewritten against the current API — and it is a much smaller diff, with none of the working-notes documents that made up ~60% of #238.

## Open question

**Timeout constants vs env vars** (see above) is the one call I would like a reviewer to make explicitly.

Two notes, neither actionable in this PR: `gobreaker/v2` is already a `main` dependency, so the sign-off #238 asked for is moot. And `docs/reviews/` currently holds 39 readiness reports committed on `main`; CLAUDE.md §5 says those should be deleted before a PR, but they are not this branch's and removing them here would be an unrelated 39-file change.

https://claude.ai/code/session_01GtKXocLa5CCVCJdasafSK2

## Summary by CodeRabbit

- **Reliability**
  - Bot requests continue when Valkey is temporarily unavailable, while duplicate in-flight requests remain protected.
  - Cache-backed features now degrade gracefully during Valkey outages instead of blocking requests.

- **Observability**
  - Added telemetry for requests admitted without Valkey-backed controls.
  - Degraded-service warnings are throttled while retaining diagnostic context.

- **Configuration**
  - Valkey connections now use purpose-specific timeout and retry profiles.
  - Startup reachability checks can be required for tools that need fail-fast behavior.

- **Documentation**
  - Updated health-probe and metrics documentation to describe outage behavior and monitoring signals.


julianshen:
**4 findings · 3 unique issues · 3 inline comments posted** — 🟠 4 warning

### Related work

- [PR #238](https://github.com/hmchangw/newchat/pull/238) — Valkey timeout and startup hardening (closed)

### Local review memory

- `pkg/valkeyutil/observability.go:66` — @coderabbitai: _🩺 Stability & Availability_ _🟠 Major_ _⚡ Quick win_ **Pass StoreProfile to the user-presence-service ConnectRaw call** ConnectRaw defaults to CacheProfile , so presence operations use 150 ms read/write timeouts and one retry.… ([thread](https://github.com/hmchangw/newchat/pull/486#discussion_r3979193178))
- `pkg/valkeyutil/profile.go:69` — @coderabbitai: _🩺 Stability & Availability_ _🟠 Major_ _⚡ Quick win_ &lt;details> &lt;summary>🔎 Supported by static analysis&lt;/summary> 🏁 Script executed: shell #!/bin/bash set -euo pipefail # Confirm the repository version and profile mapping. rg -n… ([thread](https://github.com/hmchangw/newchat/pull/486#discussion_r3979193192))
- `pkg/valkeyutil/valkey.go:80` — @coderabbitai: _🩺 Stability & Availability_ _🟡 Minor_ _⚡ Quick win_ **Make positive-TTL IncrEx atomic.** clusterClient.IncrEx uses a non-transactional pipeline. INCR can succeed while EXPIRE NX fails, leaving botrl:caller:* or botrl:glob… ([thread](https://github.com/hmchangw/newchat/pull/486#discussion_r3979193199))
- `botplatform-service/metrics.go:49` — @coderabbitai: _🩺 Stability & Availability_ _🟠 Major_ _⚡ Quick win_ **Throttle the fail-open warning.** bypassControl runs for every admitted request while Valkey is unavailable. Lines 48-49 emit a direct WARN for each request. This conflicts wit… ([thread](https://github.com/hmchangw/newchat/pull/486#discussion_r3979808234))

<details>
<summary>📒 Files reviewed (26)</summary>

* `botplatform-service/metrics.go`
* `botplatform-service/middleware_idempotency_test.go`
* `botplatform-service/middleware_ratelimit_test.go`
* `docs/health-probes.md`
* `pkg/roomsubcache/lookup.go`
* `pkg/valkeyutil/config_test.go`
* `pkg/valkeyutil/degraded.go`
* `pkg/valkeyutil/degraded_test.go`
* `pkg/valkeyutil/integration_test.go`
* `pkg/valkeyutil/observability.go`
* `pkg/valkeyutil/observability_integration_test.go`
* `pkg/valkeyutil/observability_test.go`
* `pkg/valkeyutil/profile.go`
* `pkg/valkeyutil/profile_test.go`
* `pkg/valkeyutil/readthrough.go`
* `pkg/valkeyutil/tier.go`
* `pkg/valkeyutil/valkey.go`
* `search-service/handler.go`
* `user-presence-service/main_valkey_test.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

pkg/valkeyutil/profile.go
```diff
@@ -0,0 +1,98 @@
...
+	CacheProfile = Profile{
+		DialTimeout:  time.Second,
+		ReadTimeout:  150 * time.Millisecond,
+		WriteTimeout: 150 * time.Millisecond,
+		PoolTimeout:  250 * time.Millisecond,
+		MaxRetries:   1,
+	}
+
+	// StoreProfile serves user-presence-service, where Valkey is the store of
+	// record and no fallback exists. A cache-tight ceiling on a Lua EVAL under
```
julianshen, line 49: _🎯 correctness_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`, `performance-and-concurrency` (2 rules)_

**Automatic retries can replay non-idempotent mutations.**

With MaxRetries set to 1 or 2, a read timeout can retry a command whose response was lost after the server executed it. For IncrEx, one request can increment the counter multiple times and then return success, prematurely throttling users; SetNX and Lua writes have similar ambiguous outcomes. This is reachable during the network degradation these profiles target, and cluster-level retries can compound it. Keep automatic retries disabled for mixed read/write clients, use a separate retrying read client, or make retried mutations idempotent using operation IDs.

<details>

- **`distributed-system`**:45: Both profiles change `MaxRetries` from the previous no-retry behavior to a positive value. If Valkey executes an `IncrEx` pipeline but its response is lost or exceeds `ReadTimeout`, go-redis can resend it, incrementing the counter twice for one request. The inflated count can then throttle users early. Disable transparent retries for non-idempotent writes, use separate read/write clients, or attach operation IDs and deduplicate mutations server-side.

</details>

<details>

```
In `pkg/valkeyutil/profile.go` around line 49: With MaxRetries set to 1 or 2, a read timeout can retry a command whose response was lost after the server executed it. For IncrEx, one request can increment the counter multiple times and then return success, prematurely throttling users; SetNX and Lua writes have similar ambiguous outcomes. This is reachable during the network degradation these profiles target, and cluster-level retries can compound it. Keep automatic retries disabled for mixed read/write clients, use a separate retrying read client, or make retried mutations idempotent using operation IDs.

```

</details>

---

pkg/valkeyutil/valkey.go
```diff
@@ -239,17 +251,27 @@ func (r *clusterClient) SetNX(ctx context.Context, key, value string, ttl time.D
...
+			return 0, fmt.Errorf("valkey incr: %w", err)
 		}
+		return n, nil
+	}
+	// EXPIRE NX on every increment, not a bare EXPIRE on the 0->1 transition: if a
+	// prior call's INCR landed and its EXPIRE did not, the counter is left with no
+	// expiry, never resets, and permanently throttles the caller once it passes the
+	// ceiling — long after Valkey recovered. NX repairs that on the next increment
+	// while leaving a live window alone, so this stays a fixed window rather than
+	// becoming a sliding one. Pipelined to keep the pair at one round trip; both
```
julianshen, line 266: _🏷️ distributed-atomicity_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`, `performance-and-concurrency`, `tgd-review` (3 rules)_

**The increment and expiry can still partially succeed.**

A pipeline is not a Redis transaction. A node can execute `INCR`, then fail or be promoted before `EXPIRE NX` executes; `Exec` returns an error while the increment remains without expiry. If the next request arrives after the intended window, it increments the stale count and only then installs a fresh TTL, so prior-window traffic can reject valid work for another window. Use one Lua script or a same-slot transaction to execute the increment and conditional expiry atomically, while retaining repair for pre-existing no-TTL keys.

<details>

- **`performance-and-concurrency`**:267: If failover or connection loss occurs after INCR executes but before EXPIRE NX, Exec returns an error while the key remains persistent. A later successful call repairs the TTL, but starts a fresh full window from an elevated count, so active callers can be rejected beyond the intended window; an inactive key persists until its next use. Pipelining saves a round trip but provides no atomicity. Execute both commands through MULTI/EXEC or a single server-side script.

<details>

```text
	var incr *redis.IntCmd
	_, err := r.c.TxPipelined(ctx, func(pipe redis.Pipeliner) error {
		incr = pipe.Incr(ctx, key)
		pipe.ExpireNX(ctx, key, ttl)
		return nil
	})
	if err != nil {
		return 0, fmt.Errorf("valkey incr: %w", err)
	}
	return incr.Val(), nil
```

</details>

- **`tgd-review`**:270: Pipeline batches network I/O but does not make INCR and EXPIRE NX atomic. A partial write or disconnect can let INCR execute while Exec returns an error before EXPIRE reaches Valkey. The next request then increments the stale count and only starts a fresh TTL, so a caller can be rejected for an additional full window. Execute both commands in one same-slot Lua script or transaction, and test this failure boundary.

</details>

<details>

```
In `pkg/valkeyutil/valkey.go` around line 266: A pipeline is not a Redis transaction. A node can execute `INCR`, then fail or be promoted before `EXPIRE NX` executes; `Exec` returns an error while the increment remains without expiry. If the next request arrives after the intended window, it increments the stale count and only then installs a fresh TTL, so prior-window traffic can reject valid work for another window. Use one Lua script or a same-slot transaction to execute the increment and conditional expiry atomically, while retaining repair for pre-existing no-TTL keys.

```

</details>

---

pkg/valkeyutil/profile.go
```diff
@@ -0,0 +1,98 @@
...
+		WriteTimeout: p.WriteTimeout,
+		PoolTimeout:  p.PoolTimeout,
+		MaxRetries:   p.MaxRetries,
+		// Without this go-redis ignores the caller's context deadline for socket
+		// reads and only ReadTimeout applies, which makes the rest decorative.
+		ContextTimeoutEnabled: true,
+	}
+}
+
+// WithProfile selects the timeout budget for the client. Defaults to
```
julianshen, line 76: _⚡ performance_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`, `performance-and-concurrency` (2 rules)_

**Cluster redirects multiply the configured timeout budget.**

ClusterOptionsFor leaves MaxRedirects at go-redis's default of three. A retryable timeout can therefore cause four cluster attempts, each with up to MaxRetries+1 node attempts. Without a shorter caller deadline, CacheProfile can take about 1.2 seconds and StoreProfile about 6 seconds, excluding backoff and pool waits; the test's under-two-second calculation omits this outer factor. During a blackhole this both stalls hot paths and multiplies load. Add an explicit cluster-attempt budget and include redirects, retries, and backoff in the bound test.

<details>

- **`distributed-system`**:80: `WithProfile` is newly introduced, but this diff uses it only in tests; no production connection selects `StoreProfile`. The documented user-presence store therefore inherits `CacheProfile`, imposing 150 ms read/write timeouts despite having no fallback. A presence Lua operation exceeding that budget returns an error—and may be retried ambiguously—instead of receiving the intended 500 ms budget. Pass `WithProfile(StoreProfile)` to the user-presence `ConnectRaw` call.

</details>

<details>

```
In `pkg/valkeyutil/profile.go` around line 76: ClusterOptionsFor leaves MaxRedirects at go-redis's default of three. A retryable timeout can therefore cause four cluster attempts, each with up to MaxRetries+1 node attempts. Without a shorter caller deadline, CacheProfile can take about 1.2 seconds and StoreProfile about 6 seconds, excluding backoff and pool waits; the test's under-two-second calculation omits this outer factor. During a blackhole this both stalls hot paths and multiplies load. Add an explicit cluster-attempt budget and include redirects, retries, and backoff in the bound test.

```

</details>

---

pkg/valkeyutil/valkey.go
```diff
@@ -103,37 +105,47 @@ func ConnectCluster(ctx context.Context, addrs []string, password string, opts .
...
+	case pingErr == nil:
+		slog.Info("connected to Valkey cluster", "addrs", addrs)
+	case cc.requireReachable:
 		// Close the half-constructed client on the ping-failure path so unreachable addrs don't leak internal go-redis pool state.
 		if closeErr := c.Close(); closeErr != nil {
 			slog.Warn("valkey cluster close after failed connect", "error", closeErr)
 		}
-		return nil, fmt.Errorf("valkey cluster connect: %w", err)
+		return nil, fmt.Errorf("valkey cluster connect: %w", pingErr)
+	default:
```
julianshen, line 139: _🏷️ error-handling_ | _🟠 Warning_ | _⚡ Quick fix_ | _`silent-failure`_

**Permanent probe failures are converted into successful connections**

The default branch treats every PING error as temporary unreachability. A wrong password, malformed address, or caller-canceled startup context therefore logs a warning and returns a client successfully, although later commands will continue failing or startup was explicitly canceled. Only suppress transient reachability failures; propagate `ctx.Err()` and permanent authentication/configuration errors, closing the client before returning them.

<details>

```
In `pkg/valkeyutil/valkey.go` around line 139: The default branch treats every PING error as temporary unreachability. A wrong password, malformed address, or caller-canceled startup context therefore logs a warning and returns a client successfully, although later commands will continue failing or startup was explicitly canceled. Only suppress transient reachability failures; propagate `ctx.Err()` and permanent authentication/configuration errors, closing the client before returning them.

```

</details>

---

pkg/valkeyutil/valkey.go
```diff
@@ -103,37 +105,47 @@ func ConnectCluster(ctx context.Context, addrs []string, password string, opts .
...
+	// unreachability is non-fatal by default. Note the effective deadline is
+	// min(pingTimeout, Profile.ReadTimeout): go-redis takes the earlier of the
+	// two, so the profile governs in practice and this is only a ceiling.
+	pingCtx, cancel := context.WithTimeout(ctx, pingTimeout)
 	defer cancel()
-	if err := c.Ping(pingCtx).Err(); err != nil {
+	switch pingErr := c.Ping(pingCtx).Err(); {
+	case pingErr == nil:
+		slog.Info("connected to Valkey cluster", "addrs", addrs)
+	case cc.requireReachable:
```
julianshen, line 133: _🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**Caller cancellation is incorrectly treated as nonfatal.**

The default branch treats every PING error as a reachability failure. If the parent context is already canceled or expires during PING, Ping returns that cancellation, but ConnectCluster logs it and returns a successful client, allowing startup to continue after its caller aborted. Only datastore failures should be diagnostic; route parent-context cancellation through the existing close-and-error path and add a canceled-context test.

<details>

> ‼️ **IMPORTANT**

```suggestion
	case cc.requireReachable || ctx.Err() != nil:
```

</details>

<details>

```
In `pkg/valkeyutil/valkey.go` around line 133: The default branch treats every PING error as a reachability failure. If the parent context is already canceled or expires during PING, Ping returns that cancellation, but ConnectCluster logs it and returns a successful client, allowing startup to continue after its caller aborted. Only datastore failures should be diagnostic; route parent-context cancellation through the existing close-and-error path and add a canceled-context test.

```

</details>

---

julianshen (commented):
### tGDBot review of `c011ef2` — 9 findings in 10 files

1 🔴 blocking · 8 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/486#issuecomment-5619680058), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

pkg/valkeyutil/profile.go
```diff
@@ -0,0 +1,98 @@
...
+	// CacheProfile serves consumers that have a backing store to fall through to
+	// (room meta, room subscriptions, search restricted-rooms, the session L2).
+	// Failing fast matters more than succeeding slowly — the fallback is always
+	// there, so a slow success is strictly worse than a quick miss.
+	CacheProfile = Profile{
+		DialTimeout:  time.Second,
+		ReadTimeout:  150 * time.Millisecond,
+		WriteTimeout: 150 * time.Millisecond,
+		PoolTimeout:  250 * time.Millisecond,
+		MaxRetries:   1,
```
julianshen, line 45: _🎯 correctness_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`distributed-system`, `tgd-review` (2 rules)_

**Automatic retries can replay non-idempotent mutations.**

Positive MaxRetries applies to mixed read/write clients. If Valkey executes IncrEx but the response is lost, go-redis can resend it and count one request twice; the middleware can then reject traffic below the configured ceiling. SetNX has the same ambiguity: a retry after a successful first SET can return false and reject the request as already in flight. Disable transparent retries for mutation-capable clients or separate retrying reads from writes made idempotent with operation IDs.

<details>

- **`distributed-system`**:44: Both profiles set positive MaxRetries. If an IncrEx pipeline commits but its reply is lost, go-redis can resend it, incrementing twice for one request. The inflated count can make botRateLimit return 429 before the real limit is reached; Lua mutations have the same ambiguous-outcome risk. Keep transparent retries disabled on mixed read/write clients, split retrying reads from writes, or deduplicate mutations with operation IDs, including cluster-level retries.

</details>

<details>

```
In `pkg/valkeyutil/profile.go` around line 45: Positive MaxRetries applies to mixed read/write clients. If Valkey executes IncrEx but the response is lost, go-redis can resend it and count one request twice; the middleware can then reject traffic below the configured ceiling. SetNX has the same ambiguity: a retry after a successful first SET can return false and reject the request as already in flight. Disable transparent retries for mutation-capable clients or separate retrying reads from writes made idempotent with operation IDs.

```

</details>

---

botplatform-service/middleware.go
```diff
@@ -133,8 +138,8 @@ func botRateLimit(client incrExClient, perCaller, perGlobal int) gin.HandlerFunc
 		if perCaller > 0 {
 			n, err := client.IncrEx(ctx, "botrl:caller:"+pr.UserID, window)
 			if err != nil {
-				errhttp.Write(ctx, c, errcode.Internal("bot rate limit caller", errcode.WithCause(err)))
-				c.Abort()
+				bypassControl(ctx, "rate limit", controlRateLimitCaller, err, "counter", "caller")
```
julianshen, line 141: _⚡ memory-exhaustion_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`oom`_

**Valkey failures remove all bot admission backpressure.**

When either `IncrEx` call fails, lines 141/156 admit every request and bypass both configured ceilings. During a Valkey outage, a looping or malicious caller can therefore drive unbounded concurrent downstream handler work; in-flight goroutines, request data, and dependency work scale with offered load until GC thrashing, container eviction, or process OOM. Keep availability without making admission unlimited by adding a process-local bounded fallback limiter/concurrency semaphore, and reject only after that emergency bound is exhausted.

<details>

```
In `botplatform-service/middleware.go` around line 141: When either `IncrEx` call fails, lines 141/156 admit every request and bypass both configured ceilings. During a Valkey outage, a looping or malicious caller can therefore drive unbounded concurrent downstream handler work; in-flight goroutines, request data, and dependency work scale with offered load until GC thrashing, container eviction, or process OOM. Keep availability without making admission unlimited by adding a process-local bounded fallback limiter/concurrency semaphore, and reject only after that emergency bound is exhausted.

```

</details>

---

pkg/valkeyutil/profile.go
```diff
@@ -0,0 +1,98 @@
...
+		// reads and only ReadTimeout applies, which makes the rest decorative.
+		ContextTimeoutEnabled: true,
+	}
+}
+
+// WithProfile selects the timeout budget for the client. Defaults to
+// CacheProfile, which is the right answer for every consumer that has a
+// fallback; user-presence-service passes StoreProfile because Valkey is its
+// store of record rather than a cache.
+func WithProfile(p Profile) Option {
```
julianshen, line 80: _🎯 correctness_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`tgd-review`_

**The presence store never selects StoreProfile.**

Every connection defaults to CacheProfile, but this diff does not update the user-presence-service connection to pass WithProfile(StoreProfile). Presence therefore retains the 150 ms read/write budget rather than the documented 500 ms budget. A Lua operation or BatchGet taking 150–500 ms will fail even though presence has no fallback, producing user-visible presence errors. Pass valkeyutil.WithProfile(valkeyutil.StoreProfile) to that service's ConnectRaw call.

<details>

```
In `pkg/valkeyutil/profile.go` around line 80: Every connection defaults to CacheProfile, but this diff does not update the user-presence-service connection to pass WithProfile(StoreProfile). Presence therefore retains the 150 ms read/write budget rather than the documented 500 ms budget. A Lua operation or BatchGet taking 150–500 ms will fail even though presence has no fallback, producing user-visible presence errors. Pass valkeyutil.WithProfile(valkeyutil.StoreProfile) to that service's ConnectRaw call.

```

</details>

---

pkg/valkeyutil/profile.go
```diff
@@ -0,0 +1,98 @@
...
+
+// ClusterOptionsFor builds the go-redis cluster options for a profile.
+func ClusterOptionsFor(addrs []string, password string, p Profile) *redis.ClusterOptions {
+	return &redis.ClusterOptions{
+		Addrs:        addrs,
+		Password:     password,
+		DialTimeout:  p.DialTimeout,
+		ReadTimeout:  p.ReadTimeout,
+		WriteTimeout: p.WriteTimeout,
+		PoolTimeout:  p.PoolTimeout,
```
julianshen, line 68: _🏷️ timeout-budget_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`_

**Cluster redirects multiply the configured timeout budget.**

ClusterOptionsFor leaves MaxRedirects at its default of three, while the tests bound only node retries. Without an earlier caller deadline, one retryable cluster failure can make four cluster attempts, each with MaxRetries+1 node attempts: StoreProfile can spend about six seconds on reads before backoff and pool waits. During a blackhole this defeats the intended 500ms profile and multiplies load. Configure cluster attempts explicitly and test the total bound including redirects, retries, backoff, and pool acquisition.

<details>

```
In `pkg/valkeyutil/profile.go` around line 68: ClusterOptionsFor leaves MaxRedirects at its default of three, while the tests bound only node retries. Without an earlier caller deadline, one retryable cluster failure can make four cluster attempts, each with MaxRetries+1 node attempts: StoreProfile can spend about six seconds on reads before backoff and pool waits. During a blackhole this defeats the intended 500ms profile and multiplies load. Configure cluster attempts explicitly and test the total bound including redirects, retries, backoff, and pool acquisition.

```

</details>

---

botplatform-service/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+	controlRateLimitCaller = metric.WithAttributes(attribute.String("control", "rate_limit_caller"))
+	controlRateLimitGlobal = metric.WithAttributes(attribute.String("control", "rate_limit_global"))
+	controlIdempotency     = metric.WithAttributes(attribute.String("control", "idempotency"))
+)
+
+// bypassControl records one fail-open admission. The log carries the cause for
+// diagnosis; the counter is what to alert on, since it survives log sampling
+// and does not depend on log level. Shared by every control so their bypass
+// telemetry cannot drift apart.
+func bypassControl(ctx context.Context, control string, attrs metric.MeasurementOption, err error, logArgs ...any) {
```
julianshen, line 47: _⚡ performance_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`, `performance-and-concurrency` (2 rules)_

**Fail-open admissions can create an unbounded warning-log storm**

During a Valkey outage, every failed rate-limit or idempotency call reaches this synchronous WARN; a request traversing both controls can emit multiple records. Unlike the newly throttled cache-fallback logs, this path has no rate bound, so sustained bot traffic can saturate log output and add latency while the dependency is already degraded. Route these messages through `valkeyutil.LogDegraded` (keyed by control) or another bounded logger, while continuing to increment the counter for every bypass.

<details>

- **`distributed-system`**:48: When Valkey fails, the rate limiter deliberately admits a looping caller, and every bypass synchronously emits a WARN before continuing. The disabled ceiling therefore permits arbitrary warning volume, potentially compounded by idempotency failures; a slow or full logging sink can block handlers and turn the fail-open path back into an outage. Throttle or sample warnings per control, using the existing degraded-log pattern, while continuing to increment the metric for every bypass.

</details>

<details>

```
In `botplatform-service/metrics.go` around line 47: During a Valkey outage, every failed rate-limit or idempotency call reaches this synchronous WARN; a request traversing both controls can emit multiple records. Unlike the newly throttled cache-fallback logs, this path has no rate bound, so sustained bot traffic can saturate log output and add latency while the dependency is already degraded. Route these messages through `valkeyutil.LogDegraded` (keyed by control) or another bounded logger, while continuing to increment the counter for every bypass.

```

</details>

---

julianshen (commented):
### tGDBot review of `f40dd2b` — 9 findings in 24 files

4 🔴 blocking · 5 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/486#issuecomment-5619680058), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

botplatform-service/middleware_idempotency.go
```diff
@@ -70,12 +70,16 @@ func botIdempotency(
 		key := "idem:" + opID
 
 		acquired, err := client.SetNX(ctx, key, "processing", sentinelTTL)
-		if err != nil {
-			errhttp.Write(ctx, c, errcode.Internal("bot idempotency: acquire", errcode.WithCause(err)))
-			c.Abort()
+		switch {
+		case err != nil:
```
julianshen, line 74: _🏷️ error-handling_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`silent-failure`_

**SetNX errors can leave a hidden sentinel after a successful request**

A transport error does not prove SetNX was not applied: Valkey may store the sentinel and lose the response. This branch then runs the handler and returns before normal release, so a successful response can leave the sentinel blocking retries until its TTL. The fake test only models an error with no write. Preserve fail-open behavior by using a unique acquisition token and conditionally deleting that token after a non-5xx downstream result.

<details>

```
In `botplatform-service/middleware_idempotency.go` around line 74: A transport error does not prove SetNX was not applied: Valkey may store the sentinel and lose the response. This branch then runs the handler and returns before normal release, so a successful response can leave the sentinel blocking retries until its TTL. The fake test only models an error with no write. Preserve fail-open behavior by using a unique acquisition token and conditionally deleting that token after a non-5xx downstream result.

```

</details>

---

pkg/valkeyutil/valkey.go
```diff
@@ -239,17 +251,27 @@ func (r *clusterClient) SetNX(ctx context.Context, key, value string, ttl time.D
...
+	// becoming a sliding one. Pipelined to keep the pair at one round trip; both
+	// commands share a key, so cluster routing keeps them on one node.
+	pipe := r.c.Pipeline()
+	incr := pipe.Incr(ctx, key)
+	pipe.ExpireNX(ctx, key, ttl)
+	if _, err := pipe.Exec(ctx); err != nil {
+		return 0, fmt.Errorf("valkey incr: %w", err)
 	}
-	return n, nil
+	return incr.Val(), nil
```
julianshen, line 274: _🏷️ distributed-atomicity_ | _🟠 Warning_ | _⚡ Quick fix_ | _`silent-failure`_

**The increment and expiry can still partially succeed**

Pipeline does not make INCR and EXPIRE NX atomic. A disconnect or failover after INCR executes can make Exec return an error while leaving a persistent counter. The next successful request increments that stale count and starts a new full TTL, potentially rejecting valid traffic for another window. Execute both commands in one same-slot transaction or Lua script.

<details>

> ‼️ **IMPORTANT**

```suggestion
	var incr *redis.IntCmd
	_, err := r.c.TxPipelined(ctx, func(pipe redis.Pipeliner) error {
		incr = pipe.Incr(ctx, key)
		pipe.ExpireNX(ctx, key, ttl)
		return nil
	})
	if err != nil {
		return 0, fmt.Errorf("valkey incr: %w", err)
	}
	return incr.Val(), nil
```

</details>

<details>

```
In `pkg/valkeyutil/valkey.go` around line 268: Pipeline does not make INCR and EXPIRE NX atomic. A disconnect or failover after INCR executes can make Exec return an error while leaving a persistent counter. The next successful request increments that stale count and starts a new full TTL, potentially rejecting valid traffic for another window. Execute both commands in one same-slot transaction or Lua script.

```

</details>

---

botplatform-service/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+	slog.WarnContext(ctx, "bot "+control+" unavailable; admitting request",
```
julianshen, line 48: _🏷️ failure amplification_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`, `tgd-review` (2 rules)_

**Fail-open traffic emits an unbounded WARN stream**

Every failed control call synchronously emits a WARN before admitting the request. Because the same outage disables rate limiting, one looping caller can generate warning records without a rate bound; a slow or full logging sink can then add latency or block handlers and turn fail-open into another outage. Route the warning through LogDegraded or an equivalent per-control throttle while continuing to increment controlBypassed for every admission.

<details>

- **`tgd-review`**:48: Every admitted request whose Valkey control fails emits a synchronous WARN here. Because the same outage deliberately removes request ceilings, log volume can scale with offered traffic and add latency or saturate the logging sink. Route this message through the new degraded-log throttle, keyed by control, while retaining the counter increment for every bypass.

</details>

<details>

```
In `botplatform-service/metrics.go` around line 48: Every failed control call synchronously emits a WARN before admitting the request. Because the same outage disables rate limiting, one looping caller can generate warning records without a rate bound; a slow or full logging sink can then add latency or block handlers and turn fail-open into another outage. Route the warning through LogDegraded or an equivalent per-control throttle while continuing to increment controlBypassed for every admission.

```

</details>

---

pkg/valkeyutil/profile.go
```diff
@@ -0,0 +1,98 @@
...
+		MaxRetries:   p.MaxRetries,
+		// Without this go-redis ignores the caller's context deadline for socket
```
julianshen, line 70: _⚡ performance_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**Cluster redirects multiply the intended timeout budget.**

ClusterOptionsFor leaves MaxRedirects at go-redis's default of three. Without an earlier caller deadline, StoreProfile can therefore make four cluster attempts, each with up to three node attempts, spending about six seconds on reads before backoff or pool waits. This contradicts the documented `(MaxRetries+1) × ReadTimeout` bound. Configure the cluster-attempt budget explicitly and test the complete operation bound.

<details>

```
In `pkg/valkeyutil/profile.go` around line 70: ClusterOptionsFor leaves MaxRedirects at go-redis's default of three. Without an earlier caller deadline, StoreProfile can therefore make four cluster attempts, each with up to three node attempts, spending about six seconds on reads before backoff or pool waits. This contradicts the documented `(MaxRetries+1) × ReadTimeout` bound. Configure the cluster-attempt budget explicitly and test the complete operation bound.

```

</details>

---

pkg/valkeyutil/valkey.go
```diff
@@ -239,17 +251,27 @@ func (r *clusterClient) SetNX(ctx context.Context, key, value string, ttl time.D
...
+	// ceiling — long after Valkey recovered. NX repairs that on the next increment
+	// while leaving a live window alone, so this stays a fixed window rather than
+	// becoming a sliding one. Pipelined to keep the pair at one round trip; both
+	// commands share a key, so cluster routing keeps them on one node.
+	pipe := r.c.Pipeline()
+	incr := pipe.Incr(ctx, key)
+	pipe.ExpireNX(ctx, key, ttl)
+	if _, err := pipe.Exec(ctx); err != nil {
+		return 0, fmt.Errorf("valkey incr: %w", err)
 	}
```
julianshen, line 273: _🏷️ distributed-atomicity_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**The increment and expiry can still partially succeed.**

Pipeline batches commands but is not atomic. A disconnect or failover after INCR executes can leave a persistent counter without executing EXPIRE NX. The next successful request increments the stale count before installing a fresh full TTL, potentially rejecting valid traffic for another window. Execute both commands in one same-slot transaction or Lua script.

<details>

> ‼️ **IMPORTANT**

```suggestion
	// Execute the pair in one transaction so INCR cannot land without EXPIRE NX.
	// Running EXPIRE NX on every increment also repairs a missing expiry while
	// leaving a live fixed window unchanged.
	var incr *redis.IntCmd
	_, err := r.c.TxPipelined(ctx, func(pipe redis.Pipeliner) error {
		incr = pipe.Incr(ctx, key)
		pipe.ExpireNX(ctx, key, ttl)
		return nil
	})
	if err != nil {
		return 0, fmt.Errorf("valkey incr: %w", err)
	}
	return incr.Val(), nil
```

</details>

<details>

```
In `pkg/valkeyutil/valkey.go` around line 259: Pipeline batches commands but is not atomic. A disconnect or failover after INCR executes can leave a persistent counter without executing EXPIRE NX. The next successful request increments the stale count before installing a fresh full TTL, potentially rejecting valid traffic for another window. Execute both commands in one same-slot transaction or Lua script.

```

</details>

---

julianshen (commented):
### tGDBot review of `1d859dd` — 4 findings in 26 files

4 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/486#issuecomment-5619680058), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

PR #497, opened 2026-09-16
chore(deps): bump grpc to v1.83.1 for GO-2026-6348

flywindy:
## Why

The `sast` gate is failing on **`main`** and therefore on every open PR. `gosec`, `semgrep` and the repo-owned semgrep rule tests all pass; only `govulncheck` is red:

```
Vulnerability #1: GO-2026-6348
    Heap Memory Exhaustion (OOM) via HTTP/2 DATA Frame Fragmentation in
    google.golang.org/grpc
  More info: https://pkg.go.dev/vuln/GO-2026-6348
  Module: google.golang.org/grpc
    Found in: google.golang.org/grpc@v1.82.1
    Fixed in: google.golang.org/grpc@v1.83.1
    Example traces found:
      #1: pkg/natsmetrics/metrics.go:478:19: natsmetrics.Message.MarkTerminal
          calls sync.Once.Do, which eventually calls transport.NewHTTP2Client
      ...
```

Last seen on [`main` @ `3e8eda2`](https://github.com/hmchangw/newchat/actions/runs/35046729798) (the run for #495) and on every branch cut from it.

## Where grpc comes from

It is an indirect dependency, reached through the o11y SDK's OTLP log exporter — `go mod why`:

```
pkg/natsmetrics
  -> github.com/flywindy/o11y
  -> github.com/flywindy/o11y/internal/log
  -> go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploghttp
  -> go.opentelemetry.io/proto/otlp/collector/logs/v1
  -> google.golang.org/grpc
```

The OTLP proto package links grpc even on the HTTP exporter path, so the reachable call graph runs through `natsmetrics.Message.MarkTerminal`.

`github.com/flywindy/o11y v0.12.0` is already the newest published release, so there is no upstream bump that would carry a fixed grpc transitively. Requiring the fixed version directly is the remedy; minimum version selection makes the explicit requirement win over the lower version the graph asks for.

## The change

```diff
-	google.golang.org/grpc v1.82.1 // indirect
+	google.golang.org/grpc v1.83.1 // indirect
```

Two files, three lines including `go.sum`. Nothing else moved — `go mod tidy` produced no other edits.

v1.83.1 is exactly the version the advisory names as fixed, so this is the minimal change that clears it.

## Verification

| Check | Result |
|---|---|
| `go build ./...` | pass |
| `make lint` | 0 issues |
| `go test -race ./...` (whole repo) | pass |
| `make sast-gosec` | pass |
| repo-owned semgrep rules (`.semgrep/`) | 0 findings |
| `go vet -tags integration ./...` | pass |
| `sast` in CI | **passed on this branch** |

`govulncheck` itself could not be run locally — `vuln.go.dev` is blocked from my sandbox (403 through the outbound proxy; `proxy.golang.org` is reachable, which is why the module fetch works). The `sast` job on this PR is the confirmation, and it passed.

---

### Scope note

**This PR contains the grpc bump only.** The MinIO / Docker Hub problem discussed in the comments below is a separate issue and is **not** part of what merged here — `pkg/testutil/testimages` and `docker-local/compose.deps.yaml` are untouched by this change. That work is tracked in #498.

https://claude.ai/code/session_01DpxUYusUx5rxc1HN4ypmKK


flywindy:
## `test-integration (pkg/minioutil)` is failing, and it is not this PR's

**The failing check:** `test-integration (pkg/minioutil)`. Every test in the package failed identically, at container startup, before any test body ran:

```
testutil.MinIO: start minio: run minio: generic container: create container:
Error response from daemon: pull access denied for minio/minio, repository does not
exist or may require 'docker login': denied: requested access to the resource is denied
```

**Why it is not this PR's:** the diff here is one dependency version in `go.mod` plus its two `go.sum` lines. A grpc version cannot cause a Docker Hub image pull to be denied, and no Go code was compiled or run before the failure — the container never started.

The same error is already on `main`. At this PR's base commit [`3e8eda2`](https://github.com/hmchangw/newchat/actions/runs/35046729798), `test-integration (media-service)` failed with the identical message:

```
prewarm shared containers: start minio: run minio: generic container: create container:
Error response from daemon: pull access denied for minio/minio, repository does not
exist or may require 'docker login': denied: requested access to the resource is denied
```

`test-integration (client-update-service)` failed there too. So this is a standing condition in CI that lands on whichever MinIO-dependent job happens to run — media-service and client-update-service on `main`, `pkg/minioutil` here. `minio/minio` is a real public image, so "pull access denied" here is Docker Hub refusing the anonymous pull (rate limiting), not a missing repository.

**No fix exists to port.** This is a registry/runner condition, not code. The durable remedies are outside this PR: authenticate the Docker Hub pull in CI, or mirror the image. Worth its own issue — it will keep flapping across jobs until then.

**Re-run:** this failure died before any test body ran, so it is the one class where a re-run is informative rather than noise. The run is still in flight, and other MinIO-dependent jobs (`upload-service`, `media-service`) may be hit by the same condition, so I am letting it finish and will then spend a single `rerun_failed_jobs` across all of them rather than burning it on one job now.

The check that actually matters for this PR is `sast` — that is the whole point of the change, and it is still running.

---

flywindy:
## Re-run result, and a correction to my diagnosis

I said above that the MinIO failures looked like Docker Hub **rate limiting** and that a re-run had a real chance. Both were wrong. Reporting the re-run as promised, with the evidence that corrects it.

**Re-run:** spent on run [35063072854](https://github.com/hmchangw/newchat/actions/runs/35063072854) (attempt 2). All five MinIO jobs failed again with the byte-identical error. That is my one re-run, now spent.

**What it actually is.** I queried the Docker Hub registry API directly with an anonymous pull token:

| image | anonymous manifest fetch |
|---|---|
| `minio/minio:RELEASE.2025-07-18T21-56-31Z` (what `testimages.MinIO` pins) | **401** |
| `minio/minio:latest` | **401** |
| `minio/mc:latest` | **401** |
| `library/mongo:8.2.9` | 200 |
| `library/nats:2.11-alpine` | 200 |
| `library/cassandra:4.1.3` | 200 |

The whole `minio/` namespace now refuses anonymous pulls, while the official images this repo pins keep working from the same client at the same moment. Rate limiting would return 429 and would not spare `mongo`/`nats`/`cassandra`; this is an **authentication requirement**, and it is persistent, which is why the re-run reproduced it exactly.

That also explains the shape of the failure set: the five failing jobs are precisely the five packages that call `testutil.MinIO` — `pkg/minioutil`, `pkg/poolartifact`, `media-service`, `client-update-service`, `upload-service` — and nothing else in the 61-job matrix.

**Unchanged:** this is still not this PR's failure. The diff is one dependency version; no Go code runs before the container pull fails; and `sast` — the entire point of this PR — **passes**, confirming the bump clears GO-2026-6348. 56 of 61 checks are green.

## Fixes, and what I could not verify

1. **Authenticate the pull** — add `docker/login-action` to the workflow with Docker Hub credentials. There is currently no Docker Hub login anywhere in `.github/workflows/ci.yml`. Smallest change, but **I could not verify it is sufficient**: I have no credentials here, so I cannot tell whether a free authenticated account may pull `minio/minio` or whether access is restricted further.
2. **Mirror the image** into GHCR and repoint `testimages.MinIO`. Definitely works once seeded, and removes the external dependency, but needs one pull from somewhere with access.
3. **Repin to an anonymously pullable image.** For the record I probed alternatives: `bitnami/minio` is now 404, `bitnamilegacy/minio` is 200 but unmaintained. `quay.io` I could not test at all — it is blocked from my sandbox, so its 000 is "unreachable from here", not a statement about the image.

This is CI infrastructure and a credentials decision, so I have not touched it from this PR.

---

hmchangw (approved):
Thanks!

PR #499, opened 2026-09-16
Cross-site failover: SP3 routing override + SP4 failover state (on main)

hmchangw:
## What this is

The cross-site failover ("lifeboat") **SP3 + SP4** work, applied on top of current `main`. This is the clean equivalent of "rebase #256 onto main": the existing failover branches (#249/#256) sit on a **divergent history** that can't be `git rebase`d onto `main`'s squash-merged trunk (different roots), so this branch **reconstructs the delta** directly on `main` and verifies it green.

Supersedes the stacked #249/#256 as the `main`-targeting PR. (SP1's implementation specs are intentionally **not** carried here — that's a separate effort; this branch carries the program design/roadmap + the SP2/3/4/SP6 docs.)

## Code — `portal-service`

- **SP4 — failover state + control surface.** Operator-driven `FailoverState` machine (`healthy → failed_over → failing_back → healthy`, plus `resume`), a Mongo CAS store where **portal is the sole writer** (the split-brain fence), and an **ops-token-gated control surface** (`GET`/`GET`/`POST /internal/v1/failover[/:siteId]`) on a **separate internal listener** — off the public discovery API. `FAILOVER_OPS_TOKEN` gates it; `409` on illegal transition / version conflict; `operator`+`reason` audit; token never logged.
- **SP3 — routing override.** A TTL-cached failover reader + `servingURLs` at portal's `resolve()` seam: when a site is `failed_over`, `/api/userInfo` returns the **backup's** `baseUrl`/`natsUrl` while **`siteId` stays home** (the load-bearing correctness point — the backup namespaces data by origin `siteId`). Fail-safe `home` on any read miss/error. Bot/admin `/api/v1/login` left unchanged (deferred).
- Config: `FAILOVER_STATE_TTL`, `PORTAL_BACKUP_SITE_ID` (+ reserved `_backup` registry entry); `codes_portal` reasons; `docker-compose` dev env; a `docs/client-api.md` §2.3 note.

## Docs — `docs/superpowers`

Governing design + roadmap, the **World-1 SP2 identity** spec (one shared NATS account ⇒ no per-site keys; the `chat.local.room.>` grant is already on `main` via SP0), SP3/SP4 specs + implementation plans, and the **SP6 ops runbook**.

## Verification (on `main`)

- `make build SERVICE=portal-service` ✅ · `make test SERVICE=portal-service` ✅ (race) · `make lint` **0 issues** ✅
- Failover logic (state machine, reader, `servingURLs`, handlers) **100% unit-covered**.
- Integration tests (Mongo CAS, reader-over-Mongo, `/api/userInfo`-over-Mongo) **compile under the `integration` tag** — they run in CI (Docker unavailable in the dev sandbox).

## Scope / deferred (unchanged from the SP specs)

- **Bot failover** — deliberately deferred (critical, but gated on a backup botplatform via SP1/SP2); the SP3 hook is one line in `HandleLogin` when that lands.
- **SP2 serving path & SP5 failback** — gated on SP1 (materialization). **SP6** infra items are tagged in the runbook (ready-now vs gated vs external).

https://claude.ai/code/session_01Bx1yBMCjo88ftnTB6ha5pA

---
_Generated by [Claude Code](https://claude.ai/code/session_01Bx1yBMCjo88ftnTB6ha5pA)_

## Summary by CodeRabbit

- **New Features**
  - Added operator-controlled site failover and failback workflows.
  - During failover, client connection URLs automatically switch to the backup site while account identity and home site information remain unchanged.
  - Existing clients reconnect automatically and receive updated routing during failover and recovery.
  - Added safeguards for invalid transitions, unauthorized operations, and conflicting updates.

- **Documentation**
  - Documented failover behavior, recovery procedures, system design, deployment planning, and operational requirements.


hmchangw:
Worked through the review in `53571ec9`. 15 of 16 findings applied; one declined with reasoning below.

### Code fixes

| Finding | Change |
|---|---|
| `failover_handler.go` — whitespace-only audit fields | Trim `operator`/`reason` before validating; persist and log the trimmed values. |
| `failover_handler.go` — unknown `siteId` mints state | `Post` now rejects a site absent from `PORTAL_SITE_URLS` with `404 failover_unknown_site`, before any store call. `Get`/`List` stay permissive (harmless reads, and the reader depends on the synthesized healthy default). |
| `failover_reader.go` — out-of-order cache writes | `cachedTarget` carries the state `version`; a read whose version is older than what is cached neither overwrites it nor answers from itself. |
| `main.go` — non-positive `FAILOVER_STATE_TTL` | Rejected at startup. |
| `main.go` — control-surface serve failure invisible | The serve goroutine flips an atomic; `/readyz` returns 503 while an **enabled** control surface is down. Unwired (and readiness unchanged) when `FAILOVER_OPS_TOKEN` is empty. |
| `docker-compose.yml` — `dev-failover-token` default | Now empty, matching `main.go`'s safe default. |
| SP3 plan — untested healthy-after-TTL flip | Added to `failover_userinfo_integration_test.go`: resume `site-a`, lapse the TTL, assert home coordinates with `siteId` still `site-a`. Uses the reader's injectable clock rather than a sleep (CLAUDE.md §Concurrency). |

### Declined

**`main.go:193` — TLS on the internal listener.** Not taken in this PR. The listener is deliberately off the public `:8085` API and is documented as ops-network-only (SP6 runbook §1.1); there is no TLS-termination pattern anywhere in this repo's Go services — every service terminates TLS at the gateway/mesh, so adding a bespoke cert path here would be a new convention rather than a fix. The `:8090` default binds all interfaces because that is what a container needs; restricting exposure is a NetworkPolicy/mesh concern, which is where the runbook already puts it. Happy to revisit if the platform grows a standard internal-TLS story.

### Documentation corrections

Several of these were genuine factual errors on my side, not wording:

- **Split-brain claim was overstated.** The CAS fences the stored *decision*; what each replica *serves* comes from a TTL cache that a transition does not invalidate, so replicas can disagree for up to `FAILOVER_STATE_TTL`. Runbook §1.5 and SP4 spec §7 now document the bounded overlap — harmless on `failover` (the stale answer names a down site and the client retries), material on `complete` (writes split across home and backup for that window). Cross-replica invalidation is called out as deliberately out of this slice.
- **`complete` is unguarded until SP5.** Kept the transition (removing it would make `failing_back` a dead end) but added an explicit danger notice: nothing verifies convergence today, so completing early silently strands backup-only writes.
- **`MESSAGE_BUCKET_HOURS` ≠ the lifeboat read window.** Correct — the repo default is `360`, not `72`, and conflating a partition width with a retention policy would mis-partition writes. Decoupled in the design §2 and the runbook; origin `history-service`/`message-worker` config is now named as the source of truth.
- **Replication lag** is recorded as new SP1 work; no reusable signal exists (only `oplog_replication_lag_ms` in a migration tool).
- Stale SP2/SP4 roadmap status refreshed; SP3 spec's code sample now carries the `h.failover != nil` guard the real code already had; `client-api.md` describes `baseUrl`/`natsUrl` as the currently-serving site with `siteId` staying home.

### On the `chat.local.room.>` authorization finding

Real, and worth recording — with a correction to its framing. The grant is not membership-scoped *today, within a single site*: message payloads are fanned to per-user `chat.user.{account}.…` subjects, but room-level events go to the room subject and are delivered by interest, so a non-member who knows a room id can already observe them on their own site. That is pre-existing and unchanged by this PR.

What the backup changes is the blast radius: co-locating every site's rooms in one NATS means the site boundary stops containing that exposure. Nothing is exploitable now — the serving path is gated on SP1 and does not exist — but it must close before the backup ever serves traffic. Recorded as a blocking item in the SP2 spec §4.1 and the roadmap, with the two candidate fixes (scope local-room subjects by origin site, or enforce membership at subscribe time). Agreed that the leaf deny is not a mitigation here.

### Verification

`make test SERVICE=portal-service` (race) ✅ · `make build SERVICE=portal-service` ✅ · `make lint` 0 issues ✅ · `gosec` PASS ✅ · failover logic still 100% unit-covered. Integration tests compile under the `integration` tag and run in CI (no Docker in my sandbox). `govulncheck` and `semgrep` could not run locally — proxy-blocked and not installed respectively — so I'm relying on the CI `sast` gate for those.

---

julianshen:
**19 findings · 12 unique issues · 12 inline comments posted** — 🔴 11 blocking · 🟠 8 warning

> [!NOTE]

### Related work

- [PR #256](https://github.com/hmchangw/newchat/pull/256) — Cross-site failover: SP3 — portal health-aware routing override (closed)
- [PR #249](https://github.com/hmchangw/newchat/pull/249) — Cross-site failover: SP2–SP4 design cycles \+ SP4-core failover state \(portal-service\) (closed)

- **Failover reads inherit the stale secondaryPreferred preference.**
  - `portal-service/failover_mongo.go:19` (`mongodb`) — Failover reads inherit the stale secondaryPreferred preference.
  - `portal-service/failover_handler.go:116` (`distributed-system`) — Failover commits even when no backup route is configured.
  - `portal-service/failover_handler.go:114` (`tgd-review`) — Failover succeeds even when no backup target is configured.
  - `portal-service/failover_handler.go:118` (`silent-failure`) — Failover reports success without validating the backup target

### Local review memory

- `portal-service/main.go:209` — @coderabbitai: _🔒 Security & Privacy_ _🛡️ Analyzed with Security Review_ _🟠 Major_ _🏗️ Heavy lift_ &lt;details> &lt;summary>🧩 Analysis chain&lt;/summary> 🏁 Script executed: bash #!/bin/bash set -e printf '%s\n' '--- main.go relevant symbols ---' rg… ([thread](https://github.com/hmchangw/newchat/pull/499#discussion_r4034600854))

<details>

* `docs/client-api.md`
* `docs/superpowers/plans/2026-07-28-cross-site-failover-roadmap.md`
* `docs/superpowers/plans/2026-08-12-sp3-portal-routing-override.md`
* `docs/superpowers/plans/2026-08-12-sp4-failover-state-portal.md`
* `docs/superpowers/plans/2026-08-12-sp6-ops-runbook.md`
* `docs/superpowers/specs/2026-07-28-cross-site-failover-design.md`
* `docs/superpowers/specs/2026-08-11-sp2-backup-identity-jwt-minting.md`
* `docs/superpowers/specs/2026-08-11-sp3-portal-routing-override.md`
* `docs/superpowers/specs/2026-08-11-sp4-failover-trigger-health-detection.md`
* `pkg/errcode/codes_portal.go`
* `portal-service/deploy/docker-compose.yml`
* `portal-service/failover.go`
* `portal-service/failover_handler.go`
* `portal-service/failover_handler_test.go`
* `portal-service/failover_integration_test.go`
* `portal-service/failover_middleware.go`
* `portal-service/failover_middleware_test.go`
* `portal-service/failover_mongo.go`
* `portal-service/failover_reader.go`
* `portal-service/failover_reader_integration_test.go`
* `portal-service/failover_reader_test.go`
* `portal-service/failover_test.go`
* `portal-service/failover_userinfo_integration_test.go`
* `portal-service/handler.go`
* `portal-service/handler_test.go`
* `portal-service/main.go`
* `portal-service/mock_store_test.go`
* `portal-service/store.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

portal-service/failover_reader.go
```diff
@@ -0,0 +1,67 @@
...
+	r.mu.Unlock()
+
+	// Read outside the lock so a slow Mongo call does not serialize all readers.
+	st, err := r.store.Get(ctx, siteID)
+	if err != nil {
+		slog.WarnContext(ctx, "failover reader: store read failed, defaulting to home",
+			"siteId", siteID, "error", err)
+		return ServingHome
+	}
+	target := st.ServingTarget()
```
julianshen, line 55: _🎯 cassandra-correctness_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`cassandra`_

**A transient state-read error can split Cassandra writes**

While a site is `failing_back`, home Cassandra is online but must not serve until replay converges. After this cache expires, any transient Mongo error reports `home`; a reconnecting client is then handed home URLs while other clients continue writing to backup. This creates two active Cassandra write locations and can leave messages absent from either side's view or replay. Preserve the last cached target on refresh errors, defaulting to home only when no state has ever been read.

<details>

> ‼️ **IMPORTANT**

```suggestion
	if err != nil {
		slog.WarnContext(ctx, "failover reader: store read failed, using last-known target",
			"siteId", siteID, "error", err)
		r.mu.Lock()
		defer r.mu.Unlock()
		if c, ok := r.cache[siteID]; ok {
			return c.target
		}
		return ServingHome
	}
```

</details>

<details>

```
In `portal-service/failover_reader.go` around line 51: While a site is `failing_back`, home Cassandra is online but must not serve until replay converges. After this cache expires, any transient Mongo error reports `home`; a reconnecting client is then handed home URLs while other clients continue writing to backup. This creates two active Cassandra write locations and can leave messages absent from either side's view or replay. Preserve the last cached target on refresh errors, defaulting to home only when no state has ever been read.

```

</details>

---

portal-service/failover.go
```diff
@@ -0,0 +1,109 @@
...
+
+// nextStatus returns the status resulting from applying action to cur, or
+// errIllegalTransition if the action is not valid from cur.
+func nextStatus(cur FailoverStatus, action FailoverAction) (FailoverStatus, error) {
+	switch {
+	case cur == StatusHealthy && action == ActionFailover:
+		return StatusFailedOver, nil
+	case cur == StatusFailedOver && action == ActionFailback:
+		return StatusFailingBack, nil
+	case cur == StatusFailedOver && action == ActionResume:
```
julianshen, line 83: _🏷️ failback safety_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`cassandra`, `distributed-system`, `nats`, `tgd-review` (4 rules)_

**complete and resume can flip home before backup writes are safe**

Both terminal transitions are accepted without verifying replay/convergence or that backup accepted no writes. After users have written through backup, an operator can POST `complete` or `resume`; the state becomes healthy and subsequent discovery routes home, where backup-only messages are absent. The added runbook itself identifies this as silent user-visible data loss, while SP5 is deferred. Until a convergence/no-write gate exists, reject these terminal actions rather than relying on an operator warning.

<details>

- **`distributed-system`**:82: `resume` and `complete` immediately make the persisted target `home`. Existing clients remain connected to the still-running backup because changing this document does not disconnect them, while new clients are routed home after cache expiry. The diff also contains no replay/convergence gate, so backup-only writes can be stranded both before and after the transition. Disable both backup-to-home transitions until SP5 can verify convergence and fence or drain backup sessions as part of the cutover.
- **`tgd-review`**:83: After clients have been routed to the backup, either `resume` directly or `complete` after an unverified failback changes the target to home. No code verifies that backup writes were replayed because SP5 is deferred. A client can write on the backup, an operator can invoke either transition, and after the TTL that message is absent from home and effectively lost. Reject terminal transitions until a durable convergence/no-write gate exists.
- **`cassandra`**:29: `complete` and `resume` return routing to home without a Cassandra replay or lag check. After failover, a client can receive backup URLs and get a send acknowledged into backup Cassandra; either action then sends discovery traffic home, where that row is absent. The runbook acknowledges this loss path for `complete`; `resume` has the same risk after even a brief false-alarm failover. Reject these transitions until SP5 verifies convergence, or gate them on a server-verified zero-write or zero-lag condition.

</details>

<details>

```
In `portal-service/failover.go` around line 83: Both terminal transitions are accepted without verifying replay/convergence or that backup accepted no writes. After users have written through backup, an operator can POST `complete` or `resume`; the state becomes healthy and subsequent discovery routes home, where backup-only messages are absent. The added runbook itself identifies this as silent user-visible data loss, while SP5 is deferred. Until a convergence/no-write gate exists, reject these terminal actions rather than relying on an operator warning.

```

</details>

---

portal-service/failover_reader.go
```diff
@@ -0,0 +1,67 @@
...
+	r.mu.Lock()
+	if c, ok := r.cache[siteID]; ok && r.now().Before(c.expires) {
+		r.mu.Unlock()
+		return c.target
+	}
+	r.mu.Unlock()
+
+	// Read outside the lock so a slow Mongo call does not serialize all readers.
+	st, err := r.store.Get(ctx, siteID)
+	if err != nil {
```
julianshen, line 50: _🏷️ distributed-consistency_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`_

**A Mongo outage can reactivate a fenced home site.**

After this replica caches `backup`, let the TTL expire while Mongo is unavailable. `Get` then fails and the reader returns `home`, discarding its last known authoritative target. If the home site is reachable but has not been reconciled, this replica sends clients home while other replicas continue serving the backup, splitting writes across both sites. Preserve the last successfully read target on errors; when no prior value exists, propagate an unavailable result rather than inventing `home`.

<details>

```
In `portal-service/failover_reader.go` around line 50: After this replica caches `backup`, let the TTL expire while Mongo is unavailable. `Get` then fails and the reader returns `home`, discarding its last known authoritative target. If the home site is reachable but has not been reconciled, this replica sends clients home while other replicas continue serving the backup, splitting writes across both sites. Preserve the last successfully read target on errors; when no prior value exists, propagate an unavailable result rather than inventing `home`.

```

</details>

---

portal-service/failover_handler.go
```diff
@@ -0,0 +1,159 @@
...
+			errcode.WithReason(errcode.AuthMissingFields)))
+		return
+	}
+	if !isKnownAction(req.Action) {
+		errhttp.Write(ctx, c, errcode.BadRequest(fmt.Sprintf("unknown action %q", req.Action)))
+		return
+	}
+	// A site absent from the registry has no coordinates to route to, so state
+	// for it would be inert. Reject the typo instead of minting a dead document.
+	if _, ok := h.sites[siteID]; !ok {
```
julianshen, line 143: _🏷️ failover-configuration_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`distributed-system`, `silent-failure`, `tgd-review` (3 rules)_

**Failover commits even when no backup route is configured.**

The POST validates only that the home `siteID` exists. With an empty or incorrect `PORTAL_BACKUP_SITE_ID`, an operator receives 200 and the state is committed as `failed_over`; subsequent `/api/userInfo` requests take the backup branch and return 500 because `servingURLs` cannot find that entry. Validate the configured backup ID and registry entry before accepting `failover`, preferably also failing startup when the control surface is enabled with invalid backup configuration.

<details>

- **`tgd-review`**:114: `Post` validates only the home `siteID`. With the accepted default `PORTAL_BACKUP_SITE_ID=""`, or a missing registry entry, `failover` persists `failed_over` and returns 200. Subsequent `/api/userInfo` requests call `servingURLs`, fail the backup lookup, and return 500 for every affected user. Validate that a distinct configured backup entry exists before enabling or committing failover, and reject the action without mutation otherwise.
- **`silent-failure`**:118: `Post` verifies only that the home `siteId` exists. `PORTAL_BACKUP_SITE_ID` may be empty or absent from the registry, yet `failover` still persists `failed_over` and returns 200. Subsequent `/api/userInfo` calls select `ServingBackup`, fail the backup lookup in `servingURLs`, and return 500 for every affected account. Pass the backup ID into this handler (or validate it at startup) and reject `failover` before `Transition` unless it names a valid registry entry.

</details>

<details>

```
In `portal-service/failover_handler.go` around line 116: The POST validates only that the home `siteID` exists. With an empty or incorrect `PORTAL_BACKUP_SITE_ID`, an operator receives 200 and the state is committed as `failed_over`; subsequent `/api/userInfo` requests take the backup branch and return 500 because `servingURLs` cannot find that entry. Validate the configured backup ID and registry entry before accepting `failover`, preferably also failing startup when the control surface is enabled with invalid backup configuration.

```

</details>

---

portal-service/failover_mongo.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+// mongoFailoverStore persists FailoverState in the failover_states collection,
+// keyed by _id = siteID, with optimistic-concurrency CAS on the version field.
+type mongoFailoverStore struct {
+	coll *mongo.Collection
+}
+
+func newMongoFailoverStore(db *mongo.Database) *mongoFailoverStore {
+	return &mongoFailoverStore{coll: db.Collection("failover_states")}
```
julianshen: _🏷️ mongodb-consistency_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`mongodb`_

**Failover reads inherit the stale secondaryPreferred preference.**

This collection uses the shared Mongo client, whose configured default is `secondaryPreferred`. After `Transition` writes to the primary and returns success, `Get` can read an available lagging secondary, synthesize `healthy` from `ErrNoDocuments`, and cache that result. During an actual outage, `/api/userInfo` then directs users back to the failed home site. Pin `failover_states` reads to the primary (and an appropriate read concern) without changing the directory collection's offloading preference.

<details>

```
In `portal-service/failover_mongo.go` around line 19: This collection uses the shared Mongo client, whose configured default is `secondaryPreferred`. After `Transition` writes to the primary and returns success, `Get` can read an available lagging secondary, synthesize `healthy` from `ErrNoDocuments`, and cache that result. During an actual outage, `/api/userInfo` then directs users back to the failed home site. Pin `failover_states` reads to the primary (and an appropriate read concern) without changing the directory collection's offloading preference.

```

</details>

---

portal-service/failover_reader.go
```diff
@@ -0,0 +1,67 @@
...
+		slog.WarnContext(ctx, "failover reader: store read failed, defaulting to home",
+			"siteId", siteID, "error", err)
+		return ServingHome
+	}
```
julianshen, line 54: _🏷️ NATS routing_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`nats`_

**Store errors silently route failing-back users to home**

While a site is `failing_back`, backup remains authoritative until replay converges. If Mongo times out after the cache expires, this branch returns `ServingHome`; `/api/userInfo` then successfully sends reconnecting clients to home while other clients still write through backup. That creates split writes before convergence and can strand messages. Do not substitute a routing decision for an unreadable authoritative state: return an error so discovery responds 503, or use a deliberately bounded last-known state.

<details>

```
In `portal-service/failover_reader.go` around line 54: While a site is `failing_back`, backup remains authoritative until replay converges. If Mongo times out after the cache expires, this branch returns `ServingHome`; `/api/userInfo` then successfully sends reconnecting clients to home while other clients still write through backup. That creates split writes before convergence and can strand messages. Do not substitute a routing decision for an unreadable authoritative state: return an error so discovery responds 503, or use a deliberately bounded last-known state.

```

</details>

---

portal-service/handler.go
```diff
@@ -186,25 +242,34 @@ func (h *PortalHandler) resolve(ctx context.Context, c *gin.Context, account str
...
+		site, siteOK := h.sites[e.SiteID]
+		baseURL, natsURL = site.BaseURL, site.NATSURL
+		if !siteOK {
+			// The dev-fallback site itself isn't in the registry — fall back to
+			// the legacy PORTAL_DEV_FALLBACK_NATS_URL so local logins keep working.
+			natsURL = h.devFallbackNatsURL // baseURL stays "" as before
+		}
+	} else {
+		// Provisioned account: apply SP4's failover override. siteId stays home;
+		// only the transport URLs may swap to the backup.
```
julianshen, line 257: _🏷️ NATS lifecycle_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`nats`_

**Failover does not move clients with live home NATS connections**

The override only affects a new `/api/userInfo` request. The documented client calls discovery once per login, and JWT renewal does not re-query it. If an operator fails over a partially degraded site while existing NATS WebSockets remain alive, those clients continue publishing at home while new/reconnecting clients publish at backup, violating the single-serving-site assumption and splitting messages. Add an explicit routing-change signal or server-side disconnect/drain that forces established clients through discovery before backup is designated.

<details>

```
In `portal-service/handler.go` around line 257: The override only affects a new `/api/userInfo` request. The documented client calls discovery once per login, and JWT renewal does not re-query it. If an operator fails over a partially degraded site while existing NATS WebSockets remain alive, those clients continue publishing at home while new/reconnecting clients publish at backup, violating the single-serving-site assumption and splitting messages. Add an explicit routing-change signal or server-side disconnect/drain that forces established clients through discovery before backup is designated.

```

</details>

---

portal-service/failover_handler.go
```diff
@@ -0,0 +1,159 @@
...
+	ctx := errcode.WithLogValues(c.Request.Context(), "request_id", c.GetString("request_id"))
+	siteID := c.Param("siteId")
+
+	var req failoverActionRequest
+	if err := c.ShouldBindJSON(&req); err != nil {
+		errhttp.Write(ctx, c, errcode.BadRequest("invalid request body",
+			errcode.WithReason(errcode.AuthMissingFields)))
+		return
+	}
+	// Trim before validating so a whitespace-only value can't pass as attribution.
```
julianshen, line 129: _🏷️ resource-exhaustion_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`oom`_

**The failover POST body can exhaust process memory.**

A caller with the ops token can POST a JSON body containing a multi-gigabyte `operator` or `reason`. `ShouldBindJSON` decodes these fields into strings without any byte limit; the server's 10-second read timeout limits duration, not bytes, so a fast sender can force proportional allocation and cause GC thrashing, container eviction, or process OOM. Bound the request body before decoding and optionally add smaller field-length validation.

<details>

> ‼️ **IMPORTANT**

```suggestion
	c.Request.Body = http.MaxBytesReader(c.Writer, c.Request.Body, 64<<10)
	if err := c.ShouldBindJSON(&req); err != nil {
		errhttp.Write(ctx, c, errcode.BadRequest("invalid request body",
			errcode.WithReason(errcode.AuthMissingFields)))
		return
	}
```

</details>

<details>

```
In `portal-service/failover_handler.go` around line 98: A caller with the ops token can POST a JSON body containing a multi-gigabyte `operator` or `reason`. `ShouldBindJSON` decodes these fields into strings without any byte limit; the server's 10-second read timeout limits duration, not bytes, so a fast sender can force proportional allocation and cause GC thrashing, container eviction, or process OOM. Bound the request body before decoding and optionally add smaller field-length validation.

```

</details>

---

portal-service/failover_reader.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// ServingTarget returns home or backup for siteID. On a cache hit within the TTL
+// it returns the cached value; otherwise it reads the store, derives the target,
+// and caches it. A store error resolves to home (fail-safe) and is not cached.
+func (r *failoverReader) ServingTarget(ctx context.Context, siteID string) ServingTarget {
+	r.mu.Lock()
+	if c, ok := r.cache[siteID]; ok && r.now().Before(c.expires) {
+		r.mu.Unlock()
+		return c.target
```
julianshen, line 44: _🏷️ backpressure_ | _🟠 Warning_ | _⚡ Quick fix_ | _`distributed-system`, `performance-and-concurrency`, `tgd-review` (3 rules)_

**Cache expiry causes a per-site Mongo request stampede.**

Every caller observing an expired entry releases the mutex and independently calls Mongo. During the expected whole-site reconnect burst, thousands of requests for one site can therefore issue identical reads at the same TTL boundary. Mongo slowness then amplifies the burst, and errors trigger more uncached reads on every retry. Coalesce refreshes per site with singleflight or a per-key in-flight entry while allowing different sites to refresh concurrently.

<details>

- **`performance-and-concurrency`**:50: Every concurrent request that observes a cold or expired entry releases the mutex and independently calls `store.Get`. A site outage creates a reconnect wave; when that site's five-second entry expires, the wave can issue one Mongo query per client rather than one refresh per site, increasing latency and potentially exhausting the Mongo pool. Coalesce in-flight refreshes per site with singleflight or a per-key lock while keeping different sites parallel.
- **`tgd-review`**:48: Every caller releases the mutex before `store.Get`, so a cold cache or TTL expiry during a site-wide reconnect sends one Mongo read per login rather than one per site. If the pool queues or errors under that burst, the error path routes clients back to the unavailable home site and encourages more retries. Collapse concurrent misses per site with singleflight or per-key locking while retaining parallel reads for different sites.

</details>

<details>

```
In `portal-service/failover_reader.go` around line 44: Every caller observing an expired entry releases the mutex and independently calls Mongo. During the expected whole-site reconnect burst, thousands of requests for one site can therefore issue identical reads at the same TTL boundary. Mongo slowness then amplifies the burst, and errors trigger more uncached reads on every retry. Coalesce refreshes per site with singleflight or a per-key in-flight entry while allowing different sites to refresh concurrently.

```

</details>

---

portal-service/failover_mongo.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	if err != nil {
+		return FailoverState{}, fmt.Errorf("get failover state %q: %w", siteID, err)
+	}
+	return st, nil
+}
+
+// List returns every stored failover state.
+func (s *mongoFailoverStore) List(ctx context.Context) ([]FailoverState, error) {
+	cur, err := s.coll.Find(ctx, bson.D{})
```
julianshen, line 47: _🏷️ mongodb-data-model_ | _🟠 Warning_ | _⚡ Quick fix_ | _`mongodb`_

**The fleet listing omits sites without persisted state.**

`Get` models a missing document as healthy version 0, but `List` returns only documents physically present in `failover_states`. Before any transition the fleet endpoint returns `[]`; afterward it still omits every untouched healthy site, despite the control surface and runbook describing this as the whole fleet's state. Merge stored states with the configured site registry and synthesize healthy entries for missing active sites, or initialize one document per active site.

<details>

```
In `portal-service/failover_mongo.go` around line 38: `Get` models a missing document as healthy version 0, but `List` returns only documents physically present in `failover_states`. Before any transition the fleet endpoint returns `[]`; afterward it still omits every untouched healthy site, despite the control surface and runbook describing this as the whole fleet's state. Merge stored states with the configured site registry and synthesize healthy entries for missing active sites, or initialize one document per active site.

```

</details>

---

portal-service/main.go
```diff
@@ -177,6 +248,11 @@ func run() error {
 		shutdown.Wait(ctx, 25*time.Second,
 			func(ctx context.Context) error {
 				slog.Info("shutting down portal service")
+				if internalSrv != nil {
+					if err := internalSrv.Shutdown(ctx); err != nil {
```
julianshen, line 256: _🏷️ shutdown_ | _🟠 Warning_ | _⚡ Quick fix_ | _`performance-and-concurrency`_

**Shutdown serializes the internal and public HTTP drains.**

`internalSrv.Shutdown(ctx)` completes before `srv.Shutdown(ctx)` begins. If an internal handler is blocked in Mongo, it can consume the shared 25-second shutdown deadline while the public listener continues accepting requests. The public server then receives an expired context and cannot drain those requests, causing connection resets during deployment. Start both independent server shutdowns concurrently under the same deadline, wait for both, and only then stop refresh work and disconnect Mongo.

<details>

```
In `portal-service/main.go` around line 252: `internalSrv.Shutdown(ctx)` completes before `srv.Shutdown(ctx)` begins. If an internal handler is blocked in Mongo, it can consume the shared 25-second shutdown deadline while the public listener continues accepting requests. The public server then receives an expired context and cannot drain those requests, causing connection resets during deployment. Start both independent server shutdowns concurrently under the same deadline, wait for both, and only then stop refresh work and disconnect Mongo.

```

</details>

---

portal-service/failover_handler.go
```diff
@@ -0,0 +1,159 @@
...
+		Version:       s.Version,
+		Timestamp:     s.Timestamp,
+	}
+}
+
+// List returns every site's failover state.
+func (h *FailoverHandler) List(c *gin.Context) {
+	ctx := errcode.WithLogValues(c.Request.Context(), "request_id", c.GetString("request_id"))
+	states, err := h.store.List(ctx)
+	if err != nil {
```
julianshen, line 95: _🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`tgd-review`_

**The fleet list omits sites that have never transitioned.**

`List` returns only Mongo documents, but healthy sites have no document because `Get` synthesizes their version-0 state. Consequently a normal fleet can produce an empty response despite the endpoint being documented as listing every site's state. Merge stored states with the configured home-site registry and synthesize healthy entries, excluding the reserved backup entry.

<details>

```
In `portal-service/failover_handler.go` around line 68: `List` returns only Mongo documents, but healthy sites have no document because `Get` synthesizes their version-0 state. Consequently a normal fleet can produce an empty response despite the endpoint being documented as listing every site's state. Merge stored states with the configured home-site registry and synthesize healthy entries, excluding the reserved backup entry.

```

</details>

---

julianshen (commented):
### tGDBot review of `53571ec` — 19 findings in 28 files

11 🔴 blocking · 8 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/499#issuecomment-5720281658), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

hmchangw:
@julianshen — I verified both of the findings from that thread against the code. Both are real, and both are fixed in `26cba3bb`.

**1. Stale failover reads.** Confirmed: `newMongoFailoverStore` took the collection straight off the shared client, which `main.go` connects with `secondaryPreferred` (and compose defaults to that), so `Get`/`List` could read a lagging secondary. Worth spelling out why this collection in particular can't tolerate it — there are two distinct consequences, not one:

- `failoverReader.ServingTarget` reads through it, so replication lag stacks *on top of* `FAILOVER_STATE_TTL`, widening exactly the home/backup overlap window documented in runbook §1.5.
- `Post` is a read-modify-write: a stale `Get` yields a stale `cur.Version`, so the CAS filter misses and the operator gets a spurious `409 failover_version_conflict` — during an incident, on the break-glass path. A stale *status* can likewise make a legal action look illegal.

Fixed by pinning the collection to `readpref.Primary()` via the existing `mongoutil.CollectionWithReadPreference` helper. One honest caveat: the driver exposes no read-preference accessor on `*mongo.Collection`, so there is no way to assert this in a test — the guarantee rests on construction, and I've documented it at the constructor and in the runbook's config table.

**2. Failover to a backup nothing can serve.** Confirmed, and worse than a 500 on one endpoint: `PORTAL_BACKUP_SITE_ID` defaults to empty, so in any deployment that hasn't provisioned a backup yet, a `failover` action would commit `failed_over` and then **every** `/api/userInfo` for that site returns 500. The site goes dark rather than failing over — the exact opposite of what the operator invoked. My earlier "loud internal error rather than a silent mis-route" framing was wrong: it's loud at every user's login, not at the operator's request, which is the wrong place to be loud.

Enforced in both places you suggested:
- **Startup** (`validateFailoverConfig`) fails fast when `FAILOVER_OPS_TOKEN` is set but the backup id is empty or absent from `PORTAL_SITE_URLS`. Enabling the control surface is an explicit claim that this deployment can fail over; a deployment that can't should surface that at deploy time, not mid-incident.
- **`Post`** returns `503 failover_backup_unavailable`. I gated it on `next.ServingTarget() == ServingBackup` rather than on the action name, so it covers `failover` and `failback` together and any action added later — and, importantly, never blocks `resume`/`complete`, which return a site *home* and must stay available precisely when the backup is broken. There's a test pinning that recovery path.

Agreed the two root causes are independent; they're separate commits' worth of reasoning but landed together.

**Verification:** `make test SERVICE=portal-service` (race) ✅ · `make build` ✅ · `make lint` 0 issues ✅ · `gosec` PASS ✅ · `validateFailoverConfig`, `backupIsServable` and `Post` all 100% covered. Integration tests compile under the `integration` tag and run in CI.

---

hmchangw:
**CI `test` failed on `26cba3bb`, and it is not this PR's.** Re-queued the failed job; details so nobody has to re-derive this.

**What failed**

```
--- FAIL: TestAppNameCache_NamesCollapsesConcurrentColdBatches (0.00s)
    pkg/preview/appname_test.go:329
        expected: 1
        actual  : 2
        concurrent misses on one account set must collapse to one read
FAIL	github.com/hmchangw/chat/pkg/preview	0.052s
```

`portal-service` passed in the same run, as did every other package.

**Why it isn't mine.** This branch does not touch `pkg/preview` — `git diff origin/main...HEAD -- pkg/preview` is empty; the diff is `portal-service/*`, `pkg/errcode/codes_portal.go`, `portal-service/deploy/docker-compose.yml` and docs. I reproduced the failure on `origin/main` at `c412ece3` (the PR's own merge base, none of my commits present): `go test -race -run TestAppNameCache_NamesCollapsesConcurrentColdBatches -count=40 ./pkg/preview/` fails there. Same command on this branch fails at a comparable rate — roughly 2 in 40 either way. Pre-existing flake, ~5%.

**Root cause, for whoever owns `pkg/preview`.** The test is racy by construction rather than flaky-by-timing:

```go
release := make(chan struct{})
c := NewAppNameCache(nil, func(...) (map[string]string, error) {
    calls.Add(1)
    <-release
    ...
})
for range 8 { go func() { c.Names(ctx, []string{"a.bot", "b.bot"}) }() }
close(release)   // <-- races the 8 goroutines reaching Names
wg.Wait()
```

The barrier is meant to pile all eight callers onto one in-flight leader before letting it proceed, but nothing orders `close(release)` after the goroutines have entered `Names`. When the scheduler lets the first goroutine run through — it enters the fetch, finds `release` already closed, returns, and the singleflight group retires — a straggler that calls `Names` afterwards finds no leader and opens a second read. Hence `2`.

It can't be fixed by moving `close(release)` alone: the followers block *inside* the singleflight, not anywhere the test can observe, so there's no state to wait on from outside. It needs either a counter the fetch fn can wait on (leader blocks until N−1 followers have registered) or a test hook on the cache. That's a change to `pkg/preview`'s test design and arguably its testability surface.

**Not fixing it here.** It's unrelated to SP3/SP4 failover, in a package this PR doesn't touch, and the fix is a design decision for that package's owner — folding it into a failover PR would be scope reviewers wouldn't expect. Flagging it instead; worth its own issue. If you'd rather I take it, say so and I'll do it in a separate PR.

I'll keep watching #499 and report the re-run.

---

PR #405, opened 2026-08-27
perf(room-worker): remove the large-room cliffs on the membership paths

hmchangw:
Fixes the `critical` and all four `high` performance findings from a production-readiness audit of `room-worker`. Every change is behaviour-preserving except where noted, and each was written test-first.

## The defects

**`critical` — nothing extended the ack deadline.** `AckWait` is 30s (`pkg/stream/consumer.go:17`) and no handler in the repo called `InProgress()`. Any handler that ran longer was redelivered mid-flight, so a second worker began the same job concurrently — for a large-room removal that means a duplicate key rotation and a duplicate fan-out to every survivor. Removing a member from a ~5k-member room, or reconciling a Teams batch of more than about twenty chats, crosses that line.

**`high` — rename read every subscription document in the room.** `processRoomRename` called `ListByRoom` and used only `sub.User.Account`. A 5k-member rename shipped 5k full documents (roles, read state, timestamps) to extract one string each.

**`high` — the remove paths ran an unconditional full-room double count.** `ReconcileMemberCounts` is two `CountDocuments` over the whole room, called on every removal. Dropping one member from a 50k-member room cost two full index scans; an org removal that freed nobody paid the same price for a no-op.

**`high` — the org-member aggregation carried whole user documents into two correlated `$lookup`s.** Projection was the last stage, so removing a 5,000-person department dragged 5k full user docs through both joins inside one 30s `AckWait`.

**`high` — bulk writes were unchunked.** `BulkCreateSubscriptions` and `BulkCreateRoomMembers` built one `WriteModel` per candidate and sent them as a single `BulkWrite`; a 20k-user org add became one 20k-model command held whole in memory.

## The changes

| Area | Change |
|---|---|
| `pkg/jsretry/heartbeat.go` | New `Heartbeat` (per-message ticker calling `InProgress` until stopped) and `HeartbeatInterval` (`AckWait/3`, so two ticks fit in the budget and one lost tick is survivable). A non-positive `AckWait` disables it rather than inventing an interval; a failing `InProgress` ends the loop instead of spamming the log. |
| `room-worker/main.go` | Consumer loop starts the heartbeat and stops it from the same defer that finishes the tracked message, so the goroutine terminates on the panic path too. |
| `room-worker/handler.go` | Rename uses the already-projected `GetSubscriptionAccounts`. Both remove paths `$inc` an exact delta via `ApplyMemberCountDelta` and recompute only when the TTL reports a drift check due. |
[...]

PR #501, opened 2026-09-17
perf(jsretry): keep a message's ack deadline alive while its handler runs

hmchangw:
Split 1 of 4 from #405. Independent of the other three — touches `pkg/jsretry` and `pkg/stream` plus `room-worker/main.go`, which no sibling split modifies. Merges in any order.

## The defect

`AckWait` is 30s (`pkg/stream/consumer.go`) and **nothing in the repo called `InProgress()`** — `grep -rn "InProgress()"` returned nothing. A handler that ran longer than its budget was redelivered mid-flight, so a second worker began the same job concurrently. For a `room-worker` large-room removal that means a duplicate key rotation and a duplicate fan-out to every survivor. Removing a member from a ~5k-member room, or reconciling a Teams batch of more than about twenty chats, crosses that line.

## The change

`jsretry.Heartbeat` is a per-message ticker calling `InProgress` until stopped or until its budget is spent. `HeartbeatInterval` derives the period as `AckWait/3`, so two ticks fit inside the budget and one lost tick is survivable; a non-positive `AckWait` disables it rather than inventing an interval. It paces off `cons.CachedInfo().Config.AckWait`, the deadline the server actually applied, rather than the one we requested.

**The extension is bounded, and unbounded is deliberately not expressible.** `HeartbeatBudget.Max` caps total extension, and an unset `Max` resolves to `jsretry.DefaultHeartbeatMax` rather than to forever — omission is the case a caller gets wrong. The operator knob is `CONSUMER_HEARTBEAT_MAX` (default 10m), declared once on `stream.ConsumerSettings` so every consumer that later adopts the heartbeat shares it, with a test pinning it to `DefaultHeartbeatMax` so the two cannot drift.

The bound is not decoration. `room-worker`'s `msgCtx` comes from the otel-nats iterator, which roots it in `context.Background()` on both the traced and passthrough paths — so `Heartbeat`'s `ctx.Done()` arm never fires there, leaving `stop()` as the only other exit. Without a cap, a handler that never returns would heartbeat forever: the message holds its `MaxAckPending` slot indefinitely, is never redelivered, and never reaches `MaxDeliver`. That turns a bounded failure into an invisible permanent one and neutralizes the hang arm of the server-side `BackOff` lever. Spending the budget logs at Warn and hands the message back to that lever.

It restores the message's escape hatch, not the worker slot — a wedged handler still holds its semaphore slot until it returns.

## Behaviour worth reading closely

- **`stop()` waits for the goroutine to exit**, so no `InProgress` is in flight once it returns. It does **not** order the stop before settlement: `room-worker`'s handler settles inside `runJobWithRecovery` (via `jsretry.SettleQuiet`, which `Ack`s), and `stopHeartbeat()` runs in the defer afterwards. See the known gaps below.
- **A failing `InProgress` keeps retrying.** It cannot distinguish a settled message from a transport blip, so the safe reading is transient: keep extending rather than abandoning the deadline silently.
- **`room-worker` starts the heartbeat at delivery**, not inside the worker goroutine. A message queued on the semaphore is already spending its ack deadline — heartbeating only after a worker frees up left exactly the gap this PR exists to close.

## Known gaps — follow-ups, not blockers

- **The heartbeat stops after settlement, not before.** A tick can in principle call `InProgress` on an already-settled message. The window is the microseconds between the handler's `Ack` and the deferred `stop()`, against a 10s tick, and the server ignores `AckProgress` for a sequence that is no longer pending — a correctness wart rather than a live hazard.
- **The 1s interval floor can meet or exceed a very short `AckWait`.** Dormant at the 30s default (10s interval); it would only bite below 3s. The room-worker headroom test asserts the two-ticks invariant for 30s/1m/5m only, so a lowered `AckWait` could violate it silently.
- **A heartbeat failure is logged once, at Debug.** Persistent `InProgress` failures are invisible at typical log levels, with no metric.
- **This narrows duplicate execution; it does not fence it.** If NATS is unreachable for at least `AckWait` while MongoDB stays reachable, a concurrent redelivery is still possible. Durable fencing on a stable message ID is the real fix and is out of scope here.
- **Only `room-worker` adopts the heartbeat.** Every other consumer still rides `AckWait` alone.

## Verification

CI is green on `15c8f82`: `lint`, `test` (full repo, `-race`), `sast`, and both `test-integration` jobs.

The budget resolution and the stop-on-exhaustion behaviour are covered in `pkg/jsretry`, the tag/constant pinning in `pkg/stream`, and the wiring in `room-worker`. Default-asserting tests parse an explicitly empty environment, so an operator's `CONSUMER_HEARTBEAT_MAX` or `CONSUMER_ACK_WAIT` cannot silently decide what "the default" is.

---

https://claude.ai/code/session_018N3uLbz9RRXm3YmeNA1oGL

## Summary by CodeRabbit

- **New Features**
  - Long-running message processing now periodically keeps the acknowledgment deadline active, reducing premature redeliveries.
  - Heartbeat activity stops automatically when processing ends, the context is cancelled, or the configured time limit is reached.
  - Added a configurable heartbeat limit with a default of 10 minutes.

- **Documentation**
  - Updated guidance to explain heartbeat timing, limits, and their relationship to acknowledgment wait settings.


julianshen:
**9 findings · 3 unique issues · 3 inline comments posted** — 🔴 2 blocking · 🟠 7 warning

### Related work

- **Heartbeat failures still permit concurrent MongoDB mutations**
  - `pkg/jsretry/heartbeat.go:58` (`mongodb`) — Heartbeat failures still permit concurrent MongoDB mutations
  - `room-worker/main.go:334` (`distributed-system`) — Context cancellation reopens concurrent redelivery
  - `pkg/jsretry/heartbeat.go:25` (`performance-and-concurrency`) — The interval floor can exceed the message's ack deadline
  - `room-worker/main.go:347` (`tgd-review`) — The heartbeat stops only after the message may already be Acked.
  - `pkg/jsretry/heartbeat.go:25` (`distributed-system`) — The interval floor can exceed the acknowledgement deadline
  - `pkg/jsretry/heartbeat.go:56` (`silent-failure`) — Persistent heartbeat failures are discarded while work continues
  - `pkg/jsretry/heartbeat.go:24` (`tgd-review`) — The one-second floor can place the first heartbeat after expiry.
  - `pkg/jsretry/heartbeat.go:25` (`silent-failure`) — The interval floor allows the ack deadline to expire before any heartbeat
  - `pkg/jsretry/heartbeat.go:24` (`nats`) — Short AckWait values expire before the first heartbeat

<details>

* `pkg/jsretry/heartbeat.go`
* `pkg/jsretry/heartbeat_test.go`
* `room-worker/consumer_config_test.go`
* `room-worker/main.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

julianshen (commented):
### tGDBot review of `e94bc7f` — 9 findings in 4 files

2 🔴 blocking · 7 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/501#issuecomment-5719273274), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

room-worker/main.go
```diff
@@ -319,21 +328,26 @@ func main() {
 				consumerMetrics.LoopFailed(context.Background(), err)
 				return
 			}
+			// Heartbeat from delivery, not from when a worker frees up: a
+			// message queued on the semaphore is already spending its ack
+			// deadline, and a large-room mutation can outrun what is left.
+			stopHeartbeat := jsretry.Heartbeat(msgCtx, msg, heartbeatEvery)
```
julianshen: _🏷️ delivery guarantees_ | _🔴 Blocking_ | _⚡ Quick fix_ | _`distributed-system`_

**Context cancellation reopens concurrent redelivery**

`Heartbeat` exits when `msgCtx` is canceled, but this caller stops it only after `runJobWithRecovery` returns. During shutdown, cancellation can therefore end heartbeats while a handler is still running. If processing continues beyond `AckWait`, JetStream can redeliver the message to another worker while the original performs key rotation or fan-out. Detach heartbeat cancellation here so the deferred stop remains its lifetime boundary.

<details>

> ‼️ **IMPORTANT**

```suggestion
			stopHeartbeat := jsretry.Heartbeat(context.WithoutCancel(msgCtx), msg, heartbeatEvery)
```

</details>

<details>

```
In `room-worker/main.go` around line 334: `Heartbeat` exits when `msgCtx` is canceled, but this caller stops it only after `runJobWithRecovery` returns. During shutdown, cancellation can therefore end heartbeats while a handler is still running. If processing continues beyond `AckWait`, JetStream can redeliver the message to another worker while the original performs key rotation or fan-out. Detach heartbeat cancellation here so the deferred stop remains its lifetime boundary.

```

</details>

---

pkg/jsretry/heartbeat.go
```diff
@@ -0,0 +1,66 @@
...
+			select {
+			case <-done:
+				return
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				// InProgress cannot distinguish a settled message from a
+				// transport blip, so keep extending and log only the first
+				// failure rather than giving the deadline up silently.
+				if err := msg.InProgress(); err != nil && !reported {
```
julianshen, line 96: _🎯 MongoDB correctness_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`, `mongodb`, `nats`, `performance-and-concurrency`, `silent-failure`, `tgd-review` (6 rules)_

**Heartbeat failures still permit concurrent MongoDB mutations**

An `InProgress` failure is only logged; no failure reaches room-worker and the handler continues. If NATS is unavailable for at least `AckWait` while MongoDB remains reachable, none of these retries reaches the server, JetStream redelivers, and the original worker continues the same mutation. For the described room-removal workload, this recreates concurrent key rotation and duplicate fan-out. Keep the heartbeat as an optimization, but durably fence or deduplicate the mutation using a stable event/message ID; coordinate multi-document state and fan-out transactionally or through an idempotent outbox.

<details>
<summary>Also reported by 6 other rules</summary>

- **`performance-and-concurrency`**:25: For an AckWait below one second, this returns a one-second interval; the added test explicitly demonstrates 100ms. Because Heartbeat does not send an initial InProgress, that message expires before the first ticker event. Any handler running longer than 100ms can therefore be redelivered concurrently, recreating the duplicate-work path this change addresses. Use an interval below AckWait, such as AckWait/3, or reject configurations whose AckWait is too short for the minimum interval, and update the test accordingly.
- **`distributed-system`**:25: For a server-applied `AckWait` below one second, this returns one second; the included 100ms test explicitly exercises that case. Because the ticker sends no immediate heartbeat, JetStream can expire and redeliver the message before the first `InProgress`, allowing concurrent processing. Return an interval strictly below `AckWait` (such as `AckWait/3`) or reject consumer configurations whose deadline is too short for the minimum interval, and update the floor test accordingly.
- **`silent-failure`**:56: Only the first `InProgress` error is debug-logged; subsequent failures are ignored, and `stop` cannot report that deadline extensions never succeeded. If failures persist for an AckWait while the handler runs, the message can be redelivered and processed concurrently while the original worker continues normally. Track successful extensions or consecutive failures and signal/cancel the worker before the current deadline expires, while surfacing the terminal heartbeat failure.
- **`tgd-review`**:24: For any positive AckWait below three seconds, `max` selects one second instead of AckWait/3. At AckWait <= one second, the first tick occurs at or after expiration, so the message can be redelivered before `InProgress`; between one and three seconds the promised headroom is lost. Either permit shorter heartbeat intervals or disable/reject AckWait values for which the one-second load floor cannot safely extend the deadline, and test these boundaries.
- **`silent-failure`**:25: For the explicitly tested 100ms AckWait, this returns one second, so the first `InProgress` call occurs ten times after the original deadline. A handler or semaphore wait exceeding 100ms can therefore be redelivered even though its heartbeat is active, without any reported failure. Ensure every positive returned interval is strictly below AckWait, preferably AckWait/3, and update the short-wait test accordingly.
- **`nats`**:24: For an applied AckWait of 100ms—the case explicitly covered by the new test—this returns a 1s interval. JetStream can therefore redeliver the message about 900ms before the first InProgress call; AckWait values below 3s also lose the claimed missed-tick headroom. Remove the 1s floor and update the test, or reject consumer configurations whose AckWait is too short for the minimum interval.

</details>

<details>

```
In `pkg/jsretry/heartbeat.go` around line 58: An `InProgress` failure is only logged; no failure reaches room-worker and the handler continues. If NATS is unavailable for at least `AckWait` while MongoDB remains reachable, none of these retries reaches the server, JetStream redelivers, and the original worker continues the same mutation. For the described room-removal workload, this recreates concurrent key rotation and duplicate fan-out. Keep the heartbeat as an optimization, but durably fence or deduplicate the mutation using a stable event/message ID; coordinate multi-document state and fan-out transactionally or through an idempotent outbox.

```

</details>

---

room-worker/main.go
```diff
@@ -319,21 +328,26 @@ func main() {
...
 				tracked := consumerMetrics.Track(msgCtx, msg, natsmetrics.RoomEventTypeFromSubject(msg.Subject()), consumerCfg.MaxDeliver)
 				msgCtx = tracked.Context(msgCtx)
 				// runJobWithRecovery contains handler panics (it Acks — drops — the
 				// poison message) so this async goroutine, which runs outside
 				// natsrouter's recovery middleware, can't crash the worker.
 				defer func() {
+					stopHeartbeat()
 					tracked.Finish(msgCtx)
 					<-sem
 					wg.Done()
```
julianshen, line 347: _🎯 correctness_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`tgd-review`_

**The heartbeat stops only after the message may already be Acked.**

`stopHeartbeat` runs in the defer after `runJobWithRecovery` returns. The adjacent comment confirms that the recovery path Acks a panicking handler before returning, so an in-flight `InProgress` can overlap or follow that Ack; waiting inside `stop` does not provide the claimed settlement ordering. Stop the heartbeat before every Ack/Nak/Term, including the recovery Ack, likely by integrating it with the settlement wrapper or recovery path, and test settlement against an in-flight heartbeat.

<details>

```
In `room-worker/main.go` around line 347: `stopHeartbeat` runs in the defer after `runJobWithRecovery` returns. The adjacent comment confirms that the recovery path Acks a panicking handler before returning, so an in-flight `InProgress` can overlap or follow that Ack; waiting inside `stop` does not provide the claimed settlement ordering. Stop the heartbeat before every Ack/Nak/Term, including the recovery Ack, likely by integrating it with the settlement wrapper or recovery path, and test settlement against an in-flight heartbeat.

```

</details>

---

PR #502, opened 2026-09-17
perf(mongoutil): chunk unordered bulk writes

hmchangw:
Split 2 of 4 from #405. Independent of the other three. It shares `room-worker/store_mongo.go` with #503, but in different functions (the `Bulk*` writers vs. the read pipelines), so only the second to merge needs a trivial rebase.

## The defect

`BulkCreateSubscriptions` and `BulkCreateRoomMembers` built one `WriteModel` per candidate and sent them as a single `BulkWrite`. Adding a 20k-user org turned into one 20k-model command held whole in memory.

## The change

`ChunkedBulkWrite` issues unordered batches of `MaxBulkChunk` (1000) and merges the results, rebasing `UpsertedIDs` ordinals onto the whole input so per-chunk indexes cannot collide. Chunks are not atomic with respect to each other, which is safe for these callers: every model they build is an upsert, so a retry re-applies cleanly.

Two things the naive loop got wrong, both caught in review on #405:

- **A failed chunk still returns the models it did write.** The pinned driver ends `BulkWrite` with `return &op.result, wrapErrors(err)`, so checking `err` before merging drops writes that durably landed. Merge first.
- **Chunking must not narrow unordered semantics.** A single unordered `BulkWrite` carries on past per-model write errors; chunking turned one bad model into a barrier for every later chunk — a regression against the code it replaced. `terminalBulkError` now continues on per-model write errors and stops only for context, transport or write-concern failures, aggregating what it saw.

The capacity computation uses `1+(len-1)/size`: `len+size-1` overflows for a size near `MaxInt` with two or more models, panicking in `make`.

`room-worker`'s three bulk paths adopt the helper.

## Verification

`make lint` and `make test` (full repo, `-race`) both pass on this branch alone.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QBGxDVzzDGJ8VfWxz5rhzy)_

PR #503, opened 2026-09-17
perf(room-worker): project the rename read and narrow the org-member aggregation before its joins

hmchangw:
Split 3 of 4 from #405. Independent of the other three. It shares two files with siblings but in disjoint regions — `store_mongo.go` with #502 (read pipelines vs. the `Bulk*` writers) and `handler.go`/`handler_test.go` with #504 (`processRoomRename` vs. the remove paths) — so only the second to merge needs a trivial rebase.

## The defects

Two reads fetched far more than their callers used.

**Rename read every subscription document in the room.** `processRoomRename` called `ListByRoom` and used only `sub.User.Account`. A 5k-member rename shipped 5k full subscription documents — roles, read state, timestamps — to extract one string each. The projected `GetSubscriptionAccounts` already existed.

**The org-member aggregation carried whole user documents into two correlated `$lookup`s.** Projection was the final stage, so removing a 5,000-person department dragged 5k full user docs through both joins inside one 30s `AckWait`.

## The change

Rename uses `GetSubscriptionAccounts`. `ListByRoom` stays on the interface for the Teams reconcile path, which genuinely diffs per-subscription `joinedAt` — its doc comment claimed it was test-only, which stopped being true when that path adopted it.

A narrowing `$project` now sits directly after the indexed `$match`, keeping only the fields later stages read. The pipeline moves into `orgMembersPipeline` so the stage order is unit-testable without Mongo — `store_pipeline_test.go` asserts the projection precedes the joins and that it keeps every field they depend on, since dropping one would silently blank a name or break the sibling-org join.

All four `$lookup` sites gain the `// $lookup justification:` marker CLAUDE.md requires; they were carrying the explanation as ordinary prose.

## Verification

`make lint` and `make test` (full repo, `-race`) both pass on this branch alone.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QBGxDVzzDGJ8VfWxz5rhzy)_


julianshen:
**No actionable comments.** ✅

### Related work

- [PR #502](https://github.com/hmchangw/newchat/pull/502) — perf\(mongoutil\): chunk unordered bulk writes (open)
- [PR #504](https://github.com/hmchangw/newchat/pull/504) — perf\(room-worker\): apply exact member-count deltas on the remove paths (open)

<details>

* `room-worker/handler.go`
* `room-worker/store_pipeline_test.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

PR #504, opened 2026-09-17
perf(room-worker): make removing members from large rooms much faster

hmchangw:
Split 4 of 4 from #405. Independent of the other three. Rebased onto `main` after #503 merged.

## The defect

Both remove paths called `ReconcileMemberCounts` unconditionally, which is two `CountDocuments` over the whole room. Removing one member from a 50k-member room cost two full index scans, and an org removal that freed nobody paid the same price for a no-op.

## The change

Follow the add path: `$inc` an exact delta and recompute only when `ApplyMemberCountDelta` reports the TTL drift check due.

Redelivery safety comes from keying the delta on **what the delete actually removed**, never on what it targeted:

- **Org path** — a partial delete falls back to the full recompute, since the target set no longer describes what left.
- **Individual path** — a zero delete does the same. It cannot distinguish a completed prior delivery from one that failed *before* adjusting the counters, and assuming the former would leave the count permanently high. (This was the review finding on #405; the first version skipped all count work on a zero delete.)

`finishCreateRoom` and the Teams reconcile deliberately keep `ReconcileMemberCounts`. Those pass the full intended set rather than a live-state diff, so a delta there would double-count on redelivery — recompute-and-`$set` is what makes them idempotent.

## The delta follows the persisted flag (second commit)

`ReconcileMemberCounts` counts `u.isBot == true` and derives `userCount` by subtraction, so a row is an app **iff its stored flag is exactly true** — and a subscription written before that flag existed sits in `userCount` however its account reads. Re-deriving bot-ness from the account string would decrement `appCount` for such a row, leaving `userCount` one high and `appCount` one low (possibly negative) until the next reconcile. The old unconditional recompute did not have this failure, so the first commit introduced it; this one closes it by reading the flag off the row that actually left.

- `DeleteSubscription` uses `FindOneAndDelete` and returns the deleted row's flag. Same single round trip as the `DeleteOne` it replaces, and atomic.
- `DeleteSubscriptionsWithBotSplit` counts bot rows across the target set before deleting, since `DeleteMany` cannot return documents.

### What the bot count actually costs

An earlier revision of this description called that count "index-backed over the target set", which oversold it. The count seeks `{roomId, u.account}` over the `$in` set, so it never scans the whole room — but **no index carries both `u.account` and `u.isBot`** (`subscriptions` has `{roomId, u.account}` and `{roomId, u.isBot}`, `room-service/store_mongo.go:124,137`), so it is not covered and fetches one document per target to read the flag. An org removal freeing K members therefore does K seeks plus K fetches, then the `DeleteMany` seeks the same K again.

That is a real cost on a large org, and it is the price of an exact delta. It still beats the two whole-room scans it replaced whenever the room is much larger than the org, which is the case this PR targets. Making it genuinely covered needs a `{roomId, u.account, u.isBot}` index, which is room-service-owned schema with its own write amplification — a follow-up, not this PR.

Because that cost is real, the split is a **separate store method**. The Teams reconcile recomputes authoritatively straight afterwards, so it uses the plain `DeleteSubscriptionsByAccounts` and pays nothing; only `processRemoveOrg`, which needs the split for its delta, calls `DeleteSubscriptionsWithBotSplit`. A third commit fixed that after review caught the Teams path paying for a count it discarded.

`subIsBot` keeps the single stamping-site use in `newSub` it was extracted for; the remove paths no longer re-derive it.

## Known limitation, not addressed here

Deletion, delta and reconciliation are three separate writes, so a due reconcile can publish an authoritative snapshot while an earlier deletion still has an unapplied delta, leaving the count one low.

This is the delta-plus-due-reconcile contract the add path already ships, but it is newly *reachable* on the remove paths: their previous terminal write was an authoritative `$set`, which an `$inc` cannot lose an update against. A version fence or transaction would change `ApplyMemberCountDelta`'s contract for every caller, so it belongs in a change of its own.

Worth stating precisely how such drift clears, since "until the next reconcile" reads as a bounded wait and is not one. `countsReconciledAt` advances only on a real reconcile, so any room whose last reconcile is older than the TTL recomputes on its **next** count-touching event — the longer a room has been quiet, the more certain that is. Drift therefore persists only in a room that drifts while hot and then goes permanently quiet. Note also that an org removal freeing nobody now does no count work at all, where it previously recomputed; that removes a healing opportunity, which is the intended trade for not paying two full scans on a no-op.

## Verification

`make lint`, `make test` (full repo, `-race`) and `gosec` pass on this branch. `govulncheck` and `semgrep` could not run in the dev container (blocked vulndb fetch, tool not installed); CI's `sast` job covers them. The new store-level integration tests compile locally but need Docker, so CI's `test-integration (room-worker)` job is what exercises them.

https://claude.ai/code/session_01EyLMKF1sJUndmx7reeLFNj

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved member-count accuracy after individual and organization member removals.
  - Bot and user subscriptions are now counted based on their persisted classification, including legacy subscriptions.
  - Added safer handling for missing or partially completed deletions, with authoritative reconciliation when needed.
  - Avoided unnecessary count updates when organization removals do not delete any subscriptions.
  - Improved deletion behavior so absent subscriptions are handled without errors.


julianshen:
**5 findings · 5 inline comments posted** — 🔴 2 blocking · 🟠 3 warning

### Related work

- [PR #503](https://github.com/hmchangw/newchat/pull/503) — perf\(room-worker\): shape the Mongo reads on the rename and org-remove paths (open)

<details>

* `room-worker/debug_log_test.go`
* `room-worker/handler.go`

</details>
<details>

* `cassandra`
* `dead-code`
* `distributed-system`
* `elasticsearch`
* `mongodb`
* `nats-large-payload`
* `nats`
* `oom`
* `silent-failure`
* `tgd-review`

</details>

---

room-worker/handler.go
```diff
@@ -442,12 +442,22 @@ func (h *Handler) processRemoveIndividual(ctx context.Context, req *model.Remove
 		return err
 	}
 
-	if err := h.store.ReconcileMemberCounts(ctx, req.RoomID); err != nil {
+	if deleted > 0 {
+		userDelta, appDelta := -1, 0
+		if subIsBot(req.Account) {
+			userDelta, appDelta = 0, -1
+		}
+		if err := h.applyMemberRemovalCounts(ctx, req.RoomID, userDelta, appDelta); err != nil {
```
julianshen, line 451: _🏷️ distributed-systems_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`distributed-system`_

**Concurrent redelivery can double-count a member removal**

Two deliveries can both pass the membership read. Delivery A deletes the subscription and pauses; delivery B gets a zero delete and reconciles from subscriptions, setting the count to N-1. A then applies -1 here, leaving N-2 even though only one member left. The same interleaving exists for an org partial-delete reconciliation. The incorrect count can persist indefinitely without another due reconciliation. Fence reconciliations against unapplied deletions or make deletion and counter mutation transactional; otherwise retain an authoritative reconciliation after the delta.

<details>

```
In `room-worker/handler.go` around line 450: Two deliveries can both pass the membership read. Delivery A deletes the subscription and pauses; delivery B gets a zero delete and reconciles from subscriptions, setting the count to N-1. A then applies -1 here, leaving N-2 even though only one member left. The same interleaving exists for an org partial-delete reconciliation. The incorrect count can persist indefinitely without another due reconciliation. Fence reconciliations against unapplied deletions or make deletion and counter mutation transactional; otherwise retain an authoritative reconciliation after the delta.

```

</details>

---

julianshen, line 451: _🏷️ atomicity_ | _🔴 Blocking_ | _🏗️ Heavy lift_ | _`mongodb`_

**A delayed removal delta can corrupt a concurrent recount**

Subscription deletion and the room-counter `$inc` are separate writes. If removal A deletes its subscription and pauses, removal B can delete another subscription, apply its delta, and run a due `ReconcileMemberCounts`. That recount includes both deletions, but A can then apply another `-1`, leaving the room document undercounted or negative until a later reconciliation. The previous unconditional recount did not have this ordering failure. Make deletion and delta atomic in one transaction with the session propagated to both operations, or add a version fence so reconciliation rejects or includes pending deltas.

<details>

```
In `room-worker/handler.go` around line 450: Subscription deletion and the room-counter `$inc` are separate writes. If removal A deletes its subscription and pauses, removal B can delete another subscription, apply its delta, and run a due `ReconcileMemberCounts`. That recount includes both deletions, but A can then apply another `-1`, leaving the room document undercounted or negative until a later reconciliation. The previous unconditional recount did not have this ordering failure. Make deletion and delta atomic in one transaction with the session propagated to both operations, or add a version fence so reconciliation rejects or includes pending deltas.

```

</details>

---

julianshen, line 451: _🏷️ nats-redelivery_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`nats`_

**Concurrent redelivery can apply a delta after reconciliation**

Delivery A can delete the subscription and pause before applying its delta. If AckWait expires, delivery B sees `deleted == 0` and reconciles to the authoritative subscription count; A then resumes and applies `-1`, persisting an undercount. The org branch has the same full-delete/zero-delete interleaving. Previously both remove paths reconciled, so the final write remained authoritative; this race is newly extended to them despite the PR description. Make deletion and count adjustment atomic or version-fenced, serialize operations per room, or retain reconciliation until that contract exists. Add an overlapping-delivery test.

<details>

```
In `room-worker/handler.go` around line 450: Delivery A can delete the subscription and pause before applying its delta. If AckWait expires, delivery B sees `deleted == 0` and reconciles to the authoritative subscription count; A then resumes and applies `-1`, persisting an undercount. The org branch has the same full-delete/zero-delete interleaving. Previously both remove paths reconciled, so the final write remained authoritative; this race is newly extended to them despite the PR description. Make deletion and count adjustment atomic or version-fenced, serialize operations per room, or retain reconciliation until that contract exists. Add an overlapping-delivery test.

```

</details>

---

room-worker/handler.go
```diff
@@ -1519,13 +1551,34 @@ func subscriptionRoomFor(room *model.Room, pair *roomkeystore.VersionedKeyPair)
...
+}
+
+// applyMemberRemovalCounts $inc's an exact delta, recomputing only when the TTL
+// says drift is due: ReconcileMemberCounts scans the whole room twice.
+func (h *Handler) applyMemberRemovalCounts(ctx context.Context, roomID string, userDelta, appDelta int) error {
+	reconcileDue, err := h.store.ApplyMemberCountDelta(ctx, roomID, userDelta, appDelta, h.reconcileTTL)
+	if err != nil {
+		return fmt.Errorf("apply member count delta: %w", err)
+	}
+	if reconcileDue {
```
julianshen, line 1563: _🎯 concurrency_ | _🟠 Warning_ | _🏗️ Heavy lift_ | _`performance-and-concurrency`_

**Reconciliation can include another removal before its delta**

The new reconciliation protocol is not ordered with other handlers. Starting at N, handler A deletes member 1 and pauses; B deletes member 2, applies -1, gets `reconcileDue`, and stores the authoritative N-2; A then applies its -1, leaving N-3. A concurrent duplicate can cause the same lost update through the zero-delete reconciliation branch. Counts remain low until a later reconciliation. Serialize deletion through counter adjustment per room across workers, or transaction/version-fence deletions, deltas, and snapshot writes.

<details>

```
In `room-worker/handler.go` around line 1567: The new reconciliation protocol is not ordered with other handlers. Starting at N, handler A deletes member 1 and pauses; B deletes member 2, applies -1, gets `reconcileDue`, and stores the authoritative N-2; A then applies its -1, leaving N-3. A concurrent duplicate can cause the same lost update through the zero-delete reconciliation branch. Counts remain low until a later reconciliation. Serialize deletion through counter adjustment per room across workers, or transaction/version-fence deletions, deltas, and snapshot writes.

```

</details>

---

room-worker/handler.go
```diff
@@ -656,8 +669,27 @@ func (h *Handler) processRemoveOrg(ctx context.Context, req *model.RemoveMemberR
 		return fmt.Errorf("delete room member (org): %w", err)
 	}
 
-	if err := h.store.ReconcileMemberCounts(ctx, req.RoomID); err != nil {
-		return fmt.Errorf("reconcile member counts: %w", err)
+	// Keyed on what the delete actually removed, so a redelivery is a no-op.
+	switch {
+	case len(accounts) == 0:
```
julianshen, line 675: _🎯 correctness_ | _🟠 Warning_ | _⚡ Quick fix_ | _`silent-failure`_

**Org retry can acknowledge failed count updates**

Subscriptions and the org room-member row are deleted before `applyMemberRemovalCounts` runs. If the delta write or a due reconciliation fails, the handler returns an error; on redelivery, the deleted org row can no longer produce `accounts`, so this branch skips all count work and the retry can succeed with a stale count. Move `DeleteRoomMember` after count handling so retries can reconstruct the target set, or reconcile when `accounts` is empty. Add an org-path retry test covering a count failure after deletion.

<details>

```
In `room-worker/handler.go` around line 674: Subscriptions and the org room-member row are deleted before `applyMemberRemovalCounts` runs. If the delta write or a due reconciliation fails, the handler returns an error; on redelivery, the deleted org row can no longer produce `accounts`, so this branch skips all count work and the retry can succeed with a stale count. Move `DeleteRoomMember` after count handling so retries can reconstruct the target set, or reconcile when `accounts` is empty. Add an org-path retry test covering a count failure after deletion.

```

</details>

---

julianshen (commented):
### tGDBot review of `7082c0e` — 5 findings in 3 files

2 🔴 blocking · 3 🟠 warning
Findings not shown inline are in the [review summary](https://github.com/hmchangw/newchat/pull/504#issuecomment-5718804177), which is authoritative for what this run found and where each finding went.

<details>

</details>

<details>

One command per comment.

</details>

---

room-worker/handler.go
hmchangw: Not fixing this one — the premise doesn't hold. The retry does reconstruct `accounts`, so it reconciles rather than silently acking a stale count.

`accounts` comes from `GetOrgMembersWithIndividualStatus`, whose pipeline (`room-worker/store_mongo.go:391-467`) matches on the **users** collection — `sectId`/`deptId == orgID` — and never reads the org's own `room_members` row. The two `$lookup`s read *individual* membership rows and *sibling org* rows, and the sibling lookup explicitly excludes the org being removed (`$ne` against `orgID`, line 446). Deleting the org row in `DeleteRoomMember` therefore has no effect on what the pipeline returns.

So on the redelivery you describe, `accounts` is re-derived identically and non-empty, the subs are already gone, and `DeleteSubscriptionsByAccounts` reports `deletedSubs == 0`. That fails `int(deletedSubs) == len(accounts)` and falls into the `default:` branch — `ReconcileMemberCounts` — which is the authoritative recompute. The `len(accounts) == 0` branch is not reached.

Moving `DeleteRoomMember` after the count handling would be a no-op for this scenario, and the suggested "reconcile when `accounts` is empty" would reintroduce the two full-room `CountDocuments` on exactly the no-op org removal this PR exists to make cheap.

---

PR #510, opened 2026-09-21
message-worker: fold the replier's lastSeenAt advance into their subscription upsert

hmchangw:
Part 4 of 4 splitting #331, re-applied onto current `main`. Independent of the other three, but it edits the same block of `handleSubsequentThreadReply` as #508 and #509 — see **Interaction with #509** below, which needs a small deliberate follow-up rather than a pure rebase.

## What changed

Every thread reply did two writes to `thread_subscriptions` for the replier: a `$setOnInsert` upsert of their subscription, then a separate `$max` on `lastSeenAt` (replying implies they read up to their own reply, #396).

`UpsertThreadSubscriptionAdvancingLastSeen` does both in one write. `lastSeenAt` is owned exclusively by `$max` and never appears under `$setOnInsert`, so the two operators do not conflict: a new subscription is seeded with `lastSeenAt=at`, an existing one only moves forward.

Only the replier's write folds. The parent author's subscription keeps the plain `UpsertThreadSubscription` — this reply is not theirs to have read.

`handleSubsequentThreadReply` reports whether it advanced the replier, and `processMessage` runs the standalone `$max` only when it did not: first reply, migration, self-reply and system messages still take that path.

## The race, and why the recovery is narrow

The upsert filters on `(threadRoomId, userAccount)`, which carries a unique index, so two concurrent replies by the same account in one thread can both miss the filter and both attempt the insert. The loser's `E11000` is recovered by replaying the `$max` alone rather than NAKing the reply.

If the replayed `$max` matches nothing, the duplicate came from a *different* unique key — an `_id` collision, say — and the original error is returned. An earlier revision of this change swallowed any duplicate key, which dropped such a write silently; that is what the `MatchedCount == 0` branch exists to prevent, and it has its own integration test.

## One failure mode changes

On the hot path `lastSeenAt` used to be a best-effort write whose failure was logged and swallowed. It is now part of the subscription upsert, whose failure NAKs the reply. Redelivery replays both. There is a unit test pinning this (`combined upsert failure NAKs the reply`).

## Interaction with #509

#509 drops the parent author's re-upsert. This PR keeps `UpsertThreadSubscription` in `ThreadStore` precisely because the parent still uses it. **If both merge, `UpsertThreadSubscription` loses its last `message-worker` caller and should be removed from the interface** (a mock regeneration) — that is a one-line follow-up on whichever rebases second, not a conflict to resolve blindly. `inbox-worker`'s same-named method is a separate interface and is untouched either way.

## Testing

- `make test` (full suite, race), `make lint`, `make sast-gosec`, repo-owned semgrep rules: green. `make generate` produces no diff.
- Unit tests cover the fold (the `at` argument is the reply's own `createdAt`, and the folded write is the replier's, not the parent's), the standalone `$max` still running on migration, and the new NAK-on-failure behaviour.
- New integration tests cover insert-seeds-lastSeenAt, advance-without-clobbering, never-moves-backward, an 8-way same-account concurrent race (no errors, one document, newest `lastSeenAt`), and the non-race duplicate key surfacing.
- Integration tests need Docker and could not run here; CI is the check on them.

## Scope

`message-worker` only. Not a `chat.user.*` handler, so no `docs/client-api.md` change. No `pkg/model` change.

https://claude.ai/code/session_01NebGr7Ge5ae3f4A5Ga317z

---
_Generated by [Claude Code](https://claude.ai/code/session_01NebGr7Ge5ae3f4A5Ga317z)_


hmchangw:
@coderabbitai review

Re-requesting on the rebased head (`56b186e8`) — this branch was rebased onto `main` after #507 and #509 merged and force-pushed as a single commit. You reviewed this exact head already with no actionable comments; this is a second pass at the author's request, focused on the merge resolution:

- The replier block #509 refactored into `writeReplierThreadSub` is where the `lastSeenAt` fold now lives, and both its callers set `replierLastSeenAdvanced`.
- `healParentThreadSub` deliberately keeps the plain `UpsertThreadSubscription` — the parent author did not write this reply, so it is not theirs to have read.

---

hmchangw:
@coderabbitai full review

Second pass requested by the author on the rebased head `56b186e8`, which was rebased onto `main` after #507 and #509 merged. The incremental review already covered this commit cleanly; `full review` is to re-examine the whole changeset rather than the increment — in particular the replier block, where #509's `writeReplierThreadSub` refactor and this PR's `lastSeenAt` fold had to be combined by hand.

---

